PR #306, opened 2015-02-01
fix Qobj division unit test

jrjohansson:
Attempt to solve issue with false errors in Qobj division tests on 32 bit system by using comparison that is more robust to floating-point errors.


nonhermitian:
Seems that this is merge ready.

PR #311, opened 2015-02-03
Remove call to fpmath

nonhermitian:
-This should not affect much of anything on x86 systems, and has no
effect on x64.

-Removes incompatibility with ARM systems.


jrjohansson:
+1 for removing ineffective compiler flags.

Neutral about implied/express ARM support (fine if there is no additional maintenance burden, probably not worth it otherwise).

nonhermitian:
I will merge this now.

PR #312, opened 2015-02-04
Timeslot attribute and symplectic evolution

ajgpitch:
Some dynamics attributes set when assigning timeslot durations directly
Fix to symplectic evolution


qutip/control/dynamics.py
```diff
@@ -314,6 +314,10 @@ def init_time_slots(self):
         if self.tau is None:
             self.tau = np.ones(self.num_tslots, dtype='f') * \
                 self.evo_time/self.num_tslots
+        else:
+            self.num_tslots = len(self.tau)
+            self.evo_time = sum(self.tau)
```
jrjohansson: If self.tau isn't very short, use np.sum instead of Pythons builtin sum.

ajgpitch:
It could be pretty long. Making change now

PR #313, opened 2015-02-04
Qtrl test out directory auto-create sync fix

ajgpitch:
Some issues occurred when multiple processes were checking to see if the test_out_dir needed to be created
This is now handled specifically

It would seem that my previous pull request is merged into this one. I hope that does not cause issues.


jrjohansson:
OK, no problem. Usually it is better to start new PR from branches that are created from master rather than another PR branch, but in this case it doesn't make much difference since the other PR is already merged now.

jrjohansson:
This looks fine, but can you clarify what roles the output of data to files plays in this module? Usually in QuTiP we do not have solvers create files in the file system, but rather return a for example an object that contain all the data, so that it will be up to the user if the data should be written to the filesystem or used in some other way. I assume that this mode of operation is also possible in this module? What is the use-case you had in mind for writing data to the file system?

ajgpitch:
The original purpose behind the files was to aid debugging, firstly of the library, then of implementations. Now it turns out that some of the 'test out files' are useful in terms of analysis, e.g. the iteration log, and so I generally have this turned on. Often the optimisation can take a long time, and hence having the files written out during the process can be useful to monitor progress. Also, sometimes the processing is terminated early (by the scheduler on the cluster we use), and so having some filesystem output can mean that some kind of result can be extracted.

All this being said, I can see your point that some kind of 'run data' container object would be a better method, as the data could be used interactively. Perhaps users could then optionally chose to save this at the end or during the run, whichever suits them. I can add this to my todo list.

ajgpitch:
Missing indentation meant no files created

ajgpitch:
As mentioned in the commit comment, this latest change was simply a return to something that works. How it got broken is a mystery.
Regarding this pull request as a whole, the changes are just small fixes to some feature that has been in the library from the beginning. I have tried to tidy it up a few times, but I appreciated that more work is required. This I would prefer to do in some future pull request.
To be clear, the default behaviour does not write anything to the file system.

jrjohansson:
Ok, thanks, that makes sense. Merging this now.

PR #318, opened 2015-02-20
Fix gcc44 build

d-meiser:
gcc 4.4 does not support the `-flto` option for link time optimization. This is still a common compiler because it is the system gcc in centos 6 and redhat enterprise 6.

In this pull request I've removed that option. Users who need this could set the CCFLAGS environment variable to customize the compiler flags. Alternatively, we could add options to `setup.py` to enable or disable link time optimization.

I've also added additional rows to the travis build matrix to test building with gcc 4.4 and clang. In its current form this adds 4 additional rows to the build matrix (2 additional compilers, each tested with python 2.7 and 3.4). Let me know if you think this is excessive. It's possible to add just one python version for gcc 4.4 and clang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: python
 python:
     - 2.7
     - 3.4
+matrix:
+    include:
+        - python: 3.4
+          env: CC=clang CXX=clang++
 
 before_install:
     - uname -a
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
     - nosetests --with-coverage --cover-package=qutip qutip
 
 after_success:
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]]; then
+    - if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]] && [[ $CC == 'gcc' ]]; then
           coveralls;
       fi
 
--- a/qutip/control/setup.py
+++ b/qutip/control/setup.py
@@ -8,7 +8,7 @@ import os
 
 exts = ['cy_grape']
 
-_compiler_flags = ['-w', '-ffast-math', '-O3', '-march=native', '-flto']
+_compiler_flags = ['-w', '-ffast-math', '-O3', '-march=native']
 
 def configuration(parent_package='', top_path=None):
     # compiles files during installation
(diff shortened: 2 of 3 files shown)


nonhermitian:
Since I have been using GCC 4.8 for a while now, i did not think about compilers back to 4.4. I will take a look at this code on Monday, but my initial thinking is just to drop the optimization and leave it at that. This would save us from having to run an extra test suite on Travis. The only compiler flags that matter at the end are O3 and the fast math. Everything else is just a few percent improvement at best. 

Paul

d-meiser:
Sounds good. Would you like me to amend the pull request to get rid of the additional travis builds or at least reduce them? Might be nice to have a least one build with gcc 4.4 and clang.

nonhermitian:
Perhaps keeping just clang, if we wanted to keep one of the two, would be a good way to go. If we never mess with the compiler flags again there should be no problems.

d-meiser:
Ok, I removed gcc 4.4. The matrix is now:

- gcc 4.6 + python 2.7
- gcc 4.6 + python 3.4
- clang 3.4 + python 2.7
- clang 3.4 + python 3.4

Is that what you want or just one line for clang?

nonhermitian:
Yeah, perhaps we only need the py3.4 clang tests. Nothing that we do should require testing compilers against different Python versions. 

Paul

d-meiser:
Ok makes sense. I will make that change over the weekend. Btw for whatever reason one of the builds failed in the most recent build:

https://travis-ci.org/qutip/qutip/jobs/51588003

Any idea what the problem is? Is this something I should be worried about?

nonhermitian:
I can not imagine that that error would have much to do with a compiler change. Sometimes the tests fail for odd reasons and are fine if they get rerun. I will look into it tomorrow. 

Paul

d-meiser:
I changed the build matrix to

- gcc, python 2.7
- gcc, python 3.4
- clang, python 3.4

The compiler versions are gcc 4.6 and clang 3.4 (default on travis build machines).
For the latest commit all tests pass.

nonhermitian:
I think the only thing left to do is to modify the coveralls tests so that they run for only one of the Py3.4 tests.  Currently, the tests run if the Travis python version is 3.4.. As such, they are currently being run twice.

d-meiser:
I updated the pull request to post to coveralls only for one build permutation (gcc+python 3.4).

nonhermitian:
Ok, this looks good.  Thanks for you help!

PR #320, opened 2015-02-26
change version_table in ipynbtools to use deterministic output order

jfeist:
I've been annoyed from time to time that the output of the very useful version_table is ordered randomly. This fixes that by using a list of tuples instead of a dictionary (an OrderedDict would work equally well, of course).


nonhermitian:
This is a good fix.  Thank you.

PR #322, opened 2015-03-09
Possible fix for cython codegen errors

nonhermitian:
This seems to fix the issue reported by Vlad on the Google group.

This fixes the errors when using complex cmath header and typedef variables.  Variables must be of float, int, or complex, not np.kind_t


nonhermitian:
The possibility exists of issues on 32-bit systems. However, the reported error no longer occurs when checked using int, int32, int64, float, float32, float64, complex, and complex128 types.

PR #325, opened 2015-03-10
Fixed a bug in gradient computations in control module

fhenneke:
hi,
i think there was a bug in the computation of gradients for unitary problems. im not sure if i tracked them down correctly, but the gradients seem to be more consistent now. this can be checked with the following code after initialization of an Optimizer object `optim`.
```python
from scipy.optimize import check_grad
func = optim.fid_err_func_wrapper
grad = optim.fid_err_grad_wrapper
x0 = optim.dynamics.ctrl_amps.flatten()
check_grad(func, grad, x0)
```

also, im new to github and im not sure if im doing this correctly.

### DIFF
--- a/qutip/control/cy_grape.pyx
+++ b/qutip/control/cy_grape.pyx
@@ -3,11 +3,11 @@
 #    Copyright (c) 2011 and later, Paul D. Nation and Robert J. Johansson.
 #    All rights reserved.
 #
-#    Redistribution and use in source and binary forms, with or without 
-#    modification, are permitted provided that the following conditions are 
+#    Redistribution and use in source and binary forms, with or without
+#    modification, are permitted provided that the following conditions are
 #    met:
 #
-#    1. Redistributions of source code must retain the above copyright notice, 
+#    1. Redistributions of source code must retain the above copyright notice,
 #       this list of conditions and the following disclaimer.
 #
 #    2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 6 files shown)


ajgpitch:
Hi,
Thanks very much for your input. It looks like you are using github correctly to me, although I am no expert.
Well spotted on the typo, and thanks for sorting that out.
It looks like the tau factor on the propagator gradient is required - revisiting the paper by S Machnes (cited in the code). 
I am looking into your alternative method for the PSU normalisation now.
I am testing your branch locally as well.
I will report back when I have completed these tasks.

ajgpitch:
I have a made a few suggestion on the commit (see above).
I agree that this pull request seems to fix some problems with the unitary gradient computation. 
Thanks again for looking into this and doing the fixes.

nonhermitian:
I think we need some unit tests here verify that the gradient computation is doing what it is supposed to.

ajgpitch:
I agree, I have some plans to create some unit tests. The check_grad function that fhenneke has pointed out will be part of them.

I have tested the Frechet gradient methods (which are used by Lindbladian and symplectic dynamics) using the check_grad function and they seem to be fine.

ajgpitch:
From my perspective this is now ready to merge. 
fhenneke has fixed a bug in the gradient and a typo in a class name.
It seems like the last commit overwrote the previous one, and so some of the comments here have disappeared, however the queries to the code have been addressed.
I will look to add some unit tests (including the grad_check) over the Easter holiday, i.e. over the next few weeks.

nonhermitian:
Ok, merging now.  Looking froward to the unit tests.

PR #332, opened 2015-04-13
Stinespring dilations of type="super" Qobj

cgranade:
This feature starts work towards implementing diamond norms between quantum channels by providing Stinespring dilations of ``type="super"`` Qobjs. This requires generalizing the current ``to_kraus`` feature, since the left and right Kraus operators are the same only for CP maps, whereas the arguments to diamond norms are very often not CP (in particular, will often be a difference, || \Lambda_1 - \Lambda_2 ||_\diamond).

The pull request is currently in an incomplete state, so as to solicit feedback before solidifying the design. The docstrings are missing or are incomplete, Python 3 compatibility has not yet been tested, and I still need to make unit tests. Most importantly, the ``dims`` attribute of the input Qobj is not yet preserved, in lieu of solving #331.

PR #333, opened 2015-04-13
Fix #331

### LINKED ISSUE #331: Extra indices in dims break multiplication
In the development of a new feature (PR coming soon), I ran into a problem where adding new indices to ``dims`` can break ``__mul__`` in surprising ways. The full details are in [this notebook](http://nbviewer.ipython.org/gist/cgranade/8f7fce10b6deaf06ce01), but in summary, the bug causes multiplication of Qobjs with dimensions of the form ``[[n, 1], [n]]`` to not multiply correctly with Qobjs of dimensions ``[[n], [n]]``.

cgranade:
This commit should fix #331 and add extra tests to prevent regression. The new tests are of the form encountered in #332.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -63,6 +63,9 @@ from qutip.sparse import (sp_eigs, sp_expm, sp_fro_norm, sp_max_norm,
                           sp_one_norm, sp_L2_norm)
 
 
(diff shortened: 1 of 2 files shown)


nonhermitian:
Many thanks for this fix.

PR #336, opened 2015-04-19
optimized bloch_redfield_tensor routine

jfeist:
Mostly by replacing explicit loops with vectorized numpy expressions. For a simple test Hamiltonian with a single N-level bosonic operator, it is faster by a factor of ~40 for N=100 (Hamiltonian 100x100), i.e., it takes 12 seconds instead of 460 seconds. The speedup improves about linearly with N for this system.

I checked for some test systems that the results are the same as before. The only difference in interface is that c_ops is [] by default, and c_ops=None is not accepted anymore. This seems to me more in line with the rest of qutip, but could be changed back trivially.

### DIFF
--- a/qutip/bloch_redfield.py
+++ b/qutip/bloch_redfield.py
@@ -224,10 +224,10 @@ def bloch_redfield_solve(R, ekets, rho0, tlist, e_ops=[], options=None):
 
 
 # -----------------------------------------------------------------------------
-# Functions for calculting the Bloch-Redfield tensor for a time-independent
+# Functions for calculating the Bloch-Redfield tensor for a time-independent
 # system.
 #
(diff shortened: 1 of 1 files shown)


nonhermitian:
I seems there is a few issues with Nonetypes in for loops,  but otherwise this is a great speed up.   We never got around to optimizing this function nor the floquet solver (I think) and it is about time.

jfeist:
Sorry, I had forgotten to also commit the change of the default value for `c_ops` in `brmesolve`. The tests should pass now.

In any case, I now also changed `bloch_redfield_tensor` to accept `c_ops = None` as input.

nonhermitian:
Looks good.  Merging now.  Thanks again.

PR #337, opened 2015-04-20
Fixed basis transform in to_chi, added unit tests for regression.

cgranade:
In writing the documentation for #332, I noticed that the chi-matrix produced for several test channels was incorrect due to my placing the dagger in the wrong place. I apologize for that mistake--- this PR should fix it, and add unit tests that ensure that the chi matrix produced by to_chi agrees with that for some known Pauli channels and a unitary rotation.

### DIFF
--- a/qutip/superop_reps.py
+++ b/qutip/superop_reps.py
@@ -210,7 +210,7 @@ def choi_to_chi(q_oper):
     nq = len(q_oper.dims[0][0])
     B = _pauli_basis(nq)
     B.superrep = 'choi'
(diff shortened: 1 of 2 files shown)


ajgpitch:
Is this a necessary fix for #332? Should it be merged first?

cgranade:
Ah, got it. Anyway, this PR is orthogonal to #332, it's just that I only noticed this problem in trying to write test cases and documentation for the new functionality in that PR.

ajgpitch:
I am happy with this and will merge if there are no further comments

PR #338, opened 2015-04-22
CRAB Algorithm implemented in control library

ajgpitch:
An additional algorithm has been added. The existing GRAPE algorithm remains the default
The CRAB algorithm was developed at the University of Ulm
The main difference is that it reduces the number of optimisation variables by defining the control pulses by expansions of basis functions, where the variables are the coefficients. Typically a Fourier series is chosen, i.e. the variables are the Fourier coefficients.  Therefore it does not need to compute an explicit gradient. By default it uses the Nelder-Mead method for fidelity error minimisation. Further explanation is given in the comments.

Significant changes were required to many of the classes in order to support another alg option in a tidy way.


JonathanUlm:
Thanks for the nice implementation. Could you please add some references to the papers explaining the theory about the algorithm?
p.s. Hope GRAPE is not exiting ;)

JonathanUlm:
I have added a development notebook to test this pull-request.

ajgpitch:
I also added a development notebook to test this

nonhermitian:
This is looking good.  Not being a control man myself, I am enjoying looking over the code to see how this gets implemented.  I think that at some point, we might want to think about doing a few things differently to keep the overall structure and style similar to the rest of QuTiP.  But of course this is a matter of taste only.  In addition, I think that you might want to start writing a documentation section about these methods, both theoretical and the numerical implementation as well.  This, along with the demo notebooks, will go along way with getting people to understand what is going on, and as such, more people will likely use these tools.  Starting this summer, I am planning on moving all the QuTiP documentation over to iPython notebooks, so perhaps a notebook detailing the above mentioned stuff could be put together?

ajgpitch:
Thanks for the feedback. I am very keen to spend some time making the control modules more QuTiP like in style. In particular trying to utilise QuTiP elements, e.g. maintaining Qobjs throughout the code. I am however focussed on trying to get some results out using the library, so it will probably be a month of so before I can start on this. 
Re the documentation, there is some theory in the first of our notebooks example-pulse_optim-Hadamard, however it could be a lot more detailed and better presented. I should be able to make a start on a general overview of the control modules though, i.e. one notebook that outlines the theory and then explains the different functions and classes and how / when to use them. Is that the kind of thing you mean?
I was wondering about the auto-generated API docs, how it deals with inherited and overidden methods etc. Do you use some specific tool to generate these? Can you point me to some docs about how it works.

nonhermitian:
No worries.  Nothing is in a hurry.  Obviously getting some results for a paper should take priority.  The docs are generated via Sphinx, and the docs for the project are a good place to start.  We have our own repository for that here on github.  Like I said, most of the stuff will move to IPython, but we will keep the API docs in sphinx.

PR #341, opened 2015-05-15
New module for open quantum system solvers. Code created by Neill Lambert. 

anubhavvardhan:
Groundwork for a new module that solves system-bath problems for open quantum systems.
Code created by Neill Lambert.
link.aps.org/doi/10.1103/PhysRevA.90.032114


anubhavvardhan:
useless redundant code. creates so much of a problem.

@nonhermitian...merge it please, neill will add the rest

jrjohansson:
It would be much better to thoroughly code review this pull request before merging rather than merging sooner and make additional PRs later. You can keep working on the PR while getting feedback on it without merging now. Things like model names and API is better to sort out once and for all before the first merge. 

With this PR, I think the ossolve modeule name is not very appropriate. Most of qutip is related to open system solvers. I think rcsolve (reaction coordinate solver) is a more accurate name for this module. Also, depending on how much code you have in mind to include in this module, it might be more appropriate to put all of it in a file called rcsolve rather than creating a new module directory like you propose here.

Also, don't use underscores in module names (rc_model -> rcmodel or better yet rcsolve).

anubhavvardhan:
will do.

anubhavvardhan:
@jrjohansson ....feedback?

jrjohansson:
You should first separate the problem definition and the solver. Move the solve method to a rcsolve function, and rename the rcsolve class to RCSolveXxx where Xxx  is something that describes the problem (and possibly move this to an example notebook). Otherwise, nice progres!

anubhavvardhan:
@nonhermitian , @jrjohansson ....this good enough?

nonhermitian:
It seems like this routine is geared toward a specific problem, rather than a general input.  Regardless, I think a notebook explaining this method is required, including better examples on the usage.  This can be added to the documentation once it is converted to IPython notebooks.  A few things I noticed are given below:

- The output is not a list, but a Result class object.
- Why is it not possible to directly calculate expectation values other than the coupling term?
- Why are the default input values what they are?
- The print statement at the end should be optional depending on the setting of the debugger.
- The call to Odeoptions in mesolve needs to be user accessible.
- tlist should not be the last input argument and should probably not be a kwarg.
- The call to eigenstates should also have an optional argument to call the sparse eigenstates solver if needed.

nwlambert:
Hey guys, thanks for the feedback (this is the first time I have logged in to github in about a year!)..

Just some replies to Paul's comments.   The method is basically general, under certain assumptions about the bath spectrum density and bath temperature, so the 'system' can, in principle, be anything.  I will also do so minor tweaking so it can deal with multiple baths and correlated baths.  We will add some notebooks showing different examples and explaining when its valid later.  

>Why is it not possible to directly calculate expectation values other than the coupling term?

Yeah there was a list of expectation values at some point, but it disappeared while we were making edits.  Thanks for the reminder.

Thanks for the other points, we will fix the various issues.

nwlambert:
Despite discovering the awesome reduce() function in python, generalizing this stuff to arbitrary baths is kinda tedious, and not very practical, so we decided not to do it (I tried for 2 baths a while back and it runs out of memory pretty quickly, as you might expect!).   In any case, the hierarchy can do the same thing much more efficiently.

Anubhav is working hard on some example notebooks for this RC stuff

anubhavvardhan:
travis' virtual python env is failing, can someone please restart the ci build.

qutip/rcsolve.py
```diff
@@ -0,0 +1,158 @@
...
+    -------
+    output: Result
+        System evolution.
+    """
+    if options is None:
+        options = Options()
+    output = None
+    start_time = time.time()    
+
+    dot_energy, dot_state = Hsys.eigenstates()
```
nonhermitian: Should set sparse = eigen_sparse in call to eigenstates method.

qutip/rcsolve.py
```diff
@@ -0,0 +1,158 @@
...
+         +(sprepost(PsipreX, A)) + (-spost(PsipreX * A))
+         +(spre(A * PsipreEta)) + (sprepost(A, PsipreEta))
+         +(-sprepost(PsipreEta, A)) + (-spost(PsipreEta * A)))           
+
+    #Setup the operators and the Hamiltonian and the master equation 
+    #and solve for time steps in tlist
+    psi0 = (tensor(thermal_dm(N,nb), initial_state))
+    output = mesolve(H, psi0, tlist, [L], return_vals_exp, options=options)
+    end_time = time.time()
+    if calc_time is True:
```
nonhermitian: This might be better set using the Logger and the debug setting.  The user could do a time calculation outside of the function if they so choose to.  Or they could just go to debug mode if using the Logger. Another option would be to add the calculation time as a property of the Result object.

nonhermitian:
Where are the functions in the hsolve module used?

qutip/rcsolve.py
anubhavvardhan: missed that one

qutip/rcsolve.py
anubhavvardhan: you're right. i'll just remove that part, its not of any consequence anyway.

anubhavvardhan:
hsolve module will have a hierarchical solver. i'm still working on generalizing it more. the others functions are just auxiliaries. it should be done in a few days.

anubhavvardhan:
@nonhermitian @jrjohansson ...could you take a look.

nwlambert:
quick note;  we are still trying to decide about a few things, e.g., what to do about the units, i.e., whether to make hbar and boltzmann constant user definable parameters, or rescale the coupling and the cut-off frequency of the bath in a more sensible way to take them into account.  I guess the latter is the 'more sensible' option, but will need to do it carefully.

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+#    LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#    DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#    THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+
+#Author: Neill Lambert, Anubhav Vardhan
+#Contact: nwlambert@gmail.com
+
```
jrjohansson: please add an `__all__ = [ ... ]` that explicitly determines which symbols to export from this module. See e.g. mesolve.py for an example

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+import time
```
jrjohansson: please remove unused imports

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+#    LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#    DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#    THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+
+#Author: Neill Lambert, Anubhav Vardhan
+#Contact: nwlambert@gmail.com
+
```
jrjohansson: add `__all__ = ....`

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+import time
+import warnings
+import numpy as np
+import scipy.sparse as sp
+from numpy import matrix
```
jrjohansson: is this used?

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+from numpy import linalg
+from qutip import spre, spost, sprepost, thermal_dm, mesolve, Odeoptions
+from qutip import tensor, identity, destroy, sigmax, sigmaz, basis, qeye, dims
+
+def rcsolve(Hsys, Q, wc, alpha, N, Temperature, tlist, initial_state,
```
jrjohansson: Please use the same order of arguments as other qutip solvers as far as possible. For example: `mesolve(H, rho0, tlist, c_ops, e_ops, ...)` so you should at least have the signature `rcsolve(H, initial_State, tlist, ...` in the same order.

jrjohansson: Temperature should not have a capital T. use lowercase for variable names

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+        Coupling strength.
+    N: Integer
+        Number of cavity fock states.
+    Temperature: Float
+        Temperature. 
+    tlist: List.
+        Time over which system evolves.
+    initial_state: Qobj
+        Initial state of the system.
+    return_vals: list of :class:`qutip.Qobj` / callback function single
```
jrjohansson: usually we call this argument `e_ops`. any particular reason to not follow that convention?

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+        Single operator or list of operators for which to evaluate
+        expectation values.
+    eigen_sparse: Boolean
```
jrjohansson: use `sparse` instead of `eigen_sparse` (for consistency in the library)

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+        With options for the solver.
+     
+    Returns
+    -------
+    output: Result
+        System evolution.
+    """
+    if options is None:
+        options = Options()
+    output = None
```
jrjohansson: not needed

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+    
+    dot_energy, dot_state = Hsys.eigenstates(sparse=eigen_sparse)
+    deltaE = dot_energy[1] - dot_energy[0]
+    if (Temperature < deltaE/2):
```
jrjohansson: in other places in qutip we use `w_th` to denote a temperature in frequency units. `temperature` suggests, at least to me, that it is in unit K

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+        warnings.warn("Given temperature might not provide accurate results")
+    gamma = deltaE / (2 * np.pi * wc)
+    wa = 2 * np.pi * gamma *wc #reaction coordinate frequency
```
jrjohansson: space between operator and variable `wc`

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+    g = np.sqrt(np.pi * wa * alpha / 2.0) #reaction coordinate coupling
+    nb = (1 / (np.exp(wa/Temperature) -1))
+
+    #Reaction coordinate hamiltonian/operators
```
jrjohansson: space between # and R

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+    Nmax = N * 2        #hilbert space
```
jrjohansson: hilbert space? seems like an unnecessary comment

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+        warnings.warn("Given temperature might not provide accurate results")
+    gamma = deltaE / (2 * np.pi * wc)
+    wa = 2 * np.pi * gamma *wc #reaction coordinate frequency
+    g = np.sqrt(np.pi * wa * alpha / 2.0) #reaction coordinate coupling
+    nb = (1 / (np.exp(wa/Temperature) -1))
+
+    #Reaction coordinate hamiltonian/operators
+    Nmax = N * 2        #hilbert space 
+    dimensions = dims(Q)
+    a  = tensor(destroy(N), qeye(dimensions[1]))
```
jrjohansson: double white space between a and =.  please run the pep8 command on the file and correct pep8 violations

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+    xa = a.dag() + a
+
+    # decoupled Hamiltonian
+    H0 = wa * a.dag() * a + Hsys_exp
+    #interaction
+    H1 = (g * (a.dag() + a) * Q_exp)
+    H = H0 + H1
+    L=0
+    PsipreEta=0
+    PsipreX=0
```
jrjohansson: spaces around operators.

qutip/rcsolve.py
```diff
@@ -0,0 +1,152 @@
...
+
+    all_energy, all_state = H.eigenstates(sparse=eigen_sparse)
+    Apre = spre((a + a.dag()))
+    Apost = spost(a + a.dag())
+    for j in range(Nmax):
+        for k in range(Nmax):
+            A = xa.matrix_element(all_state[j].dag(), all_state[k])
+            delE = (all_energy[j] - all_energy[k])
+            if np.absolute(A) > 0.0:
+                if abs(delE) > 0.0:
```
jrjohansson: dont need both `np.absolute` and `abs`. pick one and use it consistently

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+from copy import copy
+from numpy import matrix
+from numpy import linalg
+from scipy.misc import factorial
+from qutip.cy.spmatfuncs import cy_ode_rhs
+from qutip import spre, spost, sprepost, thermal_dm, mesolve, Options, dims
+from qutip import tensor, identity, destroy, sigmax, sigmaz, basis, qeye
+from qutip import liouvillian, mat2vec, state_number_enumerate
+
+def _heom_state_dictionaries(dims, excitations):
```
jrjohansson: why not use `enr_state_dictionaries` from `qutip.states` ? this seems like a duplicate?

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+
+    for state in state_number_enumerate(dims, excitations):
+        state2idx[state] = nstates
+        idx2state[nstates] = state
+        nstates += 1
+
+    return nstates, state2idx, idx2state
+
+
+def _heom_number_enumerate(dims, excitations=None, state=None, idx=0):
```
jrjohansson: duplicate of `qutip.state.state_number_enumerate` ?

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+    Calculate cotangent.
+    Parameters
+    ----------
+    x: Float
+        Angle.
+    """
+    return np.cos(x)/np.sin(x)
+
+
+def hsolve(H, Q,gam, lam0, Nc, N, Temperature, tlist, initial_state,
```
jrjohansson: use an argument order that is consistent with `mesolve`. rename initial_state to rho0 or psi0 (whichever is more suitable)

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+    Function to solve for an open quantum system using the
+    hierarchy model. 
+
+    Parameters
+    ----------
+    H: Qobj
+        The system hamiltonian.
+    Q: Qobj
+        The coupling between system and bath.
+   gamma: Float
```
jrjohansson: alignment

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+        Bath cutoff frequency.
+    lam0: Float
+        Coupling strength.
+    Nc: Integer
+        Cutoff parameter.
+    N: Integer
+        Number of matsubara terms.
+    Temperature: Float
```
jrjohansson: `temperature` or something like `w_th`

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+     
+    Returns
+    -------
+    output: Result
+        System evolution.
+    """
+    if options is None:
+        options = Options()
+    
+    #Set up terms of the matsubara and tanimura boundaries
```
jrjohansson: space between # and start of comment (check pep8)

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+    for i1 in range(1,N):
+        ci = (4 * lam0 * gam * kb * Temperature * i1
+        	  * gj/((i1 * gj)**2 - gam**2)) / (hbar**2)
+        L12 = L12 + (ci / gj) * LD1
+
+    #Setup liouvillian
+
+    L = liouvillian(H, [L12])
+    Ltot = L.data
+    unitthing = sp.csr_matrix(np.identity(Ntot))
```
jrjohansson: `unitthing` -> `unit`

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+                if kcount>0:                                                
+                    ci =  (4 * lam0 * gam * kb * Temperature * kcount
+                    	   * gj/((kcount * gj)**2 - gam**2)) / (hbar**2)
+                    Lbig = Lbig + sp.kron(Ltemp,(-1j
+                    				 * (np.sqrt(nlabeltemp[kcount]
+                    				    / abs(ci)))
+                    				 * ((ci) * spre(Q).data
+                    				 - (np.conj(ci))
+                    				 * spost(Q).data)))
+                nlabel = copy(nlabeltemp)
```
jrjohansson: do you really need another copy here?

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+                    			     * abs(c0)))
+                    			  * (spre(Q)- spost(Q)).data)
+                if kcount>0:
+                    ci = (4 * lam0 * gam * kb * Temperature * kcount
+                    	  * gj/((kcount * gj)**2 - gam**2)) / (hbar**2)
+                    Lbig = Lbig + sp.kron(Ltemp,-1j
+                    			  * (np.sqrt((nlabeltemp[kcount]+1) 
+                    			     * abs(ci)))
+                    			  * (spre(Q)- spost(Q)).data)
+            nlabel = copy(nlabeltemp)
```
jrjohansson: i'm guession that copy isnt needed here either, since nlabeltemp was assigned to a copy and it has not been modified

qutip/hsolve.py
```diff
@@ -0,0 +1,306 @@
...
+    
+    r.set_initial_value(rho0big1, tlist[0])
+    dt = tlist[1] - tlist[0]    
+
+    for t_idx, t in enumerate(tlist):
+        r.integrate(r.t + dt)
+        for idx,element in enumerate(rhotemp):        
+            output[idx].append(r.y[idx])
+
+    return output
```
jrjohansson: Use a Result class, and store the output in Result.states or something similar

qutip/rcsolve.py
anubhavvardhan: is it required when there is only one function in the module?

qutip/rcsolve.py
anubhavvardhan: i think it is, if matrix_element is its subfunction.

qutip/rcsolve.py
jrjohansson: Yes, otherwise this module exports all its symbols, including imported modules

qutip/rcsolve.py
jrjohansson: matrix_element is a method in qutips Qobj, not related to numpy.matrix

qutip/hsolve.py
```diff
@@ -0,0 +1,310 @@
...
+    for i1 in range(1,N):
+        ci = (4 * lam0 * gam * kb * w_th * i1
+        	  * gj/((i1 * gj)**2 - gam**2)) / (hbar**2)
+        L12 = L12 + (ci / gj) * LD1
+
+    # Setup liouvillian
+
+    L = liouvillian(H, [L12])
+    Ltot = L.data
+    unit = sp.csr_matrix(np.identity(Ntot))
```
nonhermitian: Should use sp.eye(Ntot,format='csr')

qutip/hsolve.py
```diff
@@ -0,0 +1,310 @@
...
+    Lbig = sp.kron(unit,Ltot.tocsr())
+    rho0big1 = np.zeros((Nsup * Ntot), dtype=complex)
+
+
+    # Prepare initial state:
+
+    rhotemp =  mat2vec(np.array(psi0.full(), dtype=complex))
+
+    for idx,element in enumerate(rhotemp):
+        rho0big1[idx] = element[0]
```
nonhermitian: Is there a better way to do this using rho0big1 as a sparse matrix, perhaps in COO format?

qutip/hsolve.py
```diff
@@ -0,0 +1,310 @@
...
+            Lbig = Lbig + sp.kron(Ltemp,(-nlabel[kcount] * (kcount * gj)
+            	   			 * spre(unit).data))
+        
+        for kcount in range(N):
+            if nlabel[kcount] >= 1:
+            # find the position of the neighbour
+                nlabeltemp = copy(nlabel)
+                nlabel[kcount] = nlabel[kcount] -1
+                current_pos2 = int(round(state2idx[tuple(nlabel)]))
+                Ltemp = sp.lil_matrix(np.zeros((Ntot,Ntot)))
```
nonhermitian: Should again use sp.lil_matrix((Ntot,Ntot)) here to avoid creating dense matrix of zeros.

qutip/hsolve.py
```diff
@@ -0,0 +1,310 @@
...
+                Ltemp[current_pos, current_pos2] = 1
+                Ltemp.tocsr()
```
nonhermitian: In general COO is prefered over LIL since conversion to and from CSR and CSC is faster

qutip/hsolve.py
```diff
@@ -0,0 +1,310 @@
...
+        ci = (4 * lam0 * gam * kb * w_th * i1
+        	  * gj/((i1 * gj)**2 - gam**2)) / (hbar**2)
+        L12 = L12 + (ci / gj) * LD1
+
+    # Setup liouvillian
+
+    L = liouvillian(H, [L12])
+    Ltot = L.data
+    unit = sp.csr_matrix(np.identity(Ntot))
+    Lbig = sp.kron(unit,Ltot.tocsr())
```
nonhermitian: Ltot should already be CSR format

qutip/hsolve.py
```diff
@@ -0,0 +1,310 @@
...
+                    				 * spost(Q).data)))
+                nlabel = copy(nlabeltemp)
+
+        for kcount in range(N):
+            if ntotalcheck <= (Nc-1):
+                nlabeltemp = copy(nlabel)
+                nlabel[kcount] = nlabel[kcount] + 1
+                current_pos3 = int(round(state2idx[tuple(nlabel)]))
+            if current_pos3 <= (Ntot):
+                Ltemp = sp.lil_matrix(np.zeros((Ntot,Ntot)))
```
nonhermitian: Same thing.  Just create empty LIL matrix here.

qutip/rcsolve.py
```diff
@@ -0,0 +1,156 @@
...
+    H1 = (g * (a.dag() + a) * Q_exp)
+    H = H0 + H1
+    L = 0
+    PsipreEta = 0
+    PsipreX = 0
+
+    all_energy, all_state = H.eigenstates(sparse=sparse)
+    Apre = spre((a + a.dag()))
+    Apost = spost(a + a.dag())
+    for j in range(Nmax):
```
nonhermitian: This is going to be slow since you are doing loops over Nmax=N**2.  Could be pushed to Cython?

qutip/hsolve.py
nwlambert: in the end we use the dense form of this for the initial state in the ODE solver (the sparse copy is never used so will be removed).  The ODE solver complains if I use a sparse form for the initial condition, is there anyway around that?

nonhermitian: Ahh, no that is my mistake.  The solver does not allow for a sparse RHS.

nwlambert wrote:

qutip/rcsolve.py
```diff
@@ -0,0 +1,156 @@
...
+    deltaE = dot_energy[1] - dot_energy[0]
+    if (w_th < deltaE/2):
+        warnings.warn("Given w_th might not provide accurate results")
+    gamma = deltaE / (2 * np.pi * wc)
+    wa = 2 * np.pi * gamma * wc  # reaction coordinate frequency
+    g = np.sqrt(np.pi * wa * alpha / 2.0)  # reaction coordinate coupling
+    nb = (1 / (np.exp(wa/w_th) - 1))
+
+    # Reaction coordinate hamiltonian/operators
+    Nmax = N * 2
```
nwlambert: minor reminder for me and anubhav;  this should be    Nmax = N * dimensions[1][0]      and should appear below the definition of dimensions

jrjohansson:
This looks ready for merging to me.

anubhavvardhan:
I think so too.

PR #350, opened 2015-06-19
Qtrl load params list, Qobj and ndarrays

ajgpitch:
This small change allows parameter values of data type list, Qobj and ndarray to be given in parameter files.
As with the other data types supported, the set_param function looks at the existing data type for the object property to determine how to parse the value in the parameter (ini) file.
Lists are expected to appear in the file in a format that can be pass to list() as 
setattr(obj, attrib_name, list(eval(val)))
is the call made, e.g.
[Dynamics]
evo_times=[1,2,5,10]

Similarly with arrays the call is 
setattr(obj, attrib_name, np.array(eval(val), dtype=a.dtype)) 
e.g.
target = [[1, 1], [1, 1]]
The same goes for Qobj

### DIFF
--- a/qutip/control/loadparams.py
+++ b/qutip/control/loadparams.py
@@ -50,6 +50,8 @@ defined for that object
 import numpy as np
 from ConfigParser import SafeConfigParser
 # QuTiP logging
+from qutip import Qobj
+
 import qutip.logging as logging
 logger = logging.get_logger()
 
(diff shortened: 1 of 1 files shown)


ajgpitch:
I did this work some time ago, and so I did a git rebase in order to try and avoid any conflicts
I was advised this is the thing to do in such circumstances.
It seemed to go okay, some feedback on whether I did the right thing and did it correctly would be welcome

ajgpitch:
I am giving a tutorial on the control modules at the QUAINT workshop on 30 June. I would like to include using this feature. I thought I had made this pull request some time ago, but seems not.
It would be great if it could be merged before then. Sorry for the time pressure, but it is only a small change and I don't think anyone else is using load_params yet. Hopefully this will change when I promote this to many potential new users.

jrjohansson:
Looks fine to me.

PR #351, opened 2015-06-30
Add copy=True to csr_matrix() in Qobj constructor

kiuthed:
When csr_matrix(spmat) is called with a sparse matrix spmat, the
default argument is copy=False, which then returns a reference to
spmat rather than a copy.

This behavior is the cause of at least one other bug (involving Qobj
multiplication) I have found, which I will fix separately if this
behavior is the intended result.

This probably hasn't been identified before because in many places
qobj.data is replaced with new matrices following computations. Also
copying can happen implicitly if the sparsity is changed as discussed 
here:
http://stackoverflow.com/questions/16722316/scipy-sparse-matrices-copy-argument

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -185,14 +185,14 @@ class Qobj(object):
 
         if fast == 'mc':
             # fast Qobj construction for use in mcsolve with ket output
-            self.data = sp.csr_matrix(inpt, dtype=complex)
+            self.data = sp.csr_matrix(inpt, dtype=complex, copy=True)
             self.dims = dims
             self._isherm = False
             return
 
         if fast == 'mc-dm':
             # fast Qobj construction for use in mcsolve with dm output
-            self.data = sp.csr_matrix(inpt, dtype=complex)
+            self.data = sp.csr_matrix(inpt, dtype=complex, copy=True)
             self.dims = dims
             self._isherm = True
             return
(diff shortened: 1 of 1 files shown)


nonhermitian:
I see that this can be an issue if a csr matrix is passed to Qobj, the Qobj data is modified, and then the original csr matrix is needed again.  I think that if there is indeed a multiplication bug, and it is related to this issue, then including it in this pull is a good idea.  Then there are less things to keep track of

nonhermitian:
However, as you mentioned, the underlying sparse matrices are usually copied do to a change in sparsity, and therefore, the included change should be invisible to nearly all users.

kiuthed:
The issue also arose if a `Qobj` was passed to `Qobj`, which is where the multiplication bug came from. To be more specific, this is the bug I found:
```
from qutip import *
a = basis(2,0)
print a          # [1, 0]
b = Qobj(2.0)*a
print b          # [2, 0]
print a          # [2, 0] !!!
```
This happens because [the multiplication for size-1 Qobjs](https://github.com/qutip/qutip/blob/a76dc6032cb212d98e1485043796761bb758596f/qutip/qobj.py#L442) uses the `*=` operator, and so doesn't create a copy to put into `b`, which, because `Qobj()` doesn't create a copy either, changes `a`. Everywhere else I looked in qobj.py doesn't use an augmented assignment operator, which I guess is why this never came up before.

I was just going to change this multiplication command but it seems to me that the bug happened because `Qobj(other)` [is expected to return a copy](https://github.com/qutip/qutip/blob/a76dc6032cb212d98e1485043796761bb758596f/qutip/qobj.py#L201), so this pull request fixes that instead.

nonhermitian:
I think this makes sense overall, so merging now.

PR #352, opened 2015-07-17
Make Qobj().iscp compare to -settings.atol rather than zero

kiuthed:
This commit makes `Qobj().iscp` compare the choi matrix eigenvalues to `-settings.atol` rather than `0`. Otherwise it returns `False` for cases where some eigenvalues are negative due to numerical noise but should really be zero.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -1476,7 +1476,7 @@ class Qobj(object):
                     if self.superrep in ('choi', 'chi')
                     else sr.to_choi(self)
                 ).eigenenergies()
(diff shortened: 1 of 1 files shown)


ajgpitch:
This seems like a good change to me. I would be happy to merge if no-one has any objection

cgranade:
I think this makes a lot of sense, especially since ``Qobj.istp`` and ``Qobj.isherm`` both already rely on ``settings.atol``.

ajgpitch:
Sounds good to me. I will merge it now.

Issue #353, opened 2015-07-20
Bug in control/optimizer.py 


zerodynamics:
There appears to be a bug in OptimizerLBFGSB.run_optimization .  At line 901, _build_bounds_list is called and correctly constructs the list of upper and lower bound tuples and sets self.bounds, however there is no return value.  At line 939,  the empty variable 'bounds' is passed to scipy.fmin_l_bfgs_b and therefore the constraints are not active.  

Easy solution is to change "bounds=bounds" to "bounds=self.bounds" in line 939.

ajgpitch:
Good spot. A lot of changes were made to the Optimizer class when the CRAB algorithm, I think this slipped in then.  You will see on line 518 that the base class method passes the bounds as you suggest. Therefore the if you specify optim_method='l-bfgs-b' in any of the pulseoptim methods then your bounds should work and you will be using L-BFGS-B. 

I'll create a pull request with this fix now though. 
Thanks

ajgpitch:
There appear to be a few issues with the bounds. I will take a better look later or tomorrow.

zerodynamics:
In some spot testing, I found that making the proposed change (bounds -> self.bounds) sets the constraints correctly.

ajgpitch:
You'll see I had to make a coupled of other small changes to ensure the bounds passed in pulseoptim methods are applied to the Optimizer as well as the PulseGen

Thanks again for reporting this.

PR #354, opened 2015-07-21
Amplitude bounds now applied (again)

ajgpitch:
Recent changes to the Optimiser class had left the bounds only being applied for the CRAB algorithm
Bounds now applied for GRAPE using l-bgfs-b and fmin_l_bfgs_b

fixes issue #353


ajgpitch:
As there have been no comments to this, I am merging now

PR #357, opened 2015-07-29
Feature unitarity

cgranade:
This PR introduces a new metric, unitarity, based on [work with Wallman, Harper and Flammia](https://scirate.com/arxiv/1503.07865). In that work, we show that this metric is useful for discriminating between coherent and incoherence noise. As such, I hope that contributing an implementation of our new metric to QuTiP will be of use, especially for comparison to experimental data from randomized benchmarking. While it still needs some more documentation and test cases, I wanted to open the PR now to get some initial feedback.

### DIFF
--- a/qutip/metrics.py
+++ b/qutip/metrics.py
@@ -267,4 +267,6 @@ def unitarity(oper):
     u : float
         Unitarity of ``oper``.
     """
-    return np.linalg.norm(_super_to_superpauli(oper).full()[1:, 1:], 2)
+    Eu = _super_to_superpauli(oper).full()[1:, 1:]
+    #return np.real(np.trace(np.dot(Eu, Eu.conj().T))) / len(Eu)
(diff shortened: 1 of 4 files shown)


ajgpitch:
Seems like an interesting piece of work. I have asked for comments from members of our group here
In general it seems like a great idea to get new research into qutip, so that interested groups can test, compare, use etc.

cgranade:
I'm glad that it's useful! We wanted this work to be accessible, so I thought it made sense to contribute the definition. If the current design is acceptable, then I'll go on and finish the test cases and add a corresponding PR to qutip/qutip-notebooks demonstrating the new function.

ajgpitch:
I will check through the code over the next few days

ajgpitch:
Ok, I have check through the code and see no issues.
I am happy to merge unless there are any further comments

PR #358, opened 2015-08-04
Single-qubit Clifford group

cgranade:
This PR introduces a new generator function, qubit_clifford_group, that yields each member of the single-qubit Clifford group as a Qobj. This functionality should be useful in contexts such as quantum error correction and randomized benchmarking.

### DIFF
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -47,7 +47,7 @@ from qutip.tensor import tensor
 from qutip.states import fock_dm
 
 from itertools import product
-from functools import partial
+from functools import partial, reduce
 from operator import mul
 
 
@@ -718,10 +718,14 @@ def gate_sequence_product(U_list, left_to_right=True):
     return U_overall
 
 def _powers(op, N):
+    """
+    Generator that yields powers of an operator `op`,
+    through to `N`.
+    """
     acc = qeye(op.dims[0])
     yield acc
 
-    for _ in xrange(N - 1):
+    for _ in range(N - 1):
         acc *= op
         yield acc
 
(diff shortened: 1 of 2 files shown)


ajgpitch:
Thanks Chris,
This seems like a useful addition. I have taken a quick look at the code, and it seems like it is fairly independent, and so could be merged with little risk. The interface seems good, so I can't see any issues immediately. 

Are you planning to do a notebook example?

cgranade:
I made up a quick notebook and posted it as qutip/qutip-notebooks#20. In the meantime, I notice I forgot that Python 3 is missing ``reduce()``, having moved it to ``functools.reduce()``. Is it permissible to import ``functools.reduce`` conditioned on checking ``sys.version_info`` so that I can use ``reduce``? Thanks!

ajgpitch:
I have read a bit about the reasons for moving reduce. The main one seems to be that code using it can be unclear. And that in most cases a for loop can be used to achieve the same in a clearer way.  I think that may be applicable here.  I have spent a while looking at the line that uses it and I am still not sure what it is doing. I think sacrificing a little succinctness for clarity here may be beneficial.

ajgpitch:
I think that you may have issues with xrange in Python3 as well. I believe it was dropped, as range in Python3 does what xrange does in Python2. I think the qutip philosophy is that we should be developing for Python3, but backwards compatible. As in this case, using range will have a performance hit in Python2, but not Python3.

ajgpitch:
Perhaps just for my benefit you could explain why you have chosen to use generators rather than lists. As in what is the advantage in this case.

cgranade:
I always forget about ``xrange`` versus ``range``, sorry about that. Anyway, I used generators as opposed to lists as I didn't need the additional structure of lists, and as the Python 3 style has been to express as much as possible in terms of generators. For example, ``range``, ``zip`` and ``map`` are all generators now. In that spirit, by making ``qubit_clifford_group`` a generator, I don't assume that the user needs to remember the entire group, nor that the user wants to store the group in a list as opposed to a more specialized structure. This will be especially important, I think, should there be a need later to introduce a function that enumerates over multi-qubit Clifford groups. In that respect, the current choice is in anticipation of consistency with future features.

As for the question of ``reduce`` versus a triply-nested ``for`` loop, my intent there was twofold: first, to reflect the mathematical notation more closely and second, to allow for possible generalization to generating other groups. In L758, ``partial(reduce, mul)`` is a used to express products over the tuples generated by taking Cartesian products over powers of *E*, *X* and *S*. This approach should allow generalization such that similar groups can be written out in terms of this technique.

ajgpitch:
Thanks for the reply on generators. Seems like a good policy going forward.

Your implementation seems very neat. Thanks for explaining the thinking behind it.

Having looked a bit more into reduce. It seems that from Python2.7 (min req for qutip) reduce exists in functools. Therefore, I think you can import and use it with having to check version.

cgranade:
Thanks, I'm glad to be of service! I missed that about ``functools.reduce`` in 2.7, will amend the PR to use that, fix the ``xrange`` thing and try to make the use of ``reduce`` a bit more clear. Thanks!

cgranade:
Applied fixes for Python 3, and now the new unit tests run locally on both 2.7 and 3.4, so I expect that the CI checks should pass as well.

ajgpitch:
I am happy with the changes and explanations. 
I see the Python2.7 check has failed though. I am not sure what the error message implies. @cgranade do you have any idea?

cgranade:
That's very odd, it looks like the test was killed the test midway through, but I have no idea why. I'd guess that rerunning the build may work, but answers to [a question on StackOverflow](http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit) claim that you must have write access to the repo in order to trigger a new build.

ajgpitch:
Ok, after some fiddling I found I could restart the build. Lets see what happens

cgranade:
Thanks, and I'm sorry for the hassle. No idea what happened last time. Perhaps it's related to using the legacy container? If you think it's helpful, I can play with ``travis.yml`` to try and [get the migration working](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade).

ajgpitch:
No, its fine. You were right, it passed this time. I would be happy to merge this, we will wait a few days to see if any one else has comments.

cgranade:
Ah, great, thank you!

PR #359, opened 2015-08-11
__mul__ and __add__ leave isherm lazily-evaluated.

cgranade:
This PR changes ``Qobj.__mul__`` and ``Qobj.__add__`` to leave ``isherm`` lazily-evaluated (that is, ``_isherm = None``) unless the Hermicity of the result is already known from closure under addition. This change was prompted by noticing that ``isherm`` alone took up over half of the runtime of a computation I was attempting, despite that nothing actually ever looked at the Hermicity of a product of two ``Qobj`` instances. I also added additional tests to ensure that ``__mul__`` doesn't result in incorrect values of ``isherm``, something that was not previously covered by ``qutip.tests.test_qobj.test_QobjHerm``. The new tests pass locally on Python 2.7 and 3.4 (Ubuntu 14.10) as well as on Windows with Python 2.7.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -315,7 +315,9 @@ class Qobj(object):
             if isinstance(dat, (int, float)):
                 out._isherm = self._isherm
             else:
-                out._isherm = out.isherm
+                # We use _isherm here to prevent recalculating on self and
+                # other, relying on that bool(None) == False.
+                out._isherm = True if self._isherm and other._isherm else out.isherm
 
             out.superrep = self.superrep
 
@@ -425,7 +427,7 @@ class Qobj(object):
                 else:
                     out.dims = dims
 
-                out._isherm = out.isherm
+                out._isherm = None
 
                 if self.superrep and other.superrep:
                     if self.superrep != other.superrep:
--- a/qutip/tests/test_qobj.py
+++ b/qutip/tests/test_qobj.py
@@ -40,13 +40,20 @@ from qutip.qobj import Qobj
 from qutip.random_objects import (rand_ket, rand_dm, rand_herm, rand_unitary,
                                   rand_super)
 from qutip.states import basis, fock_dm, ket2dm
-from qutip.operators import create, destroy, num, sigmax
+from qutip.operators import create, destroy, num, sigmax, sigmay
 from qutip.superoperator import spre, spost, operator_to_vector
 from qutip.superop_reps import to_super
 from qutip.tensor import tensor, super_tensor, composite
 
 from operator import add, mul, truediv, sub
 
+def assert_hermicity(oper, hermicity, msg=""):
+    # Check the cached isherm, if any exists.
+    assert_(oper.isherm == hermicity, msg)
+    # Force a reset of the cached value for isherm.
+    oper._isherm = None
+    # Force a recalculation of isherm.
+    assert_(oper.isherm == hermicity, msg)
 
 def test_QobjData():
     "Qobj data"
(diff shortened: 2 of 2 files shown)


cgranade:
It looks like this CI check was killed prematurely again. I'm not sure what is happening with Travis lately, but I suspect it's related to their move over to container-based deployment.

ajgpitch:
Looks like a good improvement to me.
I have checked over the code and it looks good.

ajgpitch:
The second Travis restart passed.

cgranade:
That's good, thanks for restarting that! I'll submit my PR for migrating to Travis containers soon, it seems to be helping with the issue of tests being killed, though it's also slower to run so far.

As for the PR at hand, I'm glad it's useful. Thanks for checking over it!

ajgpitch:
I think that more slower, but more reliable, Travis CI tests, would be a definite benefit

jrjohansson:
+1

This looks like a good PR to me.

Just a quick note: We have had some side-effect issues when changing how isherm and other cached attributes are handled in the past, in particular with respect to performance. I used to monitor the performance over time with the continous-benchmarking suite in one of the respositories in the project, and that was sometimes useful to find unexpected performance hits. 

Hopefully this PR will only improve performance, but it could still be worthwhile to check that for example the unit test suite does not take longer time on this branch than on master.

cgranade:
Ah, that seems like a good check. Given the current instability of Travis timings, I'll run a quick check on my local machine to see how this PR affects performance.

cgranade:
Running on Python 2.7 (Ubuntu 14.10) on my local system, this change significantly increased the performance. For some reason, the change was less pronounced using Python 3.4 on the same system (under 1 second change), but in both cases, this PR resulted in test times that were at least as fast as master.

```
cgranade@berith:~/Dropbox/software-projects/qutip (master u=)$ time nosetests qutip
..............................................................................................SSSSSSS............................................................................................................................................................................................................................................/home/cgranade/Dropbox/software-projects/qutip/qutip/qobj.py:434: UserWarning: Multiplying superoperators with different representations
  warnings.warn(msg)
.............................................
----------------------------------------------------------------------
Ran 382 tests in 292.925s

OK (SKIP=7)

real    4m54.023s
user    11m20.584s
sys     0m5.356s
cgranade@berith:~/Dropbox/software-projects/qutip (master u=)$ git checkout fix-isherm-lazy 
Switched to branch 'fix-isherm-lazy'
cgranade@berith:~/Dropbox/software-projects/qutip (fix-isherm-lazy)$ time nosetests qutip
..............................................................................................SSSSSSS.............................................................................................................................................................................................................................../home/cgranade/Dropbox/software-projects/qutip/qutip/qobj.py:436: UserWarning: Multiplying superoperators with different representations
  warnings.warn(msg)
.............................................
----------------------------------------------------------------------
Ran 369 tests in 283.042s

OK (SKIP=7)

real    4m43.554s
user    10m51.428s
sys     0m5.080s
```

jrjohansson:
Perfect! Thanks for checking.

cgranade:
Ah, I hadn't noticed that, sorry, but yeah, that would explain it! I'll investigate today, then.

cgranade:
The issue turned out to be that I hadn't merged with upstream/master yet to include the unitarity feature, so those tests were not being included. With those added, ``master`` and this PR take about the same amount of time on Python 2.7— the variation between runs is larger than the difference. Sorry again for missing those tests before.

jrjohansson:
OK, no problem. Good to have verified that there are no regressions. I think this could be merged now.

PR #360, opened 2015-08-12
Travis migration

cgranade:
Pursuant to discussions over the build failures observed in #359, this PR migrates Travis CI support to their new container-based configuration. I've observed that the builds with this configuration are more reliable, but they are slower for some reason. In the [instructions](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade#Faster-builds) they provide, Travis say that this configuration should be faster, and to let them know if our build is slower; that might be worth doing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 
 python:
@@ -8,10 +9,16 @@ matrix:
         - python: 3.4
           env: CC=clang CXX=clang++
 
+addons:
+    apt:
+        packages:
+            - libatlas-dev
+            - libatlas-base-dev
+            - liblapack-dev
+            - gfortran
+
 before_install:
     - uname -a
-    - sudo apt-get update -qq
-    - sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran
     - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
     - chmod +x miniconda.sh
     - ./miniconda.sh -b


ajgpitch:
As mentioned previously, to me more reliable is a big benefit, slightly slower is not an issue.
Do @nonhermitian or @jrjohansson have an opinion on this PR?

ajgpitch:
I would be keen to merge this before the other merge-ready PRs, as I am having to restart the Travis build 2 or three times per commit at the moment

cgranade:
Makes sense, I only mentioned about the speed as I thought it might be of interest to Travis. That said, I totally agree, getting a reliable build is definitely important.

ajgpitch:
I agree, if Travis say that they are interested, then we should tell them. I would be happy for you to raise this with them, if you ok with that. It would seem to make sense if you have done the tests. I'll do it if you think that is more appropriate.

cgranade:
Sounds good, will do that soon then. Thanks!

jrjohansson:
+1

Container-based testing on travis is the way to go, and qutip should make use of it as soon as possible. The speed issue could probably be fixed later, if necessary, but might also very well be fixed upstream. 

BTW: Great work on this and other recent PRs!

cgranade:
I suspect it may already be getting fixed upstream as a recent merge on my fork completed testing in just about 10 minutes, quite on par with where it was. If it becomes important, we can use the new container infrastructure to start caching things like the Anaconda installer, but looking at the detailed breakdown, it didn't seem important enough to justify adding the extra complication.

Anyway, I'm glad to be of service!

PR #372, opened 2015-09-02
Python3 compatibility fix in control classes apply_params

### LINKED ISSUE #367: control objects apply_params() is not Python3 compatible
The apply_params method, which is implemented in many control classes uses the iteritems attribute of dict to iterate through the parameters passed as a dictionary.

A python 2 and 3 compatible method should be implemented to fix this

ajgpitch:
Use of iteritems removed in apply_params to make compatible with Python3
fixes #367 
Also some changes to the way max_metric_corr and accuracy_factor method_params are applied in the optimizer

### DIFF
--- a/qutip/control/dynamics.py
+++ b/qutip/control/dynamics.py
@@ -293,8 +293,8 @@ class Dynamics:
         
         if isinstance(params, dict):
             self.params = params
-            for key, val in params.iteritems():
-                setattr(self, key, val)
+            for key in params:
+                setattr(self, key, params[key])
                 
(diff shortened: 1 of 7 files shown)


ajgpitch:
If there are no comments to this, then I will merge soon

jrjohansson:
Looks good to me. Just a note: You can use `items`  for both Python 2 and 3 compatibility (gives a list in Python 2 and an iterator in Python 3, but works in both).

ajgpitch:
Thanks Rob.

Issue #377, opened 2015-10-05
Installation issues on MS Windows 8 and 10 64 bit 


ajgpitch:
A number of posts have appeared recently on the Google group reporting problems with installing QuTiP on MS Windows 8 or 10 64 bit editions.  I have also experienced problems when trying to install on a colleagues laptop. Mainly these seem to be compilation related issues.

Is there anything we can do to the code or the instructions to help avoid these issues?

Is there any value in creating an Anaconda install package?

jrjohansson:
The problem with windows has been that there was/is no free 64 bit compiler, and because of that it has been way easier to just install anaconda 32 and use mingw32 on windows (even on windows 64). 

In principle is should be possible build and install qutip 64-bit version on windows with a commercial 64-bit compiler (e.g. Visual Studio), but since few people has access to such a compiler we have not tested this option. Perhaps mingw64 is now a viable option, but I'm not sure how useful it is now. Would be great if someone could test it. 

Binary builds for 32 and 64 conda packages for windows would be highly desirable in my opinion. I have once created a conda package for qutip (https://github.com/conda/conda-recipes/tree/master/qutip) 
so if someone got the development environment with the required compilers setup for windows it should be straightforward to build the packages.

cgranade:
I've had good luck using the mingw package for 64-bit Anaconda, which installs the 64-bit version of mingw. I've detailed this in qutip/qutip-doc#22, but I haven't had the opportunity to test those instructions on a wide range of Windows systems.

ajgpitch:
I have tried on Win7x64 using anaconda. 
under Python2.7:
I found qutip would install, and most things would work. However there were many warnings when running tests and occurrences of python.exe crashing
under Python 3.5
I found qutip would install. However nothing much would run, e.g. issues importing spmatfuncs when creating a Qobj

cgranade:
I haven't run into that failure before, but I've only tried on Windows 8.1
and 10 so far. Do you have a console log for a python.exe crash?

ajgpitch:
Ok, here it is:
btw the actual error (the last line) did not appear in the output file, but it does appear in the cmd window, so I typed it in.

Transform 2-level to eigenbasis and back ... ok
Transform 10-level real-values to eigenbasis and back ... ok
Transform 10-level to eigenbasis and back ... ok
Transform 10-level imag to eigenbasis and back ... ok
Consistency between transformations of kets and density matrices ... ok
Check diagonalization via eigenbasis transformation ... ok
[...]

PR #379, opened 2015-10-12
Fix logging

cgranade:
This should at least partially address #377, as it seems that ``from __future__ import absolute_import`` isn't causing the built-in ``logging`` to appear in preference to ``qutip.logging`` on Windows 8. Given the hassles of calling both ``logging``, the idea behind this PR is to just rename and avoid conflicts that way.


ajgpitch:
I didn't realise that was causing a problem too. Seems like a good idea generally to try and avoid conflicts with built in names where possible.

jrjohansson:
Looks like a reasonable and practical solution to me.

Regarding the logging system in qutip, I think that it also would be worth making some changes to how it is used. As it is now, qutip does not really use the Python logging system in the way it is intended. The best way to use it would be to have each submodule that use logging use something like:

    import logging
    logger = logging.getLogger(__name__)

and then log at different levels using the logger instance, for example

    logger.debug("debug msg")

or

    logger.warning("warning msg")

etc.

The only utilities that would be required in qutip.logging_utils would then be to optionally create a stream or file handler for the "qutip" package or any of its submodules. This would allow selectively tuning logging for different submodules if requried, and there we not be any need for checking if qutip.settings.debug is set or to set logging levels explicitly in any submodules (as currently in done in the steadystate module for example).

Not saying that is should be done in this PR, but would probably be worth sorting this out at some point.

ajgpitch:
I really can't see how the changes here could be causing the Travis errors. @cgranade any ideas?

cgranade:
I looked at the failures and they seem to be in parts that are completely disconnected from logging. I have no idea why they aren't also in the main branch, but I'll investigate.

Anyway, I agree with you, @jrjohansson, that the current implementation is not in keeping with the design of Python's logging package. At the time, it was a bit of a hack to try and make it fit as well as possible, but I think in retrospect that may have been misguided. Do you think the refactoring to more appropriately use Python logging would be a good 3.2.0 feature, or would you suggest that for ≥ 3.3? Thanks!

ajgpitch:
I notice that I am also getting the same error on the travis build in the latest branch on my fork. An error that I do not get on my local machine. I will try a fresh conda env later and see if I can reproduce it.

ajgpitch:
The Travis error occurred because the latest numpy v1.10 did not allow some implicit cast from object to complex. I have fixed this in #380 
@cgranade  I guess we can merge that, then you can rebase from master (or similar process)

cgranade:
That makes sense, then, though I'm surprised that the version of numpy used by the Travis CI build suddenly changed. I'll merge in your fix on #380, at which point this really should work.

ajgpitch:
The Travis build seems to create a new conda env each time it runs the test. Consequently it downloads and installs the latest [conda] numpy, scipy etc. I tried creating a fresh conda env yesterday and found that the numpy version was higher than my qutip dev env, which I created a few months ago.

ajgpitch:
I am merging this now because I need to rebase my latest branch

PR #380, opened 2015-10-14
casting fix for correlation_2op_2t 

ajgpitch:
corr_mat casting made explicit
A cast from object to complex that was acceptable in numpy 1.9
was failing in numpy 1.10
This was causing an error in #379 Travis tests


ajgpitch:
It seems likely that there was some deprecation warning in numpy 1.9.

cgranade:
Given that this now passes successfully and is a fairly minor edit, any objections to merging in now so as to fix Travis CI builds? Thanks!

ajgpitch:
Go for it

PR #383, opened 2015-10-16
Qtrl unit tests

ajgpitch:
Units tests added for the qutip.control modules
Tests based on the notebook examples
Some other small fixes applied to issues found during the testing.


ajgpitch:
This resolves #383

PR #387, opened 2015-10-30
Fix #386 by changing axis for np.sum().

### LINKED ISSUE #386: tensor_contract fails for adjacent dimensions
This one seems to be due to a quirk of NumPy's fancy indexing, but ``tensor_contract`` fails to contract out a pair ``(i, j)`` if ``j == i + 1 and i > 0``. For example::

```python
In [1]: import numpy as np

In [2]: import qutip as qt

In [3]: dat = np.arange(2 * 3 * 3 * 4).reshape((2 * 3, 3 * 4))

In [4]: qobj = qt.Qobj(dat, dims=[[2, 3], [3, 4]])

In [5]: qt.tensor_contract(qobj, (1, 2))
<ipython-input-5-cfa6198a0fa3> in <module>()
----> 1 qt.tensor_contract(qobj, (1, 2))

C:\Anaconda\lib\site-packages\qutip-3.2.0.dev_8af111f-py2.7-win-amd64.egg\qutip\
tensor.pyc in tensor_contract(qobj, *pairs)
    421
    422     # Reshape back into a 2D matrix.
--> 423     qmtx = qtens.reshape((l_mtx_dims, r_mtx_dims))
    424
    425     # Return back as a qobj.

ValueError: total size of new array must be unchanged
```

cgranade:
This PR fixes #386 by changing checking if ``j == i + 1``, in which case NumPy doesn't automatically move fancy indices to the left. In particular, for adjacent indices, we should sum over index ``i``, not over index ``0`` as was assumed before. To catch regressions, this PR also adds more robust unit tests that check more unusual tensor contractions. Thanks to @arnelg for pointing this out!

### DIFF
--- a/qutip/tensor.py
+++ b/qutip/tensor.py
@@ -359,7 +359,9 @@ def _tensor_contract_single(arr, i, j):
     idxs = np.arange(arr.shape[i])
     sl = tuple(slice(None, None, None)
                if idx not in (i, j) else idxs for idx in range(arr.ndim))
-    return np.sum(arr[sl], axis=0)
+    contract_at = i if j == i + 1 else 0
+    print contract_at
+    return np.sum(arr[sl], axis=contract_at)
 
 
 def _tensor_contract_dense(arr, *pairs):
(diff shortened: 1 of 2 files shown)


cgranade:
This now seems to work in all cases, any objections to merging it in?

cgranade:
Since there haven't been any objections, I'm going to merge this in now.

PR #390, opened 2015-11-02
Generalized superop reps to more robustly handle multiqubit dims.

cgranade:
The previous implementation of ``choi_to_chi`` and ``chi_to_choi`` required that ``dims`` be specified explicitly as qubits; for example, ``[[[2, 2, 2], [2, 2, 2]], [[2, 2, 2], [2, 2, 2]]]`` for a three-qubit system. This PR generalizes to allow for ``dims`` built from powers of 2 instead, as in ``[[[8], [8]], [[8], [8]]]``. In addition, this PR adds new test cases to check that ``to_choi`` and ``to_chi`` properly roundtrip for multiqubit dimensions.

### DIFF
--- a/qutip/superop_reps.py
+++ b/qutip/superop_reps.py
@@ -52,8 +52,10 @@ from numpy.core.multiarray import array, zeros
 from numpy.core.shape_base import hstack
 from numpy.matrixlib.defmatrix import matrix
 from numpy import sqrt, floor, log2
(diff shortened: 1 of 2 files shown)


jrjohansson:
This looks good to me.

cgranade:
Sounds good, thanks. I'll merge it in now, then, so that the other PRs can use it for test cases.

PR #393, opened 2015-11-08
Add # of CPUs to about() and iPython table & fix Travis CI

nonhermitian:
Also updated the mac hardware_info func.


cgranade:
This looks like a good change. Not sure why it would cause Travis to not see ``conda`` at all, though, that's bizarre. Any ideas?

nonhermitian:
Not sure. Conda seems to be complaining about how the update is called. Maybe something has changed in conda. Need to check.

nonhermitian:
I have also fixed the Travis CI build settings.  It should have been done in another pull but I was lazy and just wanted something that works.  Since the latest Anaconda is now Python 3.5, I have switched to that version.  This also helps to address many depreciation issues.

cgranade:
Since this is now passing, I'll read over the commit soon and merge it in. Thanks!

PR #396, opened 2015-11-10
Ial time dependent drift

ial0:
@ajgpitch,

Could you take a look at these patches, with this set I've tried to keep this backwards compatible.

With this set, the helper function `create_pulse_optimizer` has been updated to accept the new drifts.
`optimize_pulse` currently creates an array of constant drift amplitudes, adding yet a another `PulseGen` parameter to an already excessive amount seemed counter productive.

There are no examples in this patch, that's something I'll look at if wanted.

Thanks
Ian

### DIFF
--- a/qutip/control/dynamics.py
+++ b/qutip/control/dynamics.py
@@ -145,7 +145,7 @@ class Dynamics:
 
     drift_dyn_gen : Qobj
         Drift or system dynamics generator
-        Matrix defining the underlying dynamics of the system
+        List of matricies defining the underlying dynamics of the system
 
     ctrl_dyn_gen : List of Qobj
         Control dynamics generator: ctrl_dyn_gen ()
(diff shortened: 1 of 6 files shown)


ial0:
I''d missed some backwards compatibility checking, hence some the tests were failing, this should be fixed now.

ajgpitch:
I would be more keen to keep the internal drift as either single operator (array / Qobj) or a list for when working with time-dependent drift dynamics. This would then mean that methods such as combine_dyn_gen would need to check if the drift is a list or single object.
Two reasons for this:
1. In the vast majority of cases the drift will be fixed, and in may be confusing for new users as to why it is a list
2. In cases where users set the drift directly using the dynamics object (like the test that you had to change) this change would not be backwards compatible.

ajgpitch:
An IPython notebook example should be provided for using this time-dependent option.
A test should also be added.

ajgpitch:
This was implemented in #506 instead

PR #397, opened 2015-11-11
Qtrl qobj propagation

### LINKED ISSUE #370: Quantum entities should be stored as Qobj rather than ndarrays internally
As the control modules were originally developed outside of qutip the initial and target states / operators are stored internally as ndarrays. The same applies for the timeslot propagators. 

Some users have wished to use the Qobj methods within their customised classes (e.g. some custom fidelity).

ajgpitch:
The internal data type for storing dynamics generators (Hamiltonians), propagators, and the time evolution can now be set as Qobj. When using Qobj the propagators are computed using qutip methods for expm.  Resolves #370
The default for small systems remains ndarray, as this provides much faster execution (factor of 10 approx). However, it is assumed, but not yet tested, that for larger systems the Qobj propagation will be faster, and certainly less memory consumptive.  The internal data type (if not specified) is chosen automatically using the same criteria Qobj.expm() uses to determine whether to use the (cython) sparse method.
The main motivation for implementing this change is for users developing their own custom fidelity and gradient measures which use the partial trace method. Consequently, preserving the dimensions is important. This has been shown (by colleagues testing) to be more efficient than recreating the Qobjs from ndarrays.
The internal operator processing methods and attributes have been renamed with underscores preceding to indicate that these will be of the internal operator data type. Accessors have been added that will always provide Qobj representations of the operators.
Theoretically any operator data type that supports the dot() method could be used, e.g. some other sparse matrix type, however this has not been extensively tested and should be considered an experimental feature.


ajgpitch:
Issue with setting phase_option #400 resolved by using property setter rather than set_phase_option
Took the opportunity to replace all get_ and set_ functions with property accessors.
Full list of function and attrib name changes provided in 'qtrl-v3_2-release_notes.txt'
These should only affect customised sub-classes. Deprecation warnings given.

PR #398, opened 2015-11-13
Fix ipynb parallel

nonhermitian:
The call to Client in ipython parallel has moved modules in V4.


nonhermitian:
This should be ready to go.

jrjohansson:
Looks good to me. Merging.

PR #399, opened 2015-11-19
fix use of hardware_info in __init__.py

jfeist:
When `qutip.settings.num_cpus==0` in `__init__.py`, the `qutip.hardware_info` module was replaced with the `hardware_info` function through `from qutip.hardware_info import hardware_info`. This caused problems in `ipynbtools.version_table`, which tries to call `qutip.hardware_info.hardware_info()`.


nonhermitian:
Looks good to me.  Thanks for the fix.

PR #404, opened 2015-12-06
Allow superoperator as inital condition to mesolve

### LINKED ISSUE #384: mesolve: allow superoperator as initial condition
Hi all,

There are quite a few applications when it is convenient to solve the Lindblad master equation for a superoperator, i.e.,

equation

where the last equation shows how the density matrix is found from the time-propagator (i.e., the superoperator solution to the Lindblad equation). The initial condition to the first equation is typically the identity superoperator.

As an example, I've typically come across the need for this when dealing with non-Markovian problems. Certain non-Markovian problems can be dealt with using time-propagators of the above type where the Lindblad evolution is in an "enlarged" Hilbert space of some sort. In fact, to give some motivation for implementing this feature, I am currently working on a sub-package for qutip with a few solvers for non-Markovian problems, one of them being the so-called "Transfer Tensor Method" (TTM) [PRL 112, 110401 (2014)], which takes as input a list of precomputed dynamical maps in the form of superoperators. A typical usecase could be to generate these with a (costly) Lindblad equation in a large Hilbert space, and use the TTM to extrapolate to larger times. My code is not quite ready for a pull request yet (due to this issue amongst other things), but when it is I would also like to discuss the structure of such a package. I guess it is best to leave that to when I have a pull request ready though.

For now I have a local modified version of the `_mesolve_const` function that allows a superoperator in place of `rho0`. The necessary modification is actually only to the function implementing the "right hand side" (RHS) of the differential equation, as this now needs to implement a matrix-matrix multiplication instead of a matrix-vector multiplication. For illustration purposes I will attach a code for a modified `_mesolve_const` below, that does the job, but the RHS is not implemented in cython.

I think implementing this feature in qutip should be relatively straight forward, but possibly a bit tedious since it there are so many different dispatch functions from `mesolve` depending on the different types of time-dependence etc. One would also need cython implementations of the RHS functions for superoperator evolutions.

Please let me know what you think about implementing this in qutip.

Here is the code for a function that propagates the Lindblad equation with a superoperator as initial value (written as a standalone module for illustration purposes). As you will notice in the code, there is a subtle issue with needing to transpose the Lindbladian to get the correct result. I'm not sure why this is the case.. must be something with the superoperator representation. Perhaps @cgranade knows?

````python
import scipy

from qutip.mesolve import _generic_ode_solve
from qutip.superoperator import liouvillian, mat2vec
from qutip.settings import debug

if debug:
    import inspect


def _mesolve_const_super(H, E0, tlist, c_op_list, e_ops, args, opt,
                   progress_bar):
    """
    Evolve the super-operator `E0` using an ODE solver, for constant 
    Liouvillian
    """

    if debug:
        print(inspect.stack()[0][3])

    """
    #
    # check initial state
    #
    if isket(rho0):
        # if initial state is a ket and no collapse operator where given,
        # fall back on the unitary schrodinger equation solver
        if len(c_op_list) == 0 and isoper(H):
            return _sesolve_const(H, rho0, tlist, e_ops, args, opt,
                                  progress_bar)

        # Got a wave function as initial state: convert to density matrix.
        rho0 = ket2dm(rho0)
    """

    #
    # check initial value
    #
    if not E0.issuper:
        raise TypeError("Argument 'E0' should be a super-operator")

    #
    # construct liouvillian
    #
    if opt.tidy:
        H = H.tidyup(opt.atol)

    L = liouvillian(H, c_op_list)

    #
    # setup integrator
    #
    initial_vector = mat2vec(E0.full()).ravel()
    # r = scipy.integrate.ode(cy_ode_rhs)
    r = scipy.integrate.ode(_rhs)
    # r.set_f_params(L.data.data, L.data.indices, L.data.indptr)
    # not sure why I need to transpose L here:
    r.set_f_params(L.trans().data)
    r.set_integrator('zvode', method=opt.method, order=opt.order,
                     atol=opt.atol, rtol=opt.rtol, nsteps=opt.nsteps,
                     first_step=opt.first_step, min_step=opt.min_step,
                     max_step=opt.max_step)
    r.set_initial_value(initial_vector, tlist[0])

    #
    # call generic ODE code
    #
    return _generic_ode_solve(r, E0, tlist, e_ops, opt, progress_bar)


def _rhs(t,y,data):
    ym = y.reshape(data.shape)
    return (ym*data).ravel()
````

arnelg:
This branch adds support for `rho0` argument to `mesolve` being a superoperator, to resolve #384. The functionality has been added to all the different mesolvers (but not the sesolvers).

Unit tests are added to tests/test_mesolve.py in a new class, `TestMESolveSuperInit`. There are tests for all the different mesolvers, and the result with `rho0` as a density matrix is compared to the result for `rho0` being the identity super-operator, by comparing the fidelity of the resulting states in each respective case. The reason the tests are a little bit slow compared to, e.g., the tests in `TestMESolveTDDecay` is the computation of the fidelity of the states at all the different times. If you think the tests are too slow, i can change them, and for example only check the fidelity of the final state.

As a side note: As I wrote this new functionality, I also ran some simple tests comparing a pure python implementation of the right hand side for `_mesolve_const` to the cython implementation (I did this because I was considering how much cython code I had to write, or if a pure python implementation of the new functionality would suffice). I actually found that a pure python implementation was as fast (or maybe slightly *faster*) than the cython implementation. The pure python implementation of the right hand side is simply (see lines 767-769 of mesolve.py)

````python
def _ode_rho_test(t, rho, data):
    # for performance comparison of cython code
    return data*(np.transpose(rho))
````

Here is a plot of the performance compare to the standard cython implementation for a very simple test. The benchmark was a damped-cavity JC-model with Hilbert space dimension 4 for the cavity, integrated up to "integration time". "run time" is the real run time on my laptop.

benchmark

Since there is no big difference between the cython implementation and the pure python implementation (if anything, the pure python implementation is slightly faster), maybe it makes sense to replace the call to `cy_ode_rhs` with a call to `_ode_rhs_test` (see lines 744 to 747 in mesolve.py)? This makes the code a tiny bit easier to read. Of course, this was a single, very simplistic test.

Sorry about all the merge commits. I'm still learning good workflow for this type of feature branch. Will use rebasing more in the future.

Let me know if you have any questions about the implementation.


arnelg:
Fixed some causes of travis tests not succeeding. In particular  e303b0e fixes use of `round` in the function `_td_wrap_array_str` in rhs_generate.py. This function uses `round` to get an integer number that is then passed as an array index. Since `round` returns a float, this causes the deprecation warning "DeprecationWarning: using a non-integer number instead of an integer will result in an error in the future". I therefore wrapped the call to `round` with a call to `int`. The travis tests now pass succesfully for this pull request.

arnelg:
I don't understand why the first commit to this PR is a modification of steadystate.py. Sorry, I think I might have messed up the history with some merges or something (still learning good work flow!). The master branch is merged in after this commit, and github tells me that the branch is up to date with the base branch. Please review this before accepting.

ajgpitch:
This generally looks fine to me. I would prefer to hear from some of the other developers about this too before merging.

ajgpitch:
I am no expert on git, but my approach re workflow is to rebase from the master before making a pull request if it has been a long time since I started the branch (or if there is some other reason to believe there may be merge conflicts).

If starting a new branch that relies on something from a branch I am already working on then I generate the branch from that one, e.g.
git checkout -b next_new_feature new_feature

arnelg:
Cleaned up some minor things in `tests/test_meoslve.py`, like removing trailing white spaces etc.

PR #405, opened 2015-12-09
nonmarkov subpackage for non-Markovian open quantum system solvers

arnelg:
This is the initial pull request for a subpackage for non-Markovian solvers, called `nonmarkov`. This idea was mentioned in issue #384.

The subpackage now includes two modules, with two quite different non-Markov solvers:

`transfertensormethod`: This module is an implementation of the "Transfer Tensor Method" (TTM) introduced by Cerrillo and Cao in [Phys. Rev. Lett 112, 110401 (2014)](http://journals.aps.org/prl/abstract/10.1103/PhysRevLett.112.110401) ([arXiv link](http://arxiv.org/abs/1307.7743)).

TTM is a completely general purpose approach to non-Markovian problems. It takes as input a collection of precomputed dynamical maps for a list of intial "learning times", and then tries to extrapolate the non-Markovian dynamics to larger times.

`memorycascade`: This module implements the method introduced in [Phys. Rev. Lett 115, 060402 (2015)](http://journals.aps.org/prl/abstract/10.1103/PhysRevLett.115.060402) ([arXiv link](http://arxiv.org/abs/1502.06959)) to integrate the dynamics of open quantum systems coupled to a coherent feedback loop with a time-delay. I decided to refer to this method as the "memory cascade" method in qutip. The method essentially maps the non-Markovian problem onto a cascaded chain of identical system, where each system in the chain represents a "memory" of a previous version of the system.

There is a companion pull request to the qutip-notebooks repo with two new notebooks documenting the usage of these two modules, qutip/qutip-notebooks#21:

examples/example-transfer-tensor-method.ipynb
examples/example-coherent-delayed-feedback.ipynb

Together with code docstrings, these two notebooks should give a good idea of how the modules are to be used.

As you will see in the notebooks, the TTM implementation has a user interface that closely mimicks the `mesolve` interface. The solver is called `ttmsolve` and has the following syntax (explained in more detail in the docstring and notebook):

````
ttmsolve(dynmaps, rho0, times, e_ops=[], learningtimes=None, tensors=None, **kwargs)
````
where `dynamps` is a list of pre-computed dynamical maps to learn from, `rho0` is the initial state, `times` are the times at which to find the solution etc.

The "memory cascade" implementation however, diverges a bit from this standard. In general I think it will be hard to have a kind of unified interface for different types of non-Markovian solvers, because unlike the Markovian case, there is no unified theoretical approach, and different solvers work in quite different ways. For the "memory cascade" implementation I chose a more object oriented style for the interface. The user first creates a `MemoryCascade` object, storing all the relevant parameters, and subsequently calls class methods to runs simulations. I personally like this style, and I think it is a general style that could be used across a great variety of solvers. I could however in addition write a function with a more `mesolve`-like interface. It can't, however, be identical, as there are several special arguments that has to be given. There is also a separate class method that computes output field correlation functions, and I kind of like that the various methods are collected together in a class, instead of having multiple different functions that all have to be called with a long list of arguments.

I therefore suggest the style used in the `memorycascade` module as a style that can be used for new solvers going forward. In general there will be a class with a name reflecting the solver, and this class has methods for doing various types of simulations.

The two non-Markov solvers I have implemented are of course closely related to my own line of current research. The "memory cascade" method is something I have developed, and the main motivation for me to implement the TTM was that it can be used very nicely in conjunction with the "memory cascade" method (see the notebooks). Going forward, it would be great to include more solvers. If anyone has suggestions for non-Markovian solvers they would like to see in qutip, we can also use this pull request as a forum for discussion.

### CODE REVIEW

**Comment by ajgpitch** [comment]
Looks generally good to me. Personally I am happy with the class instance method of invoking the code. However, as you say, the standard approach elsewhere in the library is to provide a function call. In the control modules I took the approach of doing both, with some functionality only being available through the setting of properties and calling instance methods.

I would prefer to here from some of the other developer before merging though, as this is more in their area than mine.

**Comment by ajgpitch** [comment]
The naming convention of test class methods seems to vary a lot throughout the library. Is there any reason why we do not stick with the usual pep8 naming convention in this area?

**Comment by ajgpitch** [comment]
It may be better to rebase this from the master to ensure that history is correctly preserved.

@cgranade  @jrjohansson @nonhermitian do we have a suggested method for checking this?

File: qutip/tests/test_mesolve.py [hunk 1]
```diff
@@ -451,5 +454,214 @@ def testMETDDecayAsArray(self):
...
+        expt = expect(a.dag()*a, out1.states)
+        actual_answer = 9.0 * np.exp(-kappa * (1.0 - np.exp(-tlist)))
+        avg_diff = np.mean(abs(actual_answer - expt) / actual_answer)
+        assert_(avg_diff < me_error)
+
+
+# average error for failure
+# me_error = 1e-6
+
+class TestMESolveSuperInit:
```
jrjohansson: test classes should go in qutip/tests folder

jrjohansson:
This looks really nice. Feel free to create a new API more suitable for the non-markov solvers. I agree with @ajgpitch that all names should follow pep8

jrjohansson:
I don't think there is any reason to worry about the commit history. You can rebase or squash the commits if you like, or just merge it as is. Git will likely eliminate many of the merge and related commits that do not contribute to the final diff.

arnelg:
For the tests I added I followed the naming convention already used in `tests/test_mesolve.py`. I guess we should try to change the naming convention in all the test modules then. Probably better to do this in another pull request though.

ajgpitch:
Many thanks for this contribution

PR #406, opened 2015-12-16
Function for plotting control pulses added to matplotlib_utilities

ajgpitch:
A simple line plot is produced that illustrates clearly that the pulse is piecewise constant in the timeslots
A standard Matplotlib line plot suggests that the amplitudes vary linearly throughout the timeslot, which is not the case

### DIFF
--- a/qutip/matplotlib_utilities.py
+++ b/qutip/matplotlib_utilities.py
@@ -185,3 +185,91 @@ class MidpointNorm(Normalize):
     def __call__(self, value, clip=None):
         x, y = [self.vmin, self.midpoint, self.vmax], [0, 0.5, 1]
         return np.ma.masked_array(np.interp(value, x, y))
+
+# Added 16 December 2015
+# by Alexander Pitchford
+def plot_ctrl_pulse(time, amps, ax=None, **plot_kwargs):
+    """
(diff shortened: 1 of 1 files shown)


ajgpitch:
This is a pretty simple addition that does not affect any other part of the library, so I am marking merge ready.

jrjohansson:
Have you looked at the step method of the Axes object in Matplotlib? I'm not sure what kind of graph this method produces but is sounds related.

qutip/matplotlib_utilities.py
```diff
@@ -185,3 +185,91 @@ def __init__(self, midpoint=0, vmin=None, vmax=None, clip=False):
     def __call__(self, value, clip=None):
         x, y = [self.vmin, self.midpoint, self.vmax], [0, 0.5, 1]
         return np.ma.masked_array(np.interp(value, x, y))
+
+# Added 16 December 2015
+# by Alexander Pitchford
```
jrjohansson: Author credits are available in the git history and does not need to be added explicitly in the files (because it doesnt scale when many authors work on a project)

ajgpitch:
Yes, with a bit of fiddling you can achieve exactly what I wanted in one line.
I will revert this out.
@jrjohansson thank you

Issue #413, opened 2016-01-12
nonmarkov package not in setup


ajgpitch:
The nonmarkov folder is not being installed during the setup.
I have done a fix that I will push now

ajgpitch:
Fixed by #414 
Can anyone explain why travis did not pick this up?

ajgpitch:
solved by #414

PR #414, opened 2016-01-12
nonmarkov added to list of packages

ajgpitch:
This is to fix issue #413 that caused new installations to be non-functional.
Anyone got any ideas as to why this was not picked up by the travis tests?


jrjohansson:
Looks good.

PR #417, opened 2016-02-02
Truncation of negative eigenvalues

cgranade:
This PR introduces a new method on``Qobj`` that forces an instance to be a valid density operator. This is useful when dealing with tomographic estimators that do not constrain to the space of density operators, such as least-squares fits. In particular, this PR allows for two different truncation algorithms: ``clip``, which is fast and straightforward, and the [SGS algorithm](https://dx.doi.org/10/bb76).

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -1185,15 +1185,54 @@ class Qobj(object):
 
 
     def trunc_neg(self, method="clip"):
+        """Truncates negative eigenvalues and renormalizes.
+
+        Returns a new Qobj by removing the negative eigenvalues
+        of this instance, then renormalizing to obtain a valid density
+        operator.
+
+
+        Parameters
+        ----------
+        method : str
+            Algorithm to use to remove negative eigenvalues. "clip"
+            simply discards negative eigenvalues, then renormalizes.
+            "sgs" uses the SGS algorithm (doi:10/bb76) to find the
+            positive operator that is nearest in the Shatten 2-norm.
+
+        Returns
+        -------
+        oper : qobj
+            A valid density operator.
+        """
+        if not self.isherm:
+            raise ValueError("Must be a Hermitian operator to remove negative eigenvalues.")
+
         if method not in ('clip', 'sgs'):
             raise ValueError("Method {} not recognized.".format(method))
 
         eigvals, eigstates = self.eigenstates()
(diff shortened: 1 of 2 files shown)


ajgpitch:
Looks good to me.
Should the test function have a docstring? I seem to remember these get reported during the testing, although maybe it's different when you use this yield method

cgranade:
That's a good catch, thanks. I'll look at how nostests handles docstrings in cases, but at the least, the test generator should have a docstring, I agree.

cgranade:
That turned out to be surprisingly hairy. This is a total hack, but hopefully one that will be useful in the future.

ajgpitch:
Yes, I see quite involved. Well, looks good now anyway.

cgranade:
I checked locally by intentionally making the tests fail, and the case descriptions do show up in failure logs correctly, so at least the hack works?

ajgpitch:
All looks good and ready to merge to me

cgranade:
Given that this PR worked fine on Travis CI before I merged it, I'm at a loss as to why all of the builds are now erroring out, much less at the linking stage. It's late here, so I'll try and look at it ASAP tomorrow. Thanks for your patience.

nonhermitian:
It is from the mcf90 fortran code.  Since we are testing using Anaconda, it is looking for the MKL library that does not exist.  Therefore, it has nothing to do with the current PR.  I will disable the mcf90 flag that is really worthless anyway since the fortran tests do not run.

nonhermitian:
After fixing the fortran issue here is the error:

======================================================================
ERROR: Test Qobj: trunc_neg works for non-diagonal opers.
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/qutip/qutip/qutip/tests/test_qobj.py", line 816, in case
    pos_qobj = qobj.trunc_neg(method=method)
  File "/home/travis/build/qutip/qutip/qutip/qobj.py", line 1246, in trunc_neg
    if eigvals[idx] + acc / (idx + 1) >= 0:
IndexError: index 2 is out of bounds for axis 0 with size 2

cgranade:
Weirdly, that error didn't show up in Travis CI for me before I merged. Working on diagnosing now, thanks for catching it.

PR #419, opened 2016-02-02
Rand updates

nonhermitian:
- Added random stochastic operators.

- rand_herm and rand_dm can be passed an array of eigenvalues that will be the eigenvalues of the resulting random operators. This is done via complex Jacobi rotations, but is really only suitable for operators with dimensionality <~=1000.

- Added unit tests.

### DIFF
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -42,7 +42,7 @@ The sparsity of the ouput Qobj's is controlled by varing the
 __all__ = [
     'rand_herm', 'rand_unitary', 'rand_ket', 'rand_dm',
     'rand_unitary_haar', 'rand_ket_haar', 'rand_dm_ginibre',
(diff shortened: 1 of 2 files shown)


ajgpitch:
Looks good to me.
You'll see one test has passed and another not. I am not sure what the difference is between these, and why one would fail and the other not?

nonhermitian:
That is a test for super operators that fails quite often.  It is unrelated to the current PR.  In fact, I think I will submit an issue to Chris about this particular test.

nonhermitian:
I am rerunning the tests that failed.  Should pass unless we get unlucky twice.

nonhermitian:
All good now.

cgranade:
I think this all looks good. As for the failing test, please see #420. I think that should do it.

nonhermitian:
Merging this.

PR #420, opened 2016-02-03
Fix for Liouville test

cgranade:
This PR changes the norm used in ``qutip.tests.test_superoper.testLiouvilleImplem`` to the max-norm. Previously, this test used the tr-norm, which scaled badly with the dimension of the Liouvillian under test, such that spurious failures were generated (see #419).


nonhermitian:
Works for me.  Merging this now.

PR #421, opened 2016-02-08
ENH: About updates

nonhermitian:
- hardware_info now detects # of physical cpus under Win.
- about() and version_table() now display BLAS type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ For instructions on how to install QuTiP, see:
 
 Demos
 -----
-A selection of demonstration notebooks is available here: [![Binder](http://mybinder.org/badge.svg)](http://mybinder.org/repo/qutip/qutip-notebooks)
+A selection of demonstration notebooks is available here: [![Binder](http://img.shields.io/badge/launch-binder-ff69b4.svg?style=flat)](http://mybinder.org/repo/qutip/qutip-notebooks)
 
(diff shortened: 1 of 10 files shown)


ajgpitch:
Generally looks like a really good idea.

Are either of these libraries that Python windows users would have installed anyway? Are they in the default Anaconda for windows for instance?

cgranade:
I only had ``comtypes`` installed for something else, but at least it's in conda and pip. Sadly, ``win32com`` is not in either repository.

ajgpitch:
I which case I think we should definitely use the comtypes as Chris suggests. And add it to the recommended packages in the install docs

nonhermitian:
Win32com is in win Anaconda. That is how I tested the code.

nonhermitian:
I will recode tonight

ajgpitch:
Well, I suppose it makes sense to try both then, and hope one gives an answer.

cgranade:
I didn't realize it was in current versions of Anaconda, sorry, as it
didn't use to be on 64-bit. My sincere apologies.

ajgpitch:
Is everyone happy with this now? I am quite keen to get a version out there which fixes the version_table.

ajgpitch:
@nonhermitian it seems this now has conflicts and cannot merged

nonhermitian:
This no longer has any conflicts.  However, as usual, the errors are generated by qutip.tests.test_subsys_apply as given in Issue #430.

PR #422, opened 2016-02-08
BUG: Anaconda 2.5 does not like -march=native

nonhermitian:
- Switched to -mtune=native

The latest version of Anaconda does not play well with the -march flag.  I am not sure if this is due to the use of MKL on all platforms, or if it is something else.  -mtune works fine though.

### DIFF
--- a/qutip/control/setup.py
+++ b/qutip/control/setup.py
@@ -8,7 +8,7 @@ import os
 
 exts = ['cy_grape']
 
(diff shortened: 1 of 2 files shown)


ajgpitch:
I suppose if anyone really knows what they are doing and they think that -march offers something, then they can always set it themselves.

Seems ready to merge to me.

cgranade:
Agreed, I think it makes sense to go on and merge to get Travis back up and running ASAP.

PR #428, opened 2016-02-13
Fix for #425

### LINKED ISSUE #425: Qobj trunc_neg causing Travis errors
The tests for this are sometimes erroring in Travis
See test result for #424 
ERROR: Test Qobj: trunc_neg works for non-diagonal opers.

I saw a conversation about this somewhere, but can't find it now

cgranade:
This PR fixes how ``Qobj.trunc_neg_eigs(method='sgs')`` handles eigenvalues equal to 0, which should resolve #425.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -1227,8 +1227,9 @@ class Qobj(object):
             
             acc = 0.0
             dim = self.shape[0]
+            n_eigs = len(eigvals)
             
-            for idx in reversed(range(dim)):
+            for idx in reversed(range(n_eigs)):
                 if eigvals[idx] + acc / (idx + 1) >= 0:
                     break
                 else:


cgranade:
Both this failure and the one before it are in unrelated tests. For instance, the most recent failure is in ``qutip.tests.test_fidelity2``, which rather surprises me as I'd have thought that test case was quite stable.

cgranade:
Similar test failures are affecting #429, with ``qutip.tests.test_subsys_apply`` now raising assertion errors.

ajgpitch:
For want of another idea I restarted the tests

cgranade:
Sounds good. I restarted once earlier and it failed on a different test, such that I think we should possibly create a new Issue to help track down which tests are randomly failing and why.

ajgpitch:
Yes, I was thinking the same thing. I will open an issue now.

cgranade:
Sounds good, we should definitely make sure that the tests all work for 3.2
final. In the meantime, I think we should continue to merge in fixes when
its clear that the test failures are unrelated.

PR #435, opened 2016-02-18
Increase subsys test tolerance

ajgpitch:
Some of the norms calculated in the test_ComplexSuperApply were above the tolerance of 1e-12
In 50 repeated tests 7 errors occurred. norms as large as 5e-11 were reported.
After increasing tol to 1e-10 test were repeated 200 times and no errors occurred


cgranade:
I think that makes sense. Tolerances of 1e-12 are a bit ridiculous to insist upon for a trace norm, since that scales with the dimension.

nonhermitian:
I guess this is also good to go.

Issue #433, opened 2016-02-17
test_SuperChoiChiSuper issues a UserWarning


ajgpitch:
UserWarning: Multiplying superoperators with different representations
issued during tests.
@cgranade I think you implemented this part. Is this something to concern
I mainly raised it as I have spent my evening removing all the other warnings from the tests, so would be nice to clear them all

cgranade:
Sorry, caught me while I was travelling to Albuquerque. I'll take a look ASAP, then.

ajgpitch:
I would not by any means call it urgent. More like something to look at some time

cgranade:
All the same, I'd very much like to further reduce warnings. I've literally had Firefox crash trying to load the entire Travis logs.

PR #436, opened 2016-02-18
Suppress superrep warnings when we know what we're doing.

cgranade:
This PR addresses #433 by manually setting ``Qobj.superrep`` to suppress warnings in ``chi_to_choi``. The change of basis for chi → choi is currently represented as a ``superrep='super'``, as that makes sense in several other contextx, such that spurious warnings were raised when performing this conversion.


ajgpitch:
Looks good to me

PR #440, opened 2016-02-20
ENH: Collection of minor updates (in progress)

nonhermitian:
A collection of updates that I can find while going through the docs and code.


qutip/qobj.py
```diff
@@ -953,12 +953,12 @@ def expm(self, method=None):
         out = Qobj(F, dims=self.dims)
         return out.tidyup() if settings.auto_tidyup else out
 
-    def checkherm(self):
+    def chec_kherm(self):
```
ajgpitch: This appears to be a type

ajgpitch:
Just a question: If we can use la.sinm and la.cosm in the tests, why not use it in the qobj method? Is the expm based method likely to be more efficient?

nonhermitian:
Yes that is a typo. But as the tests pass, it shows that that method is never used. Usually q.isherm does the job. Will likely depreciate it.

nonhermitian:
I guess the main use of internal methods are to maintain the qobj properties and sparsity. Although the expm is usually dense, one also could imagine cases, a diagonal mattix being the trivial example, where the result is sparse. We could have a flag for a dense mode like expm does

Issue #437, opened 2016-02-19
Qobj with dense data array


ajgpitch:
The Qobj data that represents the state or operator is stored as a sparse (csr) array. In most cases this makes sense, as the memory requirements for many solvers would be excessive. There are many cases though for small systems where memory would not be an issue, and significant performance improvements could be gained. This has been demonstrated in the optimal control modules. Many matrix operations are much faster in the dense representation.

Originally suggested in #26. This is now more urgent. In answer to #370 I submitted #397. However, it was found that significant performance degradation occurred in the control evolution calculation for small systems, and as such a retrofit of ndarray option for internal operator type was made. This leads to pretty untidy code, and means that subclasses for propagator and gradient calculations, fidelity, and evolution calculations all need to implement conditional methods.

Implementation of a dense Qobj would resolve these problems, and as mentioned above, could lead to performance improvements in many of the other solvers.

Many of the solvers assume sparsity, and so they too may need changes to support dense states and operators. 

In general, great care would need to be undertaken in implementing the dense Qobj. Those methods that did not support it would need to check the Qobj type when called

jakelishman:
Closed in `dev.major` by #1351 (and the gajillion other PRs that came before it). 🎉

PR #443, opened 2016-02-23
BUG: Qobj*vec crashes computer

nonhermitian:
- A Qobj times a dense vector would result in a recursive loop and
crash the computer.

- Made the multi array check better about handling dtypes and added
test.

The output is a dense array.  However if we do create a dense Qobj class (as described in #437) then that should be the returned object.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -458,7 +458,15 @@ class Qobj(object):
             else:
                 raise TypeError("Incompatible Qobj shapes")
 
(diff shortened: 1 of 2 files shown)


ajgpitch:
Yes, looks good to me. Think you should merge sooner rather than later.

btw these other tests have appeared due trying to make this conda-forge package. Looks like I will have to create some kind of solution file for the appveyor one to work. However, nice to see the circleci one worked without any specific config.

nonhermitian:
Ok, merging now

PR #446, opened 2016-03-01
state numbers made positive in state_number_enumerate

ajgpitch:
This may seem like a pedantic change, but is does (very slightly) improve performance in hsolve, and also avoids addressing an array index as a non-integer.

Seems to make sense for the state numbers to be integers anyway.


ajgpitch:
As there are comments on this I will merge now. 
I have tested extensively with the work I am doing on hsolve

PR #452, opened 2016-03-07
Fix for #451, new ishp attribute, new unittests to cover.

### LINKED ISSUE #451: Qobj.iscp returns false positive for non-normal operators.
If a ``Qobj`` instance represents a non-normal Choi matrix (that is, [A, A^T] ≠ 0), then ``Qobj.eigenenergies()`` can in some cases return all zeros, such that ``Qobj.iscp`` returns True, even though ``to_choi(...).isherm`` returns False. As a concrete example:

```python
In [1]: a = create(2).dag()
In [2]: S = sprepost(a, a)
   ...: print(S)
Quantum object: dims = [[[2], [2]], [[2], [2]]], shape = [4, 4], type = super, isherm = False
Qobj data =
[[ 0.  0.  0.  0.]
 [ 0.  0.  0.  0.]
 [ 0.  1.  0.  0.]
 [ 0.  0.  0.  0.]]
In [3]: print(S.iscp, to_choi(S).isherm)
(True, False)
```

cgranade:
This PR should fix #451 by adding an additional check to ``Qobj.iscp`` to check the hermicity-preserving condition. Since this is a useful condition in its own right, the PR also adds ``Qobj.ishp`` to expose that logic separately. Finally, new unit tests have been added to check for regressions on #451, and for the new ``Qobj.ishp`` attribute.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -113,6 +113,9 @@ class Qobj(object):
     iscp : bool
         Indicates if the quantum object represents a map, and if that map is
         completely positive (CP).
+    ishp : bool
+        Indicates if the quantum object represents a map, and if that map is
+        hermicity preserving (HP).
     istp : bool
         Indicates if the quantum object represents a map, and if that map is
         trace preserving (TP).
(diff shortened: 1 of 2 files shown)


ajgpitch:
Looks good to me.
Just one slightly cheeky question: you have edited a method, and created a new method with a FIXME in it. Were you not tempted to do the fixes as part of this PR?

cgranade:
I was tempted, but it's the same FIXME as currently affects both ``iscp`` and ``istp``. In particular, it's one that I think might take a bit of discussion to make sure it's addressed in a similar way to the caching strategy for ``isherm`` and now ``type``. If you think that's necessary for this to be a useful merge, though, I'm fine to go on and add caching of the three channel attributes.

ajgpitch:
No, I think you should merge now, I was just wondering. 
Makes sense I think to stage it. I seem to remember the caching of isherm had some teething troubles.

cgranade:
Sounds good, merging now, then. Perhaps after that, I should create an Issue, then, to track unified caching of the expensive attributes (``type``, ``isherm``, ``iscp``, ``istp`` and ``ishp``)?

ajgpitch:
Yes, sounds like a good plan

PR #454, opened 2016-03-07
Class for recording and reporting performance statistics of solvers

ajgpitch:
This class allows counts and timings (and general messages) to be collected about solver performance.
These can then be reported at the end.
I could not see anything similar already in qutip. I hope I have not duplicated effort.

I have been outputting some counts and timings in hsolve in order to look for efficiency improvements. They are a bit untidy at the moment, so I wanted a more tidy way to achieve the same. 

Hopefully this can be used in other solvers too. I intend to make the control stats subclass from this at some point.


ajgpitch:
I have not put any tests in for this, because it seems like a lot of effort for something that is used as a kind of testing tool anyway.  When I make my PR for the hsolve enhancements, including tests, this file will get pretty reasonable coverage.

cgranade:
This seems like a really good start, thanks! Building on this, we may want to explore taking a more unified approach to performance testing, as there's a number of performance-sensitive features in QuTiP. One avenue may be to write performance testing into the unit testing builds, that seems to be an [older extension for Nose](https://github.com/disqus/nose-performance), not sure if there's a better one out there.

ajgpitch:
I am sure that there probably is better third party tool out there. I didn't really look, not sure why.
I will be happy to swap to some other tool in the future. For now I will just use this

cgranade:
Sorry, that was definitely meant as a suggestion for something more in the future. Looks as though NumPy might have something built-in, though, so I'll look into that shortly.

ajgpitch:
I understood your meaning. I was just excusing my laziness / foolishness. I will happily swap to. or subclass from, something else, if you find something. I only realised I was probably reinventing the wheel near the end. I am just trying to reach some kind of milestone with hsolve before I leave Japan on Thursday, so rushing a bit

ajgpitch:
I moved the new class to the solver file. Seems like natural home for it, as in it has similar usage to the Result and Options classes

PR #455, opened 2016-03-10
Hsolve enhancements - now class module

ajgpitch:
Clarification of some variable naming to attempt to link with literature and also be usable without having to read the papers.
Switch to a class module. In line with recent work from @arnelg 
Moved to nonmarkov folder, as this is a non-Markovian approx method. Old file remains with deprecated function, just in case anyone was using it. This now links to the new instance method.
Many efficiency improvements (now up to 10x faster)

Test added. 

Will provide a notebook example soon. Then some documentation.


ajgpitch:
Just squashed the commits. I learned about this while doing the conda-forge thing (they insisted). If you don't already know about it, then it is definitely worth looking up. I had 16 commits on this, now its 1 - much more tidy.

ajgpitch:
@nwlambert please could you check the description I give in the class file. Maybe I should also add more references?

ajgpitch:
I need to fix (or hide) some warning given out by the integrator during the test. I will do that before I merge

nwlambert:
This looks great!   

The references are sufficient I guess.  We can add some more in the example notebook?

One minor terminology point in the explanation:  for the Drude-Lorentz bath the correlation functions can be exactly analytically expressed as an infinite sum of exponentials which depend on the temperature, and these are called the Matsubara terms or Matsubara frequencies.  

In general, when not dealing with a Drude-Lorentz bath, or something similar, we have to approximate the correlation functions by numerically fitting them to exponentials, but in that case they don't have any particular name I guess.  When we get around to adding it as an option, lets call them Fruchtman terms, as it was Amir Fruchtman who told me to try doing it that way!

ajgpitch:
Yes, we will add a reference to the your paper with the TLS when I submit a
TLS notebook example.
I think I will just make a simple example for now. And then add the more
involved one that compares the RC method later.

Thanks for the explanation on Matsubara terms (I might have to spelling to
correct :). I did not realise they were only referred to as such in the
DL spectral density. I will make some corrections now.

ajgpitch:
@nwlambert I have made the changes as suggested

ajgpitch:
The compulsive renaming continues.
I also reduced the number of time steps in the test to speed things up, and set to ignore the IntegrationWarning

qutip/hsolve.py
```diff
@@ -31,8 +31,10 @@
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 """
-This module provides exact solvers for a system-bath setup using the
-hierachical method.
+2016-03-9: This module has been deprecated.
```
jrjohansson: The qutip.hsolver has not yet been part of a release, right? Or was it already here in 3.1? If not, then you should not have deprecation message. Just deleting this module is fine. We dont guarantee api stability until after release.

ajgpitch: Good point. No, hsolve is new since 3.1. It's unlikely that many users
would be affected anyway
I will delete the old module and name the new one in nonmarkov 'hsolve'

ajgpitch:
Old module deleted as suggested by @jrjohansson 
Now using HEOM in naming generally, as this is in general use, and more descriptive than hsolve. Also fits with style in nonmarkov

arnelg:
Thanks, I didn't know we had an implementation of the hierarchical method :) Is there a notebook example for usage?

ajgpitch:
Yes, there are lots of notebooks. I just need to convert one to the new format and tidy it up a bit, then I will put on github - hopefully later today

PR #456, opened 2016-03-16
Diamond norm, dual channels and tensor swap.

cgranade:
This PR is very much still in progress, but adds three very nice new features inspired by the rather widespread interest in diamond norms (see, for example, pyGSTi, SchattenNorms.jl, QETLAB). Having them in QuTiP should help a lot. I've run some informal checks against other libraries, and the results returned by the new dnorm method agree pretty well; I'll write more formal tests soon.

Adding tensor swap and dual channel functionality is helpful in writing specialized cases for dnorm based on the observation that the diamond norm for positive maps is the same as the completely-bounded infinity norm (see [Johnston and Kribs](http://arxiv.org/pdf/1102.0948v2.pdf) and Watrous' notes). I hope to have this case added to dnorm soon as well, but I wanted to focus on making sure that the prerequisite features are included and well-tested.

Thanks to @ysanders for pointing out that CVXPY has come a long way recently. With that library in place, I think it's a good time to get dnorm() out there.


cgranade:
As usual, I forgot that ``map`` is an iterator in 3 and not a list. Will fix shortly.

cgranade:
I don't think that this is ready quite yet, but it's getting significantly closer. I want to add some more dnorm() tests for known cases, since the current known cases are all fairly trivial.

qutip/metrics.py
```diff
@@ -37,12 +38,26 @@
...
+import qutip.settings as settings
+
+import qutip.logging_utils as logging
+logger = logging.get_logger()
+
+try:
+    import cvxpy
+except:
+    cvxpy = None
+    logger.warn("CVXPY not found, semidefinite solving is disabled.")
```
jrjohansson: logger.warning

cgranade: Thanks for catching that.

cgranade:
I've now added more tests based on checking for agreement with the [QuantumUtils for MATLAB](https://github.com/CoryGroup/quantum-utils-matlab) library, since I'm familiar with that one, and as it has been tested for agreement with other implementations as well. I've written up the test cases in [this notebook](https://goo.gl/oWXhO9).

cgranade:
Now that the tests are fairly stable, I'd like to get the difference-of-unitaries case working, improve coverage a bit, and it should be ready for review.

cgranade:
This PR is now feature-complete, but needs a bit more debugging and test coverage to make a nice feature.

ajgpitch:
Looks very good to me.
What do you think about squashing commits? The conda-forge people insisted on it from me.
I am not really sure of the benefits, but it felt like a good idea.

There are lots of resources on the web explaining how to do it. In this case it would be:
`git rebase -i HEAD~27`
choose which commits to squash, and edit commit comments
`get push --force`

cgranade:
I'm glad you like! As for squashing, I think that can make sense, yeah. I'd like to hold off on doing so with this PR I have the current test failures fixed (seems as though the difference-of-unitaries optimization still needs a bit of work). After that, though, I agree that'd make for a much more legible commit history.

cgranade:
Since this branch is unfortunately rather old, I didn't want to squash the early commits, but the most recent batch have been squashed down to just one. I think that this feature is now in a state I feel comfortable with reviewing, so I'll add the label. In the future, I would like to generalize the difference of unitaries optimization, but I think that will take a bit of Cython work to implement.

cgranade:
Since this is now passing all Travis tests without complaint, I think it makes sense to go on and merge now.

PR #457, opened 2016-03-21
BUG: Correlation fixes

nonhermitian:
The correlation solvers were rewritten using args=None that is incompatible with the mesolver for time-dependent systems.  This fixes that by replacing with args={}.


nonhermitian:
Going to merge this now.

PR #458, opened 2016-03-26
Test updates

nonhermitian:
- Removed yield commands as they output to the console.

- Removed the lgmres steady state tests since this will be removed.

- Fixed test name going over 80 chars.


nonhermitian:
Merging this now.

cgranade:
Thanks for taking care of this, looks like some nice improvements. I had only originally added the generator-based tests to allow them to fail independently (this is especially useful for randomized tests), but I agree with cutting down the amount of extraneous console output. Is there a better way that I should use in the future that will allow for independent failure without inflating the number of tests reported?

PR #459, opened 2016-04-01
ENH: Intel MKL Sparse Enhancements

nonhermitian:
This will be a collection of functions that access the sparse features in the Intel MKL:

1) Parallel CSR SpMV

2) Parallel direct sparse LU factorization

I have yet to work out a nice way of importing the MKL library, so these functions will not work unless the path to the MKL is modified by hand.  My thinking is to look at the python version string to check for the Anaconda or Intel Python distros, and if found, find the MKL in the known locations for each distro and platform.  Intel said they would expose this functionality in a later version of their distro.


cgranade:
It looks as though NumPy keeps some metadata on MKL if it was compiled with MKL support. The ``numpy.__config__`` module in particular gives a path to the directory that NumPy loads its MKL libraries from. On my Linux box, for instance:

```python
In [1]: import numpy.__config__

In [2]: print(numpy.__config__.mkl_info['libraries'])
['mkl_intel_lp64', 'mkl_intel_thread', 'mkl_core', 'iomp5', 'pthread']

In [3]: print(numpy.__config__.mkl_info['library_dirs'])
['/home/cgranade/anaconda/lib']
```

The case seems to be worse for Anaconda on Windows (naturally), with ``C:\aroot\stage\`` taking the place of the actual Anaconda root:

```python
In [1]: import numpy.__config__

In [2]: print(numpy.__config__.mkl_info['libraries'])
['mkl_core_dll', 'mkl_intel_lp64_dll', 'mkl_intel_thread_dll']

In [3]: print(numpy.__config__.mkl_info['library_dirs'])
['C:\\aroot\\stage\\Library\\lib']
```

In the case that NumPy is compiled without MKL support, it seems to just omit the ``mkl_info`` dict entirely:

```bash
cgranade@berith:~$ source activate py27-nomkl
discarding /home/cgranade/anaconda/bin from PATH
prepending /home/cgranade/anaconda/envs/py27-nomkl/bin to PATH
cgranade@berith:~$ ipython
Python 2.7.11 |Anaconda custom (64-bit)| (default, Dec  6 2015, 18:08:32)
Type "copyright", "credits" or "license" for more information.

IPython 4.0.3 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import numpy.__config__

In [2]: print(numpy.__config__.mkl_info['libraries'])
<ipython-input-2-ebe4f6e784e1> in <module>()
----> 1 print(numpy.__config__.mkl_info['libraries'])

AttributeError: 'module' object has no attribute 'mkl_info'
```

nonhermitian:
This is a good find.  Previous versions of Anaconda did not show the correct dirs in numpy.__config__.  They used to point the the Intel compiler directory that built them.  This makes things easier than I thought.

cgranade:
I'm glad that helps, then! Perhaps ``sys.executable`` could be used to resolve ``C:\aroot\stage\`` for Anaconda/Windows?

nonhermitian:
I just checked the Intel distro.  Their numpy config does not point to the correct location for the libraries.  They point toward to compiler folders just like Anaconda did in the past.  Will stick to Anaconda support for now.

ajgpitch:
Is there anything in this parallel LU decomp that will potentially speed up expm?

nonhermitian:
Yes! In fact anything that uses LU decomposition should have a huge speedup.  Up to over 100x in some cases. This is why I am so interested. Also the parallel spmv should be useful in a number of places and dynamical solvers.

nonhermitian:
Switched this pull to QuTiP 3.2 has I am almost done with it, and it will be nice to have a speedup in the steadystate and correlation functions sooner than later.

nonhermitian:
This is finally ready to go.  Just a few comments:

- I have not done anything with the MKL spmv.  The cython version I have has much lower overhead.

- There are additional places that can benefit from this solver (e.g. expm, eigs,...).  These will be added later.

ajgpitch:
All looks very good to me.
There are a few extraneous empty lines in there.
There a lot of commits, do you think there is value in commit squashing?

nonhermitian:
Ok, the tests have been fixed so I am going to merge.  I will squash merge as suggested by Alex.

PR #461, opened 2016-04-06
Correct way of calculating Wigner function

Vutshi:
A new iterative method using Clenshaw summation is implemented. It is numerically stable and fast (2-3 times faster than 'iterative').

Test:

```
from qutip import *
import numpy as np

xvec=np.linspace(-10,10,101);yvec=xvec;X,Y = np.meshgrid(xvec, yvec)
psi=coherent(100,7.3*np.exp(1j*0.5))
rho0 = ket2dm(psi)
WW = wigner(rho0,xvec,yvec,method='iterative')
WWc = wigner(rho0,xvec,yvec,method='clenshaw')

plt.contourf(X,Y,WW,100,cmap=plt.cm.viridis)
plt.show()
plt.contourf(X,Y,WWc,100,cmap=plt.cm.viridis)
plt.show()
```

### DIFF
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -59,13 +59,16 @@ def test_wigner_coherent():
 
     # calculate the wigner function using qutip and analytic formula
     W_qutip = wigner(psi, xvec, yvec, g=2)
+    W_qutip_cl = wigner(psi, xvec, yvec, g=2, method='clenshaw')
     W_analytic = 2 / np.pi * np.exp(-2 * abs(a - beta) ** 2)
 
     # check difference
     assert_(np.sum(abs(W_qutip - W_analytic) ** 2) < 1e-4)
+    assert_(np.sum(abs(W_qutip_cl - W_analytic) ** 2) < 1e-4)
 
     # check normalization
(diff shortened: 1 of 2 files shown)


cgranade:
Thank you for your contribution, this looks like a nice feature. If you could please provide some unit tests as to its correctness, that would be greatly appreciated.

nonhermitian:
Denis,  Good to hear from you again.  It looks like a solid addition!  Great to have finally squashed this issue.  I ran your code through our unit tests in place of the iterative method and it passed.  We can just add additional unit tests similar to the ones we have now for the iterative method.

Vutshi:
I'm also glad to be back and contributing to the great project. I have accumulated some more stuff to add to stochastic module (later at some point).
Back to business. Maybe we should just change the default method to 'clenshaw' and then the current tests will test it instead of 'iterative'?

Vutshi:
Tests for the new method are added.

nonhermitian:
Can you go ahead and make the Clenshaw version the default Wigner method is this pull?  Then we should be good to merge.

nonhermitian:
Merging this now.  Many thanks!

Issue #462, opened 2016-04-08
Qobj expm


nonhermitian:
Alex has asked about having a faster expm method for Qobj objects.  Having looked into it a bit, here are my thoughts:

- We should check to see if the Qobj is diag, and if so just do the exp on the elements.  Otherwise do a dense expm.  This is really what matlab does. This will be done in parallel if connected to a good BLAS.

- I have yet to find a real-world usage case where the resulting expm is not a dense matrix.  Therefore, the above suggestion seems to be optimal.

- There is a bug in our expm code for a particular kind of matrix, this happens to be a particular diagonal matrix.

- There is a bug in the scipy.sparse.linalg.expm function that is even more serious, and was reported.

- Someone discussed using approximation methods for the expm in the control module.  This is an interesting idea in general, but we would need some way of quantifying the error in the approximation.  This has probably already been done in the literature somewhere.

ajgpitch:
I tried some experiments with elementwise exp with numpy to see if mkl did any parallel processing. It did not seem to.

scipy.linalg.expm claims to use the Pade approximation. There used to be a expm2 (that used diagonalisation) and an expm3 (method?) but these have been deprecated. The scipy docs claim that the method used in expm is "numerically more robust".

In the control modules we either use a eigendecomp (diagonalisation) method or scipy.linalg.expm_frechet as we typically need to compute propagators (which either uses the eigendecomp or is provided by the frechet deriv) . Frechet also uses some Pade approximation. I think the Pade method uses something like a LU decomp, but would need more research. It would always be more efficient to use the eigendecomp if it has been computed already.

My guess is that if parallel LU decomp provides much (x100) faster eigendecomp, then the diagonalisation method would be much faster than expm. However, some experimentation would be needed to verify. 

I am told that there are many methods of computing expm, and is there is much numerical research into it. It seems strange that scipy only offer us one method. The spinach control library (http://spindynamics.org/Spinach.php) reputedly has some interesting methods of computing expm.

A quick Google search found https://github.com/scipy/scipy/pull/354, which shows that scipy considered adding a method kwarg, but have parked the idea.

We currently have multiple methods in qutip based. I assume there must have been some motivation for writing our own sparse matrix functions for expm. It would seem that a pretty thorough investigation would be required determine to value of the various options. Would make a nice self contained project for someone.

nonhermitian:
One must set the number of threads manually for MKL in Anaconda:

import mkl
mkl.set_num_threads(N)

[...]

PR #463, opened 2016-04-08
ENH: Fast checking for diagonal sparse matrix

nonhermitian:
This is added in response to my comments on the issue of expm (#462). This might also be useful elsewhere.


ajgpitch:
Looks good to me

PR #464, opened 2016-04-13
FIXED: Expm

nonhermitian:
Updated expm function.  Fixes bug in previous version.  Defaults to sparse=False as resulting matrix is almost always dense, and the algorithm runs about 10x faster.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -928,21 +928,18 @@ class Qobj(object):
         else:
             return np.real(out)
 
-    def expm(self, method=None):
+    def expm(self, method='dense'):
         """Matrix exponential of quantum operator.
 
         Input operator must be square.
 
         Parameters
         ----------
-        method : str {'dense', 'sparse', 'scipy-dense', 'scipy-sparse'}
+        method : str {'dense', 'sparse'}
             Use set method to use to calculate the matrix exponentiation. The
-            available choices includes 'dense' and 'sparse' for using QuTiP's
-            implementation of expm using dense and sparse matrices,
-            respectively, and 'scipy-dense' and 'scipy-sparse' for using the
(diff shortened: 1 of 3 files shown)


ajgpitch:
Very nice work Paul. 
I was kind of expecting to see use of the new is_diag that you implemented recently. Is it there and I have missed it?

nonhermitian:
I was kind of expecting to have it there until I spaced it out.  Will add
tonight.

ajgpitch:
Looks good to me

nonhermitian:
Well, there are still some issues with the tests not passing. Probably due to the 1e-12 tol there.

nonhermitian:
Ok, this is good now.  I tried to get away with using a lower Pade approx. but I guess 13th order is the way to go.

cgranade:
Looks good to me as well, looks like a great enhancement!

nonhermitian:
Merging this now then.

ajgpitch:
Yes, good work.
I will see if I can follow your lead and create a sp_expm_frechet

argriffing:
Please feel free to contribute improvements upstream! You can blame me for the scipy implementations of those functions, I'm just a postdoc who needed them for some computational biology calculations. Sorry if they are bad. Also they may be bit-rotting further due to the questionable 'LinearOperator' situation in scipy and the inconsistencies among the np.ndarray, np.matrix, scipy.sparse, and linear operator interfaces.

nonhermitian:
No worries. Will add a pull tonight. Our previous implementation was also sketchy. Also wrote that as a postdoc before the SciPy version existed. Luckily, Expokit knows more than I do!

nonhermitian:
Actually, with further testing it seems that the scipy method is better at larger sized matrices. Should probably stick with that one. The older qutip algorithm is closer to the matlab version per 3015b, and returns similar results. It seems the scipy one is a bit better behaved at larger sizes.

PR #465, opened 2016-04-14
ENH: Remove CXVPY logger warning in metrics

nonhermitian:
Having a warning that appears every time qutip is imported should be avoided.  That is unless we want a ton of questions on the google help group.

### DIFF
--- a/qutip/metrics.py
+++ b/qutip/metrics.py
@@ -57,7 +57,6 @@ try:
     import cvxpy
 except:
     cvxpy = None
-    logger.warning("CVXPY not found, semidefinite solving is disabled.")
 
 
 def fidelity(A, B):


cgranade:
Good point, sorry to cause that problem, then. Is it worth keeping it at a debug or info severity instead, then? Also, should the warning after a failed import of configobj also be removed or downgraded (https://github.com/qutip/qutip/blob/4fc89ce6804324f20e84039d961ac765105a7a2e/qutip/settings.py#L98)?

nonhermitian:
Perhaps at the info level.  At least that is my naive thinking.  However, I think maybe the exception that you have in the dnorm is good enough.  Then the user can install the required module if they want to use that function.  Regardless though, this is something that really needs to be specified in the docs.

cgranade:
I'll add it to the docs, then, and make a new PR for that shortly.

PR #466, opened 2016-04-16
ENH: Move to SciPy expm

nonhermitian:
The new version of expm that I grabbed from Expokit does not behave quite as well as the SciPy version.  After a bit of testing, the SciPy version performs the best overall.  For our purposes, I think that either method would be sufficient.  However, the SciPy version is the current best algorithm, and replacing as much code as possible with SciPy code is probably a good thing.

This code still defaults to the dense method.


nonhermitian:
Merging this now.

PR #467, opened 2016-04-21
Fix a complicated list-cast-map-list antipattern

Krastanov:
The previous version contained multiple antipatterns:

- casting to list
- map when comprehension can be used
- map over a comprehension
- len(range(list)) looping

I also suspect that the `zip(*vecs)` call is just a  strange way to reimplement the transpose method, but I am not familiar with the code and did not want to change it.


cgranade:
This looks like a nice fix, thank you! Since it's passing all checks, I'll mark this as merge-review ready.

nonhermitian:
Merging this now since there are no other comments.

PR #470, opened 2016-05-03
ENH: Parallel spmv function (WIP)

nonhermitian:
Here is a parallel version of the spmv_csr function that we use almost everywhere.  I get quite a bit of speedup as the dimensionality of the system goes above ~500 or so states (depends on the problem though).  Here is a plot of the speedup from serial for a spin-chain.

parallel_spmv

It uses the openmp from cython so the GCC is required.  On linux this is fine, but on OSX the GCC needs to come from homebrew or macports as the builtin clang does not support openmp.  On Win, my guess is that the intel compiler could be used.

Integrating this into the various solvers will be done in the next couple of days.


nonhermitian:
I have added this to 3.3, but it can likely be squeezed into the 3.2 branch, depending on the release timeframe.

PR #471, opened 2016-05-07
cimport spmv in codegen 

nonhermitian:
This prevents intermediate calls to Python from the time-dependent string format.


nonhermitian:
Merging this now.

PR #476, opened 2016-05-23
Hermitian eigenvalue solver used for spectral_decomp

ajgpitch:
eigh now used in place eig for all oper_dtype
This resolved loss of unitary evolution that had be observed in some cases.


ajgpitch:
Can anyone explain why all these travis errors occur?
It seems like they all relate to CVXOPT. They also seem to occur many (10) times each.

cgranade:
It looks like something broke in the diamond norm test cases, but I'm not sure why. I'll take a look, see if I figure out what happened.

ajgpitch:
@cgranade  Ok, great, thank you

ajgpitch:
The CI errors do not relate to this change, and it is only a one character bug fix, so I will merge unless anyone objects.

nonhermitian:
Go for it

Issue #477, opened 2016-05-24
Non-unitary evolution in control module


ajgpitch:
The DynamicsUnitary class in some cases does produce unitary evolution of a quantum state. This was spotted by a colleague. The cases typically for a small time interval,

ajgpitch:
This has been resolved by #476

ajgpitch:
Fixed

Issue #371, opened 2015-09-02
Some control objects debug info is output to the file system


ajgpitch:
There are various options for 'test_output' files that can be set to monitor progress of an optimisation or to debug. These are currently output to files during the run. By default none of these files are produced. 

It would be better to return an object tree, which could then saved to file if necessary. It should remain possible to output to file during the run, as in some cases this can be useful.

ajgpitch:
Done

PR #478, opened 2016-05-24
Qtrl data check and dump

ajgpitch:
Unitarity checking added for propagators of time evolution
dumping of transient calculation operators added
generation of summaries for pulse optimisation iterations and system evolution computations added

### DIFF
--- a/qutip/control/dump.py
+++ b/qutip/control/dump.py
@@ -97,17 +97,15 @@ class DynamicsDump(Dump):
     def reset(self):
         dyn = self.parent
         self.evo_dumps = []
-        self.dump_amps = True
-        self.dump_dyn_gen = True
-        self.dump_prop = True
-        self.dump_prop_grad = True
-        self.dump_fwd_evo = True
-        self.dump_onwd_evo = dyn.fid_computer.uses_onwd_evo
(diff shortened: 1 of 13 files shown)


ajgpitch:
Initially started to investigate #477 
Found that work proposed in #371 would be very helpful, decided to bite the bullet and get it done.
This also contains the fix #476

ajgpitch:
Need to add unitarity checking and data dumping into some tests. 
Also, will squash commits before merging.

ajgpitch:
The dumping tests were mainly added to increase coverage
The unitarity test would have caught #477

ajgpitch:
I have rebased to pick up all the recent bug fixes.  Hopefully this will mean this will pass all the tests now.
I have been using the new features quite a lot, so I will squash and merge if this passes the tests.

ajgpitch:
After a few attempts I found that I was unable to successfully squash any commits here. Perhaps I should have done it before rebasing from master? I think I am in danger of doing more harm than good through my squashing, so I will just merge this as it is unless anyone objects.

nonhermitian:
There is a squash merge option when you merge.

PR #481, opened 2016-05-25
check H and rho0 are the same shape, for #479

### LINKED ISSUE #479: mesolve allows different dimensions for H and rho0
I'm still fairly new to QuTiP but I've found that `mesolve` allows `H` and `rho0` to have different dimensions. Of course getting it right is the users responsibility, an informative warning may be useful. A minimal example is as follows:
```
from qutip import *
import numpy as np

N = 3
a = tensor(destroy(N),identity(2))
sz = tensor(identity(N),sigmaz())

H = sz*a.dag() + sz.dag()*a
collapse = [a]
tlist = np.linspace(0,1,100)
solution = mesolve(H, basis(2,0), tlist, collapse, [a.dag()*a])
```
which gives a numerical integration error:
`UserWarning: zvode: Repeated convergence failures. (Perhaps bad Jacobian supplied or wrong choice of MF or tolerances.)`
Rather than a more direct warning about the (more likely) problem of mis-matched Hilbert spaces. I'm happy to add a check to the `mesolve` function and submit a PR, but I wanted to be sure that I'm not missing some use case where `H` and `rho0` should be allowed to differ.

amcdawes:
an attempt to fix #479 
Doesn't check `H` and `rho0` in the case of `H` as `list` type (but that could be added).

### DIFF
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -264,6 +264,10 @@ def mesolve(H, rho0, tlist, c_ops, e_ops, args={}, options=None,
 
         if isinstance(H, Qobj):
             # constant hamiltonian
+            if H.shape != rho0.shape:
+                raise TypeError("Hamiltonian (H) and initial state (rho0)" +
+                                "do not have the same shape. ")
+
             if n_func == 0 and n_str == 0:
                 # constant collapse operators
                 res = _mesolve_const(H, rho0, tlist, c_ops,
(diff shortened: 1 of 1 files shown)


arnelg:
You should take care of the fact that both H and rho0 are allowed to be superoperators as well. H can be a Liouvillian, and rho0 can be a superoperator to allow for the type of situation mentioned in #384. If one of the two, but not both, are supers, your simple test will fail. 

We should fix the docstring of `mesolve` so that it mentions that rho0 can be a super.

nonhermitian:
We also need to be careful as the propagator function makes use of this relaxed checking.  That would need to be changed as well.

nonhermitian:
The propagator does not respect the usual rules when inputing to mesolve. So that wiuld also have to be addressed

nonhermitian:
Having played around a bit, there are many functions that take advantage of the loose checking in mesolve.  The propagator is one that comes to mind.  One way to accommodate everything would be to have an extra kwarg, such as `_check_inputs` or something, that we can have True by default, but then can turn off internally.

amcdawes:
That makes a lot of sense and would certainly let us handle specific cases that might be more problematic for new users.

I apologize, I haven't had a chance to look at this again more recently but should be able to soon.

Andy

nonhermitian:
This was taken care of by Pull #537

Issue #480, opened 2016-05-25
final_state output of mesolve has a strange dimension


fabiananders:
Hello.

As qutip is new to me, I'am not shure if there is a real problem or if it is just poor understanding. Well, here is my issue:
I'd like to get the the final state after solving some unitary time evolution with `mesolve`. The input state is a tensor product of three Fock states.
`psi0 = tensor(basis(N+1,0), basis(N+1,0), basis(N+1,N))`

Then I do the evolution like this:
`H = tensor(a, b, c.dag() * c.dag()) - tensor(a.dag(), b.dag(), c * c)`
Operators a,b and c are defined as `destroy(N+1)`.
`times = np.linspace(0.0, 2.0, 100.0)`
`opts = Options(store_states=False, store_final_state=True)`
`result = mesolve(H, psi0, times, [], [numAop,numBop,numCop], options=opts)`

Now I observe different dimension of input and output state:
`print(psi0.dims, psi0.shape)`
`print(result.final_state.dims, result.final_state.shape)`
gives:
[[41, 41, 41], [1, 1, 1]] [68921, 1]
[[68921], [1]] [68921, 1]
So somehow the dimension structure is lost. The results for the expectation values of the number operators look reliable. So I guess that the time evolution itself worked fine.
Kind regards
Fabian

nonhermitian:
The difference in dimensions is due to the fact that the density matrix must be flattened to an array for the computation.  You can see this by noting that 41**3 = 68921.  Basically the vector is not converted back to a matrix at the end.  This will be fixed in the next day or so.  Thanks for pointing it out!

nonhermitian:
This is fixed in Pull #486.  Turns out that the Qobj dims attribute was not set.

PR #486, opened 2016-05-27
FIXED: Wrong dims (Issue #480)

nonhermitian:
- Fixed the dims on the final_state output in mesolve and sesolve.

- Also explicitly set the isherm on ouput from mesolve to be True, since the output should be viewed as aa DM regardless if small numerical errors may the isherm check fail.


nonhermitian:
This is also ready to merge.

cgranade:
I agree, looks good.

fabiananders:
Works perfectly. Thanks!

Issue #484, opened 2016-05-27
Travis CI failures for dnorm()


cgranade:
As noted by @ajgpitch in #476, Travis CI is currently raising a large number of errors, as every call to ``dnorm()`` results in a ``SolverError``. I am in the process of investigating this problem, but have not yet been able to reproduce the test failures locally. It appears as though the failures only occur when MKL and CVXPY are used together, but I am able to use MKL and CVXPY together fine locally.

cgranade:
After investigating a bit more, I think this may be related to 3a5ab42200b3f1851343081614064e67be3b02bf, in that SCS and CVXOPT are not yet compatible with MKL, and require OpenBLAS instead. If I locally use pip to install SCS and CVXOPT with an MKL-enabled conda env, then importing cvxpy fails due to undefined symbols. On the other hand, if I use the Omnia conda channel to install cvxopt and scs, then ``dnorm()`` works, but the conda env switches to use OpenBLAS. To resolve the build failures, then, I'd suggest temporarily marking the tests as skipped if MKL is enabled.

nonhermitian:
A sticky situation given that Anaconda is MKL by default now.  I agree that perhaps the tests can be skipped if MKL is present.  This can easily be checked via the qutip.settings.has_mkl property

cgranade:
Agreed, it is a bit sticky. I'd guess that since the MKL switchover was relatively recent, they haven't had time to get things working. At any rate, I'll go on and make a PR that skips the relevant tests then.

cgranade:
Since #487 is now merged in, I'll go on and close this for now, but it's worth revisiting once MKL is better supported upstream.

nonhermitian:
Can you submit an issue with the relevant projects?

cgranade:
Yeah, I'd like to get a better isolated failure case first, but once I've done that, I think it makes sense to report this to CVXOPT especially.

PR #487, opened 2016-05-27
Skipping dnorm tests when MKL is present.

cgranade:
This PR addresses #484 by skipping dnorm() tests in the known failure case that MKL is present (see #484 and cvxopt/cvxopt#57).


cgranade:
It looks as though the known failures are now properly being skipped. Given that this is causing build failures elsewhere, I'd suggest that this PR is ready for merge.

nonhermitian:
I would agree.  Merge when ready.

Issue #426, opened 2016-02-12
Wrong import order in setup.py


Phyks:
Hi,

I think the import order is not correct in `setup.py`, importing `numpy` at the beginning of the file, before giving a chance to `setuptools` to actually install it (because it is in required packages). This line is faulty https://github.com/qutip/qutip/blob/master/setup.py#L34 and results in

``` zsh
% pip install qutip
Collecting qutip
  Downloading qutip-3.1.0.tar.gz (800kB)
    100% |████████████████████████████████| 802kB 595kB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-w53v9c3b/qutip/setup.py", line 34, in <module>
        import numpy as np
    ImportError: No module named 'numpy'
```

Then, instead of `pip install qutip`, one has to `pip install numpy; pip install qutip`.

Phyks:
Note that the same holds for Cython:

```zsh
% pip install qutip
Collecting qutip
  Using cached qutip-3.1.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-hg437fn6/qutip/setup.py", line 159, in <module>
        configuration=configuration
      File "/home/phyks/.local/share/virtualenvs/physique/lib/python3.5/site-packages/numpy/distutils/core.py", line 135, in setup
        config = configuration()
      File "/tmp/pip-build-hg437fn6/qutip/setup.py", line 135, in configuration
        config.add_subpackage('qutip')
      File "/home/phyks/.local/share/virtualenvs/physique/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 1002, in add_subpackage
        caller_level = 2)
      File "/home/phyks/.local/share/virtualenvs/physique/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 971, in get_subpackage
        caller_level = caller_level + 1)
      File "/home/phyks/.local/share/virtualenvs/physique/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 908, in _get_configuration_from_setup_py
        config = setup_module.configuration(*args)
      File "qutip/setup.py", line 8, in configuration
        config.add_subpackage('cy')
[...]

PR #489, opened 2016-06-07
Fix setup.py for use from external requirements.txt

cgranade:
As raised in #426, QuTiP currently does not install correctly if listed in the requirements.txt for an external project unless the target environment already has numpy installed. This is due to the use of ``numpy.distutils`` in the current ``setup.py``, such that this PR attempts to solve the issue by falling back to setuptools and then distutils when numpy is missing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,16 +38,23 @@ try:
     import setuptools
     TEST_SUITE = 'nose.collector'
     TESTS_REQUIRE = ['nose']
-    TESTING_KWARGS = {
+    EXTRA_KWARGS = {
         'test_suite': TEST_SUITE,
         'tests_require': TESTS_REQUIRE
     }
 except:
-    TESTING_KWARGS = {}
+    EXTRA_KWARGS = {}
 
 try:
     import numpy as np
     from numpy.distutils.core import setup
+
+    # If we use NumPy's distutils, it will also
+    # try to import Cython due to the add_subpackage
+    # calls below. We want to fail early instead, so that
+    # we branch off to the setuptools/distutils fallbacks
+    # if Cython isn't present.
+    import Cython
 except ImportError:
     # Use a more basic implementation of setup
     # from setuptools so that we can bootstrap install_requires.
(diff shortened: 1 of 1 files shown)


cgranade:
Testing locally in a conda environment without NumPy, this seems to work. In particular, I tested with the following commands:

```bash
$ conda create -n nonp python=2.7
$ source activate nonp
$ pip install git+https://github.com/cgranade/qutip.git@fix-setup-numpy#egg=QuTiP
```

cgranade:
As a test, I've used this PR from the RTD build for [qinfer/python-qinfer](https://github.com/QInfer/python-qinfer), since the RTD virtualenv is by default rather restrictive. The modified setup.py works for that build process, which I think is a fairly good sign.

cgranade:
After more examination, it appears as though ``qutip/cy/setup.py`` breaks the fix introduced by this PR if Cython is not installed before QuTiP. I'll investigate, but I'd like to get ``pip install`` working even from very minimal environments.

cgranade:
My current understanding is that if NumPy is installed but not Cython, then when ``egg-info`` is called, NumPy distutils will happily follow ``add_subpackage`` to pull in ``qutip/cy/setup.py``, then choke on the missing Cython dependency. On the other hand, if both are missing, the new fallbacks introduced in this PR will be tripped, such that the ``add_subpackage`` call will not happen, and ``qutip/cy/setup.py`` will not get called. I'll readily admit that NumPy but not Cython is an edge case, but it's one that we can probably deal with by trying to import Cython at the top and letting the ``ImportError`` branch us onto the fallback path. Thoughts?

cgranade:
The latest attempt seems to correctly generate ``egg_info`` if NumPy is missing, if NumPy is present by not Cython, or if both are present. That should make it much easier to list QuTiP as an external dependency in ``requirements.txt``, as  our``setup.py`` should no longer assume a particular install order. If the latest CI passes, I'll add the review merge ready label.

ajgpitch:
I don't claim to understand the technical details, but this seems like a very good idea. And would seem to be ready to merge

cgranade:
Cool, merging now, then. Thanks!

PR #494, opened 2016-06-10
Fix #493

### LINKED ISSUE #493: Qobj.expm() not functioning properly for diagonal states which have a zero element in the diagonal
Qobj.expm() is currently not working properly for diagonal states which have a zero in the diagonal (these zeros don't become ones). An example:

```python
>>> foo = qutip.Qobj([[1.,0.],[0.,0.]])
>>> foo.expm()
Quantum object: dims = [[2], [2]], shape = [2, 2], type = oper, isherm = True
 Qobj data =
 [[ 2.71828183  0.        ]
  [ 0.          0.        ]]
```

While the correct result should have been (notice the element in second row, second column):
```python
Quantum object: dims = [[2], [2]], shape = [2, 2], type = oper, isherm = True
 Qobj data =
 [[ 2.71828183  0.        ]
  [ 0.          1.        ]]
```

On analysing the source code I found the the function `sp_expm` in `sparse.py` is responsible for this computation. The particular code is reproduced below:
```python
def sp_expm(A, sparse=False):
    """
    Sparse matrix exponential.    
    """
    if _isdiag(A.indices, A.indptr, A.shape[0]):
        A.data = np.exp(A.data)
        return A
    if sparse:
        E = spla.expm(A.tocsc())
    else:
        E = spla.expm(A.toarray())
    return sp.csr_matrix(E)
```
Since `A.data` is stored in sparse matrix format, only non-zero elements are exponentiated. A way to avoid this problem would be to either completely remove the case of optimizing for diagonal states or modifying the appropriate code as:

```python
def sp_expm(A, sparse=False):
    """
    Sparse matrix exponential.    
    """
    if _isdiag(A.indices, A.indptr, A.shape[0]):
        A = sp.diags(np.exp(A.diagonal()),format='csr')
        return A
    if sparse:
        E = spla.expm(A.tocsc())
    else:
        E = spla.expm(A.toarray())
    return sp.csr_matrix(E)
```

I am running qutip 3.2.0.dev-cb698ef
Python 3.5.1, numpy 1.10.4, scipy 0.17.0

nonhermitian:
Fix for #493.

### DIFF
--- a/qutip/sparse.py
+++ b/qutip/sparse.py
@@ -383,7 +383,8 @@ def sp_expm(A, sparse=False):
     Sparse matrix exponential.    
     """
     if _isdiag(A.indices, A.indptr, A.shape[0]):
-        A.data = np.exp(A.data)
+        A = sp.diags(np.exp(A.diagonal()), shape=A.shape, 
+                    format='csr', dtype=complex)
         return A
     if sparse:
         E = spla.expm(A.tocsc())
--- a/qutip/tests/test_qobj.py
+++ b/qutip/tests/test_qobj.py
@@ -40,7 +40,7 @@ from qutip.qobj import Qobj
 from qutip.random_objects import (rand_ket, rand_dm, rand_herm, rand_unitary,
                                   rand_super, rand_super_bcsz, rand_dm_ginibre)
 from qutip.states import basis, fock_dm, ket2dm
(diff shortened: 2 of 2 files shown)


nonhermitian:
Merging this now as I need it for another pull.

PR #496, opened 2016-06-10
Updates to propagator

nonhermitian:
- Propagator can now be run in parallel.

- Simplified dissipative calling by unifying 'sparse' and dense calls when dissipative.

- Added propagator unit tests.


nonhermitian:
Will merge shortly unless there are any comments.

PR #502, opened 2016-07-02
Use propagator in nonmarkov/memorycascade

arnelg:
Changed from `mesolve` to `propagator` to find the propagator
in the `nonmarkov/memorycascade` module. Gives significant speedup for larger systems and slightly slower for smaller systems. "Larger systems" correspond to long times and "smaller systems" to short times in the memorycascade method.

### DIFF
--- a/qutip/nonmarkov/memorycascade.py
+++ b/qutip/nonmarkov/memorycascade.py
@@ -46,6 +46,7 @@ method in qutip.
 """
 
 import numpy as np
+import warnings
 
 import qutip as qt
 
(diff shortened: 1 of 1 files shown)


nonhermitian:
Just note that the propagator gives a dense matrix, so if simulating high dim systems is needed then you may want to keep mesolve as an option.

arnelg:
I see. `propagator` performed better with a system of d=2^5 in a simple test. I better do some testing of even larger systems then before merging this. Any thoughts on why `propagator` might be slower also for small systems and short integration times? Some overhead related to parallelization? Thanks for the info.

nonhermitian:
If your using parallel mode then yes, could be overhead. For larger systems the parallel propagator is probably advantageous. I have been working on a parallel mesolve, but it only works on linux. I am hoping to have it in qutip 3.3. Would be interesting to see which method is better then.

arnelg:
Added optional argument `integrator` to choose between `propagator` or `mesolve` (also a `parallel={True,False}` argument).

Realized there is probably not much to gain from `mesolve` returning a sparse matrix for large systems, however. This is because in the memorycascade method there is some postprocessing of the propagator using the `tensor_contract` method (written by @cgranade) which does not support sparse matrices, so it gets converted to dense anyway. Still `mesolve` actually performs a bit faster for small systems and small times, so still useful to be able to choose the method I guess.

nonhermitian:
Arne, is this ready to go, or are there more things to add?

arnelg:
Yeah, it's ready. It's not actually completely clear when `propagator` outperforms `mesolve`. In the future I will likely completely change the algorithm based on work done by @swhalen who has written a faster implementation..

PR #509, opened 2016-08-04
Ability to calculate two-time correlations with time-dependent c_ops and bug fixes related to the correlator module

kafischer:
This branch adds the ability to calculate two-time correlations with fully time-dependent Hamiltonians and collapse operators. The time-dependence may now either be in the Hamiltonian or the collapse operators, or both. Additionally, support for the numpy list format was added. New tests were made to test all of these functionalities. Bugs were fixed.

### DIFF
--- a/qutip/correlation.py
+++ b/qutip/correlation.py
@@ -1340,9 +1340,11 @@ def _correlation_mc_2t(H, state0, tlist, taulist, c_ops, a_op, b_op, c_op,
                 ]
 
                 # final correlation vector computed by combining the averages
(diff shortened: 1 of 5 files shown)


nonhermitian:
This looks good.  Are you done with this pull, or are you still adding stuff?  On another note, I see that you are using the interp1d function.  I actually have a cubic spline function (real only) that is much faster, and can be called from the str td-format.  I will upload it when I get a chance.

kafischer:
Cool, thanks! Yes, I'm otherwise done with this pull.

nonhermitian:
Merging now.

PR #512, opened 2016-08-10
MKL rename

nonhermitian:
Rename MKL -> _MKL so as to prevent a namespace conflict with the conda mkl service module.


nonhermitian:
Merging now.

PR #513, opened 2016-08-11
Steadystate fixes

nonhermitian:
Moved a section of debugging code inside of a superlu block as it does not work with the MKL sovler


nonhermitian:
Merging now

PR #517, opened 2016-08-18
WIP: Cubic spline compatible with time-dependent format.

nonhermitian:
This is a cubic spline interpolation object that is compatible with being called from the string-based time-dependent format in the solvers.


nonhermitian:
@kafischer : We can actually use the SciPy CubicSpline function to calculate a non-equidistant spline fit, and then modify some of the Cython calling code to fit the str based format.  I can implement this in relatively short order.  This however would require SciPy 0.18+

nonhermitian:
Going to merge this now so that I can start to embed it into the time-dependent stuff.

kafischer:
Cool, that sounds good - I'm sure its much more efficient than what i had proposed.

PR #520, opened 2016-08-21
Modified _td_wrap_array_str to allow for non-linear tlists

kafischer:
I'm not sure if this is the most efficient way to do this, but here's what I came up with.

The previous version only worked for linear tlists, this one now works for arbitrary tlists


kafischer:
I tried using a long string of conditionals to define the non-linear function instead, but this was only faster for some functions and anyways caused cython to crash for long tlists

nonhermitian:
What is the motivation behind having the np array time dependence?  Is this some thing that the cubic spline function could be used for?

kafischer:
I'm a bit confused by your comment: Are asking about the purpose of the np array format itself or did you mean why it should accept non-linear tlists?

The reason it should accept non-linear tlists is because such a thing is good to have when you simulate rapid dynamics followed by slow dynamics. This currently works for all other types of time-dependent formats, so it should also work for this one to avoid confusion. The cubic spline would indeed be nicer here, except it seems to me like it assumes the x values are linear?

nonhermitian:
I mean the motivation behind the np array format itself.  And indeed, the spline assumes equidistant points.

kafischer:
To be honest, I'm not 100% sure myself since it's just turned into a string and cythonized anyways. When it was initially added, I had assumed the purpose was because occasionally a function doesn't work well in the string format or is slow in the list-function format.

Though I'm working on adding time-dependence to the Bloch Redfield solver by using piece-wise constant Liouvillians and there I guess the format makes a bit more sense intuitively.

kafischer:
@nonhermitian Hi Paul, I don't think it's super important to necessarily have the most efficient algorithm for allowing non-linear tlists, but I do think it's good to have to avoid tripping up more advanced users. Could you merge this solution and we can always re-visit later?

nonhermitian:
Let me look at this tonight.  I have uploaded the spline interp method, and just need to integrate it with the solvers.  I have done something similar elsewhere so it should be straightforward (provided I find the time).  The nonlinear list is also easy to incorporate, for SciPy 0.18+, so I think that there interp methods will replace the array based format in the future.

kafischer:
OK, sounds good. Thanks!

nonhermitian:
This needs a unit test of some kind.

kafischer:
Sure, I just added something - just a simple mesolve that uses a nonlinear tlist - I put it in test_td_formats.

kafischer:
@nonhermitian 

Hi Paul, is there something here that you're still waiting on me for?

thanks!

nonhermitian:
I am waiting to do the splines. Now that mcsolve td code is cleaned up that is next on the list.

kafischer:
great, thanks!

nonhermitian:
Please update this pull to the latest master so that we can check if it passes with all of the new changes.  Also, your last correlation pull did not clean up the *.pyx files generated for the time-dependent code.  Please check that no *.pyx files are left hanging around in this pull.

kafischer:
Apologies for the delay and oversight in that last pull. This one does not have that problem.

PR #522, opened 2016-08-24
Revert "ENH: PEP440 versioning fix"

nonhermitian:
Reverts qutip/qutip#518. This has been causing some issues when being downloaded directly from git. Will rework it.


nonhermitian:
Merging now

PR #524, opened 2016-08-24
More graceful way to check for parallel components in ipynbtools

nonhermitian:
This check does not give depreciation warning like last one.


nonhermitian:
Going to merge this now.

PR #527, opened 2016-08-26
ENH: Update to Wigner Clenshaw

nonhermitian:
Just a minor update where the diags of the sparse matrix are grabbed directly rather than converting the matrix to dense form.  For larger density matrices, this method outperforms the dense method due to the time it takes to convert the format and do the multiplication.

- Also added more Wigner unit tests.


nonhermitian:
Merging now.

qutip/tests/test_wigner.py
```diff
@@ -94,16 +94,19 @@ def test_wigner_fock():
         # calculate the wigner function using qutip and analytic formula
         W_qutip = wigner(psi, xvec, yvec, g=2)
```
Vutshi: Currently the default method is 'clenshaw'. Do we want to specify 'iterative' method here or remove the line completely?

PR #528, opened 2016-08-29
Fixes to countstat and pseudo-inverse functions

nwlambert:
This is my first attempt at fixing and slightly extending the functionality of the countstat_current_noise()
function in countstat.py.   There were several issues with this function, which I have attempted to rectify as described below.  Also its my first pull request (instead of asking anubhav or alex to help), so if I messed it up let me know.

1) The default method, a direct solution using sparse matrices, would fail with an "exactly singular" error on small examples (like the well studied single-resonant level or double quantum dot examples).  The other sparse methods also suffered from the same issue, as did the 'direct' method for dense solutions.

The latter, the dense matrices solver, I have fixed by capturing the error using lstsq() if np.linalg.solve() fails.

For the sparse cases there seems to no equivalent fix.  One can solve the issue by adding a small finite frequency to the definition of the shot noise.  However, this seems like a bit of a bodge, and could cause issues if this small frequency is of the order of some system frequencies.  So I have changed the default behavior to use the dense solvers, and added a description of the sparse matrix issues in the function description and in an example notebook I will upload to the example notebooks github.  

Maybe someone can suggest a better fix?

2) There was, in addition, an error in the definition of the shot noise formula in the default sparse matrix direct solution.  In addition, the dense methods which used pinv() were also incorrectly defined (they were missing pre and post- multiplication by the 'Q' operator), and would thus give different results from the direct solutions.  These issues have been fixed, and everything now fits expected analytical results (see example notebook).

3) I have added a frequency dependence via a list of possible frequencies at which to evaluate the shot noise.

4)  I removed the option of passing a pseudo-inverse to the function directly, and changed various other interface things.

I am not sure if everything conforms to PEP8 etc.

Edit: the example notebook is here https://github.com/qutip/qutip-notebooks/pull/35

### DIFF
--- a/qutip/countstat.py
+++ b/qutip/countstat.py
@@ -43,7 +43,11 @@ from qutip.expect import expect_rho_vec
 from qutip.steadystate import pseudo_inverse, steadystate
 from qutip.superoperator import mat2vec, sprepost, spre
 from qutip import operator_to_vector, identity, tensor
-
+import qutip.settings as settings
+from qutip.qobj import Qobj, issuper, isoper
+# Load MKL spsolve if avaiable
+if settings.has_mkl:
+    from qutip._mkl.spsolve import (mkl_splu, mkl_spsolve)
 
 
 def countstat_current(L, c_ops=None, rhoss=None, J_ops=None):
@@ -99,7 +103,7 @@ def countstat_current(L, c_ops=None, rhoss=None, J_ops=None):
 
 
 def countstat_current_noise(L, c_ops, wlist=None, rhoss=None, J_ops=None, 
-                            sparse=False, method='direct'):
+                            sparse=True, method='direct'):
     """
     Compute the cross-current noise spectrum for a list of collapse operators
     `c_ops` corresponding to monitored currents, given the system
@@ -174,10 +178,10 @@ def countstat_current_noise(L, c_ops, wlist=None, rhoss=None, J_ops=None,
     else:
         S = np.zeros((N, N,len(wlist)))
         
-    if sparse==False: 
+    if sparse == False: 
         rhoss_vec = mat2vec(rhoss.full()).ravel()
         for k,w in enumerate(wlist):
-            R = pseudo_inverse(L, rhoss, w, sparse = sparse, method=method)
+            R = pseudo_inverse(L, rhoss=rhoss, w= w, sparse = sparse, method=method)
             for i, Ji in enumerate(J_ops):
                 for j, Jj in enumerate(J_ops):
                     if i == j:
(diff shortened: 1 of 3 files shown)


nwlambert:
I guess I need to edit the tests to match the new function behavior.

nwlambert:
One minor thing;  I guess finite n_th was included in the test functions because it also removes the singular behavior in the sparse solvers.

nonhermitian:
Hey buddy.

There is a much faster (100x or more) sparse solver in the `_mkl` folder.  The call is `mkl_spsolve` and it is a replacement for the SUPERLU solver that comes with scipy.  The mkl module is only loaded for the Anaconda and Intel Python distributions.  So you will need to have an if statement to check for it.  The setting `has_mkl` tells you if the module is found and can be loaded.

nonhermitian:
If you add a small nonzero term (~1e-15) to the shot noise then shouldn't that help?  I would find it difficult to imagine that someone would care about frequencies in that range.  Especially if they properly normalized the problem.

nwlambert:
Hi paul,

Thanks for the feedback. I think maybe Rob had originally included the mkl stuff, but I took it out when I was bug fixing.  Will put in, I didn't realize it gave such a big performance boost. thats cool.

I will also follow your advice and  stick in a finite frequency to make sure all the solvers work.  

I had a poke around in discussion forums for numpy and scipy, and it seems like the problem is that they solve an exception when they decide L is "exactly singular" (which technically it is).  Sometimes numerical rounding errors prevent this exception, as occurs in larger systems, or adding a finite frequency, so it tries to solve it anyway and succeeds, because the projector Q on the right-hand side guarantees a solution.  Conversely, matlab will just throw a warning, and try and solve it anyway.

qutip/countstat.py
```diff
@@ -144,45 +163,87 @@ def countstat_current_noise(L, c_ops, rhoss=None, J_ops=None, R=False):
...
-        tr_op = tensor([identity(n) for n in L.dims[0][0]])
-        tr_op_vec = operator_to_vector(tr_op)
-
-        Pop = sp.kron(rhoss_vec.data, tr_op_vec.data.T, format='csc')
-        Iop = sp.eye(N*N, N*N, format='csc')
-        Q = Iop - Pop
-        A = L.data.tocsc()
+        S = np.zeros((N, N,len(wlist)))
+        
+    if sparse==False:
```
ajgpitch: PEP8 requires spaces either side of ==
;)

ajgpitch:
Looks like good work to me. Well done for taking that brave new step into the PR world!

I agree with @nonhermitian that if the negligible frequency fixes things then, then so long as it is clearly commented as to why it's there then it's valid.

Re. PEP8, you should see I made one inline comment about " == ". There also seems to be a few unnecessary empty lines that break the rules, but I could not spot anything else with my quick check.

@nwlambert do you need an example of where the mkl solver has been used elsewhere in qutip?

ajgpitch:
The mkl solver usage looks good to me. Travis is happy.
I would have no objections to this being merged.

PR #530, opened 2016-08-30
ENH: Faster tensor function

nonhermitian:
- This pull contains a sparse kron function that does not need to convert to COO format to do the product.  It is possible to get up to ~20x performance over the SciPy implementation.

- This function can likely be recast to use OPENMP on the outer-loop.

- Added random tests comparing this method vs the SciPy sparse kron.

Here are speed ups for various random objects (dims in range [2,100]) with random densities:
oper-oper
oper-ket
ket-ket

The graphs show a relatively constant speedup.  However, the results are a bit better than displayed here when using the `timeit` module directly.


Vutshi:
Will something like liouvillian benefit from this new _csr_kron?

nonhermitian:
Ahh yes, good point.  I finished this at 1am, so there are probably things that I am missing.

Vutshi:
Maybe a wrapper (like we have for spmv) can replace every sp.kron in qutip

nonhermitian:
I have tested an OPENMP version of the zcsr_kron function and there is not much to gain overall when compared to the serial version.

nonhermitian:
Going to merge now.

PR #532, opened 2016-09-08
ENH: Bump QuTiP dev version to 4.0

nonhermitian:
We have done quite a bit of work, and added a lot of functionality for a point release.  I am suggesting that we bump the dev version to 4.0, and have this be the version that corresponds to the PRX paper.


nonhermitian:
Since there are no comments about this, merging now.

Issue #534, opened 2016-09-09
vector_to_operator() conflicts with updated kron and tensor


Vutshi:
vector_to_operator() returns CSC matrix which is in conflict with CSR assumption of the new kron and tensor.
@nonhermitian

nonhermitian:
Was fixe in Pull #535.

PR #535, opened 2016-09-11
FIX: Fix for Issue #534

nonhermitian:
Fix for issue #534.  As usual it is the implicit conversion from CSR -> CSC when taking the transpose.


nonhermitian:
Merging this now

PR #536, opened 2016-09-12
BUG: hardware_info - physicalcpu not found on MacOS Sierra

nonhermitian:
The physicalcpu item from the commandline has shifted locations in the latest MacOS release.  This fi should fix this and all later versions.


nonhermitian:
Merging this now so that I can continue with other pulls.

PR #541, opened 2016-09-22
ENH: Simplified & optimized mcsolve codegen

nonhermitian:
- Replaced csr_spmv & for-loop at end with spmvpy calls.

- Removed codegen steps that are no longer needed.


nonhermitian:
Merging now.

PR #545, opened 2016-10-04
Bugfix in bra vector permute

Vutshi:
A bug is fixed in Qobj permute and bra vector tests are added.
The bug is found by a student of mine Slava Kuzmin (IQOQI).


nonhermitian:
Many thanks to your student for hunting this down.  Merging now.

PR #546, opened 2016-10-06
BUG: spmvpy is not void

nonhermitian:
We were outputting unnecessary arrays for spmvpy when the function should be void


nonhermitian:
Merging now

PR #547, opened 2016-10-07
BUG:Fix for broken Travis tests

nonhermitian:
MKL library has moved to envs dir


nonhermitian:
Ok, it seems that for some reason the 'mkl_info' key from the NumPy config info is no longer available in some miniconda installs.  This took forever to find as I could not reproduce the error at home when using a fresh miniconda install.  Now I look in the libs as well for BLAS info.

- Also removed matplotlib from the requirements as it is not a requirement.  At least not at present.

nonhermitian:
Merging now so that the master now passes.

PR #550, opened 2016-10-14
ENH: Use Cubic_Spline in dynamics

nonhermitian:
- Cubic_Spline objects can now be used in the me, se, and mc dynamics
solvers as a time-dependent term.

- If possible, the Cubic_Spline object will be evaluated in the
str-based format.

- I have yet to add this functionality to the collapse terms .


nonhermitian:
Merging this now.

PR #551, opened 2016-10-14
ENH: Faster spmv code

nonhermitian:
- One can get a 10-40%+ speed enhancement on the core spmv and spmvpy functions by moving to c-code and intrinsics on SSE3 compatible processors.  The biggest gains come on platforms that use clang (e.g. OSX) where the speedups are in the 40% or higher range compared to the current Cython based method.  GCC seems to be better suited to compile Cython code, so the gains are only 10-20% there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 dist: trusty
 language: python
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,16 +7,49 @@ python:
 matrix:
     include:
         - python: 3.5
-          env: CC=clang CXX=clang++
+        - compiler: clang
+          addons:
+            apt:
+              sources:
+                - ubuntu-toolchain-r-test
+                - llvm-toolchain-precise-3.6
+              packages:
+                - clang-3.6
+          env: COMPILER=clang++-3.6
+        
         - python: 3.5
-          env: NOMKL=1
+            env: NOMKL=1
+        - compiler: gcc
+            addons:
+              apt:
+                sources:
+                  - ubuntu-toolchain-r-test
+                packages:
+                 - g++-4.9
+            env: COMPILER=g++-4.9
+        
+        - python: 3.5
+        - compiler: gcc
+            addons:
+              apt:
+                sources:
+                  - ubuntu-toolchain-r-test
+                packages:
+                 - g++-4.9
+            env: COMPILER=g++-4.9
+        
+        - python: 2.7
+        - compiler: gcc
+            addons:
+              apt:
+                sources:
+                  - ubuntu-toolchain-r-test
+                packages:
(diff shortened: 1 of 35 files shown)


Vutshi:
This is very interesting! I was looking for some time already a way to incorporate assembler here. Now it is possible to use FMA (fused multiply add) which should greatly improve speed of almost anything or there is a very neat implementation of complex number product with FMA which is exact (avoids loss of precision) with almost no performance penalty.

nonhermitian:
Ok, it looks like the tests are finally going to pass.  As the number of commits suggests, this was a pain. Note because the code was wrong, but because:

1) The default config used by Travis has GCC 4.6 which contains an error that happens to get raised in our case.  Thus, had to move to the Trusty build stack.

2) Although it took forever to hunt down, it appears that when .pyx files are included as package data, Py27 may try to rebuild them upon calling __init__.  This would throw a missing header error.  Since these files are not needed after building, removing them clears up this issue.

nonhermitian:
@Vutshi That sounds interesting.  I have not done much coding at the intrinsics level before, so I am finding out about all sorts to tricks.  I saw a couple of lectures by a guy at Intel whom suggested that the best way to go is to let the compiler figure it out after hinting and adding pragma.  For vectorization, he suggested OPENMP 4+ as the way to go.  They have a omp for simd pragma that was suggested.  However, OPENMP 4+ requires GCC 4.9+ and some corresponding LLVM version that probably makes those methods a bit too specialized for a general qutip release.

Vutshi:
Apparently, gcc and clang with proper compiler options can produce code with FMA from the bare c code [https://godbolt.org/g/pte65q](url)

Here are also some useful links regarding complex product with FMA (just in order not to forget it):
[Crunching Numbers with AVX and AVX2](http://www.codeproject.com/Articles/874396/Crunching-Numbers-with-AVX-and-AVX)
[Stackoverflow: multiplying two arrays of complex numbers](http://stackoverflow.com/a/39521257/224132)

nonhermitian:
I will have to go back and take a look, but I could not get the compiled gcc code to go faster than the intrinsics. Also the plain c code version under clang was quite slow, sometimes performing worse that the Cython version. 

Also note that the vector data in the spmv is fetched out of order, so I am not sure if grabbing multiple elements as can be done in AVX is going to give a performance benefit.

Vutshi:
I agree that the vectorization like in the links above is not what we need for sparse operations. The links are mostly only for reference on how to work with avx2.
For us the FMA is important since it should be faster and more precise. Maybe the compiler code is not optimal in usage of the fused operation. 
In principle there should be 4 FMA operations for one line:
`dot += data[jj]*vec[ind[jj]]`
and the compiler produces them with -ffast-math and -mfma options on. Maybe some over operations around are not optimal.

PR #554, opened 2016-11-03
adding an opencl version of mcsolve

cwassert:
This pull request contains a reimplementation of `mcsolve` using [OpenCL](https://www.khronos.org/opencl/). More specifically [pyopencl](https://mathema.tician.de/software/pyopencl/).

The function is called `mcsolve_cl` and can be used as a drop-in replacement for `mcsolve` in many cases.

### DIFF
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -188,6 +188,12 @@ except:
 else:
     qutip.settings.fortran = True
 
+# try to import the opencl mcsolver
+try:
+    from qutip.opencl import *
+except:
+    pass
+
 # -----------------------------------------------------------------------------
 # Check that import modules are compatible with requested configuration
 #
--- a/qutip/cy/sparse_utils.pyx
+++ b/qutip/cy/sparse_utils.pyx
@@ -33,8 +33,11 @@
 import numpy as np
 import scipy.sparse as sp
 cimport numpy as np
-from libc.math cimport abs
+from libc.math cimport abs, sqrt
 cimport cython
+
+cdef extern from "complex.h":
+    double cabs(double complex x)
 
 cdef inline int int_max(int x, int y):
     return x ^ ((x ^ y) & -(x < y))
@@ -321,4 +324,19 @@ cpdef _csr_kron(np.ndarray[complex, ndim=1, mode="c"] dataA,
                 ptr_end += distB
     
     return sp.csr_matrix((out_data,out_inds,out_ptr), shape=(rows_out,cols_out))
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def unit_row_norm(complex[::1] data, int[::1] ptr, int nrows):
+    cdef size_t row, ii
+    cdef double total
+    for row in range(nrows):
+        total = 0
+        for ii in range(ptr[row], ptr[row+1]):
+            total += cabs(data[ii])**2
+        total = sqrt(total)
+        for ii in range(ptr[row], ptr[row+1]):
+            data[ii] /= total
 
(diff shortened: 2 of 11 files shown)


nonhermitian:
This looks like a very interesting pull.  I will have to play around with it. In the mean time, you will need to get the Travis tests to run and probably make a jupyter notebook to demo the usage. Hopefully the former just requires adding pyopencl to the install packages in Travis.yml

nonhermitian:
I am getting a variety of errors when running the test scripts under OSX:

- For `test_mcsolve_cl_time_dep_h` I get "atol could not be reached for 1 trajectories".

I am also seeing:

"CompilerWarning: Non-empty compiler output encountered. Set the environment variable PYOPENCL_COMPILER_OUTPUT=1 to see more.
  "to see more.", CompilerWarning)"

"AttributeError: 'NoneType' object has no attribute 'clobj__delete'
Exception ignored in: <object repr() failed>"

nonhermitian:
Also, instead of adding a third entry to the nested list characterizing the time-dependence, I think the real and imag parts should be in the [1] list entry as a tuple.  That way, all of the time-dependence is in a single list item.  We may in the future want to add additional items to the list (e.g. on-off times for the time-dependent terms.)

I also think that we should get some feedback from @ajgpitch and @cgranade.  Chris has some experience with opencl, and he probably is bette suited for feedback than myself.  

Another question would be the issue of maintainability.  If, for some reason, you are unable to address bug issues or incompatibilities with future QuTiP versions, then can we support this large code submission without having authored it ourselves.  For instance, the mcsolve_f90 routine has largely been non-maintained since Arne added it years ago.

cwassert:
First of all I'm glad you could get it to run under OSX. Which OpenCL environment are you using?
* Is the failure of `test_mcsolve_cl_time_dep_h` reproducible?
* You can probably ignore the compiler warning. I assume it is related to the pragma statement. Some OpenCL implementations refuse to work without it, others give you a warning that it is unnecessary. Maybe we can omit the statement based on the platform version and hide some warning, but that would require more testing (on different platforms). At the moment there is a lot of debug information to ease testing.
* The AttributeError you get is something I have not encountered yet.

If you do consider merging this branch, it is a good idea to get more people involved in testing. The code has not been reviewed and probably contains some bugs.

The questions you raised concerning the maintainability are all valid. On top of that setting up OpenCL is somewhat troublesome, which makes `mcsolve_cl` a kind of experimental feature. Ultimately it is up to the maintainers of qutip to decide whether such a feature should be included. The advantage `mcsolve_cl` has over `mcsolve_f90` is that it supports some time-dependence and introduces no build dependencies. If, for some reason `mcsolve_cl` does not work, the user can always fall back to `mcsolve`.

By the way, is it wise to rebase this branch all the time or only when necessary?

ajgpitch:
Potentially GSoC student could look at this, so leaving open for now

hodgestar:
@cwassert Hello! I know this is a somewhat old PR, but I wanted to mention that for QuTiP v5 we are in the process of rewriting mcsolve (#1710) to use the new data layer. We already have a data layer for CuPy (https://github.com/qutip/qutip-cupy) and one could easily imagine using the OpenCL functions defined here to make an OpenCL QuTiP v5 data layer which would then allow many Qobj operations to use OpenCL.

I'd like to close the PR in favour of the new approach, but I'd like to hear thoughts from others first.

cwassert:
I made this pull request as part of my bachelor thesis and to give something back to qutip. But this branch has not been touched for 5 years. Turns out @nonhermitian had it right from the start with his concerns about the maintainability. : )
  I think you should close it. Keep up the good work!

nonhermitian:
Haha. I read this and remembered this discussion.

hodgestar:
@cwassert @nonhermitian Thanks! I think this is still a good idea, but will close this PR.

PR #557, opened 2016-11-09
ENH: Batch solve unitary propagator

nonhermitian:
- Solve for all basis vectors in unitary propagator simultaneously.

- Use this method as default.


nonhermitian:
This looks good to go, so merging.

PR #559, opened 2016-11-11
ENH: CFLAGS updates

nonhermitian:
- The current choice of CFLAGS made generating the str-based c-code slower than necessary for little to no speed gain.  Here we choose flags that optimize less since there is no benefit.


nonhermitian:
Merging this now

PR #560, opened 2016-11-15
ENH: optimise unit_row_norm

Vutshi:
cabs( z )**2 is too expensive. Now it is ~ 5 times faster.


nonhermitian:
Thanks for another good pull.

PR #563, opened 2016-11-28
Fix for control pulseoption tau array for timeslots

ajgpitch:
Bug fixed where using tau array cause self referential loop


ajgpitch:
Test also added that would have caught this bug

ajgpitch:
As this was a simple bug fix and all tested passed, then merging now.

PR #565, opened 2016-11-30
Fix redundancy and sign error in jmat.

phkx:
In the case when no args were given, a sign was missing in Jy in the extra piece of code written specifically for this purpose. I propose recursive calls to the function with the respective arguments.

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -102,10 +102,7 @@ shape = [3, 3], type = oper, isHerm = True
         raise TypeError('j must be a non-negative integer or half-integer')
 
     if not args:
-        a1 = Qobj(0.5 * (_jplus(j) + _jplus(j).conj().T))
-        a2 = Qobj(0.5 * 1j * (_jplus(j) - _jplus(j).conj().T))
-        a3 = Qobj(_jz(j))
-        return np.array([a1, a2, a3])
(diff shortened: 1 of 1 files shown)


phkx:
Sorry, I'm not sure of the meaning of the failed Travis CI check shown above. I actually came back to comment, that just adding the missing sign may be safer than recursive calls of the function to itself. In the present case I don't see an actual danger of getting stuck in a loop, even with future changes to one of the involved functions, but it now feels safer to just add the missing sign. Maybe someone more experience can comment or just merge the request...

nonhermitian:
Yes, this is a bug in the most recent cvxopt package from pip.  I am trying to fix it in #566. I do not see any issues with the changes you made, so once I get the tests up and running again I will look into merging.  Thanks for the fix!

nonhermitian:
I am going to merge this now as I am sure it passes the tests

nonhermitian:
Actually this does not pass tests, so it is being reverted in #567.

phkx:
I think the test ist faulty. The spin matrices in test_operators.py to which the output of jmat is compared are missing the same sign. Should I fix this and prepare a new pull request?

nonhermitian:
Yes please. Also make sure that you update to the latest master in your pull so that the travis tests pass.

PR #569, opened 2016-12-05
Fix sign error in jmat.

phkx:
Corrected sign in jmat when called without arguments and the respective tests in test_operators.


phkx:
Sorry for the undescriptive name of the branch. I accidentally created this pull request before fixing that.

nonhermitian:
Many thanks for correcting this.  Merging now.

PR #570, opened 2016-12-06
BUG: Move pyximport to end of init.

nonhermitian:
- This should (hopefully) fix any issue where *.pyx files are being compiled twice.

- Remove control module *.pyx package data.

### DIFF
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -113,14 +113,10 @@ try:
               ("(%s), requiring %s." %
                (Cython.__version__, _cython_requirement)))
 
-    import pyximport
-    os.environ['CFLAGS'] = '-O2 -w -ffast-math'
-    pyximport.install(setup_args={'include_dirs': [numpy.get_include()]})
-
 except Exception as e:
     print("QuTiP warning: Cython setup failed: " + str(e))
 else:
-    del Cython, pyximport
+    del Cython
 
 # -----------------------------------------------------------------------------
 # Load user configuration if present: override defaults.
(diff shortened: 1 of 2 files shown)


ajgpitch:
Ok, well seems great that simply moving the pyximport to the end will solve the issue. Good work.

ajgpitch:
I was wondering about the CFLAGS bit. Should we not prepend or append our flags, rather than overwrite?
I was trying to install qutip on our (University) lab machines (for which I did not have sudo). They did not have Cython or pydev. I got quite far, but only by changing the qutip setup to accept include dirs from CPATH. I never got as far as making a PR, because ultimately it failed due to reasons I could not understand, so I parked it.
However I think I could have made something work by setting include dirs in CFLAGS.

nonhermitian:
I believe that all of the include_dirs can go into the setup_args call to pyximport.  Similar to the way the numpy headers are handled.

ajgpitch:
That is what I did in fact, but I had to add it myself. That is load them in from CPATH or C_INCLUDE_PATH. They were not included automatically.

I am not suggesting that we do that in this PR, but maybe extending rather than overwriting the CFLAGS might be better. Unless there is good reason not to?

nonhermitian:
At least on my macbook, the 'CFLAGS' environ variable is not set until we do so manually in init.  Is this not true on other platforms?  If not, then we can easily append.

ajgpitch:
You are right. In general it is not set, so we would get a KeyError. Maybe its best for now just to merge this as it is. We can discuss whether there is any value in allowing user flags and CPATHs elsewhere. I will raise an 'issue' now.

nonhermitian:
Merging now.

Issue #572, opened 2016-12-06
Bloch-Redfield master equations - a possible bug in the presence of multiple subsystems


fbeaudo:
The problem appears in qutip version 4.0.0 when considering the tensor product of a few systems, and including more than one dissipation channel.

I noticed a similar issue on a much more complex system, but in the code attached I consider the simplest system on which I have found the bug to appear.

Consider three uncoupled qubits, each with a distinct frequency. The Hamiltonian is (ignoring tensor products with identity matrices) :

H = (w1 / 2) * sigmaz_1 + (w2 / 2) * sigmaz_2 + (w3 / 2) * sigmaz_3.

Let us couple qubits 2 and 3 to independent white-noise environments through operators sigmax_2 and sigmax_3. Qubit 1 is decoupled from the other qubits, and does not couple to any relaxation channel.

We consider the initial state |1:e, 2:g, 3:g>, where g(e) labels the ground (excited) state of a qubit, and evaluate the population of the excited state of qubit 1 as a function of time using bloch_redfield_tensor() and bloch_redfield_solve().

Since qubit 1 is decoupled from the other qubits, its excited-state population should remain 1 at all times. Yet, qutip yields exponential decay of the population of qubit 1.

I noticed that when I remove dissipation for qubit 2 (keeping only dissipation through sigmax_3), the problem disappears and qubit 1 forever stays in its excited state. So the problem seems to arise when combining two relaxation sources.

I am using qutip version 4.0.0. A colleague of mine tried the attached test code on qutip version 3.1.0 and did not catch any bug (i.e. qubit 1 did not relax with qutip 3.1.0). So the problem seems to be related to versions of qutip later than 3.1.0. Can anyone replicate the bug with the attached code ? 

(I tried putting the code directly here but the hashtag signs for the comments made a mess, and github does not accept attached .py files, so I have put the code in a .txt file instead.)

[redfield.txt](https://github.com/qutip/qutip/files/635122/redfield.txt)

nonhermitian:
I will take a look.  That module underwent some optimizations that may be the cause.

fbeaudo:
Thanks a lot !

nonhermitian:
This is fixed in Pull #573

fbeaudo:
Thanks, Paul. It seems to be working now.

PR #573, opened 2016-12-07
BUG: Fix for Issue #572

nonhermitian:
- Took first element rather than calculating the sum, which was incorrect for more than one a_op.


nonhermitian:
Merging now.

PR #574, opened 2016-12-23
ENH: Faster csr pad for HEOM method

nonhermitian:
A faster cython based csr padding function.


nonhermitian:
Alex, please merge this when you can.

ajgpitch:
@nonhermitian A bit late now, but does this need to go in a heom specific file? Seems it is a pretty generic function, do you think it would be better in spmatfuncs or sparseutils?

nonhermitian:
No other functions use it, and there are likely to be other functions from HEOM that can be Cythonized.  Hence, the new file.

PR #576, opened 2016-12-23
ENH: Update cython aralloc

nonhermitian:
A faster way to get the cdef'ed version of numpy arrays.  Only updated the spmatfuncs module since the others have some compatibility issues that should be addressed first.


nonhermitian:
Merging

PR #582, opened 2016-12-24
Update superoperator.py

ylzhang91:
Add the functions to calculate the adjoint Liouvillian


nonhermitian:
Why do we need this function when one can just take `L.dag()` to get the adjoint?

ylzhang91:
@nonhermitian 
Yes, you are right. Furthermore, the functions 'spre' and 'spost' have the properties:
spre(a).dag() = spre(a.dag()), spre(a) * spost(b) = spost(b) * spre(a).

PR #590, opened 2016-12-28
Corrected typo in documentation

sahmed95:
The documentation for controlled phase gate did not mention the word 'controlled'

### DIFF
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -182,7 +182,7 @@ shape = [2, 2], type = oper, isHerm = False
 
 def cphase(theta, N=2, control=0, target=1):
     """
-    Returns quantum object representing the phase shift gate.
+    Returns quantum object representing the controlled phase shift gate.
 
     Parameters
     ----------
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -773,17 +773,16 @@ def qubit_clifford_group(N=None, target=0):
     S = phasegate(np.pi / 2)
     E = H * (S ** 3) * w ** 3
 
-    for op in map(
+    for op in map(partial(reduce, mul), product(_powers(E, 3),
+                  _powers(X, 2), _powers(S, 4))):
+
         # partial(reduce, mul) returns a function that takes products
         # of its argument, by analogy to sum. Note that by analogy,
-        # sum can be written
-        # partial(reduce, add).
-        partial(reduce, mul),
+        # sum can be written as partial(reduce, add).
+
         # product(...) yields the Cartesian product of its arguments.
         # Here, each element is a tuple (E**i, X**j, S**k) such that
         # partial(reduce, mul) acting on the tuple yields E**i * X**j * S**k.
-        product(_powers(E, 3), _powers(X, 2), _powers(S, 4))
-    ):
 
         # Finally, we optionally expand the gate.
         if N is not None:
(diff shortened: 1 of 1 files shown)


qutip/qip/gates.py
```diff
@@ -918,8 +922,8 @@ def gate_expand_3toN(U, N, controls=[0, 1], target=2):
         raise ValueError(
             "control and not target is None must be integer < integer N")
 
-    if (controls[0] == target or controls[1] == target
-            or controls[0] == controls[1]):
+    if controls[0] == target or controls[1] == target \
```
ajgpitch: Is this better practice? I often use brackets to avoid the use of \
just cause I don't like it :)

qutip/qip/gates.py
```diff
@@ -771,16 +773,17 @@ def qubit_clifford_group(N=None, target=0):
...
+        # partial(reduce, mul) returns a function that takes products
+        # of its argument, by analogy to sum. Note that by analogy,
+        # sum can be written
+        # partial(reduce, add).
+        partial(reduce, mul),
+        # product(...) yields the Cartesian product of its arguments.
+        # Here, each element is a tuple (E**i, X**j, S**k) such that
+        # partial(reduce, mul) acting on the tuple yields E**i * X**j * S**k.
+        product(_powers(E, 3), _powers(X, 2), _powers(S, 4))
+    ):
```
ajgpitch: Is this the correct indentation for this?

ajgpitch:
Hi, I see you started tidying up.
Hopefully you will see I have put a few inline comments. Just questions really

qutip/qip/gates.py
ajgpitch: I see now that your commit name suggests that this is pep8

qutip/qip/gates.py
ajgpitch: I guess this is pep8 also?

qutip/qip/gates.py
ajgpitch: pep8 says brackets as I read it
https://www.python.org/dev/peps/pep-0008/#multiline-if-statements

qutip/qip/gates.py
ajgpitch: I would say borderline and confusing
I would better to indent all inside map to the level of the opening bracket and move the closing bracket to the end.
See first example in:
https://www.python.org/dev/peps/pep-0008/#indentation

ajgpitch:
O, perhaps a bit more than questions :)

qutip/qip/gates.py
sahmed95: How about this ? 
```
    for op in map(partial(reduce, mul), product(_powers(E, 3),
                  _powers(X, 2), _powers(S, 4))):

        # partial(reduce, mul) returns a function that takes products
        # of its argument, by analogy to sum. Note that by analogy,
        # sum can be written
        # partial(reduce, add).

        # product(...) yields the Cartesian product of its arguments.
        # Here, each element is a tuple (E**i, X**j, S**k) such that
        # partial(reduce, mul) acting on the tuple yields E**i * X**j * S**k.
        
```

Having the comments after the for loop shouldn't make too much of a difference I guess.

qutip/qip/gates.py
sahmed95: Yes, I prefer brackets too but I was not able to break the line as a \n before * gives rise to another pep8 (\n before binary operator). I am tempted to change this into two if statements using something like 
```
if (controls[0] in {target, controls[1]} or controls[1] == target):
    ... 
```

or 

```
if (target in {controls[0], controls[1]} or (controls[0] == controls[1])):
    ...
```

ajgpitch:
Yes, very happy with this now. I will go ahead and merge if you are finished?

ajgpitch:
@sahmed95 There you go, you are now officially a QuTiP contributor!

sahmed95:
Thank you ! I am happy to be on board !

PR #609, opened 2017-01-02
ENH: Enforce Qobj data is fast_csr

nonhermitian:
This pull implements a safety check where the Qobj data class is validated as fast_csr_matrix type everytime the Q.data attribute is set.  This guarantees that the Qobj's have the correct data type.  This was not always true previously.

In addition, many items are fixed so that the above rule passes.  Finally, we no longer need the tests that check for Qobj csr_matrix type.  This is now strictly enforced.


ajgpitch:
Is this merge ready? I am looking to make the 4.0.2 release

nonhermitian:
Should be

ajgpitch:
I see you did a lot of work here!

nonhermitian:
Just something to do while jet lagged.

ajgpitch:
I am going to give it a final test on Windows, then build into 4.0.2

sahmed95:
Hi, I am getting this error on the current master. I guess this will be solved once this PR is merged ?

```
air:tests shahnawaz$ nosetests test_sparse.py
E
======================================================================
ERROR: Failure: ImportError (No module named 'qutip.cy.spconvert')
```

sahmed95:
Never mind, I needed a re build.

PR #626, opened 2017-01-07
Floquet lil data fix

ajgpitch:
Setting of `Qobj.data = lil_matrix` changed

PR #628, opened 2017-01-08
Floquet fast csr fix2

ajgpitch:
Fix for error that popped up when running:
```
qutip-doc/guide/scripts/floquet_ex3.py
```
Had to fix the last fix #626 as well.


qutip/floquet.py
```diff
@@ -833,7 +833,7 @@ def floquet_markov_mesolve(R, ekets, rho0, tlist, e_ops, f_modes_table=None,
         if not r.successful():
             break
 
-        rho.data = vec2mat(r.y)
```
ajgpitch: @nonhermitian would you have expected this `vec2mat` to return a `fast_csr_matrix`

PR #629, opened 2017-01-10
ENH: Faster isherm function

nonhermitian:
A faster isherm function.  3-5x faster if operator is Hermitian, 100x+ faster if not.  Essentially does a modified adjoint calculation.  This requires having a csr matrix in canonical format, which is not guaranteed under sparse multiplication.  Since canonical format is important for time-dependent dynamics as well, fast_csr_matrix now does the indices and data sorting for canonical format after multiplication.  This gives a slight performance hit, but having all Qobj in canonical format outweighs this.


ajgpitch:
sounds good to me

PR #632, opened 2017-01-13
BUG: Fix rand_ket with passed dims not working.

nonhermitian:


nonhermitian:
Had to fix the random Haar ket as well.

PR #634, opened 2017-01-15
ENH: Faster partial trace implementation

nonhermitian:
Moved the loops to cython, and reworked the indices selection.  Now 2x - 100x+ faster, depending on size.

This pull also switches the fast_csr_matrix multiplication to our own version.  This eliminates the need to call A.prune() after the SciPy matmat call.  We also handle the nnz=0 case better.

Still need to do:
- ptrace unit tests
- move old ptrace to a 'legacy' folder for testing against.


nonhermitian:
I have added the unit tests ad moved the old ptrace to a new 'legacy folder. 

The permute method for Qobj's has a similar routine to those for selecting indices in ptrace.  I have switched that code to the new ptrace functions as well, but there is some additional cleanup that needs to be done.

ajgpitch:
@nonhermitian  I made a bespoke version of the partial trace function in which the permutation matrix could be passed in place of the `sel` parameter. I did this for a particular project where I was taking the partial trace repeatedly where the tensor structure and selection were the same. I was planning to add this into the master. Do you think it is worth doing?

nonhermitian:
That sounds useful. Perhaps having a flag that tells ptrace to spit out 
the perm matrix, so one could then pass it back to the function?

-P

ajgpitch:
Sounds like a good plan. I will attempt something.

PR #640, opened 2017-02-02
Qobj copy

ajgpitch:
`copy` method added. 
Does not add a great deal, but I have had want of it a few times.
Some PEP8 fixes also completed.


nonhermitian:
Merging now.

PR #648, opened 2017-02-15
ENH: Move cython build to c++

nonhermitian:
I have managed to get everything moved over to c++.  Should address #647. There are still a few issues to address, but the tests pass on unix based systems.

- CFLAGS need to be set for visual studio builds.
- zspmv src files need to be platform dependent (i.e. win/unix versions).  This is mainly due to a few minor variations in how compiler hinting is done.  No big deal.  Also, when moving to openmp, visual studio uses v2.0, which is a bit dated.

PR #650, opened 2017-02-23
ENH : Last 41 fixes

nonhermitian:
Some updates to the Cython code.  Also removed the test scripts for the mcf90 routine as Pull #648 removed the ability to install the module.


nonhermitian:
@ajgpitch, this should be the last pull that I do for 4.1

PR #651, opened 2017-02-28
BUG: Minor updates

nonhermitian:
- Remove unnecessary setting of parallel threads.
- Remove CLFAGS setting that I forgot to remove earlier.
- Add -Wno-strict-prototypes to unix compile flags to get around c++ build issue.


nonhermitian:
merging now

PR #652, opened 2017-03-01
ENH: OPENMP based SpMV

nonhermitian:
A pull for OPENMP based parallel sparse matrix - vector multiplication.


nonhermitian:
The problem with OPENMP is that the point where parallel processing begins to have an advantage is platform, and hardware dependent.  As such, I have added a simple benchmark routine that calculates a good number of NNZ that a matrix should have for OPENMP to be utilized.  This also required me to change the way the qutiprc file is done, i.e. it now uses the standard configparser module.

ajgpitch:
Sounds like a good plan. I was hoping to standardise the way we read in file data. I used the ConfigParser in control/load_params

ajgpitch:
On a similar subject I was doing some matrix multiplication benchmarking recently. I found that:
```
q1.full().dot(q2.full())
```
was faster than
```
q1.data*q2.data
```
for nnz > 10% using randomly generated matrices (on my linux desktop). Full nnz=100% the full matrix multiplication was 20x faster. I was testing with 1024x1024 matrices.
Should we consider having some benchmarking like this pull for when we should convert to full() for matrix multiplication?

nonhermitian:
We do not use dense matrices by default.  This was a design decision that favored targeting larger Hilbert spaces.  For some time now, Robert has advocated having a dense version of the Qobj class that would do the dense linear algebra for smaller system sizes.  That is a ton of work though, as almost all of our algorithms are sparse specific.

nonhermitian:
To get OPENMP to work on OSX one needs to install the LLVM from homebrew.  However, there is a libomp clash with the OPENMP from MKL as two libraries with the same name are simultaneously initialized.  I believe that this is due to the fact that the llvm has its own openmp lib, where as the MKL was compiled against a different openmp lib (perhaps from the Intel compiler as Apple's clang does not have openmp).  This likely also occurs on Windows if using mingwpy.  The workaround is to set
`os.environ['KMP_DUPLICATE_LIB_OK'] = 'True'`.  This is not recommended at all, but seems to work fine.

Perhaps we should give an unsupported warning on platforms where this issues arrises.  Or, we could just roll the dice and see how it goes.

nonhermitian:
@ajgpitch I think this is ready to go.  Let me know if you see any issues with it.

ajgpitch:
Sorry, I was a bit slow there Paul. Looks like a very good addition this.
I see this pull implements its use in sesolve. Is it applicable in the other solvers (me, heom etc)?

nonhermitian:
It is me as well.  Will look into heom.

PR #653, opened 2017-03-03
BUG: pyx still in package data

nonhermitian:
Looks like the pyx files are back.  Probably due to a Pull that was committed but not up to date with master.


ajgpitch:
No, it was deliberate. I added these in as part of the 4.0 release. They need to be distributed with the pip installer, otherwise it cannot build. At least that is why I added them in.

nonhermitian:
Then I guess we would need to add the Cy_grape one in the control folder back.  I can do that here.  Should be fine as I think moving the pyximport call to the end of __init__ fixed the double cythonize.

ajgpitch:
Yes, pretty sure the double cythonize was fixed by your moving the pyximport call to the end of init fix.
Not sure the cy grape is used by anyone, but probably best to put it back for now.

ajgpitch:
I think we will need the `pyx` in the `cy` folder as well.

nonhermitian:
You are indeed correct.

PR #654, opened 2017-03-08
ENH: Faster RCM and inf/one norms

nonhermitian:
Updated RCM and one/inf-norms.


nonhermitian:
Note that int_argsort uses the c++ sort routine that does not guarantee that the array remains unchanged if sorting is not necessary.  Thus the bucky test had to be removed as the returned perm array was compiler specific.

PR #655, opened 2017-03-10
BUG: Cython cleanup not removing temp files

nonhermitian:
In moving to c++, we did not update the Cython cleanup routine.  Also, using Visual Studio leads to a different set of temp files than GCC.  Here, I take advantage of the fact that all the temp files start with 'tdname' and then just walk the dir to clean them, regardless of extension and what not.


ajgpitch:
Deleting everything that starts with 'tdname' sounds sub-optimal, but also not too risky.
I have not looked into it myself, and with this pull we are definitely better off than without.
I will merge now, as I am about to make the 4.1 release

PR #663, opened 2017-03-13
Convert pre-computed operator list A into a dict

quantshah:
This PR is to discuss the conversion of the pre-computed list of operators "A" passed to the smesolver functions (D) to a dictionary with descriptive key values. The initial commit is just to start the discussion. What should be the key values.


ajgpitch:
superseded by #815

PR #668, opened 2017-03-15
Handle mp num cpus on init

ajgpitch:
The `from scipy.misc import factorial` generated some error when building win py27 builds in conda-forge/qutip-feedstock#18. So I moved it to the only function in `utilities` that used it. Not strictly good practice. Would perhaps be better to move `_version2int` and `_blas_info` to another module, but quite a bit more work and I was not sure what to call the module: `sysutils`?

The error came from inside scipy due to a call to `multiprocessing.cpu_count()`. As we had one of those inside `__init__()` I wrapped that up as well.


nonhermitian:
This looks fine. Although, not sure the multiprocessing imports needs to be in try as it is a standard lib.  The issue is the the underlying cpu count code cant find the number of cpus.  This can occur and is mentioned in the mp docs.  I think just catching the exception is fine.

Also curious as to why factorial is causing problems, but that can be saved for another day.

nonhermitian:
Close my pull request, as this one covers everything we need.

ajgpitch:
Looks like the `factorial` import made a call to `num_cpus`

PR #673, opened 2017-03-22
ENH: Minor updates

nonhermitian:
- Force fast_csr_matrix shape to be ints (just in case).
- Get rid of annoying -Wstrict-prototypes flag not for c++ warning.
-


ajgpitch:
Good work Paul,
will be glad to see the back of that warning

PR #676, opened 2017-04-03
ENH : Time-Dependent Bloch-Redfield Solver

nonhermitian:
This is the Pull that will contain the time-dependent Bloch-Redfield solver.


nonhermitian:
It seems the checks are not passing for the openblas build due to an issue with SciPy scipy/scipy/#7295.

nonhermitian:
All of the individual terms in the dense version of the time-dependent BR solver are now tested and working.  Now I need to write the codegen module.  A sparse version is just a simple extension once the codegen is done.

nonhermitian:
The solver seems to be working nicely, and passes all the tests that I have come up with.  The last few tests still need to be added, interpolation functions should be able to be used for H and c_op terms, and the docstrings need to be updated as there is a new preferred way to input the a_ops that eliminated the need for a spectra_cb kwarg.

Issue #678, opened 2017-04-04
"Input" and "output" states in quantum circuit


sahmed95:
We should modify the `QubitCircuit` class to have input and output states specified for a quantum circuit. Currently the `QubitCircuit` class has only the following variables - N, reverse_gates, gates, U_list. We could add two more variables - input_states, output_states which could contain the input and output states as a list or a dictionary with the qubit number as key and the state as the value. This can be added to QubitCircuit [here](https://github.com/qutip/qutip/blob/master/qutip/qip/circuit.py#L168). 

Then we can modify the function [`latex_code`](https://github.com/qutip/qutip/blob/master/qutip/qip/circuit.py#L939) to add in the input and output states similar to how the gates are added so the qcircuit latex package compiles it.

The original [discussion](https://groups.google.com/forum/#!topic/qutip/QRu54GlED9A) can be found in the QuTiP Google help group where Morten Kjaergaard suggested this. 

To do
-----

- [x] Add input and output states to `QubitCircuit` class.
- [x] Implement latex visualisation of input and output states.
- [ ] Create a dictionary to label known states and get their representation using the `qubit_states` function for computation. {"+": (|0> + |1>)/sqrt(2), "0": 0}. This can be used with the `qubit_state` function to get a `Qobj` so that it can be used for computing.
- [ ] Add a method to get output states from input states by acting the propagators of the circuit on the given input states.

sahmed95:
As a hack, I created new gates for the input and output states.

![screen shot 2017-04-04 at 4 14 34 pm](https://cloud.githubusercontent.com/assets/6968324/24653828/153f4f96-1955-11e7-8dc2-99f97ffc828c.png)

sahmed95:
It also seems like we have a [`qubit_states`](https://github.com/qutip/qutip/blob/master/qutip/qip/qubits.py#L41) function which returns a tensor give the input states. This function is not used anywhere right now. Maybe it can be used to get the output state by acting the circuit's propogator function on the input state. But, for large circuits this may be an issue it may take a long tome to calculate the propagators and get the output. Instead, we could initialise the output to `None` or a user provided list and only if `output_states = True` the output is calculated.

sahmed95:
Closed via #679

PR #679, opened 2017-04-04
Adds input and output states to `QubitCircuit`

sahmed95:
This is an attempt to solve #678. The `QubitCircuit` class now has a method to add input and ouput states to a circuit. By default, all inputs and outputs are initialised to None. A new input state can be added similar to adding a gate by using the new function `add_state`. The user can specifying the state, the list of qubits where it should be added and the `state_type` as `input` or `output`.

States can be anything provided they are entered as a string.
```
qc = QubitCircuit(3)
qc.add_state("0", targets=[0])
qc.add_state("+", targets=[1], state_type="output")
qc.add_state("-", targets=[1])
```

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -1019,10 +1019,13 @@ class QubitCircuit(object):
             col.append(r" \qw ")
             rows.append(col)
 
+        input_states = ["\lstick{\ket{"+ x + "}}" if x is not None else "" for x in self.input_states]
+
         code = ""
         n_iter = (reversed(range(self.N)) if self.reverse_states
                   else range(self.N))
         for n in n_iter:
+            code += r" & %s" % input_states[n]
             for m in range(len(gates)):
                 code += r" & %s" % rows[m][n]
             code += r" & \qw \\ " + "\n"
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -36,7 +36,7 @@ import warnings
 
 from qutip.qip.circuit_latex import _latex_compile
 from qutip.qip.gates import *
-
+from qutip.qip.qubits import qubit_states
 
 __all__ = ['Gate', 'QubitCircuit']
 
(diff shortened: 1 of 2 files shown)


sahmed95:
@ajgpitch Hi, this adds input and output states to the QubitCircuit class. The next bit is to get the latex circuit visualisation for this. I will get that working as soon as I can. I guess there is not testing for that and I have to check it by trying out different plots. The next bit will be to get a list of known states and create a map so that they can be converted to a proper `Qobj`. Finally, we should have a method in `QubitCircuit` to calculate the output states, given an initial input state and add it to the circuit. But, I guess I could work on that in a different PR. All this is dicussed in the orginal issue #678

sahmed95:
Hi, this is ready for review. We now have the option to add input states to a circuit and the visualisation for the same. If the particular wire does not have an input, it is ignored.

Circuit with all input states defined

screen shot 2017-04-04 at 11 00 51 pm

Circuit with only one input state

screen shot 2017-04-04 at 11 04 28 pm

mortenkjaergaard:
This is great, thank you! 

Would it possible to instead of just accepting strings, it would also be able to accept LaTeX code when defining the state? E.g. often you'd want to send in a state $\psi_\text{in}$ and have a state $\psi_\text{out}$?

sahmed95:
Hi @mortenkjaergaard. Thanks for the suggestion. This code is "latex proof" as in any valid latex code can be directly passed for defining the state. Have a look at the following. You don't have to use the dollar sign as qcircuit takes care of it.

screen shot 2017-04-05 at 3 50 10 am

mortenkjaergaard:
Fantastic @sahmed95! I would love to know how to pull this into my own release of QuTiP. But potentially this is not something we should discuss here in this thread?

Thanks again for the swift and beautiful resolution 👍

sahmed95:
Well the most naive way I can think of is to overwrite the `circuit.py` file in your local installation of qutip with this new `circuit.py` file. It is located in `qutip/qutip/qip` in your local python package installation folder. If you can't find it then you can go for a fresh install with this file

- Remove your current qutip version with a `pip uninstall qutip`
- Download the latest qutip version from the [website](qutip.org) and replace the `circuit.py` file in `qutip/qip` folder with the one from here.
- Reinstall qutip by running `python setup.py install` from your qutip folder.

Of course there are better ways to do this such as pulling this branch locally and updating your qutip version. If you need any help ping me on our gitter channel https://gitter.im/qutip/Lobby.

ajgpitch:
This looks really good to me. I am happy to merge it now if @sahmed95 is happy with that

ajgpitch:
@mortenkjaergaard once merged there are documented procedures for you to install the latest version from this repo

mortenkjaergaard:
@ajgpitch great, yes, those I am familiar with. If you choose to to merge this, it will be even easier to update my qutip. Thanks!

sahmed95:
You can go ahead and merge it. The tests are passing and I checked the
latex code and visualization myself.

mortenkjaergaard:
Hi again @ajgpitch and @sahmed95.

I had a chance to work a little with the ".add_state" feature this morning, and wanted to raise an issue in terms of the visualization. The problems are shown in the figure below:
image
(NB: this could just be due to jupyter notebook? In any case, I see the problem is also reproduced in @sahmed95's post here: https://github.com/qutip/qutip/pull/679#issuecomment-291652067)

Two things:
1) The cropping seems to go bonkers if you use anything else than one symbol in the ket-state
2) It's not possible to use \textrm{xxx} in the ket-names.

sahmed95:
Hi, the cropping issue was there before. Even I had faced it when
visualizing circuits. I used to just add a redundant qubit to the circuit
to get that extra padding. I ll look into the cropping and resize it to
make the visualization better. Could you please open a new issue for that?

PR #683, opened 2017-04-07
Improve analytic calculation of coherent states.

bastikr:
Instead of calculating the alpha**n/sqrt(n!) for every entry, calculate
it recursively by x_{n+1} = x_{n}*alpha/sqrt(n), thereby avoiding the
explicit calculation of the factorial. This way it works also for
dimensions>400 where it would fail before.

This comes with a speed improvement:
benchmark_coherentstate

### DIFF
--- a/qutip/states.py
+++ b/qutip/states.py
@@ -123,11 +123,6 @@ def qutrit_basis():
     return np.array([basis(3, 0), basis(3, 1), basis(3, 2)], dtype=object)
 
 
-def _sqrt_factorial(n_vec):
-    # take the square root before multiplying
-    return np.array([np.prod(np.sqrt(np.arange(1, n + 1))) for n in n_vec])
-
-
 def coherent(N, alpha, offset=0, method='operator'):
     """Generates a coherent state with eigenvalue alpha.
 
@@ -188,9 +183,14 @@ def coherent(N, alpha, offset=0, method='operator'):
     elif method == "analytic" or offset > 0:
 
         data = np.zeros([N, 1], dtype=complex)
-        n = arange(N) + offset
-        data[:, 0] = np.exp(-(abs(alpha) ** 2) / 2.0) * (alpha ** (n)) / \
-            _sqrt_factorial(n)
+        if offset == 0:
+            data[0] = np.exp(-abs(alpha)**2 / 2.0)
+        else:
+            s = np.prod(np.sqrt(np.arange(1, offset + 1))) # sqrt factorial
+            data[0] = np.exp(-abs(alpha)**2 / 2.0) * alpha**(offset) / s
+        for n in range(N-1):
+            data[n+1] = data[n]*alpha/np.sqrt(offset + n + 1)
+
         return Qobj(data)
 
     else:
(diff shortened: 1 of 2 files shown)


ajgpitch:
Looks good to me. @bastikr could you please explain the difference between the green and orange lines. And what is the red line that seems to trump all?

bastikr:
The green line is after the first commit of this pull request where I used a for loop. The orange  line is after the second commit when I switched to using numpy.cumprod.
The red line is from the quantum optics framework written in julia that I'm currently working on (https://github.com/bastikr/QuantumOptics.jl). While doing some benchmarks I realized that for large N qutips coherent function failed which is why I wrote this patch. I don't know where the constant offset in the benchmark originates from but I think the speed is fine now, creating coherent states is probably never a bottleneck anyway.

ajgpitch:
Thanks for the explanation.
Happy to merge if no-one has any further comments

nonhermitian:
I am happy with this.  However, the analytic method is rarely used as it does not give correct probabilities.  An interesting thing to check would be if the analytic coefficients, after normalization, return the same matrix values as the standard coherent call.  That would be useful as the expm could go.

bastikr:
What do you mean by it does not give correct probabilities? As long as the cutoff is chosen appropriately both yield the same result. When the cutoff is too small they both give wrong results, which however have different properties. For the displacement operator method the state is normalized but P(n) is not a poissonian distribution anymore and if you look at the q-function of this state it will look considerably wrong. The analytic method doesn't have this problem but of course is unnormalized. None of both methods is in my opinion really superior.  I also think that returning non-normalized states is fine since this gives a rather obvious indication that something, in this case the choice of cutoff, was chosen inappropriately.

nonhermitian:
What you are asking about is a common issue when trying to simulate infinite dimensional systems, and has been asked many times.  In short, for self-consistency, you must begin with the assumption that your Hilbert space is finite, built the corresponding operators, and go from there.  You would check convergence by increasing the state space size until a desired tolerance is met.  Using the infinite analytic formulas are incorrect for the truncated space, but people use them a lot since they are familiar.

bastikr:
It seems to me that what counts in the end is how good the result of the simulation in the truncated Hilbert space coincides with the exact solution for the infinite Hilbert space. As an example I tried it out on a pumped mode in a cavity (https://gist.github.com/bastikr/136908ce690b299e2e82bd13ab2cb47a). Both approaches, i.e. using the analytically generated coherent state or generating it with the displacement operator as initial states, yield pretty much the same error compared to the analytic result. I may have made a mistake or simply misunderstood your statement. Can you maybe further help me to understand this issue?

nonhermitian:
The difference is that you are doing inconsistent simulations unless the Hilbert space size is large enough. By inconsistent I mean state vectors that do not have unit probability, your displacement operator would not return the same state as a call to coherent... In contrast, QuTiP always does a self-consistent simulation regardless of Hilbert space size. I dont concern myself with textbook formulas derived in an infinite dimensional space, as that is not where my system lives.  Convergence is always checked by doing self-consistent simulations of increasing size.

If you, or the user, is not guaranteed the standard rules apply for all system sizes, then it will come back to bite you.  Of course, for your own project, you are free to do as you please.  Just think of it as some advise from someone who has been doing this for seven years now.

nonhermitian:
Merging now.  Many thanks.

PR #684, opened 2017-04-08
Fixes #682: Reverse circuits

### LINKED ISSUE #682: Reverse Circuit doesn't work
Whenever i try to reverse some Circuit it throws an exception telling that temp does not have append method implemented.
I checked the source code and i think that instead o append the developers meant add_gate.

sahmed95:
The `reverse_circuit` function was broken as the gates should have been added to the temporary variable's `QubitCircuit.gates`. A better way is to use the `add_gate` method.

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -367,8 +367,8 @@ class QubitCircuit(object):
         """
         temp = QubitCircuit(self.N, self.reverse_states)
 
-        for i in range(self.N):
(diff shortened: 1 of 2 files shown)


sahmed95:
@ajgpitch

ajgpitch:
Looking good to me
@sahmed95 I will merge if you say it's ready.

sahmed95:
Hi, this is ready to be merged from my side!

PR #685, opened 2017-04-11
ENH: Faster tidyup

nonhermitian:
Tidyup is now done in Cython and is an in-place operator (10x+ faster).


ajgpitch:
Looks good

nonhermitian:
At some point this function should also do the eliminate zeros step, so we do not have to loop twice.

PR #691, opened 2017-05-04
Tiny bug-fix in def add_circuit(self, qc, start=0):

IBM-Research-Ismail:
The semantics seems incorrect.

qc.N should be used instead of len(qc.gates), we are checking if the added circuit (with a certain number of tracks) fits onto the parent circuit.

The number of gates don't feature.

Before:
<         if self.N - start < len(qc.gates):
After:
>         if self.N - start < qc.N:

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -285,7 +285,7 @@ class QubitCircuit(object):
             The qubit on which the first gate is applied.
         """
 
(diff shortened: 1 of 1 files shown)


sahmed95:
@ajgpitch This can be closed via #693

nonhermitian:
This has been superseded.

PR #693, opened 2017-05-07
Tiny bug-fix in def add_circuit(self, qc, start=0):

essence-of-waqf:
Apologies for re-doing the pull request, I had to delete my other account.

qc.N should be used instead of len(qc.gates), we are checking if the added circuit (with a certain number of tracks) fits onto the parent circuit.

The number of gates don't feature.

Before:
if self.N - start < len(qc.gates):
After:
if self.N - start < qc.N:

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -284,8 +284,7 @@ class QubitCircuit(object):
         start : Integer
             The qubit on which the first gate is applied.
         """
-
-        if self.N - start < len(qc.gates):
+        if self.N - start < qc.N:
             raise NotImplementedError("Targets exceed number of qubits.")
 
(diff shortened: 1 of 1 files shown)


sahmed95:
Hi, thanks for the PR. Let me see if there is a test written to check this. You may add your own test to ensure this works fine. 
@ajgpitch

ajgpitch:
@sahmed95 ok, just to be clear, are we saying merge this and close #691

sahmed95:
@ajgpitch Yes ! We can close #691 as the account associated with that is deleted by @essence-of-waqf

essence-of-waqf:
Thank you very much, now I can dockerise and pull from qutip git :)

PR #695, opened 2017-05-17
ENH: Faster td_func based me and se solve

nonhermitian:
Do not add sparse matrices.  Instead, directly do a spmv for each term.


nonhermitian:
Merging now.

PR #699, opened 2017-05-31
BUG: integer overflow check in tensor product

Vutshi:
Tensoring of large objects can lead to crash because index runs out of int32. Now it crashes gracefully and reports the overflow problem.

### DIFF
--- a/qutip/cy/spmath.pyx
+++ b/qutip/cy/spmath.pyx
@@ -348,7 +348,7 @@ def zcsr_kron(object A, object B):
     cdef int rowsB = B.shape[0]
     cdef int colsB = B.shape[1]
 
-    cdef int out_nnz = dataA.shape[0] * dataB.shape[0]
+    cdef int out_nnz = _safe_multiply(dataA.shape[0], dataB.shape[0])
     cdef int rows_out = rowsA * rowsB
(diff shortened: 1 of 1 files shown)


nonhermitian:
Ahh, someone has finally hit the int32 limit.  

The Pull looks fine to me.  However, the real solution is to used fused types in the Cython code so that both int32 and int64 can be handled seamlessly.  This has been something that I have known will need to be done for sometime now, but have pushed off until it was needed.

I think that this Pull should suffice until the fused types are implemented throughout.

Vutshi:
I know nothing about the fused types, but it sounds like a cool solution.
Maybe there is a better solution for the huge quantum objects. Soon we will see.

ajgpitch:
@nonhermitian can we merge this?

nonhermitian:
We can, but I think it is best to wait.  This is just one example of where
we need to add int64 support.  It is going to be a bit of work to do that,
so should wait till next release.

Vutshi:
The problem is that at the moment this error crashes the entire python kernel unexpectedly for a user without any messages. It can be very frustrating for a newbie.

nonhermitian:
Ok no worries, lets merge it in there.

-P

PR #704, opened 2017-06-22
Solve index error in function wigner cmap

ishxiao:
# solve the IndexError when call the function wigner_cmap
By trying to converter  a float number to an int number
zero_pos = int(zero_pos)  # solve the IndexError


nonhermitian:
I believe that this was fixed i Pull #703.

PR #708, opened 2017-07-02
Fix deprecated call to NumPy linspace in mesolve test.

avandeursen:
Since NumPy 1.12, `np.linspace` raises a DeprecationWarning when its third `num` parameter cannot be safely interpreted as an integer. This method is used in the `mesolve` tests.

Fixed by feeding the calls to `linspace` 100 instead of 100.0.

The warning (in Travis CI):
 
image

The deprecation: https://github.com/numpy/numpy/releases/tag/v1.12.0


nonhermitian:
Many thanks for this.  Merging now.

PR #709, opened 2017-07-04
Fix creation of swap gates in qft_gate_sequence (#688)

### LINKED ISSUE #688: Quantum Fourier Transform error
Hello, i discovered an error in the implementation of the QFT. Namely when i want to obtain the gate sequence. The is when tha code runs this 

if swapping is True:
            for i in range(N // 2):
                qc.add_gate(r"SWAP", targets=[i], controls=[N - 1 - i, i])

it gives an error because the swap  gate needs two targets and it is only passing one.

Here is a test code

qft.qft_gate_sequence(N=2)

avandeursen:
This fixes #688, following the suggestion by @mrcalixe.

The fix sets the swap targets to `[N - i - 1, i]`, instead of setting a target and a control separately.

The method `qft_gate_sequence` was untested: This pull request provides various test cases of varying length (1-6) and with and without swapping enabled.

### DIFF
--- a/qutip/qip/algorithms/qft.py
+++ b/qutip/qip/algorithms/qft.py
@@ -101,7 +101,7 @@ def qft_steps(N=1, swapping=True):
                 U_step_list.append(cphase(np.pi / (2 ** (i - j)), N,
                                           control=i, target=j))
             U_step_list.append(snot(N, i))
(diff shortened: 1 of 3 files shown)


nonhermitian:
Thank you for this one as well!

PR #716, opened 2017-07-13
ENH: Qtrl time depend ctrl dyn gen

ajgpitch:
Can now set different controls for a each timeslot. This allows time-varying controls to be used in pulse optimisation


ajgpitch:
As this is a self contained (to control modules only) PR, then I will merge this is a few hours.

PR #720, opened 2017-07-14
BUG: Updates to MKL solver and steady state module

nonhermitian:


nonhermitian:
The particular example that was given as an MKL error, is a tricky problem.  Seems that the MKL cannot solve for it in the `power` steady mode.  As such, I have implemented a 'solver' kwarg that allows for picking the solver that is used in the direct and power methods.  The default is SciPy, which could be the usual SuperLU solver, or UMFPACK, if installed via scikits.umfpack.  The 'mkl' option can be passed to do the mkl library.

I will see what the Intel people have to say about solving the particular matrix that was given:

https://groups.google.com/d/msg/qutip/axGZ-7vtnhI/2Rlsl7ExAgAJ

nonhermitian:
I have submitted a bug report to Intel:

 https://software.intel.com/en-us/node/738375

nonhermitian:
I believe that I am going to rework this into a new Pull.  As such, this should not be merged.

nonhermitian:
Intel seems to think that this is a bug as well.

nonhermitian:
Closing this as it is superseded

PR #721, opened 2017-07-17
Reset master to Beta non-released and bump point release

ajgpitch:


ajgpitch:
@nonhermitian There is some BR test failing here. Do I need to worry about this? What are the two Travis tests now?

nonhermitian:
I am guessing that the test has too low a tolerance.  Also not sure about the two tests.  Didnt see it a day or two ago.

ajgpitch:
Ok, thanks for restarting the test, we'll see what happens

PR #724, opened 2017-07-21
BUG: rcsolve imported old Odeoptions

nonhermitian:
Now imports the correct Options class. 

Probably need a good unittest for this @nwlambert.


nwlambert:
Will do.  Also going to update a few things in the function itself soon.

PR #726, opened 2017-07-21
Remove star import in tests

ajgpitch:
Only test_mcsolve done for now.


ajgpitch:
When running nosetest with just test_mcsolve in a folder, on Windows, I found that somehow the tests appeared to respawn recursively. For some reason this change seemed to fix that problem.

ajgpitch:
There are ten other test files with `from qutip import *`. I will fix these soon, as this may fix the problem we are experiencing with tests on Windows with the release we are trying to make

ajgpitch:
It seems that #729 fixed the strange recursively calling of the tests mentioned above. Hence I will close this as it has nothing to offer

sahmed95:
Just an observation, in a lot of places I have seen the use of `from qutip import *` (examples, code) which is supposed to be a bad practice. Should we open an issue to clean that up ?

ajgpitch:
I think we should open up an issue to discuss it. I am keen that we remove them

ajgpitch:
But perhaps we should do it in small PRs like this one, rather than one big one

Issue #723, opened 2017-07-20
Installation error: "ambiguous overloaded method"


kcgthb:
Hi!

I'd like to report an installation issue on CentOS/RHEL platforms. I've tested various Python versions (2.7, 3.4) on both RHEL 6.x and 7.x, and the error is consistent, and happens during compilation of some Cython files.

The error is: `ambiguous overloaded method` referring to the `abs()` function, and happens in two different places.

Full installation output:
```
$ pip install qutip
Collecting qutip
  Using cached qutip-4.1.0.tar.gz
    Complete output from command python setup.py egg_info:
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        cdef double * col_sum = <double *>PyDataMem_NEW_ZEROED(ncols, sizeof(double))
        cdef double max_col = 0
        for ii in range(nrows):
            for jj in range(ptr[ii], ptr[ii+1]):
                k = ind[jj]
                col_sum[k] += abs(data[jj])
                                ^
    ------------------------------------------------------------
    
    qutip/cy/sparse_utils.pyx:316:29: ambiguous overloaded method
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        cdef size_t ii, jj
        cdef double * row_sum = <double *>PyDataMem_NEW_ZEROED(nrows, sizeof(double))
        cdef double max_row = 0
        for ii in range(nrows):
            for jj in range(ptr[ii], ptr[ii+1]):
                row_sum[ii] += abs(data[jj])
                                 ^
    ------------------------------------------------------------
    
    qutip/cy/sparse_utils.pyx:335:30: ambiguous overloaded method
    Compiling qutip/cy/spmatfuncs.pyx because it depends on /home/kilian/qutip/lib/python2.7/site-packages/Cython/Includes/numpy/__init__.pxd.
    Compiling qutip/cy/stochastic.pyx because it depends on /home/kilian/qutip/lib/python2.7/site-packages/Cython/Includes/numpy/__init__.pxd.
    Compiling qutip/cy/sparse_utils.pyx because it depends on /home/kilian/qutip/lib/python2.7/site-packages/Cython/Includes/numpy/__init__.pxd.
    Compiling qutip/cy/graph_utils.pyx because it depends on /home/kilian/qutip/lib/python2.7/site-packages/Cython/Includes/numpy/__init__.pxd.
[...]

PR #731, opened 2017-07-25
Fixes for Cython 0.26

nonhermitian:
Cython 0.26 changed a few internal items, and was causing come issues with our builds.  This should alleviate those problems.


ajgpitch:
As described in https://github.com/qutip/qutip/issues/723#issuecomment-318100070
I have tested this and it seems to fix #723, so I would suggest we merge this

ajgpitch:
Just finished all tests on Centos with cython 0.26 - all passed

nonhermitian:
Ok then lets merge it.

PR #732, opened 2017-07-27
4.2 release ready

sahmed95:
Added changes from master for new release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ as well as in the classroom.
 DOCLINES = __doc__.split('\n')
 
 CLASSIFIERS = """\
-Development Status :: 5 - Production/Stable
+Development Status :: 4 - Beta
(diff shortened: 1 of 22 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
 DOCLINES = __doc__.split('\n')
 
 CLASSIFIERS = """\
-Development Status :: 5 - Production/Stable
+Development Status :: 4 - Beta
```
ajgpitch: This needs to be `5 - Production/Stable`

setup.py
```diff
@@ -58,7 +58,7 @@
 MAJOR = 4
 MINOR = 2
 MICRO = 0
-ISRELEASED = True
+ISRELEASED = False
```
ajgpitch: This needs to be `True`

PR #733, opened 2017-08-02
ENH: Use OPENMP for tidyup if installed.

nonhermitian:
It is much faster to use openmp to do the tidyup operation when NNZ > ~500.  Also, in many cases the tidyup operation does not actually result in any zero values in the sparse data, and as such, it is unnecessary to run the eliminate_zeros function.  Tidyup now returns a flag that indicates whether eliminate_zeros should be run.


nonhermitian:
I was not shortening the output data and indices arrays in zcsr_mult.  This was being caught, and fixed, by tidyup.  I have fixed that issue here as well.

PR #735, opened 2017-08-03
Remove block of code that traced out 1d systems

nonhermitian:
This trace out only occurred when adding system, and not things like mult.   Caused several confusing errors.


ajgpitch:
@nonhermitian this seems to generate some errors, are you able to fix?

nathanshammah:
I resolved the conflict. Maybe @Ericgig can you look into it and decide whether to merge or close?

jakelishman:
This PR is obviously very old now, and the state of the library as it is now means that there are a few places where this behaviour is assumed.  I'd consider this to be a breaking change for the 4.x cycle right now, so I'll close this PR.

Note that we _are_ actually merging this functionality in the future - it already exists in `dev.major`, where we took care to handle all the uses of it in the rest of the library.  See commit 85afb6f in PR #1332 (though you're probably better looking in the git log - that's a big PR that includes a merge commit.  I started doing merges in separate PRs after that one...).

PR #736, opened 2017-08-04
ENH: 'clenshaw' method by default for plot_wigner

Vutshi:
Make the useful function even better


Vutshi:
Btw, the function creates a plot with a weird aspect ration 2:1 by default. May I change it to a more reasonable 1:1?

nonhermitian:
Sounds good to me.

PR #747, opened 2017-08-23
New attribute evo_full_initial added to OptimResult

ajgpitch:
It is the full evolution based on the initial pulse
It is set when the optimisation starts


ajgpitch:
This is small change to add a new attribute that is convenient in my current work. Presumably would be to others too

PR #746, opened 2017-08-18
calc_omega comments updated

ajgpitch:
clarity and numpydoc compliance.

PR #749, opened 2017-08-30
Fixes #746: Trace norm for non-herm

sahmed95:
This is the most naive way perhaps to do this. But since it uses Qobj functions to do it I am guessing it is the most efficient and takes into consideration sparse matrices etc.

@ajgpitch


nonhermitian:
This should be superseded by #751.

Issue #748, opened 2017-08-25
Trace norms incorrect for non-Hermitian operators


ngedwin98:
The implementation of the trace norm (as called through the `norm` function) produces incorrect results for non-Hermitian operators. For an operator `A`, the [trace norm is defined as](https://www.quantiki.org/wiki/trace-norm)
```
(A*A.dag()).sqrtm().tr()
```
However, `qutip` currently provides the following behavior: 
```python
>>> a = destroy(10)
>>> (a*a.dag()).sqrtm().tr()
19.30600052603572
>>> a.norm()
0.0
```
```python
>>> s = sigmap()
>>> (s*s.dag()).sqrtm().tr()
1.0
>>> s.norm()
0.0
```
Aside from the discrepancies above, there is also the problem that an operator should have zero norm if and only if it is the zero operator. This mathematical requirement is not satisfied by the current implementation of `norm`.

Based on a cursory inspection, it appears the issue arises from the following lines:
https://github.com/qutip/qutip/blob/6eb6a2762831129ca2dedd2d9ea00c8265f6c01e/qutip/qobj.py#L956-L958
Here, the trace norm is calculated as the sum of absolute values of the eigenvalues. While this formula is correct for Hermitian operators (like a density matrix), it is not correct for non-Hermitian operators, since the singular values are not in general equal to the magnitudes of the eigenvalues.

sahmed95:
Hi, Thank you for bringing this to notice. I guess there should have been an `isherm` check to determine the trace norm using `(a*a.dag()).sqrtm()` instead of just summing over absolute eigenvalues based on the operator. Let me open a PR to fix this.

@nonhermitian @ajgpitch

nonhermitian:
This was fixed in #751.

PR #751, opened 2017-08-30
BUG: Fix for trace-norm bug

nonhermitian:
This is a fix for the trace-norm bug, as reported in Issue #748.  Unlike Fix #749, this minor correction works for all operators, and does not do the sqrtm operation.


nonhermitian:
@sahmed95, Please take a look.

sahmed95:
Of course this is a better way to make sure it works for all operators. We can close the other PR and merge this. 

The unit tests could be added though.

sahmed95:
Hi, this looks good. I closed the other PR.

Issue #664, opened 2017-03-14
Reminders for users to cite qutip


ajgpitch:
There is some concern that some users are forgetting to cite qutip when using it in their work. This thread is for ideas of how to add reminders into the library.

ajgpitch:
One idea was to randomly print a message on `import qutip`. I am not sure on whether this would go against established etiquette. Anyone have a view on this?

ajgpitch:
Adding a reminder in `qutip.about()` would seem to be a good idea, but not sure how often it would be seen

ajgpitch:
Having reminder in the repr of the Result object would seem appropriate.
"If this result is interesting, please cite QuTiP in your publication"

sahmed95:
Something to also make citing easier such as a bib tex generator? A
qutip.cite() call within the result class that makes a .bib file where the
plots are saved or code is run. That way it will surely come to the users
notice when he checks out the saved plots / code.

ajgpitch:
@sahmed95 Yes, very nice idea. Making life easier for people is always a good idea. We could have something on the website as well that generates bibtex for the qutip papers and the Zenodo DOI that people can copy direct into their bib files, like on the Zenodo site.

sahmed95:
I use doi2bib.org. Using the Qutip 2 doi we get this :

http://www.doi2bib.org/#/doi/10.1016/j.cpc.2012.11.019

@article{Johansson2013,
  doi = {10.1016/j.cpc.2012.11.019},
  url = {https://doi.org/10.1016%2Fj.cpc.2012.11.019},
  year  = {2013},
  month = {apr},
  publisher = {Elsevier {BV}},
  volume = {184},
  number = {4},
  pages = {1234--1240},
  author = {J.R. Johansson and P.D. Nation and Franco Nori},
  title = {{QuTiP} 2: A Python framework for the dynamics of open
quantum systems},
  journal = {Computer Physics Communications}
}

Phyks:
My two cents on this but if providing an easy citation formatting function, I think it would be relevant to have the QuTiP version mentionned in the citation. Many QuTiP citations I see in papers just mention something like "Thanks to QuTiP developpers, all simulations were made under QuTiP", without any explicit mention of version. Having a versionned citation might help track the impact of potential bugs in QuTiP code on results.

ajgpitch:
@Phyks I agree that it would help having a version specific citation, for the repeatability reasons you mention. We do have a Zenodo DOI
[...]

PR #752, opened 2017-09-19
WIP: Citation message and bibtex generator

sahmed95:
As discussed in #664 this PR is regarding citation reminders for QuTiP as it seems that many times users are unaware of the papers to cite when they use QuTiP.

I have added a bibtex file in the source now and into the MANIFEST file. We should discuss how to get this to the users attention. Creation of a cite() function which generates a bibtex entry in the working directory whenever qutip is imported could be one option. The other option is to put it in /home/qutip or print it out whenever a result class is called. We can also add a welcome message at setup which reminds users regarding the papers and points them to the bibtex file.

I am in favour of having a bibtex generator in __init__.py such that qutip.bib appears in the code folder whenever QuTiP is used.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include README.md
 include LICENSE.txt
 include requirements.txt
+include qutip.bib
--- a/qutip.bib
+++ b/qutip.bib
@@ -1,0 +1,26 @@
+@article{Johansson2013,
+  doi = {10.1016/j.cpc.2012.11.019},
+  url = {https://doi.org/10.1016/j.cpc.2012.11.019},
+  year  = {2013},
+  month = {apr},
+  publisher = {Elsevier {BV}},
+  volume = {184},
+  number = {4},
+  pages = {1234--1240},
+  author = {J.R. Johansson and P.D. Nation and Franco Nori},
+  title = {{QuTiP} 2: A Python framework for the dynamics of open quantum systems},
+  journal = {Computer Physics Communications}
+}
+@article{Johansson2012,
+  doi = {10.1016/j.cpc.2012.02.021},
+  url = {https://doi.org/10.1016/j.cpc.2012.02.021},
+  year  = {2012},
+  month = {aug},
+  publisher = {Elsevier {BV}},
+  volume = {183},
+  number = {8},
+  pages = {1760--1772},
+  author = {J.R. Johansson and P.D. Nation and Franco Nori},
+  title = {{QuTiP}: An open-source Python framework for the dynamics of open quantum systems},
+  journal = {Computer Physics Communications}
+}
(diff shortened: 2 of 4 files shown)


ajgpitch:
It would be good to include the zenodo bib reference that relates to the specific version that is being used. It would be good if users included this zenodo reference, as this pinpoints a version, which could be important for people trying to repeat the results.

In the same vein it could be a good idea to automatically output a file into the cwd that contains the `qutip.about()` output. We could include a citation reminder in the `qutip.about()` and also a message that explains that bib files have also be produced for their convenience.

In order to placate those who may object to all of this, we should provide some setting that allows users to turn off this bib and about() output. We should mention this setting in the `about()`

We also need to handle file errors that may occur.

ajgpitch:
We should avoid any file output related to Result objects. This could be costly in loops that produce a lot of results.

sahmed95:
Hi, we have the cite function in `qutip.about` now and can decide when to invoke it. I don't think users usually call the `about` function much so it may not be very effective to use it there. The current `cite` function generates a qutip.bib` file and prints it out. If the filename is not specified, it creates a `qutip.bib` file in cwd.

sahmed95:
@nonhermitian The test failure is with NOMKL=1 and OPENMP=0. Specifically, a solver is unavailable. This is not related to the addition of this function. Something else is breaking.

```
cvxpy.error.SolverError: Solver 'CVXOPT' failed. Try another solver.
```

sahmed95:
@Phyks Thanks for the suggestion. I made the qutip.cite() function to print and generate a bibtex file. Perhaps this is sufficient. With a message during install mentioning that you can invoke"
```
from qutip.about import cite
cite("path_to_generate_bibtex")
```
to get the bibtex file and printing the output on the screen. @ajgpitch Should I add this to the setup.py file. Can we move this PR forward?

ajgpitch:
@sahmed95 what are you thinking is needed in setup.py

ajgpitch:
I don't think we should print the whole bibtex content to screen by default when calling about(). I think a message to say.
"Please cite QuTiP in your publication - for you convenience bibtex file saved to:\n".join(filepath)

I think really we should provide some way to turn off the saving of the bibtex file in settings as well

ajgpitch:
I think it would be better if the article names were 'qutip1' and 'qutip2' rather than 'Johansson2013' and 'Johansson2012'

sahmed95:
In the current code there is no automatic generation of the bibtex files but a polite message to cite qutip and use the cite function to generate the same. I hope that is okay.

```
print("Please cite QuTiP in your publication - for you convenience a bibtex file can be easily generated using `qutip.about.cite()`"
```

ajgpitch:
Ok, I think I get it now. Seems good to me.
I did a few tiny edits. If all tests are good and @sahmed95 happy with my edits, then I will merge

sahmed95:
Thanks for taking a look. This is good for me and you can merge it once the tests pass. Thank you.

PR #755, opened 2017-09-26
ENH: Faster ket expectation values

nonhermitian:
This is a pure sparse based method for doing expectation values with kets.  About 10x+ faster than previous version.


nonhermitian:
Can get rid of kk for loop.

nonhermitian:
This seems good to go.

PR #758, opened 2017-09-28
BUG: Fix Travis CI

nonhermitian:


nonhermitian:
The issue was with the conda update that pulled a libgcc package that broke Travis

PR #762, opened 2017-10-04
Reduce computation time of `floquet.floquet_master_equation_rates`

Phyks:
The scalar product `(f_modes_t[a].dag() * c_op * f_modes_t[b])[0, 0]` is computed at each loop in the `for` loop on `k`. But it only depends on `a` and `b` and can then be computed outside of the loop.

In my case, I was dealing with Hilbert spaces of size 100, which resulted in a significantly reduced run time.

Please, let me know about any feedback about this PR :)


nonhermitian:
The Floquet module probably needs a ton of work, so any effort in that direction is much appreciated.

Phyks:
Thanks for the review, I'll update my MR accordingly :)

Phyks:
Not sure why Github is not taking my modifs into account in this MR. I'll close it and open https://github.com/qutip/qutip/pull/766 to track these changes. Sorry for the noise :/

PR #763, opened 2017-10-08
ENH: ~30x faster Q.overlap() [inner product]

nonhermitian:
Faster routine to compute inner products.  About 30x faster than previous versions.


nonhermitian:
This Pull was a bit tricky as some unrelated test failed when using OPENMP for no apparent reason.  The same code, when copied to a notebook, worked as expected.  Therefore, I modified the non-Hermitian eigen test, moving from a unitary to a complex sum of Hermitians for the non-Hermitian operator.  The tests now pass.

PR #766, opened 2017-10-17
Reduce computation time of `floquet.floquet_master_equation_rates`

Phyks:
The scalar product (f_modes_t[a].dag() * c_op * f_modes_t[b])[0, 0] is computed at each loop in the for loop on k. But it only depends on a and b and can then be computed outside of the loop.

In my case, I was dealing with Hilbert spaces of size 100, which resulted in a significantly reduced run time.

Please, let me know about any feedback about this PR :)

Follow up of !762.

### DIFF
--- a/qutip/floquet.py
+++ b/qutip/floquet.py
@@ -434,7 +434,7 @@ def floquet_state_decomposition(f_states, f_energies, psi):
         The coefficients :math:`c_\\alpha` in the Floquet state decomposition.
 
     """
-    return [(f_states[i].dag() * psi).data[0, 0]
+    return [f_states[i].overlap(psi)
             for i in np.arange(len(f_energies))]
 
 def fsesolve(H, psi0, tlist, e_ops=[], T=None, args={}, Tsteps=100):
@@ -615,12 +615,14 @@ def floquet_master_equation_rates(f_modes_0, f_energies, c_op, H, T,
 
         # f_modes_t = floquet_modes_t(f_modes_0, f_energies, t, H, T, args)
         f_modes_t = floquet_modes_t_lookup(f_modes_table_t, t, T)
-        for a in range(N):
-            for b in range(N):
+        for b in range(N):
+            ket_b = c_op * f_modes_t[b]
+            for a in range(N):
+                scalar_product = ket_b.overlap(f_modes_t[a])
                 k_idx = 0
                 for k in range(-kmax, kmax + 1, 1):
                     X[a, b, k_idx] += (dT / T) * exp(-1j * k * omega * t) * \
-                        (f_modes_t[a].dag() * c_op * f_modes_t[b])[0, 0]
+                        scalar_product
                     k_idx += 1
 
     Heaviside = lambda x: ((np.sign(x) + 1) / 2.0)
@@ -934,10 +936,10 @@ def fmmesolve(H, rho0, tlist, c_ops=[], e_ops=[], spectra_cb=[], T=None,
         An instance of the class :class:`qutip.solver`, which contains either
         an *array* of expectation values for the times specified by `tlist`.
     """
-    
+
     if _safe_mode:
         _solver_safety_check(H, rho0, c_ops, e_ops, args)
-    
+
     if T is None:
         T = max(tlist)
 


Phyks:
@nonhermitian I went for the `overlap` solution, as timing using this approach were better than using the `c_op.matrix_element(f_mode_a, f_mode_b)` solution.

For reference,

```
from qutip import *

N = 20
f_mode_a = coherent(N, alpha=1)
f_mode_b = coherent(N, alpha=0.5)
c_op = destroy(N)

for a in range(100):
    for b in range(100):
        c_op.matrix_element(f_mode_a, f_mode_b)
```

runs in 2.13s (± 0.03s)

and

```
from qutip import *

N = 20
f_mode_a = coherent(N, alpha=1)
f_mode_b = coherent(N, alpha=0.5)
c_op = destroy(N)

for a in range(100):
    ket_a = c_op * f_mode_a
    for b in range(100):
        ket_a.overlap(f_mode_b)
```
runs in 1.88s (± 0.04s).


Additionnally, I replaced a manual `.dag()` construction with an `overlap` call in `floquet_state_decomposition`.



Also, I have a few other changes to `floquet` module that I can contribute. I will open an issue to discuss them.

qutip/floquet.py
```diff
@@ -117,13 +112,7 @@ def floquet_modes(H, T, args=None, sort=False, U=None):
     else:
```
Phyks: Not sure this is intended to sort according to `-e_quasi`. I would have expected the sort to be made according to `e_quasi` so that energies are in increasing order. But I should have missed something :/

qutip/floquet.py
```diff
@@ -100,15 +99,11 @@ def floquet_modes(H, T, args=None, sort=False, U=None):
...
 
     eargs = angle(evals)
-
-    # make sure that the phase is in the interval [-pi, pi], so that
-    # the quasi energy is in the interval [-pi/T, pi/T] where T is the
-    # period of the driving.  eargs += (eargs <= -2*pi) * (2*pi) +
-    # (eargs > 0) * (-2*pi)
-    eargs += (eargs <= -pi) * (2 * pi) + (eargs > pi) * (-2 * pi)
+    # note: angle is in the interval [-pi, pi], so that the quasi energy is in
+    # the interval [-pi/T, pi/T] where T is the period of the driving.
```
Phyks: The comment was misleading, using `2 * pi` whereas the actual line below was using a mix of `pi` and `2*pi`.

From https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.angle.html and https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.arctan2.html#numpy.arctan2, it seems that the value returned by `angle` is always in the range `[-pi, pi]` anyways, so I guess there should not be any need for this correction.

qutip/floquet.py
```diff
@@ -100,15 +99,11 @@ def floquet_modes(H, T, args=None, sort=False, U=None):
         U = propagator(H, T, [], args)
 
     # find the eigenstates for the propagator
-    evals, evecs = la.eig(U.full())
+    evals, evecs = U.eigenstates()
```
Phyks: I am not well aware of the internal code of QuTiP, but it seemed to me as I read the code that `Qobj.eigenstates` will do the same as the previous line. Using it makes the eigenmodes ordering simpler at the end of the function call. Plus, I guess it is better to use a single method to compute eigenstates across the whole library, so that any part of the library will take advantage of potential enhancements made in eigenstates computation in the future.

Phyks:
I added another commit to this PR, working a bit on Floquet modes computation. I added comments in https://github.com/qutip/qutip/pull/766/commits/95a9c1f7014e394f00195f391b650599f7883c7b to better explain my reasoning for the proposed changes.

qutip/tests/test_floquet.py
```diff
@@ -67,6 +70,94 @@ def testFloquetUnitary(self):
 
         assert_(max(abs(sol.expect[0] - sol_ref.expect[0])) < 1e-4)
 
+    def testFloquetDissipativeEOps(self):
```
Phyks: These test cases are directly taken from the [docs](http://qutip.org/docs/latest/guide/dynamics/dynamics-floquet.html#floquet-theory-for-dissipative-evolution).

qutip/tests/test_floquet.py
```diff
@@ -67,6 +70,94 @@ def testFloquetUnitary(self):
...
+            f_modes_t = floquet_modes_t_lookup(f_modes_table_t, t, T)
+            p_ex[idx] = expect(num(2), output.states[idx].transform(f_modes_t,
+                                                                    True))
+
+        # For reference: calculate the same thing with mesolve
+        output = mesolve(H, psi0, tlist, [np.sqrt(gamma1) * sigmax()],
+                         [num(2)], args)
+        p_ex_ref = output.expect[0]
+
+        assert_(max(abs(np.real(p_ex) - np.real(p_ex_ref))) < 1e-1)
```
Phyks: Not sure exactly what happens with these computations.

In QuTiP 4.2.0 (installed from `pip`),

```
> max(abs(np.real(p_ex) - np.real(p_ex_ref)))
0.022044446457344991
```

whereas in QuTiP 4.3.0 (from the github `master` branch, installed using `pip install .`),

```
> max(abs(np.real(p_ex) - np.real(p_ex_ref)))
 0.098117562402980751
```

Not sure exactly what changed between these versions (but as far as I could test, this is pre-existing in the `master` branch before this PR).

qutip/floquet.py
```diff
@@ -702,20 +699,14 @@ def floquet_master_equation_tensor(Alist, f_energies):
         for J in range(N * N):
             c, d = vec2mat_index(N, J)
 
-            R = -1.0j * (f_energies[a] - f_energies[b])*(a == c)*(b == d)
```
Phyks: I think I might be missing something here, but I can't find where it comes from, while looking at eqs.(251-252) from the [Gri98] reference.

qutip/floquet.py
```diff
@@ -702,20 +699,14 @@ def floquet_master_equation_tensor(Alist, f_energies):
...
+            if a == b and c == d:
+                Rdata_lil[I, J] += sum(A[a, c] for A in Alist)
 
-                if dR != 0.0:
-                    Rdata_lil[I, J] += dR
+            if a == c and b == d:
+                Rdata_lil[I, J] += -0.5 * sum(
+                    sum(A[n, a] + A[n, b] for n in range(N))
+                    for A in Alist
+                )
```
Phyks: Reworked the way computation is done. Using my simple code from #772, tensor computation times drops from around 50 seconds to less than a second.

Output was the same as the previous code, except from the `f_energies` part.

qutip/floquet.py
Phyks: Actually, it looks like it was [added a few years ago](https://github.com/qutip/qutip/commit/6942deb25cee0114a7ac0c9f9098dfd40006cbbe).

@jrjohansson would you have more infos about it please? :)

qutip/tests/test_floquet.py
usmanmacer: This code is not working for other hamiltonians. Does it support two level systems only?

Phyks:
This PR is starting to diverge both from the qutip codebase and my own codebase, and I can't really commit to maintain this open and in sync. I'd rather close it than leave an out of sync PR open for too long. For anyone coming across this and having issues with Floquet code, I will post the latest version which is working fine for me in #772.

PR #774, opened 2017-10-25
Cythonizing mcsolve

Ericgig:
This contain a new solver for `mcsolve`, `dopri5` which don't go through scipy and is faster for small system.

It also contain a new class for time-dependent Qobj. It allow to create an object from the usual list of Qobj and string/function. The object can then be called to obtain the Qobj at a time. It is used in `mcsolve`.


qutip/cy/inter.pyx
```diff
@@ -0,0 +1,63 @@
...
+    if t > dt*(N-2):
+        return str_array_0[N-2]*(dt*(N-1)-t)/dt + \
+                str_array_0[N-1]*(t-dt*(N-2))/dt
+
+    # In the middle: 4th order polynomial approximation
+    cdef int ii = <int>(t/dt)
+    cdef double a = (t/dt - ii)
+    cdef complex approx = 0.
+
+    approx += (-a**3 + 3*a**2 - 2*a) * 0.1666666666666666 * str_array_0[ii-1]
```
Vutshi: This is a bit inefficient for two reasons. First, calculation of powers (like  a**2) is substantially slower than simple multiplication (a*a). Second, it is inefficient and numerically unstable to calculate polynomials in such a direct way. One can use Horner form of a polynomial instead.

qutip/cy/td_qobj_cy.pyx
```diff
@@ -0,0 +1,583 @@
...
+        if self.super:
+            return self._expect_mat_super(t, &vec[0], isherm)
+        else:
+            return self._expect_mat(t, &vec[0], isherm)
+
+
+cdef class cy_td_qobj(cy_qobj):
+    def __init__(self):
+        self.N_ops = 0
+        self.ops = <CSR_Matrix**> malloc(0 * sizeof(CSR_Matrix*))
```
nonhermitian: Calls to `malloc` should be replaced with `PyDataMem_NEW(size_t size)` from `numpy/arrayobject.h`. This is because the `malloc` called by the complied Cython code may not be the same as the `malloc` used when building Python.  For example, using GCC to compile QuTiP on Win.  Freeing data should also be done with ` PyDataMem_FREE(void * ptr)`.  Learned this the hard way.

qutip/td_qobj.py
```diff
@@ -0,0 +1,1111 @@
...
+        +,- : td_Qobj, Qobj
+        * : Qobj, C-number
+        / : C-number
+    and some common linear operator/state operations. The td_Qobj
+    are constructed from a nested list of Qobj with their time-dependent
+    coefficients.
+
+    *Examples*
+
+        H = [H0, [H1, 'sin(w*t)'], [H2, f1_t], [H3, np.sin(3*w*tlist)]]
```
nonhermitian: As it stands now, we do not support mixing str and Python func formats.  Of course we could, but then all the solvers would need to add this support.

nonhermitian:
- QuTiP already has some interpolation functions that can be compiled down into Cython.  Is there some reason we need new ones?  Does the new solver support the current interp functions?

- I don't see any changes / additions to unit tests.  Are these still coming?

- The new `td_qobj` objects should be able to be passed to all the solvers. This is most easily done by just looking for a `td_qobj` instance and then grabbing the `Q_object` from it.

qutip/cy/td_qobj_cy.pyx
Ericgig: I will make the replacement this week.

qutip/td_qobj.py
Ericgig: While the different format work together, there are not many advantages to mix then. In some methods, there are optimizations for only func format or only str format.  Should I just remove it from the documentation?

Ericgig:
I removed the malloc and rewrote the examples in a way to discourage the mixing of str and func format.

I have some unit tests for the td_Qobj, but I still have to add new ones for mcsolve.
It seems I do not have the same version of the pickle module than the tests server. Will check that next week.

nonhermitian:
Hey @Ericgig, what is the status of this Pull?

Ericgig:
The mcsolve is Ok. But I did not test it on windows yet.
td_Qobj work fine for the solvers. However, there are probably a few methods/properties that would be useful but I did not think about. If somebody would have the time to try it and send feedback, that would be great.
For now, sesolve and mesolve also accept td_Qobj. I am working, in another branch, to have the stochastic solvers work with it.

Ericgig:
Split in 2 other pull request: mcsolve and td_qobj. Should be removed.

PR #776, opened 2017-10-30
BUG: Fix partial_transpose

rganardi:
partial_transpose doesn't work properly if mask is boolean array instead of integer array.

### DIFF
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -165,7 +165,7 @@ def negativity(rho, subsys, method='tracenorm', logarithmic=False):
 
         Experimental.
     """
-    mask = [int(idx == subsys) for idx, n in enumerate(rho.dims[0])]
+    mask = [idx == subsys for idx, n in enumerate(rho.dims[0])]
     rho_pt = partial_transpose(rho, mask)
 
     if method == 'tracenorm':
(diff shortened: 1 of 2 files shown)


nonhermitian:
Many thanks for this.  What was the exact error that was occurring?  This might be better treated in the partial transpose call.

rganardi:
```
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/qutip/entropy.py", line 169, in negativity
    rho_pt = partial_transpose(rho, mask)
  File "/usr/lib/python3.6/site-packages/qutip/partial_transpose.py", line 82, in partial_transpose
    return _partial_transpose_dense(rho, mask)
  File "/usr/lib/python3.6/site-packages/qutip/partial_transpose.py", line 96, in _partial_transpose_dense
    np.array(rho.dims).flatten()).transpose(pt_idx).reshape(rho.shape)
TypeError: only integer scalar arrays can be converted to a scalar index
```

I guess `_partial_transpose_dense` was expecting `mask` to be an array of 0's or 1's, but negativity is passing in boolean array.

Now that I read the docstring for partial_transpose, I agree with you. Should I just cast the mask to integers in `partial_transpose`?

nonhermitian:
Yeah, probably.  Best to fix it at the root of the problem.

rganardi:
I've fixed it in `partial_transpose` in 82fd28f

nonhermitian:
I think this needs to be returned to at some point to clean up.  But the current solution works good enough for now.  Thanks for finding this.

PR #779, opened 2017-11-15
Added exception to use GCC in Windows if running in MSYS2

vnegnev:
When installing Qutip from source in an MSYS2 environment, the build flags should be set to GCC-style like in *nix installations - I updated the relevant files to do this. I just check for the existence of the MSYSTEM environment variable to check for this case.


nonhermitian:
I think this looks good. Has no idea about this new win+gcc install methods.  Thanks.

ajgpitch:
Yes, thanks @vnegnev

PR #780, opened 2017-11-23
hardware_info(): Replacing lscpu with /proc/cpuinfo

Zomtir:
lscpu is localisation dependant and breaks if the OS uses e.g. German. "Core(s) per socket" turns into "Kern(e) pro Socket" and results in a crash.

/proc/cpuinfo on the other hand remains the same same, therefore it is suited more as information source.

### DIFF
--- a/qutip/hardware_info.py
+++ b/qutip/hardware_info.py
@@ -58,21 +58,23 @@ def _linux_hardware_info():
     # get cpu number
     sockets = 0
     cores_per_socket = 0
+    frequency = 0
     for l in [l.split(':') for l in open("/proc/cpuinfo").readlines()]:
         if (l[0].strip() == "physical id"):
             sockets = np.maximum(sockets,int(l[1].strip())+1)
         if (l[0].strip() == "cpu cores"):
             cores_per_socket = int(l[1].strip())
+        if (l[0].strip() == "cpu MHz"):
+            frequency = int(l[1].strip())
     results.update({'cpus': sockets * cores_per_socket})
     # get cpu frequency directly (bypasses freq scaling)
     try:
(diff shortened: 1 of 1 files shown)


Zomtir:
To be honest, I have no clue why this fails... the exact code works for itself without any issue: https://pastebin.com/pVUz86wJ

The _linux_hardware_info() functions crashes as soon as a "physical id" is found, but only in QuTip.

nonhermitian:
The hardware info is wrapped in a `try` statement, thus hiding the error.  To see what the problem is, just remove the try statement, and see what error Travis is returning.

Zomtir:
The NOMKL build is the only one out of the 6 that fails:
`ERROR: Metrics: checks that special cases for dnorm agree with SDP solutions.`
`cvxpy.error.SolverError: Solver 'CVXOPT' failed. Try another solver.`

Unfortunately I'm not seeing how this is related to my changes. I noticed a huge chunk of warnings for scipy.misc.factorial and will make a second pull request. If that ones fails at the same position, it might be unrelated to this commit.

nonhermitian:
It is not related.  I restarted, and it should pass.

nonhermitian:
Now that the issue has been solved, the question is do we want the try statement back?  I am thinking not.

Zomtir:
I'd agree on dropping the try-catch as well. The about.py is dependant on a successful hardware_info() call. So there's little use on catching an error just to invoke another one with an empty array one level higher.

Shall I make a new clean pull request with a more sophisticated commit message?

nonhermitian:
I think the current Pull is good enough.  Many thanks for finding this issue.  Would have never thought of it.

PR #781, opened 2017-11-28
qutip/wigner.py: scipy.misc.factorial is deprecated

Zomtir:
There's a bunch of warnings in the build log that indicate that scipy.special.factorial should be used in favor of scipy.misc.factorial.


nonhermitian:
Many thanks for this.

PR #785, opened 2017-12-12
tensors and br solver in correlation func

IIAOPSW:


qutip/correlation.py
```diff
@@ -539,9 +545,9 @@ def spectrum(H, wlist, c_ops, a_op, b_op, solver="es", use_pinv=False):
         print(inspect.stack()[0][3])
 
     if solver == "es":
-        return _spectrum_es(H, wlist, c_ops, a_op, b_op)
+        return _spectrum_es(H, wlist, collapse_ops, A, B)
     elif solver == "pi":
-        return _spectrum_pi(H, wlist, c_ops, a_op, b_op, use_pinv)
+        return _spectrum_pi(H, wlist, collapse_ops, A, B, use_pinv)
```
nonhermitian: This is one of the lines that is causing the problems in the Travis tests.  The `A` and `B` operators are dense NumPy arrays and not Qobj objects, thus throwing an error.

qutip/correlation.py
```diff
@@ -1297,20 +1369,20 @@ def _correlation_mc_2t(H, state0, tlist, taulist, c_ops, a_op, b_op, c_op,
...
+    L = H if issuper(H) else liouvillian(H, collapse_ops)
 
     tr_mat = tensor([qeye(n) for n in L.dims[0][0]])
     N = np.prod(L.dims[0][0])
 
     A = L.full()
-    b = spre(b_op).full()
-    a = spre(a_op).full()
+    b = spre(B).full()
+    a = spre(A).full()
```
nonhermitian: `A` is a dense array when passed to `spre`.  This is the error.  It should be sparse, and then the result of `spre` is converted to dense.

qutip/tensor.py
```diff
@@ -328,6 +300,8 @@ def tensor_swap(q_oper, *pairs):
     return Qobj(inpt=data, dims=dims, superrep=q_oper.superrep)
 
 
+import qutip.states
+
 def tensor_contract(qobj, *pairs):
```
nonhermitian: Just FYI: This whole routine can be made much faster by moving the loops to Cython.

qutip/correlation.py
```diff
@@ -70,9 +71,9 @@
 
 # low level correlation
 
-def correlation_2op_1t(H, state0, taulist, c_ops, a_op, b_op,
+def correlation_2op_1t(H, state0, taulist, collapse_ops, A, B,
```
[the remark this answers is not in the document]
IIAOPSW: The reason for the rename is that c_ops could just as easily be mnemonic for "coupling operator" as it is for "collapse operator". To add to the confusion the correlation solver also has a "c_op" argument which is short for "C operator". To add even more confusion the coupling operators are often named "a_ops" in the documentation (despite having no lexicographical similarity) and might further be confused with "a_op" which is short for "A operator".

I thought that having "a_op", "a_ops", "c_op" and "c_ops" all in the same function call is a bad idea and decided to just spell everything out. If you want to switch back to the other format for consistency then sure, I can do that.

qutip/tensor.py
IIAOPSW: noted and appended to todo

qutip/correlation.py
```diff
@@ -1048,27 +1057,90 @@ def _correlation_2t(H, state0, tlist, taulist, c_ops, a_op, b_op, c_op,
...
+    coupling_ops_shifted = coupling_ops #allops_shifted[len(collapse_ops):] #time shifting coupling ops is irrelevant? 
+    if config.tdname:
+        _cython_build_cleanup(config.tdname)
+    rhs_clear()
+    for t_idx, rho in enumerate(rho_t):
+        if not isinstance(H, Qobj):
+            _args["_t0"] = tlist[t_idx]
+
+        temp = B.isherm
+        B.isherm = False #temporarily pretend B is non hermitian otherwise a crash will happen later on
```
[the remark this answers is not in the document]
IIAOPSW: The error originates from bloch_redfield line 326.

Obviously the reason for this line is that the br solver expects Hermitian operators to always evolve as real valued (which is true). I'm not sure of the deeper reason for the failure but I think it has to do with the initial states being passed in. 

My original workaround was to us brme_solve to evolve the wavefunction completely and then calculate <B> after the fact using expect(). It gave the exact same result as my current code but took a lot longer.

qutip/correlation.py
IIAOPSW: This is deeply unusual since I didn't touch that part of the code and spre has an if statement to throw an error when input is not a Qobj. 

Any idea how this happened?

qutip/correlation.py
```diff
@@ -70,9 +71,9 @@
 
 # low level correlation
 
-def correlation_2op_1t(H, state0, taulist, c_ops, a_op, b_op,
+def correlation_2op_1t(H, state0, taulist, collapse_ops, A, B,
                        solver="me", reverse=False, args={},
-                       options=Options(ntraj=[20, 100])):
+                       options=Options(ntraj=[20, 100]),coupling_ops=None):
```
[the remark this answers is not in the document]
IIAOPSW: when you're right you're right. fixed.

PR #753, opened 2017-09-22
Add permutational invariance methods to QuTiP

sahmed95:
#This pull request will deal with a new module for solving quantum systems exploiting permutational symmetries. A somewhat complete code for this is available at https://github.com/nathanshammah/piqs which is built on several papers @nathanshammah can link to. The code is almost complete and tested but here we will upload it in parts to make the review process easy.

@ajgpitch @nonhermitian @nwlambert

PR #793, opened 2018-01-11
WIP: General HEOM class for arbitrary spectral density functions

sahmed95:
Hi, this is just a refactoring of @nwlambert 's code for a generic HEOM solver which takes in a set of exponents and coefficients for a spectral density function and can be used to work with arbitrary spectral density functions. In comparison, the current `heom` solver in QuTiP is specialized. 

Specific quantum systems in the `qutip/models` sub-module could have the following signature:
```
model = System(hamiltonian, **params)
result = model.solve(initial_state, tlist, options)
```
System specific functions could be in the model class. For instance,
```
model._rhs()
```
gives the RHS (a matrix equivalent to an effective Liouvillian) of the density matrix evolution equation.

If any system can be written down as the evolution of an initial density matrix as dp/dt = (RHS)p, and there is an optimised way to solve, it can be written down as a qutip.model. The other necessary function would be a basis change map or a lookup dictionary for translating a state tuple to the index in the larger density matrix which is used for the evolution using the effective Liouvillian. This is the `_heom_state_dictionaries` in this case and `jmm1_dictionary` in case of the WIP #753 

The code in the current form is good enough to be used but it needs to be optimized further. The next steps are:
-  [x] Tests for RHS generation and the solve function
-  [ ] Optimize the generation of the effective Liouvillian (RHS) similar to the `heom` solver
-  [ ] Refactoring and deciding on the API (how to pass in the coefficients etc)


sahmed95:
Closing this for now. Will get back with a new PR implementing a time-dependent implementation as well as general spectral densities for the HEOM method.

gharib85:
General HEOM class could be easily implemented if spectral density called inside heom class   as  a parameters , by implementing spectral density itself as a separate class
That will facilitate implementation 
You can see new package pyheom 
https://aip.scitation.org/doi/abs/10.1063/5.0007327

PR #795, opened 2018-01-14
Fixes #794: Changed dimensions for allowing tensor product of states

### LINKED ISSUE #794: HEOM solver bug for a tensor product of systems: requires a change of dimensions
In the construction of the RHS of the equation for the Hierarchy Equations of Motion (`non_markov/heom`) there is a bug found by @nwlambert . If the 'system' is actually a tensor product of systems it fails as the dimensions are incorrect. This can be fixed by changing the construction of the RHS in `non_markov/heom.py` (line 310) from

```
sup_dim = H_sys.dims[0][0]**2
unit_sys = qeye(H_sys.dims[0])
```
to

```
N_temp = reduce(mul, H_sys.dims[0], 1)
sup_dim = N_temp**2
unit_sys = qeye(N_temp)
```

I will open a PR to do the same. Just raising the issue so that it is documented and helps others find the problem in the previous versions.

sahmed95:
This was raised by @nwlambert and the code change fixes the issue of setting appropriate dimensions while constructing the RHS when the system is a tensor product. This is a fix for the issue: https://github.com/qutip/qutip/issues/794

### DIFF
--- a/qutip/nonmarkov/heom.py
+++ b/qutip/nonmarkov/heom.py
@@ -55,7 +55,6 @@ from qutip.cy.heom import cy_pad_csr
 from qutip.cy.spmath import zcsr_kron
 from qutip.fastsparse import fast_csr_matrix, fast_identity
 
-from operator import mul
 
 class HEOMSolver(object):
     """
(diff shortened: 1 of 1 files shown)


ajgpitch:
Thanks @sahmed95 for doing this fix. It seems that there is an issue with the Python3 running. You have to import reduce with using Python3

nwlambert:
Might need to import mul as well. I think its

from operator import mul

sahmed95:
Yes. I will change it. It's recommended to use a loop instead of reduce in
python3.

sahmed95:
@ajgpitch This can be merged now

PR #809, opened 2018-01-29
Better citation message

sahmed95:
Hi,

I noticed that the current message regarding the citations gets lost in the end of the build messages and warnings. A small modification to make the message stand out. @ajgpitch


ajgpitch:
Looks good, thanks
I'll merge when tests have passed

PR #814, opened 2018-02-07
td qobj

Ericgig:
The include a new class to represent time dependent Qobj system.
It's take the list of Qobj and coefficient in the format normally accepted by solvers like mesolve and mcsolve and make an object of it.
The td_Qobj is callable and return the Qobj at the desired time, it support math (+,-,*) and unitary transformations: conj, dag, transpose.
The object has a cython compiled version with cython call to spmv and expect in order to be efficient in solvers.
I made it as a tool used by solver (stochatic/mcsolve), therefore may lack some features for users. Also the naming of some methods only make sense in a solver (rhs).
Lastly, this branch include it's own cubic spline code since the version available when I started working on it only accepted uniformly sampled data. This part may have become redundant.

This branch only contain the new object, not the solvers using it.
This pull request replace half the "cythonizing mcsolve" pull request.


Ericgig:
Not up to date with the version in stochastic and useless by itself.

PR #816, opened 2018-02-07
Mcsolve

Ericgig:
This contain a new solver for mcsolve, dopri5 which don't go through scipy and is faster for small system, and replace the treatment of time dependent cases with the use of td_Qobj.

Replace the "cythonizing mcsolve" pull request: The new pull request split the td_qobj part and mcsolve part so that the pull request for stochastic stays independent from this one (both use td_Qobj).


Ericgig:
No longer compatible with the stochastic pull.
Not to merge now.

PR #817, opened 2018-02-08
qobj isunitary attrib

ajgpitch:
add isunitary attribute for checking if qobj is a unitary operator:
unitary checking func and property funcs added
initialisation of isunitary
clear flag on operations that risk unitarity
qobj unitarity testing added
preset isunitary for qeye

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -101,7 +101,7 @@ shape = [3, 3], type = oper, isHerm = True
         raise TypeError('j must be a non-negative integer or half-integer')
 
     if not args:
-        return jmat(j, 'x'), jmat(j, 'y'), jmat(j, 'z') 
+        return jmat(j, 'x'), jmat(j, 'y'), jmat(j, 'z')
 
     if args[0] == '+':
         A = _jplus(j)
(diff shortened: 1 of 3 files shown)


ajgpitch:
Nothing too clever here. As in cached flag cleared whenever unitarity threatened
The check could probably be more efficient, but as not expected to be used intensively, then this would seem acceptable.

nonhermitian:
Formally you need to check U*U.dag and U.dag*U.  There is a more optimal way of doing this by slighly modifying the zcsr_mult routine, but may not be worth the effort.

ajgpitch:
Thanks @nonhermitian. I will add the other check.
Its not immediately apparent to me what I can do with zcsr_mult to optimise things. Do you mean have a routine like (pseudo code?):
```
zcsr_mult_adj(A):
    return A*A.dag()

```

nonhermitian:
I mean you can do the check without doing the full multiplication by manipulating the sparse structure itself.

nonhermitian:
Again, not straight forward, but would be faster.

ajgpitch:
As I am only planning to use this in solver setup, then I think this will be faster enough.
Hence merging now.

PR #821, opened 2018-02-13
spectrum_correlation_fft edit: 2-sided spectra and Fourier conventions

obmatt1234:
Finding the full power spectrum given a time-dependent Hamiltonian usually requires a 2-sided power spectrum. For convenience, it would be useful to be able to access this data in a properly ordered array.

Additionally, the majority of literature regarding cavity optomechanics uses a positive e^(i w t) Fourier convention. Passing a simple boolean parameter to this function would allow for the use of scipy's ifft function, scaled by the number of elements.

### DIFF
--- a/qutip/correlation.py
+++ b/qutip/correlation.py
@@ -547,7 +547,7 @@ def spectrum(H, wlist, c_ops, a_op, b_op, solver="es", use_pinv=False):
                          "%s (use es or pi)." % solver)
 
 
-def spectrum_correlation_fft(tlist, y):
+def spectrum_correlation_fft(tlist, y, inverse = False):
     """
     Calculate the power spectrum corresponding to a two-time correlation
     function using FFT.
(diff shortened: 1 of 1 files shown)


qutip/correlation.py
```diff
@@ -547,7 +547,7 @@ def spectrum(H, wlist, c_ops, a_op, b_op, solver="es", use_pinv=False):
                          "%s (use es or pi)." % solver)
 
 
-def spectrum_correlation_fft(tlist, y):
+def spectrum_correlation_fft(tlist, y, inverse = False):
```
ajgpitch: pep8 says no spaces in `inverse = False`

ajgpitch:
Aside from one small PEP8 fix I see no reason not to merge this
@kafischer @nonhermitian any comments

nonhermitian:
Seems fine to me.

ajgpitch:
Ok, thanks. I made the pep8 fix. I will merge when tests have passed again

PR #827, opened 2018-02-26
Fixes #825: Added the axis handle to

### LINKED ISSUE #825: Visualization of density matrix via `visualization.hinton` doesn't work with subplots
Consider the following MWE:

```
%matplotlib notebook
import qutip
fig, axs = plt.subplots(1, 2)
dm1 = qutip.ket2dm(qutip.basis(2, 0))
dm2 = qutip.ket2dm(qutip.basis(2, 1))

qutip.visualization.hinton(dm1, ax=axs[0]);
qutip.visualization.hinton(dm2, ax=axs[1]);
```

The result is the following:
capture

I'm guessing this should work? Or should I do this kind of plotting differently?

sahmed95:
This adds the axes to the `_blob` function so that the plots are made on the subplot axes if the user supplies them. @ajgpitch

### DIFF
--- a/qutip/visualization.py
+++ b/qutip/visualization.py
@@ -70,7 +70,7 @@ from qutip import settings
 
 
 # Adopted from the SciPy Cookbook.
-def _blob(x, y, w, w_max, area, cmap=None):
+def _blob(x, y, w, w_max, area, cmap=None, ax=None):
     """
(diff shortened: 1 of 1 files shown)


ajgpitch:
@sahmed95 Thanks for this. 
Just restarted failed test. Fail not related to this change, so hopefully just a blip

PR #832, opened 2018-02-27
sesolve unitary operator evolution

ajgpitch:
Can now give unitary operator as initial `psi0` in sesolve. This way we solve the Schrodinger operator equation.
Also added tests for all the sesolve methods.

Kind of a bit of waste of time, as `propagator` does almost exactly the same thing. However, I think it does add a bit of completeness. The tests at least should be valuable


ajgpitch:
I think this is pretty much ready to go, but I will leave it here for a few days in case anyone wants to check and comment

PR #834, opened 2018-03-02
fixes #833: updated docstring

### LINKED ISSUE #833: Incorrect docstring of spin coherent state
In qutip.states the docstring of the `spin_coherent` state is the same of `spin_state`. The correct description should be: "Generate the coherent spin state |theta, phi>."

nathanshammah:


ajgpitch:
Thanks Nathan

PR #835, opened 2018-03-08
independent citation generator

sahmed95:
@ajgpitch Hi, there were some issues with bundling the `cite` function inside about as it was itself a module. So `qutip.about.cite()` would not work and one had to do `from qutip.about import cite` and then run it. Instead, I think it is better to make a `cite` function such that it can be called as `qutip.cite()`. I made the necessary changes. Please let me know if this makes sense. `qutip.about.cite()` was failing.


ajgpitch:
Agreed. Makes sense

Issue #672, opened 2017-03-17
Q&A: Why so many old commits in PR branch?


ajgpitch:
In all my recent PR branches many commits are shown. I think these date back to when we started Squash & Merge as our policy. The same policy is in use on our conda-forge/qutip-feedstock but I don't find this issue there. I did a lot of reading, but could not find a nice simple way to stop this happening. Do we know why this is happening? I see @nonhermitian has a similar string of old commits on his PR branches too.

I see we have a new 'Rebase and Merge' option for merging PRs now. Do we think we should be using this?

ajgpitch:
I think the answer to why this happens is that there are some commits in the origin/master that conflict with the upstream/master, even though the final state of the origin/master has no conflicts with the upstream/master. 
A solution therefore is to (on the origin/master branch)
```
git rebase upstream/master
```
skipping any conflicting commits, then
```
 origin/master
```
Will confirm when I have tested a bit more

ajgpitch:
I managed to remove all the old commits by running
``` 
git rebase upstream/master
```
skipping any conflicting commits, then
```
git push --force
```
on my feature branch. As evidenced in #745 
But I still don't seem to have sorted out my master branch

ajgpitch:
The above did actually sort out my master branch - I just needed to run the commands on my laptop as well :-)
I did not need to rebase #746 
So I am going to call this answered

PR #838, opened 2018-03-16
Enr fock_state fix

IIAOPSW:


ajgpitch:
@IIAOPSW I think something has gone wrong here with your git branch. Seems a bunch of other changes got included in. I assume, due to the name, that you are wishing to push the change in `states.py` only.

If so, I would suggest that you start again with a new branch. You may need to rebase your master branch first. See #672

PR #851, opened 2018-04-13
Fixes #847 scipy requirement superceeds error catching

### LINKED ISSUE #847: qdiag throws incorrect error
Hi all

When I ran this code

```python
import qutip as q
q.qdiags(q.basis(3, 0), 0)
````

The error that was thrown indicated an improper version of Scipy instead of an error associated with
the function itself. I think it's because the [exception clause here](https://github.com/qutip/qutip/blob/master/qutip/operators.py#L753) is too broad. I think it should be catching ``NameError``.

sahmed95:
This is a fix for #847. Since we already have a minimum requirement for Scipy, the error catching here does not make sense. It is better to leave the error returned from the scipy function `scipy.diags` untouched as the qutip version just wraps the result into a `qutip.Qobj`. 

@ajgpitch

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -717,6 +717,7 @@ def qdiags(diagonals, offsets, dims=None, shape=None):
     ----------
     diagonals : sequence of array_like
         Array of elements to place along the selected diagonals.
+
     offsets : sequence of ints
         Sequence for diagonals to be set:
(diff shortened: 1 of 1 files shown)


ajgpitch:
@sahmed95 Thanks for fixing this. Merging now.

PR #855, opened 2018-04-25
BUG: Dnorm complex casting bug

nonhermitian:
This fixes a complex casting warning in the diamond norm function.


ajgpitch:
Thanks Paul,
I was hoping we could get rid of that warning

PR #865, opened 2018-05-22
Fix depreciation warnings in iterative solvers

nonhermitian:
Setting `atol='legacy'`.


sahmed95:
@nonhermitian Thanks for the fix. But the tests are failing for the older scipy version. Maybe we should bump up the min scipy requirement in the setup for the next version?

ajgpitch:
@sahmed95  The principle is fine. What is the minimum version we can support?

nonhermitian:
The issue is that conda py27 does not have the latest scipy.  Perhaps it is
coming, or perhaps they are depreciating it.  I would suggest dropping py27
support, and then bump scipy to 1.1.0 or higher.

It is actually annoying because the warning is called at each iteration,
and the fix breaks back compatibility, which of course they should not do.

sahmed95:
Ok there are two issues here.

1. Scipy `atol` keyword argument has only been introduced in Scipy 1.1, the latest version. So we should bump it to 1.1 or should do nothing at all and ignore the warnings.

2. The tests are failing for the sparse eigenval function in [`qutip.tests.test_sp_eigs`](https://github.com/qutip/qutip/blob/master/qutip/tests/test_sp_eigs.py#L45). So there is something else which is breaking and this is some weird segmentation fault issue, see below. Note that the test itself was known to give trouble before. But for some reason, with Python 3+ and the latest scipy, everything works fine. 

I am very much in favor of dropping support for Python 2.7 as even Numpy seems to be gearing up to drop Python 2.7. [Link](https://docs.scipy.org/doc/numpy/neps/dropping-python2.7-proposal.html). 

With regards to the Scipy version, 1.1 seems okay as @nonhermitian mentioned, but it is the latest version though. 

```
@unittest.skipIf(_version2int(scipy.__version__) < _version2int('0.10'),
                 'Known to fail on SciPy ' + scipy.__version__)
def test_SparseHermValsVecs():
    """
    Sparse eigs Hermitian
    """    
```

Failure
```
Sparse eigs Hermitian ... /home/travis/.travis/job_stages: line 57:  6793 Segmentation fault      (core dumped) nosetests --verbosity=2 --with-coverage --cover-package=qutip qutip
```

sahmed95:
So, have we come to a conclusion here? We should decide and make the necessary changes on this as soon as possible. The way forward would be a Python 3+ shift and updating the Scipy requirements.

ajgpitch:
I agree we need to prioritise this

Not sure we should remove Python 2.7 support without plenty of warning. I think we should definitely remove it in v5. I am assuming we will be making a v4.3 release before that.

(I have not examined this in any detail, so excuse me if I have missed some crucial point)
One option could be to only support the latest scipy in py3?

Has anyone checked the scipy github to see if others are suffering / offered solutions?

ajgpitch:
Unfortunately 'legacy' did not cause legacy behaviour. 
Hopefully #870 will do the trick

ajgpitch:
Superseded by #870

PR #868, opened 2018-05-31
Sesolve fix unitary op evo rhs matrix

ajgpitch:
The unitary operator evolution implemented in #832 was flawed.
This is a complete overhaul. It uses the method from mesolve map evolution.
The tests implemented in #832 did not uncover the flaws, and so extra ones are added that would have.


sahmed95:
Hi, what was the flaw in the previous PR? This looks good to me overall. I have not worked much with the `sesolve` but all the checks on the initial state input look fine. Feel free to merge this.

Issue #872, opened 2018-06-03
Metrics tests failing "multiple values for argument 'name'"


ajgpitch:
Current 106 tests are failing. All seeming 'Metrics' test.
All failing with:
```
TypeError: __init__() got multiple values for argument 'name'
```
I will try and reproduce locally

ajgpitch:
I have managed to reproduce.
It would that conda is currently installing an openblas version of numpy even when `mkl` version is set. I guess this is a mistake by Anaconda that they will fix soon. Probably because they rushed the fix for #871 
This means that the 'metrics' tests with "known issues" are not skipped.
What does not make sense is why the 'NOMKL' tests are failing.

Some questions:
- why do we install a specific version cvxopt in our test build?
- These "known issues" must be quite old, so perhaps may have been fixed.

ajgpitch:
The "known issues" are definitely still there. They cause whatever `dnorm` does using cvxpy to fail, just through having `mkl` as a conda package. `test_metrics` uses the `qutip.settings.has_mkl` to determine whether to skip the dnorm tests. This however is False whenever blas_info is not mkl, mkl could still be installed.

ajgpitch:
The main issue causing these errors though is that `cvxpy.Variable` and `cvxpy.Parameter` expect the first arg to be a tuple with dimensions. We are passing `num_rows, num_cols` is two args. Who knows how this has happened - cvxpy changed its interface?

ajgpitch:
As far as the "known issues" go, these are discussed in #484.
I don't think the issues I saw recently  are the same as in that one.
```
Intel MKL FATAL ERROR: Cannot load libmkl_avx.so or libmkl_def.so.
```
I think this is just an MKL install problem.
It's difficult to test at the moment, as I don't seem to be able to create a conda env with mkl blas

PR #873, opened 2018-06-03
fix - semidefinite cvxpy Variable and Parameter

ajgpitch:
cvxpy Variable and Parameter dims passed as tuple
cvxpy and its depends only installed in tests for NOMKL builds
Extra test added for NOMKL with OPENMP

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ matrix:
           env: NOMKL=1 OPENMP=0
         - python: 3.6
           env: OPENMP=1
+        - python: 3.6
+          env: NOMKL=1 OPENMP=1
 addons:
     apt:
         packages:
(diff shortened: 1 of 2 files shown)


ajgpitch:
Attempted fix for #872

sahmed95:
Thanks for this. Merging now.

ajgpitch:
@sahmed95 I am assuming that tests passed then? 
I can't find any Travis record of them having run.

sahmed95:
Yes, the tests ran locally on a test environment I had created. The metrics test ran fine on Travis indirectly on this other PR : https://travis-ci.org/qutip/qutip/builds/379109476. 

Note that the python 2.7 failures are not related but are of similar nature with something changing for supplying arguments in tuples vs directly passing them. However, in hindsight, I should have rebuild on this PR. But I could find no way of re-starting the tests here as travis didn;t see any change to trigger a new test. I will test this with a fresh PR.

sahmed95:
Testing the master on: https://travis-ci.org/qutip/qutip

ajgpitch:
Ok, seems like everything is super fragile at the mo.

Can't see the extra test I added running.
NOMKL=1 OPENMP=1
Does anything else need to be done, other than editing the .travis.yml?

sahmed95:
You are right. The travis.yml could not be parsed in the last PRs you made: https://travis-ci.org/qutip/qutip/requests

sahmed95:
There are conda versions for cvxopt and cvxpy (https://anaconda.org/mcg/cvxpy) but nobody seems to be maintaining or updating them. Installing the conda version downgrades everything else. I will first try to fix the yml issue. Nothing seemed to be incorrect in the YML file, and the travis CI linter is not available, so I am not sure what is wrong with the file. Will update if I make progress.

ajgpitch:
Yes, I also tried using these packages. If you see #484 and all the
corresponding links you will see why we switched to the pip install.

It's questionable whether we should have a feature that relies on a library
that is not maintaining a conda package. But that is a question for another
day.

sahmed95:
There is an issue in the travis yml file. Attempting to fix this.

PR #883, opened 2018-06-18
tensor contract on sparse matrix

IIAOPSW:
This is *exactly* my old pull request #785 but I went through and rebased my local branches and disentangled it from the BR fix. I also went ahead and cythonized the main loop as per request.


sahmed95:
Please rebase this against the current master @IIAOPSW . There are some changes in the travis.yml file put in by @ajgpitch which selectively avoids some py2.7 tests. There were many new additions to tackle some Python 2 issues too.

IIAOPSW:
@sahmed95 shouldn't the merge automatically take care of that without bothering to rebase? 

I'd like to avoid wasting time on a rebase if there's no tangible benefit

sahmed95:
Yeah. Just do a 'git pull upstream master' on this branch. The Travis
config file was updated and it's not the same as this branch. You don't
have to do a painful rebase. The tests need to run online.

IIAOPSW:
@sahmed95 weird that the travis test still fails. I merged with master and it still does this.

sahmed95:
It is strange. The failure is not dependent on the changes that you made here since the failure is due to some missing header files in Numpy. I am at a loss on what is happening. But your other PRs are going through so I am sure, it is something small. I will try restarting the builds for this PR.

ajgpitch:
@IIAOPSW when you refer to your 'old pull request', please do so using its # This way there there is a link back.
Alternatively you could just give a little detail on what we doing in this PR

ajgpitch:
Not really sure on the source of the error here. Stange one, as it's from a file not changed in this PR.
All I can suggest is to rebase from the upstream master again, which is not really a very onerous process
```
git rebase upstream/master
```
skipping any conflicting commits, then
```
git push --force
```

ajgpitch:
There are quite a few pep8 issues. The most obvious ones are comment related.
@IIAOPSW please see https://www.python.org/dev/peps/pep-0008/#comments
Perhaps we should focus on these first.
Maybe @sahmed95 could explain to @IIAOPSW how to use the pep8 linter?

ajgpitch:
I tried to replicate locally but seemed difficult to create the same mkl/numpy/scipy env.
Anaconda Python 2.7 now seems to have mkl 2018, scipy 1.1, numpy 1.14
Travis is creating mkl 2017, scipy 0.19, numpy 1.13
I suspect it will update soon, so lets fix the other issues first and see what happens when it's merge-ready

ajgpitch:
@IIAOPSW I cannot actually find your original PR, so lets make like it never happened. Edit the first comment to give some background on this.

IIAOPSW:
@ajgpitch I believe I fixed the pep8 issues. 

There was also a cython issue where I was compiling tensor.pyx each time qutip starts (aka dev mode) rather than leaving it pre-compiled. I think this might be related to the travis test failures. We are about to find out.

ajgpitch:
I will attempt to work out what this is about and get it merged

hodgestar:
I think it would be great to have a fast implementation of tensor contraction. However, QuTiP has moved on since this PR and I think the right way to implement this now would be to add a new specialization to the data layer for tensor contraction (maybe something like einsum) and to implement that instead (or even just to re-implement methods for the new CSR representation).

@IIAOPSW If you're up for giving that a try, that would be awesome. If not, I might close this PR and we can revisit it after QuTiP v5 alpha has been released.

@IIAOPSW Separately, I've read through the code the PR, but I'd still like to get a high-level summary of the algorithm you're implementing (or a reference). Quite a lot of work has been done in optimizing these contractions in recent years, and it would be nice to implement a good algorithm in QuTiP.

hodgestar:
Closing this for now, but happy to reopen it later if needed.

PR #884, opened 2018-06-19
Br in correlation function

IIAOPSW:
this was the other half of the fork that I just finished rebasing into its own branch.

### DIFF
--- a/qutip/bloch_redfield.py
+++ b/qutip/bloch_redfield.py
@@ -323,7 +323,7 @@ def bloch_redfield_solve(R, ekets, rho0, tlist, e_ops=[], options=None, progress
 
     for e_eb in e_eb_ops:
         if e_eb.isherm:
-            result_list.append(np.zeros(n_tsteps, dtype=float)) #this line gives grief 
+            result_list.append(np.zeros(n_tsteps, dtype=float))
         else:
             result_list.append(np.zeros(n_tsteps, dtype=complex))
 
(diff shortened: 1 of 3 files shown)


ajgpitch:
@IIAOPSW Could you just remind us here what we are doing with this PR please

ajgpitch:
Some initial feedback:
There are a number of lines that overrun the pep8 limit

ajgpitch:
@IIAOPSW I cannot actually find your original PR, so lets make like it never happened. Edit the first comment to give some background on this.

IIAOPSW:
The original was #785 which auto-closed when I deleted the branch and rebased to split it into 2. I'm just undoing my branch management sloppiness. 

This was about the correlation function supporting bloch-redfield solver.

ajgpitch:
Will check and fix any pep8 issues, then merge

hodgestar:
@Ericgig Do you think it's a good idea to add the br_solver as an option for the correlation functions? If so, there a plan to make selecting a solver more flexible in QuTiP v5? I found https://github.com/qutip/qutip/projects/6#card-72925447 which suggest there might be, but I don't know the full plan.

Ericgig:
We don't need to add `br_solver` as a correlation option, I think we can already do it by passing a bloch redfield tensor as `H` and  using `solver="me"`.  
The plan for v5 is a simple cleaning as it still calls `cython_build_cleanup` and `rhs_clear` which don't do anything anymore.

hodgestar:
In that case I think I will close this for now -- please shout if it needs to be reopened.

Issue #823, opened 2018-02-21
Tracing out systems with dimensions 1 on multipilcation creates dimention missmatch errors


tacruc:
I'm woundering what is the reason for doing this partial trace
https://github.com/qutip/qutip/blob/0b068fc4876727303644c42c181af82e1f089700/qutip/qobj.py#L490
becauseit leads to errors doing:
```
O=tensor(sigmax(), qeye(1))
Psi = tensor(basis(2,0),basis(1,0))
O*O*Psi
```

would it be possible to implement an Option in settings to dissable this behavior? I would be willing to provide a PR.

IIAOPSW:
AFAIK this error only manifests if you have a space of dimension 1.
and a space of dimension 1 is trivial. 

Is there a use case where this actually manifests?

tacruc:
I use the qutip to do some Grover simulations. having a Hilbertspace of Grover an an additional entangled Hilbert space. In a simpler version there is only one state in the entangled Space so the program automatically chooses dimension 1. 
And then there are allot of errors coming because of this qutip error. 
At the moment I fixed it using a patched qutip version.

ajgpitch:
Arne,

Please propose your patch in a PR on our Github. So long as it is
mathematically sound, then the context does not matter in this case.

If you need any advice on how to do this, then let us know. Email me direct
if you prefer. alex.pitchford@gmail.com

Thanks

Alex

tacruc:
This issue can be closed since the PR #886 is merged or?

PR #886, opened 2018-06-21
Added options auto_tidyup_dims to disable deletion of 1 dim Systems after multiplication

tacruc:
In the multiplication is a mechanism deleting 1 dimensional Systems. This can lead to problems on Systems having multiple one dimensional Subsystems. Examples are in #823.

Because I coundn't figure out in which cases this deleting mechanism needed, I didn't want to remove it, but provide an option to disable it.

Signed-off-by: Arne Hamann <kontakt+github@arne.email>

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -488,8 +488,8 @@ class Qobj(object):
                 out.dims = dims
                 if settings.auto_tidyup: out.tidyup()
                 if (settings.auto_tidyup_dims 
-                        and not isinstance(dims[0][0], list) and
-                        not isinstance(dims[1][0], list)):
+                        and not isinstance(dims[0][0], list)
+                        and not isinstance(dims[1][0], list)):
                     # If neither left or right is a superoperator,
                     # we should implicitly partial trace over
                     # matching dimensions of 1.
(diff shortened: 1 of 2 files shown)


ajgpitch:
This looks fine to me
@nonhermitian do you see any reason not to merge this?

qutip/qobj.py
```diff
@@ -487,7 +487,7 @@ def __mul__(self, other):
                 dims = [self.dims[0], other.dims[1]]
                 out.dims = dims
                 if settings.auto_tidyup: out.tidyup()
-                if (not isinstance(dims[0][0], list) and
+                if (settings.auto_tidyup_dims and not isinstance(dims[0][0], list) and
```
ajgpitch: this line is too long - pep8 wise

ajgpitch:
Actually it needs one pep8 fix please

ajgpitch:
@tacruc for future reference you should really create a new branch on your fork and submit that as a PR for qutip. It's okay in this case, as it is a simple change, but would cause issue if you were possibly making multiple PRs at the same time.

ajgpitch:
If there are no comments on this the I will merge when tests pass

tacruc:
@ajgpitch sorry for the long time without reply. Thanks for the commit with pep8 and the tip with the branch.

ajgpitch:
No worries.
Merged now.
Thanks for the contribution

PR #895, opened 2018-07-10
FreeBSD compatibility: add _freebsd_hardware_info.

yurivict:


ajgpitch:
@yurivict Thanks for your contribution. 
Merged just now.
For future contributions, you should really create a branch on your fork to make the PR.
See this for more info:
https://github.com/qutip/qutip-doc/blob/master/qutip_dev_contrib.md

yurivict:
> For future contributions, you should really create a branch on your fork to make the PR.

Ok, but what difference does it really make? The difference would be if I needed to make several contributions, but I don't need to do this in this case.

PR #899, opened 2018-07-11
Bump to 4.3.1

sahmed95:
Bumping micro after pypi release.


ajgpitch:
@sahmed95 Looks like you are all done. You want me to make an announcement on the Google Group?

ajgpitch:
@sahmed95 looks like we should hold off on the announcement. See Slack

Issue #907, opened 2018-08-02
random_objects.py won't open in IDLE (python's default editor)


IIAOPSW:
Every other file works fine. 

For example, double click qobj.py and this opens.

![example](https://user-images.githubusercontent.com/32380833/43563341-b0c54b2a-965c-11e8-917c-4f2c18d8b457.png)

Double click random_objects.py and it freezes and does nothing. I can open it in other editors like vim but I think it *should* work with the python default editor.  

When I try to open up with IDLE from command line I get this error.

![error](https://user-images.githubusercontent.com/32380833/43563415-f443f842-965c-11e8-94f2-c28503bc6637.png)

Which leads me to think there's a character somewhere in the file that causes the crash. But I don't know where.

ajgpitch:
Well, what char is U+1d7d9?

IIAOPSW:
Appears to be fancy 1

http://www.fileformat.info/info/unicode/char/1d7d9/index.htm

Not sure how it got in there. Perhaps as a comment?

ajgpitch:
I think the recommended way to include mathematical symbols is to use the :math: construct.
This is an example from piqs
```
:math:`|j, m, \\alpha\\rangle`
```

As the comments get built into API docs, then it is good to be able to be include math symbols

jakelishman:
In some comments (not the docstrings) in `random_objects.py` and `metrics.py` (that I can think of - possibly there are others), there are high-range unicode symbols used to make the mathematics clearer.  This couldn't be done with the `:math:` sphinx directive because it's about code comment readability.

I'll close this for now because it's very old, and really the "proper" solution is just to use a text editor that knows how to handle all text.

PR #909, opened 2018-08-03
removed troublesome unicode char

IIAOPSW:
This fixes the issue I bought up yesterday (#907)


ajgpitch:
I think the recommended way to include mathematical symbols is to use the :math: construct.
This is an example from piqs
```
:math:`|j, m, \\alpha\\rangle`
```
I assume we must be able to find a way to display the identity matrix and the tensor product

PR #912, opened 2018-08-24
Added calculating the wigner distribution and plotting it on the Bloch

NithinRamu:
... sphere functionality. See wigner_visualization notebook in examples for details.

### DIFF
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -42,6 +42,8 @@ from qutip.random_objects import rand_dm, rand_ket
 
 
 def test_wigner_bell1_su2parity():
+    """wigner: testing the SU2 parity of the first Bell state.
+    """
     psi = bell_state('00')
 
     steps = 100
@@ -52,19 +54,22 @@ def test_wigner_bell1_su2parity():
     wigner_analyt = np.zeros((steps, steps))
     for t in range(steps):
         for p in range(steps):
-            wigner_analyt[t, p] = ((1 + np.sqrt(3) * np.cos(theta[0, t]))
-                                   * (1 + np.sqrt(3) * np.cos(theta[1, t]))
-                                   + 3 * (np.sin(theta[0, t])
-                                   * np.exp(-1j * phi[0, p])
-                                   * np.sin(theta[1, t])
-                                   * np.exp(-1j * phi[1, p])
-                                   + np.sin(theta[0, t])
-                                   * np.exp(1j * phi[0, p])
-                                   * np.sin(theta[1, t])
-                                   * np.exp(1j * phi[1, p]))
-                                   + (1 - np.sqrt(3) * np.cos(theta[0, t]))
-                                   * (1 - np.sqrt(3)
-                                   * np.cos(theta[1, t]))) / 8.
+            wigner_analyt[t, p] = np.real(((1 + np.sqrt(3)
+                                            * np.cos(theta[0, t]))
+                                           * (1 + np.sqrt(3)
+                                           * np.cos(theta[1, t]))
+                                           + 3 * (np.sin(theta[0, t])
+                                           * np.exp(-1j * phi[0, p])
+                                           * np.sin(theta[1, t])
+                                           * np.exp(-1j * phi[1, p])
+                                           + np.sin(theta[0, t])
+                                           * np.exp(1j * phi[0, p])
+                                           * np.sin(theta[1, t])
+                                           * np.exp(1j * phi[1, p]))
+                                           + (1 - np.sqrt(3)
+                                           * np.cos(theta[0, t]))
+                                           * (1 - np.sqrt(3)
+                                           * np.cos(theta[1, t]))) / 8.)
 
     wigner_theo = wigner_transformation(psi, 0.5, False, steps, slicearray)
 
@@ -72,6 +77,8 @@ def test_wigner_bell1_su2parity():
 
 
 def test_wigner_bell4_su2parity():
+    """wigner: testing the SU2 parity of the fourth Bell state.
+    """
     psi = bell_state('11')
 
     steps = 100
@@ -88,6 +95,9 @@ def test_wigner_bell4_su2parity():
 
 
 def test_wigner_bell4_fullparity():
+    """wigner: testing the parity of the fourth Bell state using the parity of
+    the full space.
+    """
     psi = bell_state('11')
 
     steps = 100
@@ -106,6 +116,8 @@ def test_wigner_bell4_fullparity():
 
 
 def test_parity():
+    """wigner: testing the parity function.
+    """
     j = 0.5
     assert_(_parity(2, j)[0, 0] - (1 - np.sqrt(3)) / 2. < 1e-11)
     assert_(_parity(2, j)[0, 1] < 1e-11)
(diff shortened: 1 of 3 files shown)


Ericgig:
Seems interesting but tests are failing...

NithinRamu:
> 
> 
> Seems interesting but tests are failing...

Hi Ericgig!

Thank you very much for looking into this! I am not able to figure out why the build is failing, the error messages quote parts of the code that I didn't even modify. Could you assist me in finding out why?

Thanks a bunch!
Nithin

qutip/tests/test_wigner.py
```diff
@@ -36,11 +36,158 @@
...
+                                   * np.exp(-1j * phi[0, p])
+                                   * np.sin(theta[1, t])
+                                   * np.exp(-1j * phi[1, p])
+                                   + np.sin(theta[0, t])
+                                   * np.exp(1j * phi[0, p])
+                                   * np.sin(theta[1, t])
+                                   * np.exp(1j * phi[1, p]))
+                                   + (1 - np.sqrt(3) * np.cos(theta[0, t]))
+                                   * (1 - np.sqrt(3)
+                                   * np.cos(theta[1, t]))) / 8.
```
Ericgig: The formula return a complex but `wigner_analyt` is real resulting in a warning.
Please add a `np.real` to remove the warning.

qutip/tests/test_wigner.py
```diff
@@ -36,11 +36,158 @@
...
+                                   * np.exp(-1j * phi[2, p])
+                                   + np.sin(theta[0, t])
+                                   * np.exp(1j * phi[0, p])
+                                   * np.sin(theta[1, t])
+                                   * np.exp(1j * phi[1, p])
+                                   * np.sin(theta[2, t])
+                                   * np.exp(1j * phi[2, p])) + (1 - np.sqrt(3)
+                                   * np.cos(theta[0, t])) * (1 - np.sqrt(3)
+                                   * np.cos(theta[1, t])) * (1 - np.sqrt(3)
+                                   * np.cos(theta[2, t]))) / 16.
```
Ericgig: Here also there is a complex to real warning.

qutip/tests/test_wigner.py
```diff
@@ -36,11 +36,158 @@
...
 
-from qutip.states import coherent, fock
-from qutip.wigner import wigner
+from qutip.states import coherent, fock, ket, bell_state
+from qutip.wigner import wigner, wigner_transformation, _parity
 from qutip.random_objects import rand_dm, rand_ket
 
 
+def test_wigner_bell1_su2parity():
+    psi = bell_state('00')
```
Ericgig: Under the function definition there should be a 1 line docstring to describe the test.
Can be simple like `"testing bell su2 parity"`

qutip/wigner.py
```diff
@@ -50,6 +50,179 @@
...
+    """
+    return (_j_plus(j, h) - _j_minus(j, h)) / 2j
+
+
+def _j_z(j, h=1):
+    """Private function to calculate the z angular momentum operator matrix.
+    """
+    mult = np.int32(2 * j + 1)
+    matrix = np.zeros((mult, mult))
+    np.fill_diagonal(matrix, np.int32(h * np.linspace(-j, j, mult)))
```
Ericgig: Here the `np.int32` round the 0.5 to 0. So `_j_z(0.5,1)` return an empty matrix. Removing it make the test passes.

Ericgig:
Hi Nithin,
I think I found the bug: `_j_z`.
Could you correct it and add make the complex to real conversion explicit to remove the warning.

I would suggest to merge in the present qutip/master since we changed the tests script. The test you have now don't work with the recent version of scipy and cython. If all but mac test passed we can merge.

NithinRamu:
What do I have to do to merge with the current master? Just get all the new files from the main fork and commit and push them here?

nonhermitian:
What is different between this and the built in routine that seems to do the sane thing?

NithinRamu:
@nonhermitian The introduction of [https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.117.180401 ] answers your question better than I could. :)

nonhermitian:
Ok I see what you are doing.  It is not clear from the title that you are doing multi-partite Wigner functions, and there is already a `spin_wigner` function in QuTiP.

Also, the use of `@` notation pegs the Python version at 3.5+, but I believe 2.7 is still supported.  Personally I like support for 3.5+ only, but hate the `@` notation.

NithinRamu:
@Ericgig All tests except Mac are passing. Can we merge now?
@nonhermitian Would you want to my Wigner function to be renamed?

qutip/visualization.py
```diff
@@ -69,6 +69,82 @@
 from qutip import settings
 
 
+def plot_wigner_sphere(fig, ax, wigner, reflections):
+    """Plots a coloured Bloch sphere.
```
Ericgig: Adding it to all will make it easier to use.

qutip/tests/test_wigner.py
```diff
@@ -36,11 +36,181 @@
...
+from qutip.wigner import wigner, wigner_transformation, _parity
 from qutip.random_objects import rand_dm, rand_ket
 
 
+def test_wigner_bell1_su2parity():
+    """wigner: testing the SU2 parity of the first Bell state.
+    """
+    psi = bell_state('00')
+
+    steps = 100
```
Ericgig: Could you set all those steps to 20 are 25, using 100 make clean images but it takes a lot of time and not needed for tests.

qutip/wigner.py
```diff
@@ -48,12 +48,149 @@
...
+def wigner_transformation(psi, j, fullparity, steps, slicearray):
+    """Calculates the wigner function of a quantum state at each theta, phi.
+
+    Parameters
+    ----------
+        psi : qobj
+              a state vector or density matrix.
+        j : int
+            the total angular momentum of the quantum state.
+        sun : int
```
Ericgig: What is `sun`?
Also adding `wigner_transformation` to all could be nice.

Ericgig:
It seems good and could be accepted, but I would prefer it if you could make the tests faster.
Maybe have a better description so that users knows when to use `wigner_transformation` and when to use `wigner`. Both having only a 1 line description, it can be confusing.

(Sorry for the delay.)

qutip/tests/test_wigner.py
NithinRamu: Done

> It seems good and could be accepted, but I would prefer it if you could make the tests faster.
> Maybe have a better description so that users knows when to use `wigner_transformation` and when to use `wigner`. Both having only a 1 line description, it can be confusing.
> 
> (Sorry for the delay.)

I made the changes you requested. Are we ready for merge?

qutip/wigner.py
Ericgig: Here `wigner_transform` is defined with 5 parameters but 6 are listed in the description.
`sun` is hard coded to be 2 and not an input, so listing it as a parameter will only cause confusion.

NithinRamu: Removed this from the docstring in the latest commit.

Ericgig:
Looks good, waiting for the tests to finish then will merge if there are no surprises.

PR #917, opened 2018-10-02
implemented possibility of chosing random seed

marekyggdrasil:


Ericgig:
Thank you Marek,
It seems you forgot to add a seed variable in the definition of the rand_herm function (line 112~114). Because of this, many tests fails. 
Could you add it?

marekyggdrasil:
@Ericgig sorry about that, I just added the argument, I'm also planning adding some tests but I haven't yet figured out what would be best way to do it, any suggestions welcome!

marekyggdrasil:
How about instead of checking if it is Hermitian, we could put a seed parameter and check if the output is exactly equal (or [close](https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html)) to given hardcoded matrix?

https://github.com/qutip/qutip/blob/d91a4c93f35d97b8bfc633420ffa6ddebadc06a1/qutip/tests/test_rand.py#L57

Ericgig:
Hardcoded matrix will cause problem in later modification of the code. (Or if numpy update it's random number generator, etc.)
Keeping the present tests, I would add something like this:
```
O1 = rand_herm(..., seed=seed)
O2 = rand_herm(..., seed=None)
O3 = rand_herm(..., seed=seed)
assert(O1!=O2)
assert(O1==O3)
```
You check that random Qobj are reproducible using a seed and work as expected when not.

marekyggdrasil:
@Ericgig good idea about the tests, I'll implement them in this way

now there seems to be some problem with continuous integration, it fails on

```
ERROR: Failure: ImportError (libgfortran.so.1: cannot open shared object file: No such file or directory)
```

I attempted to fix it by installing `libgfortran` using `conda`, however it didn't fix the issue

https://github.com/travis-ci/travis-ci/issues/4816#issuecomment-148789525
https://github.com/menpo/landmarkerio-server/issues/23#issuecomment-278361601

I don't want to mess up too much with continous integration as it takes long time to rebuild it, any suggestions on that?

Ericgig:
Do worry about the failing python2.7 tests, it's a common error on all recent checks. Your pull should not affect them.

marekyggdrasil:
@Ericgig tests implemented and passing, please code review if you can.

Ericgig:
As it is in your code, if no seed is given, then np.random.seed(None) is called. np.random.seed(None) reset the seed based on the computer time. 
It would break the reproducibility of code which already set the seed once:
```
np.random.seed(seed)
my_rand_qobj = qt.rand_herm(N)
```

It would be better if you did:
```
if seed is not None:
    np.random.seed(seed=seed)
```

marekyggdrasil:
@Ericgig thanks for suggestion

weird, my local tests all passed, let me examine the problem

Ericgig:
It seems to be caused by a new version of cython which affect which files are used for tests by nose. I made a patch.

marekyggdrasil:
Thats great, thanks @Ericgig 

If any other modifications requested don't hesitate to let me know. Looking forward to have my changes merged!

qutip/random_objects.py
```diff
@@ -150,8 +154,10 @@ def rand_herm(N, density=0.75, dims=None, pos_def=False):
             _check_dims(dims, N, N)
         nvals = N**2*density
         while M.nnz < 0.95*nvals:
-            M = rand_jacobi_rotation(M)
+            M = rand_jacobi_rotation(M, seed=seed)
```
Ericgig: Putting the seed here will reseed at each iteration of the while loop.
Better to to it just once at the beginning of the function.

qutip/random_objects.py
```diff
@@ -352,7 +360,7 @@ def rand_dm(N, density=0.75, pure=False, dims=None):
             _check_dims(dims, N, N)
         nvals = N**2*density
         while H.nnz < 0.95*nvals:
-            H = rand_jacobi_rotation(H)
+            H = rand_jacobi_rotation(H, seed=seed)
```
Ericgig: Here the code is also calling seed inside a loop.

qutip/random_objects.py
```diff
@@ -494,15 +502,17 @@ def rand_super(N=5, dims=None):
         pass
     else:
         dims = [[[N],[N]], [[N],[N]]]
-    H = rand_herm(N)
+    H = rand_herm(N, seed=seed)
+    if seed is not None:
```
Ericgig: The seed is already applied in rand_herm, no need to do it twice.

Ericgig:
I marked a few spot where the seed was applied more than once.
The failed test is a known issue.

marekyggdrasil:
@Ericgig I marked your remarks as resolved and made some new changes. If anything more required don't hesitate to ping me.

PR #923, opened 2018-10-10
Update travis build

nonhermitian:
- Remove Python 2.7 (probably good to drop support at this stage)
- Bumped Python version numbers up.
- Added OSX test.


nonhermitian:
OSX build is causing troubles, so will reopen with linux only mods

PR #927, opened 2018-10-19
change to fit with cython0.29

Ericgig:
With cython0.29 that came out this weeks there where warning at "cythonization" about language_level. Complied with it by adding `#cython: language_level=3` at the start of every .pyx and .pxd.

Also rename the cy/*testing* files to cy/*checks* since nose detected them as tests to run which caused automated tests to fail.


ajgpitch:
@Ericgig thanks for sorting this out.
Merging now

PR #929, opened 2018-10-25
Updated plot_fock_distribution

Rajiv-B:
removed the offset value 0.4 to center the fock state distribution on the ticks for the bar plot.


ajgpitch:
Do we have any idea why this 0.4 offset is there?

ajgpitch:
I will try and understand this better then

PR #930, opened 2018-11-03
Qutip 4.3.x

Chapa1:
Dear ladies and gentlemen,
Is it possible to calculate expactation values of time-dependent operators (using, for example, mesolve)?
I do not know how to evaluate <psi | O(t) | psi>, where O(t) is time-dependent and psi is the solution of the Schroedinger equation.


ajgpitch:
This seems like a help group question.
closing

PR #934, opened 2018-11-08
Fast Husimi Q using precomputation

dweigand:
A lot of the cost of the Husimi Q function does not depend on the state. If it is called many times (e.g. for a density matrix or when doing multiple states), this can be used for a speedup. I wrote a function to precompute everything that does not depend on the state and store it as a 3d array. The Q function itself is then a dot product between the last axis of the precomputed array and a pure state.

Performance: Everything tested by computing Q of a coherent state with N dimensions, evaluated at 401 points in x and y, respectively.
Speedup is 3x for N=10, 4x for N=20, 7x for N=50. 
For density matrices, precomputing is already paid for by the first call. 

Possible further changes:
Because precomputing is always worth it if the state is a density matrix, one could do that automatically. Because the memory cost can be large (len(xvec) * len(yvec) * N * 128 bit) , I still leave the decision to the user.

The third if statement in _qfunc_pure can be avoided at a small performance cost in either the polyval or np.dot variant. This seemed the best compromise between performance and nice code.


qutip/wigner.py
```diff
@@ -440,15 +448,15 @@ def qfunc(state, xvec, yvec, g=sqrt(2)):
...
+        qmat = _qfunc_pure(state, amat, amat_pwr)
     elif isoper(state):
         d, v = la.eig(state.full())
         # d[i]   = eigenvalue i
         # v[:,i] = eigenvector i
 
         qmat = zeros(np.shape(amat))
         for k in arange(0, len(d)):
-            qmat1 = _qfunc_pure(v[:, k], amat)
+            qmat1 = _qfunc_pure(v[:, k], amat, amat_pwr)
```
Ericgig: Why not compute amat_pwr before the loop for large enough state. 
From my tests for 10x10 computing amat_pwr first is already faster.

Ericgig:
Without documentation I would not expect anybody to use it. 
For density matrices I would reverse the default: precomputation unless the users says not to. 
Maybe check `amat_pwr` size and build it only if under 1G?

Could you add a test to check both method give the same output.

nathanshammah:
Hi @dweigand, this looks interesting. It would you be nice to push this forward, adding some tests for this and integrating @Ericgig suggested changes.

I am wondering whether there is a more intuitive (explicit) option than `qfunc_amat `. Something like a Boolean `precompute`?

Writing some documentation for this, in https://github.com/qutip/qutip-doc, would allow users to discover this option more easily at http://qutip.org/docs/latest/guide/guide-visualization.html#husimi-q-function, it's very nice you have some performance checks.

nathanshammah:
Thank you. The checks fail with this message `ERROR: Husimi Q: Compare with/without precomputation for rand. ket`.

dweigand:
Hi, I fixed the bug and improved the handling of the default cases.

If the user wants to compute a very large Q func (e.g. 512*512 points, State size >256), a ResourceWarning will be given and the function falls back to the old method as that would use >1GB memory. I did only a manual test of this behavior, as a test function would be very slow.

There is an automated test to ensure that precompute() raises a MemoryError for very large Q, if it is not manually set to use >1GB Memory. At least on my end, all the tests are now running fine.

I might add some documentation in the future, but I am currently in the middle of writing my PhD Thesis, so no guarantees.

dweigand:
Concerning the failed test: I suspect this is because the test temporarily uses almost 1Gb of memory.
Can you confirm this as a potential issue? In that case, I could change the test to reduce memory usage.

nathanshammah:
The tests ran fine, besides for the option with OpenMP, where it gave 
```
/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/qutip-4.5.0.dev0+2fe7701-py3.7-linux-x86_64.egg/qutip/__init__.py:166: UserWarning: matplotlib not found: Graphics will not work.
  warnings.warn("matplotlib not found: Graphics will not work.")
Calibrating OPENMP threshold...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

dweigand:
@nathanshammah: It seems that travis fails before it even gets to the testing, when calibrating openmp. I cannot replicate the error on my end, the calibration takes only about a second for me.

I did make 535010d more similar to 5501b2e by no longer using numpy.testing.assert_raises, numpy.testing.assert_warns and a ResourceWarning. Instead I use a try-except construction now.

That seems to have fixed things, so I suspect that importing/calling one of the functions or the warning makes openmp fail on the rig used by travis. As I'm pretty sure the reason for the previous failure is either a numpy or openmp issue and therefore a hard fix, I would propose to simply not use those functions, the try-except construction does well enough.

qutip/wigner.py
```diff
@@ -35,8 +35,9 @@
            'spin_wigner', 'wigner_transform']
 
 import numpy as np
-from numpy import (zeros, array, arange, exp, real, conj, pi,
-                   copy, sqrt, meshgrid, size, conjugate,
+import warnings
+from scipy import (zeros, array, arange, exp, real, conj, pi,
```
Ericgig: Most of these should be imported from `numpy` not `scipy`. As of scipy 1.4 they raise a [deprecation warning](https://scipy.github.io/devdocs/release.1.4.0.html#deprecated-features) on use and will be removed in scipy 2.0.

nathanshammah: @dweigand can you do this simple change? Then we could merge if it is ok for all.

nathanshammah:
We could merge or this [issue](https://travis-ci.org/qutip/qutip/jobs/640430207?utm_medium=notification&utm_source=github_status) needs to be solved, @Ericgig ?

Ericgig:
I have the same error in the current master, seems parallel can sometime freeze, not sure why or how common it is.

Ericgig:
This is the issue fixed by #1169, you can merge @nathanshammah

nathanshammah:
@Ericgig shall we merge it against the qutip-4.5.X branch, instead master branch? Or let me know if it is instead better to merge to master and be part of a micro (patch) release.

Ericgig:
It depend how far you are into 4.5. If it is too late for 4.5.0 then master, if not then both.

qutip/wigner.py
```diff
@@ -603,20 +633,86 @@ def qfunc(state, xvec, yvec, g=sqrt(2)):
 # |psi>   = the state in fock basis
 # |alpha> = the coherent state with amplitude alpha
 #
-def _qfunc_pure(psi, alpha_mat):
+def _qfunc_pure(psi, alpha_mat, precompute=False):
```
jakelishman, line 636: `None` is the correct type to indicate absence of value, rather than `False` - it's a Python guarantee that all references to `None` are referentially equal.

qutip/wigner.py
```diff
@@ -35,8 +35,9 @@
            'spin_wigner', 'wigner_transform']
 
 import numpy as np
+import warnings
 from numpy import (zeros, array, arange, exp, real, conj, pi,
-                   copy, sqrt, meshgrid, size, conjugate,
+                   copy, sqrt, meshgrid, size, polyval, fliplr, conjugate,
                    cos, sin)
 from scipy import polyval, fliplr
```
jakelishman, line 42: How come `polyval` and `fliplr` are imported from both numpy and scipy?  Should probably just be `numpy`.

qutip/wigner.py
```diff
@@ -603,20 +633,86 @@ def qfunc(state, xvec, yvec, g=sqrt(2)):
...
+            + f"{max_memory} MB.Turn precomputation off with precompute=False "
+            + f"or use a larger max."
+        )
+    X, Y = meshgrid(xvec, yvec)
+    amat = 0.5 * g * (X - Y * 1j)
 
-    return real(qmat) * exp(-abs(alpha_mat) ** 2) / pi
+    powers = np.arange(n)
+    precomputed = np.power(np.expand_dims(amat, axis=-1), powers)
+    precomputed /= sqrt(factorial(arange(n)))
```
jakelishman, line 713: Is this possibly dangerous?

It hits the limit of floating-point computation earlier than it needs to.  For one,
```python
np.sqrt(scipy.special.factorial(np.arange(n)))
```
is technically a quadratic dependency, but its run cost is completely overshadowed by the other stuff so it's not a big deal.  _Somewhat_ more importantly, it starts doing funny business at `n` is 171 due to the maximum factorial that can be represented by a double-precision float, even though we then take the square root.  We can extend the limit to 302 by instead doing
```python
precomputed[:, :, 2:] /= np.cumprod(np.sqrt(np.arange(2, n)))
```
That said, it's not _so_ important because we're using it as a divisor, and floating point multiplies overflow to `inf`, and it's defined for all cases that `x / inf == 0.0` except for `x` is `inf` itself.  This could become a problem if `xvec` and `yvec` are sufficiently wide, though - it's possible for the powers call to overflow and produce `inf`, then get divided by an `inf`, even though the true result is representable by floating-point calculations.

I'm not certain that this is very important (presumably no-one uses values of `n` that large), but I saw a value of 256 in a test, and the current form technically can overflow that.  Just wanted to point it out.

jakelishman (commented):
I think this PR must have fallen through the cracks from our side, sorry.  @dweigand, are you still around and interested in getting this merged?  If not, I'll take it over and fix up the last minor bits before QuTiP 4.7 (with you getting credit for the contribution, of course).

I've left a few minor comments and one more major query in the review notes.  I think the major query is about code that's been copied from existing stuff, but I don't know a huge amount about using the Husimi Q function in real-world situations, so I thought I'd flag it to check.

There's also some docstring formatting bits that I didn't comment on inline, but I can do if you'd like to finish off the PR.  They can be a little difficult to describe, so if I don't hear from you in a while, I'll just make the changes myself.

qutip/wigner.py
```diff
@@ -581,16 +594,33 @@ def qfunc(state, xvec, yvec, g=sqrt(2)):
 
     qmat = zeros(size(amat))
 
+    if precompute is None:
+        # Decide whether precomputation should be used, if not already set
+        precompute = isoper(state)
+
+    memory = len(xvec) * len(yvec) * state.shape[0] * 16 / 1024**2
+    if (precompute is True) and (memory > 1024):
```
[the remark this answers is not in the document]
hodgestar: It would be nice if there was a way for a user to ask for precompute to happen even if memory usage is `> 1024`. My suggestion is to have precompute never happen if `precompute=False` is passed, always happen if `precompute=True`, and happen if `isoper(state) and memory > 1024` is `precompute=None`.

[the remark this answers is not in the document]
jakelishman: There is, as I understand it - you just manually have to call `qfunc_precompute` with the increased memory limits, and pass that as the argument to this function.  Now I think of it, though, that means you can't do `if precompute` or `if precompute == True`, which presumably the author realised (since `x is True` isn't a very natural expression).  Still, `precompute is True` _feels_ wrong, and it could perhaps be a little more obvious from within the code that `precompute` is allowed to be both a Boolean and a numpy array (I know the docstring says it).

I'm not super opposed to it being a little difficult and non-intuitive to ask for precomputation of over a gigabyte, including needing you to be very explicit.  The amount of memory stored is an implementation detail, and the size could be rather surprising, so gating it behind an "advanced" function to prevent you from crashing your Python session accidentally doesn't seem terrible (though I agree too that a hard-coded warning value is very much nonideal).

Issue #940, opened 2018-11-24
Overlap of density matrices


goerz:
Consider the following example:
~~~
>>> import qutip
>>> ket = (qutip.tensor(qutip.ket('0'), qutip.ket('0')) - 1j * qutip.tensor(qutip.ket('1'), qutip.ket('1')))/np.sqrt(2)
>>> rho = qutip.ket2dm(ket)
>>> rho.overlap(rho)
Traceback (most recent call last):
  File "<ipython-input-57-e8684f44b66f>", line 1, in <module>
    rho.overlap(rho)
  File "/Users/goerz/anaconda3/lib/python3.5/site-packages/qutip/qobj.py", line 1488, in overlap
    raise TypeError("Can only calculate overlap for state vector Qobjs")
TypeError: Can only calculate overlap for state vector Qobjs
~~~

This should return the Hilbert-Schmidt overlap, defined as
~~~
>>> (rho.dag() * rho).tr()
0.9999999999999996
~~~

For proper density matrices, the `dag()` is superfluous, but I'd leave it in for the overlap of general operators.

PR #941, opened 2018-11-27
__array__ for Qobj

Ericgig:
Added the __array__ method to Qobj so that np.array(Qobj) return the array.
Asked in issue 937


Ericgig:
This branch also include qobj.overlap(qobj) for density matrix and operator (Overlap of density matrices #940).
I planned to do it in another branch, but had a problem with git...

ajgpitch:
This looks fine to me
@Ericgig thanks for doing this

PR #942, opened 2019-01-18
fix rare bad random Qobj

Ericgig:
In some rare case, the matrix obtain from random_objects.rand_stochastic would have an empty row or column. This would cause the automated test to fail (about 0.8% of the time.)

Small fix that ensure the matrix always has a least 1 element on each rows and columns.


sahmed95:
This faces the same issue as other new PRs which are failing on Mac due to how the test suites handle C code. I am guessing that since the tests pass on Linux, we can merge this? Till we find a fix for the test suite issue, we may need to ignore tests on Mac. @ajgpitch I can merge this if this is okay.

PR #943, opened 2019-01-18
Bug fix for `piqs`

sahmed95:
In `qutip.piqs` the dok matrices should be complex valued in-general. This was pointed out by Jose and @nwlambert when trying to create the CSS state. The imaginary part of the density matrix was missing. 
@nathanshammah


yosefrajwulf:
Thank you @sahmed95 for the quick turnaround on this!

sahmed95:
@yosefrajwulf Thanks for pointing out the bug. I got an unrelated error in MacOS so restarting the tests. Hopefully, they pass and we can merge this.

PR #949, opened 2019-02-13
Remove -march=native from compiler flags

### LINKED ISSUE #920: Optimization flags in setup.py should be completely avoided
Hard-coding compiler flag `-march=native` in setup.py completely destroys possibility to set up Qutip on heterogeneous cluster. In general, it brings a lot of problems for people that don't have a good experience in debugging "illegal instruction" errors, that often happen, if you compile the module on different machine than you use.

If you are sure you need optimized build for localhost, you might use
```
export CFLAGS="-O3 -march=native"
export CXXFLAGS="$CFLAGS"
pip install qutip
```
instead or provide separate option for setup.py script.

slavoutich:
Due to the fact that this flag is hard-coded to the `setup.py`, I am completely unable to install QuTiP to the cluster environment. The only way to overcome it is clone the repo and remove it by hands.

Closes #920

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -159,7 +159,7 @@ if (sys.platform == 'win32' and int(str(sys.version_info[0])+str(sys.version_inf
     _compiler_flags = ['/w', '/Ox']
 # Everything else
 else:
-    _compiler_flags = ['-w', '-O3', '-march=native', '-funroll-loops']
+    _compiler_flags = ['-w', '-O3', '-funroll-loops']
     if sys.platform == 'darwin':
         # These are needed for compiling on OSX 10.14+
         _compiler_flags.append('-mmacosx-version-min=10.9')


quantshah:
@slavoutich Thanks for the PR. @ajgpitch, can we merge this? Or was it hard-coded for some reason?

ajgpitch:
I have been discussing this with @Ericgig, so I would like to hear his opinion first.
This is not the only place compiler flags are set, although this seems to be the only one where `march=native` is set.
Maybe we need a more comprehensive review of how and why we set compiler flags and how this can be overridden by users.
However, no-one seems has put forward any reason for keeping this flag, so I am happy for this to be merged as is.

slavoutich:
Well, `-march=native` gives you couple of percents of performance in the compiled binary (I wonder, did anyone benchmark how much, actually?), but it is a promise "I will never run this binary on another machine, than this". It is better not to do it by default. Even worse if it propagates somewhere on Conda: typically it uses some modern Inter servers to build packages, and if one has AMD CPU, it will definitely segfault with an illegal instruction. I vaguely remember such problems with conda-forge packages, but not sure if it was QuTiP or some other package with the same problem.

Ericgig:
I believe this is the only `-march=native` we have. It will cause problem if left here so it's better to remove it. 

However it may create a bigger slowdown than expected. If you look at [qutip/cy/src/zspmv.cpp](https://github.com/qutip/qutip/blob/master/qutip/cy/src/zspmv.cpp), you will see low level optimization that only activate on certain cpus. This is the main function used in solver (mat*vec) and can be 30% slower or faster than than scipy's version depending on the situation. Without this low level optimization, we are essentially doing the same thing as scipy so should get similar times.

ajgpitch:
I think that @Ericgig makes a good point though. I think we should make the default install from source compile with `-march=native`. I think in the majority of cases this will be the best option for users. However, we need someway to turn it off, so that the conda-forge build will be more portable, and for the use-case illustrated in #920.

Ok, I am going to merge this now, while we think of a more complete solution.

slavoutich:
Thanks a lot!

PR #957, opened 2019-02-26
Docstring sigmap

nathanshammah:
Fix typo in docstrings of `sigmap()` referring incorrectly to `sigmam()`.


ajgpitch:
And then the MacOS tests went and failed when this was merged in, so there is still an issue

Issue #892, opened 2018-07-10
matplotlib is missing from requirements in setup.py


yurivict:

sahmed95:
Yes, that is to keep the installation light. Matplolib is surely used for the visualization but most runtime modules can do without it.

yurivict:
> Matplolib is surely used for the visualization but most runtime modules can do without it.

There is syntax for optional dependencies in ```setup.py```, you should consider using it.

nathanshammah:
Closing as it is addressed by @abhisekupadhyaya PR #967.

PR #961, opened 2019-03-02
Adding Optional dependency for matplotlib.

agarwal1998:
issue - #892 
Inserted the syntax for optional dependency in setup.py and requirements.txt

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,8 @@ MINOR = 4
 MICRO = 0
 ISRELEASED = False
 VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
-REQUIRES = ['numpy (>=1.8)', 'scipy (>=0.15)', 'cython (>=0.21)']
+REQUIRES = ['numpy (>=1.8)', 'scipy (>=0.15)', 'cython (>=0.21)','matplotlib(>=1.2.1)']
+EXTRAS_REQUIRE = {'graphics':['matplotlib(>=1.2.1)']}
 INSTALL_REQUIRES = ['numpy>=1.8', 'scipy>=0.15', 'cython>=0.21']
 PACKAGES = ['qutip', 'qutip/ui', 'qutip/cy', 'qutip/cy/src',
             'qutip/qip', 'qutip/qip/models',
@@ -273,6 +274,7 @@ setup(
     package_data = PACKAGE_DATA,
     zip_safe = False,
     install_requires=INSTALL_REQUIRES,
+    extras_require=EXTRAS_REQUIRE,
     **EXTRA_KWARGS
 )
 _cite = """\
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 cython>=0.21
 numpy>=1.8
 scipy>=0.15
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -60,7 +60,8 @@
 MICRO = 0
 ISRELEASED = False
 VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
-REQUIRES = ['numpy (>=1.8)', 'scipy (>=0.15)', 'cython (>=0.21)']
+REQUIRES = ['numpy (>=1.8)', 'scipy (>=0.15)', 'cython (>=0.21)','matplotlib(>=1.2.1)']
```
ajgpitch: Does this line not make this a required, rather than optional module?

nathanshammah: It was also my question. I wonder whether the presence of INSTALL_REQUIRES and REQUIRES is redundant?

https://stackoverflow.com/questions/41268863/difference-between-extras-require-and-install-requires-in-setup-py 
There might be differences between `python setup.py install` and `pip install qutip` or `conda install qutip`.  
https://github.com/pypa/pip/issues/3147?

ajgpitch: I just tried looking up this `requires` argument. Cannot find much about. Seems maybe it is / was used by `distutils` rather than `setuptools`. I cannot remember why we still have the `distutils` option (see lines 34-47), most discussions say use `setuptools` over `distutils`.
@cgranade  was the last person to do an overhaul of our `setup.py`
I would suggest that this PR would be fine only adding the `extras_require`

nathanshammah:
Somehow tests pass here, also for MacOS. Some of the previous PRs have passing tests now. I am rerunning some of them.

nathanshammah:
On a subsequent run, the build on Travis CI failed. 

The two runs were prompted by the same branch, minutes apart, but differed for the version of Python that Travis CI picked up (3.72 working, 3.73 failing) and some minor decimal difference in release version of one dependency (Cython if I remember correctly). 

Making a branch with a special .travis.yml file with additional osx_image instructions for the Mac OS build might help to sort this out. Two days ago Travis switched to Python 3 as default installation language and unexpectedly yesterday I noticed that some of the failing builds on previous branches were successful. We are always building with Python 3, as it says in the change log, I am just wondering if some dependency was doing something incorrectly.

nathanshammah:
Copying here the message of the failed build, as I cannot find the message of the successful one in the build history (since it refers to the same PR, it looks like it was not kept). 

```
QuTiP Version:      4.4.0.dev0+109c09d
Numpy Version:      1.16.2
Scipy Version:      1.2.1
Cython Version:     0.29.7
Matplotlib Version: None
Python Version:     3.7.3
Number of CPUs:     2
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True
Installation path:  /Users/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/qutip-4.4.0.dev0+109c09d-py3.7-macosx-10.7-x86_64.egg/qutip
```
from https://travis-ci.org/qutip/qutip/builds/521534881

nathanshammah:
This seems addressed by #967.

PR #962, opened 2019-03-05
WIP: Hierarchical Equations of Motion (new features and restructuring)

quantshah:
I would like to add the Underdamped Lorentzian spectral density to the HEOM and realised that there should be some discussion on the design.

For this specific PR, I am in two minds: 

A. I follow the previous code and make a `HSolverUnderDampedBrownian` class. This will be quite different from the HSolverDL class as now, we will take the coefficients ck and vk separately for the non-Matsubara part and then the Mastsubara part and combine them to get the RHS instead of computing them from the coupling strength as before.

B. Rewrite a new `Heom` class with the `configure` method taking the ck and vk values and not caring about the spectral density. We add some additional functions to compute the cks and vks based on Drude-Lorrentz, Underdamped Lorrentzian or Ohmic spectral density. We can even do a numerical fitting for the coefficients using the `bath_correlation` function that I added now for any arbitrary spectral density and then feed it to the `configure` method of the new `Heom` class to build the RHS. This is a more generic version of the HEOM which separates the method from specific spectral densities and is close to implementing the following version that is simplified from @nwlambert 's paper :

screenshot 2019-03-05 14 41 25

I am leaning more towards B and re-structuring the HEOM functions to seperate the SD and the implementation itself. This has some advantages in the future too perhaps if we want to work specificially on the HEOM part without caring about the form of the spectral density.

The second point is:

Most of QuTiP's solvers are function based eg., `mesolve`. In `qutip.piqs` however, we tried to do things with the class called `Dicke`. We defined a `pisolve` method similar in spirit to `mesolve` which takes in an initial state, tlist and computes the evolution. For many sophisticated approaches like `piqs` or `heom`, there needs to be some amount of pre-processing such as computing the Liouvillian/Lindbladian or the levels of the Hierarchy that needs to be done to get the RHS which is solved to get the dynamics. I saw similar approaches in `qutip.pdpsolve` with `StochasticSolverOptions`.

I really like the `HEOMSolver` class defined here and was wondering if we can we have a similar abstract class which can be shared across all future solver methods and not just HEOM or the stochastic solver? I add an example below.

@ajgpitch Does this make sense and should we strive to get some uniform class for all future solvers? 
@nathanshammah @nonhermitian 

```
class Solver(object):
    """
    A super-class specifying the basic methods required in a solver.

    Parameters
    ----------
    name: str
        The name of the specific solver the sub-class represents, eg., `heom`.
        default: None
    """
    def __init__(self, solver=None):
    	self.solver = solver

    def configure(self, *args, **kwargs):
    	"""
    	Configures the solver by computing the necessary objects required
    	to run the `solve` method.
    	"""
    	raise NotImplementedError("This is a abstract class only. "
        "Use a subclass, for example `Dicke.configure`")

    def solve(self, initial_state, tlist, options, *args, **kwargs):
    	"""
    	Runs the solver to compute the evolution using the initial state and tlist.
    	"""
    	raise NotImplementedError("This is a abstract class only. "
        "Use the solve method from a subclass, for example `Dicke.solve`")
```

### DIFF
--- a/qutip/nonmarkov/heom.py
+++ b/qutip/nonmarkov/heom.py
@@ -747,7 +747,7 @@ class HSolverUB(HEOMSolver):
     HEOM solver based on the underdamped Brownian motion spectral density.
     
     We need to express the correlation funtion as a sum of exponentials.
-    Two types of exponentials can be constructed a non-Matsubara part
+    Two types of exponentials can be constructed - a non-Matsubara part
     (two terms) and an infinite Matsubara sum. The non-Matsubara exponents
     can be analytically computed from the parameters of the spectral density
     and temperature while the Matsubara terms have to be approximated by
@@ -760,13 +760,68 @@ class HSolverUB(HEOMSolver):
 
     :math:`J(\omega) = \frac{\gamma \lambda^2 \omega}{(\omega^2
             - \omega_0^2)^2 + \gamma^2 \omega}`
+
+    Parameters
+    ==========
+    H_sys: Qobj
+        The system hamiltonian.
+
+    coup_op: Qobj
+        The coupling between system and bath.
+
+    coup_strength: Float
+        Coupling strength.
+
+    temperature: float
+        The temperature of the bath.
+
+    lam0: Float
+        Coupling strength.
+
+    N_cut: Integer
+        Cutoff parameter for the Hierarchy.
+
+    N_exp: Integer
+        Number of Matsubara terms.
+
+    cut_freq: float
+        The cutoff frequency for the spectral density.
+
+    cav_freq: float
+        The frequency of the single cavity mode.
+
+    cav_broad: float
+        The broadening of the single cavity mode.
+
+    planck, boltzmann: float
+        The value of the planck and boltzmann constants used.
+
+    renorm: bool
+        Renormalize the coefficients of HEOM method.
+
+    band_cut_approx: bool
+        Approximation to cut off Hierarchy.
+
+    options : :class:`qutip.Options`
+        With options for the solver.
+
+    progress_bar:
+
+    stats:
+
+    ck2: array
+        A 1D array of the matsubara coefficients.
+
+    vk2: array
+        A 1D array of the matsubara frequencies.
     """
 
-    def __init__(self, H_sys, coup_op, coup_strength, ck1, vk1,
-                 ck2, vk2,
-                 temperature, N_cut, N_exp, cut_freq, planck=1.0,
-                 boltzmann=1.0, renorm=True, bnd_cut_approx=True,
-                 options=None, progress_bar=None, stats=None):
+    def __init__(self, H_sys, coup_op, coup_strength, temperature,
+                 N_cut, N_exp, cut_freq, cav_freq, cav_broad,
+                 planck=1.0, boltzmann=1.0,
+                 renorm=True, bnd_cut_approx=True,
+                 options=None, progress_bar=None, stats=None,
+                 ck2=None, vk2=None):
 
         self.reset()
 
@@ -790,7 +845,13 @@ class HSolverUB(HEOMSolver):
         self.N_cut = N_cut
         self.N_exp = N_exp
         self.cut_freq = cut_freq
-        self.configure(ck1, vk1, ck2, vk2)
+        self.cav_freq = cav_freq
+        self.configure(H_sys, coup_op, coup_strength, temperature,
+                       N_cut, N_exp, cut_freq, cav_freq, cav_broad,
+                       planck, boltzmann,
+                       renorm, bnd_cut_approx,
+                       options, progress_bar, stats,
+                       ck2, vk2)
 
     def reset(self):
         """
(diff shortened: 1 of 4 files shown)


qutip/nonmarkov/heom.py
```diff
@@ -624,3 +626,173 @@ def _pad_csr(A, row_scale, col_scale, insertrow=0, insertcol=0):
         raise ValueError("insertrow must be >= 0 and < row_scale")
 
     return A
+
+
+class HSolverUnderDampedLorrentzian(HEOMSolver):
```
ajgpitch: I think you want Lorentzian (single r)

ajgpitch:
I think it may be better to have a single class, but we can think about that at some future time.
I am happy with this method for now.
I will take a better look at this next week.

quantshah:
Okay. I will go ahead and push the rest of the code then. Thanks for taking a look.

ajgpitch:
@quantshah I really like your option B idea, but for now I think we should do whatever is quickest to get something in the dev branch that can be used in some notebook to accompany your paper.

I am also **very** much behind the Solver class. I think we have some work to do to convince @Ericgig , who is currently re-writing all the solvers. Perhaps we should start a discussion about this on Slack. There are lots of methods that could be used across solvers. I think the main interface can still be function based, but the code structured in classes

nwlambert:
Just some quick comment, for B, when we get to it, perhaps the general solver can just take arbitrary prefactors/correlation functions, and then we can have special pre-defined cases based on spectral density parameter cases for the following:

1) overdamped + explicit number of matsubara modes  (basically the default code we have in qutip now)
2) overdamped + fitted matsubara 
3) underdamped + explicit number of matsubara modes 
4) underdamped + fitted matsubara (what shahnawaz is uploading now)
5) in principle we can do fitting for arbitrary spectral densities. this might be better as an example notebook though, as it its a little hard to guarantee sensible results.  Ditto with the fitting cases above?

Other stuff that could be added:
1) The pade decomposition (this is just an alternative decomposition to the normal matsubara one which converges more quickly) https://aip.scitation.org/doi/10.1063/1.3484491
2) Alightly more generic ''ishizaki-tanimura'' boundary condition.  this basically approximates any residual matsubara contributions as white noise/lindblads, to give better convergence/detailed balance, etc.   This is already in the default code, but there's some funny cases which I don't know how to deal with (e.g., what happens when those residual contributions are negative?)

quantshah:
Hi, This is not ready for review so I added a new tag `WIP`. The tests etc are not proper and I just wanted to get a basic version out. The latest commits are @nwlambert's code to construct the Liouvillian for the Heom. I will start refactoring things now and update the tag to `review in progress` when this is ready.

quantshah:
So, I have a basic first implementation now that our paper is out: https://arxiv.org/abs/1903.05892

We have a special `HSolverUnderdampedBrownian` and I compare it with the more general `Heom` which takes in some ck and vk and gives the same results. For fun, I try to reproduce some plot from the paper where we see the effect of having the Matsubara terms in the picture vs removing them. I need to do more testing and re-factoring but we can start thinking of a more general `Heom` class now. Any comments will be helpful. 

```python
import numpy as np

from qutip import Qobj, sigmaz, basis, expect
from qutip.nonmarkov.heom import (Heom, HSolverUnderdampedBrownian)
from qutip.solver import Options
import warnings
from qutip.operators import sigmax, sigmaz

```

# Parameters


```python
Q = sigmax()
wq = 1.
wrc = 1.
Hsys = 0.5*wq*sigmaz()
initial_ket = basis(2, 1)
Nc = 9
rho0 = initial_ket*initial_ket.dag()
Nexp = 4
lam, gamma, w0 = 0.2, 0.05, 1.
omega = np.sqrt(w0**2 - (gamma/2.)**2)
a = omega + 1j*gamma/2.
wcav = omega
lam_renorm = ((1/np.pi)*(-gamma*lam**2*(4*np.arctan(0.5*gamma/wrc) \
    -2*np.pi)/(8*0.5*gamma*wrc)))
tlist = np.linspace(0, 20, 1000)

ck1 = np.array([0.0 + 0.j, 0.02000625])
vk1 = np.array([-0.025 + 0.99968745j, -0.025 - 0.99968745j])
ck_mats = np.array([-0.00011662, -0.00020182])
vk_mats = np.array([-0.34959062, -1.75226554])

options = Options(nsteps=1500, store_states=True, atol=1e-12, rtol=1e-12)
```

# Keep the Matsubara terms


```python
mats_solver1 = HSolverUnderdampedBrownian(Hsys, Q, lam_renorm,
                                     ck1, -vk1, ck_mats, -vk_mats,
                                     0., Nc, Nexp, 1, options=options)

mats_solver2 = Heom(Hsys, Q, np.concatenate([ck1, ck_mats]),
               np.concatenate([-vk1, -vk_mats]),
               ncut=Nc,
               lam=lam_renorm)

tlist = np.linspace(0, 200, 1000)
mats_output1 = mats_solver1.run(rho0, tlist)
mats_output2 = mats_solver2.run(rho0, tlist)
mats_result1 = (np.real(expect(mats_output1.states, sigmaz()))+1)/2
mats_result2 = (np.real(expect(mats_output2.states, sigmaz()))+1)/2
```

# Ignore Matsubara


```python
no_mats_solver1 = HSolverUnderdampedBrownian(Hsys, Q, lam_renorm,
                                     ck1, -vk1, ck_mats*0, -vk_mats*0,
                                     0., Nc, 4, 1, options=options)

no_mats_solver2 = Heom(Hsys, Q, ck1,
                       -vk1, ncut=Nc, lam=lam_renorm)

no_mats_output1 = no_mats_solver1.run(rho0, tlist)
no_mats_output2 = no_mats_solver2.run(rho0, tlist)
no_mats_result1 = (np.real(expect(no_mats_output1.states, sigmaz()))+1)/2
no_mats_result2 = (np.real(expect(no_mats_output2.states, sigmaz()))+1)/2
```

    /Users/shahnawaz/Dropbox/dev/qutip/qutip/nonmarkov/heom.py:924: RuntimeWarning: divide by zero encountered in double_scalars
      * cin * (spre(Q).data - spost(Q).data)))
    /Users/shahnawaz/Dropbox/dev/qutip/qutip/nonmarkov/heom.py:918: RuntimeWarning: divide by zero encountered in double_scalars
      * cin * (spre(Q).data - spost(Q).data)))
    /Users/shahnawaz/anaconda3/lib/python3.7/site-packages/scipy-1.2.1-py3.7-macosx-10.7-x86_64.egg/scipy/integrate/_ode.py:1009: UserWarning: zvode: Repeated convergence failures. (Perhaps bad Jacobian supplied or wrong choice of MF or tolerances.)
      self.messages.get(istate, unexpected_istate_msg)))
    /Users/shahnawaz/anaconda3/lib/python3.7/site-packages/scipy-1.2.1-py3.7-macosx-10.7-x86_64.egg/scipy/integrate/_ode.py:1009: UserWarning: zvode: Illegal input detected. (See printed message.)
      self.messages.get(istate, unexpected_istate_msg)))


# Plots


```python
plt.plot(mats_result1, "--", label = "Specialized Underdamped solver (Matsubara)")
plt.plot(mats_result2, label = "General HEOM solver (Matsubara)")

plt.plot(no_mats_result1, "--", label = "Specialized Underdamped solver (No Matsubara)")
plt.plot(no_mats_result2, label = "General HEOM solver (No Matsubara)")

plt.legend()
plt.show()
```


output_8_0




```python

```

nathanshammah:
What is the fate of this PR, also in view of more recent developments?

nwlambert:
Soon (TM).  The refactor is all done (thanks to @tehruhn ).  We are just mostly working up the example notebooks.  

There is some uncertainty about how to include this in qutip, given recent discussions about qutip "prime" and external modules. But we will get it uploaded, and see what happens.

nathanshammah:
@nwlambert @tehruhn @quantshah do you suggest closing this PR and addressing it from a new one, given theses changes? If so, can you close this PR?

nwlambert:
Yeah, if @quantshah  is ok with it, that is probably tidier

quantshah:
Hi, I think its better to close this and continue in a new PR.

quantshah:
@tehruhn Please link the PR here.

PR #971, opened 2019-04-08
fixed and updated some links to docstrings

bdmj12:


nathanshammah:
Seems the required changes to make the links work. We will check if they work as expected once the documentation is built in version 4.4. Thank you.

Issue #958, opened 2019-02-27
Unit testing software: nose, nose2, pytest 


nathanshammah:
This is post for discussion over whether it might be beneficial to switch from `nose` to another unit testing software. 

- QuTiP is using nose for unit testing. A [disclaimer](https://nose.readthedocs.io/en/latest/) on that project homepage warns that nose might not be maintained in the future. 

- [`nose2`](https://nose2.readthedocs.io/en/latest/differences.html) is the successor to `nose`. It is actively maintained. There are some [differences](https://nose2.readthedocs.io/en/latest/differences.html).  

- [pytest](https://pytest.readthedocs.io/en/latest/) is another open-source solution. By comparing [nose2](https://github.com/nose-devs/nose2/graphs/contributors?from=2010-08-01&to=2019-02-27&type=a) and [pytest](https://github.com/pytest-dev/pytest/graphs/contributors?from=2009-04-05&to=2019-02-27&type=a) activity it seems that the latter has grown and is actively updated. 

I wonder whether it would make sense to migrate to a more popular (possibly more robust?) software and how much work this would mean. Other implications are not clear to me at this moment.

ajgpitch:
I think we should swap, but I don't think it's necassarily high priority.

My reckoning is that nose2 could be easier to switch to than pytest. I don't think that anything we do is particularly special. It might be a case of suck it and see.

quantshah:
I have been trying out pytest for the last couple of days and we can start by using it for the testing as it is now. We need to replace installation [lines](https://github.com/qutip/qutip/blob/master/.travis.yml#L23) like this in the .travis.yml file to install pytest:
```
conda install mkl blas=*=mkl numpy scipy pytest cython coveralls pytest-cov
```

and then change the runner script to use pytest instead of `nose` by changing:
```
nosetests --verbosity=2 --with-coverage --cover-package=qutip qutip
```
to
```
pytest --verbosity=2 --cov=qutip qutip
```

BassemSafieldeen:
> 
> 
> I have been trying out pytest for the last couple of days and we can start by using it for the testing as it is now. We need to replace installation [lines](https://github.com/qutip/qutip/blob/master/.travis.yml#L23) like this in the .travis.yml file to install pytest:
> 
> ```
> conda install mkl blas=*=mkl numpy scipy pytest cython coveralls pytest-cov
> ```
> 
> and then change the runner script to use pytest instead of `nose` by changing:
> 
> ```
[...]

PR #972, opened 2019-04-08
WIP: nose to pytest: first step

BassemSafieldeen:
In .travis.yml: replaced nose installation lines with pytest installation lines.
Suggested in issue #958.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ stage_generic_linux: &stage_generic_linux
   # command to run tests
   script: 
     - python -m qutip.about
-    - pytest --verbosity=1  --cov=qutip qutip
+    - nosetests --verbosity=2 --with-coverage --cover-package=qutip qutip
   install:
     - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
     - bash miniconda.sh -b -p $HOME/miniconda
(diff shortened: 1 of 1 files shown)


quantshah:
Hi, thanks for the PR. Needless to say this isn't expected to be that easy and without hiccups. But hopefully it works out. I will have a better look and get back to this.

nathanshammah:
@BassemSafieldeen thank you for addressing this enhancement. It would be nice to find out what's broken with pytest.

BassemSafieldeen:
I think the problem is that ["pytest interprets the function name as a file name and it fails because the file does not exist." ](https://github.com/pytest-dev/pytest/issues/1110#issue-109663494). I am thinking about a fix.

BassemSafieldeen:
Ok, that particular problem has been fixed. Now the build fails for 1 function: TestPulseOptim

quantshah:
Hi, this is great. @agpitch, do you have any idea about the test? Looks like something related to optimal control and you could provide some insight. We can investigate the issue separately and fix this. Good job on the PR. Moving to PyTest will be great for us in the long run.

nathanshammah:
- All builds fail a test regarding optimal control in `` fails with `AttributeError: 'TestPulseOptim' object has no attribute 'tmp_dirs'`.
- The MacOs build still exits with an error once it encounters `ptrace`.

nathanshammah:
Cc @tehruhn.

nathanshammah:
Thank you @BassemSafieldeen, this has been solved by #1131.

BassemSafieldeen:
Happy to have helped!

PR #974, opened 2019-04-11
codeclimat config

Ericgig:
Will take some trial and error to make this useful but has some interesting stats.


nathanshammah:
Since codeclimate is since already an API in the repository, I am merging this PR. So far I did not personally really benefit from the tool. If we opt to remove it in the future, we can remove the badge, but since it's implemented right now, it looks fine to have it.

PR #976, opened 2019-04-12
Updated qobj docstrings with hyperlinks (Issue 71)

bdmj12:
updated docstrings in qutip/qobj.py containing instances of qobj with hyperlinks, as per issue 71


nathanshammah:
In the future, better to create a branch of your master, and then do a `push origin [branch-name]`, opening the PR that way.

PR #985, opened 2019-04-20
WIP: Travis osx

nathanshammah:


nathanshammah:
This is just a test. 
Since many failed Travis builds had a malloc error, also according to [this StackOverflow thread](https://stackoverflow.com/questions/19840671/malloc-error-incorrect-checksum-for-freed-object-object-was-probably-mod/19841133) it seemed that XCode was the source of the problem. 

I added as osx environment in the travis configuration file xcode 10.2 (should be on osx 10.14 Mojave, not sure though). Build still fails, now at 
```
Qobj subtraction ... python(2491,0x1192585c0) malloc: Incorrect checksum for freed object 0x7ffe6ff1ad50: probably modified after being freed.
Corrupt value: 0x3ff0000000000000
python(2491,0x1192585c0) malloc: *** set a breakpoint in malloc_error_break to debug
/Users/travis/.travis/functions: line 104:  2491 Abort trap: 6           nosetests --verbosity=2 --with-coverage --cover-package=qutip qutip
The command "nosetests --verbosity=2 --with-coverage --cover-package=qutip qutip" exited with 134.
```
which seems a pretty basic test.

PR #990, opened 2019-05-15
dynamic args

Ericgig:
This PR is part of  PR #969 , splitting that PR in two parts.

It contains:
- `qobjevo`'s args that update with the state. (replacing rhs_with_state functionality) 
- method `norm` renamed to `_cdc` as it was not used as a norm.
- cython side of expect method no longer take `isherm` as an arguments.
- security check in propagator for `H` as a function with `batch` method.
- `propagator` uses `sesolve` capability to evolve operators.
- floquet now call `sesolve` instead of `mesolve` with no collapse.

PR #969, opened 2019-04-04
Qobjevo solvers

Ericgig:
Rewrote sesolve, mesolve and mcsolve to use QobjEvo.
Compatible with previous version, solvers' test pass without modification.

Feature/imporvement:
- state can be obtain as an argument as a vector or Qobj.
- expectation values also can be arguments.
- The system is saved in a object to be reuse.
- mcsolve is fully class based. (but the use as object is not documented)
- s/mesolve follow a similar pattern to be easily made to class.
- Experimental: diagonalized mcsolve.

Issue #964, opened 2019-03-08
Branch for next major version features


ajgpitch:
We (mainly @Ericgig ) are developing some changes that would only be included in the next major release. It would make sense to merge these into one branch, but not the `master` branch, so that we can test their compatibility. Maybe `development` is a good name, although we often refer to our `master` as the 'development' branch.

I have seen some template for this in another project, but cannot remember where. If anyone has some ideas for this, then would be good if they could share here

skyme5:
Why not create a feature branch?

PR #991, opened 2019-05-15
Qoe mcsolve

Ericgig:
Rewrote, cythonized and simplified `mcsolve`.
`mcsolve` implementation based on `qobjevo`.
It is now class based with the general structure I would propose for solvers.
It has (experimental, hidden for now) support for diagonalized evolution.
Collapse history can be obtained by the Hamiltonian.
Otherwise the same algorithm than the previous version.

Between #990 and #969

### DIFF
--- a/qutip/mcsolve.py
+++ b/qutip/mcsolve.py
@@ -487,17 +487,19 @@ class _MC():
     def final_state(self):
         dims = self.psi0.dims[0]
         len_ = self._psi_out.shape[2]
-        if not self._experimental or self.options.average_states:
-            dm_t = np.zeros((len_, len_), dtype=complex)
-            for i in range(self.num_traj):
-                vec = self._psi_out[i,-1,:] # .reshape((-1,1))
-                dm_t += np.outer(vec, vec.conj())
-            return Qobj(dm_t/self.num_traj, dims=[dims, dims])
-        else:
-            psis = np.empty((self.num_traj), dtype=object)
-            for i in range(self.num_traj):
-                psis[i] = Qobj(self._psi_out[i,-1,:].reshape((-1,1)))
-            return psis
+        dm_t = np.zeros((len_, len_), dtype=complex)
+        for i in range(self.num_traj):
+            vec = self._psi_out[i,-1,:]
+            dm_t += np.outer(vec, vec.conj())
+        return Qobj(dm_t/self.num_traj, dims=[dims, dims])
+
+    @property
+    def runs_final_states(self):
+        dims = self.psi0.dims[0]
+        psis = np.empty((self.num_traj), dtype=object)
+        for i in range(self.num_traj):
+            psis[i] = Qobj(self._psi_out[i,-1,:].reshape((-1,1)), dims=dims)
+        return psis
 
     @property
     def expect(self):
@@ -587,13 +589,16 @@ class _MC():
 
         if options.steady_state_average:
             output.states = self.steady_state
-        elif options.average_states:
+        elif options.average_states and options.store_states:
             output.states = self.states
         elif options.store_states:
             output.states = self.runs_states
 
         if options.store_final_state:
-            output.final_state = self.final_state
+            if not self._experimental or options.average_states:
+                output.final_state = self.final_state
+            else:
+                output.final_state = self.runs_final_states
 
         if options.average_expect:
             output.expect = [self.expect_traj_avg(n) for n in ntraj]
--- a/qutip/mcsolve.py
+++ b/qutip/mcsolve.py
@@ -34,28 +34,25 @@
 __all__ = ['mcsolve']
 
 import os
-from types import FunctionType
 import numpy as np
 from numpy.random import RandomState, randint
 from scipy.integrate import ode
-import scipy.sparse as sp
 from scipy.integrate._ode import zvode
-from scipy.linalg.blas import get_blas_funcs
+
+from types import FunctionType, BuiltinFunctionType
+from functools import partial
+
 from qutip.qobj import Qobj
+from qutip.qobjevo import QobjEvo
 from qutip.parallel import parfor, parallel_map, serial_map
-from qutip.cy.spmatfuncs import cy_ode_rhs, cy_expect_psi_csr, spmv, spmv_csr
-from qutip.cy.codegen import Codegen
-from qutip.cy.utilities import _cython_build_cleanup
-from qutip.cy.spconvert import dense2D_to_fastcsr_cmode
-from qutip.solver import Options, Result, config, _solver_safety_check
-from qutip.rhs_generate import _td_format_check, _td_wrap_array_str
-from qutip.interpolate import Cubic_Spline
+from qutip.cy.mcsolve import CyMcOde, CyMcOdeDiag
+# cy_mc_run_ode = None
+from qutip.sesolve import sesolve
+from qutip.solver import (Options, Result, ExpectOps,
+                          solver_safe, SolverSystem)
 from qutip.settings import debug
 from qutip.ui.progressbar import TextProgressBar, BaseProgressBar
-from qutip.fastsparse import csr2fast
 import qutip.settings
-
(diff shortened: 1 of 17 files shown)


fminga:
I am reviewing the pull request with @nathanshammah: 
- test on the latest notebook on Schrodinger cats produced expected results in comparable times;
- test on other bosonic models (where numerical instability may be present, `H.shape= (500, 500)`) gave expected results (checked with `mcsolve` of Qutip 4.3.1) and in agreement with semi-analytical results;
- I notice that the structure of the `Result.expect` class changes according to the option `average_states`: if `True`, the result is a ndarray of dimension `len(e_ops)`. If  `average_states` is `False`, the array has dimension ntraj x `len(e_ops)`. 
This is in contrast with the previous choice for `mcsolve`, where the result for `average_states`:  `True` where array of dimension 1 x`len(e_ops)`. Clearly, the option implemented makes more logical sense, but a deprecation warning may be better for backward compatibility.

WIP: test on spin systems.

qutip/stochastic.py
```diff
@@ -105,7 +105,7 @@ def stochastic_solvers():
         Scheme keeping the positivity of the density matrix. (smesolve only)
         -Order strong 1.0?
         -Code: 'rouchon', 'Rouchon'
-        Eq. 4 of arXiv:1410.5345 with eta=1 
+        Eq. 4 of arXiv:1410.5345 with eta=1
```
nathanshammah: Please add published reference too: Phys. Rev. A 91, 012118, (2015).

qutip/stochastic.py
```diff
@@ -244,7 +244,7 @@ class StochasticSolverOptions:
 
     store_measurement : bool (default False)
         Whether or not to store the measurement results in the
-        :class:`qutip.solver.SolverResult` instance returned by the solver.
+        :class:`qutip.solver.Result` instance returned by the solver.
```
nathanshammah: Thanks for spotting this.

qutip/stochastic.py
nathanshammah: This occurs in the mcsolve PR but seems related to the other solvers.

qutip/mcsolve.py
```diff
@@ -80,10 +74,9 @@ def step(self, *args):
         self.call_args[2] = itask
         return r
 
-
-def mcsolve(H, psi0, tlist, c_ops=[], e_ops=[], ntraj=None,
-            args={}, options=None, progress_bar=True,
-            map_func=None, map_kwargs=None,
+def mcsolve(H, psi0, tlist, c_ops=[], e_ops=[], ntraj=0,
+            args={}, options=Options(),
+            progress_bar=True, map_func=parallel_map, map_kwargs={},
```
nathanshammah: `map_func` default changed from `None` to `parallel_map `. @fminga is this related to that modification you were proposing of avoiding parallel for one trajectory?

fminga: More or less, my point is that the default `map_func` is always `parallel_map`. Probably we can change it to `serial_map` if `ntraj=1`. This would allow to easily run in parallel several different single trajectories (for example, with different parameters) with `parfor` or `parallel_map`.

qutip/mcsolve.py
```diff
@@ -368,964 +234,548 @@ class _MC():
...
+
+        if options.steady_state_average:
+            output.states = self.steady_state
+        elif options.average_states:
+            output.states = self.states
+        elif options.store_states:
+            output.states = self.runs_states
+
+        if options.store_final_state:
+            output.final_state = self.final_state
```
fminga: We could implemente the option that if  `store_final_state=True` and `average_states=True`, `Results.final_state` is a density matrix.
Instead, if `average_states=False`, the final results are the wavefunctions at the last time.

fminga:
@Ericgig we verified the code also for spins.
The code is consistent with previous versions and it seems to run slightly faster when `e_ops` is given.
I checked the results with `mcsolve` from Qutip 4.3.1, also I checked the density matrix with respect to `mesolve` and `steadystate `

However, when the `e_ops` are not given, and the option `average_states=True`, the `mcsolve` in your version requires much more time than in the 4.3.1 case.
Please, check that this is not a machine dependent bug. 
I wrote a code which does that for a simple three spin system:
https://gist.github.com/fminga/c3e75a75dc528cb509c7d82d3eb9823b

There is, moreover, a minor inconsistency: when the option  `average_states=False` and `store_final_state=True`, one would expect to obtain a set of wavefunctions, instead the code returns the density matrix at the last time.

qutip/stochastic.py
Ericgig: Updated it in #990.
Change in 990 are included in 991, then 969 include both.
master -> 990 -> 991 -> 969

qutip/mcsolve.py
Ericgig: Do it later at line 413.

qutip/mcsolve.py
Ericgig: There are ntraj final wavefunctions, returning a list of wavefunctions won't cause confusion with other solvers output which return a list of state for each time?
For now I will code it but not activate it.

Ericgig: With commit "final_state average flag", `mcsolve` take `_exp=True/False`. On `True` act as you described.

Ericgig:
@fminga Thank you for catching the slow down, It should be better now.

qutip/mcsolve.py
fminga: @Ericgig  : sorry, maybe what I meant was not clear.

We have three possible options which can occur: `store_states`, `store_final_state`, `average_states`.
What I think should be implemented is the following:
- if `store_states` and `store_final_state` are both `True`, the `Results.states` is an array with the states for each time step, while `Results.final_state` contains the results only for the last time.
- if `store_states=False` and `store_final_state=True`, the `Results.states` should be empty and `Results.final_state` contains the final state.

In this sense, there should be no ambiguity between `store_states` and `store_final_state`.
On top of that, the option `average_states` can be used to say if we want the wavefunctions to be averaged, or if we want the non-averaged output (i.e., one list of Qobj per trajectory). Of course, we can  either keep or remove `_exp`, whatever seems more natural.

Moreover, in this version the option `store_final_state=True` and `store_states=False`, returns a `Results.states` which is non-empty (please, find a code here https://gist.github.com/fminga/cfacb90ac4d9498fe514b6694fbee275).

fminga:
@Ericgig : the averaged trajectory is now much faster (local with the previous code https://gist.github.com/fminga/c3e75a75dc528cb509c7d82d3eb9823b
give me a speedup also with respect to Qutip 4.3.1). Thank you very much!

qutip/mcsolve.py
Ericgig: There is also the options `steady_state_average` that affect `results.states` and has priority over both `store_states` and `average_states`...
For now, I will let it like this, but we can set the default value of `_exp=True` later if better.

I hope that we use an object interface later. With that we will be able to do:
```
final_state_averaged = system.final_state
final_state_runs = system.runs_final_states
```
Thus simplifying the use of options. 

Non-empty `Results.states` is expected. By default `store_final_state=False` and `store_states=False` so there would be no outputs if no `e_ops` are given. So when no options are given, with `e_ops` you get the expectations values, otherwise you get the states. This behavior is common with other solvers (`mesolve`, `sesolve`) . If the user only want the final state, he could set `tlist=[0,end]`.

fminga: Perfect! Thank a lot!

nathanshammah:
I am happy to merge into master, I only want to double check with @ajgpitch @quantshah about #964. 
All issues seem to be resolved and work fine. I was a bit We thoroughly tested the solver and now I am more than confident into merging it into master. If I understand correctly we can then move on to assess #990.

Ericgig:
No, this PR include #990. 
Once This is merged, only the updated citation (Phys. Rev. A 91, 012118, (2015).) will be left to merge from #990.

nathanshammah:
So #990 can be merged after this one, and then it's time to assess #969, right?

PR #994, opened 2019-05-17
Piqs prune

nathanshammah:
Eliminate the function `prune_eigenstates()` as it is not correct. Not affecting any other part of code.


quantshah:
This looks good. Merging now.

PR #995, opened 2019-05-18
Some patches for qip.Gate

BoxiLi:
Some small changes:

1. remove repetitive assignments of `self.arg_value` and` self.arg_label`
2. move the description to the definition of class so the document will show it
3. raise the corresponding error if targets is None while required or not an integer, e.g `Gate("SWAP")` or `Gate("SWAP", [0.5, 2])`

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -50,22 +50,22 @@ class Gate(object):
     
     Parameters
     ----------
-    name : String
+    name : string
         Gate name.
-    targets : List or Integer
+    targets : list or int
         Gate targets.
-    controls : List or Integer
+    controls : list or int
         Gate controls.
-    arg_value : Float
+    arg_value : float
         Argument value(phi).
-    arg_label : String
+    arg_label : string
         Label for gate representation.
     """
 
     def __init__(self, name, targets=None, controls=None, arg_value=None,
                  arg_label=None):
         """
-        Creates a gate with specified parameters.
+        Create a gate with specified parameters.
         """
         self.name = name
         self.targets = None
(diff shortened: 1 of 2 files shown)


qutip/qip/circuit.py
```diff
@@ -31,9 +31,11 @@
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
-import numpy as np
+from collections.abc import Iterable
```
nathanshammah: Can you update this import according to #1003 ?

qutip/qip/circuit.py
```diff
@@ -45,46 +47,50 @@ class Gate(object):
     """
     Representation of a quantum gate, with its required parametrs, and target
     and control qubits.
+    
+    Parameters
+    ----------
+    name : String
```
nathanshammah: lower capital unless this is a class. String -> string. Same applies for other parameters belows.

qutip/qip/circuit.py
```diff
@@ -45,46 +47,50 @@ class Gate(object):
...
+        Gate name.
+    targets : List or Integer
+        Gate targets.
+    controls : List or Integer
```
nathanshammah: Integer might not be defining an object (but I might be wrong). Maybe `int` is more appropriate? But maybe also with a float that is an integer, the function works.

qutip/qip/circuit.py
```diff
@@ -45,46 +47,50 @@ class Gate(object):
...
+    arg_value : Float
+        Argument value(phi).
+    arg_label : String
+        Label for gate representation.
     """
 
     def __init__(self, name, targets=None, controls=None, arg_value=None,
                  arg_label=None):
         """
         Creates a gate with specified parameters.
```
nathanshammah: Create a gate (etc).

qutip/qip/circuit.py
BoxiLi: I'm a bit confused. In my python version 3.7.3, this warning occurs when I run
`from collections import Iterable`
but doesn't when I run
`from collections.abc import Iterable`
It seems to me that the second one is the up-to-date one?

BoxiLi: @nathanshammah  Thanks for the suggestions. I have corrected all other doc formats. I think the importation I used here `from collections.abc import Iterable` should be the up-to-date one.

Issue #989, opened 2019-05-07
Bloch sphere (not Bloch3d) is visualised only once in jupyter notebook


ziofil:
After creating a Bloch sphere object, I can add vectors and visualise it with the method `show()`. All fine so far.

This however works only the first time. If I call `show()` a second time, it doesn't output anything.

nathanshammah:
Indeed 
```
b = Bloch()
b.show()
```
renders the Bloch sphere. Calling again `b.show()` does not work as you say. 

Interestingly, invoking just
```
b
```
 gives an error, but plots the correct sphere again. 
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/miniconda3/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/miniconda3/lib/python3.6/site-packages/qutip/bloch.py in _repr_svg_(self)
    298         from IPython.core.pylabtools import print_figure
    299         self.render()
--> 300         fig_data = print_figure(self.fig, 'svg').decode('utf-8')
    301         plt.close(self.fig)
    302         return fig_data

AttributeError: 'str' object has no attribute 'decode'
```
Regarding the no-show, I guess there is something to fiddle with at
https://github.com/qutip/qutip/blob/master/qutip/bloch.py#L625-L632

wrzadkow:
I tried to solve this issue. It seems like this is known, and expected, behavior of `plt.show()`, described e.g. here https://stackoverflow.com/questions/50452455/plt-show-does-nothing-when-used-for-the-second-time/51005555

The workaround depends on the user environment. For Ipython one uses `display()` function from `IPython.display`, while in other cases using `fig.show()` instead of `plt.show(fig)` works.

This would however make the code more involved and environment-dependent, so I didn't decide to submit a PR right away, but I can do with the above fix if somebody from qutip team agrees.

nathanshammah:
@wrzadkow thank you. It sounds good to me. Could [detect](https://stackoverflow.com/questions/5376837/how-can-i-do-an-if-run-from-ipython-test-in-python) whether the environment  is `IPython`. 
If you can write a test too for it to avoid environment-dependent errors that would be great.

wrzadkow:
@nathanshammah thanks! Submitted a PR that seems to be fixing the issue in the above mentioned way.

[...]

PR #996, opened 2019-05-19
Address issue #989 (modify Bloch.show() to show the sphere not only once)

wrzadkow:


nathanshammah:
I tested the code locally on my machine (OSX 10.13) on
-  python from Terminal
- jupyter notebook
- IPython
It turns out that `Bloch.render()` followed by `show()` had the same functionality of showing multiple times the Bloch sphere in a Jupyter Notebook but this is now more immediate, as `render` alone would work multiple times in jupyter notebook, but still having `show` not showing anything, whereas in the Terminal `render` would only "reactivate" the plot for `show` to subsequently plot it. I personally like it better as it is, also given that `render` is not as well documented as `show` in the Users Guide or tutorials.

PR #979, opened 2019-04-17
Gates can be added to specified position in a QubitCircuit

abhisekupadhyaya:
This enables QubitCircuit.add_gate() to insert some gate in the middle of the circuit. This should address issue #615 .

PR #1005, opened 2019-05-30
Typo In Specifying the datatype is fixed

abhisekupadhyaya:
Fix for the typo in #979


abhisekupadhyaya:
> Thank you very much. Since we're here, could you change all instances of List into list? Line 59, 61, 215, 217, 262.

abhisekupadhyaya:
This should take care of all the List typo in the file

PR #1012, opened 2019-06-10
OptPulseProcessor for simulating quantum circuit with optimal control pulse

BoxiLi:
Here is the newly implemented method for the `qip` module. I implemented a class `OptPulseProcessor`, which is the subclass of `circuitprocessor` in `qip.models`. It is defined by a set of Hamiltonians, usually determined by the hardware realization. It can take a `QubitsCircuit`, find the corresponding Hamiltonian representation of it (`self.load_circuit`) with `optaoptimize_pulse_unitary ` and then use these control pulses to simulate the circuit at the level of driven Hamiltonian (`self.run_state`). 

A large part of it is actually more general and can be moved to circuitprocessor after I refactor `cqed `and `spinchain`

Glad to hear if you have any suggestions.


ajgpitch:
Hi Boxi,
Are you you proposing this for review and merge now, or is this work in progress? If the latter, then please add 'WIP:' to the start of the PR name.
Thanks

BoxiLi:
@ajgpitch It is still work in progress, thanks for the advice

BoxiLi:
@ajgpitch Hi Alex, this PR is ready to be merged into qip_opt

quantshah:
Closed via #1042

PR #1013, opened 2019-06-11
nosetests segmentation fault in Metrics docstrings

nathanshammah:
Fix a docstring error prompting a segmentation fault when run on python 3.7.3 on Mac Mojave.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Installation
 
 For instructions on how to install QuTiP, see:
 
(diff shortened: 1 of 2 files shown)


Ericgig:
Still failing, can the 
```    
Calculate the propagator U(t) for the density matrix or wave function such
that :math:`\psi(t) = U(t)\psi(0)` or
:math:`\\rho_{\mathrm vec}(t) = U(t) \\rho_{\mathrm vec}(0)`
where :math:`\\rho_{\mathrm vec}` is the vector representation of the
density matrix.```
In propagator.py be the next place to look?

PR #1014, opened 2019-06-11
Updating qip_opt

quantshah:
Updating the qip_opt branch with the master as base.

### DIFF
--- a/qutip/cy/cqobjevo.pxd
+++ b/qutip/cy/cqobjevo.pxd
@@ -41,6 +41,7 @@ cdef class CQobjEvo:
     cdef object dims
     cdef int super
     cdef int num_ops
+    cdef int dyn_args
 
     #cdef void (*factor_ptr)(double, complex*)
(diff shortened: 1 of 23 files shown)


quantshah:
This feels a bit weird. To update the qip_opt branch from the master. But I guess this is how it should be. @ajgpitch @BoxiLi

BoxiLi:
I'm not sure how this should be down. What I was thinking is that there are some changes to `circuitprocessor.py` and `gate.py` in my commits in the above list. And I might work on those files further in the coming weeks. If they are not included in qip_opt, we will have to eventually come back and resolve the confliction. Maybe in the next few months all PR related to qip should be merged to qip_opt instead of master?

quantshah:
Yes @BoxiLi. When you open a PR now related to QIP, open it to be merged in this branch if it is part of the GSoC project. If it is something general or unrelated, you can open the PR against the master.

PR #1021, opened 2019-06-18
Bugs leading to error in building the documentation

BoxiLi:
1. `Qobj` is included separately. `qt` is not defined.
2. See Also: `func_a : Function a with its description.` As described in https://numpydoc.readthedocs.io/en/latest/format.html


nathanshammah:
Thanks for spotting these @BoxiLi.

Issue #1023, opened 2019-06-19
Long simulation time in `mcsolve` for `average_states = False, store_states = True`


nathanshammah:
The latest changes in `mcsolve` with option `average_states = False, store_states = True` is much slower than `average_states = True, store_states = True`. 

Strangely, the cell runs for much longer after the progress bar is completed.

nathanshammah:
https://gist.github.com/nathanshammah/f96f6c1db5c323243a4729467cb7351f

nathanshammah:
Addressed by @Ericgig in the very fast solver in 4.4.0.

PR #1024, opened 2019-06-19
faster mcsolve postprocessing

Ericgig:
Faster output for each runs states.  Issue #1023.

### DIFF
--- a/qutip/cy/spconvert.pyx
+++ b/qutip/cy/spconvert.pyx
@@ -110,6 +110,40 @@ def dense2D_to_fastcsr_cmode(complex[:, ::1] mat, int nrows, int ncols):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
+def dense1D_to_fastcsr_ket(complex[::1] vec):
+    """
+    Converts a dense c-mode complex ndarray to a sparse CSR matrix.
+
+    Parameters
+    ----------
+    mat : ndarray
+        Input complex ndarray
+
+    Returns
+    -------
+    out : fast_csr_matrix
+        Output matrix in CSR format.
+    """
+    cdef int nnz = 0
+    cdef size_t ii, nrows = vec.shape[0]
+    cdef np.ndarray[complex, ndim=1, mode='c'] data = np.zeros(nrows, dtype=complex)
+    cdef np.ndarray[int, ndim=1, mode='c'] ind = np.zeros(nrows, dtype=np.int32)
+    cdef np.ndarray[int, ndim=1, mode='c'] ptr = np.zeros(nrows+1, dtype=np.int32)
+
+    for ii in range(nrows):
+        if vec[ii] != 0:
+            data[nnz] = vec[ii]
+            nnz += 1
+        ptr[ii+1] = nnz
+
+    if nnz < (nrows):
+        return fast_csr_matrix((data[:nnz], ind[:nnz], ptr), shape=(nrows,1))
+    else:
+        return fast_csr_matrix((data, ind, ptr), shape=(nrows,1))
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
 cdef void fdense2D_to_CSR(complex[::1, :] mat, CSR_Matrix * out,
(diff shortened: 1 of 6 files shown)


nathanshammah:
By running the same file as in the gist of issue #1023, now the times are back to comparable with qutip 4.3.1 for the choice `Average States = False, Store States = True` and the storing doesn't get stuck two minutes. 

With the default options, QuTiP 4.4 (both 4.4.0.dev0+e4b2415a and dev0+da029386) is slower than 4.3.1 of a factor 2 (Case 4 below). 
With `Average States = True, Store States = False`, version 4.3.1 is faster than the old patch that is faster than this one. 

Here the full cases:
### Case 1) Average states = False, Store States = True

#### QuTiP 4.3.1:
Total run time:  29.72s
** Real cell time: 30.454669713974 ** 

#### QuTiP 4.4 before this patch:
Total run time:  17.49s
**Real cell time: 127.9s** 

#### QuTiP 4.4 with this patch:
Total run time:  19.96s
**Real cell time: 28.90908122062683**

### Case 2) Average states = False, Store States = False
#### QuTiP 4.3.1:
Total run time:   9.55s
real time 10.40579104423523

#### QuTiP 4.4 before this patch:
Total run time:  20.02s
**Real cell time: 21.579365968704224**

#### QuTiP 4.4 with this patch:
Total run time:  18.61s
**Real cell time: 18.642597913742065**

### Case 3) Average states = True, Store States = True

#### QuTiP 4.3.1:
Total run time:  35.35s
**Real cell time: 65s**

#### QuTiP 4.4 before this patch:
Total run time:  18.35s
**Real cell time:  21s**

#### QuTiP 4.4 with this patch:
Total run time:  19.52s
**Real cell time:  23s** 

### Case 4) Default options
#### QuTiP 4.3.1:
Total run time: 8s
**Real cell time: 8s**

#### QuTiP 4.4 before this patch:
Total run time:  18s
**Real cell time:  22s**

#### QuTiP 4.4 with this patch:
Total run time:  19s
**Real cell time: 19s** 

### Case 5) Average states = True, Store States = False
#### QuTiP 4.3.1:
Total run time: 8s
**Real cell time: 8s**

#### QuTiP 4.4 before this patch:
Total run time:  19s
**Real cell time:  18s**

#### QuTiP 4.4 with this patch:
Total run time:  19s
**Real cell time: 28s** 

I am merging this in any case as it removes the slowing down when storing the states flagged in #1023.

Ericgig:
Could you send me the code you are using for the test.
I find some of the timing strange.

nathanshammah:
Note that all this is when giving `e_ops`. When the `e_ops` are not given, as tested with @fminga in #991, the speedup of the current qutip 4.4 is considerable: Instead of 127s, they are at 47s. The code is in the gist of #1023.

nathanshammah:
Let me know when you are happy to merge, for me it's fine as it is now, as it is improving the current master.

nathanshammah:
This code https://gist.github.com/nathanshammah/f96f6c1db5c323243a4729467cb7351f
Also, let me mention that the progress bar was giving some errors in qutip 4.3.1 and now it does not.

Ericgig:
Should be about the same speed as 4.3 with `e_ops`.

Ericgig:
@nathanshammah as soon as the tests finish if the usual tests pass.

nathanshammah:
It now all fares similarly to 4.3.1. or considerably better, also when `e_ops` are given:

### QuTiP 4.4 with this patch (updated):

* Case 1) Average states = False, Store States = True
**Real cell time: 17s (previously: 28.9s. version 4.3.1: 25s)**

* Case 2) Average states = False, Store States = False
Real cell time: 8s (previously: 18.64s. version 4.3.1: 11s)

* Case 3) Average states = True, Store States = True
**Real cell time: 9s (previously: 23s. version 4.3.1: 65s)**

* Case 4) Default options
Real cell time: 8s (previously: 19s. version 4.3.1: 8s )

* Case 5) Average states = True, Store States = False
Real cell time: 8.7s (previously: 28s. version 4.3.1: 8.7s)

PR #1025, opened 2019-06-21
Add user defined gates, MS gate and fix some bugs

### LINKED ISSUE #914: Custom gates for QubitCircuit
At the moment it is possible to make a custom `Gate` and pass it to a `QubitCircuit` so that it will appear on the diagram.
Is it also possible to connect it to an actual operator, so that when  `QubitCircuit.propagators` is called it also returns the propagator of the custom gate?

BoxiLi:
1. `expand_oper`: expand a qubits operator to dim N
2. User-defined gate can be given by a python function. By passing a list of user-defined gate function to the attribute `QubitCircuit.user_gate`,   `QubitCircuit.propagators()` can return the corresponding matrices. Resolve #914 
3. MS gate and physical single qubit rotation. (Is the name `mc_gate `and `qrot `appropriate?) Resolve #814
4. Fix some bugs in resolving circuit gates. Allow more than one 2-qubits gates in the basis. This is actually already required in `spinchain.py` but covered by a bug

### DIFF
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -527,16 +527,16 @@ def molmer_sorensen(theta, N=None, targets=[0, 1]):
 
     Parameters
     ----------
-    theta : float
+    theta: float
         The duration of the interaction pulse.
-    N : int
+    N: int
         Number of qubits in the system.
-    target : int
+    target: int
         The indices of the target qubits.
 
     Returns
     -------
-    molmer_sorensen_gate : :class:`qutip.Qobj`
+    molmer_sorensen_gate: :class:`qutip.Qobj`
         Quantum object representation of the Mølmer–Sørensen gate.
     """
     if targets != [0, 1] and N is None:
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -44,9 +44,11 @@ from qutip.tensor import tensor
 from qutip.states import fock_dm
 
 
-__all__ = ['rx', 'ry', 'rz', 'sqrtnot', 'snot', 'phasegate', 'cphase', 'cnot',
+__all__ = ['rx', 'ry', 'rz', 'sqrtnot', 'snot', 'phasegate', 'qrot', 
+           'cphase', 'cnot',
            'csign', 'berkeley', 'swapalpha', 'swap', 'iswap', 'sqrtswap',
-           'sqrtiswap', 'fredkin', 'toffoli', 'rotation', 'controlled_gate',
+           'sqrtiswap', 'fredkin', 'ms_gate', 
+           'toffoli', 'rotation', 'controlled_gate',
            'globalphase', 'hadamard_transform', 'gate_sequence_product',
            'gate_expand_1toN', 'gate_expand_2toN', 'gate_expand_3toN',
            'qubit_clifford_group', 'expand_oper']
@@ -509,7 +511,7 @@ shape = [4, 4], type = oper, isHerm = False
                               [0, 0, 0, 1]]), dims=[[2, 2], [2, 2]])
 
 
-def MS_gate(theta, N=None, targets=(0, 1)):
+def ms_gate(theta, N=None, targets=(0, 1)):
     """
     Quantum obejct of a Mølmer–Sørensen gate
 
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -177,6 +177,27 @@ shape = [2, 2], type = oper, isHerm = False
                     dims=[[2], [2]])
 
 
+def qrot(theta, phi, N=None, target=0):
+    """
+    Single qubit rotation driving by Rabi oscillation with 0 detune.
+
+    Parameters
+    ----------
+    phi : float
+        The inital phase of the rabi pulse
+    theta : float
+        The duration of the rabi pulse
+    """
+    if N is not None:
+        return expand_oper(qrot(theta, phi), N=N, targets=target)
+    else:
+        return Qobj(
+            [
+                [np.cos(theta/2.), -1.j*np.exp(-1.j*phi)*np.sin(theta/2.)],
+                [-1.j*np.exp(1.j*phi)*np.sin(theta/2.), np.cos(theta/2.)]
+            ])
+
+
 #
 # 2 Qubit Gates
 #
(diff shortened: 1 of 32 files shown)


qutip/qip/gates.py
```diff
@@ -488,6 +509,29 @@ def sqrtiswap(N=None, targets=[0, 1]):
                               [0, 0, 0, 1]]), dims=[[2, 2], [2, 2]])
 
 
+def MS_gate(theta, N=None, targets=(0, 1)):
```
nathanshammah: what about being more explicit and call it `molmer_sorensen` instead of `MS_gate`? This way we also avoid using initial upper case that is used for Classes.

qutip/qip/gates.py
```diff
@@ -177,6 +177,27 @@ def phasegate(theta, N=None, target=0):
                     dims=[[2], [2]])
 
 
+def qrot(theta, phi, N=None, target=0):
```
nathanshammah: what about being a bit more explicit and call it `qubit_rotate`, `qrotate, or `rotate_qubit`? Maybe `qrot` is fine too.

qutip/qip/gates.py
```diff
@@ -177,6 +177,27 @@ def phasegate(theta, N=None, target=0):
...
+    """
+    Single qubit rotation driving by Rabi oscillation with 0 detune.
+
+    Parameters
+    ----------
+    phi : float
+        The inital phase of the rabi pulse
+    theta : float
+        The duration of the rabi pulse
+    """
```
nathanshammah: Shouldn't it have also "Returns" in the docstring?

qutip/qip/gates.py
```diff
@@ -488,6 +509,29 @@ def sqrtiswap(N=None, targets=[0, 1]):
...
 
 
+def MS_gate(theta, N=None, targets=(0, 1)):
+    """
+    Quantum obejct of a Mølmer–Sørensen gate
+
+    Parameters
+    ----------
+    theta : The duration of the interaction pulse
+    """
```
nathanshammah: Also here: Shouldn't it have also "Returns" in the docstring?

qutip/qip/gates.py
nathanshammah: Add target to docstring.

qutip/qip/gates.py
```diff
@@ -1045,3 +1092,84 @@ def gate_expand_3toN(U, N, controls=[0, 1], target=2):
         p = [p[p1[p2[k]]] for k in range(N)]
 
     return tensor([U] + [identity(2)] * (N - 3)).permute(p)
+
+
+def _check_qubits_oper(oper):
```
nathanshammah: add docstring

qutip/qip/gates.py
nathanshammah: Add ` target ` docstring. Maybe default should be a list instead of a tuple for coherence with rest of code where `integer` is integer or list?

qutip/tests/test_gates.py
```diff
@@ -283,6 +285,35 @@ def _rand_3qubit_gate(N=None, controls=None, k=None):
                         o2 = psi_ref_out.overlap(psi_ref_in)
                         assert_(abs(o1 - o2) < 1e-12)
 
+    def test_expand_oper(self):
```
nathanshammah: Add docstring. This way the test comment will be printed to screen if verbose unit testing is chosen.

qutip/qip/gates.py
```diff
@@ -487,6 +511,29 @@ def sqrtiswap(N=None, targets=[0, 1]):
                               [0, 0, 0, 1]]), dims=[[2, 2], [2, 2]])
 
 
+def ms_gate(theta, N=None, targets=(0, 1)):
+    """
+    Quantum obejct of a Mølmer–Sørensen gate
```
nathanshammah: obejct --> object.

qutip/qip/gates.py
```diff
@@ -488,6 +509,29 @@ def sqrtiswap(N=None, targets=[0, 1]):
...
+    if targets != (0, 1) and N is None:
+        N = 2
+
+    if N is not None:
+        return expand_oper(MS_gate(theta), N, targets=targets)
+    else:
+        return Qobj(
+            [
+                [np.cos(theta/2.), 0, 0, -1.j*sin(theta/2.)],
+                [0, np.cos(tehta/2.), -1.j*sin(theta/2.), 0],
```
nathanshammah: `np.cos(tehta/2.)` should give a bug once run: `tehta` --> `theta`

qutip/qip/gates.py
```diff
@@ -488,6 +509,29 @@ def sqrtiswap(N=None, targets=[0, 1]):
...
+                [0, -1.j*sin(theta/2.), np.cos(tehta/2.), 0],
```
nathanshammah: `tehta` --> `theta`

nathanshammah:
Also: maybe this PR, which seems self-contained with respect to deliverables and independent on the "`qutip.nisq`" module, could be made directly to the master, so that we can already include it in the next QuTiP release.

qutip/qip/gates.py
BoxiLi: `molmer_sorensen` looks good.

qutip/qip/gates.py
BoxiLi: Sorry, it seems that I forgot to push the most recent change

BoxiLi:
@nathanshammah Thanks a lot for the review. I'm not sure how gist works, should I upload only the example or with the definition of `expand_oper`? Maybe we can talk about this tomorrow at the meeting. I have some examples in notebook. 
Yes, you are right, I think this PR could be merged directly to master, but I also used `expand_oper` in my `circuitprocessor` a lot so I may need you to update qip_opt from master branch.

nathanshammah:
Maybe we can change the base branch to master. Never did it before but it seems straightforward https://help.github.com/en/articles/changing-the-base-branch-of-a-pull-request and then update qip opt branch.

BoxiLi:
@nathanshammah @quantshah I tried it, but a lot of commits appear which don't belong to me. They are updates from the master branch.

quantshah:
Great. I was trying to see if I can add commits to this PR directly and it is possible. I will post here how to do that in a while. Just going to resolve these conflicts and rebase against master

BoxiLi:
@quantshah I find it a bit weird because those commits are merged from master to qip_opt and I'm now just trying to merge back to master. Where do those conflicts come from?

quantshah:
The conflicts are because first we branched qip_opt from master, but then qip_opt was not updated to master itself and I tried to update it against master since this will be a PR which gets merged into master and not your GSoC project. This led to many changes which happened over master and were not there in your PR and led to some conflicts between  `master` and `qip_opt` I suppose.

quantshah:
We could take this PR as an example of how to work with a very messy git history. I see some forced pushes which should probably have been rebases to avoid conflicts.

But at this point I would suggest to Boxi that you open a fresh PR branching from the current master on your local qutip repo. So do:

```
git checkout master
git pull upstream master
git push origin master
```

This syncs your master to QuTiP. Then, start a new branch and make the same changes that you did here on that new branch. 

Make the changes, add commit etc.
```
git checkout -b user_gate2
...
git add .
git commit -m "message"
```

and then make a new PR and close this (without merging). Then delete this branch (locally and remote)

```
git branch -D user_gate
git push origin -delete user_gate
```

BoxiLi:
But if I haven't changed any of those files, shouldn't git ignore them and believe that master is the up-to-date one? May the workflow here "master->qip_opt->user_gate->master" is not what git expected. The correct way should be "master->qip_opt->user_gate->qip_opt->master" or "master->user_gate->master".
So I should have started from master if I want to be merged back to master, not through qip_opt.

BoxiLi:
Sure, I'll do it.

quantshah:
> So I should have started from master if I want to be merged back to master, not through qip_opt.

It is not necessary to do so. You just have to make sure that the branch you want to merge to (in this case we changed this midway to `master` from `opt_qip`) is not having conflicts with your code. And maybe that is what happened. Not sure though.

PR #1033, opened 2019-07-01
ENH: Attribution in about

ajgpitch:
QuTiP team members split into groupings based on role
reference to Github for list of contributors


ajgpitch:
@Ericgig sorry for the late hour of the PR, but we have to make sure Franco is credited.

Issue #963, opened 2019-03-07
Travis tests failing for MacOS - incorrect checksum for freed object


ajgpitch:
Originally reported in #944. Tests seem to be failing for MacOS. We get errors like
```
python(2234,0x7fff95aef380) malloc: *** error for object 0x7fde6f2f0ff0: incorrect checksum for freed object - object was probably modified after being freed.
```
The build is pretty much always failing, but sometimes passes.

I think we only get the errors when tests are run, the functions work fine otherwise. I think someone tried pytest and still got same errors. So kind of a nasty one to investigate, but there might be an easy fix if we can work out what to do.

I believe @nathanshammah has done some quite thorough investigation and could probably add to this.

nathanshammah:
I report local tests failing on my MacOS 10.13.6 by running qutip.testing.run().

Tests fail for python 3.7 and 3.6 environments, if QuTiP 4.4. is installed in develop mode (openmp = True). 
Tests fail also for QuTiP 4.3.1 installed with conda. 

### py36
Broken test for python 3.6 with: (1) QuTiP 4.4-dev,  with openmp =True and (2) conda install qutip, QuTiP 4.3.1: 
Error Message below: 
``` 
Metrics: Trace dist. & Fidelity mixed/pure inequality ... ok 
Metrics: Check avg gate fidelities for random ... 
python(35852,0x7fff9757f380) malloc: *** error for object 0x7fe189168a10: incorrect checksum for freed object - object was probably modified after being freed. *** 
set a breakpoint in malloc_error_break to debug 
Abort trap: 6 ----------------------------------------------   
``` 

Building the dev environment: 
``` conda create --name qutip-dev-44-py36  
source activate qutip-dev-44-py36 
conda install python=3.6 cython numpy scipy nose jupyter matplotlib
python setup.py develop --with-openmp
``` 

### py37
Broken test for python 3.7 with: (1) QuTiP 4.4-dev,  with openmp =True and (2) conda install qutip, QuTiP 4.3.1

Error Message below: 
``` 
qutip.tests.test_qobj.test_expect_exception ... ok Qobj data ... ok Qobj type ... ok Qobj Hermicity ... ok 
Qobj unitarity ... ok 
Qobj shape ... ok 
Qobj: multiplication w/ non-square qobj.dims ... ok 
Qobj addition ... ok 
Qobj subtraction ... ok 
Qobj multiplication ... ok 
Qobj division ... ok 
Qobj power ... 
[...]

PR #1034, opened 2019-07-02
Skip avg gate fid tests on mac

ajgpitch:
As we don't have a fix for #963 then this PR skips those tests on the `average_gate_fidelity` on Mac for the time being


Ericgig:
Thank you, the failing tests were bugging me.

Issue #1039, opened 2019-07-08
Testing code errors: TypeError


lrunze:
Hi, 

I installed qutip on my windows system and tried the testing codes and faced the problem (as the figure shows):

TypeError: Cannot convert _UnitedFuncCaller to qutip.cy.cqobjevo_factor.CoeffFunc

![image](https://user-images.githubusercontent.com/13745165/60830800-9a9e5e00-a1ea-11e9-99d7-2ec24390083e.png)

Here is the about() information:

QuTiP: Quantum Toolbox in Python
Copyright (c) 2011 and later.
A. J. Pitchford, P. D. Nation, R. J. Johansson, A. Grimsmo, and C. Granade

QuTiP Version:      4.4.0.dev0
Numpy Version:      1.16.4
Scipy Version:      1.2.1
Cython Version:     0.29.11
Matplotlib Version: 3.1.0
Python Version:     3.7.3
Number of CPUs:     8
BLAS Info:          INTEL MKL
OPENMP Installed:   True
INTEL MKL Ext:      True
Platform Info:      Windows (AMD64)
Installation path:  C:\Anaconda\envs\qutip\lib\site-packages\qutip-4.4.0.dev0-py3.7-win-amd64.egg\qutip

I would appreciate it if you can give me some advice to help me pass the testing codes!

Ericgig:
Hi,
This a manual installation of the master branch?
Are you using the qutip tests in qutip.testing.run(), if so could you tell us which test is failing? (And which similar only are not.)

lrunze:
Hi Ericgig,

Thanks for your response!

I installed the qutip manually with master branch. And yes, I used the qutip.testing.run() to test if the installation and functions are ready.

I re-runned the testing and the error happened on the "**Monte-carlo: Collapse terms constant (func format)**".

Attached file is the error output with context.

[issue1039.txt](https://github.com/qutip/qutip/files/3373987/issue1039.txt)

Thanks!

Ericgig:
Hi Irunze,
PR #1040 should fix that bug. 
Thank you for reporting.
Eric

zhedie:
Has this problem been solved?

> Hi,
> 
> I installed qutip on my windows system and tried the testing codes and faced the problem (as the figure shows):
> 
> TypeError: Cannot convert _UnitedFuncCaller to qutip.cy.cqobjevo_factor.CoeffFunc
> 
[...]

PR #1040, opened 2019-07-09
picking bug on windows

Ericgig:
Should fix the issue #1039 
Added matching automated test.


Ericgig:
To merge to 4.4.X for 4.4.1 instead?

PR #1041, opened 2019-07-10
QobjEvo do not need to start from 0 anymore

Ericgig:
QobjEvo array coefficients needed to start from 0. 
Made it to be able to start from anywhere.


Ericgig:
To merge to 4.4.X for 4.4.1 instead?

PR #1045, opened 2019-07-15
Purity

nathanshammah:
Add a simple method onto the Qobj to calculate the purity of a quantum object, `(rho*rho).tr()`.

Issue #993, opened 2019-05-17
qutip.piqs Dicke.liouvillian() gives wrong result for Liouvillian spectrum and nonlinear functions of steadystate


nathanshammah:
The spectrum of the Liouvillian obtained with qutip.piqs in the permutational symmetric basis from Dicke.liouvillian() or Dicke.lindbladian() is not to be used to look at spectral properties, such as the Liouvillian gap and overall spectrum. Working on a fix and possible disclaimer meanwhile.

nathanshammah:
The optional function `prune_eigenstates` needs to be removed.

nathanshammah:
Also, nonlinear functions, such as the Von Neumann entropy of the steadystate can be wrong. 

Any test on expected values of collective spin operators the steadystate, compared with the steadystate calculated in the full basis, give correct results.

nathanshammah:
A quick patch at least for collective processes would be to change the piqs Liouvillian internal function calculating `collective_emission`, `collective_dephasing`, and `collective_pumping` properties of the `Dicke` are set by using standard `qutip.liouvillian()` with collective operators defined in the Dicke basis (`qutip.piqs.jspin`).

nathanshammah:
Nonlinear functions on the density matrix need to be evaluated on each block by factoring in its degeneracy, given by `state_degeneracy(N,j)` for each block `j` for an ensemble of `N` two-level systems. Working on efficient functions and a notebook to highlight this feature.

nathanshammah:
It was incorrect to close the issue, as only the nonlinear function part.

PR #1046, opened 2019-07-15
Entropy vn dicke

nathanshammah:
Addressing #993: nonlinear functions such as the Von Neumann entropy or purity of a density matrix written in the Dicke basis need to account for the identical copies of the blocks relative to non-symmetrical Dicke states. 

This can be done by cycling into the blocks and applying the correct weight accounting for degeneracy (using `piqs.state_degeneracy`). 

This PR also contains auxiliary functions to visualize and unpack the Dicke-basis density matrix into the full 2^N space still of collective spin but with degenerate blocks. 

It uses the `purity` method proposed in #1045. 

Will add info in User Guide and notebooks to highlight this features that need to be considered to correctly calculate nonlinear functions of a density matrix in the Dicke basis.

### DIFF
--- a/qutip/piqs.py
+++ b/qutip/piqs.py
@@ -58,11 +58,13 @@ from qutip.cy.piqs import (jmm1_dictionary, _num_dicke_states,
                            m_vals, j_vals)
 
 __all__ = ['num_dicke_states','num_dicke_ladders','num_tls','isdiagonal',
-           'mask_dicke_matrix','purity_dicke','expand_dicke_matrix',
-           'entropy_vn_dicke','Dicke','state_degeneracy','m_degeneracy',
-           'ap','am','spin_algebra','jspin','collapse_uncoupled',
-           'dicke_basis','dicke','excited','superradiant','css','ghz',
-           'ground','identity_uncoupled','block_matrix','tau_column','Pim']
+           'mask_dicke_matrix','expand_dicke_matrix',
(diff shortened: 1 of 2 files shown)


nathanshammah:
In the latest commit I refactored the `entropy_vn_dicke` and  `purity_dicke` functions in terms of a more general function, `dicke_trace`, which accounts for adding with the correct weight the blocks of a density matrix in the Dicke basis. The `dicke_trace` function can be used by other similarly nonlinear functions that involve taking the trace of a Taylor-expandable function.

quantshah:
Thanks Nathan. Lets see if the tests pass. Meanwhile, perhaps you could do `black -l 79` to keep the lines only 79 characters long. Also there are some trailing white spaces and spaces in a blank line which seem not to be picked up by black. Could you please quickly check that so that we get the code climate green check? If not then we can still merge this.

nathanshammah:
@quantshah: what about if we make a PR uniforming all code in QuTiP with `black` and automatically enforce  also the 79 line limit? Also, what about adding it to the guidelines for contributions?

quantshah:
I am not completely familiar with black yet and how much it adheres to pep8 so I will keep that for a big PR some other time. Right now we can slowly start blacking our code for the new modules. 

This PR passes all the tests so I will be happy to merge.

quantshah:
@nathanshammah Could you please also take care of the "blank line contains spaces" issues from code climate?

PR #1050, opened 2019-07-17
Time-dependent coefficients without cubic interpolation in QobjEvo

BoxiLi:
Implement an option for using array-like coefficients as step-function instead of cubic-spline:
e.g. 
```
tlist = np.array([0., 1., 2., 3.])
coeff = np.array([1., 2., 3., 3.])
```
The coefficient is 1 in the time interval [0,1), 2 in [1,2) and so on. The last element in `coeff` is just a place holder and has no effect.

@Ericgig Do you think the following might be useful?
1. In `inter.pyx` many functions have two versions, one for complex input and one for float. Now there are 4 pairs of them, will it helps if I try to merge them with fused type in Cython?
2. For non-uniform `tlist`, `_binary_search` is used to find the current index inside the method such as `_spline_float_t_second`, but very often we call this method within a loop for the same `t`. It would improve the performance if `_binary_search` is called outside once for all `t`.

### DIFF
--- a/qutip/cy/inter.pyx
+++ b/qutip/cy/inter.pyx
@@ -196,3 +196,53 @@ cdef complex _spline_complex_t_second(double x,
     cdef complex Mb = M[p] * dt2
     return te * (Mb * te * te + (y[p]   - Mb)) + \
            tb * (Me * tb * tb + (y[p+1] - Me))
+
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+@cython.cdivision(True)
+cdef double _step_float_cte(double x, double[::1] t, double[::1] y, int n_t):
+    if x < t[0]:
+        return 0
+    elif x >= t[n_t-1]:
+        return 0
+    cdef int ind = <int>(x/t[n_t-1]*(n_t-1))
+    return y[ind]
+
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+@cython.cdivision(True)
+cdef complex _step_complex_cte(double x, double[::1] t, complex[::1] y, int n_t):
+    if x < t[0]:
+        return 0
+    elif x >= t[n_t-1]:
+        return 0
+    cdef int ind = <int>(x/t[n_t-1]*(n_t-1))
+    return y[ind]
+
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+@cython.cdivision(True)
+cdef double _step_float_t(double x, double[::1] t, double[::1] y, int n_t):
+    if x < t[0]:
+        return 0
+    elif x >= t[n_t-1]:
+        return 0
+    # TODO this can be moved out for better performance
+    cdef int ind = _binary_search(x, t, n_t)
+    return y[ind]
+
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+@cython.cdivision(True)
+cdef complex _step_complex_t(double x, double[::1] t, complex[::1] y, int n_t):
+    if x < t[0]:
+        return 0
+    elif x >= t[n_t-1]:
+        return 0
+    # TODO this can be moved out for better performance
+    cdef int ind = _binary_search(x, t, n_t)
+    return y[ind]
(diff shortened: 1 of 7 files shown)


qutip/cy/inter.pyx
```diff
@@ -196,3 +196,53 @@ cdef complex _spline_complex_t_second(double x,
...
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+@cython.cdivision(True)
+cdef double _step_float_cte(double x, double[::1] t, double[::1] y, int n_t):
+    if x < t[0]:
+        return y[0]
+    elif x >= t[n_t-1]:
+        return y[n_t-1]
+    cdef int ind = <int>(x/t[n_t-1]*(n_t-1))
```
Ericgig: This will only work if tlist[0]=0, there was a case where this was not so recently and spline function no longer expect tlist to start at 0. Could you make it so tlist cans start from anywhere?

qutip/qobjevo.py
```diff
@@ -139,12 +140,26 @@ def __call__(self, t, args={}):
...
+            self.func = interp1d(
+                self.tlist, self.coeff, kind="previous",
+                bounds_error=False, fill_value=0.)
+        else:
+            self.func = CubicSpline(self.tlist, self.coeff)
 
     def __call__(self, t, args={}):
+        try:
+            (self.func([t])[0])
+        except:
```
Ericgig: If it sometime fails, better raise an exception.

qutip/cy/inter.pyx
```diff
@@ -196,3 +196,53 @@ cdef complex _spline_complex_t_second(double x,
...
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+@cython.cdivision(True)
+cdef double _step_float_t(double x, double[::1] t, double[::1] y, int n_t):
+    if x < t[0]:
+        return y[0]
+    elif x >= t[n_t-1]:
+        return y[n_t-1]
+    # TODO this can be moved out for better performance
```
Ericgig: What do you have in mind for 'move out'?

Ericgig:
Also I would add a note about `args={"_step_func_coeff":True}` in the QobjEvo docstring.
I saw the docstring for args is incomplete so don't add it in the parameter section but in the array coefficient description (line 227).

BoxiLi:
@Ericgig Thanks for the review and comments. I'll make some improvement accordingly.

1. About the test, I think the very last test I added in `test_mesolve`, the case of "mixed callable", uses the `scipy.interpolate.interp1d`. But I agree with you that there should be a separate test in `test_qobjevo` for calling the coefficients. I'll add a test similar to `test_QobjEvo_call` for the step coefficient case.
2. About that "move out", I mean move out the `_binary_search` inside functions as `_spline_complex_t_second`. 

To be more concrete, for example, in `InterCoeffT` we have the following:
```
    cdef void _call_core(self, double t, complex* coeff):
        cdef int i
        for i in range(self._num_ops):
            coeff[i] = _spline_complex_t_second(t, self.tlist,
                                    self.y[i,:], self.M[i,:], self.n_t)
```
For the same time `t` and `tlist`, we call `_spline_complex_t_second` multiple times. Each time we call it, it will have to do a `_binary_search` for `t` in `tlist`, which has complexity log(`len(tlist)`). But since `t` and `tlist` does not change in the function, it's enough to do it only once outside the loop.

qutip/qobjevo.py
BoxiLi: My mistake, used it for debugging, should be deleted

qutip/qobjevo.py
```diff
@@ -139,12 +140,22 @@ def __call__(self, t, args={}):
...
+            use_step_func = 0
+        if use_step_func:
+            self.func = interp1d(
+                self.tlist, self.coeff, kind="previous",
+                bounds_error=False, fill_value=0.)
+        else:
+            self.func = CubicSpline(self.tlist, self.coeff)
 
     def __call__(self, t, args={}):
+        self.func([t])[0]
```
Ericgig: This line is not needed.

Ericgig:
Moving out _binary_search would could be done, but I don't believe it will make visible speed increase. I most cases the matrix operations are the bottleneck.

Fused types are a good idea, but if you still have a lot to do in your project, they are not a priority.

BoxiLi:
Alright, thank you. Let's keep it like this then. If you have no more suggestions, this PR is ready to be merged.

PR #1052, opened 2019-07-18
expand_oper for composite system with arbitrary dimensions

BoxiLi:
1. Generalize expand_oper to arbitrary dimensions instead of only qubits system. Together with user-defined gate, this would allow quantum circuit of n-level-system.
2. A method `expand_oper_periodic` can be used to expand gate for all cyclic permutation of the given target qubits. ([0, 1], [1, 2], [2, 0] for 3 qubits system)

### DIFF
--- a/qutip/qip/gates.py
+++ b/qutip/qip/gates.py
@@ -1166,7 +1166,7 @@ def _targets_to_list(targets, oper=None, N=None):
         targets = list(range(len(oper.dims[0])))
     if not isinstance(targets, Iterable):
         targets = [targets]
-    if not all([isinstance(t, numbers.Integral) for t in targets])):
+    if not all([isinstance(t, numbers.Integral) for t in targets]):
         raise TypeError(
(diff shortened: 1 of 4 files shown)


qutip/qip/gates.py
```diff
@@ -1144,66 +1195,77 @@ def expand_oper(oper, N, targets):
...
+    rest_qubits = list(range(len(targets), N))
+    for i, ind in enumerate(rest_pos):
+        new_order[ind] = rest_qubits[i]
+    id_list = [identity(dims[i]) for i in rest_pos]
+    return tensor([oper] + id_list).permute(new_order)
+
+
+def expand_oper_periodic(oper, N, targets=None, dims=None):
+    """
+    Expand a qubis operator to one that acts on a N-qutbis system for
```
quantshah: qubit

BoxiLi:
@quantshah This RP is ready for merging.

PR #1058, opened 2019-07-26
Revised random ket generators for dims

### LINKED ISSUE #1057: _check_ket_dims fails
The dimension check [here](https://github.com/qutip/qutip/blob/181a98e910a1f34bb29be31b91d61f765d6c5a37/qutip/random_objects.py#L672) should be
`if np.prod(dims[0]) != N1 or np.prod(dims[1]) != 1`

(alternately the whole function could just call _check_dims(dims, N1, 1) )

In particular this makes rand_ket_haar fail if dims is specified, presumably other ket creators as well.

yarrriv:
Revised rand_ket, rand_ket_haar to allow specifying the ket dimensions.

Previously _check_ket_dims was broken, and definition of dims in the two functions was conflicting.

Changes here:
- removed _check_ket_dims, using the more general _check_dims instead.
- Changed the input method for the rand_ket functions to receive either int, as before, _or_ a dims specifications, rather than both. 
- If dims is specified, N is inferred from dims.

Since both functions were broken previously if dims was specified, and this is backwards compatible if only N is specified, it should be entirely backwards compatible.

Fixes #1057

### DIFF
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -265,6 +265,7 @@ def rand_unitary_haar(N=2, dims=None, seed=None):
 
 def rand_ket(N, density=1, dims=None, seed=None):
     """Creates a random Nx1 sparse ket vector.
+    
     Parameters
     ----------
     N : int
@@ -284,15 +285,13 @@ def rand_ket(N, density=1, dims=None, seed=None):
     """
     if seed is not None:
         np.random.seed(seed=seed)
-    if dims:
-        if not N :
-            N = prod(dims[0])
+    if N and dims:
         _check_dims(dims, N, 1)
-    else:
-        if N:
-            dims = [[N],[1]]
-        else:
-            raise TypeError("Must provide either N or dims")
+    elif dims:
+        N = prod(dims[0])
+        _check_dims(dims, N, 1)
+    elif:
+        dims = [[N],[1]]
     X = sp.rand(N, 1, density, format='csr')
     X.data = X.data - 0.5
     Y = X.copy()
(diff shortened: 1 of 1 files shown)


Ericgig:
Hi yarriv,
Thank you for the PR.
I saw that you removed `dims`, this break backward compatibility which I would like to keep for a bugfix like this. 
Could you restore `dims` and `_check_ket_dims` with the version you suggested in #1057.

yarrriv:
Eric,

Sure, I can revert to just that. On the issue page you suggested automatically getting N from the dims specifications so I thought I'd incorporate that.

(Note that currently neither random ket generator works when you supply dims, so changing the definition now isn't breaking anybody's code)

One thing I wasn't sure about was that the documentation for rand_ket seems to imply dims should be just the left dims (i.e. [n1, n2, n3]) while the codes itself assigns Qobj.dims = dims, implying it should be a full dims array (i.e. [[n1,n2,n3],[1,1,1]]). Which one should I expect from the user?

Ericgig:
`dims` is an argument of all random Qobj creator, we should keep it.
It should be in [[m,n],[1,1]] format. documentation is not clear / wrong. If used with this format, it does work, but the check and documentation indicate to use [[n,m]] which is wrong and will cause errors.

What I was thinking was something along the line of:
```
def rand_ket(N=0, ...) : 
    if not N: 
        N = np.prod(dims[0])
```

yarrriv:
Ok. Do you prefer to keep the _check_ket_dims function or eliminate it and use _check_dims(dims,N,1)?

PR #1068, opened 2019-08-09
Fixed PolyDataMapper construction

sjgriffiths:
Attempting to use `add_states()` on a Bloch3d instances results in an exception delayed to when attempting to show it. The line
`cone_mapper = tvtk.PolyDataMapper(input=cone.output)`
is the culprit, as the `input` and `output` attributes were deprecated for VTK 7.x. A quick updating of the attribute names fixes this.


nathanshammah:
#1027

nathanshammah:
Thank you @sjgriffiths.

PR #1070, opened 2019-08-11
Fix shebangs

### LINKED ISSUE #1059: Wrong shebangs in .pyx and .pxd files: #!python
For example https://github.com/qutip/qutip/blob/master/qutip/cy/checks.pyx#L1

They should be ```#!/bin/sh python".

sameshl:
Fix shebangs in `.pyx` and `.pxd` files

changed `#!python` --> `#!/bin/sh python`

closes #1059

### DIFF
--- a/qutip/control/cy_grape.pyx
+++ b/qutip/control/cy_grape.pyx
@@ -1,4 +1,4 @@
-#!python
+#!/bin/sh python
 #cython: language_level=3
 # This file is part of QuTiP: Quantum Toolbox in Python.
 #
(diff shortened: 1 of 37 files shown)


Ericgig:
Hi sameshl,
Why do it need fixing? 
As I understand it, the shebang tell which interpreter should execute the file but:
- None of those file should be executed by itself.
- cython code cannot be be interpreted by a python interpreter, you would need an exotic interpreter like [shython](https://gitlab.com/ole.tange/tangetools/tree/master/shython)

So unless I a missing something, the shebang are useless for cython files.
Wouldn't it be better to remove them?

hodgestar:
Agreed that the shebangs are not useful for these files and should just be removed.

sameshl:
@Ericgig I understand. I thought the same but continued nevertheless as it was mentioned in this issue #1059.

Ericgig:
Replaced by #1091

Issue #1072, opened 2019-08-13
Restructuring QIP module - `device`, `circuit`, `operations`, `measurement` ... 


quantshah:
With the Google Summer of Code 2019 project by @BoxiLi on noise models, we are making some big changes to update the QIP module. One of the things that came up is having a `DeviceProcessor` or `Device`` class which interfaces with hardware implementations directly and runs quantum circuits on models such as SpinChain, cQED. We will also work towards linking the OptPulseProcessor to find optimal pulses to implement a circuit and add a `noise` module.

I am opening this issue to discuss the restructuring of the QIP module and also adding a `measure` function/module (later) to get samples and expectation values from a circuit. We c can use this thread to discuss other issues too regarding the QIP module and how we can interface it better to QuTiP core functionalities. 

- qip
    - device/
        - processor (which has DeviceProcessor (base class for spinchain, cqed), OptPulseProcessor etc)
        - spinchain
        - cqed
    - circuit.py
    - operations.py (gates.py + qft.py)
    - noise.py
    - measurement

@ajgpitch @nathanshammah @qutip/owners 
@BoxiLi please feel free to suggest, change or discuss more on this issue as you continue working on it. Thanks.

BoxiLi:
It seems more natural for me to put gates, qft, measurement all under a module called operations. From the view of quantum information theory, measurement no different than a non-unitary operation.

- operations/
     - gates.py
     - qft.py
     - measurement.py

On the other hand, functions in gate.py seem to be used quite a lot (also inside qutip, like `swap`). I'm not sure if `from qutip.qip.operations import swap` is too cumbersome?

nathanshammah:
I think this has been addressed by #1042 (GSoC project on noise in quantum circuit). Feel free to reopen if this is not the case or it only partially addresses this.

PR #1074, opened 2019-08-16
Deprecate importing qip functions to the qutip namespace

BoxiLi:
This could be a possible solution of deprecating the importing `qip` functions to the `qutip `namespace. I wrote a wrapper in the top-level `__init__.py` so that functions from the `qip` module will be wrapped with a deprecation warning when they are imported to the qutip namespace. In this way the all the members of `qip` module still work as usual and the warning will only be given if they are imported in the wrong way.

The following code will give a warning
```
>>> from qutip import *
>>> cnot()
__main__:1: DeprecationWarning: Importing functions/classes of the qip submodule directly from the namespace qutip is deprecated. Please import them from the submodule instead, e.g. qutip.qip import `func_name`.
Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[1. 0. 0. 0.]
 [0. 1. 0. 0.]
 [0. 0. 0. 1.]
 [0. 0. 1. 0.]]
```
```
>>> import qutip
>>> qutip.cnot()
__main__:1: DeprecationWarning: Importing functions/classes of the qip submodule directly from the namespace qutip is deprecated. Please import them from the submodule instead, e.g. qutip.qip import `func_name`.
Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[1. 0. 0. 0.]
 [0. 1. 0. 0.]
 [0. 0. 0. 1.]
 [0. 0. 1. 0.]]
>>> import qutip  #  warning appears only once.
>>> qutip.cnot()
Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[1. 0. 0. 0.]
 [0. 1. 0. 0.]
 [0. 0. 0. 1.]
 [0. 0. 1. 0.]]
```

The help function still works correctly:
```
>>> help(cnot)
Help on function cnot in module qutip.qip.gates:
cnot(N=None, control=0, target=1)
    Quantum object representing the CNOT gate.

    Returns
    -------
    cnot_gate : qobj
        Quantum object representation of CNOT gate

    Examples
    --------
    >>> cnot()
    Quantum object: dims = [[2, 2], [2, 2]], shape = [4, 4], type = oper, isHerm = True
    Qobj data =
        [[ 1.+0.j  0.+0.j  0.+0.j  0.+0.j]
         [ 0.+0.j  1.+0.j  0.+0.j  0.+0.j]
         [ 0.+0.j  0.+0.j  0.+0.j  1.+0.j]
         [ 0.+0.j  0.+0.j  1.+0.j  0.+0.j]]
```

The only flaw I found is that the traceback will stop at the wrapper function if the arguments are wrong
```
>>> cnot(wrong_arg=1)
__main__:1: DeprecationWarning: Importing functions/classes of the qip submodule directly from the namespace qutip is deprecated. Please import them from the submodule instead, e.g. qutip.qip import `func_name`.
  File "<stdin>", line 1, in <module>
  File "c:\users\boxi\onedrive\studium\qutip-project\qutip\qutip\__init__.py", line 240, in deprecated_func
    return func(*args, **kwargs)
TypeError: cnot() got an unexpected keyword argument 'wrong_arg'
```
Anyone would like to try it out and see if the behaviour is the same? For instance, running `test_entropy.py` now will give a warning because the importation is wrong there. All tests still pass.


qutip/qip_depracation.py
```diff
@@ -0,0 +1,44 @@
...
+# this forces it to show up.
+warnings.simplefilter('once', DeprecationWarning)
+
+
+def _qip_importation_warning():
+    warnings.warn(
+        "Importing functions/classes of the qip submodule directly from "
+        "the namespace qutip is deprecated. "
+        "Please import them from the submodule instead, e.g. "
+        "qutip.qip import `func_name`.", DeprecationWarning, stacklevel=3)
```
nathanshammah: Change
from
```qutip.qip import `func_name`.```
to
```from qutip.qip import `func_name`.```
Also, maybe could be on multiple lines for readability

Also, maybe this option is more user friendly:
```from qutip.qip import *```

BoxiLi: Thanks for the comments. I have updated the code.

BoxiLi:
There is one thing I was wondering, should we leave the functions in  "qip/gates.py" in the namespace of qutip or not? Those functions, such as cnot, are the most often used functions I guess. They are used a lot outside the qip context just for a matrix representation.

Ericgig:
Could you fix the import in `test_entropy.py`. It would be better in we don't raise warning in our tests. Otherwise it seems good.

Ericgig:
You probably need to add `qutip/qip/ops` in `PACKAGES` of setup.py (line 64) for the import to work.

BoxiLi:
Made two changes:
- Move `qutip.qip.gates` to `qutip.qip.operations.gates` as discussed in #1072. Later we might introduce classical control and measurement, which all belong to `operations`. Since we are going to give warning anyway, better give all warning at once.
- Move the deprecation inside the `qip` folder. The behaviour is the same as described at the beginning.
In the end, we are only going to support the use such as
`from qutip.qip.operations import swap`
`from qutip.qip.circuit import QubitCircuit`
`from qutip.qip.device import Processor`
but not
`from qutip import cnot`
`from qutip.qip import cnot`
like in ProjectQ

BoxiLi:
Thanks Eric, you are correct. I didn't know that. I guess that's why it was totally fine on my own machine. Now all the Python 3.7 test passes except for the Python 3.6 one.
I guess it is because 3.7 supports some circular importation. "Circular imports involving absolute imports with binding a submodule to a name are now supported." from https://docs.python.org/3/whatsnew/3.7.html

Ericgig:
Sometime placing the import at the end of the file can solve the circular import. It's bad style but work. We have quite a few at the end of qobj.py.

BoxiLi:
Looks all good now. @Ericgig

PR #1075, opened 2019-08-17
Fix potential cyclic importation in the control module

BoxiLi:
change `from qutip import Qobj` to `from qutip.qobj import Qobj`


ajgpitch:
Very happy to hear that this fixes the issue.

PR #1082, opened 2019-09-09
QobjEvo no longer require cython for string coefficient

Ericgig:
Added the option to import string to python instead of compiling to cython for QobjEvo.
This is default if no cython, or set manually if cython available. `QobjEvoInstance.use_cython = False`
Added test and updated docstring.
Work in parallel.
Did not do any benchmark yet.

### DIFF
--- a/qutip/qobjevo_codegen.py
+++ b/qutip/qobjevo_codegen.py
@@ -77,11 +77,6 @@ def f(double t, args):
                           '<string>', 'exec')
     exec(import_code, locals())
 
-    try:
-        os.remove(filename+".pyx")
-    except:
-        pass
-
     return str_func[0], filename
 
 
@@ -103,12 +98,7 @@ def _compiled_coeffs(ops, args, dyn_args, tlist):
     exec(import_code, locals())
     coeff_obj = import_list[0](ops, args, tlist, dyn_args)
 
-    try:
-        os.remove(filename+".pyx")
-    except:
-        pass
-
-    return coeff_obj, code, filename
+    return coeff_obj, code, filename+".pyx"
 
 
 def _make_code_4_cimport(ops, args, dyn_args, tlist):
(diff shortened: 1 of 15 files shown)


ajgpitch:
Nice work Eric. I had a quick look at the code.
Does this PR mean that string type td will be used to construct Python functions when Cython not available? Just checking that I understand.

Ericgig:
Yes, string will be turned to python function when cython is not available.
It is possible to use python even if cython is available if preferred.

nathanshammah:
Is this merge-ready?

Ericgig:
Wait 1 more commit so I check the test skipped without cython.

Ericgig:
@nathanshammah 
That was a lot more than 1 commit but it work better.
- I added a test suite without cython. It only skip brmesolve time dependent tests.
- Error is raised when trying to use brmesolve without cython, but not at qutip import.

I tested without cython with both conda and pip. It work with both when done properly but conda is not as stable: uninstalling cython can change numpy, scipy, etc.'s versions which break qutip. So I am not sure if it will work properly we downloaded as binaries from conda. But this error should already be present for conda installation.

qutip/qobjevo_codegen.py
```diff
@@ -38,12 +38,48 @@
...
+
+
+def _try_remove(filename):
+    try:
+        os.remove(filename)
+    except Exception:
+        pass
+
+
+def import_str(code, basefilename, obj_name, cythonfile=False):
```
nathanshammah: Can you add docstrings?

Ericgig: I made a few new functions to make codeclimat happy: small simple function, little repeated code.

nathanshammah:
Tests still run fine, also against the current master, cc @tehruhn.

PR #1083, opened 2019-09-10
core: sesolve and codegen little fixes

ajgpitch:
Add import of os in codegen
Remove prints in sesolve


ajgpitch:
@Ericgig I think these are mishaps. I found them when experimenting with the time-dependence options.
Maybe merge these into #1082 ?

PR #1090, opened 2019-09-29
Add measurement functions.

hodgestar:
Hello! This is the first of hopefully a few PRs to add measurement functions to qutip.

This PR adds a new module named `qutip.measurement` that contains two functions:

* `measure(op, state)` that performs a classic quantum measurement as described in most introductory texts (measurement collapses a state to an eigenvector of an hermitian operator and the result is the corresponding eigenvalue). It is intended to simulate a simple measurement that could be performed in a real experiment.

* `measurement_statistics(op, state)` returns information that is not available via a real experiment, but is nevertheless often useful in simulator (the eigenvalues and vectors or projection operators for the measurement operator and the probabilities of each measurement result).

Both functions support either kets or density matrices and return sensible results based on the kind of input.

Not currently supported:

* Partial measurements of a system.
* Destructive measurements (aka von Neumann measurements of the second kind)
* Density matrix ensembles (applying the measurement probabilities to the density matrix rather than making a specific measurement)
* Any custom support for qubit measurement (my plan is to add those to a `qutip.qip.measurements` module once we have the basics of this module agreed on).

Questions / things I'd like feedback on:

* Do the inputs and outputs of the current functions seem intuitive / sensible?
* What are the most important other features to support?
* I don't want to have too many functions, but I also don't necessarily want to cram all possible kinds of measurements into one function. Any thoughts on how to split these up?
* What other unit tests should I be writing?
* I haven't given any thought to performance yet. Anything egregious I should fix now or any important cases to optimize for?

General stylistic feedback welcomed too.


nathanshammah:
Thank you @hodgestar, glad to see many details coming with this nicely drafted PR. 

Functions design seems sensible. 
What is the main difference between `expect` and `measure`? That's the main thing that comes up to my mind. What different question are you answering? I also wonder if `expect` could be called. The description of the `measure` function is a bit general maybe with this regard. But I might have missed the point completely (my bad).  

Tests seem quite elaborate, I wouldn't know how to add more. They can be even very simple in general. 

Somehow `smesolve` tests fail with Travis CI, not sure how this PR could have affected it.

nathanshammah:
Sorry I am missing something:
```python
psi_plus = (basis(2,0)+basis(2,1))/np.sqrt(2)
psi_minus = (basis(2,0)-basis(2,1))/np.sqrt(2)
```
the eigenstates of `sigmax()`. Their braket with `sigmaz()` is zero, e.g. given by `expect(sigmaz(),psi_plus)`. 
Instead,  `measure(sigmaz(),psi_minus)` gives 
```python
(1.0, Quantum object: dims = [[2], [1]], shape = (2, 1), type = ket
 Qobj data =
 [[-1.]
  [ 0.]])
```
I don't get what is the meaining of the first 1 and also of the (-1,0) state in output.

hodgestar:
> the eigenstates of `sigmax()`. Their braket with `sigmaz()` is zero, e.g. given by 
> `expect(sigmaz(),psi_plus)`.

Yes, it's common for the expected value to not be a possible measurement result in the same way that a weighted average is not necessarily one of the values the average is over.

> Instead, `measure(sigmaz(),psi_minus)` gives
> 
> ```python
> (1.0, Quantum object: dims = [[2], [1]], shape = (2, 1), type = ket
>  Qobj data =
>  [[-1.]
>   [ 0.]])
> ```
> 
> I don't get what is the meaining of the first 1 and also of the (-1,0) state in output.

The `1` is the measurement result, i.e. one of the eigenvalues of sigmaz. The `(-1, 0)` state is the corresponding eigenstate. The global phase that makes the eigenstate is `(-1, 0)` instead of `(1, 0)` is mathematically fine, but I'm not sure what if anything we should do about it to surprise humans less.

hodgestar:
> What is the main difference between `expect` and `measure`? That's the main thing that comes up 
> to my mind. What different question are you answering? I also wonder if `expect` could be called. 

`expect` returns the expectation value (i.e. probability weighted average of the measurements).

`measure` returns a *single( measurement (i.e. an eigenvalue and an eigenstate, selected randomly with weighted probabilities).

> The description of the `measure` function is a bit general maybe with this regard. But I might have 
> missed the point completely (my bad).

Open to suggestions. I will also re-read it and see if I can think of something clearer.
 
> Somehow `smesolve` tests fail with Travis CI, not sure how this PR could have affected it.

I'll dig a bit and see if the failure is related somehow.

hodgestar:
Re failing `smesolve` -- the tests pass together locally. I see Travis builds master with the PR merged in. I've merged master in to this branch in case that helps.

Ericgig:
Hi hodgestar,
Thank you for the PR.
The broken tests (smesolve) is not  caused by your addition, you do not have to worry about it.

hodgestar:
@Ericgig Woot. Tx.

qutip/measurement.py
```diff
@@ -0,0 +1,102 @@
...
+        The list of eigenvalues of the measurement operator.
+    eigenstates_or_projectors: List of Qobj
+        If the state was a ket, return the eigenstates of the measurement
+        operator. Otherwise return the projectors onto the eigenstates.
+    probabilities: List of floats
+        The probability of measuring the state as being in the
+        corresponding eigenstate (and the measurement result being
+        the corresponding eigenvalue).
+    """
+    eigenvalues, eigenstates = op.eigenstates()
```
ajgpitch: Check validity of input types here. Check they are `Qobj` - TypeError("xxx should be Qobj") if not. Check `Qobj.type` valid ValueError("xxx should be 'oper'") etc. 
To avoid unnecessary type checking when called internally make a `_measurement_statistics` function that does the actual work.

qutip/measurement.py
```diff
@@ -0,0 +1,102 @@
+# This file is part of QuTiP: Quantum Toolbox in Python.
+#
+#    Copyright (c) 2011 and later, Paul D. Nation and Robert J. Johansson.
```
ajgpitch: I think we are putting the (c) to the "QuTiP Project" these days.

qutip/measurement.py
```diff
@@ -0,0 +1,102 @@
...
+#    PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+#    HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#    SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+#    LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#    DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#    THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+"""
```
ajgpitch: You could add "Original contribution by Simon Cross, yyyy-mm-dd" if you like. It's not really necessary, as GitHub knows all, but feel free to if you want.

qutip/measurement.py
```diff
@@ -0,0 +1,102 @@
...
+    else:
+        projectors = [v * v.dag() for v in eigenstates]
+        probabilities = [(p * state).tr() for p in projectors]
+        return eigenvalues, projectors, probabilities
+
+
+def measure(op, state):
+    """
+    Perform a measurement specified by an operator on the given state.
+
```
ajgpitch: Would be nice to be explicit here about what it does (and what it currently does not do). Could even give a 
```
Example
-------
```

qutip/tests/test_measurement.py
```diff
@@ -0,0 +1,170 @@
...
+    else:
+        projs = ess_or_projs
+        assert_(len(projs), len(pairs.projectors))
+        for a, b in zip(projs, pairs.projectors):
+            assert_(isequal(a, b))
+    assert_almost_equal(probs, probabilities)
+
+
+def test_measurement_statistics_sigmaz():
+    check_measurement_statistics(
```
ajgpitch: Each of the `test_` functions need a docstring. This is displayed when `nose` is running (in some modes).

ajgpitch:
Going forward.... thinking about compound systems.... would seem like a good idea to keep these high level functions that check qobj type and then make appropriate all to internal `_fullsys_measurement_statistics`, `_subsys_measurement_statistics` etc

ajgpitch:
Just an 'out there' question... what about superoperators?

qutip/measurement.py
hodgestar: Corrected.

qutip/measurement.py
hodgestar: I couldn't find other files that have this, so I'll leave myself out rather than make this file a special snowflake. :)

qutip/tests/test_measurement.py
hodgestar: Docstrings added. :)

qutip/measurement.py
hodgestar: Good idea! Added.

qutip/measurement.py
hodgestar: Examples and more of an explanation added.

hodgestar:
> Just an 'out there' question... what about superoperators?

I'm still reading, but I imagine superoperator measurements will correspond to partial measurements on systems where only part of the system is available for measurement (e.g. quantum channels), so perhaps lets do partial measurements next and then superoperator measurements after that?

hodgestar:
@ajgpitch I think I responded to all your comments -- ready for a second round of reviewing.

hodgestar:
@ajgpitch Re-pinging.

ajgpitch:
@hodgestar Hi Simon, really sorry about the delay. 
All the requested changes look good to me. I am happy this is ready to merge this now. 

Normally we ask for a new feature PR to be accompanied by an example notebook and a section in the documentation. I guess we can work on the documentation later, it is a bit of a hassle getting the doc build working. However, an example Jupyter notebook would be really nice to go along with this. Can be quite simple - just take the examples from the docstring. Would be good to show the maths as well. We can then build on the notebook as we add more measuring features.

ajgpitch:
I saw an interesting talk last week showing how measuring a sub-system of entangled systems (qubits) leads to teleportation of states. I guess we need to make sure that occurs when we add sub-system measurement. Would make a nice example too I guess. Let me know if you want me to find out a nice reference for this - I guess there are many.

hodgestar:
Happy to create a notebook with some examples. Should it live at https://github.com/qutip/qutip-notebooks/tree/master/docs/guide ? If so, I'll make a PR there.

For documentation, do I add them at https://github.com/qutip/qutip-doc/blob/master/apidoc/functions.rst?

Re measurement and teleportation -- that sounds like ordinary quantum teleportation, and probably would be a fun test to write once we get there.

ajgpitch:
The notebook book would live in here:
https://github.com/qutip/qutip-notebooks/tree/master/examples
We would then link it in here:
https://github.com/qutip/qutip.github.io/blob/master/tutorials.html
I guess in the Basics section.

As for the qutip-docs:
It would be good to update functions.rst. This is for the auto-generated API docs.

If you are feeling super keen then you could add a guide-measurement.rst in:
https://github.com/qutip/qutip-doc/tree/master/guide
and link it from:
https://github.com/qutip/qutip-doc/blob/master/guide/guide.rst

Barring a few build problems here and there the docs building has been okay recently (I think).
Instructions can be found here:
https://github.com/qutip/qutip-doc/blob/master/README.md

nathanshammah:
It would be nice to merge this if everything is fine. Not sure of why tests are failing, it seems to pick `smesolve` that is unaffected by these changes.

nathanshammah:
Happy to merge this soon if the conflicts are solved. cc @hodgestar @hodgestar. We are aiming for a new qutip release this week.

hodgestar:
@nathanshammah I've merged in master, waiting for builds to complete.

The code climate error is seems largely unrelated to this PR (the PR adds a tiny snippet of code that is similar to 4 other existing pieces of code). I'm happy to refactor this a bit if someone makes a suggestion, but the code climate error seems overly zealous to me.

hodgestar:
Woot, Travis builds have passed, so just the code climate issue to resolve.

sarsid:
@hodgestar any updates on merging this, I'm hoping to extend some of this functionality!

hodgestar:
@ajgpitch I've started a guide and docs in https://github.com/qutip/qutip-doc/pull/113. Would you mind having a quick look and letting me know if the direction is good. I'll try finish the measurement_statistics section tomrrow.

Could we perhaps land this PR with just the guide? I'm happy to add a notebook straight afterwards (and it should be fairly straight forward once we have the guide), I just don't want to end up having too many different pieces in flight and holding up @sarsid more than necessary.

hodgestar:
Guide and docs are ready for review.

ajgpitch:
@hodgestar It seems @nathanshammah is happy with the documentation, and so I am happy to merge this now. Thank you for persisting with this.

hodgestar:
Yay! Thank you! Now to think / discuss the next steps. :D

PR #1091, opened 2019-10-01
Remove wrong shebangs

sameshl:
Fix unnecessary shebangs in `.pyx` and `.pxd` files by removing them.

related to #1059
ref :https://github.com/qutip/qutip/pull/1070


Ericgig:
Look good,
Thank you sameshl.
We can ignore the test errors as they are not related to this PR.

sameshl:
@Ericgig Thanks! Glad I could help.:smile:

Issue #1094, opened 2019-10-08
ode2es fails for certain initial states


peterkirton:
For an initial state for which the elements of the density operator sum to 0 ode2es returns an empty list. Minimal example below

>from qutip import *
>H = sigmaz()
>psi0 = (basis(2,0) - basis(2,1)).unit() 
>L = liouvillian(H, [])
>es = ode2es(L, psi0)

To fix we need to change the checks in ode2es to be
>if abs(rho0.full()).sum() < 1e-10 + 1e-24:

which then checks the sum of the absolute values rather than the absolute value of the sum

PR #1095, opened 2019-10-08
fix error checking for null matrix in essolve

peterkirton:
Fix bug #1094


nathanshammah:
I'm disregarding the fails in tests on Travis CI as this is due to pre-existent conditions due to `smesolve`, not affected here.

PR #1098, opened 2019-10-09
Enhancement nonsquare superopreps

arnelg:
Convert between superop reps with non-square shape. Needed because quantum channels can have input and output spaces of different dimension.

Affected functions so far:
`vec2mat`, `_super_tofrom_choi`, `choi_to_kraus`, `kraus_to_choi`

Not sure how to deal with tensor product structures.


arnelg:
Some of the checks are failing, but this seems unrelated to the changes made in this pull request. Any thoughts?

nathanshammah:
Yes, the `smesolve` testit is failing elsewhere already and it's unrelated. Sorry could you explain me a bit the tensor issue and if this is merge-ready?

nathanshammah:
Cc @fminga.

arnelg:
@nathanshammah: Sure. An assumption had previously been hard-coded into various conversion functions between super-operator representations that the input and output Hilbert spaces for the superoperator have the same dimension. An example is `_super_tofrom_choi` which converts between Liouville and Choi matrix representations, for example the line:

``sqrt_shape = int(sqrt(data.shape[0]))``

This enhancement removes these assumptions and instead looks at the input and ouput dimensions of the `Qobj`s, which is stored in the `dims` attribute, to correctly convert in the case when the input and output dims are different. It's needed in all sorts of applications.

Ericgig:
Operator being square are assumed at many other places. Is it fine to only improve this part? If so, I am ok with merging as is.

arnelg:
> Operator being square are assumed at many other places. Is it fine to only improve this part? If so, I am ok with merging as is.

Ideally this assumption should be removed everywhere unless the function really do need the Qobj to be square for some fundamental reason. Unfortunately this assumption has been hardcoded in in many places for no good reason. I do not have a good overview over all the places this has been done. Fixing it in this one place does not break anything and at least allows this function to work generally for superoperators with different intput and output spaces. I don't see any disadvantage to fixing it on a case by case basis... It seems better than doing nothing.

fminga:
@arnelg What if we add a variable `shape` also to the `_super_tofrom_choi`, `choi_to_kraus`, and `kraus_to_choi` functions?
In that way,  if `shape=None` the function uses the old code, while if `shape="rectangular"` it implements the new version of the code.
Moreover, for the sake of completeness, I think this option should be implemented also in the Stinespring and $\chi$ representation.
Commenting the docstring on this issue could help the user, even if these are private functions.
I have been reviewing this with @nathanshammah.

arnelg:
> @arnelg What if we add a variable `shape` also to the `_super_tofrom_choi`, `choi_to_kraus`, and `kraus_to_choi` functions?
> In that way, if `shape=None` the function uses the old code, while if `shape="rectangular"` it implements the new version of the code.

I don't think this is a good idea. The "square" case is really just a special case of "rectangular", and there is no reason the user should need to specify that it is square or not through an additional argument. The code as submitted already correctly handles a square input (as before), so why create a separate code path? Also by sticking to one code path we ensure that if there is an error, it only needs to be fixed once.

> Moreover, for the sake of completeness, I think this option should be implemented also in the Stinespring and $\chi$ representation.
> Commenting the docstring on this issue could help the user, even if these are private functions.
> I have been reviewing this with @nathanshammah.

Yeah good point. I will have a look at this together with @joshcombes. (It is fine to accept the current pull request, and we can open another one for further improvements, however.)

joshcombes:
I'd be happy to help on the future improvements (e.g. chi and Stinespring) and integrating these changes with other parts of qutip. The more piecemeal we can make the changes the easier it will be to review.

nathanshammah:
Thank you for the details on the new updates. @fminga's suggestions seemed reasonable to me: personally, I would be somewhat reluctant to modify something that is so much at the heart of the library machinery (master equations, etc.) for what seems a very specific use. I admit that part of this is this resistance is that I cannot really foresee the implications of this change. Now the Qobj permute tests are failing for example. I also see your point for more general definitions. I would recommend that docstrings and tests be included and possibly even some dev notebook or tutorial, or even just a gist here.

nathanshammah:
I'm not sure the `Qobj permute` failing test is related, by the way.

arnelg:
> I'm not sure the `Qobj permute` failing test is related, by the way.

Yeah we actually figured this one out. It's a numerical round off error. `Qobj permute` uses `tidyup`, with the same numerical tolerance as is used when the quality of two `Qobj`s are compared in the test. Kind of lucky that it didn't fail before. We either have to change the test to use a larger tolerance in the comparison, or give an argument to `Qobj permute` to not use `tidyup`. This is not really related to this PR but I'll implement the change once we decide on the best solution.

Does anyone have thoughts on why `Qobj permute` uses `tidyup` by default in the first place? I don't know how `permute` actually implements the permutation, but naively I would have thought that a permutation could be done without any numerical error

nonhermitian:
I was probably being stupid. It does not need tidyup.

arnelg:
> Thank you for the details on the new updates. @fminga's suggestions seemed reasonable to me: personally, I would be somewhat reluctant to modify something that is so much at the heart of the library machinery (master equations, etc.) for what seems a very specific use.

OK let me think about this some more. Our plan is to not modify any core parts of the code, only superoperator representation conversion functions, and in such a way that passing a square superoperator gives exactly the same behavior as before. My argument against passing a "shape=rectangular" flag is that the information we need is already stored in the `Qobj` `shape` and `dims` attributes.  But I've removed the "merge ready" flag until we have thought it through some more.

nathanshammah:
We aim at delivering the next release in a couple of weeks. Maybe just adding tests to be sure nothing is broken here would help including this PR.

nathanshammah:
It is giving an error at 
```
ERROR: mesolve: super vs. density matrix as initial condition
File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/qutip-4.5.0.dev0+294d62c2-py3.7-linux-x86_64.egg/qutip/tests/test_mesolve.py", line 543, in testSuperJC
    fid = self.fidelitycheck(out1, out2, rho0vec)
  File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/qutip-4.5.0.dev0+294d62c2-py3.7-linux-x86_64.egg/qutip/tests/test_mesolve.py", line 475, in fidelitycheck
    rhot = vector_to_operator(E*rho0vec)
  File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/qutip-4.5.0.dev0+294d62c2-py3.7-linux-x86_64.egg/qutip/superoperator.py", line 280, in vector_to_operator
    q.data = sp_reshape(op.data.T, shape[::-1]).T
  File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/qutip-4.5.0.dev0+294d62c2-py3.7-linux-x86_64.egg/qutip/sparse.py", line 135, in sp_reshape
    return zcsr_reshape(A, shape[0], shape[1])
  File "qutip/cy/spconvert.pyx", line 254, in qutip.cy.spconvert.zcsr_reshape
Exception: Total size of array must be unchanged.
```
Also, can you do a small change so that we can check if with the new `pytest` on Travis CI this is still the case?

joshcombes:
The error is genuine.

I think I have narrowed down the problem. The issue is reshaping (specifically vector_to_operator ) when the vector represents a tensor product system.

```python
from qutip import *


psi0 = tensor(basis(4, 0), basis(2, 0))
rho0vec = operator_to_vector(ket2dm(psi0))

print(rho0vec)


vector_to_operator(rho0vec)
```

joshcombes:
So the last commit fixed the **Travis CI** problems.

Aside from writing test and documentation I think there is one issue left. It has to do with constructing non square operators. Consider

```python

from qutip import *

zero = basis(2,0) 
one = basis(2,1)

zero_log = tensor(zero, zero, zero)
one_log = tensor(one, one, one)

print(zero_log)

print(zero.dag())

zero_log * zero.dag()

```

The last line fails. 

```python
<ipython-input-69-9ea42be37847> in <module>
     11 print(zero.dag())
     12 
---> 13 zero_log * zero.dag()

~/qutip/qutip/qobj.py in __mul__(self, other)
    544 
    545             else:
--> 546                 raise TypeError("Incompatible Qobj shapes")
    547 
    548         elif isinstance(other, np.ndarray):

TypeError: Incompatible Qobj shapes
```

It seems like '*' should allow one to construct non square operators, otherwise they need to constructed in numpy and then turned into a Qobj. In the latter case the `dims` won't be correct as the tensor product structure is not understood.

Ericgig:
Presently the condition in `Qobj.__mul__` is `left.dims[1] == right.dims[0]`. So the tensor dims `[1,1,1] `is causing the problem. You could create a special rule for this case (`self.shape[1] ==1 and  other.shape[0] == 1`).

joshcombes:
Update.

### Summary 

This PR helps convert between superop reps with non-square shape. This is needed because quantum channels can have input and output spaces of different dimension.

It works between super, choi, kraus. It correctly handles tensor product structures and systems with different dimensions e.g. a qubit and a qutrit. 

It does not work for Chi rep or the Steinspring rep. In the Chi rep only the qubit Pauli basis has been implemented. One would need to first get the Chi rep working for Qudits.  At the moment it is unclear how to generalize to the Stinespring rep. 

**changes**
- in `type_from_dims` the flag `enforce_square` was set to `False`. This helps in the conversion.
- in `qobj`,  `__mul__` was extended to allow construction of non square operators from an outer product using the suggestion of @Ericgig 
- `_super_tofrom_choi`, `choi_to_kraus`, `kraus_to_choi` have been changed so they work with non-square shapes.
- to support the above changes `vector_to_operator` and `vec2mat` were changed to allow for non square shapes

**other changes**
- `Qobj permute`. `tidyup` was removed from  to resolve a failing test at the suggestion of @nonhermitian 
- `choi_to_kraus` and `to_kraus`. When converting from Choi to to Kraus there are many small norm Kraus operators. A `tol` parameter was added to help remove these spurious operators.


### Added Tests
**test_superoperator_reps**
- Convert non-square Kraus operator to Super + Choi matrix and back.
- Neglect tiny Kraus operators when converting Choi to Kraus.

**test_superoper**
- Operator - vector - operator conversion with a tensor product state.
- Operator - vector - operator conversion for non-square matrix

**test_qobj**
- bra and ket multiplication with different dims e.g. [N, 1] and [1, M] for M \neq N

joshcombes:
Update Travis build passes.

### Summary 

This PR helps convert between superop reps with non-square shape. This is needed because quantum channels can have input and output spaces of different dimension.

It works between super, choi, kraus. It correctly handles tensor product structures and systems with different dimensions e.g. a qubit and a qutrit. 

It does not work for Chi rep or the Steinspring rep. In the Chi rep only the qubit Pauli basis has been implemented. One would need to first get the Chi rep working for Qudits.  At the moment it is unclear how to generalize to the Stinespring rep. 

**changes**
- in `type_from_dims` the flag `enforce_square` was set to `False`. This helps in the conversion.
- in `qobj`,  `__mul__` was extended to allow construction of non square operators from an outer product using the suggestion of @Ericgig 
- `_super_tofrom_choi`, `choi_to_kraus`, `kraus_to_choi` have been changed so they work with non-square shapes.
- to support the above changes `vector_to_operator` and `vec2mat` were changed to allow for non square shapes

**other changes**
- `Qobj permute`. `tidyup` was removed from  to resolve a failing test at the suggestion of @nonhermitian 
- `choi_to_kraus` and `to_kraus`. When converting from Choi to to Kraus there are many small norm Kraus operators. A `tol` parameter was added to help remove these spurious operators.


### Added Tests
**test_superoperator_reps**
- Convert non-square Kraus operator to Super + Choi matrix and back.
- Neglect tiny Kraus operators when converting Choi to Kraus.

**test_superoper**
- Operator - vector - operator conversion with a tensor product state.
- Operator - vector - operator conversion for non-square matrix

**test_qobj**
- bra and ket multiplication with different dims e.g. [N, 1] and [1, M] for M \neq N

Ericgig:
There is a small pep8 typo to fix. Otherwise seems good to me.

joshcombes:
> There is a small pep8 typo to fix. Otherwise seems good to me.

Thanks, it should be fixed now.

Issue #1077, opened 2019-08-26
Issue with steadystate on windows.


HiggsBoson3310:
Hello,

I'm using the following system and versions:
```
QuTiP: Quantum Toolbox in Python
Copyright (c) QuTiP team 2011 and later.
Original developers: R. J. Johansson & P. D. Nation.
Current admin team: Alexander Pitchford, Paul D. Nation, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, and Eric Giguère.
Project Manager: Franco Nori.
Currently developed through wide collaboration. See https://github.com/qutip for details.
QuTiP Version:      4.4.1
Numpy Version:      1.16.4
Scipy Version:      1.2.1
Cython Version:     0.29.12
Matplotlib Version: 3.1.0
Python Version:     3.7.3
Number of CPUs:     4
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True
Platform Info:      Windows (AMD64)
Installation path:  C:\Users\migue\Anaconda\lib\site-packages\qutip
==============================================================================
Please cite QuTiP in your publication.
==============================================================================
For your convenience a bibtex reference can be easily generated using `qutip.cite()` 
```

and I get this error repeatedly when using the `steadystate` function. 

```
OSError                                   Traceback (most recent call last)
<ipython-input-3-4ef4797a881f> in <module>
     30 g2_v = []
     31 for p, x in enumerate(Gamma_vec):
---> 32     j, z = calculate_avg_photons(N, x)
     33     print(p)
     34     n_avg.append(j)

<ipython-input-3-4ef4797a881f> in calculate_avg_photons(N, Gamma)
     18         c_ops.append(np.sqrt(rate)*sm.dag())
     19 
---> 20     rho = steadystate(H,c_ops)
     21     n_cav = expect(a.dag()*a,rho)
     22     g2_cav = expect(a.dag()*a.dag()*a*a,rho)/(n_cav**2)

~\Anaconda\lib\site-packages\qutip\steadystate.py in steadystate(A, c_op_list, method, solver, **kwargs)
    288         if (ss_args['solver'] == 'scipy' and ss_args['sparse']) \
    289                 or ss_args['solver'] == 'mkl':
--> 290             return _steadystate_direct_sparse(A, ss_args)
    291         else:
    292             return _steadystate_direct_dense(A, ss_args)

[...]

Issue #1017, opened 2019-06-12
Return info error with steadystate.py and scipy solver


JetSethi:
We were running the "Steady State Solvers: Optomechanical System" from QuTiP examples, but were repeatedly getting errors. Specifically, line 466 of `steadystate.py` tries to query `max_iter_refine` from `ss_args`, but this line is also executed when using a scipy solver, causing a `KeyError`. 

We were using scipy specifically so that `use_precond` is a valid option.

Ericgig:
Hi the `steadystate` has some problem with optional keyword and scipy/mkl solver.
To have the notebook work, if you have mkl installed, you can change the cell calling `steadystate` to:
```
solvers = ['direct','eigen','power','iterative-gmres','iterative-bicgstab']
mech_dms = []

for ss in solvers:
    if ss in ['iterative-gmres','iterative-bicgstab']:
        rho_ss, info = steadystate(H, c_ops, method=ss, use_precond=True, 
                    use_rcm=True, tol=1e-15, return_info=True)
    else:
        rho_ss, info = steadystate(H, c_ops, method=ss,
                    use_rcm=False, return_info=True)
    print(ss,'solution time =',info['solution_time'])
    rho_mech = ptrace(rho_ss, 1)
    mech_dms.append(rho_mech)
mech_dms = np.asarray(mech_dms)
```
Without mkl installed, skip the 'direct' method.


Help Wanted: Some rework of the keywords_args treatment, testing solver/method mixes ('direct' + 'scipy' should work but don't) updating notebooks and docstring, etc. would be appreciated.

tehruhn:
Hi @Ericgig , could you be a little more specific about what reworking the kwargs treatment would entail? Also if you could emphasize more concretely what you're looking for in updates to the notebooks and docs, that would be great.

I do have some other observations.
1. I do not get the KeyError message if I run the same jupyter notebook locally with the same versions of code as specified by version_table( ). I get another error instead, which actually seems more reasonable to me (something along the lines of 'use_precond' wasn't found in the keys of the dict, which clearly indicates that there's something off with the optional args thing).
2. Also using mech_dms = np.asarray(mech_dms) fails, I think because it sees a Qobj instead of a real number.
3. The direct patch to the KeyError mentioned above, based on my understanding of the code, would simply be to nest the assignment to 'max_iter_refine' in the 'info' dictionary inside a condition that checks if the solver is 'scipy'?

Ericgig:
Hi @tehruhn.
[...]

PR #1100, opened 2019-10-10
Fixes issue 1077, removes separate args for mkl and scipy

tehruhn:
Unified the args list for scipy and mkl solvers, and fixed issue 1077


tehruhn:
So Travis CI says there's an error in smesolve, which I haven't touched. :/

nathanshammah:
There is a bunch of .xml and .iml files that keep being tracked in your PR, I expect this is not part of the refactoring but just due to your local environment? You can fix it setting up a specific gitignore in your workspace I think.

nathanshammah:
I am referencing #1077 explicitly here.

nathanshammah:
This addresses also #1017 if I understand correctly. Looks good to me but would prefer to know @Ericgig's opinion on this.

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.so
 .DS_Store
 .f2py_f2cmap
+.idea/
```
Ericgig: Why is this needed?

qutip/steadystate.py
```diff
@@ -255,12 +243,13 @@ def steadystate(A, c_op_list=[], method='direct', solver=None, **kwargs):
...
-    elif solver == 'mkl':
-        ss_args = _mkl_steadystate_args()
-    else:
-        raise Exception('Invalid solver keyword argument.')
+    # if solver == 'scipy':
+    #     ss_args = _default_steadystate_args()
+    # elif solver == 'mkl':
+    #     ss_args = _mkl_steadystate_args()
+    # else:
+    #     raise Exception('Invalid solver keyword argument.')
```
Ericgig: Remove these comments before merging.

qutip/steadystate.py
```diff
@@ -255,12 +243,13 @@ def steadystate(A, c_op_list=[], method='direct', solver=None, **kwargs):
...
+    ss_args = _default_steadystate_args()
     ss_args['method'] = method
```
Ericgig: The `solver` value need to be set to `mkl` or `scipy` as needed. It is always `scipy` now.

.gitignore
tehruhn: Because I use PyCharm for its debugger, and PyCharm automatically creates the .idea/ folder in my workspace. I thought I'd add it there in case someone else did the same, and to avoid having to manually not commit the folder in my workspace.

qutip/steadystate.py
tehruhn: Done in latest commit.

qutip/steadystate.py
tehruhn: I assumed that it would be up to the user to specify if scipy or mkl was needed? I set scipy as default because I thought it was the most stable solver. Could you tell me what specifically I should set the default as, or what changes can be made?

Ericgig: Yes but it is in the variable `solver`, not in `kwargs` so it will not be updated in `ss_args`. Also it should be done before the line `ss_args['info']['solver'] = ss_args['solver']`. Or the info will always says `sicpy` even if mkl was used.

tehruhn: Aah, I see the issue. Apologies for the oversight.

tehruhn: I've fixed it in the latest commit, please check again. And thanks!

Issue #1103, opened 2019-10-12
Add resolution parameter when saving a Bloch sphere image


Titanosaurus:
Hi there! You people have done an incredible work with Qutip, but I still think there are many small settings that could be added probably without much effort;

When saving a Bloch sphere
`
from qutip import *

b = Bloch()
b.save(dirc = 'temp')
`

It would be awesome to have a way to specify the resolution of the png image just like in pyplot. Something like this

`b.save(dirc = 'temp', dpi = 600)`

Not having this option currently forbids these images to be added to conference posters or HD animations since the resolution is low.

Is there any workaround for this? I really need to know how to do it if there's a way.

tehruhn:
@nathanshammah I'm looking into this right now.

tehruhn:
HI @Titanosaurus , have submitted a PR, you will have your fix when it gets merged. :)

Usage would be as follows:
```b = Bloch()
    b.save("test", dpsi=500)
```

MichelangeloPantaleoni:
Thank you so much. You are quick!

Titanosaurus:
Wow, thank you for the quick response!

PR #1104, opened 2019-10-12
fixes issue #1103, adds dpsi arg to Bloch.save()

tehruhn:
added an argument to take dpsi (dots per square inch) in Bloch.save(), whose usage would be as follows :
``` b = Bloch()
     b.save("test", dpsi = 1000)
```
Only modified bloch.py, hence merge ready


nathanshammah:
There is a `.idea/workspace.xml` file that got picked up. I imagine that's specific of your environment, you would need to add it to the gitignore.

nathanshammah:
The dpsi option worked fine on my mac from jupyter. 
- please update the docstring to highlight this option (could also add the presence of the `name` option, besides directory and format)
- It is a nice feature, it could be highlighted in the documentation,  here https://github.com/qutip/qutip-doc/blob/master/guide/guide-bloch.rst
- add to 4.4.1 enhancements in the changelog, https://github.com/qutip/qutip-doc/blob/master/changelog.rst
Let's skip the codeclimate  issues for the moment, but in the future it might be nice to refactor the Bloch class inner workings, as it took me a while to sort out issue #1027.

nathanshammah:
Thanks @tehruhn for swiftly addressing this issue opened by @Titanosaurus. Please remember update the change log and .rst documentation file, when you manage. (It may be nice to find a way to remind contributors to update the Change log for the new version release, or understand how this can be partly done automatically, as in qiksit).

PR #1115, opened 2019-10-28
FIX: sesolve average states

ajgpitch:
When using the `average_states` solver option `general_stochastic` seemed to error when summing the timeslot states. This PR provides a fix for this.

Also included is some hopefully more succinct messages from the safety checks in `general_stochastic`.

### DIFF
--- a/qutip/stochastic.py
+++ b/qutip/stochastic.py
@@ -1038,6 +1038,7 @@ def general_stochastic(state0, times, d1, d2, e_ops=[], m_ops=[],
     Solve stochastic general equation. Dispatch to specific solvers
     depending on the value of the `solver` keyword argument.
 
+
     Parameters
     ----------
 
@@ -1090,8 +1091,8 @@ def general_stochastic(state0, times, d1, d2, e_ops=[], m_ops=[],
     sso = StochasticSolverOptions(False, H=None, state0=state0, times=times,
                                   e_ops=e_ops, args=args, **kwargs)
     if sso.solver_code not in [50, 100, 150]:
-        raise Exception("Only Euler, platen, platen15 can be " +
-                        "used for the general stochastic solver")
+        raise ValueError("Only Euler, platen, platen15 can be " +
+                         "used for the general stochastic solver.")
 
     sso.d1 = d1
     sso.d2 = d2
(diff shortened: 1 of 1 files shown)


ajgpitch:
I also added a few more safety checks. These may be helpful for anyone trying to convert some old example to v4.4 compatible.

ajgpitch:
@Ericgig as a side comment: is there some reason why we expect user to provide dm states vectorized already and also e_ops as superops. Could we not do the conversion for them?

ajgpitch:
I also switched the `general_stochastic` parameter checking exceptions type from `Exception` to:
`TypeError` for incorrect object type
`ValueError` for some incorrect property of object
`RuntimeError` for unknown source of error

I think this is good practice.

Ericgig:
`general_stochastic` is very manual, halfway to scipy's ode solver. It can be used for both open and closed system and does not correspond to any particular physics equation. It the function that you use if you do weird stuff that does not fit already available template and I would discourage it's use. 
If other type of stochastic evolution are commonly used, it would be better to make a new specialized function than use `general_stochastic`.

Old example should be adapted to `smesolve` as much as possible.

The solver does not know if it is solving an close or open system (or something else). So e_ops are not touched. 
d1, d2 are working in array for performance issue. We could easily have them work in Qobj, but I would expect a significant slowdown. These function, taking no args, and being call many time at each step could benefit from being compiled with numba.

ajgpitch:
Thanks @Ericgig for feedback.
I see (having read more carefully) in
https://github.com/qutip/qutip-notebooks/blob/master/examples/smesolve-inefficient-detection.ipynb
there is an example showing the equivalence of using `smesolve` and `general_stochastic` with a clear performance enhancement in the former. So I think I can easily switch my example to use `smesolve`. Thanks.

ajgpitch:
The `general_stochastic` function does process the `e_ops`. 
I see your point about it being it being a general solver of SDEs and so it makes sense that the variable is a vector. Certainly no value in changing the processing to work on Qobj.
I just think it would make usage easier if `state0` and `e_ops` (and I guess `m_ops`) could be provided as dm and opers rather than vectorised and superoperators. These could be recognised as such and converted before processing. We could then convert the `result.states` accordingly. 

Anyway, I am only just beginning to understand this stuff, so I will leave it up to your judgement as to whether this is worth the effort. If the aim to try always to avoid `general_stochastic`, then I suppose it is not.

ajgpitch:
Anyway, I think the changes proposed in this PR are needed to make `average_states` work. @Ericgig Do you agree.

ajgpitch:
re latest commit: I have tested some but not all of the StochasticSolverOptions.__init__() checking exceptions.

ajgpitch:
@Ericgig if you are happy with the way I have changed the parameter checking exception raising, then I will change the rest of them in the `stochastic.py` to match as part of this PR, as I think that makes sense. 

Some are not just cosmetic. In particular the ones where I added `raise .... from e` were very helpful me when working on my current example (which I why I added them of course :)

ajgpitch:
`traj_states` property added to result object. This is just `states` unless `average_states` is used. When `average_states` is used then it is the original trajectory states. Sometimes it can be useful to access the original trajectory states. 

`avg_states` property added to result object. This is the same as `states` when `average_states` is used, otherwise it is `None`.

PR #1123, opened 2019-10-31
Qobjevofunc

Ericgig:
Add a wrapper for function returning `Qobj` to be treated as `QobjEvo`: `QobjEvoFunc`
Wrapped function have access to all method of `QobjEvo` therefore only one version of the code should be sufficient for time dependant problems using both list format and function.
Any callable is considered a function, allowing for class method etc.

The new function `qobjevo_maker` can take any valid time dependant quantum object and make it a `QobjEvo` like object. Added wrapper for function using the old `rhs_with_state` format and using no `args`. 

Tests for `QobjEvoFunc` have been added but not for `qobjevo_maker`.

Solver are not updated to use it yet.

@ajgpitch I added it last minute in stochastic, but did not tried it yet.


nathanshammah:
Shall we disregard the no-cython build fail and merge?

Ericgig:
It's not ready for 4.5. I will add other PR that use the feature. Not useful by itself.

PR #1125, opened 2019-11-07
Update tests

Ericgig:
Add a linux python 3.8 test
Fix MacOS python version to 3.7 as our test environment result in conflict on 3.8
* Does not fix mac tests' errors


Ericgig:
I got a hard crash (segfault?) on macOS 10.14 but not on  macOS 10.13: xcode10 vs xcode11.2
We should wait a bit more for tests python 3.8

Issue #1120, opened 2019-10-31
"malloc: Incorrect checksum" error in qutip.testing qt.run()


SachaRG:
After installing qutip via conda (conda install qutip) on Mac 10.15.1 High Sierra, I ran the testing script using:

>> import qutip.testing as qt 
>> qt.run()

I am getting the following error: 

Qobj subtraction ... python(4259,0x119a26dc0) malloc: Incorrect checksum for freed object 0x7fdbb05c5910: probably modified after being freed.
Corrupt value: 0x3ff0000000000000
python(4259,0x119a26dc0) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6

Does anyone know what is causing this, how to fix it, or what kinds of problems it might cause later? Thank you!

Sacha

amcdawes:
Can you post the output of qutip.about()

amcdawes:
I see this issue as well but with a different test (error came up with ptrace test). I think this is a larger py3.7 issue, some comments have come up in the mailing list too.

In a py3.7 env I see malloc errors occasionally with the es solver in correlation_2op_1t calls as used here: https://github.com/gsteele13/gary-exploring-qutip/blob/master/Correlations.ipynb

QuTiP Version:      4.4.1
Numpy Version:      1.17.2
Scipy Version:      1.3.1
Cython Version:     0.29.13
Matplotlib Version: 3.1.1
Python Version:     3.7.4
Number of CPUs:     2
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True
Platform Info:      Darwin (x86_64)

amcdawes:
This also appears to be inconsistent in terms of when the issue comes up. I tried a quick use of `tracemalloc` to begin poking around, and that seems to change the occurrence too (like a true quantum system, measuring it disturbs it :)

<img width="858" alt="Screen Shot 2019-11-08 at 10 13 00 AM" src="https://user-images.githubusercontent.com/406657/68500451-9aea4c80-0210-11ea-8c86-0023b71613a7.png">

`Malloc.py` listed below:

```
from qutip import *
import numpy as np

N = 5
w = 1 * 2 * np.pi              # 1 Hz oscillator frequency
Q = 40
kappa = w / Q

a = destroy(N)
n = num(N)
x = (a + a.dag())/np.sqrt(2)
p = -1j * (a - a.dag())/np.sqrt(2)
H = w * a.dag() * a

c_ops = []

[...]

PR #1126, opened 2019-11-08
explicitly set pointer to NULL in spconvert

amcdawes:
This is a real shot in the dark, but it appears to have fixed an issue with lots of random malloc errors popping up on the mac platform (see #1120 and other recent discussions on the google group. It appears this is literally the only call to `free()` in all of qutip so perhaps there is a better way to do low-level memory work? I only know enough C to be dangerous, and found a random hint about setting pointers to NULL after freeing them. I gave it a shot and no longer see the issue on my mac when running py2.7.


amcdawes:
Sorry, cart before the horse. I should have checked that the latest code from github still had this issue... I'll roll back my changes and test against py37

PR #1131, opened 2019-12-02
Migrating from nose to pytest

tehruhn:
Updated travis.yml to move to pytest. 
Also updated qutip.testing to use pytest.


nathanshammah:
Referencing #972 and #958.

tehruhn:
So, in travis, only test_control_pulseoptim.py fails, but for all containers.

nathanshammah:
Great for the PR. 

Somehow `nose` was not picking this error. 
It would be nice to investigate and fix. 
The problem seems to be at test 03 there,
```
AttributeError: 'TestPulseOptim' object has no attribute 'tmp_dirs'
```
Indeed, TestPulseOptim seems to have this attribute where `tmp_dirs`
are defined, 
```
    def setUp(self):
        # list of file paths to be removed after test
        self.tmp_files = []
        # list of folder paths to be removed after test
        self.tmp_dirs = []
```
Investigating this locally (running the pytest only for this module) would be great.

nathanshammah:
Something is wrong in setting up the environment.

tehruhn:
The problem was that I was using a pytest plugin to output the docstrings, which are not available in conda. Now I just increased verbosity to 2.

nathanshammah:
All builds  pass, besides the build on Python 3.7 without Cython. I would still merge this PR and focus on that issue separately. 

Now we need to update places where `nose` is mentioned, I found these in `qutip-doc`
- the [changelog](https://github.com/qutip/qutip-doc), 
- the instructions on the dependencies in the documentation [installation instructions](https://github.com/qutip/qutip-doc/blob/402ec10de7a314ba9f124ddac7514621e92496df/installation.rst), 
- and [development guidelines](https://github.com/qutip/qutip-doc/blob/2ffd6d1cc581e858674d43fe97a9297e9476a278/qutip_dev_contrib.md).

tehruhn:
Now only tests which fail are on the "Python 3.7 no Cython container". Everything else passes. Looking into these right now.

tehruhn:
Latest commit changes the code in qutip.testing to run pytest instead of nose.

qutip/tests/cqobjevo_compiled_coeff_827871948660708379.pyx
```diff
@@ -0,0 +1,41 @@
+#!python
```
Ericgig: These files should not be included, could you add it to gitignore.
Without cython,".py" temp files will be created instead, but the "compiled_coeff" part is constant.

tehruhn: Done, thanks for pointing out.

tehruhn:
The first time these tests failed was after I removed the compiled_coeff file, they always passed in previous travis builds. I would like to roll back the last commit and see if it works.

tehruhn:
Sorry for this mess of reverts.

tehruhn:
Hi @Ericgig , 
Could you please take a look at this? Me and @nathanshammah  were wondering if this should be merged as-is, but a new issue should be opened to reflect that the Python no Cython tests fail? Or maybe we could fix them, and could you help with what's going wrong?

Ericgig:
Is the code failing or only the tests?
Does it fail in local test or only in automated tests?
The file that I asked you to add to gitignore was a temp file created in the tests directory, which is not where they should be. If pytest run the tests from that directory, this may be the cause of the error. Otherwise I have no idea.

tehruhn:
@nathanshammah I think @Ericgig is right because it doesn't fail tests locally. Should this be merged for now? If it causes problems we can create new issues later.

nathanshammah:
@tehruhn let's look into this together this week so that I can understand better also Eric comments. I would be happy to merge, and open a separate issue on the np-Cython bit. nose to pytest migration has been dragging on for ages. Also, please let's try to find a way to have the docstrings print to screen, the verbosity=2 thing does not seem to work. With `nose` and current development master now I get a segmentation fault on Mac 10.14 so I cannot test locally.

nathanshammah:
locally on my machine, also with pytest, it gives a segmentation fault at brmesolve

nathanshammah:
Also in the conda distribution. I am on gcc 4.8.5

Ericgig:
Could you make 2 changes:
`verbosity=1`, it looks a lot cleaner.
The tests that "XFAIL" use `yield` and are skipped as it is not supported by pytest. Could you rewrite them without it. (4 in `test_qobj`, not sure if there are more.) I will be looking at the failing tests in more detail next week.

tehruhn:
@Ericgig I was trying to use pytest's paramterize feature which allows us to parametrize test functions variable by variable. I was able to successfully rewrite some tests which use yield, but this one (`test_trunc_neg`) in particular throws an error. Could you please run this locally and tell me if there is any mistake I'm making? Thanks for the tip about yield, that might indeed be what is wrong.

```
from qutip import *
import numpy as np
import pytest
from numpy.testing import (assert_equal, assert_, assert_almost_equal,
                            run_module_suite)

@pytest.mark.parametrize("method", ['clip', 'sgs'])
@pytest.mark.parametrize("qobj", [rand_dm(5), Qobj(np.diag([1.1, -0.1])),
    rand_unitary(3)*Qobj(np.diag([1.1, 0, -0.1]))*rand_unitary(3).dag()])
@pytest.mark.parametrize("expected", [None, Qobj(np.diag([1.0, 0.0])),
    rand_unitary(3)*Qobj(np.diag([1.0, 0.0, 0.0]))*rand_unitary(3).dag()])
def test_trunc_neg(method, qobj, expected):
    "Checks trunc_neg in several different cases."
    pos_qobj = qobj.trunc_neg(method=method)
    assert(all([energy > -1e-8 for energy in pos_qobj.eigenenergies()]))
    assert_almost_equal(pos_qobj.tr(), 1)
    if expected is not None:
        assert_almost_equal(pos_qobj.data.todense(), expected.data.todense())
```

Ericgig:
I believe that all permutation are run with `parametrize`. So the input (`qobj`) is compared to each outputs `(expected`), not just the matching one. Also the rotation matrix (`U = rand_unitary(3)`) must be constant. Here, at least 4 different random operators are created.

Instead of `parametrize` and `yield` simply calling the function `case` would work:
```
yield case("Test Qobj: trunc_neg works for positive opers."), \
            rand_dm(5), method
```
to
```
case(rand_dm(5), method, None)
```
To keep the message,  `assert_almost_equal` has a `err_msg` keyword to raise if the test fails.

I do not believe the `yield` is the cause of the no cython fails.

ps. I will make a branch of this PR and will add `mark` to prepare a subset of tests for faster testing #1147.

tehruhn:
@nathanshammah @Ericgig all tests seem to pass. Please review. Also thanks for your fix @Ericgig.

nathanshammah:
This looks great. Any clue on why the `verbosity=1` is not showing the docstrings? Is it working fine locally?

tehruhn:
Can confirm that on running qutip.run( ) locally, it prints a one line description of tests.

Ericgig:
I get the not the docstring but the function name. `verbosity=2` also give the test file full path, but not the docstring.

Ericgig:
I found this discussion on pytest: https://github.com/pytest-dev/pytest/issues/445 . It's old but seems it's not an option they support. I would be happy to merge this as is.

PR #1139, opened 2019-12-06
QASM v1 support added

abhisekupadhyaya:
Class Function for importing QASM v1.0 files added. Supports loops and quantum gates. There is no support for the initialization of qubits, measurement, averaging and classical control of the circuits yet. These parts are ignored while reading the .qasm file

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -1147,3 +1147,174 @@ class QubitCircuit(object):
             code += "\n"
 
         return code
+        
+    @classmethod
+    def read_qasm(cls, file_path):
+        """
+        Import a QSAM 1.0 file and return an equivalent QuTiP quantum
+        circuit.
+
+        Parameters
+        ----------
+        file_path : string
+            Path to the file.
+        """
+
+        # Defining the gates
+        version = ["version"]
+        qubit_registary = ["qubits"]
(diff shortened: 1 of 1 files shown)


BoxiLi:
Hi, @abhisekupadhyaya nice work! 

Usually a high cognitive complexity indicates that the function is too long or too deep for understanding and maintenance, but sometimes it exaggerates a bit, so it's fine as long as everything is clear. To reduce the function length, for example, if the checking part and each gate parsing can be resued (e.g for QASM v2), they can be defined in a separate function.

Another minor thing, what about leaving a short warning when ignoring the unsupported code?

abhisekupadhyaya:
> Hi, @abhisekupadhyaya nice work!
> 
> Usually a high cognitive complexity indicates that the function is too long or too deep for understanding and maintenance, but sometimes it exaggerates a bit, so it's fine as long as everything is clear. To reduce the function length, for example, if the checking part and each gate parsing can be resued (e.g for QASM v2), they can be defined in a separate function.
> 
> Another minor thing, what about leaving a short warning when ignoring the unsupported code?

abhisekupadhyaya:
@BoxiLi : I will introduce the warnings. I think it will be better to work on developing classical control and measurement tools for QuTiP in order to fully support QASM v1 before moving on to QASM v2. Let me know more suggestions.

PR #1140, opened 2019-12-07
QASM v1 support added

abhisekupadhyaya:
Same pull request as #1139 . Reopening it after it was accidentally closed.

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -1147,35 +1147,38 @@ class QubitCircuit(object):
             code += "\n"
 
         return code
-        
+
     @classmethod
     def read_qasm(cls, file_path):
         """
         Import a QSAM 1.0 file and return an equivalent QuTiP quantum
-        circuit.
+        circuit object of the class QubitCircuit.
 
         Parameters
         ----------
         file_path : string
             Path to the file.
-        """
 
+        Returns
+        -------
+        qc : QubitCircuit
+            Return equivalent qubit circuit object.
+        """
         # Defining the gates
-        version = ["version"]
         qubit_registary = ["qubits"]
-        static_gate = {'h':'SNOT', 'x':rx(np.pi), 'y':ry(np.pi), 'z':rz(np.pi),
-                       'x90':rx(np.pi/2), 'y90':ry(np.pi/2), 'z90':rz(np.pi/2),
-                                         'mx90':rx(-np.pi/2), 'my90':ry(-np.pi/2), 
-                                         'mz90':rz(-np.pi/2),'s':phasegate(np.pi/2),
-                                         'sdag':phasegate(-np.pi/2), 't':phasegate(np.pi/4),
-                                         'tdag':phasegate(-np.pi/4)}
-        param_gate = {'rx':'RX', 'ry':'RY', 'rz':'RZ'}
-        c_gate = {'cnot':'CNOT', 'cz':'CSIGN'}
-        crk = {'crk':'CPHASE'}
-        cr= {'cr':'CPHASE'}
-        toffoli_gate = {'toffoli':'TOFFOLI'}
+        static_gate = {'h': 'SNOT', 'x': rx(np.pi), 'y': ry(np.pi),
+                       'z': rz(np.pi), 'x90': rx(np.pi/2), 'y90': ry(np.pi/2),
+                       'z90': rz(np.pi/2), 'mx90': rx(-np.pi/2),
+                       'my90': ry(-np.pi/2), 'mz90': rz(-np.pi/2),
+                       's': phasegate(np.pi/2), 'sdag': phasegate(-np.pi/2),
+                       't': phasegate(np.pi/4), 'tdag': phasegate(-np.pi/4)}
+        param_gate = {'rx': 'RX', 'ry': 'RY', 'rz': 'RZ'}
+        c_gate = {'cnot': 'CNOT', 'cz': 'CSIGN'}
+        crk = {'crk': 'CPHASE'}
+        cr = {'cr': 'CPHASE'}
+        toffoli_gate = {'toffoli': 'TOFFOLI'}
         swap_gate = {'swap': 'SWAP'}
-        
+
         gates_keys = []
         gates_keys.extend(static_gate.keys())
         gates_keys.extend(param_gate.keys())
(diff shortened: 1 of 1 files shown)


qutip/qip/circuit.py
```diff
@@ -1147,3 +1147,174 @@ def qasm(self):
             code += "\n"
 
         return code
+        
+    @classmethod
+    def read_qasm(cls, file_path):
+        """
+        Import a QSAM 1.0 file and return an equivalent QuTiP quantum
```
nathanshammah: QSAM-->QASM. Specify the class of the Quantum Object.

qutip/qip/circuit.py
```diff
@@ -1147,3 +1147,174 @@ def qasm(self):
...
+        circuit.
+
+        Parameters
+        ----------
+        file_path : string
+            Path to the file.
+        """
+
```
nathanshammah: No space needed

qutip/qip/circuit.py
```diff
@@ -1147,3 +1147,174 @@ def qasm(self):
...
+                    lines.append(command)
+                    
+                    # Checking for the size of the qubit register.
+                    if command[0][0] in qubit_registary:
+                        N = int(command[1][0])
+                        qc = QubitCircuit(N)
+                        loop_time = 1
+                        qc_loop = QubitCircuit(N)
+
+                    # Checking for if the line is a function defination.
```
nathanshammah: definition

qutip/qip/circuit.py
```diff
@@ -1147,3 +1147,174 @@ def qasm(self):
...
+                                       int(bit_list[1][0])]
+                            
+                            qc_loop.add_gate(swap_gate[gate_name], 
+                                             controls=targets)
+        
+        # Adding the gate in the last function into the circuit.
+        for i in range(loop_time):
+            qc.add_circuit(qc_loop)
+        
+        return qc
```
nathanshammah: Add newline to end file

qutip/qip/circuit.py
```diff
@@ -1147,3 +1147,174 @@ def qasm(self):
...
+        
+    @classmethod
+    def read_qasm(cls, file_path):
+        """
+        Import a QSAM 1.0 file and return an equivalent QuTiP quantum
+        circuit.
+
+        Parameters
+        ----------
+        file_path : string
```
nathanshammah: Missing `cls` as parameter. Please add `Returns:` with the output.

nathanshammah:
Regarding this: "Also, more generally, I am thinking of where to add the project online: whether in the QuTiP Wiki, as a new issue, or as an issue in a new qutip-projects repository in the qutip organization." I see there is a projects tab on the right side of the menu, which maybe could be used.

nathanshammah:
I think it is better to keep this on hold, and merge it after the upcoming release, also as my comments have not been addressed.

abhisekupadhyaya:
> I think it is better to keep this on hold, and merge it after the upcoming release, also as my comments have not been addressed.

Sorry for the late reply. I think it would be better to keep it in the project tab. Code Climate Issues will be delt soon enough. I suppose it would be better if the advantage of #1090 is taken to support measurement in the qutip-circuit before moving on to Qiskit.

nathanshammah:
I was checking the status of this PR, sorry for the delay, @abhisekupadhyaya. 

1) **Address comments**: There's been some updates and CI tests pass. However, my comments are still not addressed, e.g., updates in docstrings, see above. Addressing this before the next months is crucial as then there may be quite some activity on the `qip` module. 


2) **Add tests**:  We also need to add at least a couple of tests with some examples circuits. The tests could check:
2.a) that the circuit does the same thing
2.b) that the circuit is exactly the same (no change of gates into other gates)

2) **Qasm export**: Adding an export feature, e.g. `to_qasm`, would be great. But this PR could address just the import and be fine (changing name would be best). If the `to_qasm` feature was here, one could simply test it one way and return, and check that, e.g.
 ```
read_qasm(qutip_circuit1.to_qasm()) == qutip_circuit1
read_qasm(qasm_circuit2).to_qasm() ==qasm_circuit2
```
if this feature was there, one could also test randomly generated circuits. @abhisekupadhyaya do you think you can address 1) and 2)?

**Comment by nathanshammah** [comment]
So, the "to qasm" feature is already implemented, e.g.,
```python
>>> from qutip.qip.circuit import QubitCircuit

>>> qc = QubitCircuit(N=3)
>>> qc.add_gate("SNOT", targets=0)
>>> qc.add_gate("SNOT", targets=1)
>>> qc.add_gate("SNOT", targets=2)

>>> qasm_qc = qc.qasm()
>>> print(qasm_qc)

# qasm code generated by QuTiP

	qubit	q0
	qubit	q1
	qubit	q2

	SNOT	q0
	SNOT	q1
	SNOT	q2
```

nathanshammah:
This branch could merge the current master and possibly add to the qasm conversion also the recently  added controlled-S(Z, etc) gate, #1224, which in QASM may be `c-S` and `c-Z`, etc.

nathanshammah:
@abhisekupadhyaya, I cannot find where the input `cls` is used in the method.

qutip/qip/circuit.py
jakelishman: @nathanshammah `cls` is a pseudo-parameter, like `self`.  When you decorate a function with `@classmethod`, then you call it as (say) `QubitCircuit.read_qasm(file_path)`, and `cls` becomes the `type` `QubitCircuit` (as opposed to a specific instance).  It's like a static method in any other OOP language.

qutip/qip/circuit.py
```diff
@@ -1147,3 +1147,181 @@ def qasm(self):
...
+                    command = [x.replace(',', ' ') for x in command]
+                    command = [x.split() for x in command]
+                    command = [x for x in command if x]
+
+                    lines.append(command)
+
+                    # Checking for the size of the qubit register.
+                    if command[0][0] in qubit_registary:
+                        N = int(command[1][0])
+                        qc = QubitCircuit(N)
```
jakelishman, line 1215: If this is a class method, then calls to `QubitCircuit` should be via the parameter `cls` so that subclassed versions behave correctly.  Alternatively, you can remove this from the class and make it a regular function, since Python allows this, and subclassed versions may modify the `__init__` method.

nathanshammah: Thanks.

nathanshammah:
Closed by #1291.

PR #1142, opened 2019-12-09
long interger based clebsch

### LINKED ISSUE #1141: High Clebsch-Gordan coefficients give incorrect results because of numeric instability
The current version of `clebsch` gives wildly incorrect results for high `j`, because it relies on the precise cancellation of extremely large quantities. This leads to incorrect answers for anything that depends on this function, such as the spin Wigner function for Rydberg atoms.

Example of problem:

```
clebsch(15, 15, 30, 0, 0, 0) 
> 7.03237620084722 
clebsch(20, 20, 40, 0, 0, 0)
> 8986860.153685257
clebsch(30, 30, 60, 0, 0, 0)
> nan
```

Is it possible to implement one of the many higher-precision methods to compute Clebsch-Gordan coefficients? If that would be slower, is it possible to add a flag to toggle between the methods?

Ericgig:
Rewrote `clebsch` function based on long integer fraction. Thus escaping the rounding error mentioned in #1141. Surprisingly it's a little faster than the float implementation. 

Also added tests for the function.

Closes #1141

### DIFF
--- a/qutip/tests/test_utilities.py
+++ b/qutip/tests/test_utilities.py
@@ -34,7 +34,7 @@
 import numpy as np
 from numpy.testing import assert_, run_module_suite
 
-from qutip import convert_unit
+from qutip import convert_unit, clebsch
 
 
 def test_unit_conversions():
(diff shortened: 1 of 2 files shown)


nathanshammah:
Wonderful!

Issue #1147, opened 2019-12-30
Recommended subset of tests for faster testing


costrouc:
I am packaging qutip for [nixpkgs](https://github.com/NixOS/nixpkgs) https://github.com/NixOS/nixpkgs/pull/76661 from version 2.2.0 -> 4.4.1. Right now the only issue with the build is that qutip takes way too long to run the test suite in proportion to its usage in nixpkgs. Do you have a recommended set of tests that we should run? For nixpkgs I will likely just delete some of these files https://github.com/qutip/qutip/tree/master/qutip/tests. Currently the test suite takes around 2 hours and would like to get it to around 10 minutes or less.

nathanshammah:
This is issue has a two-fold interest: decreasing test time in general and expand the platforms through which QuTiP is available. 

What is the reason it takes so long on Nix? QuTiP, in the [latest PR](https://travis-ci.org/qutip/qutip/builds/628865624?utm_source=github_status&utm_medium=notification), takes on average 20 minutes to run its tests on a given platform. Much of this I think is due to the Cython configurations, indeed the no-cython option takes only 13 minutes. My first suggestion would be to try without Cython if things speed up, from those requirements (@Ericgig set that up and I am not knowledgeable about it). 

From your PR on `nixpkgs` do you build on Travis CI or a similar cloud? If you could give a list of time taken by the different tests that would help. 

The idea of a subset of tests is interesting (essential tests vs. thorough tests), in general; for example through tests could be done only in production and a lighter version could be called with `qutip.testing.run()`. Although, in specific, it is not clear to me how triaging which tests to skip.   I would be glad to have the others opinion on this too (pinged them as assignees).

Ericgig:
Testing without cython is faster but cython is needed to install qutip. Most tests are only useful to us. We should make a subset of tests to check installation without checking all options. We could have only 1 stochastic test instead of running all 8 solvers, unless you are working on this part of the code, they will all fail or pass together.

@costrouc running one test per file would be better than removing some file. Are you in an hurry? I could make it once I am back from vacation (unless someone else have time during new year)

costrouc:
@Ericgig not in any huge hurry. Once there is a nice subset of tests that can be run to test installation I will update the nixpkgs package for qutip. Thanks for being so responsive! For now I'll just delete some of the tests and can always update the package once the ability to test only a subset is added to a release.

PR #1151, opened 2020-01-07
Marked test

Ericgig:
Mark some of the slower test to be skipped when running.

`qutip.testing.run()`: about 5 min.
`qutip.testing.run(full=True)`: about 30 min.

Keep test for every function: stochastic only test the default solver... 
Also speed up some test by limiting the tests (shorter time for solver, less "pixel" for wigner).

Build on top of #1131, must be merged after that PR.
Should solve #1147


nathanshammah:
No cython tests are failing.

Ericgig:
The fix for the no cython should go in #1131, I sent the fix directly to @tehruhn :  tehruhn/qutip#1

nathanshammah:
Need to resolve conflicts here.

nathanshammah:
Can we add information about this feature in the documentation?

Ericgig:
I added the feature explanation in the docstring. I don't think we have special documentation for `testing.run`.

Issue #1152, opened 2020-01-07
qutip.propagator does not set dims for unitary_mode='single' with parallel=False


lukjp:
When using qutip.propagator with unitary_mode='single' and parallel=False, the result does not have the same dims as the Hamiltonian, for other setting it does work.
Minimal example (tested with qutip 4.4.1):
```
In [1]: import qutip                                                                                                                                                    

In [2]: H = qutip.tensor([qutip.qeye(2)]*2)                                                                                                                             

In [3]: H.dims                                                                                                                                                          
Out[3]: [[2, 2], [2, 2]]

In [4]: qutip.propagator(H, 1 , unitary_mode='single', parallel=True).dims                                                                                              
Out[4]: [[2, 2], [2, 2]]

In [5]: qutip.propagator(H, 1, unitary_mode='batch').dims                                                                                                               
Out[5]: [[2, 2], [2, 2]]

In [6]: qutip.propagator(H, 1 , unitary_mode='single').dims                                                                                                             
Out[6]: [[4], [4]]```

**Comment by nathanshammah** [comment]
It should be addressed by PR #1153.  
The problem seems to be in line 161 of `qutip.propagator`, [here](https://github.com/qutip/qutip/blob/89ddbf4d2c238cba880bb40acdc448fc1c7c951e/qutip/propagator.py#L161), 
```
            if unitary_mode == 'single':
                output = sesolve(H, qeye(N), tlist, [], args, options,
                                 _safe_mode=False)
```
the `qeye(N)` resets the dimensions of `H`, which up to then retains the correct Hilbert space structure, which should be `H.dims[0]`.

PR #1153, opened 2020-01-07
fix dims incongruence in `propagator`

nathanshammah:
Solves #1152, in which it is found that the `propagator` gives the wrong dimensions when it is run for `unitary_mode= single` and `parallel=False`. 

The problem was arising in line 161 of `qutip.propagator`, [here](https://github.com/qutip/qutip/blob/89ddbf4d2c238cba880bb40acdc448fc1c7c951e/qutip/propagator.py#L161), 
```
            if unitary_mode == 'single':
                output = sesolve(H, qeye(N), tlist, [], args, options,
                                 _safe_mode=False)
```
the `qeye(N)` resets the dimensions of `H`, which up to then retains the correct Hilbert space structure, `dims`.  Now they are set by `H.dims[0]` in the state.

### DIFF
--- a/qutip/propagator.py
+++ b/qutip/propagator.py
@@ -158,7 +158,7 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
                     u[:, n, k] = output[n].states[k].full().T
         else:
             if unitary_mode == 'single':
-                output = sesolve(H, qeye(H.dims[0]), tlist, [], args, options,
+                output = sesolve(H, qeye(dims[0]), tlist, [], args, options,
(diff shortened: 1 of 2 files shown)


lukjp:
I think you should use `dims[0]` as it is set to `H0.dims` here https://github.com/qutip/qutip/blob/89ddbf4d2c238cba880bb40acdc448fc1c7c951e/qutip/propagator.py#L148 instead of `H.dims[0]` as `H` may be a callback function or a list.

Ericgig:
Thank you for finding the bug.

PR #1155, opened 2020-01-11
Pulse-level control in qip device simulator

BoxiLi:
We have been talking about pulse level control in the `qip` device simulator for a while, but I just released that we didn't really define it in as separate "pulse". I find it really important and therefore I rewrite the data part of the `qip` numerical simulator as a new class `Pulse`.

It's quite interesting that I end up to define a class very similar to `EvoElement` in `Qobjevo` (I called it `_EvoElement` for internal use, cannot find a better name). Instead of the full Hamiltonian, I only save the target qubits label and the non-trivial Hamiltonian (e.g. `sigmaz()` instead of `tensor(sigmaz(), identity(2)`). It is defined in this way for easy inspection after initialized, such as adding noise.

`_EvoElement` consists of
- Hamlitonian
- Target qutbis label
- `tlist`
- `coeff`

`Pulse` consists of

- ideal part: `_Evoelement`
- coherent noise: list of `_Evoelement`
- lindblad noise: list of `_Evoelement`

The advantages are:
- A list of `Pulse `is easier to understand and manipulate compared to the old design: a list of H and a list of `coeff`. Now different `tlist `can also be defined for different pulses.
- Adding noise is much easier as we now can get the non-trivial part of the Hamiltonian and the targets of each pulse, instead of only a huge Hamiltonian in `QobjEvo`. One can now define noise for each Pulse.

Still working on the docs.


BoxiLi:
@nathanshammah @ajgpitch 

Hi Nathan and Alex, this PR is ready. Is it be possible to get it merged before 1st Feb? It would be great if this update is ready before the talk.

The main change I did was in
- processor.py
- pulse.py
- noise.py

The others are just updating, renaming and moving around.

qutip/qip/device/cqed.py
```diff
@@ -54,6 +56,8 @@ class DispersivecQED(ModelProcessor):
     For a given pulse amplitude matrix, the processor can
     calculate the state evolution under the given control pulse,
     either analytically or numerically.
+    (Only additional attributes are documented here, others please
```
nathanshammah: *for* others

qutip/qip/device/modelprocessor.py
```diff
@@ -55,6 +55,8 @@ class ModelProcessor(Processor):
     The processor can simulate the evolution under the given
     control pulses either numerically or analytically.
     It cannot be used alone, please refer to the sub-classes.
+    (Only additional attributes are documented here, others please
```
nathanshammah: for others

qutip/qip/device/modelprocessor.py
```diff
@@ -184,7 +143,7 @@ def run_state(self, rho0=None, analytical=False, qc=None, states=None,
             and then calculate the evolution.
 
         states: :class:`qutip.Qobj`, optional
-         Old API, same as rho0.
+         Old API, same as init_state.
```
nathanshammah: I understand this will be removed later on.

qutip/qip/device/processor.py
```diff
@@ -67,126 +68,275 @@ class Processor(object):
...
         self.noise = []
+        self.drift = Drift()
         if dims is None:
             self.dims = [2] * N
         else:
             self.dims = dims
         self.spline_kind = spline_kind
 
-    def add_ctrl(self, ctrl, targets=None, cyclic_permutation=False):
+    def add_drift_ham(self, ham, targets, cyclic_permutation=False):
```
nathanshammah: I would replace instances of `ham` with `hamiltonian`, or remove either `add` or `hamiltonian` to make it shorter

qutip/qip/device/processor.py
```diff
@@ -67,126 +68,275 @@ class Processor(object):
...
+        Parameters
+        ----------
+        tlist: array-like, optional
+            A list of time at which the time-dependent coefficients are
+            applied. See :class:`qutip.qip.Pulse` for detailed information`
+        """
+        for pulse in self.pulses:
+            pulse.tlist = tlist
+
+    def add_pulse(self, pulse):
```
nathanshammah: cool

qutip/qip/device/processor.py
```diff
@@ -308,107 +456,100 @@ def read_coeff(self, file_name, inctime=True):
...
-        return self.tlist, self.coeffs
+            return self.get_full_tlist, self.coeffs
 
-    def get_unitary_qobjevo(self, args=None):
+    def get_noisy_pulses(self, ind_device_noise=False, ind_drift=False):
         """
-        Create a :class:`qutip.QobjEvo` without any noise that can be given to
-        the QuTiP open system solver.
+        It takes the pulses defined in the `Processor` and
+        add noise according to `Processor.noise`. It does not modify the
```
nathanshammah: adds

qutip/qip/device/processor.py
```diff
@@ -308,107 +456,100 @@ def read_coeff(self, file_name, inctime=True):
...
+        pulses saved in `Processor.pulses` but return a new list.
```
nathanshammah: returns

qutip/qip/device/processor.py
```diff
@@ -308,107 +456,100 @@ def read_coeff(self, file_name, inctime=True):
...
+        The length of the new list of noisy pulses might be longer
+        because of drift Hamiltonian and device noise. They will be
+        added to the end of the pulses list.
 
         Parameters
         ----------
-        args: dict, optional
-            Arguments for :class:`qutip.QobjEvo`
+        ind_device_noise: bool, optional
+            If include pulse independent noise such as single qubit
```
nathanshammah: If "true"? By the way, I would remove `ind` for independent, also in `ind_drift`, or would spell it `independent`.

BoxiLi: I see, it is confusing. What I mean is `include_device_noise`. What about remove `ind` and make it a  boolean  parameter `device_noise`?

qutip/qip/device/processor.py
BoxiLi: Make sense. I'll change it to `add_drift(self, hamiltonian, targets, cyclic_permutation=False)`.

qutip/qip/device/modelprocessor.py
BoxiLi: This is an old API. I think I will just remove it as we already broke some default use of the old `Processor` anyway.

BoxiLi:
Thanks for the comments Nathan. Yes, I'm already updating the notebooks, will open a PR in the next few days.

qutip/qip/device/modelprocessor.py
BoxiLi: Deprecation warning exists

BoxiLi:
Move compiler to a separate package. The new code climate issues are all from there.

After #1074 get merged, one has to update all the `qutip.qip.gates` to `qutip.qip.operations` in this PR. Everything else should be fine.

BoxiLi:
Conflicts resolved @Ericgig

BoxiLi:
@quantshah 
Hi Shahnawaz, do you still have suggestions on this PR? It would be great if we can get this merged before Saturday.

BoxiLi:
I'll update the notebook and test this more.

nathanshammah:
There are some issues with OptPulseProcessor

BoxiLi:
@nathanshammah All tests passed now

Issue #363, opened 2015-08-28
Complete tensor product shorthand for state, operator initialization


cgranade:
Currently, the`identity` and `qeye` functions each take lists of dims as an alternative to an integer. For instance, `qeye([2, 3])` is equivalent to `tensor(qeye(2), qeye(3))`. While convenient, this is inconsistent with `zero_oper`, `basis`, `maximally_mixed_dm`, `projection` and `zero_ket`. Would it be worth extending the shorthand used by `identity` to these functions?

agarwal1998:
I think there should be the shorthand notation for zero_oper, basis, maximally_mixed_dm, projection and zero_ket . Since it provides a uniformity which helps to grasp the coding style easily and supports abstraction to some extent.
 But this functionality seems to be hidden as I couldn't find it in documentation or examples

ajgpitch:
There is an example in `qutip.operator.qeye`. It is possible it is only described in the API documentation

jakelishman:
I'd be interested in providing these changes.

I have a few points to ask/consider:
 1. which function is meant by `zero_oper()`?  I can't find any reference to it in the code, and judging by the name, I'd guess it's equivalent to `qzero()`, which already supports this.
 2. should we extend the automatic tensoring to other operators, e.g. `create()`, `destroy()`, and so forth?  I can't necessarily think of any particular use-case, but _not_ having it causes an odd inconsistency with `qeye()`.
 3. for `projection()` and `basis()` in particular: should the interface allow only scalar `m` and `n`, which must be the same for all dimensions, or should there be broadcasting rules?  The latter seems like a more useful choice to me, with the rules (and equivalent for however many arguments):
    - if `m` is a list, then `len(dims) == len(m)`
    - if `m` is a scalar, then `m = [m] * len(dims)`
 4. should there be error-checking that supplied scalars/lists are the correct format?  For lists in particular, the current behaviour is to silently flatten nested lists, though this is presumably a user error.

Ericgig:
This issue is open since 2015 and no one did anything about it mostly because there are no use case for adding list support for most of those function and removing them from `qeye` may break something.

What I would see as an improvement: if `N` could be the dimension of the resulting object.
`qeye(dims).dims = [dims,dims]`. This would be clear for `qeye` and `qzero` (`zero_oper` does not exist anymore),  since only the dimension of the resulting object is affected. For `create`, etc. hiding the tensor product would be hiding the physic, so it would be better to leave them as is.

[...]

PR #1157, opened 2020-01-16
Remove unnecessary circular call in qeye

jakelishman:
Hello,

There is currently a circular function call in the tensor produce part of `qeye()` - the list comprehension calls `identity()` for each element of the space, which is just a pass-through back to `qeye()`, rather than explicitly recursing.  I've just made the recursion explicit.

Also, as it stands, the "list of ints" argument will accept arbitrarily-nested lists of integers, and silently flatten it out.  Is it worth adding an error check to catch what is (presumably) user error?

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -454,7 +454,7 @@ shape = [3, 3], type = oper, isHerm = True
 
     """
     if isinstance(N, list):
-        return tensor(*[identity(n) for n in N])
+        return tensor(*[qeye(n) for n in N])
     N = int(N)
     if N < 0:
         raise ValueError("N must be integer N>=0")


Ericgig:
Thank you for catching this.
I do not believe that adding a check is needed. Ideally it should not flatten the list so `qeye` could create superoperator.

jakelishman:
Thanks.

For the superoperators part, I think there's perhaps a larger discussion to be had about how the whole process should generalise; `qzero()` already has the same behaviour due to the same kind of recursion, but it could be handled better by a decorator which handles broadcasting rules for dimensions.  I think this pull request could be ignored in favour of a proper solution to #363?

nathanshammah:
There is a small conflict to resolve here.

jakelishman:
This was deprecated by #1159 - this version was the solution before we revamped how the implicit tensor product worked. The new one has no recursive call at all, so this is unnecessary.

PR #1159, opened 2020-01-18
Implicit tensor product for qeye, qzero and basis.

jakelishman:
This makes the additions referred to in #363.  As `fock`, `fock_dm` and `projection` are thin passes-through to `basis`, the automatic tensor-producting applies to them too.

`qeye` and `qzero` are reworked to remove all calls to `tensor`, and instead just build their underlying matrix and override the `dims` when converting to a `Qobj`.  The new functionality in `basis` also follows a similar form.

This deprecates my previous pull request #1157, since the recursive call is entirely removed.

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -440,7 +440,8 @@ def _implicit_tensor_dimensions(dimensions):
     size = 1
     for dimension in dimensions:
         if not isinstance(dimension, numbers.Integral):
-            raise TypeError("'dimensions' must be an integer or list of integers.")
+            message = "'dimensions' must be an integer or list of integers."
+            raise TypeError(message)
         if dimension < 0:
             raise ValueError("All dimensions must be integers >= 0")
         size *= dimension
@@ -493,13 +494,15 @@ def qeye(dimensions):
     Examples
     --------
     >>> qeye(3)
-    Quantum object: dims = [[3], [3]], shape = (3, 3), type = oper, isherm = True
+    Quantum object: dims = [[3], [3]], shape = (3, 3), type = oper, \
+isherm = True
     Qobj data =
     [[ 1.  0.  0.]
      [ 0.  1.  0.]
      [ 0.  0.  1.]]
     >>> qeye([2,2])
-    Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
+    Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, \
+isherm = True
     Qobj data =
     [[1. 0. 0. 0.]
      [0. 1. 0. 0.]
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -42,6 +42,7 @@ __all__ = ['jmat', 'spin_Jx', 'spin_Jy', 'spin_Jz', 'spin_Jm', 'spin_Jp',
            'qutrit_ops', 'qdiags', 'phase', 'qzero', 'enr_destroy',
            'enr_identity', 'charge', 'tunneling']
 
+import numbers
 import numpy as np
 import scipy
 import scipy.sparse as sp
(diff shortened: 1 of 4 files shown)


jakelishman:
I split the two formatting commits up so that `6e2ca88` can be squashed into `6a19041` and `cac2304` into `a315b0` on rebase/merge.

qutip/operators.py
```diff
@@ -422,17 +423,65 @@ def create(N, offset=0):
     return qo.dag()
 
 
+def _implicit_tensor_dimensions(dimensions):
```
Ericgig: Dimensions are not just a list of integer but can be nested list for superoperator:
```
>>> liouvillian(qeye(2)).dims
[[[2], [2]], [[2], [2]]]
```
There are function in qutip.dimensions that could be useful like `flatten`

jakelishman: Ah, thanks for pointing me to `qutip.dimensions` - I'd not come across that yet.  I'm still working on getting more familiar with the internals.

jakelishman:
beb3d66 replaces my previous (fairly poor) `_implicit_tensor_dimensions` with one that should accept anything that `qutip.dimensions.type_from_dims` would accept as an operator or super-operator.  Thanks for the code review - I've never really worked with anyone else before, so it's good for me to get into good habits.

qutip/operators.py
```diff
@@ -422,20 +424,78 @@ def create(N, offset=0):
...
+        Dimension list in the form required by ``Qobj`` creation.
+    """
+    if not isinstance(dimensions, list):
+        dimensions = [dimensions]
+    flat = flatten(dimensions)
+    if not all(isinstance(x, numbers.Integral) and x >= 0 for x in flat):
+        raise ValueError("All dimensions must be integers >= 0")
+    dimensions = [dimensions, dimensions]
+    if type_from_dims(dimensions) not in ['oper', 'super']:
+        raise TypeError("`dimensions` must be a valid single dimension of a "
```
Ericgig: `type_from_dims([[1],[1]])` is `"bra"`, not `"oper"`. So `qeye(1)` fails.

jakelishman: This is a good point.  After a bit more testing, I think the handling of 1D Hilbert spaces is a rather inconsistent throughout `qutip`, with the bug here manifesting itself in a few different places.

Some things I found:
  1. `qutip.create(1).type == qutip.basis(1, 0).type == 'bra'`.  This is the same type-inference bug as you've pointed out here, but those two functions don't test for it, so it silently gives an unexpected result (there are probably more, as `create` is one I haven't touched, and the bug existed in the old `basis` too), which fortunately is then mostly ignored by other calling code.
  2. `1 + qutip.tensor(qutip.create(1), qutip.create(1))` raises `TypeError` for incompatible dimensions, but making either call `create(2)` succeeds (as it should).  Incidentally, `1 + qutip.basis(5, 0)` succeeds - adding a scalar to a ket -, which I personally find rather odd.
  3. Take a tensor-product operator where any one or more of the Hilbert spaces have one subspace of dimension one, for example `a = qutip.tensor(qutip.create(2), qutip.create(1))`.  Now `a + a` and `1 * a` are the correct dimensions, but `a * a` contracts all scalar dimensions to a maximum of one (and that remains only if _all_ the spaces' dimensions are one), so `a * a * a` is a `TypeError`.

Looking at the code, this final multiplication one seems to be deliberate, but it's pretty surprising behaviour in this case, no matter how mathematically useless those operators are, and it's probably a bug that addition and multiplication are inconsistent.  I assume this is to allow tensor-product bra/ket multiplication to contract to a canonical scalar representation.

### Possible fixes

`type_from_dims` could return a `set` of valid types rather than picking one at random, and allow the caller to decide what is sensible in this situation.  In this case, `Qobj` initialisation should fail loudly if the type isn't explicitly passed and there's more than one possibility. Alternatively, and also not entirely satisfactorily, the default for `type_from_dims([[1], [1]])` (or other equivalent scalar forms) could be `'oper'`, not `'bra'`.  It's not really either (could be a new `'scalar'` type?), but numeric types behave like they're multiplied by an implicit identity operator, so `'oper'` seems more reasonable.

Mostly, I think points 2 and 3 are inconsistent with each other, rather than either being necessarily an error.  Still, I should be able to add a scalar to a square operator of any tensor-product size, and the default behaviour allowing a square operator to have `a**3` be valid and `a*a*a` be a `TypeError` is wrong.  Perhaps there should be a check in the automatic dimension contraction to contract only if that dimension changed size compared to either `Qobj`?

Ericgig: The questions is mostly how much energy do we want to spend on "scalar" Hilbert space, I don't see any use for them other than quick tests. I would be fine with simply removing the dimension check in `qeye`. It is redundant as a test.

`type_from_dims` returning a set of all possible type could be a good option. But I am not sure it's needed. Setting 'oper' as the default is probably sufficient. Not knowing the type should not cause an error.

`1 + qutip.tensor(qutip.create(1), qutip.create(1))` There is actually an exception to stop these cases. `1 + qutip.basis(5, 0)`: Quite surprising behavior. I expect this is an error that come from cut/pasting the product code. Updating `Qobj.__add__` deserve it's own issue.

3 - Setting `qutip.settings.auto_tidyup_dims = False` make `a*a*a` behave as expected. There is also an old PR to correct this: #735.

jakelishman: In theory I don't like removing the test because I think library functions should be conservative in what they return, and it shouldn't be possible to use `qutip` library functions to get an invalid `Qobj`.  That said, `type_from_dims` assumes that the dimensions are properly formatted, so it wasn't a great test.

I've removed the test for now, but I think strengthening `Qobj.type` and the dimension checks could make it easier to handle some of the edge cases we were discussing above.

jakelishman: For the other points:

If `def op(n): return qutip.tensor(qutip.create(n), qutip.create(n))`, then surely either `1 + op(1)` and `1 + op(2)` should both error, or neither of them should?  The exception at the moment catches it at the moment presumably to prevent things like `qutip.create(1) + qutip.tensor(qutip.create(1), qutip.create(1))`, which _would_ be an error, but because scalars are promoted to `Qobj` at the top of the function, they get caught up in it too.

`1 + qutip.basis(5, 0)` seems to have been added deliberately - there's an `if` branch to catch the case of `Qobj.type not in ['oper', 'super']`.  Without knowing the reason, it seems a very odd choice, especially because it only adds the scalar to the filled entries, which again, was clearly a deliberate decision in the code.

I'll have a look at the interplay of `Qobj.type`, `__add__` and `__mul__` and see if I can iron out some of those bugs without introducing a whole bunch more.  But that would be a different PR.

Issue #1158, opened 2020-01-18
qutip.testing.run() NameErrors


jakelishman:
I think there's been some form of merge conflict in `qutip.testing` - for one, it looks like the tests are intended to run twice (once under `pytest` and once under `nose`), and secondly it refers to the name `qset`, which isn't defined.

nathanshammah:
I think you are right. According to [this version](https://github.com/qutip/qutip/commit/0d23a5b72e3eb2c9567d8a3918388cf96370438a#diff-b979016620c6998841cf7cefe791ebfe) and [this one](https://github.com/qutip/qutip/commit/97a65a3fee0983ae2d823a4fdf886aa02965ee5b#diff-b979016620c6998841cf7cefe791ebfe), before some of the latest and recent commits, the changes should be 
- `import qutip.settings as qset` 
- set the settings properties before running the tests
- delete the `nose` tests.
 
Would you like to fix it?

jakelishman:
I've sent a PR #1161 that fixes this, but a bigger worry is why this wasn't caught on Travis on other commits?  I'm not familiar with the CI system, but at a cursory glance it looks like it just tests that QuTiP builds, and not that the tests succeed.  Is that intentional?

nathanshammah:
I think that `qutip.testing` and the tests in CI are different. The former are in that file, the latter look at the instructions from [`.travis.yml`](https://github.com/qutip/qutip/blob/master/.travis.yml), which basically call `python setup.py install` (or `test`). Also looking at the Travis CI reports, it seems that they work fine. Locally, I'm getting tests to fail on my machine, with malloc or abort trap 6 errors every other time, although some time they pass (hence I opened #1160). It may be that they fail on my machine due to some mess with the gcc compiler.

jakelishman:
Yeah, that was what I was referring to - Travis just tests that the Cython parts compile.  I'm new to working on code in any kind of organisation, so I wasn't sure if it's normal to have the unit tests run as part of the CI process, or if that was only meant to test the installation path.

PR #1161, opened 2020-01-18
Remove duplicated test runners

jakelishman:
Removes the `nose` test runners found in `qutip.testing.run()` in #1158.

This isn't ready to be merged yet, I think - there's still references to `nose` in `setup.py`, and a comment in `test_qobj.py`, the context of which suggests that the tests are organised specifically for `nose`.  I'm not familiar enough with the differences between `nose`, `pytest` and `unittest` to know how much of a problem the latter is, though.

### DIFF
--- a/qutip/testing.py
+++ b/qutip/testing.py
@@ -31,18 +31,15 @@
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 from qutip.about import about
-from qutip import settings
+from qutip import settings as qset
 
 def run():
     """
-    Run the nose test scripts for QuTiP.
+    Run the test scripts for QuTiP.
     """
     # Call about to get all version info printed with tests
     about()
     import pytest
-    # runs tests in qutip.tests module only
-    pytest.main(["--verbosity=1",
-                "--disable-pytest-warnings", "--pyargs", "qutip"])
     real_num_cpu = qset.num_cpus
     real_thresh = qset.openmp_thresh
     if qset.has_openmp:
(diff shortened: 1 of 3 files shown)


nathanshammah:
Good points. I suggest to simply remove the comment in `test_qobj`, and change to pytest in the `setup.py`. Something like what's [here](https://doc.pytest.org/en/3.0.4/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner) should work (basically changing `nose` with `pytest`). It seems however that testing it's a thorny [issue](https://github.com/pypa/setuptools/issues/1684), but these changes for the time being make the whole package at least consistent.

jakelishman:
So I pushed two commits here - one which makes the changes to `setup.py`, and another which doesn't just remove the commented reference to `nose`, it completely overhauls `test_qobj.py` into a more `pytest`-friendly format.

I found that removing the comment would have left a very odd function that was used to decorate test cases, which isn't necessary in `pytest` so much because you can use parametrisation, more description function names, and grouping in `Test*` classes to achieve the same goal with less custom wrappers.  Once I was on that, I kind of got carried away changing the rest of the file.

Let me know if that's too much - I could rebase this PR (and resubmit it) to split up the commits if you'd rather, or I can I just remove that last commit which makes all those changes.

At the time of writing, the CI tests haven't run, and I'm a little worried that they'll fail due to the added dependency on `pytest-runner` in `setup.py`, when I haven't modified `.travis.yml`.  Even if the CI tests don't fail, I think there might be some knock-on effects for documentation or packaging from that change?

jakelishman:
Also, when you run `pytest` on the file (`pytest qutip/tests/test_qobj.py` from the git root), it moans a lot about `np.matrix` being essentially deprecated, and `Qobj` uses it as a data representation in some cases.  I haven't touched that, but I also didn't want to hide the warning.

jakelishman:
Ok, I force-pushed a removal of the last commit (which just added the `# TODO` comment), so it's gone now without adding chaff to the commit history.

nathanshammah:
Great.

PR #1170, opened 2020-02-01
Activate Python 3.8 builds on Linux.

hodgestar:
Python 3.8 was released awhile ago, so I thought I'd try activate its travis build and see if it passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -151,4 +151,6 @@ jobs:
     - stage: test
       <<: *stage_linux_37_openblas
     - stage: test
+      <<: *stage_linux_38
+    - stage: test
       <<: *stage_osx


hodgestar:
Build succeeded. Should we merge this? Or should I extend this to perhaps do more builds on 3.8 and reduce 3.7 to just the basic build? Should 3.6 be dropped or kept?

Ericgig:
We will keep 1 test for each 3.6 and 3.7. If the test passes, other tests should be moved to 3.8. But this may be another PR is you are ready to merge this as is.

nathanshammah:
Sorry what is the difference between `stage_linux_38: &stage_linux_38` and this? And in general  between the `stage_linux_...` and the `test`?

Ericgig:
`stage_linux_38: &stage_linux_38` define what to do in the test, but does not run it. This addition tell travis to run the test. When I defined the test, 3.8 just came out and there were missing dependencies. We still have no automated test running on 3.8. I am merging this and will make another PR to move other tests from 3.7 to 3.8.

PR #1168, opened 2020-01-29
Update setup file for dev to bump up on top of 4.5.0 release

nathanshammah:

PR #1171, opened 2020-02-01
update version to4.6.0 development in setup.py

nathanshammah:
I had forgotten to do this version bump in the setup.py file after the version release. This PR needs to be merged before any other PR.


nathanshammah:
Actually, nothing bad should happen even if it is merged later.

nathanshammah:
Addressed already by #1168

PR #1177, opened 2020-02-06
Fix clebsch function not working with non-int inputs

taw181:
Changed `clebsch` to work with non-int inputs. Test file shows it working for both int and float inputs, giving expected results.


Ericgig:
Hi taw181, thank you for the PR. 
There are already test for `clebsch` in `test_utilities`. Could you move your new test there?

qutip/utilities.py
```diff
@@ -161,29 +161,29 @@ def clebsch(j1, j2, j3, m1, m2, m3):
...
-    _factorial_prod(j3 - j1 + j2, c_factor)
-    _factorial_prod(j1 + j2 - j3, c_factor)
-    _factorial_prod(j3 + m3, c_factor)
-    _factorial_prod(j3 - m3, c_factor)
-    _factorial_div(j1 + j2 + j3 + 1, c_factor)
-    _factorial_div(j1 - m1, c_factor)
-    _factorial_div(j1 + m1, c_factor)
-    _factorial_div(j2 - m2, c_factor)
-    _factorial_div(j2 + m2, c_factor)
+    c_factor = np.zeros((int(j1 + j2 + j3 + 1)), np.int32)
```
Ericgig: Wouldn't it have been easier to add the `int` inside the `_factorial_prod` and `_factorial_div` functions.

taw181: Yes it would have.

taw181:
> There are already test for `clebsch` in `test_utilities`. Could you move your new test there?

Yes sorry I didn't realise that was there, I put something together quickly because I was asked to include a test. I'll write something that fits better with the existing test and include it

Ericgig:
Are you still working on a test?

PR #1179, opened 2020-02-06
Removed converting error on windows.

JiihwanKim:
Removed error when converting qcirc.pdf to qcirc.png on windows.

### DIFF
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -58,8 +58,8 @@ def _latex_compile(code, filename="qcirc", format="png"):
     os.system("mv %s-tmp.pdf %s.pdf" % (filename, filename))
 
     if format == 'png':
-        os.system("magick convert -density %s %s.pdf %s.png"
-                      % (100, filename, filename))
+        os.system("magick convert -density %s %s.pdf %s.png" % (100,
+                                               filename, filename))
         with open("%s.png" % filename, "rb") as f:
             result = f.read()
     else:
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -31,6 +31,7 @@
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 import os
+import sys
 
 _latex_template = r"""
 \documentclass{standalone}
@@ -59,6 +60,9 @@ def _latex_compile(code, filename="qcirc", format="png"):
     if format == 'png':
         os.system("convert -density %s %s.pdf %s.png" % (100, filename,
                                                          filename))
+        if sys.platform == 'win32':
+            os.system("magick convert -density %s %s.pdf %s.png" 
+                      % (100, filename, filename))
         with open("%s.png" % filename, "rb") as f:
             result = f.read()
     else:
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -58,10 +58,7 @@ def _latex_compile(code, filename="qcirc", format="png"):
     os.system("mv %s-tmp.pdf %s.pdf" % (filename, filename))
 
     if format == 'png':
-        os.system("convert -density %s %s.pdf %s.png" % (100, filename,
-                                                         filename))
-        if sys.platform == 'win32':
-            os.system("magick convert -density %s %s.pdf %s.png"
+        os.system("magick convert -density %s %s.pdf %s.png"
                       % (100, filename, filename))
         with open("%s.png" % filename, "rb") as f:
             result = f.read()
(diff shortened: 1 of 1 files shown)


jakelishman:
If I understand correctly, this is more to do with the `ImageMagick` system changing, rather than Windows in particular - ImageMagick v7 introduced the command `magick` in place of the previous `convert` (though at least on my Mac, `convert` still works as expected): see https://www.imagemagick.org/discourse-server/viewtopic.php?t=29582.

qutip/qip/circuit_latex.py
```diff
@@ -59,6 +60,9 @@ def _latex_compile(code, filename="qcirc", format="png"):
     if format == 'png':
         os.system("convert -density %s %s.pdf %s.png" % (100, filename,
                                                          filename))
+        if sys.platform == 'win32':
+            os.system("magick convert -density %s %s.pdf %s.png"
+                      % (100, filename, filename))
```
jakelishman: This duplicates the system command if the platform is Windows (and also duplicates the code).  Perhaps it would be better to do something like `command  = magick_command + magick_arguments.format(100, filename, filename)`, and detect/set `magick_command` only once (could be cached)?

JiihwanKim: From your response, I searched a little about ImageMagick. 
As you pointed out, command 'convert' was replaced with 'magick' from version 7 or higher.
But I'm not sure whether this works well in Mac. According to your response, 'convert' seems to still work well in Mac. Could you check 'magick' command also works in Mac? If it works, it would be better to adopt the command 'magick'. And if it does not work, could you check your ImagaMagick's version? Sorry to bother you. I want to check by myself, but I don't have Mac.

nathanshammah:
Only `convert` works for me, on Mac 10.14.5, ImageMagick 7.0.8-49 Q16 x86_64

qutip/qip/circuit_latex.py
nathanshammah: Edit: Both `convert` and `magick` work for me, on Mac 10.14.5, ImageMagick 7.0.8-49 Q16 x86_64

jakelishman: Both `convert` and `magick` work for me with ImageMagick installed by homebrew, version `7.0.9-16`.

While this is beside the point of the PR, perhaps you're missing a symlink to `magick`, Nathan?  My install has both the `magick` and `convert` binaries (although they're identical...).

JiihwanKim: I think my problem depends on OS. As far as I know command 'convert' is already native command in Windows ( it is used to change disk type: completely different purpose). In my guess, ImageMagick's changing command is reflecting this fact.

jakelishman:
If it helps in the future, you can use `pycodestyle` (available in `pip` and `conda`) to check pep8 compliance on your end before pushing a commit.

As to the actual command, we might not actually want to remove the usage of `convert` - it's quite possible that a lot of people are still using ImageMagick 6, since for most people it's just a dependency that got installed by something else, and isn't frequently updated.  `qutip` doesn't list it as an official dependency, so it wouldn't get updated on our behalf.

We could catch which version of `ImageMagick` to use with something like
```
def _magick_comamnd():
    for command in ['magick', 'convert']:
        try:
            subprocess.run((command, '--version'), capture_output=True)
            return command
        except OSError:
            pass
    raise FileNotFoundError("Couldn't locate system ImageMagick")
```
and cache the result on import (below into the variable `_magick`, then the convert command would become
```
subprocess.run((_magick, "-density", 100, filename + ".pdf", filename + ".png"))
```
or something like that.

JiihwanKim:
Thank you for your comments. I can understand what you mean. But apart from this problem, I have one question. If someone use ImageMagick 6, everythings work well? (including windows users)

jakelishman:
You're right in that everything works at the moment if somebody is using IM6, but with the proposed commit it would no longer work (since IM6 doesn't have the `magick` command).

That's not necessarily a problem as long as it's a conscious decision, and I can't provide that since I'm not a maintainer.

JiihwanKim:
Oh thank you a lot. Version of IM was the point. I'm sorry to bother you but is there any document commenting the required version?

BoxiLi:
ImageMagick is only mentioned in our tutorial page and not listed elsewhere. So we didn't specify a required version for it. 


> We could catch which version of `ImageMagick` to use with something like
> 
> ```python
> def _magick_comamnd():
>     for command in ['magick', 'convert']:
>         try:
>             subprocess.run((command, '--version'), capture_output=True)
>             return command
>         except OSError:
>             pass
>     raise FileNotFoundError("Couldn't locate system ImageMagick")
> ```
> 
> and cache the result on import (below into the variable `_magick`). There could also be logic based on OS if needed. The convert command would become
> 
> ```python
> subprocess.run((_magick, "-density", 100, filename + ".pdf", filename + ".png"))
> ```

This solution looks neat. One minor thing is that maybe we should use `check=True` instead of  `capture_output=True` since the latter one is only available from Python 3.7+.


Despite those has already been discussed above, the plot sill does not work for me. I find the following lines also wrong on Windows
``` python
    os.system("rm -f %s.aux %s.log" % (filename, filename))

    os.system("mv %s-tmp.pdf %s.pdf" % (filename, filename))
```
because `rm` and `mv` are not Windows command. We should probably replace it by python script `os.remove` and `os.replace` to make it system independent.

PR #1164, opened 2020-01-20
Move tests to pytest

jakelishman:
To get more familiar with bits of the internals, I'm going through and migrating the tests to `pytest`.  While I'm at it, I'm also fixing PEP8 issues in the testing files, since CodeClimate will moan at me if I don't.

I'm basically learning how to use `pytest` and how to code as part of an organisation as I go, so please feel free to correct me if I'm doing something wrong.

I'll leave this PR as in draft mode until I've made it through more of the testing files.

PR #1181, opened 2020-02-08
Convert tests to pytest

jakelishman:
Continuing the work of #1164 - this is part two of many more to come, depending on how long I can keep doing this!  Of particular note is the first commit in this PR, which registers `pytest.mark.slow` with `pytest` to stop it warning about an unknown mark in use.

There are several additional comments and explanations in the expanded commit messages, where I've tried to explain a little bit about bugs I found and reasons for swapping over to certain libraries (e.g. for temporary file handling).

I'm in part opening a new PR now so that I can see if the CodeClimate tests are passing.  I shouldn't have many pep8 issues because I have a tool to check, but I don't have CodeClimate working locally for things like the complexity (if it even checks them for tests...).

I've been really rather aggressively using the parametrisation that `pytest` offers for fixtures, which is cutting out an awful lot of code duplication, and hopefully making it a bit clearer when things are exactly the same.

One common theme I'm finding when I'm running these tests is that there's an awful lot of warnings caused by `pyximport`, a part of `Cython`, which is using the Python 2 era `imp` instead of the newer `importlib`.  I'm not sure there's anything we can or should do about this, though.

### DIFF
--- a/qutip/tests/test_interpolate.py
+++ b/qutip/tests/test_interpolate.py
@@ -62,7 +62,7 @@ def test_equivalence_to_scipy(noise):
     expected = scipy.interpolate.CubicSpline(x, y, bc_type="natural")
     # We use the bc_type="natural", i.e. zero second-derivatives at the
     # boundaries because that matches qutip.
-    assert np.allclose(test(x), expected(x), atol=1e-10), "Array argument"
+    np.testing.assert_allclose(test(x), expected(x), atol=1e-10)
     centres = 0.5 * (x[:-1] + x[1:])
     # Test some points not in the original.
     for point in centres:
@@ -149,4 +149,4 @@ def test_usage_in_solvers(solver, coefficients):
         H_test.append([operator, other])
     expected = solver(H_expected, state, times, e_ops=e_ops).expect[0]
     test = solver(H_test, state, times, e_ops=e_ops).expect[0]
-    assert np.allclose(test, expected, atol=1e-4)
+    np.testing.assert_allclose(test, expected, atol=1e-4)
--- a/qutip/tests/test_interpolate.py
+++ b/qutip/tests/test_interpolate.py
@@ -30,325 +30,123 @@
 #    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
-import pytest
+import functools
 import numpy as np
-import scipy.interpolate as sint
-from numpy.testing import assert_, assert_equal, run_module_suite
-import unittest
-from qutip import *
-from qutip import _version2int
+import scipy.interpolate
+import pytest
+import qutip
 
-try:
-    import Cython
-except:
-    Cython_OK = False
-else:
-    Cython_OK = _version2int(Cython.__version__) >= _version2int('0.14')
+pytestmark = [pytest.mark.usefixtures("in_temporary_directory")]
 
 
-def testInterpolate1():
-    "Interpolate: Sine + noise (array)"
-    x = np.linspace(0,2*np.pi,200)
-    y = np.sin(x)+0.1*np.random.randn(x.shape[0])
-    S1 = Cubic_Spline(x[0],x[-1],y)
-    S2 = sint.interp1d(x,y,'cubic')
-    assert_(np.max(np.abs(S2(x)-S1(x))) < 1e-9)
+def pytest_generate_tests(metafunc):
+    """
+    Perform more complex parametrisation logic for tests in this module.  This
+    was originally needed because the `brmesolve` solver cannot mix splines and
+    functions in the Hamiltonian (it can only accept string-formatted
+    time-dependence), but we want to parametrize this for all the other tests.
+    """
+    if ("solver" in metafunc.fixturenames
+            and "coefficients" in metafunc.fixturenames):
+        _parametrize_solver_coefficients(metafunc)
 
 
-def testInterpolate2():
-    "Interpolate: Sine + noise (point)"
-    x = np.linspace(0,2*np.pi,200)
-    y = np.sin(x)+0.1*np.random.randn(x.shape[0])
-    S1 = Cubic_Spline(x[0],x[-1],y)
-    S2 = sint.interp1d(x,y,'cubic')
-    for k in range(x.shape[0]):
-        assert_(np.abs(S2(x[k])-S1(x[k])) < 1e-9)
+@pytest.mark.parametrize('noise', [
+    pytest.param(lambda n: 0.1*np.random.rand(n), id="real"),
+    pytest.param(lambda n: 0.1j*np.random.rand(n), id="complex"),
+])
+def test_equivalence_to_scipy(noise):
+    x = np.linspace(0, 2*np.pi, 200)
+    y = np.sin(x) + noise(x.shape[0])
+    test = qutip.Cubic_Spline(x[0], x[-1], y)
+    expected = scipy.interpolate.CubicSpline(x, y, bc_type="natural")
+    # We use the bc_type="natural", i.e. zero second-derivatives at the
+    # boundaries because that matches qutip.
+    assert np.allclose(test(x), expected(x), atol=1e-10), "Array argument"
+    centres = 0.5 * (x[:-1] + x[1:])
+    # Test some points not in the original.
+    for point in centres:
+        assert np.abs(test(point) - expected(point)) < 1e-10, "Scalar argument"
 
 
-def testInterpolate3():
-    "Interpolate: Complex sine + noise (array)"
-    x = np.linspace(0,2*np.pi,200)
-    y = np.sin(x)+0.1*np.random.randn(x.shape[0]) + \
-        0.1j*np.random.randn(x.shape[0])
-    S1 = Cubic_Spline(x[0],x[-1],y)
-    S2 = sint.interp1d(x,y,'cubic')
-    assert_(np.max(np.abs(S2(x)-S1(x))) < 1e-9)
+# We make a bunch of different test cases for the coefficients, all as a
+# three-tuple of
+#   (string, other, type)
+# where `string` is the string-format time-dependence (we use this as the
+# reference), `other` is what will become the time-dependence for this part of
+# the test, and `type` is a string in ["spline", "function", "string"] which
+# tells us how to interpret the type of `other`.  If `type` is "function" or
+# "string", then nothing is done to `other` and it is used as-is.  If `type` is
+# "spline", then `other` is given as a vectorised function which we turn into a
+# qutip.Cubic_Spline using appropriate `x` and `y` values.
+_case_real_spline = ('0.25*sin(t)', lambda t: 0.25*np.sin(t), 'spline')
+_case_real_string = ('0.25*cos(t)', '0.25*cos(t)', 'string')
+_case_real_function = ('0.25*sin(t)', lambda t, *_: 0.25*np.sin(t), 'function')
+_case_complex_spline = ('0.1j*sin(t)', lambda t: 0.1j*np.sin(t), 'spline')
+_case_complex_string = ('-0.1j*sin(t)', '-0.1j*sin(t)', 'string')
 
 
-def testInterpolate4():
-    "Interpolate: Complex sine + noise (point)"
-    x = np.linspace(0,2*np.pi,200)
-    y = np.sin(x)+0.1*np.random.randn(x.shape[0]) + \
-        0.1j*np.random.randn(x.shape[0])
-    S1 = Cubic_Spline(x[0],x[-1],y)
-    S2 = sint.interp1d(x,y,'cubic')
-    for k in range(x.shape[0]):
-        assert_(np.abs(S2(x[k])-S1(x[k])) < 1e-9)
+def _parametrize_solver_coefficients(metafunc):
+    """
+    Perform the parametrisation for test cases using a solver and some
+    time-dependent coefficients.  This is necessary because not all solvers can
+    accept all combinations of time-dependence specifications.
+    """
+    solvers = [
+        (qutip.sesolve, 'sesolve'),
+        (qutip.mesolve, 'mesolve'),
+        (functools.partial(qutip.mcsolve, ntraj=500), "mcsolve"),
+        (qutip.brmesolve, 'brmesolve')
+    ]
+    coefficients = [
+        ([_case_real_spline], "real-spline"),
+        ([_case_real_spline, _case_real_string], "real-spline,string"),
+        ([_case_real_spline, _case_real_function], "real-spline,function"),
+        ([_case_real_function], "real-function"),
+        ([_case_complex_spline, _case_complex_string],
+            "complex-spline,string"),
+    ]
+
+    def _valid_case(solver_id, coefficient_id):
+        invalids = [
+            "brmesolve" in solver_id and "function" in coefficient_id,
+        ]
+        return not any(invalids)
+
+    def _make_case(solver, solver_id, coefficient, coefficient_id):
+        marks = []
+        if "brmesolve" in solver_id:
+            marks.append(pytest.mark.requires_cython)
+        id_ = "-".join([solver_id, coefficient_id])
+        return pytest.param(solver, coefficient, id=id_, marks=marks)
+
+    cases = [_make_case(solver, solver_id, coefficient, coefficient_id)
+             for solver, solver_id in solvers
+             for coefficient, coefficient_id in coefficients
+             if _valid_case(solver_id, coefficient_id)]
+    metafunc.parametrize(["solver", "coefficients"], cases)
 
 
 @pytest.mark.slow
-def test_interpolate_evolve1():
+def test_usage_in_solvers(solver, coefficients):
     """
-    Interpolate: sesolve str-based (real)
+    Test that the Cubic_Spline can be used as a time-dependent argument to all
+    three principle solvers, both alone and in combination with other forms of
+    time-dependence.
     """
-    tlist = np.linspace(0,5,50)
-    y = 0.25*np.sin(tlist)
-    S = Cubic_Spline(tlist[0], tlist[-1], y)
-    N = 10
-    psi0 = fock(N,1)
-    a = destroy(N)
-    H = [a.dag()*a,[a**2+a.dag()**2,'0.25*sin(t)']]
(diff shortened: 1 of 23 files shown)


jakelishman:
Weird CI error on f56d27f - the build log suggests there was an error in setup (looks like `numpy` failed to load?), but I don't think that commit actually caused it, since it doesn't touch `setup.py`.

jakelishman:
You're welcome!  I've just `qutip` a lot throughout my PhD, so it seems fair that I'd contribute back to development.  While I'm still getting used to the code base, I'm quite happy to do some "boring" software engineering-y maintenance bits, and plus, this helps me learn `pytest` a lot better.

This PR can be merged whenever if you're happy with the changes - I'll open a new one if/when I convert more files.

BoxiLi:
Hi @jakelishman , Thanks for the great work and effort! I'm reading your code and also try to learn some pytest. 

If I may, I have a minor suggestion. Could you maybe leave some more descriptive comments in the tests? I find it a bit hard to follow the logic when tests are automatically generated or parametrized, e.g. in test_control_pulseoptim. A lot of things are defined for the fixtures in the beging. It took me some time to understand what is happening. I guess it is because I'm not yet very familiar with the tools in pytest. Maybe some description like 
> We first define 4 test cases: ...

or

> Test function1 against case1, case2 ...

This might make it easier for whoever wants to add more tests later to understand what has been tested without too much digging.

And another personal idea, maybe it is good to make a separate PR every 10 or 15 tests? A new branch can still base on this one. Just after the old PR is merged, fewer changes will be shown the new PR. It might make the later review a bit easier.

Just personal ideas. Thanks again!

jakelishman:
Hi!

About the comments: yes, absolutely I'll leave more.  It's a bit silly that I haven't yet, since one thing that was frustrating me previously was not understanding what was trying to be done!  I think as time goes by I'm getting a bit better about being clearer in the code - some of the earlier parametrizations I did were too aggressive because I wasn't used to the tools yet.  The `test_control_pulseoptim` parts are very complex, in part because the underlying calling convention of that code is complex too.  I'll try and add more comments to it to ease that along, especially in the part where I just defined a load of systems to test.  That certainly should have been commented.

There are general comments about what I was using the `pytest` tooling for in the commit messages, but I didn't necessarily want to leave "intro to pytest" comments throughout, because it would mostly crowd out the actual tests, and the `pytest` documentation online is probably going to be a better explanation than what I can do.  I'm also learning `pytest` by doing this (I'd not used it before I opened the first PR)!

About the PRs: I'm happy with whatever workflow works best for you guys - I don't have any control over when the PR gets merged.  I just opened this one after the first one I wrote got merged.  I'm happy to just keep adding commits here until I'm all the way through.

Right now I'm a bit busier with work so my conversion rate has slowed a little.  I'll probably be back on it in a week or so.

BoxiLi:
Hi, you are absolutely correct that this is not a good place "intro to pytest". Comments on what is tested would be most helpful.

I see your point. If some files need to be updated all the time then you are probably right. I'm also no maintainer of QuTiP so merging is beyond my ability :) But I'm sure there is no hurry in this. I'm also using my spare time browsing around here.

jakelishman:
This latest commit fabd9f4 fixes several bugs in the testing of `qutip.qip.operations.gates` - see the commit message for more details.

We should also add tests that the explicit forms of the gates are correct (comparing to hard-coded arrays) in their base forms.  These tests should never really fail, but they should be there as a fail-safe sanity check against somebody changing things in that module in the future.

jakelishman:
I also note that Coveralls isn't running on any of these commits, and hasn't been run on `master` since December last year.  While so far I've just been converting tests, rather than writing new ones (for the most part), it seems wrong that it isn't being run.

jakelishman:
Skipping `test_graph.py` for the time being, pending decision/discussion on #1203.

Ericgig:
Hi jakelishman,
Thank you for your work.
Why do you change the `from qutip import ...` to `import qutip`? Seems to be a lot of work with no real improvement.

Also, do you know if pytest has a way to order tests? The tests are running in alphabetical order of the file name. It would be nice to test more basic function first, then does which build on them: `Qobj` -> `sesolve`, `mesolve` -> `floquet`, `correlation`. Say sesolve is broken, then a lot of other tests will also fail because of it, so catching it earlier in the test suite would be great.

jakelishman:
Hi!

#### Imports
Honestly it wasn't really a particularly conscious decision, and there's no problem with `from qutip import (x, y, z)` lines though I don't think `from qutip import *` is a very good idea, especially with quite how many names `qutip` exports.  The `from qutip.very.deep.nested.module import name` is certainly good, and I think I've generally left those.  I think I started converting in a few places because there were things like `from qutip import Cubic_Spline` and `from scipy.interpolate import CubicSpline`, and then the test was unreadable as to which was which.  Having `qutip.Cubic_Spline` and `scipy.interpolate.CubicSpline` is unambiguous.  The smaller cases like `from qutip import Qobj` are highly unlikely to have any conflicts, but in general I just went for consistency.

That all said, there's certainly arguments both ways, e.g. explicit `qutip.member` is good because it's clearly explicit at point of use which makes reading the code much easier, but bad because it adds unnecessary chaff which makes scanning lines take longer.  I hadn't put much thought into it before, and I think I was just generally writing in my default style.

#### Test ordering
I hadn't much considered this, but there are certainly ways of doing it.  Essentially `pytest` goes through a few stages when you run `pytest` on the command line: discovery, generation, running, reporting, and it has lots of hook functions we can define to modify the state inbetween and during each of them.

Default discovery is as you say: files in alphabetic order (I _think_ it's strictly shell globbing order, but generally that's alphabetical), then the contents of each file in order defined by the shared fixtures used.  We tend not to need large amounts of state to be held in fixtures, so our files largely just run top-to-bottom.

We can add in some logic into the collection-time hooks to re-order the tests.  There's a `pytest` plugin called [`pytest-ordering`](https://github.com/ftobia/pytest-ordering), but it looks like it's not really actively maintained any more and the logic is pretty straightforwards, so perhaps not a terrible idea just to do it ourselves.  Maybe I'm not javascript enough to enjoy a good micro-dependency?

It would also be quite nice if we could define the whole logical structure of the `qutip` package, to help keep everything in order as the package grows larger.  By that I generally just mean the internal structure, though cutting down the amount of global names exported could also be done in a sort of organisation like this: we could get a little more aggressive at not exporting things like `enr_*` or `three_level_*` to the global namespace, but instead put them in submodules.

While I think this isn't what you were asking, it's also quite easy to just run specific parts of the testing suite if you're doing it from the command line.  You can do `pytest file.py` to run only those in a specific file, or `pytest -m "mark logic"` (e.g. `pytest -m "not slow"`) to hit specifically marked tests.  Perhaps most powerfully, and one I use a lot when writing these, you can do `pytest -k "substring logic"` to run every test (potentially in the specified file(s)) that contains the elements specified in `"substring logic"`.

jakelishman:
Whoops, that's a mistake in fef7a9c - I accidentally removed a `skipif` checking for the Cython version.  I've updated my local setup so I have two environments, one with and one without `Cython` so I can catch that kind of thing locally.

I'm force-pushing over fef7a9c and bf6f75c  after a nice little rebase to keep the commit history tidier.  (Ideally I would have caught this locally and wouldn't have needed to force-push a rebase commit, but c'est la vie, and `--force-with-lease` didn't complain).

jakelishman:
Hmm.  Travis didn't run on the force-pushed version.  That's frustrating, particularly because I was force-pushing to correct a silly error that CI caught.

jakelishman:
@nathanshammah, @Ericgig, @BoxiLi: it might be good to merge in this PR now if you're happy with the changes. There seems to be quite a lot of renewed movement in other areas (especially `qutip.qip.gates`), and we're going to start having merge conflicts if this branch is too long-lived.

I can happily fix the merge conflicts if they arise, but having too many makes the commit history rather harder to navigate in the future when somebody's hunting for why things changed or bug hunting.

nathanshammah:
You're right in calling for review attention. I am currently swamped with other stuff so I hope that others can look into it soon.

jakelishman:
No problem.

I would suggest that at merge time, this PR is _not_ squashed but is simply rebased onto or merged into master.  A merge would ensure that the PR number is included in the git history if that's desired, but all the relevant explanations are also already in the commit messages.

It's a very large change, there's a lot of exposition in the commit messages, and each commit is a modular unit of work already, while the PR title won't actually be true (since not all of the tests are converted).  I know that the commit messages are all combined on github, but it's much more difficult to search for the small bit you need out of several pages of writing, rather than having only the relevant parts presented to you.

jakelishman:
Those last two CI failures are exceedingly odd, especially since they're not consistent with each other, and those last two commits essentially don't do anything.

I'm unable to reproduce either test failure on my machine.  The `mcsolve` error _could_ have just been unfortunate chance, since `mcsolve` is inherently a bit random, but the errors in `test_interpolate.py` are really odd.  I can't reproduce it at all, and it's inconsistent - it says that `mesolve` and `mcsolve` have (the same!) issue with one particular case, but in that case `mcsolve` decays exactly to `sesolve`, and that test passed...

BoxiLi:
Is it possible somehow to split the review? 21 test files seems a lot. I'll start from `qip` related tests and see how far I can go.

jakelishman:
If you click into "Files changed" you can just tick the files as you've looked at them.  To make it easier for you, I'd suggest having a copy of the old version next to them and reading the full commit message, because I usually explain what I'm doing in them and why I'm doing it that way.

If you're worried about the size of the PR in total, then I'd strongly recommend _not_ squashing this PR if/when it's merged, but instead just rebasing it onto `master` or doing a direct merge.  That way it doesn't look huge, it looks like a series of commits which each change one file, which is what you'd get if the PR was initially split.  The review doesn't end at the point the code is merged - somebody in the future may very well want to see why I made certain changes.  I know I've spent a long time looking through the old `git` logs to work out why tests were the way they were, and it's always much much more difficult when you've got one of those squash commits with like 30 sub-parts.  The commits in this PR are already modular, and with the exception that 7a0d6e6 could be squashed into 
1d60b46, they all represent an isolated unit of progression and there aren't any "work in progress" commits.  I didn't perform that squash locally because I would have had to force-push over 10+ commits.

Squashing the PR does make sense when there are a lot of wip commits for a small unit of work, but for a very large unit of work like this one, it's much better to interactive rebase to keep it neat locally (I've already done that for the most part), and never push wip commits in the first place.

jakelishman:
Everyone: (@ajgpitch, @BoxiLi, @Ericgig, @nathanshammah).  Apologies for bumping this again.

It would be good to get a strategy in place to handle a review for this.  Would it be easier if I close this PR and then make a whole series of new PRs with (say) 5 files in at a time?

I'm particularly keen to get the `test_gates.py` refactor merged in, because there's seemingly quite a lot of movement in the `qip` module.  If I do split the PR, then I'll rebase onto `master` so that the new tests added in #1224 are included without a merge resolution.

Ericgig:
It's certainly easier if it was multiple smaller ones but we can do it like this. You can simply continue in another Pr for the remaining tests.
I will start evaluating file by file starting with parts of the code I am the most familiar with:
- [ ] test_mcsolve.py
- [ ] test_lattice.py 
- [ ] test_interpolate.py 
- [ ] test_fastsparse.py 
- [ ] test_expect.py 
- [ ] test_eigenstates.py 
- [ ] test_dimensions.py 
- [ ] test_cy_structs.py

qutip/tests/test_mcsolve.py
```diff
@@ -30,561 +30,283 @@
...
+    """
+    Test that `mcsolve` correctly solves the system when there is a constant
+    Hamiltonian and no collapses.
+    """
+    def pytest_generate_tests(self, metafunc):
+        tol = 1e-8
+        expect = (qutip.expect(self.e_ops[0], self.state)
+                  * np.ones_like(self.times))
+        # Use partial to allow pickling.
+        zero_function = functools.partial(_return_constant, 0)
```
Ericgig, line 92: I would prefer you use `args` instead of partial. For now there is no real advantage, but when the object version arrive, it will be possible to change `args` without recompiling string. It is already possible with the `rhs_reuse` options, but not used much.

jakelishman: Oh yeah, I actually completely forgot about the `args` dictionary when I wrote that. That's a better solution regardless of what comes next.

qutip/tests/test_interpolate.py
```diff
@@ -30,325 +30,123 @@
...
+# three-tuple of
+#   (string, other, type)
+# where `string` is the string-format time-dependence (we use this as the
+# reference), `other` is what will become the time-dependence for this part of
+# the test, and `type` is a string in ["spline", "function", "string"] which
+# tells us how to interpret the type of `other`.  If `type` is "function" or
+# "string", then nothing is done to `other` and it is used as-is.  If `type` is
+# "spline", then `other` is given as a vectorised function which we turn into a
+# qutip.Cubic_Spline using appropriate `x` and `y` values.
+_case_real_spline = ('0.25*sin(t)', lambda t: 0.25*np.sin(t), 'spline')
```
Ericgig, line 82: The reference could be converted to a function to save time.

qutip/tests/test_interpolate.py
```diff
@@ -30,325 +30,123 @@
...
+        (qutip.sesolve, 'sesolve'),
+        (qutip.mesolve, 'mesolve'),
+        (functools.partial(qutip.mcsolve, ntraj=500), "mcsolve"),
+        (qutip.brmesolve, 'brmesolve')
+    ]
+    coefficients = [
+        ([_case_real_spline], "real-spline"),
+        ([_case_real_spline, _case_real_string], "real-spline,string"),
+        ([_case_real_spline, _case_real_function], "real-spline,function"),
+        ([_case_real_function], "real-function"),
```
Ericgig, line 105: `_case_real_function` does not use spline, does it need to be included here?

qutip/tests/test_interpolate.py
```diff
@@ -30,325 +30,123 @@
...
+    e_ops = [a.dag() * a]
+    H_expected = [a.dag()*a]
+    H_test = H_expected.copy()
+    operator = a**2 + a.dag()**2
+    for string, other, type_ in coefficients:
+        if type_ == 'spline':
+            other = qutip.Cubic_Spline(times[0], times[-1], other(times))
+        H_expected.append([operator, string])
+        H_test.append([operator, other])
+    expected = solver(H_expected, state, times, e_ops=e_ops).expect[0]
```
Ericgig, line 150: `mesolve` and `mcsolve` needs `c_ops`, otherwise they just call `sesolve`.

jakelishman: As far as I remember, I didn't add these tests, I just converted the pre-existing ones.  I noticed this behaviour, but I'd left it in these commits, because I wanted to be certain I wasn't accidentally removing any.

I'll convert them to tests of a type "`mcsolve` calls `sesolve` is `c_ops is None`" in another commit.

qutip/tests/test_interpolate.py
jakelishman: No, it can go.

This may have been a mistake that it was committed - originally I was getting failures on these cases, and I had to debug it.  It's possible I forgot to remove that case if it wasn't in the previous batch.

qutip/tests/test_interpolate.py
jakelishman: Sorry, I'm not sure what you mean.

The reason there's a layer of indirection at all is so that the test case can "feed in" the number of times that are to be used, but I think I had to add in a special case for splines anyway, so I might re-think that whole parametrisation.

Ericgig: What I meant was to use a function format coefficient instead of a string-format time-dependence for the reference. String format need compilation which take about a sec each time.

qutip/tests/test_interpolate.py
Ericgig: Looking at it, the original tests also did not have any c_ops... The fallback to `sesolve` was probably added after these tests. For the test, could you add a `qutip.qzero` as c_ops. It would ensure that the right solver is used.

qutip/tests/test_interpolate.py
jakelishman: Ah, so the original reason that the base was a string was to accommodate `brmesolve`, which can't accept the function forms.  I suspect they were originally just copy+paste code, so they were all the same.

Are you planning a complete overhaul of `brmesolve`?  If so, this switch-over could wait until then.  If not, I'll think rather more about how to handle these cases so it doesn't turn into a huge mess.

nathanshammah:
There are some conflicts.

BoxiLi:
The conflicts are due to some recent change in `qip`. I had a look at the PRs, those not merged #1187, #1226 do not modify tests. So it should be fine now to resolve it I suppose.
@jakelishman  Would you like me to review the current version of `test_gates` or after the conflicts are resolved? I suppose resolving the conflicts won't change the existing test functions.

I'll start with
- [ ] test_cavityqed.py
- [ ] test_gates.py

jakelishman:
Hi guys,
yeah, the conflicts are because of the now merged #1224, which we knew about when we merged that one.  It's a simple fix, but I'm hesitant to do it directly on this PR, depending on how we want to merge it.

The simplest way is that I make the required changes, fix the merge conflict, and we keep going until this whole behemoth merges in one go.  This ends up in the commit history being a bit of a mess, and it'll take ages.

Instead, I think it might be better if I instead split this PR into a few different ones:
1. add `pytest` markers and other test niceties
2. the test files that Eric is checking
3. the QIP-related ones so that Boxi can check those
4. `brmesolve` tests
5. trivially changed files
6. whatever's left

This has the advantage that I can rebase them all individually to neaten them up with any changes suggested, and we can get parts merged in faster.  PRs 2 to 6 will depend on PR 1, so this would require getting that first one merged ASAP, then the others will all be modular and can be merged in any order at any speed.

If you're happy with that plan, I'll cherry-pick/rebase out the necessary bits of the history and form the new PRs.  The end result will be 6 PRs, which, when merged, will have no conflicts, and will leave the repository in the exact same state it would be if this PR was merged in one lump (with the relevant conflict fixed).

For each PR, then we just have a simple process:
1. you review files
2. I make suggested changes
3. repeat until happy
4. I rebase the PR to squash it into a logical form
5. you merge

BoxiLi:
Sounds excellent to me! Let's say what the others have to say

qutip/tests/test_mcsolve.py
jakelishman: Fixed in #1250.

qutip/tests/test_interpolate.py
jakelishman: Fixed in #1250.

jakelishman:
This PR is now obsolete.  Please see #1249, #1250, #1251 and #1252!

Issue #1096, opened 2019-10-09
ENH: Update `qcircuit` version for LaTeX circuit visualization and consider `Quantikz`


nathanshammah:
QuTiP currently uses QCircuit 2 (version of 2011) to generate LaTeX circuits that are printed to screen or saved to file. A newer version of QCircuit is available, 2.6.0 (2018) https://www.ctan.org/pkg/qcircuit 
* Also, the use of this open source package could be highlighted more evidently in the documentation.

* Another package has been recently released, https://arxiv.org/abs/1809.03842, Quantikz, which has some features related to noise that could be useful especially with regard to PR #1065, Google Summer of Code project by @BoxiLi on noise modeling in QIP. As this other package is very young, maybe an option to draw the circuit with this package, instead of qcircuit, could be given, also to avoid issues.

christian512:
Hi, I just added the support for QCircuit 2.6 in my fork and tested it with [quantum-gates.ipynb](https://github.com/qutip/qutip-notebooks/blob/master/examples/quantum-gates.ipynb). Do you have any other notebook suggestion, which I could use for testing? 

I didn't look into Quantikz yet.

BoxiLi:
Hi @christian512, sounds great! This tutorial page covers all that we have for plotting circuits, so its should be good if all figures work well there. Would you like to open a Pull Request?

christian512:
Hi @BoxiLi, yes I'll do so :) 

There were some deprecation warnings and one cell not running in the notebook. But this was also the same for the old version of QCircuit. Already looked into the mistake and might open an issue in the repo.

BoxiLi:
The deprecation is because of the QuTiP 4.5 update. The notebook is a bit outdated. The importation should be `from qutip.qip.operations import *`. This should solve the problem

hodgestar:
`qcircuit` is no longer vendored by QuTiP, so it uses whatever version of `qcircuit` is installed along with LaTeX (e.g. TexLive includes `qcircuit`).

Mayank447:
@BoxiLi This issue should be moved to qutip-qip.

Currently the latex generated in texrenderer in qutip-qip is as per qcircuit, the renderering part is completely on user's system. Whether to provide an option to generate latex for quantikz is a separate issue.

veronikakurth:
Closing this issue here. @BoxiLi feel free to migrate it to qutip-qip if it's still something to be considered. Thanks!

PR #1187, opened 2020-02-29
Qcircuit update to version 2.6

christian512:
This PR updates the version of QCircuit used for plotting circuits.
LaTeX code was taken from the [QCircuit repository](https://github.com/CQuIC/qcircuit) and small modifications were done to change from .sty file to .tex file.


christian512:
@BoxiLi, should I try to format that string for QCircuit into pep8 conform 80 chars per line or keep it like this for better readability? 
Before it was also more than 80 chars per line. 
Unfortunately autopep8 does not format the multiline string automatically.

BoxiLi:
Since its a string for latex code, I think it's ok to keep it like this if you find this way is better for the readability.  It might still worth it to remove the Trailing whitespace and add an empty line at the end of the file.

nathanshammah:
Thank you for picking up my suggestion in #1096.

### Suggested Changes
- 1) There are some minor issure with CodeClimate ([here](https://codeclimate.com/github/qutip/qutip/pull/1187)). @christian512 can you please reformat to 79 lines?
This can be done with tools like `black`:
```
pip install black
black circuit.py --line-length 79
```
- 2) @christian512 could you add something also in `qutip-doc` to acknowledge the use of this library in the documentation? @BoxiLi recently added circuits to the User Guide, if I remember correctly. 

## Questions and comments
- 3) I then leave it to @BoxiLi to review the PR more closely, if he has time, otherwise I'll look into it this week. 

- 4) Is there a problem in theory using qcircuit which is with license `gpl-2.0` into a BSD-new (maybe @Ericgig @ajgpitch or @goerz know more).

christian512:
Thanks for the comment, @nathanshammah! 

- i. The problem is that the string is a raw string in order to keep backslashes and not use them as escape characters. So we would need to convert all of it to a string literal. Unfortunately Black does not provide an automatic tool for this. So I could do it manually, but this will be an extensive task...

- ii. Okay, I'll look into the documentation repository. Currently I am experiencing some problems with setting up the sphinx, but I'll try to figure out, what the problem is.

nathanshammah:
I see @christian512, so maybe we can skip that. 

With regards to the documentation, Sphinx can be a mess. I wrote some guidelines [here](https://github.com/nathanshammah/scikit-project/blob/master/5-docs.md).

I think you could skip installing it properly and just fork the repo and modify
the [guide-qip.rst](https://github.com/qutip/qutip-doc/blob/master/guide/guide-qip.rst) file, adding some information on the visualization of quantum circuits using this LaTeX-based tool. It could be a standalone subsection for simplicity. `.rst` files are restructured text file very similar to markdown.

christian512:
I've created a new PR for the qutip-doc repository containing these changes: https://github.com/qutip/qutip-doc/pull/108
Might need some feedback on the text I wrote.

BoxiLi:
@nathanshammah I'll first review #1179 and then come to this one.

jakelishman:
@BoxiLi #1179 is possibly not a suitable solution, because it would remove support for IM6.  I implemented an alternative at #1194 which should be a bit more resilient.  Both would be easy to merge in with this one.

BoxiLi:
Thanks for the PR @christian512. I tested on my machine, looks all good. 

> Is there a problem in theory using `qcircuit` code, which is with license `gpl-2.0` into a BSD-new (maybe @Ericgig @ajgpitch or @goerz know more)?

I'm not very familiar with license, but QuTiP was using this package (older version) before. Any idea on how to proceed? @Ericgig @ajgpitch or @goerz

jakelishman:
Out of interest, I had a look around.  BSD-3 clause and GPL-v2 are "compatible", but only in the sense that the resultant _must_ be GPL-licensed (see [a bit of discussion here](https://opensource.stackexchange.com/questions/7180/can-i-combine-gplv2-and-bsd-3-clause-code-and-keep-the-bsd-permissions)).  That probably means that we're technically not following the rules at the moment.

That said, my understanding is that licences are just a general statement of rights that are given to everyone.  If we get explicit permission from the creator of the library to include it in this library with a rider that this particular section cannot be copied except under GPL-v2 terms, then I think we'd be fine.

In general I think the primary reason that we can't release GPL'd code under a BSD licence is that the BSD licence doesn't have the restriction that source code must be available.  I _think_, however, that since the circuit drawing isn't an integral part of QuTiP, it would be allowed for us to distribute this portion as a stand-alone extension to QuTiP which is individually GPL'd, even if we couldn't get a relicensed version of the LaTeX package.

nathanshammah:
Thank you @jakelishman for the summary on the use of a GPL-2.0 in BSD-3.0. 
Maybe we can ping directly @Travis-S regarding the use of `qcircuit` 2.6 here for input on how to proceed. Let me point out that recently, reference to this project and a link was added in QuTiP's [user guide](https://github.com/qutip/qutip-doc/blob/f3592e9ad5a4083c19a8c7b81db8b2ae282cf4b4/guide/guide-qip.rst#plotting-a-quantum-circuit).

nathanshammah:
We can keep this in mind for the next micro version release.

jakelishman:
After some discussion in #1579, I'm fairly sure we cannot continue to vendor `Qcircuit.tex` in with QuTiP without breaking its licensing agreement.  I'm closing this PR, and I'll open one to remove the vendoring of `Qcircuit.tex` completely.

Issue #1172, opened 2020-02-03
ENH: qutip-doc refactoring, changelog, PR template


nathanshammah:
## Changelog
The changelog file is present in `qutip/qutip-doc`. This means that for each new release, the maintainers need to go through the merged PRs and update the file accordingly. 

I propose introducing a copy of the changelog file, which would simplify maintenance: each contributor could update their own contribution without opening a PR in the separate repository, which often does not happen. A message of a basic template with checklist could be prompted in the PR for itself.  

## PR template
To prompt these actions, just like for the changelog, [a template](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) with suggested checklist could be made in the PRs. 

## Repositories refactoring 
Actually, this is all part of a wider problem for which the `qutip-doc` does not get well updated in general. In Qutip 4.5.0 I added a lot of missing elements to `functions.rst` and `classes.rst` in the api-doc. Moreover, the User guide is seldom updated. Maybe, moving the `qutip-doc` repository in the main `qutip/qutip` repository could help in this regard.

Issue #1108, opened 2019-10-15
Possible solution to automating changelog/release notes, and adding PR templates


tehruhn:
Referencing this : https://github.com/qutip/qutip-doc/pull/70

1. Having manually compiled the changelog for v4.4.1 and v4.5.x today, I think that maybe we should try automating the release notes/changelogs with something like this : https://github.com/github-tools/github-release-notes
I'm attaching release notes generated by this for QuTiP.
[RELEASE_NOTES.txt](https://github.com/qutip/qutip/files/3728179/RELEASE_NOTES.txt)

2. We could also, in pursuit of having an automated release notes/changelog system start with making commits in more of a standard format, using some pre-decided keywords like [bugfix] or [improvement] in the commit message. I understand that this is a rather ambitious thing to say, but we can then use `grep` and `git log` itself to generate these in markdown. As a first step, I am proposing to use **Pull request /Issue** templates, something like this for instance : https://github.com/angular-translate/angular-translate/blob/master/.github/ISSUE_TEMPLATE.md
This way, all our Pull Requests and Issues will be homogeneously formatted.

hodgestar:
Superseded by #1894.

PR #1188, opened 2020-03-01
Create issue templates

BoxiLi:
I created two issue templates for bugs and features using the GitHub template in Settings->Options->Featuress. I adjusted the bug issue template for our use. The Feature request template stays almost the same as the default one.

The bug report issue will automatically get a label `BUG`, while the feature request issue will get a label `ENH`.

Anyone like to have a look and see if there is anything that should be added or modified?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,28 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: BUG
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Please provide a minimal working example. For instance:
+
+```python
+from qutip import identity
+# do something
+```
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**You Environment**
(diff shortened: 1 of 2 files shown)


BoxiLi:
I don't know what the `continuous-integration/travis-ci/push` failing means...

nathanshammah:
Wonderful! Thank you for picking up #1172 and #1108. Adding this to the modernization of docs project too. Travis CI seems to pass on merge, not on the push branch where the job is canceled.

nathanshammah:
I've also thought of making these ENH proposals I made as sort of QuTiP Enhancement Proposals (QEP), similarly to PEP and NEP. So far, keeping them as ENH and Q&A github issues may be fine. Or one could possibly use the Projects section of Github, too, but there is no possibility for others to chip in there.

nathanshammah:
What is strange about this Travis CI thing is that usually tests do not run on push, I think.

jakelishman:
I believe that the `push` action has run only because this PR has been created on a branch of the main `qutip` repository, rather than on a fork.

BoxiLi:
you are correct. It's on a branch of the main qutip repository because it is created on the GitHub Setting page. I guess it is cancelled because of this https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation. The two builds are created at the same time. So when a second build comes in, the first one got cancelled.

PR #1191, opened 2020-03-02
Patch4.5 fix1169

nathanshammah:
Adding a commit that should fix tests in 4.5.1.

### DIFF
--- a/qutip/qobjevo_codegen.py
+++ b/qutip/qobjevo_codegen.py
@@ -54,14 +54,15 @@ def _import_str(code, basefilename, obj_name, cythonfile=False):
     Import 'obj_name' defined in 'code'.
     Using a temporary file starting by 'basefilename'.
     """
-    filename = basefilename + str(hash(code))[1:6]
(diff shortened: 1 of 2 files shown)


qutip/qobjevo_codegen.py
```diff
@@ -54,14 +54,15 @@ def _import_str(code, basefilename, obj_name, cythonfile=False):
     Import 'obj_name' defined in 'code'.
     Using a temporary file starting by 'basefilename'.
     """
-    filename = basefilename + str(hash(code))[1:6]
+    filename = (basefilename + str(hash(code))[1:4] +
+                str(os.getpid()) + time.strftime("%M%S"))
```
jakelishman: Perhaps it would be simpler to use `tempfile.mkstemp(prefix=basefilename, suffix=suffix)`?  This explicitly avoids race conditions between all threads and it's part of the standard library.  While the chance of a collision in this current code is slight (you'd need a 1.5-byte hash collision of the code and for the same thread to call this function within a second), it's still not impossible, and all the external calls somewhat obfuscate what's happening.

Issue #1185, opened 2020-02-26
Cannot find qcirc.svg


nathanshammah:
In QuTiP 4.5.0 conda release on Mac, I get an error if I call a quantum circuit to print in jupyter notebook. However, the circuit is printed after the message. 

This did not happen with QuTiP 4.4.1. Imagemagick is installed. This issue may be related to the closed issues #1179 and #702.

```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
/miniconda3/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

/miniconda3/lib/python3.7/site-packages/qutip/qip/circuit.py in _repr_svg_(self)
   1117 
   1118     def _repr_svg_(self):
-> 1119         return _latex_compile(self.latex_code(), format="svg")
   1120 
   1121     @property

/miniconda3/lib/python3.7/site-packages/qutip/qip/circuit_latex.py in _latex_compile(code, filename, format)
     64     else:
     65         os.system("pdf2svg %s.pdf %s.svg" % (filename, filename))
---> 66         with open("%s.svg" % filename) as f:
     67             result = f.read()
     68 

FileNotFoundError: [Errno 2] No such file or directory: 'qcirc.svg'
```
<img width="1137" alt="Screenshot 2020-02-26 17 44 05" src="https://user-images.githubusercontent.com/14573436/75327357-9af58180-58bf-11ea-9989-6892f5b7af41.png">

jakelishman:
`pdf2svg` isn't part of ImageMagick, (at least on `brew`) is http://www.cityinthesky.co.uk/opensource/pdf2svg.  It feels like all of these system commands should be made a bit more robust.

In particular, the bare `os.system` is not the recommended way of calling out to system commands any more.  It's now preferable to use [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run).  Here, the bug is, I would assume, actually on the previous line:
```python
os.system('command_which_does_not_exist')
```
will return a non-zero error code which isn't checked for, but will not raise an exception.  

If Jupyter supports displaying embedded PDF files and that's all it's being used for, then it might be worth removing the conversion to SVG.  (pdf)LaTeX certainly supports embedded PDF files as graphics.  The fewer little dependencies we require, the better.

jakelishman:
[...]

PR #1194, opened 2020-03-02
Make LaTeX image conversion more resilient

jakelishman:
Notionally fixes #1179, #1185.  Converts the `os.system` calls to `subprocess.run` because they're now the recommended way of doing things.

I check for existence of the dependencies at import time, and emit a warning (`ImportWarning`) if they don't exist.  This is ignored by the typical warning filters, so it won't actually show up for most users.

If a dependency doesn't exist, then we don't define the relevant methods on the `QubitCircuit` class to avoid IPython introspecting the class object and thinking that it can make a PNG/SVG.

This may not be the best way of doing things, but in this delocalised setting I think it's easier to talk/discuss around a problem with an example in front of us.

In particular, I think raising `NotImplementedError` on failing to find `pdflatex` might not be the right call: if it doesn't exist, then the PNG and SVG methods of `QubitCircuit` are currently still created, and then IPython would try to call them and produce an error that isn't the fault of the user.  Also, maybe there's a nicer way to make sure that the user sees the "not found" warnings the first time that they do something which involves a conversion?  At the moment, the class just doesn't get the methods defined.

### DIFF
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -217,8 +217,8 @@ if _pdflatex is not None:
         # leftover files if something goes wrong (or we get a
         # KeyboardInterrupt) during conversion.
         previous_dir = os.getcwd()
-        try:
-            with tempfile.TemporaryDirectory() as temporary_dir:
+        with tempfile.TemporaryDirectory() as temporary_dir:
+            try:
                 os.chdir(temporary_dir)
                 with open(filename + ".tex", "w") as file:
                     file.write(_latex_template % (_qcircuit_latex_min, code))
@@ -237,8 +237,10 @@ if _pdflatex is not None:
                     raise ValueError("".join(["Unknown output format: '",
                                               file_type, "'."]))
                 out = CONVERTERS[file_type](filename)
-        finally:
-            os.chdir(previous_dir)
+            finally:
+                # Leave the temporary directory before it is removed (necessary
+                # on Windows, but it doesn't hurt on POSIX).
+                os.chdir(previous_dir)
         return out
 else:
     def image_from_latex(*args, **kwargs):
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -30,7 +30,12 @@
 #    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
+import collections
+import functools
 import os
+import shutil
+import subprocess
+import warnings
 
 _latex_template = r"""
 \documentclass{standalone}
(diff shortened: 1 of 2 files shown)


qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+                                 stdout=subprocess.DEVNULL,
+                                 stderr=subprocess.DEVNULL)
+_run_command.__doc__ = \
+    """
+    Run a command with stdout and stderr explicitly thrown away, raising
+    `subprocess.CalledProcessError` if the command returned a non-zero exit
+    code.
+    """
+
+if not _pdfcrop:
```
BoxiLi: Better use `if _pdfcrop is None:`?

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+    def _crop_pdf(_):
+        warnings.warn("Could not locate system 'pdfcrop':"
+                      " image output may have additional margins.")
+else:
+    def _crop_pdf(filename):
+        temporary = ".tmp." + filename
+        _run_command((_pdfcrop, filename, temporary))
+        os.rename(temporary, filename)
```
BoxiLi: I got an error there saying `[WinError 183] Cannot create a file when that file already exists: '.tmp.qcirc.pdf' -> 'qcirc.pdf'`. I need to add `_force_remove(filename)` before renaming to delete `qcirc.pdf`.

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+    def converter(file_stem):
+        in_file = file_stem + ".pdf"
+        out_file = file_stem + "." + possible.file_type
+        _run_command((which, *possible.arguments, in_file, out_file))
+        with open(out_file, mode) as file:
+            return file.read()
+    return converter
+
+
+for possible in _POSSIBLE_CONVERTERS:
```
BoxiLi: What about the name 
`possible` -> `converter_config`
`_POSSIBLE_CONVERTERS` -> `_CONVERTER_CONFIGS`

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+    if converter:
+        CONVERTERS[possible.file_type] = converter
     else:
-        os.system("pdf2svg %s.pdf %s.svg" % (filename, filename))
-        with open("%s.svg" % filename) as f:
-            result = f.read()
+        message = "".join(["Could not find system ", possible.dependency, ".",
+                           " Image conversion to '", possible.file_type, "'",
+                           " is unavailable."])
+        # ImportWarning is ignored by default, so the user won't actually see
```
BoxiLi: Actually this part of code will be executed even if the user does not intend to plot anything, for example it executes when `from qutip.qip.circuit import QubitCircuit` runs. Is Showing the warning here confusing?

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+        # this warning...
+        warnings.warn(message, category=ImportWarning)
+
 
-    return result
+if not _pdflatex:
```
BoxiLi: Is `if _pdflatex is None:` a better choice?

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+    def make_image(*args, **kwargs):
+        raise NotImplementedError("Could not find system 'pdflatex'.")
```
BoxiLi: Is this a `NotImplementedError`? An external dependency is missing, sounds to me more like a more general `RuntimeError`?

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+else:
+    def make_image(code, filename="qcirc", file_type="png"):
+        _force_remove(*[filename + suffix
+                        for suffix in [".tex", ".pdf", "." + file_type]])
+        with open(filename + ".tex", "w") as file:
+            file.write(_latex_template % (_qcircuit_latex_min, code))
+        _run_command((_pdflatex, '-interaction', 'batchmode', filename))
+        _force_remove(filename + ".aux", filename + ".log")
```
BoxiLi: I guess we can also remove the tex file here?

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+    if which is None:
+        return None
+    mode = "rb" if possible.binary else "r"
+
+    def converter(file_stem):
+        in_file = file_stem + ".pdf"
+        out_file = file_stem + "." + possible.file_type
+        _run_command((which, *possible.arguments, in_file, out_file))
+        with open(out_file, mode) as file:
+            return file.read()
```
BoxiLi: Can we delete the pdf and png/svg file after reading them? So there won't be any file left in the users' folder.

qutip/qip/circuit_latex.py
```diff
@@ -42,31 +47,121 @@
...
+        file_stem: str -> 'T,
+    where 'T is data in the format to be converted to.
+    """
+    which = _find_system_command(possible.executables)
+    if which is None:
+        return None
+    mode = "rb" if possible.binary else "r"
+
+    def converter(file_stem):
+        in_file = file_stem + ".pdf"
```
BoxiLi: Could you add NumPy-style documentation for `converter`? 
```
"""
It takes the stem string of the filename, transforms the corresponding pdf to png/svg and returns the plot.

Parameters
------------ 
...
Returns
--------
...
"""
```
 like other functions in QuTiP.

qutip/qip/circuit_latex.py
jakelishman: I would say `if not x` is pythonic and understandable - the alternative is a string with the system command name in it, and the python coding guidelines suggest using `if not x` to check for empty strings.

qutip/qip/circuit_latex.py
jakelishman: Thanks - I've not got a Windows box, and I forgot that it doesn't have the "move is overwrite" semantics that unix derivatives do.

qutip/qip/circuit_latex.py
jakelishman: Yeah, that's a better name.  I seem to remember I couldn't think of what to call those.

qutip/qip/circuit_latex.py
jakelishman: This was one of the reasons I put the PR up in the first place - I didn't think this was the best idea, but at the time it was what I thought of.

The reason it is here rather than elsewhere is so that the check for the converters only happens once, and the warning is emitter by the checkers.  It would be very convoluted logic to have the checks put the warnings somewhere, then have them later be emitted when the code is actually accessed.

The warning actually won't show at all to most people - `ImportWarning` is by default ignored by the filters (which is another problem in its own right...).

I wanted to have a conversation about how would be best to implement these warnings, since this clearly isn't perfect.

qutip/qip/circuit_latex.py
jakelishman: Same as before, really.

qutip/qip/circuit_latex.py
jakelishman: Yeah, you're probably right on that one.

qutip/qip/circuit_latex.py
jakelishman: Yeah, I'm not sure why I didn't before to be honest...

qutip/qip/circuit_latex.py
jakelishman: To be honest, it's probably just better to do this whole process in a temporary directory and clean it up at the end of the function.  Then nothing can possibly get left behind even on things like `KeyboardInterrupt`, and we don't need any of this `os.remove` stuff.

qutip/qip/circuit_latex.py
jakelishman: Sure.  Despite not being marked as such, the whole module is pretty internal and the original `_latex_compile` didn't have a docstring, but I'll add one.

jakelishman:
I'll make the changes probably tomorrow (UK time), so if you want to reply to any of my comments there's plenty of time.

qutip/qip/circuit_latex.py
jakelishman: Looking again, it's pretty questionable why I'm inconsistent within the same block of code - I do `if which is None` later on which is exactly the same case.

qutip/qip/circuit_latex.py
jakelishman: The other reason that I put the warning in at import time is because if the dependency is missing, then the `svg` and `png` methods of `QubitCircuit` aren't added (so IPython doesn't try and introspect them), and so it's not actually really possible to call the function without them.  The user would then never even know that they're missing functionality.

If there's a way to get IPython to _not_ use an existing `class.svg()` method, then we can emit the warning in `make_image` relatively neatly.  We'd add a sub-branch of the `if file_type not in _CONVERTERS` - we add a second check `if file_type in _MISSING_CONVERTERS` to determine whether it's one we could do if we have the dependency, or one that's clearly something totally screwed up in the code.

jakelishman:
To get the proper links on github:
- Fixes #1185 
- Closes #1179

BoxiLi:
I observed something strange after your update. On windows, with Jupyter Notebook, the kernel crashes every time when I try to plot the circuit. No error message. Does it run normally on your OS?

jakelishman:
Yeah, it runs completely normally for me... jupyter output

I can only think that there is some problem with the temporary directory on Windows, since the calls to all the external programmes are the functionally the same.

Do you get any error messages in the kernel log?  If not, can you try running it in raw `python` or `ipython` on the command line, and just run
```python
qc = QubitCircuit(2)
qc.add_gate("CNOT", 0, 1)
qc.png
qc.svg
```
and see if there are any error messages?

I kind of hope it's a mistake in my code somewhere, because if it is the `tempfile.TemporaryDirectory` on Windows then I'll have to re-think some parts of #1181 too.

jakelishman:
My computer is a Mac, by the way - I don't have direct access to a Windows machine at the moment.

BoxiLi:
Nothing at all. IPython also crushes. None of them leaves any message... Let me try to make some breakpoints to locate it.

jakelishman:
Raw python would be very interesting - if Jupyter crashes then I'd largely expect IPython to crash too, since it's just using that as a backend, but the base `python` executable is a completely different beast.

BoxiLi:
`python test_file.py` also gives nothing, if this is what you mean. :(

I located that the crush happens when it exits the `with tempfile.TemporaryDirectory() as temporary_dir` environment. So your suspicion might be correct...

BoxiLi:
I'm trying to make some simple example to see what went wrong.

jakelishman:
Ouch ouch ouch. I found [this bug](https://bugs.python.org/issue22107) in the Python tracker which suggests that there are some Windows-specific bugs in the handling of temporary files in Windows even in the standard library, but a further search for `TemporaryDirectory` doesn't show up the behaviour you're seeing.

The interpreter crashing without an exception sounds pretty bug-ish to me since we're not calling out to C code there, loathe as I am to try and pass the blame to the core interpreter.  What version of `python` are you on?  Does the bug persist if you use 3.6, 3.7 or 3.8?

BoxiLi:
It's probably caused by `os.chdir(temporary_dir)` inside `tempfile.TemporaryDirectory()`

jakelishman:
Using `tempfile.TemporaryDirectory()` as a context manager doesn't change the current directory, so we should be allowed to `cd` into the directory it returns...  The documentation doesn't mention anything about not being able to as far as I can see, and regardless, if it fails it should raise an `OSError` and not crash the interpreter.

I need to get a Windows box to try and debug this myself - it's difficult to know how to proceed without working out exactly what the problem is.  We could try using the `cwd` kwarg of `subprocess.run`, but if you're not allowed to change into a temporary directory on Windows in any manner, then that may also fail.  We don't want to run `pdflatex` in the user's directory, but "officially" we don't know exactly what files it will produce, so we can't reliably clean them all up.  That's also not to mention the potential race condition that we get if two processes are trying to draw circuits simultaneously.

BoxiLi:
Oh I have it, it does not switch back to the previous folder

jakelishman:
Oh damn!  That's especially annoying because I thought a lot about exactly that case, and then decided that putting it in the `finally` block was the best way to ensure it always ran.

I think I'll change it over to use `tempfile.mkdtemp()` and then put the `temporary_directory.cleanup()` after the `os.chdir()` in the `finally` block, and remove the context manager so we have better control over when the removal happens.  I'll have to modify a similar fixture in #1181 too.  Frustrating, since the context managers are cleaner, but I suppose if Windows is going to be silly about it, then we'll have to live with it.

Thanks for finding the cause!

jakelishman:
[mkdtemp.txt](https://github.com/qutip/qutip/files/4396123/mkdtemp.txt)

Could you try and apply this patch and see if it fixes the problem?  Just download it somewhere, then
```
git apply /path/to/mkdtemp.txt
```
should do it.

If so, I'll commit and push.

BoxiLi:
Yes, it works. No more crush.

Just curious, why not just add `os.chdir(previous_dir)` inside the `tempfile.TemporaryDirectory()` environment to switch back?

jakelishman:
If any element of the code within the context manager threw an exception, then the context manager would have the `__exit__` method called deleting the temporary directory before the `finally` statement ran, so we'd have the same bug.

jakelishman:
For example, note how `"exiting"` prints before `"finally"`:
Screenshot 2020-03-27 at 22 25 24

BoxiLi:
I see, thank for the explanation. What about put `try` inside?

```python
with tempfile.TemporaryDirectory() as temporary_dir:
    try:
                os.chdir(temporary_dir)
                ...
    finally:
                os.chdir(previous_dir)
```

jakelishman:
That's actually a very nice solution that I hadn't thought about.  I guess it's kind of weird that we have a `try/finally` directly inside a context manager, but effectively we're just monkey-patching the `__exit__` method so that `os.chdir(previous_dir)` gets called first, so it's always going to look a little funky.

There's our nice commit that swaps them over.

jakelishman:
For future posterity if anybody comes across this in the future: being in a directory that is deleted is explicitly not a problem on Linux because of how the kernel handles deletions, and doesn't appear to be a problem on Macs either (and I'm more used to Unix-y systems...).  Windows however throws a hissy fit and apparently crashes the Python interpreter.

BoxiLi:
I'm trying it now on a Windows without ImageMagick installed. However, it finds a wrong executable `convert`, which is a default command on Windows...... Is there a way to identify if the `convert` is the one we want? Otherwise, we can also just catch the CalledProcessError at line 148 and add an instruction.

jakelishman:
Yeah, this is just one of those silly things caused by ImageMagick choosing a _very_ generic name and then getting burned (hence the swap to `magick`...).

Probably the best thing to do is to try and run `convert` once at discovery and to see what happens.  If you run `convert -version` on an IM6 install, does it output a string that looks like
> Version: ImageMagick 6.blahblahblah
or something similar?

If so, try this patch: [check_convert.txt](https://github.com/qutip/qutip/files/4397692/check_convert.txt).  It rearranged the functions a little for a more logical order (I'll rebase the PR before merge), but the main change is that it now tests the output of `convert -version` to see if it contains the substring `imagemagick` (case insensitive).

BoxiLi:
> If you run convert -version on an IM6 install, does it output a string that looks like
> `Version: ImageMagick 6.blahblahblah`
> or something similar?

Yes, it does.

In the patch, I guess by 
`return "imagemagick" in proces.decode('utf-8').lower()`
you mean 
`return "imagemagick" in process.stdout.decode('utf-8').lower()`

Otherwise, I think it's all good!

jakelishman:
Oh man, yeah, that's embarrassing.  Yeah, that's exactly what I _meant_ to put.  I dunno how I screwed up the patch that badly!

I'll rebase these commits onto `master` and squash them into one or two nice modular commits.  I think a little bit of the development here is worth preserving in the commit history for if someone looks at it in the future (there's no guarantee that this PR discussion will _always_ be up, but commit messages are forever).

I'll force-push to this branch, so you'll have to force-pull a copy (or just delete your copy of this branch).  Not the best practice, but better for keeping the `master` commit history neat.

jakelishman:
There we go.  A gross history according to this Github PR, but a nice clean `git log`.

jakelishman:
It would be good to get this merged if it's approved.  Perhaps @nathanshammah wants to look too, since he'd also had bugs in #1185?

BoxiLi:
I'll merge it now. @nathanshammah can check later again when he gets time. But since you both use MacOS. It should be good.

PR #1195, opened 2020-03-04
Fix spin husimi/wigner functions

### LINKED ISSUE #1128: Husimi Q and Wigner spin functions don't give correct results
The Husimi Q function for spins gives incorrect output for the states |J, -J>, for all J.
Generating a spin-coherent state with some theta and phi, and then calculating the Husimi Q function for the same theta and phi should always give 1/pi, but this value tends toward 0 the closer to the bottom of the Bloch sphere you go.

A similar problem exists for the spin wigner function, where a the spin-coherent state |J, -J> has some negativity in the wigner function.

maij:
This will resolve #1128.

I've allowed the `spin_q_function` (Husimi function) and the `spin_wigner` functions to handle half-integer spin values.

I've also improved the docstrings.

### DIFF
--- a/qutip/wigner.py
+++ b/qutip/wigner.py
@@ -561,10 +561,10 @@ def qfunc(state, xvec, yvec, g=sqrt(2)):
         A state vector or density matrix.
 
     xvec : array_like
-        x-coordinates at which to calculate the Wigner function.
+        x-coordinates at which to calculate the Husimi-Q function.
 
     yvec : array_like
-        y-coordinates at which to calculate the Wigner function.
+        y-coordinates at which to calculate the Husimi-Q function.
 
     g : float
         Scaling factor for `a = 0.5 * g * (x + iy)`, default `g = sqrt(2)`.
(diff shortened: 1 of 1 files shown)


nathanshammah:
Thank you @maij, looks good to me. Good catch on the docstring.

hodgestar:
@maij @nathanshammah Do you remember how you tested `spin_q_function` when these changes were made? The changes were merged, but I only get the expected results if I invert the direction of theta:
```python
    psi = qutip.spin_coherent(spin, theta, phi)
    [[husimi_q]], _, _ = qutip.spin_q_function(psi, np.pi - theta, phi)
    # assert husimi_q == 1 / np.pi
```
Without this bug persists for me (regardless of whether the spin is integer or half integer).

I have a fix that works in my test cases, but I was wondering about the history of the changes.

maij:
@hodgestar Good catch, I was testing this by plotting 2D images over theta and phi, but I was a bit careless because the top of the plot is theta=pi (for theta on the y-axis) which isn't the top of the Bloch sphere, but the bottom. So the result was what I expected, but for the wrong reason.

Thankfully, this looks like an easy fix.

maij:
@hodgestar I think I have found a fix, the problems were two-fold:
- The addition/subtraction for the cos/sin(theta/2) terms were swapped, which gives you a flipped result (high Q value on opposite side of sphere)
- The phase factor for the diagonal terms of the density matrix were swapped, which gave you flipped result in the Y axis (high Q value for -Y if you give it +Y state)

I can submit a new fix for this alongside some test code that does some basic verification (sorry -- I'm not sure what would be a good set of tests, the ones I have come up with are certainly not exhaustive.)

hodgestar:
@maij Thank you for investigating and opening #1632. I will review & continue the discussion there.

PR #1198, opened 2020-03-07
Create PR and issue templates

BoxiLi:
**Description**
I created three issue templates and one PR template and tested on my own fork. You can have a look and also test it at https://github.com/BoxiLi/qutip. Try to open a PR or issue there, you should see how the template works.

As an example, I opened this PR following the template.

The bug report issue will automatically get a label BUG, while the feature request issue will get a label ENH.

Free feel to give suggestions.

**Related issues or PRs**
This PR addresses #1172 #1108 but does not totally solve the issues. It is a replacement of #1188, and therefore closes #1188

**change log**
Added PR and issue templates


.github/pull_request_template.md
```diff
@@ -0,0 +1,20 @@
+**Checklist**
+Please make sure you have finished the following tasks before opening the PR.
+
+- [ ] Please read [Contributing to QuTiP Development](https://github.com/qutip/qutip-doc/blob/master/qutip_dev_contrib.md)
+- [ ] Contributions to qutip should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/).
+You can use [pycodestyle](http://pycodestyle.pycqa.org/en/latest/index.html) to check your code automatically
+
+Delete this section after you have complete all the tasks.
```
nathanshammah: have completeD

.github/pull_request_template.md
```diff
@@ -0,0 +1,20 @@
...
+
+**Description**
+Describe here the proposed change.
+
+**Related issues or PRs**
+Please mention the related issues or PRs here. If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id, e.g. fix #1184
+
+**Changelog**
+Give a short description of the PR in a few words. This will be shown in the QuTiP change log after the PR gets merged.
```
nathanshammah: If we want them to update the Changelog, this should be more explicit. I feel like the Changelog in `qutip-doc` puts a lot of strain for people to comply, as they have to fork and open a PR also there. My proposal is to add a `changelog` file also in `qutip/qutip` that can be copy pasted every time there is a new release also in `qutip-doc`.  What do you think? Can you add such file? It could be named the same or something like `changelog-PRs`, I don't know.

nathanshammah:
I am very favorable for this as it can really help with sporadic contributors and save up time of maintainers. At the same time, I don't want it to make PR/Issues interactions too stiff. So I would be happier if the core devs would chip in and tell their opinion. Cc @quantshah @ajgpitch  @Ericgig.

Ericgig:
I only have doubt about the 'other' form. Does that mean that issues cannot be open without using a template? Having the changelog in the PR is the best I believe. The file in qutip-doc can be undated at each release. A file in qutip could create a lot of conflict since each PR would have to add something at the end of 1 file and a lot of PR would be waiting on it. The same way we have good PR that are only waiting for a test to be added before merging. In the pull request template, I would mention the need for tests.

nathanshammah:
I agree with @Ericgig on the possibility of having no template and adding the suggestions for tests in the template. 

I understand now that the changelog in qutip/qutip was a bad idea for the reasons you mention.
I still think it would be nice to somehow automate the way the changelog is compiled (right now by whoever is in charge of the release). Is there a simple and smart way to do this? Otherwise we could just look up the changelog section in each PR as per this template.

BoxiLi:
You can still open an empty issue or PR without the template, although the button for that is a bit smaller and is at the end of the template list. I set this up in my fork https://github.com/BoxiLi/qutip, you can try open one there. The behaviour should be exactly the same for the QuTiP main repo after this PR is merged. Let me know if you have suggestions on a better name as "others".

Sure I can add a suggestion for tests.

A proposal for automatization is in https://github.com/qutip/qutip/issues/1108. I'm not yet sure if that can be directly applied to this template.

PR #1211, opened 2020-03-23
Fix add_1q_gates + test

### LINKED ISSUE #1210: QubitCircuit.add_1q_gate missing last qubit
Expected behavior: add_1q_gate applies the specified gate to the whole register.
Actual behavior: add_1q_gate sets end to N-1, and iterates with range(start, end) to apply the single qubit rotations. Since range() does not include the stop, the last qubit is not affected.

sarsid:
In response to issue #1210 , added a quick fix for the add_1q_gates so that it now includes the "end" qubit (specified by the end argument). I also added a test in test_add_gates in test_qubitcircuit.py to cover this issue. The function now applies the 1 qubit gate to every qubit in [start, end] (0-indexed), start and end inclusive. This should resolve #1210

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -317,7 +317,9 @@ class QubitCircuit(object):
         else:
             if end is None:
                 end = self.N - 1
-            for i in range(start, end):
+            if (end >= self.N):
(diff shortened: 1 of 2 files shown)


BoxiLi:
Looks good to me too. Thanks @sarsid

PR #1209, opened 2020-03-21
Add more single-qubit gates to QubitCircuit

mlaguna10:
1. Gate T and S were defined at qutip.qip.operation. Since Pauli-X, Pauli-Y and Pauli-Z exist in qutip.operators, they were imported as functions into qutip.qip.operation. All functions were tested. 
2. Each of the new gates was added to the circuit.py file. 
3. Not apply.

PR #1212, opened 2020-03-23
More single qubit gates

### LINKED ISSUE #1208: Add more single-qubit gates to QubitCircuit
QuTiP `qip` module currently supports six different single-qubits gate, including rx, ry, yz, http://qutip.org/docs/latest/guide/guide-qip.html. However, the commonly used Pauli X, Y, Z  gates, T gate and S gate are not there. Although in principle they can all be realized by the existing gate set. It would still be good to have them supported in `qutip.qip.QubitCircuit`.

- Define those gates in `qutip.qip.operation`. If they already exist in `qutip.operators`, import them as functions such as `x_gate`, `y_gate`, `z_gate`, `s_gate`, `t_gate`
- Add the gates to `quip.qip.QubitCircuit`
- Add them also to the LaTeX code if applicable.

MartinSandeCosta:
closes #1208

### DIFF
--- a/qutip/qip/operations/gates.py
+++ b/qutip/qip/operations/gates.py
@@ -159,7 +159,7 @@ def cy_gate(N=None, control=0, target=1):
                      [0, 1, 0, 0],
                      [0, 0, 0, -1j],
                      [0, 0, 1j, 0]],
-                    )
+                    dims=[[2, 2], [2, 2]])
 
 
 def z_gate(N=None, target=0):
@@ -198,7 +198,7 @@ def cz_gate(N=None, control=0, target=1):
                      [0, 1, 0, 0],
                      [0, 0, 1, 0],
                      [0, 0, 0, -1j]],
-                    )
+                    dims=[[2, 2], [2, 2]])
 
 
 def s_gate(N=None, target=0):
--- a/qutip/qip/operations/gates.py
+++ b/qutip/qip/operations/gates.py
@@ -106,6 +106,178 @@ def rz(phi, N=None, target=0):
                      [0, np.exp(1j * phi / 2)]])
 
 
+def x_gate(N=None, target=0):
+    """Single-qubit rotation through pi radians around the x-axis.
+
+    Returns
+    -------
+    result : qobj
+        Quantum object for operator describing the rotation.
+
+    """
+    if N is not None:
+        return gate_expand_1toN(x_gate(), N, target)
+    else:
+        return Qobj([[0, 1],
+                     [1, 0]],
+                    dims=[[2, 2], [2, 2]])
+
+
+def y_gate(N=None, target=0):
+    """Single-qubit rotation through pi radians around the y-axis.
+
+    Returns
+    -------
+    result : qobj
+        Quantum object for operator describing the rotation.
+
+    """
+    if N is not None:
+        return gate_expand_1toN(y_gate(), N, target)
+    else:
+        return Qobj([[0, -1j],
+                     [1j, 0]],
+                    dims=[[2, 2], [2, 2]])
+
+
+def cy_gate(N=None, control=0, target=1):
+    """Controlled Y gate.
+
+    Returns
+    -------
+    result : qobj
+        Quantum object for operator describing the rotation.
+
+    """
+    if (control == 1 and target == 0) and N is None:
+        N = 2
+
+    if N is not None:
+        return gate_expand_2toN(cy_gate(), N, control, target)
+    else:
+        return Qobj([[1, 0, 0, 0],
+                     [0, 1, 0, 0],
+                     [0, 0, 0, -1j],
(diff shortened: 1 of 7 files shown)


MartinSandeCosta:
Commit #bea29a1 breaks some tests since it refactors some attribute names to follow PEP8 snake_case standard (i.e., qubit.qit.circuit.N ->qubit.qit.circuit.n_qubits )

MartinSandeCosta:
Should a new issue be created to replace deprecated method unittest.TestCase.assert_() with newer assertEqual/NotEqual?

jakelishman:
Not to discourage you from contributing, but there's already been some movement on #1208 in #1209.

About tests: `qutip` uses `pytest` now, so the preferred method is actually just a bare `assert`, because `pytest` does introspection on the variables on an assertion failure.  But yes, any changes to that should be in a new PR, because it's a different logical unit of work.  There's some movement on that too (e.g. #1181).

About pep8 changes: I'm not a maintainer, but I'd be much more conservative about changing actual public APIs in the name of pep8.  It's not a good idea to change public APIs that work fine because then the users, who are the reason we're writing the code in the first place, have to adjust, and the pep8 spec explicitly warns against breaking backwards compatibility to match the pep.  `N` is fairly consistent within `qutip` for the number of qubits, and the `U` in `U_list` refers to unitary matrices, which are conventionally a capital `U` in mathematical writing.

MartinSandeCosta:
*Hi Jake,*

It is not discouraging at all, quite the contrary. Thank you very much for
looking at it and taking the time to let me know your thoughts. I will take
peek on  #1181 <https://github.com/qutip/qutip/pull/1181> and provide any
help I can.

Regarding the PEP changes: I am extremely inexperienced with this project
and contrasted my ideas with others such as Qskit and projectQ before
proceeding. I also find that this refactoring can be excessively thorough
and not needed since it breaks backwards compatibility.
On the other side, I knew these changes would not go through to master but
they can serve some purpose as to help me get a better understanding of the
library as well as get in touch with the community. It also helped me on
how to approach potential improvements for the tool.

*Thanks once again,*
*Martín*

qutip/qip/operations/gates.py
```diff
@@ -106,6 +107,178 @@ def rz(phi, N=None, target=0):
                      [0, np.exp(1j * phi / 2)]])
 
 
+def x_gate(N=None, target=0):
+    """Single-qubit rotation through pi radians around the x-axis.
+
+    Returns
+    -------
+    result : qobj
```
nathanshammah: in general "qobj"---> "Qobj" in docstring. Even better would be ":class: `qutip.Qobj`" as that should get picked up by Sphinx in the docs and make it clieckable.

qutip/tests/test_qubitcircuit.py
```diff
@@ -32,9 +32,9 @@
 ###############################################################################
 
 import numpy as np
-from numpy.testing import assert_, assert_allclose, run_module_suite
+from numpy.testing import assert_raises, assert_, assert_allclose, run_module_suite
 from qutip.qip.operations.gates import (
-    gate_sequence_product, rx, expand_operator)
+    gate_sequence_product, rx)
```
nathanshammah: why is `expand_operator` removed? Cc @BoxiLi.

nathanshammah:
Ah ok, I did not catch what @jakelishman noticed on similarly duplicate stuff in other PRs. Better to address the early ones first.

qutip/tests/test_qubitcircuit.py
MartinSandeCosta: Removed unused import

ajgpitch:
Firstly, just to underline @jakelishman comment. We *cannot* change the public interface unless we make a new major release. I also don't like the `.N` property, but it snuck in before my time. We could look to change this in v.5, but not before. 

It seems this PR does contain some enhancements that are not part of #1209. However, @mlaguna10 did seem to make quite clear that they were already working on the issue. I believe #1209 is ready to merge, but we'll wait for @BoxiLi to confirm. 

So what I suggest is that we merge #1209. @MartinSandeCosta  can then merge the master into this PR and see what is left that adds to enhancement request outlined #1208. Assuming (with good reason) that there are valuable additions, then we will merge this too.

Thanks everyone for contributing.

BoxiLi:
Hi @MartinSandeCosta, I have merged the PR of @mlaguna10. You can see now there is a few conflicts. Could you make a fresh PR including your changes based on the current master branch? 

It could include:
- The controlled version of the gates
- Changes in QubitCircuit.propagator()
- Tests
and another enhancement you have related to #1208 

We shouldn't include the change of the attribution name `N` and `U_list` in this PR, as @jakelishman and  @ajgpitch explained.

MartinSandeCosta:
Sure thing!

Issue #1202, opened 2020-03-13
spawn-multiprocessing and frozen mcsolve


goerz:
Different platforms have some subtle differences in how multiprocessing works: On linux, processes "fork". My basic understanding is that sub-processes are created as copies of the parent process, thus inheriting their entire state (global variables/functions). Windows, and [macOS with Python >= 3.8](https://twitter.com/raymondh/status/1180209729609420800) uses "spawn". There, I believe subprocesses start from blank slate, and relevant global state is injected via IPC. Sadly, limitations of the pickle protocol can cause "spawn"-based multiprocessing to have problems.

Jupyter notebooks are affected by this: you cannot use multiprocessing-map using functions defined within the notebook. 

Within qutip, this definitely affects the `parallel_map` function: It currently will cause freezes on Windows (See https://qucontrol.github.io/krotov/v1.0.0/notebooks/08_example_ensemble.html). With Python 3.8, macOS is also affected by this. See https://github.com/qutip/qutip-notebooks/issues/100

One possible workaround is to use a third-party replacement for multiprocessing. An work-in-progress implementation of this is in https://github.com/qutip/qutip/pull/1092, using [loky](https://loky.readthedocs.io).

In #1197, we also identified a freeze in `mcsolve` when running tests on macOS/Python 3.8. This is suspected to be an issue with spawn-based multiprocessing, although we haven't been able to determine this with complete confidence.

If the freeze is indeed caused by spawn-multiprocessing, the problem *should* also show up on Windows. Is anyone running Windows able to reproduce this?

nonhermitian:
This may help https://github.com/Qiskit/qiskit-terra/pull/3491

Ericgig:
I can get the same error on linux with:
```
from qutip.testing import run
import multiprocessing as mp
mp.set_start_method('spawn')
run()
```
It hangs on qutip/tests/test_mcsolve.py::test_MCTDFunc.
So maybe, set_start_method('fork') could probably fix the problem on mac.
Also this hang, both at console and jupyter:
```
import multiprocessing as mp
mp.set_start_method('spawn')

def f(i):
    return i+1

p = mp.Pool()
p.map(f, [1,2,3,4])
```

berryboy2012:
> 
> 
[...]

PR #1214, opened 2020-03-25
Udated tests to 3.8

Ericgig:
Moved the openmp, mkl, no cython tests to python 3.8
Also added (temporaly?) a mac test on python 3.8.

### DIFF
--- a/qutip/parallel.py
+++ b/qutip/parallel.py
@@ -45,11 +45,13 @@ import signal
 import qutip.settings as qset
 from qutip.ui.progressbar import BaseProgressBar, TextProgressBar
 
+
 if sys.platform == 'darwin':
     # from 3.8 mp on mac use 'spawn', suposely better than fork, but freeze
     # in our tests.
     try:
         set_start_method('fork')
+
 
 def _task_wrapper(args):
     try:
--- a/qutip/parallel.py
+++ b/qutip/parallel.py
@@ -37,7 +37,7 @@ mappings, using the builtin Python module multiprocessing.
 __all__ = ['parfor', 'parallel_map', 'serial_map']
 
 from scipy import array
-from multiprocessing import Pool, set_start_method
+import multiprocessing
 from functools import partial
 import os
 import sys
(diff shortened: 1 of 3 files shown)


Ericgig:
test passes on mac with python 3.8 withour hanging on parallel mcsolve by forcing "fork" multiprocessing (#1202). Still on xcode 10. segfault probably not fixed.

qutip/parallel.py
```diff
@@ -46,6 +46,15 @@
 from qutip.ui.progressbar import BaseProgressBar, TextProgressBar
 
 
+if sys.platform == 'darwin':
+    # from 3.8 mp on mac use 'spawn', suposely better than fork, but freeze
+    # in our tests.
+    try:
+        set_start_method('fork')
```
jakelishman, line 53: I believe that `multiprocessing.set_start_method` [can only be called once](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method) during any program's execution.  If we do this in `qutip`, then a user cannot call it themselves either before or after import.

If you remove the `except` catch all and try
```python
import multiprocessing
multiprocessing.set_start_method('spawn')
import qutip
```
or don't remove it and try to set the start method after the `qutip` import, you'll see the `RuntimeError`.

Instead we can do something like
```python
import multiprocessing
if sys.platform == 'darwin':
    Pool = multiprocessing.get_context('fork').Pool
else:
    Pool = multiprocessing.Pool
```
which limits the effects, and also guarantees that no matter what the caller set the start method to, we'll be using the one that works for us.

Ericgig: Well seen, I did not know of `get_context`. `set_start_method` did work, but felt a little bit strong-handed, changed it.

jakelishman: Multiprocessing is such a nightmare at the best of times...  I didn't know about it until recently, and I've kept my local python on 3.7 because I couldn't figure out why some of my personal multiprocessing code was no longer working on macOS python3.8.  Our lives would all be so much easier if pickling could support local function definitions or lambdas.

Ericgig:
Finally I settled on 5 tests sets: 3 linux and 2 MacOS. 
The test on python 3.6 use no-cython and openblas.
mkl and openmp run together on python 3.7 and python 3.8 use conda's default.
On Mac, tests use openblas, no openmp and cython, repeated for python 3.7 and 3.8.

nathanshammah:
Is this going into 4.5.1? Btw, we could use Github milestones more effectively.

Ericgig:
It would be nice since it makes the python 3.8 test not freeze on mac. But if you don't thinks it's a good PR we can wait. It is a PR that does 2 things. Changes which tests are run and fix a bug that made one of the new tests fails. I can split it into 2 PRs if you prefer.

jakelishman:
LGTM.

Eric already made the changes I'd suggesting.  This will cause another merge conflict in #1181, but that's not a problem at all (that PR will make this same logical change anyway).

PR #1215, opened 2020-03-25
4.5.1v2

Ericgig:
Update to the original 4.5.1 which is already a month old.

New inclusions: 

Better error message for failed string coefficient compilation.
Run the python 3.8 test in travis. (by Simon Cross )
Fix clebsch function for half-integer (by Thomas Walker)
Deprecate graph function (by Jake Lishman)
Fix travis warning (by Ivan Carvalho)
Fix spin husimi/wigner functions (by maij)

I would like the mac stuff to be fixed and moving the mac test to 3.8, but that will take some time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # After changing this file, check it on:
 #   http://lint.travis-ci.org/
 
-stage_generic_linux: &stage_generic_linux
+_stage_generic_linux: &stage_generic_linux
   os: linux
   language: python
   # change dir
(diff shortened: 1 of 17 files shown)


Ericgig:
@ajgpitch @nathanshammah 
I would like to make 4.5.1 from this. All test passes, there is only windows and mac 3.8 to check.

Ericgig:
@nathanshammah 
I am keeping the release up to date adding merged PR to this one.
This version has most bug fix PR since 4.5.0 and some testing/travis one.
Idealy I would also like to add 2 more waiting approval:
#1219 : Bug on windows using int32
#1215: parallel_map run on mac in python 3.8.
I did not add the PR about gates, they are new feature.
The latex one (#1194 ), I am not sure, so I left it for now.
To be really happy about the release I would like the segfault to be fixed, but I don't have the error and travis test also does not. So I am blind for that one. It will have to wait.

nathanshammah:
So, I'd need to first review 1215 and 1219 and then this one?

Ericgig:
Ideally, or if you judge adding them is moving the goal, we can go with just this one.
Those 2 are ready to merge and fix bugs that have issues raised by users.

Ericgig:
I saw  you merged #1219 to master but merged this one also.  Should I also push it to 4.5.1 (another PR) or are we not including #1219 and #1215 in 4.5.1?
My process was to cherry-pick commits merged to master to this PR to then when this was ready to be 4.5.1 merge it to the branch 4.5.1 and release it.

PR #1216, opened 2020-03-25
perf: force bloch view update on render

### LINKED ISSUE #1106: Updating Bloch sphere attributes seems impossible
Hi again,

In the documents [there is an example](http://qutip.org/docs/4.1/guide/guide-bloch.html#animating-with-the-bloch-sphere) of how to do an animation of a Bloch sphere by saving the frames of the movie using a for loop like so:

```python
b = Bloch()
b.vector_color = ['r']
b.view = [-40,30]
for i in range(len(sx)):
    b.clear()
    b.add_vectors([np.sin(theta),0,np.cos(theta)])
    b.add_points([sx[:i+1],sy[:i+1],sz[:i+1]])
    b.save(dirc='temp')
```

With `b.clear()` we erase the points and vectors in the image, which is nice for updating the data, but What if I want to update also the attributes of the Bloch sphere?

To give an example let's say that I want the viewing angle of the Bloch sphere to smoothly change throughout the animation. I can define a function like so:

```python
def bloch_camera_path(t, dt, azim_0, azim_f, elev_0, elev_f):
    azim = ((azim_f-azim_0)/dt)*t+azim_0
    elev = ((elev_f-elev_0)/dt)*t+elev_0
    return [azim, elev]
```

This function changes the viewing angle across time from the initial `[azim_0, elev_0]` configuration to the final `[azim_f, elev_f]` configuration in a linear manner.

So I thought that by doing

```python
b = Bloch()
b.vector_color = ['r']
for i in range(len(t)):
    b.clear()
    b.view = bloch_camera_path(t[i], t[-1]-t[0], -80, -70, 45, 40)
    b.add_vectors([np.sin(theta),0,np.cos(theta)])
    b.add_points([sx[:i+1],sy[:i+1],sz[:i+1]])
    b.save(dirc='temp')
```
My animation would had that change in position of the "camera", but the result is just as if `b.view` wouldn't update.

The only way I can make this work is by making a new bloch sphere on each iteration and defining all the attributes again and again (which I find highly inconvenient in terms of speed), like so;

```python
import os

for i in range(len(t)):
    b = Bloch()
    b.vector_color = ['r']
    b.view = bloch_camera_path(t[i], t[-1]-t[0], -80, -70, 45, 40)
    b.add_vectors([np.sin(theta),0,np.cos(theta)])
    b.add_points([sx[:i+1],sy[:i+1],sz[:i+1]])
    b.save(os.getcwd()+'/temp/bloch_'+str(i)+'.png', dirc='temp')
    b.clear()
```

Could Qutip be modified so that attributes (such as viewing angle or the color of vectors) can be updated in the way I showed before? So that we don't have to create a new bloch sphere on each iteration of the animation?

MartinSandeCosta:
closes #1106 

Adds update_view method to change view variable. Thus abstracting from attribute naming.

Also forces view update in render.

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -390,6 +390,30 @@ class Bloch():
         else:
             self.vectors.append(vectors)
 
+    def update_view(self, azim=-60, elev=30):
+        """Update view with new azim and elev
+
+        Parameters
+        ----------
+        azim : float
+            Azimuthal viewing angle, defaults to -60.
+        elev: float
+            Elevation viewing angle, defaults to 30.
+
+        """
+        self.view = [azim, elev]
+
+    def _force_axe_update(self):
+        """Forces axe update with view
+
+        Parameters
+        ----------
+        void
+
+        """
+        self.axes.azim = self.view[0]
+        self.axes.elev = self.view[1]
+
     def add_annotation(self, state_or_vector, text, **kwargs):
         """Add a text or LaTeX annotation to Bloch sphere,
         parametrized by a qubit state or a vector.
@@ -454,6 +478,8 @@ class Bloch():
 
         if not axes:
             self.axes = Axes3D(self.fig, azim=self.view[0], elev=self.view[1])
+        else:
+            self._force_axe_update()
 
         if self.background:
             self.axes.clear()


MartinSandeCosta:
Added new notebook [notebook](https://github.com/MartinSandeCosta/qutip-notebooks/blob/bloch-sphere-animation/examples/Animating-with-the-Bloch-sphere.ipynb)

MartinSandeCosta:
Closed due to stagnation.

PR #1221, opened 2020-03-28
Remove executable permission from text files

jakelishman:
Remove superfluous executable permission bits being set on a few Python files.  A very minor change, though it is that rarest of things - a commit with +0/-0.

(deleted changelog section because this probably shouldn't go in - it's just git repo bookkeeping.)


jakelishman:
Something is way off about that test failure again...  Is it possible Travis isn't cleaning up properly?  I can't reproduce the failure (it's also clearly nothing to do with this change) and the error message doesn't match up.  If anything it looks like the error message is coming from a hash collision inside QobjEvo _maybe_?

BoxiLi:
Do you want me to restart the Travis tests? Or should we look into these random failing tests a bit before? Maybe @Ericgig knows something about it.

jakelishman:
You could try restarting them if you like, but it's been an issue hitting a few PRs now - I've had random test failures on two or three commits in #1181, now here, and I think one of your recent ones did too?

The oddest thing is that I can't reproduce the failures on my machine at all, so it's difficult to find out what's going on. I may try and make a fake PR which disables all but the nocython tests to see if I can pin it down (or if Eric knows). I ran the tests like 30 times back to back and no failure.

BoxiLi:
Yes, I also met one. It must be somehow random because I only changed the markdown templates. Let me wait for @Ericgig to have a look before restarting it.

Ericgig:
I can't reproduce it, but a few points are strange. 
There is a bug in QobjEvo related to correlation that I corrected in my solver branch but forgot to push to master. (https://github.com/qutip/qutip/blob/09913b461371ea25a2445212385f504ab963e13d/qutip/qobjevo.py#L1348)
Also, this case does not need to be compiled on the fly, but it is.
I will look at it further later. Please tag me if you see something similar in other commit.

PR #1224, opened 2020-04-02
Single qbtgates changes

### LINKED ISSUE #1225: More single qubit changes
Include the improvements made on #1208. These changes base from #1209 and #1212: 
- The controlled version of the gates
- Changes in QubitCircuit.propagator()
- Tests

MartinSandeCosta:
**Description**
- The controlled version of the single-qubit gates
- Changes in QubitCircuit.propagator()
- Tests

**Related issues or PRs**
fixes #1225 

**Changelog**
Added controlled version of the single-qubit gates
Added controlled single-qubit gates to Qubit.Qip.Circuit.resolve_gates()
Added controlled single-qubit gates to Qubit.Qip.Circuit.propagator()
Added controlled single-qubit gates to Qubit.Qip.Circuit.latex_code()
Added controlled single-qubit gates to Qubit.Qip.Circuit.add_1q_gate()
Added tests for controlled single-qubit gates in Qubit.Qip.Operations.Gates and Qubit.Qip.Circuit

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -354,22 +354,22 @@ class QubitCircuit:
                                        arg_value=arg_value,
                                        arg_label=arg_label))
 
-    def add_circuit(self, qubit_circuit, start=0):
+    def add_circuit(self, qc, start=0):
         """
         Adds a block of a qubit circuit to the main circuit.
         Globalphase gates are not added.
 
         Parameters
         ----------
-        qubit_circuit : QubitCircuit
+        qc : QubitCircuit
             The circuit block to be added to the main circuit.
         start : int
             The qubit on which the first gate is applied.
         """
-        if self.N - start < qubit_circuit.N:
+        if self.N - start < qc.N:
             raise NotImplementedError("Targets exceed number of qubits.")
 
-        for gate in qubit_circuit.gates:
+        for gate in qc.gates:
             if gate.name in ["RX", "RY", "RZ", "SNOT", "SQRTNOT", "PHASEGATE"]:
                 self.add_gate(gate.name, gate.targets[0] + start, None,
                               gate.arg_value, gate.arg_label)
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -38,8 +38,14 @@ import inspect
 import numpy as np
 
 from qutip.qip import circuit_latex as _latex
-from qutip.qip.operations.gates import *
-from qutip.qip.qubits import qubit_states
+from qutip.qip.operations.gates import (rx, ry, rz, sqrtnot, snot, phasegate,
+                                        x_gate, y_gate, z_gate, cy_gate,
+                                        cz_gate, s_gate, t_gate, cs_gate,
+                                        ct_gate, cphase, cnot, csign,
+                                        berkeley, swapalpha, swap, iswap,
+                                        sqrtswap, sqrtiswap, fredkin,
+                                        toffoli, controlled_gate, globalphase,
+                                        expand_operator)
 
 try:
     from IPython.display import Image as DisplayImage, SVG as DisplaySVG
@@ -55,7 +61,7 @@ except ImportError:
 __all__ = ['Gate', 'QubitCircuit']
 
 
-class Gate(object):
+class Gate:
     """
     Representation of a quantum gate, with its required parametrs, and target
     and control qubits.
@@ -107,18 +113,20 @@ class Gate(object):
             if self.controls is not None:
                 raise ValueError("Gate %s cannot have a control" % name)
 
-        elif name in ["CNOT", "CSIGN", "CRX", "CRY", "CRZ"]:
+        elif name in ["CNOT", "CSIGN", "CRX", "CRY", "CRZ", "CY", "CZ",
+                      "CS", "CT"]:
             if self.targets is None or len(self.targets) != 1:
                 raise ValueError("Gate %s requires one target" % name)
             if self.controls is None or len(self.controls) != 1:
                 raise ValueError("Gate %s requires one control" % name)
 
-        elif name in ["SNOT", "RX", "RY", "RZ", "PHASEGATE"]:
+        elif name in ["SNOT", "RX", "RY", "RZ", "PHASEGATE", "X", "Y",
+                      "Z", "S", "T"]:
             if self.controls is not None:
                 raise ValueError("Gate %s does not take controls" % name)
 
         elif name in ["RX", "RY", "RZ", "CPHASE", "SWAPalpha", "PHASEGATE",
-                            "GLOBALPHASE", "CRX", "CRY", "CRZ"]:
+                      "GLOBALPHASE", "CRX", "CRY", "CRZ"]:
             if arg_value is None:
                 raise ValueError("Gate %s requires an argument value" % name)
 
@@ -130,10 +138,10 @@ class Gate(object):
         self.arg_label = arg_label
 
     def __str__(self):
-        s = "Gate(%s, targets=%s, controls=%s)" % (self.name,
-                                                   self.targets,
-                                                   self.controls)
-        return s
+        string_name = "Gate(%s, targets=%s, controls=%s)" % (self.name,
+                                                             self.targets,
+                                                             self.controls)
+        return string_name
 
     def __repr__(self):
         return str(self)
@@ -142,12 +150,16 @@ class Gate(object):
         return str(self)
 
 
-_gate_name_to_label = {
+_GATE_NAME_TO_LABEL = {
     'X': r'X',
     'Y': r'Y',
+    'CY': r'C_y',
     'Z': r'Z',
+    'CZ': r'C_z',
     'S': r'S',
+    'CS': r'C_s',
     'T': r'T',
+    'CT': r'C_t',
     'RX': r'R_x',
     'RY': r'R_y',
     'RZ': r'R_z',
@@ -174,19 +186,18 @@ _gate_name_to_label = {
 
 def _gate_label(name, arg_label):
 
-    if name in _gate_name_to_label:
-        gate_label = _gate_name_to_label[name]
+    if name in _GATE_NAME_TO_LABEL:
+        gate_label = _GATE_NAME_TO_LABEL[name]
     else:
         warnings.warn("Unknown gate %s" % name)
         gate_label = name
 
     if arg_label:
         return r'%s(%s)' % (gate_label, arg_label)
-    else:
-        return r'%s' % gate_label
+    return r'%s' % gate_label
 
 
-class QubitCircuit(object):
+class QubitCircuit:
     """
     Representation of a quantum program/algorithm, maintaining a sequence
     of gates.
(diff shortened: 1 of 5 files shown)


MartinSandeCosta:
[Notebook](https://github.com/MartinSandeCosta/qutip-notebooks/blob/795652a0bc1e1b2c0c0ab67527cc97212523639d/development/development-single-qubit-gates.ipynb) added

MartinSandeCosta:
These tests seem a little too scant in my opinion, I am not fully satisfied with the final result. They do not provide full coverage of these new functionalities.

I am thinking of refactoring the whole thing and start all over again.

ajgpitch:
@MartinSandeCosta well, it is most important that you are happy with PR. Let us know when you have something you want us to review. Thanks

ajgpitch:
> closes #1225

A little more detail in the opening comment is preferred. It helps whoever has to update the changelog.

BoxiLi:
@MartinSandeCosta, free feel to complete the tests in the way you want. However, maybe it is good to discuss a bit with @jakelishman? Since there is a refactoring of `qutip/tests/test_gates.py` in #1181. A lot of change in this test might make it hard to resolve two PRs later.

jakelishman:
Hmmm.  You're right that there will be some nasty merge conflicts with #1181, but I don't see anyway around those, to be honest without that PR being merged before this one (which doesn't look very likely).  Barring #1181 getting merged in the next day or two, I'd just go ahead with these commits as written.

If the new-style tests were merged, `test_gates.py` would have a much tinier diff.  The single-qubit gates would just go between lines 222 and 223 in https://github.com/qutip/qutip/blob/77c21f8a97ddfff2353490d3587cc1effdfe9abc/qutip/tests/test_gates.py#L222-L239 as
```python
    pytest.param(gates.x_gate, 0, "X"),
    pytest.param(gates.y_gate, 0, "Y"),
    ...
```
(and we'd probably change the function name to `test_one_qubit`).  The controlled qubits would similarly go into the next test's parametrisation list: https://github.com/qutip/qutip/blob/77c21f8a97ddfff2353490d3587cc1effdfe9abc/qutip/tests/test_gates.py#L241-L259

Overall, it's clear that you can't write sensible new-style tests without #1181 having been merged, so I wouldn't worry about it too much.  I did drop in a couple of comments in-line (minor nitpicking), but given that this file may well get pretty much overwritten - keeping the actual tests, of course! - in short order, I wouldn't consider them changes that need to be made, unless Boxi asks.

Also, as you noted, `test_gates.py` doesn't really do much testing of the actual form of the gates yet.  I haven't actually been adding new tests as I go, because I wanted to keep #1181 roughly focussed on just converting what we've got so far.

qutip/tests/test_gates.py
```diff
@@ -33,14 +33,17 @@
...
-                       sqrtswap, molmer_sorensen,
-                       toffoli, fredkin, gate_expand_3toN, 
-                       qubit_clifford_group, expand_operator)
+from qutip.qip.operations.gates import (rx, ry, rz, x_gate, y_gate, z_gate,
+                                        s_gate, t_gate, cy_gate, cz_gate,
+                                        cs_gate, ct_gate, phasegate, qrot,
+                                        cnot, swap, iswap, sqrtswap,
+                                        molmer_sorensen, toffoli, fredkin,
+                                        gate_expand_3toN, qubit_clifford_group,
+                                        expand_operator)
```
jakelishman, line 46: While this is technically pep8-compliant, I think the spirit of the rules would suggest that
```python
from qutip.qip.operations.gates import (
    rx, ry, rz, ...,
    ...,
)
```
would make more sense (since it gives us much more space).

qutip/tests/test_gates.py
```diff
@@ -70,6 +73,71 @@ def testSwapGate(self):
         psi_res = swap() * swap() * psi_in
         assert_((psi_in - psi_res).norm() < 1e-12)
 
+    def test1QubitGates(self):
```
jakelishman, line 76: The actual name of the test function is typically the most accessible "documentation" of the test in the log, so more descriptive is always better (any surrounding class will also appear).  It's not like you have to call them manually, so they really can be like 60 characters long if it helps.

I know that you've generally followed what's in the rest of the file, though - this is just a general comment.

qutip/tests/test_gates.py
```diff
@@ -70,6 +73,71 @@ def testSwapGate(self):
...
+
+                 psi_in = tensor([psi1 if k == m else psi_rand_list[k]
+                                  for k in range(N)])
+                 psi_out = tensor([psi2 if k == m else psi_rand_list[k]
+                                   for k in range(N)])
+
+                 G = g(N, m)
+                 psi_res = G * psi_in
+
+                 assert_((psi_out - psi_res).norm() < 1e-12)
```
jakelishman, line 99: Again, I know you're just following what's going on in the rest of the file (and that's typically a very good thing!).  Now that we've moved to using `pytest`, `np.testing.assert_` is unnecessary - you can just use the keyword `assert`.  It used to be used because numpy abstracted their own tests away from the framework that they used to run them (sort of), and then this provided consistency.  Nowadays, `pytest` will do introspection on any variables that appear in a bare `assert` statement and print out their values, so we actually get much better error reporting by just doing `assert`.

qutip/tests/test_gates.py
```diff
@@ -70,6 +73,71 @@ def testSwapGate(self):
...
+
+    def testCtrld1QubitGates(self):
```
jakelishman, line 101: I'm strongly against arbitrary contraction in names, and that counts too in test descriptions - I'd have called this test `test_controlled_one_qubit_gates` (though I know `qutip` is a bit of a fiend from randomly chopping bits out of words in keywords all over the place).

Given that you've been chopping and changing everything in the name of pep8, I'm surprised you went along with the rest of the file's camel-cased functions!

qutip/tests/test_qubitcircuit.py
```diff
@@ -203,31 +203,88 @@ def test_add_state(self):
...
+        assert_(qc1.input_states[0] is None)
 
         assert_(qc1.input_states[2] == "0")
         assert_(qc1.input_states[3] == "0")
         assert_(qc1.input_states[6] == "0")
         assert_(qc1.input_states[1] == "+")
         assert_(qc1.input_states[4] == "+")
 
-        assert_(qc1.output_states[2] == None)
+        assert_(qc1.output_states[2] is None)
```
jakelishman, line 224: _If_ we were to keep using the `np.testing` functions (which we don't need to because `pytest` makes it all easier), the better one here would be `np.testing.assertEqual` (even for the `None` test).

We always should think about what information we want to get out of a failed test.  In this case, we'd be best served by knowing what the incorrect value was, so we should pass the two parts of the comparison separately.  That way, `np.testing.assertEqual` will print them out on a test failure, whereas this way `np.testing.assert_` just gets fed the value `False` and we wouldn't know what it was (was it 3, 0, "", "None"?) and we'd have to do more work offline to find out.

qutip/tests/test_qubitcircuit.py
```diff
@@ -203,31 +203,88 @@ def test_add_state(self):
...
         assert_(qc1.output_states[0] == "beta")
 
+    def test_exceptions(self):
+        """
+        Text exceptions are thrown correctly for inadequate inputs
+        """
+        qc = QubitCircuit(2)
+        for gate in ["X", "Y", "Z", "S", "T"]:
+            assert_raises(ValueError, qc.add_gate, gate,
+                          targets=[1], controls=[0])
```
jakelishman, line 238: In new `pytest`-style, the `for` loop would be done with parametrised tests and the raising test uses the `pytest.raises` context manager, so you get easier-to-read code (and it's selectable if you only want to test a certain gate) like
```python
@pytest.mark.parametrize('gate', ['X', 'Y', 'Z', 'S' ,'T'])
def test_exceptions(self, gate):
    qc = QubitCircuit(2)
    with pytest.raises(ValueError):
        qc.add_gate(gate, targets=[1], controls=[0])
```
Again, thinking of the test log, this way is nice because the test log will tell you exactly which gate failed - `pytest` adds the parametrised arguments into the test name.

MartinSandeCosta:
Thank you for taking the time to look at these changes, @jakelishman! It must have been a true nightmare looking at my code haha 😄 

Now on a more formal note. Regarding the pytest side of things, should I reformat the tests (related to my single_qubit implementation) in test_qubitcircuit.py and test_gates to follow the changes in #1181? It would definitely make it easier to merge them later on.

qutip/tests/test_gates.py
MartinSandeCosta: Fixed on ad74aeb

qutip/tests/test_gates.py
MartinSandeCosta: Fixed on ad74aeb

qutip/tests/test_gates.py
MartinSandeCosta: Fixed on ad74aeb

jakelishman:
Nah, don't disparage yourself - your code is good and you've done well to copy the style of the rest of the files.  Typically consistency is better than having "the one true style"; code's read more than it's written, and having a module which is clearly written by four different people makes the logic very very difficult to follow, which in turns makes it much harder to spot bugs.

If it were up to me (which is isn't, really), I would say that your changes are good as they are.  I didn't mean for the comments to be changes you _had_ to make.  You can't reduce the number of merge conflicts that are going to happen with #1181 anyway, because that PR is a massive refactor of large swathes of the testing suite.  You've done a good job to keep your work consistent with the rest of the file, so it will be very easy to merge in your logical changes.  Merge conflicts are pretty much just par for the course; just the fact that we're discussing this here shows that logically, there _is_ a conflict between two pieces of work being done at the same time.

I'd suggest merging this as-is.  We'll handle the problems caused by the refactor elsewhere.

qutip/qip/operations/gates.py
```diff
@@ -44,9 +44,9 @@
 from qutip.states import fock_dm
 
 
-__all__ = ['x_gate', 'y_gate', 'z_gate', 's_gate', 't_gate',
-           'rx', 'ry', 'rz', 'sqrtnot', 'snot', 'phasegate', 'qrot',
-           'cphase', 'cnot',
+__all__ = ['rx', 'ry', 'rz', 'sqrtnot', 'snot', 'phasegate', 'qrot',
+           'x_gate', 'y_gate', 'z_gate', 'cy_gate', 'cz_gate', 's_gate',
```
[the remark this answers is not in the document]
MartinSandeCosta: ***CX_GATE*** aka ***CNOT*** was already implemented

qutip/qip/circuit.py
```diff
@@ -331,26 +343,30 @@ def add_1q_gate(self, name, start=0, end=None, qubits=None,
                                        arg_value=arg_value,
                                        arg_label=arg_label))
 
-    def add_circuit(self, qc, start=0):
+    def add_circuit(self, qubit_circuit, start=0):
```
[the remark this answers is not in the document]
MartinSandeCosta: fixed on 8a844d1 :D

qutip/qip/operations/gates.py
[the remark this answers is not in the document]
BoxiLi: Aha, yes you are right.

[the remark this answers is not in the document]
jakelishman: It might be good to put in a `cx_gate` just for consistency.  You can just define the gate through `cx_gate = cnot` (or whatever it is) and add the extra line to `circuit.py`.

People are more likely to access it through `cnot`, but it's kind of weird to define a `cy_gate` and a `cz_gate` but have `cx_gate` only accessible by a completely different name.  It just seems like an unnecessary complication to remember that you can't have it that way.

The duplication is similar to how we have both `qeye` and `identity` to construct an identity operator.

BoxiLi:
@MartinSandeCosta Congrats!

By the way, the changelog in the PR description will later be used in the documentation (http://qutip.org/docs/latest/changelog.html). It's better to be a summary of the PR in one-line :)

qutip/qip/operations/gates.py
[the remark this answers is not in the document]
BoxiLi: Aha... sorry I missed this comment.

PR #1242, opened 2020-04-30
fix function QubitCircuit.remove_gate

Canoming:
1. `self.N` in `QubitCircuit` is not the number of gates but the number of qubits.
2. The error of index excesses number of gates was not captured.
3. `self.gates.remove(self.gates[self.N - i])` removes the first match but not the last.

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -438,7 +438,7 @@ class QubitCircuit:
                     break
 
         elif name is not None and remove == "all":
-            for i in range(len(self.gates)):
+            for i in reversed(range(len(self.gates))):
                 if name == self.gates[i].name:
                     self.gates.pop(i)
 
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -414,7 +414,9 @@ class QubitCircuit:
         remove : string
             If first or all gate are to be removed.
         """
-        if index is not None and index <= len(self.gates):
+        if index is not None:
+            if index > len(self.gates):
+                raise ValueError("Index exceeds number of gates.")
             if end is not None and end <= len(self.gates):
                 for i in range(end - index):
                     self.gates.pop(index + i)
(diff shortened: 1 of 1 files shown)


qutip/qip/circuit.py
```diff
@@ -414,14 +414,16 @@ def remove_gate(self, index=None, end=None, name=None, remove="first"):
...
                     self.gates.pop(index + i)
             elif end is not None and end > self.N:
                 raise ValueError("End target exceeds number of gates.")
             else:
                 self.gates.pop(index)
+        else:
+            raise ValueError("Index exceeds number of gates.")
 
         elif name is not None and remove == "first":
             for gate in self.gates:
```
jakelishman, line 429: While they do do it in a somewhat obscure manner, the CI errors are telling you that there's invalid syntax here; you've inserted an `else` statement before an `elif`.

Canoming: Sorry, my bad. I've fixed the problem

Canoming: Sorry, my bad. It should work now

BoxiLi:
Hi @Canoming, thanks for spotting this error and fixing it. I'm surprised that this error is hidden here for so long.

There is still a flaw in the option for `remove="all"`. When you remove the first element in a list, the rest elements will move forward and eventually leads to an index out of range error.
For example, if you have a list ``gates = [a, b, c]`` and use ``gates.pop(0)``. 
The list will become
 ```
>>> gates
[b, c]
```
if you now iterate to ``i=1``, you skip element `b`
This can be solved by iterating in the inverse order.
```
        elif name is not None and remove == "all":
            for i in reverse(range(len(self.gates))):
```

Canoming:
> Hi @Canoming, thanks for spotting this error and fixing it. I'm surprised that this error is hidden here for so long.
> 
> There is still a flaw in the option for `remove="all"`. When you remove the first element in a list, the rest elements will move forward and eventually leads to an index out of range error.
> For example, if you have a list `gates = [a, b, c]` and use `gates.pop(0)`.
> The list will become
> 
> ```
> >>> gates
> [b, c]
> ```
> 
> if you now iterate to `i=1`, you skip element `b`
> This can be solved by iterating in the inverse order.
> 
> ```
>         elif name is not None and remove == "all":
>             for i in reverse(range(len(self.gates))):
> ```

Ah, sure. Thanks to point it out. It's fixed now.

I actually used the reversed order but then deleted it as I forgot why I did that. LOL

PR #1248, opened 2020-05-04
Floquet changes

CamilleLCal:
**Description**
1) Changes in floquet_master_equation_rates to make it faster.
2) Correction transposition errors in floquet_master_equation_tensor and changes to make it faster. NB: This version works only for 1 rate matrix but could be generalized easily.
3) Correction in floquet_markov_mesolve. The basis change that returns the results in the computational had errors. NB: Results are still given in the interaction picture.

**Related issues or PRs**

**Changelog**
Fixed transposition error in floquet_master_equation_tensor
Fixed basis change error in floquet_markov_mesolve
Made floquet_master_equation_rates faster
Made floquet_master_equation_tensor faster

### DIFF
--- a/qutip/floquet.py
+++ b/qutip/floquet.py
@@ -614,15 +614,21 @@ def floquet_master_equation_rates(f_modes_0, f_energies, c_op, H, T,
         # instead of the propagator.
 
         # f_modes_t = floquet_modes_t(f_modes_0, f_energies, t, H, T, args)
-        f_modes_t = floquet_modes_t_lookup(f_modes_table_t, t, T)
-        for a in range(N):
-            for b in range(N):
-                k_idx = 0
-                for k in range(-kmax, kmax + 1, 1):
-                    # [:1,:1][0, 0] patch around scipy 1.3.0 bug
-                    X[a, b, k_idx] += (dT / T) * exp(-1j * k * omega * t) * \
-                        (f_modes_t[a].dag() * c_op * f_modes_t[b])[:1, :1][0, 0]
-                    k_idx += 1
+        #f_modes_t = floquet_modes_t_lookup(f_modes_table_t, t, T)
+        #for a in range(N):
+        #    for b in range(N):
+        #        k_idx = 0
+        #        for k in range(-kmax, kmax + 1, 1):
+        #            # [:1,:1][0, 0] patch around scipy 1.3.0 bug
+        #            X[a, b, k_idx] += (dT / T) * exp(-1j * k * omega * t) * \
+        #                (f_modes_t[a].dag() * c_op * f_modes_t[b])[:1, :1][0, 0]
+        #            k_idx += 1
+
+  		#Does the same as before but much faster
+        f_modes_t = np.hstack([f.full() for f in floquet_modes_t_lookup(f_modes_table_t, t, T)])
+        FF = f_modes_t.T.conj() @ c_op.full() @ f_modes_t
+        phi = exp(-1j * np.arange(-kmax, kmax+1) * omega * t)
+        X += (dT / T) * np.einsum("ij,k->ijk",FF, phi)
 
     Heaviside = lambda x: ((np.sign(x) + 1) / 2.0)
     for a in range(N):
--- a/qutip/floquet.py
+++ b/qutip/floquet.py
@@ -713,25 +713,51 @@ def floquet_master_equation_tensor(Alist, f_energies):
         N, M = np.shape(Alist[0])
 
     Rdata_lil = scipy.sparse.lil_matrix((N * N, N * N), dtype=complex)
-    for I in range(N * N):
-        a, b = vec2mat_index(N, I)
-        for J in range(N * N):
-            c, d = vec2mat_index(N, J)
 
-            R = -1.0j * (f_energies[a] - f_energies[b])*(a == c)*(b == d)
-            Rdata_lil[I, J] = R
+    ###
+    # Here there are transposition errors because the paper used in
+    # the documentation is wrong. Moreover line 726 should not be.
+    ###
+    #for I in range(N * N):
+    #    a, b = vec2mat_index(N, I)
+    #    for J in range(N * N):
+    #        c, d = vec2mat_index(N, J)
+    #
+    #        R = -1.0j * (f_energies[a] - f_energies[b])*(a == c)*(b == d)
+    #        Rdata_lil[I, J] = R
 
-            for A in Alist:
-                s1 = s2 = 0
-                for n in range(N):
-                    s1 += A[a, n] * (n == c) * (n == d) - A[n, a] * \
-                        (a == c) * (a == d)
-                    s2 += (A[n, a] + A[n, b]) * (a == c) * (b == d)
+    #        for A in Alist:
+    #            s1 = s2 = 0
+    #            for n in range(N):
+    #                s1 += A[a, n] * (n == c) * (n == d) - A[n, a] * \
+    #                    (a == c) * (a == d)
+    #                s2 += (A[n, a] + A[n, b]) * (a == c) * (b == d)
 
-                dR = (a == b) * s1 - 0.5 * (1 - (a == b)) * s2
+    #            dR = (a == b) * s1 - 0.5 * (1 - (a == b)) * s2
 
-                if dR != 0.0:
-                    Rdata_lil[I, J] += dR
+    #            if dR != 0.0:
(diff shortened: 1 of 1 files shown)


Ericgig:
I tried the notebook: https://github.com/qutip/qutip-notebooks/blob/master/examples/floquet-dynamics.ipynb
Under "Floquet-Markov master equation: comparison with other solvers", one of the result is strange with this version. 
I would change the Hamiltonian (everywhere is the notebook) from
`H = [H0, [H1, 'sin(w * t)']]`
to:
`H = [H0, [H1, lambda t, args: np.sin(args['w'] * t)]]`
as there is not reuse of the compiled string.

qutip/floquet.py
```diff
@@ -707,28 +722,38 @@ def floquet_master_equation_tensor(Alist, f_energies):
...
-                    s2 += (A[n, a] + A[n, b]) * (a == c) * (b == d)
-
-                dR = (a == b) * s1 - 0.5 * (1 - (a == b)) * s2
-
-                if dR != 0.0:
-                    Rdata_lil[I, J] += dR
+    ###
+    # Here there were transposition errors because the paper used in
+    # the documentation has typos.
+    # NB: Works only for 1 rate matrix.
```
Ericgig, line 729: Does that mean that Alist must have at most 1 element? 
Is it a new restriction? What happen if many rate matrix are given?

CamilleLCal: I forgot to remove the comment "# NB: Works only for 1 rate matrix.", Alist can have as many elements as necessary.

CamilleLCal:
**Summary of what was done:**

- Changes in floquet_master_equation_rates to make it faster.
- Corrected transposition errors in _floquet_master_equation_tensor_ and changes to make it faster. I also removed a line so that the ME is solved in the interaction picture, this implies a different basis change in the _floquet_markov_mesolve_ function.
- Correction in _floquet_markov_mesolve_. As explained in the point above, the Floquet-Markov ME was not being solved in the interaction picture, therefore I modified _floquet_master_equation_tensor_ and modified the basis change in _floquet_markov_mesolve_. This modification requires an additional variable in the entry of _floquet_markov_mesolve_ which can not be chosen by default. For the moment, if this variable is not given and _floquet_basis=False_, then the density matrix or expectation values are  returned in the interaction picture, in the computational basis, and a warning message is sent. 
- Added integration options in: _floquet_modes, floquet_modes_t, floquet_modes_table, floquet_states_t, floquet_modes_t, floquet_wavefunction_t, fsesolve, floquet_master_equation_rates,  fmmesolve_
- Added test functions to verify the dissipative dynamics:
   a) Compare _fmmesolve_ and _mesolve_ for the cases: (NO drive but dissipation) and (Drive but NO dissipation)
   b) Compare numerical and analytical matrix elements for a driven-dissipative TLS with RWA

There are some places where the code could be modified to optimize the running time, especially in the basis change in _floquet_markov_mesolve_.

Ericgig:
@jakelishman Could you look at this? I beleive you have some experience with floquet equation and this should eventually go to `dev-major` so it would be nice if we could look at it before you finish the merge `master` -> `dev-major`.

jakelishman:
Sure, I'll try soon.  I can add stuff to the merge I've already done - I learned some `git` magic to combine merge commits into the same object without losing my previous work.

qutip/tests/test_floquet.py
```diff
@@ -32,24 +32,267 @@
...
+        H = [H0, [H1, lambda t, args: np.sin(args['w'] * t)]]
+        e_ops = [num(2)]
+
+        # Solve schrodinger equation with floquet solver
+        sol = fsesolve(H, psi0, tlist, e_ops, T, args)
+
+        # Compare with results from standard schrodinger equation
+        sol_ref = mesolve(H, psi0, tlist, [], e_ops, args)
+
+        assert_(max(abs(sol.expect[0] - sol_ref.expect[0])) < 1e-4)
```
jakelishman, line 71: This sort of thing is better spelled as
```python
np.testing.assert_allclose(sol.expect[0], sol_ref.expect[0], atol=1e-4)
```
because then on a failure we'll see exactly what was in both arrays, and how much they differ by.

qutip/tests/test_floquet.py
```diff
@@ -32,24 +32,267 @@
...
+                                                  args={})
+            # Get X delta
+            DeltaMatrix, X, frates, Amat = floquet_master_equation_rates(
+                                            fmodes0, fenergies,
+                                            c_ops, H, T, {},
+                                            noise_spectrum, 0, 5)
+            # Check energies
+            deltas = np.ndarray.flatten(DeltaMatrix)
+
+            assert_(min(abs(deltas-array_ana_delta[idx])) < 1e-4)
```
jakelishman, line 278: `min` doesn't seem right?  That succeeds as long as any pair of elements is close, rather than requiring _all_ pairs to be close.  If it _is_ meant to be the latter, it should again be `np.testing.assert_allclose`.

qutip/floquet.py
```diff
@@ -176,7 +186,9 @@ def floquet_modes_t(f_modes_0, f_energies, t, H, T, args=None):
 
     return f_modes_t
 
-def floquet_modes_table(f_modes_0, f_energies, tlist, H, T, args=None):
+
+def floquet_modes_table(f_modes_0, f_energies, tlist, H, T, args=None,
+                        options=Options()):
```
jakelishman, line 191: `Options()` is a mutable object, so shouldn't be used as a function default.  You can instead do
```python
def floquet_modes_table(..., options=None):
    options = options or Options()
    ...
```

jakelishman (changes requested):
Heya, thanks so much for looking at the Floquet module.  I haven't had time to properly properly dig into the transposition error to verify that (though I'm sure you knew what you were doing), but I also have a couple of other concerns that I've put in the reviews below.

jakelishman:
I couldn't work out how to attach a patch to a GitHub comment, so I've made PR (CamilleLCal/qutip#1) against your fork to show the sort of changes I think `floquet_markov_mesolve` could benefit from.

CamilleLCal:
Hi @jakelishman, thank you very much for looking at the Pull Request! I will reply to your questions/comments as soon as possible.

jakelishman:
Sounds good!  When I made the review, I actually didn't even realise how slow the module was at all - I didn't realise how much of an improvement your PR represented.  It'll still be good to possibly fix the quartic memory usage (I can help if necessary), but it's already an order of magnitude improvement on speed, not to mention the incorrect dynamics.

qutip/floquet.py
```diff
@@ -716,130 +757,177 @@ def floquet_basis_transform(f_modes, f_energies, rho0):
...
     output.times = tlist
-
     if isinstance(e_ops, FunctionType):
-        n_expt_op = 0
         expt_callback = True
-
-    elif isinstance(e_ops, list):
-
-        n_expt_op = len(e_ops)
+        store_states = options.store_states or False
```
Ericgig, line 840:
```suggestion
        store_states = opt.store_states or False
```

qutip/floquet.py
```diff
@@ -716,130 +757,177 @@ def floquet_basis_transform(f_modes, f_energies, rho0):
...
             output.num_expect = n_expt_op
             for op in e_ops:
-                if op.isherm:
-                    output.expect.append(np.zeros(n_tsteps))
-                else:
-                    output.expect.append(np.zeros(n_tsteps, dtype=complex))
-
+                dtype = np.float64 if op.isherm else np.complex128
+                output.expect.append(np.zeros(len(tlist), dtype=dtype))
+        store_states = options.store_states or (n_expt_op == 0)
```
Ericgig, line 856:
```suggestion
        store_states = opt.store_states or (n_expt_op == 0)
```

Ericgig:
@jakelishman I meet with @CamilleLCal and we went over your comments. Do you have time for another review. If you don't please tell us and we will do it.

There is a small change to be made in the qutip/qutip-notebooks#121 with `floquet_markov_mesolve`. @CamilleLCal said she will look at it in the following weeks.

jakelishman (commented):
I'm so sorry I ended up dropping the ball on this PR, particularly to Camille, since it was her work.  I just glanced through, and as best as I remember it, the major concerns I had are now solved (large duplication of transformation-handling code in the solver loop).  I think there's a more efficient method of constructing the sparse tensor operator in COO/CSR form, but this PR is already a huge improvement in speed on the previous one, and I very much don't want to hold it up any more.

PR #1249, opened 2020-05-04
Add pytest helper functions and fixtures

jakelishman:
This is the base PR for the split-up of #1181.  This one puts in all the helper functions and updates to previously converted files that are present in that PR, and needs to be merged first before any of the others can be merged.  The commits are tidied up and rebased onto `master`.

This is the base PR for #1250, #1251 and #1252.

**Changelog**
Add pytest helper functions for large-scale test refactoring

### DIFF
--- a/qutip/tests/conftest.py
+++ b/qutip/tests/conftest.py
@@ -50,9 +50,24 @@ def _add_repeats_if_marked(metafunc):
                              ids=["rep({})".format(x+1) for x in range(count)])
 
 
+def _skip_cython_tests_if_unavailable(item):
+    """
+    Skip the current test item if Cython is unavailable for import, or isn't a
+    high enough version.
+    """
+    if item.get_closest_marker("requires_cython"):
+        # importorskip rather than mark.skipif because this way we get pytest's
+        # version-handling semantics.
+        pytest.importorskip('Cython', minversion='0.14')
+
+
 @pytest.hookimpl(trylast=True)
 def pytest_generate_tests(metafunc):
     _add_repeats_if_marked(metafunc)
+
+
+def pytest_runtest_setup(item):
+    _skip_cython_tests_if_unavailable(item)
 
 
 @pytest.fixture
--- a/qutip/tests/conftest.py
+++ b/qutip/tests/conftest.py
@@ -32,8 +32,10 @@
 ###############################################################################
 
 import pytest
+import functools
 import os
 import tempfile
+import numpy as np
 
 
 def _add_repeats_if_marked(metafunc):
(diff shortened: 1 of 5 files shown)


jakelishman:
@Ericgig: another random failure in the CI here (errored test is unchanged in this PR).

Ericgig:
Computed error 1.08705544e-07, tol 1e-7... The initial ket in `rand_ket(2)`.

I believe that with `nosetest`, the random seed was kept between test file. One of our early test would fix the seed and all following tests would be deterministic. With `pytest` the seed is probably reset at each file. We could loosen the tolerance. Ideally we would get the stats on the error first. Maybe `pytest` has a support for test that pass 99% of the time.

jakelishman:
I'd be vastly surprised if there's a way of permitting a test which passes 99% of the time - on any particular test run how would it know?

I don't remember having seen anything about setting the initial seeds, but relying on a previous test to set a seed isn't a good idea - you'd always have to ensure that no tests were ever added or changed in-between that ever used random state or everything would get lost.  We could fix the random seed for any test which uses random methods with a fixture to get around that.

I actually think this is a true failure; the default solver `rtol = 1e-6`, but here we're using `rtol=1e-7` as the test, so the test is bugged (it's probably not the only one, and I may have missed some of these in `mcsolve` too).  I should change this when I convert `test_mesolve.py`.

Ericgig:
Repeat the test 20 times and accept if at least 19 pass?

I remember adapting the tests when numpy's rng changed. 1~2 where failing. I believe that the tests pass over 99% with a random seed, but with the number of tests, it's not enough.

I did not realize the test tolerance was smaller than the test one. But the solver tolerance is for the state itself, not expectation values and other scalars obtained from them. The test is not to check numerical error. When there is an mistake in the code, the error is orders of magnitude greater. Not sure if using the sovler's tol is right, but probably better than using a number that just seems to work.

jakelishman:
Repeating just changes the probability of failure, rather than actually detecting whether it was a rare fail, and `mesolve` tests aren't always the fastest, so repetition here isn't great.

For simple unitary qubit operations, I think it should be easy enough to work out how the tolerance translates.  It's difficult to say that if there's a numerical error then the error will be much greater - I caught some errors in the `test_gates.py` (I think) that were sneaking through because the tolerances were set too high.  It's probably best to keep as tight a tolerance as is reasonable - here I think `2 * qutip.Options.rtol` would always be correct if the numerics are.

Ericgig:
`2 * Options.rtol` could be good enough for most the tests.
 
But expecting `Options.rtol` to be a good estimation of the numerical error in all case is risky. There is no promise by the solver that the result is within rtol of the analytical solution and it ignore error propagation in the post-treatment steps.

jakelishman:
I realised I never replied to this.

I suggested `2*rtol` in this case because the error propagation in this case would guarantee that this is always sufficient.  For other tests, we can always design the test so that it's calculable what the error term should be.  I'd say it's generally a good idea to have the tests be as tight as we can possibly make them without risking false-negatives on the results, because otherwise we're not properly testing that the parameters do what they say.

Regardless, that test failure isn't due to any of my PRs - it's a pre-existing problem that we happened to get unlucky on this time round.  It'll get fixed when I modify `test_mesolve.py`.

Can we review this PR to get the ball rolling on getting all of the test refactoring considered?

Possible reviewers: @Ericgig, @BoxiLi, @nathanshammah

qutip/tests/conftest.py
```diff
@@ -0,0 +1,197 @@
...
+        # version-handling semantics.
+        pytest.importorskip('Cython', minversion='0.14')
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_generate_tests(metafunc):
+    _add_repeats_if_marked(metafunc)
+
+
+def pytest_runtest_setup(item):
```
BoxiLi, line 71: If I understand correctly, this applies to all tests, where `item` is the test function (what pytest called "Node")?

qutip/tests/conftest.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    previous_dir = os.getcwd()
+    with tempfile.TemporaryDirectory() as temporary_dir:
+        os.chdir(temporary_dir)
+        yield
+        # pytest should catch exceptions occuring in functions using the
+        # fixture, so this should always be called.  We want it here rather
+        # than outside to prevent the case of the directory failing to be
+        # removed because it is 'busy'.
+        os.chdir(previous_dir)
```
BoxiLi, line 191: What goes after yield will be executed even if there are Exceptions in the test functions. It is like the `finally` in Python.

qutip/tests/conftest.py
jakelishman: Yeah exactly.  There's a [whole bunch of hook functions](https://docs.pytest.org/en/2.7.3/plugins.html#well-specified-hooks) that you can fill in if you like.  Each hook applies to every relevant item in its scope.  Here, `conftest.py` is a magic filename that acts at directory-scope.  We could also set this function with a file, or within a class, and it would only apply to tests in the same scope.  We do that a little bit in some files with `pytest_generate_tests`.

I set this hook up so it can be added to easily, but right now all it does is just skip tests marked with `requires_cython` if Cython isn't available.

qutip/tests/conftest.py
jakelishman: I don't think this is right - coroutines don't modify the flow of exception handling.  This is a mocked up and simplified version of `pytest` running a test:
```python
def fixture():
    """A possible fixture"""
    print("Initialising fixture.")
    yield 12
    print("Finalising fixture.")

def test(f):
    """A test case"""
    if f == 12:
        raise ValueError()

def run_tests(finalise=True):
    """MWE of how pytest runs tests."""
    fixtures = [fixture()]  # Store all fixture functions
    args = [next(f) for f in fixtures]  # Advance coroutines to get values
    successes = 0
    caught = []
    try:
        print("Running tests.")
        test(*args)
        successes += 1
    except Exception as e:  # Catch Assertion/WhateverError
        caught.append(e)
    finally:
        if finalise:
            for f in fixtures:
                # Try to finish the coroutine, i.e. control will start from the
                # `yield` statement, and this is when the finaliser is run.  If
                # it raises StopIteration that means it completed.  If it
                # doesn't, that means it must have tried to return a new value.
                try:
                    next(f)
                except StopIteration:
                    pass
                else:
                    raise Exception("fixture tried to yield more than once")
    return successes, caught
```
Now note that `run_tests(finalise=True)` will print "Finalising fixture.", but `run_tests(finalise=False)` won't.

In fact, it's not really anything to do with exceptions - it's just that control doesn't return to the coroutine until someone passes it back.  In this case, `pytest` does (by calling `next` again), so it does finalise correctly.

BoxiLi: Aha, this is an interesting use of `yield`. I find it quite similar to a decorator, or a context manager. You wrap the function with it every time you detect the marker.

jakelishman: Yeah, it's pretty funky.  We don't yield anything here because we don't actually care about providing a value, we only care about ensuring that the setup/teardown code is executed before and after the relevant testing scopes.

jakelishman:
I would suggest that this PR is merged as-is (not squashed) once it's ready.  There's several separate things that happen, and it helps the history to keep those entries apart.

BoxiLi:
Hi, Someone still likes to check this PR or shall we merge? @Ericgig @ajgpitch @nathanshammah

qutip/tests/conftest.py
jakelishman: Also, while this is quite an old comment now, `item` refers to an object `pytest.Item`, which subclasses `pytest.Node`.

jakelishman:
@Ericgig: I see you've been on a merging spree!

If you're about to merge, please don't squash this - I've taken care to have good commit messages and modular commits, and I'll have to rebase all the other PRs if this one gets squashed.

jakelishman:
Thanks, I'm really glad to see this merged!  I've marked #1250, #1251 and #1252 as ready for review/merge since this has been merged.

PR #1251, opened 2020-05-04
Convert tests to pytest (gates and cavityqed)

jakelishman:
These are the files that Boxi (@BoxiLi) was checking in #1181, but the commits are tidied up and rebased onto `master`.  The merge conflict in `test_gates.py` that's present in #1181 is fixed (with all the tests safely maintained).

This is based on #1249, and that one needs to be merged first.

**Changelog**
Major test refactor into a pytest style.

### DIFF
--- a/qutip/tests/test_gates.py
+++ b/qutip/tests/test_gates.py
@@ -31,430 +31,330 @@
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 
+import pytest
+import functools
 import itertools
 import numpy as np
-from numpy.testing import (assert_, assert_allclose, assert_array_equal,
-                           run_module_suite)
-from qutip.states import basis, ket2dm
-from qutip.operators import identity, qeye, sigmax, sigmay, sigmaz
-from qutip.qip.operations.gates import (
-    rx, ry, rz, x_gate, y_gate, z_gate, s_gate, t_gate, cy_gate, cz_gate,
-    cs_gate, ct_gate, phasegate, qrot, cnot, swap, iswap, sqrtswap,
-    molmer_sorensen, toffoli, fredkin, gate_expand_3toN, qubit_clifford_group,
-    expand_operator)
-from qutip.random_objects import rand_ket, rand_herm, rand_unitary, rand_dm
-from qutip.tensor import tensor
-from qutip.qobj import Qobj
+import qutip
+from qutip.qip.operations import gates
 
 
-class TestGates:
+def _permutation_id(permutation):
+    return str(len(permutation)) + "-" + "".join(map(str, permutation))
+
+
+def _infidelity(a, b):
+    """Infidelity between two kets."""
+    return 1 - abs(a.overlap(b))
+
+
+def _remove_global_phase(qobj):
     """
-    A test class for the QuTiP functions for generating quantum gates
+    Return a new Qobj with the gauge fixed for the global phase.  Explicitly,
+    we set the first non-zero element to be purely real-positive.
     """
+    flat = qobj.full().flat.copy()
+    for phase in flat:
+        if phase != 0:
+            # Fix the gauge for any global phase.
+            flat = flat * np.exp(-1j * np.angle(phase))
+            break
+    return qutip.Qobj(flat.reshape(qobj.shape), dims=qobj.dims)
 
-    def testSwapGate(self):
+
+def _make_random_three_qubit_gate():
+    """Create a random three-qubit gate."""
+    operation = qutip.rand_unitary(8, dims=[[2]*3]*2)
+
+    def gate(N=None, controls=None, target=None):
+        if N is None:
+            return operation
+        return gates.gate_expand_3toN(operation, N, controls, target)
+    return gate
+
+
+def _tensor_with_entanglement(all_qubits, entangled, entangled_locations):
+    """
+    Create a full tensor product when a subspace component is already in an
+    entangled state.  The locations in `all_qubits` which are the entangled
+    points in the output are ignored and can take any value.
+
+    For example,
+        _tensor_with_entanglement([|a>, |b>, |c>, |d>], (|00> + |11>), [0, 2])
+    should product a tensor product like (|0b0d> + |1b1d>), i.e. qubits 0 and 2
+    in the final output are entangled, but the others are still separable.
+
+    Parameters:
+        all_qubits: list of kets --
+            A list of separable parts to tensor together.  States that are in
+            the locations referred to by `entangled_locations` are completely
+            ignored.
+        entangled: tensor-product ket -- the full entangled subspace
+        entangled_locations: list of int --
+            The locations that the qubits in the entangled subspace should be
+            in in the final tensor-product space.
+    """
+    n_entangled = len(entangled.dims[0])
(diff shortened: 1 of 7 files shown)


jakelishman:
Unfortunately not, or at least, certainly not the underlying cause - I was only able to reproduce the segfaults maybe once, but then it disappeared once I attached a debugger.  It's very unlikely to be that particular test function, since I used to get them in fairly random locations.  Just rearranging the test functions may have caused it to temporarily disappear, but it'll be back eventually until we find out whatever use-after-free or out-of-bounds error's actually causing it.

qutip/tests/test_gates.py
```diff
@@ -31,430 +31,329 @@
...
+                                                 [q1, q2, q3])
+            assert _infidelity(test, expected) < 1e-12
+
+
+class Test_expand_operator:
+    # Conceptually, a lot of these tests are complete duplicates of
+    # `TestGateExpansion`, except that they explicitly target an underlying
+    # function in `qutip.qip.operations.gates` which (as of 2020-03-01) is not
+    # called in the majority of cases---it appears to be newer than the
+    # surrounding code, but the surrounding code wasn't updated.
```
BoxiLi, line 295: Yes, its kind of my fault. In principle, `expand_operator` can replace all the `gate_expand_DtoN` functions. We will fix it this year.

BTW, do you think `expand_operator` can somehow be included in the core qutip?  Probably during your GSoC project? It is quite general and can also be used outside `qip`.

qutip/tests/test_gates.py
```diff
@@ -31,430 +31,329 @@
...
+                             controls=permutation[:2],
+                             target=permutation[2])
+        base = (qutip.tensor(1 - qutip.basis([2, 2], [1, 1]).proj(),
+                             qutip.qeye(2))
+                + qutip.tensor(qutip.basis([2, 2], [1, 1]).proj(),
+                               qutip.sigmax()))
+        # Iterate the permutation once, equivalent to finding its inverse,
+        # because for us, `permutation[n]` is where qubit `n` should be placed,
+        # whereas `Qobj.permuate` interprets that qubit `n` should be at the
+        # location `i` where `permutation[i] == n`.
```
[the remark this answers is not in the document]
jakelishman: Yeah - that's somewhat deliberate.  I reuse the `_apply_permutation` function to generate tests for `expand_operator()` further down, since hopefully the code is very clear about what's happening.

The comment here is because the old test of the Toffoli gate had some very _very_ convoluted logic (that I don't even think was correct in the end...), and it took me a long time to work out what it was trying to do.

qutip/tests/test_gates.py
```diff
@@ -31,430 +31,329 @@
...
+        test = gates.toffoli(N=3,
+                             controls=permutation[:2],
+                             target=permutation[2])
+        base = (qutip.tensor(1 - qutip.basis([2, 2], [1, 1]).proj(),
+                             qutip.qeye(2))
+                + qutip.tensor(qutip.basis([2, 2], [1, 1]).proj(),
+                               qutip.sigmax()))
+        # Iterate the permutation once, equivalent to finding its inverse,
+        # because for us, `permutation[n]` is where qubit `n` should be placed,
+        # whereas `Qobj.permuate` interprets that qubit `n` should be at the
```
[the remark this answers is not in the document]
jakelishman: You wouldn't believe how many times I mis-typed `permute` writing these tests!  I'm surprised it's only one mistake that got left in!

qutip/tests/test_gates.py
jakelishman: It'll be good to get it all updated to be uniform, though I wouldn't be super worried about it right now because everything still does work.

I won't change `expand_operator` as part of the upcoming project, because it's acting conceptually on a higher level of abstraction than what we're changing (i.e. it uses `Qobj` methods for data access, not directly mutating `fast_csr_matrix` types), but I do agree that it could be promoted up into the global namespace and out of the `qip` module.  It's effectively a helper function wrapping around `tensor`, and I think it's a really useful operation that would really ease operator construction (though the cyclic permutation option is perhaps a bit specific to be given such a prominent place).

qutip/tests/test_gates.py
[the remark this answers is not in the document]
BoxiLi: I also made exactly this mistake before!  It must have something to do with the human hand and the keyboard layout :)

I'll commit this suggestion here.

BoxiLi:
This "commit suggestion" feature is indeed very convenient @nathanshammah. Thanks!

jakelishman:
I'm not sure you've suggested a commit so much as just added one!

jakelishman:
#1249 should be reviewed and merged before this (because if it needs changed, then this PR has to be rebased on top of it), but also I'm going to change the Clifford group tests a little, and rebase the typo fix commit into the general refactor before this one should be merged.

BoxiLi:
> I'm going to change the Clifford group tests a little, and rebase the typo fix commit into the general refactor before this one should be merged.

Sure

qutip/tests/test_gates.py
```diff
@@ -31,430 +31,330 @@
...
+    Return a new Qobj with the gauge fixed for the global phase.  Explicitly,
+    we set the first non-zero element to be purely real-positive.
+    """
+    flat = qobj.full().flat.copy()
+    for phase in flat:
+        if phase != 0:
+            # Fix the gauge for any global phase.
+            flat = flat * np.exp(-1j * np.angle(phase))
+            break
+    return qutip.Qobj(flat.reshape(qobj.shape), dims=qobj.dims)
```
jakelishman, line 62: Force pushed changes here.

qutip/tests/test_gates.py
```diff
@@ -31,430 +31,330 @@
...
 
-                        o1 = psi_out.overlap(psi_in)
-                        o2 = psi_ref_out.overlap(psi_ref_in)
-                        assert_(abs(o1 - o2) < 1e-12)
+    def test_all_elements_different(self):
+        clifford = [_remove_global_phase(gate) for gate in self.clifford]
+        for i, gate in enumerate(clifford):
+            for other in clifford[i+1:]:
+                # Big tolerance because we actually want to test the inverse.
+                assert not np.allclose(gate.full(), other.full(), atol=1e-3)
```
jakelishman, line 194: ... and here

qutip/tests/test_gates.py
```diff
@@ -31,430 +31,330 @@
...
+        pauli_gates = [_remove_global_phase(x) for x in self.pauli]
+        normalised = [_remove_global_phase(gate * pauli * gate.dag())
+                      for pauli in self.pauli]
+        for gate in normalised:
+            for i, pauli in enumerate(pauli_gates):
+                if np.allclose(gate.full(), pauli.full(), atol=1e-10):
+                    del pauli_gates[i]
+                    break
+        assert len(pauli_gates) == 0
+
```
jakelishman, line 214: ... and here.

jakelishman:
The `_hashable_without_global_phase` made the tests nice and easy to read, but honestly it was just a timebomb waiting to go off and break the tests.  Floats can't reasonably be hashed as an equality test, because they should always include a delta in the comparison.

This technically makes the tests slower, but as long as we're just testing the single-qubit Clifford group, it's only a fraction of a second (because there's 24 of them ignoring global phase and 24^2  is only 576).

Issue #1254, opened 2020-05-07
solver c_ops callback function


lfry512:
Dear qutip developers,

In the documentation of version 4.5 the c_ops argument is said to support callback functions, however they are not implemented in that version and as far as I can tell neither in the current branch. It would be nice to have this feature added to allow for arbitrary time-dependent Lindbladians to be implemented. There are cases where Lindbladians cannot simply be described as the product of a time-dependent scalar and a time-independent Lindbladian, this is where the callback becomes useful.

I envisage a single callback passed to c_ops as being useful, where the user has to manually add multiple decoherence channels. This way, arbitrary data can be passed through the args parameter of the callback to build the operators at a given time step. I believe this would also allow channel cross-correlations to be implemented too.

I have hacked my local copy of qutip 4.5 to check this can in principle be done with mesolve, and it has worked. Here is a link to a patch file that shows the changes I made to get this working:
https://pastebin.com/iZCgBbNQ

I am happy to add this functionality properly to a clone and make a pull request.

Ericgig:
The recently released v5 support callback for `c_ops`.

PR #1261, opened 2020-05-12
Correct documentation in mesolve

### LINKED ISSUE #1260: Callback function form for c_ops in mesolve does not work
This issue is on behalf of [a post in the forums by Louis Fry](https://groups.google.com/forum/#!topic/qutip/93uSuJGK_Jk) who is trying to implement Lindblad operators which cannot be written in the form of a known `f(t) * A`, because the operators are related to the instantaneous eigenstates of a time-dependent Hamiltonian.

The documentation in the docstring of `mesolve` says that `c_ops` may be a callback function of the signature `f(t, args) -> Qobj`, similarly to the Hamiltonian.  This is not the case, and as best as I can tell, has never been the case.

Basic reproduction (though the functionality is just completely missing, so this won't ever succeed).
```python
import qutip
def c_ops(t, args):
    return qutip.create(3)
qutip.mesolve(qutip.num(3), qutip.basis(3, 1).proj(), [0, 1], c_ops)
```

## History

- @jrjohansson originally wrote the docstring saying that `c_ops` could be a callback function in 2012 (pre-QuTiP 1.1.4) in commit 1bf006d4, although the code did not support it at this time
- in issue #40 (2013), he commented that this functionality should be implemented, and this issue was closed without further action in 2018.
- in issue #223 (2014) there is some further discussion, but nothing is done until @Ericgig closes it in 2019, saying it is addressed by QuTiP 4.4.0 (the big swap to QobjEvo), but this doesn't appear to be the case.

As of tag `v4.4.0`, the code does not support it.  The failing line of code would be https://github.com/qutip/qutip/blob/fc9d9915d0f6242587828ce716ff3cd444edf82a/qutip/mesolve.py#L227-L236
as a function type is truth-y but has no length, but the underlying problem is more just that there's no support for it at all - there's no check for `callable(c_ops)`.

At the current `master` at the time of writing (QuTiP ~4.5.1, 4102b99), we're in the same place as at tag `v4.4.0`.

## Solution

In previous issues, people have been told to rewrite the Liouvillian in the supported form, however that's not reasonably possible here.  Technically it can be well-approximated by using an interpolation for each individual matrix element, as the time-dependent Hamiltonian can be solved ahead of time, but that's not at all sensible.

Most pressingly, the documentation is currently wrong and should be changed to reflect reality.  Secondly, we have to decide whether we're going to implement this functionality at all, perhaps as a part of @Ericgig's refactor of the solvers?

jakelishman:
Fix misleading documentation which said that `c_ops` could be a callback function.  It currently cannot, though all functionality that could be achieved through that can be emulated by passing a callback function returning the full time-dependent Liouvillian as the `H` parameter.

See #1254, fixes #1260.

### DIFF
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -65,7 +65,7 @@ def mesolve(H, rho0, tlist, c_ops=None, e_ops=None, args=None, options=None,
     set of collapse operators, or a Liouvillian.
 
     Evolve the state vector or density matrix (`rho0`) using a given
-    Hamiltonian (`H`) and an [optional] set of collapse operators
+    Hamiltonian or Liouvillian (`H`) and an optional set of collapse operators
     (`c_ops`), by integrating the set of ordinary differential equations
     that define the system. In the absence of collapse operators the system is
     evolved according to the unitary evolution of the Hamiltonian.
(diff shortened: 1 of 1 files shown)


qutip/mesolve.py
```diff
@@ -78,25 +78,25 @@ def mesolve(H, rho0, tlist, c_ops=None, e_ops=None, args=None, options=None,
...
-    Hamiltonian or Liouvillian for the system at that point in time
-    (*callback format*).
-
-    Alternatively, `H` and `c_ops` can be a specified in a nested-list format
-    where each element in the list is a list of length 2, containing an
-    operator (:class:`qutip.qobj`) at the first element and where the
-    second element is either a string (*list string format*), a callback
+    For time-dependent problems, `H` and `c_ops` can be a specified in a
+    nested-list format where each element in the list is a list of length 2,
+    containing an operator (:class:`qutip.qobj`) at the first element and where
```
nathanshammah, line 88:
```suggestion
    nested-list format where each element in the list is a list of length 2,
    containing an operator (:class:`qutip.qobj`) at the first element and when
```

PR #1264, opened 2020-05-14
Remove incorrect use of 'is' in comparisons

jakelishman:
Corrects some fragile usage of `is` to compare strings and types.  This may have worked so far, but it's fragile and liable to break (e.g. strings with the same contents will always `==` each other, but they won't always satisfy `string1 is string2`).


jakelishman:
False test failure again - we really have too many tests which fail probabilistically!  This particular one should have been smoothed out by #1250, since the current test is rather convoluted and there's several imprecise floating-point literals in use.

PR #1269, opened 2020-05-15
Fix function `QubitCircuit.add_circuit`

Canoming:
The SWAP gates couldn't be added to circuits with `add_circuit`. The function handles the `controls` of the gates while SWAP gates only take the argument `targets`

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -383,9 +383,8 @@ class QubitCircuit:
                               gate.arg_label)
             elif gate.name in ["BERKELEY", "SWAPalpha", "SWAP", "ISWAP",
                                "SQRTSWAP", "SQRTISWAP"]:
-                self.add_gate(gate.name, None,
-                              [gate.controls[0] + start,
-                               gate.controls[1] + start], None, None)
+                self.add_gate(gate.name, [gate.targets[0] + start,
+                              gate.targets[1] + start], None, None)
             elif gate.name in ["TOFFOLI"]:
                 self.add_gate(gate.name, gate.targets[0] + start,
                               [gate.controls[0] + start,
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -60,6 +60,16 @@ except ImportError:
 
 __all__ = ['Gate', 'QubitCircuit']
 
+__single_qubit_gates = ["RX", "RY", "RZ", "SNOT", "SQRTNOT", "PHASEGATE",
+                        "X", "Y", "Z", "S", "T"]
+__para_gates = ["RX", "RY", "RZ", "CPHASE", "SWAPalpha", "PHASEGATE",
+                "GLOBALPHASE", "CRX", "CRY", "CRZ"]
+__ctrl_gates = ["CNOT", "CSIGN", "CRX", "CRY", "CRZ", "CY", "CZ",
+                "CS", "CT"]
+__swap_like = ["SWAP", "ISWAP", "SQRTISWAP", "SQRTSWAP", "BERKELEY",
+               "SWAPalpha"]
+__toffoli_like = ["TOFFOLI"]
+__fredkin_like = ["FREDKIN"]
 
 class Gate:
     """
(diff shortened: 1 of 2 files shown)


quantshah:
Thanks for the fix. Could you also please add some tests for the addition of SWAP gates and removal?

BoxiLi:
Thank for spotting another bug @Canoming, I'll check it and merge if everything is good.

@quantshah I'd say we let the test wait for a while. @jakelishman 's refactoring is still not merged and adding tests now will lead to conflicts. He will need to rewrite the test again. Maybe we can build a more thorough test for the circuit module during the GSoC project @sarsid ?

BoxiLi:
I wonder why does `remove_gate` still appears in the changes? It is already fixed in #1242. But Git doesn't raise any merge conflict.

jakelishman:
Don't worry about the tests - progress on merging the refactor is slow (though if you want to review #1249 it'll help!).  Besides, these tests would probably go in `test_qubitcircuit.py` which I haven't touched yet.  Add the tests here, and I'll rebase #1251 onto `master` after this is merged if it's necessary to add them to the new style.  Hopefully this will be the last time!

About the merge oddness: it comes about because this PR is built on a false version of the upstream (our) `master`.  As long as you apply this merge as a squash, it will actually be ok and the history will be clean.  Don't apply the merge as a history-preserving merge (i.e. a regular one), because it will make the commit history confusing (and defeats the purpose of having merged #1242 as a squash).


## Unnecessary detail

The exact reason is that this commit is build on top of #1242, but this is inconsistent with the upstream `master`; because #1242 was merged as a squash, `qutip/master` and `Canoming/master` diverged at fd13ae0 (i.e. before #1242), so `git` sees commits ea1c129 to b4f75cf as "new" (the first three commits of this PR). However, when those commits are taken together, they make the same changes as commit 4102b99 (the one which merged #1242) in `qutip/master`.

The Github "Files changed" dialog is a pretty-printed version of the `git` command `git diff qutip/master...Canoming/master`, which means "show all changes on either branch since `qutip/master` and `Canoming/master` diverged".  Because they diverged at fd13ae0 (i.e. _before_ #1242 was merged), then the diff includes the changes.

The merge completes without conflict, however, because `git` is clever and recognises that the changes introduced in the two branches `qutip/master` and `Canoming/master` are identical, even though the commits are different.  As a consequence, however, merging this PR without squashing it would cause commits ea1c129 to b4f75cf to suddenly appear in our commit history, which we don't want.


### If Canoming wanted to fix this (not necessary)

First run
```
git remote add qutip https://github.com/qutip/qutip.git
git fetch qutip
```
to add the upstream `qutip` remote.  Then run
```
git checkout master
git rebase -i $(git merge-base qutip/master master)
```
and modify the file so that the instruction list looks like (only the first word of each line changes)
```
pick ea1c129a fix function QubitCircuit.remove_gate
fixup 3bd94603 fix function QubitCircui.remove_gate
fixup b4f75cfe fix
pick 4aa90f56 fix bug in QubitCircuit.add_circuit
pick 3a2676de fix function QubitCircuit.add_circuit
pick 58aebc7f fix indent
```
Should you save and close the file, it would modify your commit history.  We'd only do this to avoid a merge conflict which would otherwise occur in the next step. Then, run
```
git rebase qutip/master
```
which should complete without error.  Finally, your local `master` would now have diverged from Github's copy (and this PR's), so you'd have to force-push the changes by doing
```
git push --force
```
which would update this PR accordingly.

Canoming:
@jakelishman Thanks for the instructions. I'll try to clean it up after it's merged

BoxiLi:
@jakelishman, thanks for the explanation! I forget the tests here belong to the circuit module.

@Canoming Would you still like to add some tests to `tests\test_qubitcircuit.py` for `remove_gate` and `add_circuit`? Otherwise, I can also merge this and we will update the tests later.

jakelishman:
@Canoming: if this gets merged, it's much easier to bring your `master` up-to-date because we'd want to overwrite everything on it.  In that case, it would be as simple as
```
git remote add qutip https://github.com/qutip/qutip.git
git fetch qutip
git reset --hard qutip/master
```
and you'd have forced your `master` to mirror ours (destructively - if you had additional work you wanted saving, this would overwrite it).

In the future, it will be a bit easier for you if you develop pull requests on a separate branch, and always keep `master` as mirroring ours.

Canoming:
@BoxiLi Sure, the tests looks not too complex. I'm working on it. I'm grouping the gates for better error handling.

@jakelishman Thanks. Sorry, I was just too lazy to set up the tests on my machine, LOL. I simply create the pull request and see if the code pass. That's why there are always one or two minor fixes here.

qutip/qip/circuit.py
```diff
@@ -60,6 +60,16 @@ def DisplaySVG(data, *args, **kwargs):
...
+__single_qubit_gates = ["RX", "RY", "RZ", "SNOT", "SQRTNOT", "PHASEGATE",
+                        "X", "Y", "Z", "S", "T"]
+__para_gates = ["RX", "RY", "RZ", "CPHASE", "SWAPalpha", "PHASEGATE",
+                "GLOBALPHASE", "CRX", "CRY", "CRZ"]
+__ctrl_gates = ["CNOT", "CSIGN", "CRX", "CRY", "CRZ", "CY", "CZ",
+                "CS", "CT"]
+__swap_like = ["SWAP", "ISWAP", "SQRTISWAP", "SQRTSWAP", "BERKELEY",
+               "SWAPalpha"]
+__toffoli_like = ["TOFFOLI"]
+__fredkin_like = ["FREDKIN"]
```
Canoming, line 72: I grouped the gates for exception handling. Hopes that doesn't conflict with #1249 . I think it should be fine, since nothing is affected outside `Gate` class. If it works, I'll use the list to clean up `QubitCircuit`. It looks messy.

BoxiLi: It won't lead to any conflict, no worries. It is a big help grouping the gates!

Canoming:
Does anyone has an idea about what happened to the conflict? I only changed 2 lines, but git lists a bit chunk of conflicts. There are no changes to the `qip/circuit.py` after my last pull request #1242 . I don't see where the conflict comes from.

jakelishman:
After you created a fork to work on #1242, a different PR (#1226) was merged into `master`.  It so happens that #1242 and #1226 don't conflict so that merge succeeded, but the new changes you're now trying to make do.  This is fundamentally a true merge conflict and so unfortunately there's no fancy way around it, though it could have been avoided if you'd matched the state of our `master` before beginning work here, and it would be a bit easier for you for future PRs to branch off before starting work (GitHub has [some nice material about this workflow](https://guides.github.com/introduction/flow/)).

I've rebased your PR onto our `master` and fixed the resulting merge conflict for you.  You should _immediately, without doing anything else_ reset your branch state to match mine, and then force-push your changes to GitHub to update the PR.  To do this, do
```bash
git remote add jakelishman https://www.github.com/jakelishman/qutip.git
git fetch jakelishman
git checkout master
git reset --hard jakelishman/Canoming-circuit-patch
git push --force
git remote remove jakelishman
```
This will destroy anything on `master` that you may have but haven't yet pushed.  You should do it anyway, because a lot of the underlying structure has changed from underneath you, and your changes aren't consistent with the current state of `circuit.py`.

Please also note that the tests are failing after your modifications in (the current) 10e457e.

Canoming:
@jakelishman Oh, thanks. I've cleaned it up. Everything should be working now.

qutip/qip/circuit.py
```diff
@@ -383,9 +399,8 @@ def add_circuit(self, qc, start=0):
                               gate.arg_label)
             elif gate.name in ["BERKELEY", "SWAPalpha", "SWAP", "ISWAP",
```
BoxiLi, line 400: Maybe here we can also use the nice `_swap_like` category you defined?

Canoming: Sure. I'm working on modify the if/else here. I just found more bugs, like the `SWAPalpha` is a parameterized gate, but its parameter will be set to `None` after `add_circuit`. I need to double check if there are more bugs.

But another thing worries me is that, do we really need these if/else? The parameter of an unparameterized gate is already `None`. It looses nothing to pass `None` to the new gate. What we really need is to check if `targets` or `controls` are `None` right? The type of gate doesn't matter here.

BoxiLi: Indeed, it could be that we don't need those if/else at all. It seems to me that we can `deepcopy` the Gate and then modify only `targets` and `controls`. Maybe I'm missing something.

BoxiLi:
Besides, it would be great if we can get this merged before June as there is going to be intensive development on the circuit module from then on by @sarsid.

@sarsid, having a brief look at what this PR has changed might help prevent conflicts, in case you are already working on something :)

BoxiLi:
@Canoming, I'll merge this one now. If you have any new update, you are welcome to open another PR. 

I also recommend to open a new branch for further changes and keep your master up to date with the qutip master.

Issue #1243, opened 2020-05-02
A scheduler for the quantum gates compiler


BoxiLi:
**Is your feature request related to a problem? Please describe.**
In the device simulator of QuTiP (`qutip.qip.device`), a quantum computing device is characterized by the class `Processor` and its subclasses. It contains information about the available control pulses. In addition, a compiler is used to compile a set of universal quantum gates (e.g. RX, RY and ISWAP) into the corresponding control pulses. Together, they can automatically compiler an arbitrary quantum circuit into control pulses, create the time-dependent Hamiltonian representation and run the simulation using QuTiP solvers.

However, in the current version, this compilation is done for each gate sequentially. At each time step, there is only one control pulse active. In practice, it is inefficient because usually several gates can be performed parallelly (e.g. X gate on qubit 0 and qubit 1). To make the compilation more practical, it is necessary to introduce a scheduler that schedules the control pulses. A similar example is https://qiskit.org/documentation/apidoc/pulse.html#schedules.

**Describe the solution you'd like**
There is existing research (https://arxiv.org/pdf/1708.00023.pdf) on gate-level scheduling. It studies which gates can be performed at the same time. In the most general case, however, gate-level scheduling is not enough, since different gates often have different runtime (.e.g. CNOT usually takes longer time than X gate).

Despite the difference, pulse-level scheduling can be done in a similar way as gate-level scheduling by generating a graph (see the reference above). The difference is that each pulse will have its own duration.

The possible implementation in `qutip.qip` could be in two steps:
- Modernize the current compiler and create an `Instruction` class representing each compiled control instruction (pulse, qubit index, amplitude, duration).
- Implement the scheduling algorithm similar as in the above reference.

PR #1273, opened 2020-05-24
A quantum gate(pulse) scheduler

BoxiLi:
**Description**
A gate (pulse) scheduler for quantum circuits (instructions). It schedules a `QubitCircuit` or a list of instructions to reduce the total execution time by exciting them parallelly. It uses heuristic methods mainly from https://doi.org/10.1117/12.666419. The scheduler aims at pulse schedule and therefore does not consider merging gates to reduce the gates number. It assumes that the input circuit is optimized at the gate level (e.g. has a minimal number of CNOT gates) and matches the hardware topology.

The scheduler includes two methods, "ASAP", as soon as possible, and "ALAP", as late as possible. The later is commonly used in quantum computation because of the finite lifetime of qubits.

**Methods**

- The scheduler first generates a quantum gates dependency graph, containing information about which gates have to be executed before some other gates because they do not commute. The graph preserves the mobility of the gates, i.e. commuting gates are not dependent on each other, even if they use the same qubits. 
- Next, it computes the longest distance of each node to the start and end nodes. The distance for each dependency arrow is defined by the execution time of the instruction (By default, it is 1 for all gates). This is used as a priority measure in the next step. The gate with a longer distance to the end node and a shorter distance to the start node has higher priority.
- In the last step, it uses a list-schedule algorithm with hardware constraint (e.g. gates using the same qubits) and priority and returns a list of cycles for gates.
- Since it's heuristics, sometimes it does not find the optimal. It offers an option that randomly shuffles the gates (if commuting) and repeats the scheduling a few times to get a better result.

For scheduling quantum instructions, of which the runtime of different gates might differ, an additional step is required to compute the scheduled start time of each instruction.
It adds the additional dependency caused by hardware constraint to the graph and recomputes the distance of each node to the start and end node. This distance is then converted to the start time of each instruction.

**Example**
```python
>>> from qutip.qip.circuit import QubitCircuit
>>> from qutip.qip.scheduler import Scheduler
>>> circuit = QubitCircuit(7) 
>>> circuit.add_gate("SNOT", 3)  # gate0
>>> circuit.add_gate("CZ", 5, 3)  # gate1
>>> circuit.add_gate("CZ", 4, 3)  # gate2
>>> circuit.add_gate("CZ", 2, 3)  # gate3
>>> circuit.add_gate("CZ", 6, 5)  # gate4
>>> circuit.add_gate("CZ", 2, 6)  # gate5
>>> circuit.add_gate("SWAP", [0, 2])  # gate6
>>>
>>> scheduler = Scheduler("ASAP")
>>> scheduler.schedule(circuit, gates_schedule=True)      
[0, 1, 3, 2, 2, 3, 4]
```
The result list is the cycle indices for each gate. It means that the circuit can be executed in 5 gate cycles:
```
gate0
gate1
gate3, gate4
gate2, gate5
gate6
```
Notice that gate3 and gate4 commute with gate2, therefore, the order is changed to reduce the number of cycles.

**Related issues or PRs**
Solve #1243 

**Changelog**
Add a gate(pulse) scheduler for quantum circuit

**Others**
@sarsid Maybe you will be interested in it. It can be used to determine which gates can be executed at the same time. So maybe we can improve the efficiency of the QubitCircuit by combining several gates together and generate the full matrix using `qutip.qip.expand_operation`. However, writing such scheduler in Python is usually not the most efficient choice and there will be some overhead. I'm not sure if it will bring any real improvement in circuit simulation and where is the threshold.

I'm also pinging @tehruhn because I remember you mentioned something similar before (apologize if not).

More tests are on the way, but you are very welcome to play with it a bit and see if there is any bug!

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -150,12 +150,12 @@ class Gate:
         if name in _single_qubit_gates:
             if self.targets is None or len(self.targets) != 1:
                 raise ValueError("Gate %s requires one target" % name)
-            if self.controls is not None:
+            if self.controls:
                 raise ValueError("Gate %s cannot have a control" % name)
         elif name in _swap_like:
             if (self.targets is None) or (len(self.targets) != 2):
                 raise ValueError("Gate %s requires two targets" % name)
-            if self.controls is not None:
+            if self.controls:
                 raise ValueError("Gate %s cannot have a control" % name)
         elif name in _ctrl_gates:
             if self.targets is None or len(self.targets) != 1:
--- a/qutip/qip/scheduler.py
+++ b/qutip/qip/scheduler.py
@@ -30,10 +30,12 @@
 #    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
-from qutip.qip.circuit import QubitCircuit, Gate
 from collections import deque
 from copy import deepcopy
 from functools import cmp_to_key
+from random import shuffle
+
+from qutip.qip.circuit import QubitCircuit, Gate
 
 
 class Instruction():
@@ -221,7 +223,8 @@ class InstructionsGraph():
                 pass
         self.start, self.end = self.end, self.start
 
-    def find_topological_order(self, priority=True, apply_constraint=None):
+    def find_topological_order(
+            self, priority=True, apply_constraint=None, random=False):
         """
         A list-schedule algorithm, it
         finds the topological order of the directed graph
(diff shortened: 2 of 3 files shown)


BoxiLi:
Hi @sarsid, if you are going to play with it. It would be very helpful if you could give some feedback ;) Like if the doc is clear or if the interface is reasonable, or if there are any bugs. Thanks!

ajgpitch (commented):
This seems like a very good addition to qutip.qip. It seems very well implemented. I have made a few picky suggestions/comments, as you would expect.

ajgpitch:
Is there a notebook to accompany this new feature? And has user documentation been written?

BoxiLi:
@ajgpitch Here is the notebook: https://github.com/qutip/qutip-notebooks/pull/118

Nothing has been added to the documentation yet. That's a good point.

ajgpitch:
As there is a good notebook accompanying the feature, then I am happy for this to be merged without specific user docs.
@BoxiLi perhaps add user docs for this feature to your todo list :)
@BoxiLi Please go ahead and merge when the docstring fixes are done. Nice work, thanks.

Issue #1268, opened 2020-05-15
Make testing more robust especially for random objects


nathanshammah:
**Is your feature request related to a problem? Please describe.**
A lot of issues in test fail seem to arise from random objects (as well as low-level math in MKL and/or cython issues). 

**Describe the solution you'd like**
There are several options at hand. 

### stick to pytest and be creative
A possible fix in testing may be to add randomly generated data that then is pointed at or fix seeds.
**Pros**: fast (?)
**Cons**: technical debt.

### pytest-randomly plugin
[pytest-randomly](https://github.com/pytest-dev/pytest-randomly) is a pytest plugin that addresses this kind of issues. It allows to control `random.seed`, rather than `numpy.random.seed`.  
**Pros**: pytest plugin, supports doctest. 
**Cons**: not super popular, not designed for numpy
  
### property-based testing with Hypothesis
Hypothesis is a library that aims at changing the way tests are designed, allegedly: it should go from testing an instance to designing a test that applies to a domain of instances ([property-based testing](https://hypothesis.works/)). It is not super clear to me right now.  
It contains various randomness-related features, including a [seed function](https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.seed).

**Pros**: sounds powerful and clever, popular and growing, well documented, more robust even beyond this randomness problems.
**Cons**: radical change of testing framework (?), steep learning curve (?), overkill (?).

I also admit I used nose until recently / used with pytest tests thought for nose, without taking advantage of pytest full power.

nathanshammah:
"Assigned" this just to stimulate discussion...

jakelishman:
I think `hypothesis` is the best method here in the long term, but it will most likely have to be a long-term goal.  I think the main pro in favour of it is that it actually is making an attempt to remove randomness; it's attempting to comprehensively test a spanning set of input parameters, rather than just Monte-Carlo'ing our way through and hoping. There's a couple of points which make it difficult to implement:

1. QuTiP can be quite fragile with respect to unexpected input formats, particularly in older parts of the code.
2. Various components are only accurate up to some tolerance, and the error propagation to work out how that corresponds to useful measurable quantities can be rather tricky.

Those are certainly both solvable problems, and point 1 in particular is just general improvement of usability.  The second point is about designing the tests well, which again is certainly doable, but will take a while (it takes long enough just to refactor them, let alone a total rewrite of large chunks of them!).

hristog:
Hi @nathanshammah,

[...]

Issue #1217, opened 2020-03-25
Faster automated test


Ericgig:
Travis only runs 5 test suite in parallel. Each tested commit have 6 to 8 test suite (add 3.8 and 3.8 for mac) and each test take 20 to 30min. So if many of us are working at the same time, automated test can take a few hours to run (like today).
I propose:
1) Run the short suite for older python version (3.6 and 3.7), keeping the full test for only the latest for each os (linux + mac).
2) For the specialised test (mkl, openmp, no cython) only run test affected by it.

While this will probably not speed up for when only one PR is done, it should help when many are lined up.

(2) could be done by adding new marks for the tests. @jakelishman, will this cause conflict with your unmerged PRs.

jakelishman:
For point 2: adding marks right now will cause conflicts, but if #1181 gets merged then it'll be much simpler.  The main issue is that a lot of the testing structure has changed for tests up to `test_mcsolve.py` alphabetically, and so in general the tests are just not the same.

The principle is pretty straightforwards - just like #1181 introduces a `requires_cython` mark, we would add `uses_cython`, `uses_mkl` and `uses_openmp` marks.  These would be even simpler to add than `requires_cython` because unlike it, they don't require any test-generation logic at all.

The change would simply be to add them to `pytest.ini`, mark the requisite functions and then change the pytest invocation of the specialised test to `pytest -m "uses_cython"` or whatever.  As far as I'm aware, the only parts of their codebase which modify their behaviour based on whether `Cython` is available is `QobjEvo`.  The other parts (like `brmesolve`) just don't work if Cython isn't there.

The MKL tests would become very short; by my reckoning, MKL is only used by `steadystate` and `countstat`.  OpenMP is much more pervasive, as `sesolve`, `mesolve` and `brmesolve` all use it, and parts which depend on those are generally the slowest tests.

The short tests could just become `pytest -m "not slow"`?  That's not super speedy, but it takes about half the time, maybe.  If you want more, then a good starting point is `pytest --durations=0` which will tell us which tests are the slowest (perhaps that was what you already did when marking the slow ones?).

Ericgig:
I am hoping to save more time with "not slow", I am hoping for about a 3~4x faster. Maybe I will remove 3.6 to save more time. Other test that need mark will wait for #1181 for now.

jakelishman:
[...]

PR #1274, opened 2020-05-25
density matrix measurements, comp basis measurements, circuit run

sarsid:
**Checklist**

- [ ] If the behavior of the code has changed or new feature has been added, please also update the [documentation](https://github.com/qutip/qutip-doc) and the [notebook](https://github.com/qutip/qutip-notebooks). Feel free to ask if you are not sure.

**Description**
This PR is supposed to mark on-going additions to a new class `qutip.qip.circuit.Measurement`  as well as an elementary `qutip.qip.circuit.Measurement.QubitCircuit.run()` function to exactly simulate quantum circuits. The major functions added right now are:

- `Measurement.density_measurement(self, measurement_ops, state)` : This adds generalized state/density matrix measurements given a list of observables. Still needs addition of checks to see if the list of observables is valid 
- `Measurement.measurement_comp_basis(self, state)` : This adds measurement statistics for specific qubit measurements in the computational basis and also returns the collapsed states (without changing the dimensions). 
- `QubitCircuit.run(self, state, cbits)`: Given a ket input to the circuit, applies gates and measurements from the circuit and returns the resultant ket. 

Right now, the measurement module is very bare-bones and splintered between the previous code 
from @hodgestar and the new code added by me. The eigenstate based measurements are slightly different than the measurements usually used on circuits (or I might be missing something). Ofc, in the case of measurements on all qubits, they amount to the same result. I have not added tests yet but I have added a basic quantum teleportation notebook. It will be nice to get feedback on testing methods as well as what kind of notebooks the community is interested in. Also, any other feature requests on this module would also be appreciated. 

PS: I have also added measurement code from Simon's PR here so maybe we can either add him as a contributor to this PR or figure out something else. 
 
Update: 

1. Addressing some concerns about the PR containing multiple diverging pieces of code, I think classical registers, measurement on the circuit and the classically-controlled gates are best grouped together as they all rely on each other in some ways. The valid concern is that of the Measurement class encapsulating all projective style measurements. I also want to point out that it was raised by @BoxiLi  that we may want to move the Gate class and Measurement class completely to qutip.qip.operations to which I agree but I believe that should be addressed in a clean-up PR with naming changes suggested by @ajgpitch. 

**Related issues or PRs**
Adds to #1090

**Changelog**
Adds partial and complete measurements to state vectors/density matrices and a basic run function


BoxiLi:
This might be helpful in adding @hodgestar as co-authors in a commit. https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

hodgestar:
Hi @sarsid! I'd like to recommend that we land #1090 separately. It's a stand alone piece of work and coupling PRs together into big PRs where it can be avoided generally makes things harder to review and land.

I can show you how to merge that branch into this one if you need to it make progress before #1090 lands.

I would also recommend making the formatting changes you have made (I assume using black or your editor?) in a separate PR to this one. Lumping them into one PR also makes reviewing harder since one doesn't know which changes are intentional or meaningful.

I'm also happy to help with this if you need!

I've started work on the docs for #1090, so hopefully it can land this week.

sarsid:
I think I agree that the two PRs should be kept separate. Initially, I wanted to make them the same since I was trying to use the code in #1090 but right now the two don't interact much. 

Wrt the formatting changes, they were put inadvertently and I'll make sure to revert them but thanks for pointing them out. Let me know what you think of the code in this PR and if you think there's some redundancy between the two (I think there should be but I can't quite put a finger on it).

ajgpitch:
I have now merged #1090. 
@sarsid  @hodgestar good to see you two working together on this. 
We would like as much as possible that is generic to measurement of quantum systems to be in the qutip core, with then only what is specific to QIP circuits being in the qip subpackage. Perhaps you two can continue to develop on this?
We can have qubit specific functions in the qutip core if that is necessary to preserve execution speed.

sarsid:
I think I am pretty happy now with all my functions and variable names (for now). I would say we should leave the variable name changing and moving gates and measurement to separate files for a different PR (it can just be a restructuring one !). I was wondering how to check if it passes current tests (I am not sure what changes were there on Travis).

BoxiLi:
You can either run it on your local machine by the command `pytest` or turn it from a draft PR into a ready-for-reveiw PR, which will trigger the Travis CI tests.

ajgpitch:
I will reiterate @BoxiLi suggestion. If you find that tests pass locally, then switch from draft to review ready. You can also set the 'ready for review' label with your group membership permissions. I will take a look over the code now, but will not fully review it until you mark it as ready for such. Would be best if tests were in place before this.

Take a look at the codeclimate issues. They don't have to fixed, they are just suggestions.

ajgpitch:
I am happy with the general style of the code.
It seems like this PR implements classical control as well as measurement, or am I missing something? As far as is possible PR should be self contained to one additional feature on an existing sub-package.
I am also not sure about having the `Measurement` class in the `circuit.py` file. We should try to keep to smaller code module files where possible.

sarsid:
Some Updates: 

1. I have added some tests for both the Measurement (in qutip/tests/test_circuit_measurement.py) portion and the circuit run (qutip/tests/test_qubitcircuit.py) portions of the code. There is a slight hiccup with the Python 3.8 run of one of the tests (test_run_statistics). I am not sure why it is only failing in the Python 3.8 case. 
2. I have converted all the tests in qutip/tests/test_qubitcircuit.py to pytest style tests. Maybe, @jakelishman  can comment if further changes are required. 
3. Finally, since `measurement_density` and `measurement_ket` allow for POVM style measurements, it might be better to move those functions to qutip/measurement.py but I am not sure. Maybe @hodgestar can comment on that.

BoxiLi:
> I am not sure why it is only failing in the Python 3.8 case.

Well, it's a statistical average, so it can deviate. I don't think it has anything to do with Python 3.8. But probably it can be avoided? I guess its the same question as #1268 

I'm thinking, what about adding a `targets` parameters to `QubitCircuit.run_statistics`. Usually, people don't interested in the full output state of the circuit because many of them are ancillary qubits. They can use `targets` to specify what is the qubits they want to look at. In the function, we can use `ptrace(state, targets)` to trace out the ancilla.

sarsid:
> > I am not sure why it is only failing in the Python 3.8 case.
> 
> Well, it's a statistical average, so it can deviate. I don't think it has anything to do with Python 3.8. But probably it can be avoided? I guess its the same question as #1268
> 
> I'm thinking, what about adding a `targets` parameters to `QubitCircuit.run_statistics`. Usually, people don't interested in the full output state of the circuit because many of them are ancillary qubits. They can use `targets` to specify what is the qubits they want to look at. In the function, we can use `ptrace(state, targets)` to trace out the ancilla.

Seems like a good idea, maybe we can add it as optional parameter to both `QubitCircuit.run_statistics` and `QubitCircuit.run`.

Also, regarding Python 3.8, I was saying because both times it only failed in that test run, also never failed in any of my runs. Should I make tolerance higher ?

qutip/qip/circuit.py
```diff
@@ -848,6 +1122,101 @@ def _basis_SQRTISWAP(self, qc_temp, temp_resolved):
...
+                returns the number of times each state is observed as
+                output state on running the circuit num_runs times.
+        '''
+
+        state_freq = defaultdict(int)
+        tuple_to_state = defaultdict()
+
+        for i in range(num_runs):
+            final_state = self.run(state, cbits=[])
+            state_freq[tuple([complex(a) for a in final_state.full()])] += 1
```
BoxiLi, line 1209: This seems to be quite vulnerable to floating-point error and global phase. The ideal way would be to check if the fidelity of two states is smaller than some threshold.

BoxiLi:
> Also, regarding Python 3.8, I was saying because both times it only failed in that test run, also never failed in any of my runs. Should I make tolerance higher ?

Ok.... Let me try restart the CI test once

qutip/qip/circuit.py
sarsid: Good idea, will do ! I guess the state output can just be the first instance (I'm assuming it doesn't matter much what the small deviations are)

jakelishman:
Consider what you're doing from a statistical sense here.  If everything works correctly, then you're effectively trying a two-tail hypothesis test that your test binomial distribution has a probability of 0.25, given you made `n` observations of it.  The analytic success rate of your test is then
```python
import scipy.stats
def success(n, p, tol):
    dist = scipy.stats.binom(n, p)
    return dist.cdf(n * (p+tol)) - dist.cdf(n * (p-tol))
```
where `success(4096, 0.25, 0.02)` is 99.67%.  Since there are 5 independent runs on Travis, that's equivalent to a failure rate of 1.6% on every single CI run, which is several orders of magnitude too high.  It's probably just luck that it was Python 3.8 on both occasions - it's a one-in-five chance that in two failing CI runs, it was the same setup both times.

You have to consider what is most appropriate to fix the test, and what you can reasonably test in ~1 second of runtime.  You don't want to increase the tolerance too much, because then you can get a lot of false positives even if something is broken.

If you can bump the number of runs up to 100,000 and set the tolerance at 0.01, you'll have a per-test failure rate of ~3e-13, which is more like what we'd want.  If you keep it at 4096, the tolerance should be more like 0.05 (which is pretty big tbh).

sarsid:
> Consider what you're doing from a statistical sense here. If everything works correctly, then you're effectively trying a two-tail hypothesis test that your test binomial distribution has a probability of 0.25, given you made `n` observations of it. The analytic success rate of your test is then
> 
> ```python
> import scipy.stats
> def success(n, p, tol):
>     dist = scipy.stats.binom(n, p)
>     return dist.cdf(n * (p+tol)) - dist.cdf(n * (p-tol))
> ```
> 
> where `success(4096, 0.25, 0.02)` is 99.67%. Since there are 5 independent runs on Travis, that's equivalent to a failure rate of 1.6% on every single CI run, which is several orders of magnitude too high. It's probably just luck that it was Python 3.8 on both occasions - it's a one-in-five chance that in two failing CI runs, it was the same setup both times.
> 
> You have to consider what is most appropriate to fix the test, and what you can reasonably test in ~1 second of runtime. You don't want to increase the tolerance too much, because then you can get a lot of false positives even if something is broken.
> 
> If you can bump the number of runs up to 100,000 and set the tolerance at 0.01, you'll have a per-test failure rate of ~3e-13, which is more like what we'd want. If you keep it at 4096, the tolerance should be more like 0.05 (which is pretty big tbh).

Thanks for the excellent analysis. I don't think it's feasible to do 100,000 given the current. efficiency. It did give me incentive to make it somewhat faster. What is a good guideline for the maximum time a test can take. It seems like I can maybe do 150 runs in ~ 1s. In any case, it doesnt seem feasible to do even 4096 claims without taking quite a bit of time. Any ideas on how to structure the test differently ?

BoxiLi:
Hmmm... 4000/150~25 second. That's not very short actually. The total time for a whole round of qutip CI test costs 15min~20min on Travis. 

Just ideas. Tests here seem to be two-folded:

- Test classical controlled gates in a circuit
- Test the measurement functions

So maybe we can split it. The teleportation circuit test can be done without measurement. Just check the final state tracing out the ancillary qubits. The test for measurement can probably be done for single or two `Qobj` along without circuit. But `run_circuit_statistics`... Probably we then only need to check if the number of elements in the result is correct after some 50 runs?

On a different matter, I'm wondering if running the statistics takes so long, for such a small teleportation circuit, is it still advantageous in any case? Since we are doing simulation and have the full quantum state, one can actually calculate all 4 possible final states, classically mix them into a density matrix with the corresponding measurement probability and calculate the exact statistics distribution. I doubt that will be slower than 25s, although coding will be harder I suppose.

sarsid:
> Hmmm... 4000/150~25 second. That's not very short actually. The total time for a whole round of qutip CI test costs 15min~20min on Travis.
> 
> Just ideas. Tests here seem to be two-folded:
> 
> * Test classical controlled gates in a circuit
> * Test the measurement functions
> 
> So maybe we can split it. The teleportation circuit test can be done without measurement. Just check the final state tracing out the ancillary qubits. The test for measurement can probably be done for single or two `Qobj` along without circuit. But `run_circuit_statistics`... Probably we then only need to check if the number of elements in the result is correct after some 50 runs?
> 
> On a different matter, I'm wondering if running the statistics takes so long, for such a small teleportation circuit, is it still advantageous in any case? Since we are doing simulation and have the full quantum state, one can actually calculate all 4 possible final states, classically mix them into a density matrix with the corresponding measurement probability and calculate the exact statistics distribution. I doubt that will be slower than 25s, although coding will be harder I suppose.

I added the teleportation circuit because it seemed like a simple enough example to test both classically controlled gates and measurements. I have some separate (non-circuit based) examples in the other file. Maybe the run_statistics test can be not on the teleportation circuit and something even simpler ? 

Re: the idea for run_statistics, do you mean tracking the various probability elements during each measurement (along with the state) ? Seems like a decent idea !

BoxiLi:
>I added the teleportation circuit because it seemed like a simple enough example to test both classically controlled gates and measurements. I have some separate (non-circuit based) examples in the other file. Maybe the run_statistics test can be not on the teleportation circuit and something even simpler ?

Maybe, but I find teleportation already a very short circuit, isn't it?

>Re: the idea for run_statistics, do you mean tracking the various probability elements during each measurement (along with the state) ?

Yes, if we have n qubit measurements, we will have 2^n possible final state. One can calculate the state for each of them along with its probability and then get the exact statistics.

jakelishman:
1 second was meant to be the order of magnitude rather than a hard limit, but it's a good rule of thumb for testing one piece of functionality.  [Here's a (slightly out-dated) list of our current longest tests](https://github.com/qutip/qutip/issues/1217#issuecomment-604116373), bearing in mind that some of those ones actually test many many things in the same test function.

The test seems structured ok to me.  Certainly the testing algorithm isn't the major cause of speed loss.  You could try timing some "real-world" examples of using your measurements, and try running them in a profiler to see if there are any obvious ways to speed things up.

With any luck the new data-layer tools and the possibility of using dense `Qobj` storage will give you a fair amount of speed-up once they're complete, since I imagine in your small circuit you're paying quite a lot of sparse matrix overhead.

I'll prefix this last part with: _you should always profile your code before optimising it_.  That said, let me wildly break that rule and guess at where I think you'll find big speed-ups: caching and pre-computation.  If you know you're going to run the same thing loads and loads of times, then you shouldn't have to call `self.propagators` each time, or build `measurement_ops` inside `measure_comp_basis`.  Do those once at the start of `run_statistics`, and save them.

Minor things that _may_ speed up - check the profiling!
1. "compress" the output of `self.propagators`: consecutive propagators with no measurements or classical controls inbetween can be pre-multiplied together at the start to make one matrix.  This will save you ~3 matrix multiplications per run in the test.
2. use more efficient ways to construct large tensor spaces: `qutip.basis`, `qutip.projection` and `qutip.qeye` can all directly constructor product spaces (e.g. `qutip.basis([2, 2, 2], [0, 1, 0])`).  `tensor` is a comparatively expensive operation, and tensoring a list of elements internally is effectively `tensor(...(tensor(tensor(els[0], els[1]), els[2]), ...), els[-1])`.

sarsid:
> 1 second was meant to be the order of magnitude rather than a hard limit, but it's a good rule of thumb for testing one piece of functionality. [Here's a (slightly out-dated) list of our current longest tests](https://github.com/qutip/qutip/issues/1217#issuecomment-604116373), bearing in mind that some of those ones actually test many many things in the same test function.
> 
> The test seems structured ok to me. Certainly the testing algorithm isn't the major cause of speed loss. You could try timing some "real-world" examples of using your measurements, and try running them in a profiler to see if there are any obvious ways to speed things up.
> 
> With any luck the new data-layer tools and the possibility of using dense `Qobj` storage will give you a fair amount of speed-up once they're complete, since I imagine in your small circuit you're paying quite a lot of sparse matrix overhead.
> 
> I'll prefix this last part with: _you should always profile your code before optimising it_. That said, let me wildly break that rule and guess at where I think you'll find big speed-ups: caching and pre-computation. If you know you're going to run the same thing loads and loads of times, then you shouldn't have to call `self.propagators` each time, or build `measurement_ops` inside `measure_comp_basis`. Do those once at the start of `run_statistics`, and save them.
> 
> Minor things that _may_ speed up - check the profiling!
> 
> 1. "compress" the output of `self.propagators`: consecutive propagators with no measurements or classical controls inbetween can be pre-multiplied together at the start to make one matrix.  This will save you ~3 matrix multiplications per run in the test.
> 2. use more efficient ways to construct large tensor spaces: `qutip.basis`, `qutip.projection` and `qutip.qeye` can all directly constructor product spaces (e.g. `qutip.basis([2, 2, 2], [0, 1, 0])`).  `tensor` is a comparatively expensive operation, and tensoring a list of elements internally is effectively `tensor(...(tensor(tensor(els[0], els[1]), els[2]), ...), els[-1])`.

Excellent Suggestions ! I already added some of the speed-ups you suggested, just haven't pushed them yet ! However, I think it might just be a better idea to go with Boxi's idea of removing the need for way too many runs in the first place !

quantshah:
Hi, sorry for coming late to the party. This is a very nice PR and I have a couple of comments on the naming of the functions. Instead of having three different measurement functions - `measurement_ket`, `measurement_density` and `measurement_comp_basis` why not have a single `measure` function that has the same signature as `qutip.measurement.measure`, i.e., measure(op, state) where internally you check if the state is a ket or dm and just apply the operation internally? You could also have a `basis='computational`` keyword if you need to specify the basis. But I do not see why you make that distinction? 

Also, in a real experiment you will not be able to make multiple measurements on the same state which seems to be possible now in your PR. We need to either not allow this or leave a warning message.

qutip/qip/circuit.py
```diff
@@ -3,7 +3,7 @@
 #    Copyright (c) 2011 and later, Paul D. Nation and Robert J. Johansson.
 #    All rights reserved.
 #
-#    Redistribution and use in source and binary forms, with or without
+#    Redistribution and use in sourc e and binary forms, with or without
```
nathanshammah, line 6:
```suggestion
#    Redistribution and use in source and binary forms, with or without
```

ajgpitch:
> Instead of having three different measurement functions - `measurement_ket`, `measurement_density` and `measurement_comp_basis` why not have a single `measure` function that has the same signature as `qutip.measurement.measure`, i.e., measure(op, state) where internally you check if the state is a ket or dm and just apply the operation internally? You could also have a `basis='computational`` keyword if you need to specify the basis. But I do not see why you make that distinction?
> 
In the GSoC meeting we discussed. There may be some efficiency benefits of specific functions for certain tasks. There could be some advantage of an overarching func too. So maybe both is the best approach.

As far as it makes sense the measurement stuff should go it the main package. Anything that is specific to circuits could go in qip. We have suggested @sarsid and @hodgestar discuss and come up with a design.

sarsid:
> Thanks @ajgpitch for the comments. I agree that there should be the specific functions but I think for simplifying user experience there should be a simple overarching function. E.g, when you plot Wigner functions in QuTiP you can send it kets or dms or operators and it just works.
> 
> Internally, it still calls specific functions to compute the Wigner function based on efficient methods specific to the inputs. But it is just a small convenience to the average user to have a simpler clean overarching function to do the same.

I have now made it so that the measure and measurement_statistics function precisely do this. Should we keep either mode (one mode is of the "observable" type and the other with the "projective" type) also as a api-exposed function? If that is not the case, what would be the correct way to write doc_strings ?

sarsid:
> > Instead of having three different measurement functions - `measurement_ket`, `measurement_density` and `measurement_comp_basis` why not have a single `measure` function that has the same signature as `qutip.measurement.measure`, i.e., measure(op, state) where internally you check if the state is a ket or dm and just apply the operation internally? You could also have a `basis='computational`` keyword if you need to specify the basis. But I do not see why you make that distinction?
> 
> In the GSoC meeting we discussed. There may be some efficiency benefits of specific functions for certain tasks. There could be some advantage of an overarching func too. So maybe both is the best approach.
> 
> As far as it makes sense the measurement stuff should go it the main package. Anything that is specific to circuits could go in qip. We have suggested @sarsid and @hodgestar discuss and come up with a design.

I have "proposed" a preliminary design that basically works by adding a dispatch function that appropriately calls either of the two types of measurements. The new concern is writing good documentation for the same. I am wondering if both kinds of measurements should both have their own api-exposed function (or only one function which dispatches to both types, as is right now)

sarsid:
> > I added the teleportation circuit because it seemed like a simple enough example to test both classically controlled gates and measurements. I have some separate (non-circuit based) examples in the other file. Maybe the run_statistics test can be not on the teleportation circuit and something even simpler ?
> 
> Maybe, but I find teleportation already a very short circuit, isn't it?
> 
> > Re: the idea for run_statistics, do you mean tracking the various probability elements during each measurement (along with the state) ?
> 
> Yes, if we have n qubit measurements, we will have 2^n possible final state. One can calculate the state for each of them along with its probability and then get the exact statistics.

I have changed the run_statistics function to now do this method in place of running it a number of times. This solves the problem of slow testing of the function as well !

sarsid:
Some updates:

1. All measurements (except the one in the computational basis) have now been moved to qutip/measurement.py. The interface for the previous measurement style (using observables and eigenstates) remains the same. The only difference is that the `measure` and `measurement_statistics` functions now act as dispatch functions that can carry out projective measurements instead when the `op : list of projection operators` 
2. I have moved all measurement tests to qutip/tests/test_measurement.py as well as  moving all measurement tests to pytest style

sarsid:
I have addressed some of the potential confusion on measurements in my GSOC blogpost [here](https://sarsid.wordpress.com/2020/06/14/an-introduction-to-quantum-measurements-in-qutip/).
Hope that helps and welcoming feedback.

BoxiLi:
Hi @sarsid, there are still some tests failing. When you think the PR is really, just make it a ready-for-review PR and request a review.

Besides, it would be great to address some of the style issues in code climate like "Continuation line under-indented for visual indent". Not all of them must be fixed though, it's more like a guide. The `pycodestyle` package can help you detect them locally if you would like to use it.

sarsid:
Not sure why `../tests/test_interpolate.py::test_usage_in_solvers[mcsolve-complex-spline,string] `
is failing in this case, seems completely unrelated ?

jakelishman:
Yeah, don't worry about that.  That's a pretty unfortunate error (I've never actually seen it before) but it's certainly not related to your changes.

It means that the numerical intergrator we use for solvers couldn't proceed properly (failed to converge in, or that kind of thing).  It will be related to the random seed that `mcsolve` used that time round, but certainly not your fault.

BoxiLi:
I realised that so far for measurement, `ops` must be an operator. But in projective measurement, we can allow the user to give a list of ket as `ops`, and we construct the projective operators for them by `op=ket_op.dag()*ket_op`. This should make things easier for the user.

Also, if we have this, in the doc, we can just introduce measurement by saying that "I have a plus state and want to measure in zero states, what's the probability (and post-measurement state)?". The user don't have to know that we need to construct the projective operator to do this, but rather we do it for them implicitly. Of course we can then explain in the doc further how the projector is constructed as additional information. But this might make the usage easier for their first step.

qutip/qip/circuit.py
```diff
@@ -213,6 +230,87 @@ def _gate_label(name, arg_label):
...
+                        the probability of measuring a state in a the state
+                        specified by the index.
+        collapsed_states : List of Qobjs
+                        the collapsed state obtained after measuring the qubits
+                        and obtaining the qubit specified by the target in the
+                        state specified by the index.
+        '''
+
+        n = int(np.log2(state.shape[0]))
+        target = self.targets[0]
```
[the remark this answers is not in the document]
sarsid: I was intending self.targets to be an array to maintain consistency with self.gates. Measurement must only be on one qubit at a time (atleast right now)

qutip/tests/test_measurement.py
```diff
@@ -71,157 +75,308 @@ def pairs2dm(pairs):
...
+                            SIGMAX, [0.5, 0.5], id="sigmax_ket"),
+                    pytest.param(sigmax(), ket2dm(basis(2, 0)),
+                            SIGMAX, [0.5, 0.5], id="sigmax_dm"),
+                    pytest.param(sigmay(), basis(2, 0),
+                            SIGMAY, [0.5, 0.5], id="sigmay_ket"),
+                    pytest.param(sigmay(), ket2dm(basis(2, 0)),
+                            SIGMAY, [0.5, 0.5], id="sigmay_dm")])
+def test_measurement_statistics_observable(op, state, pairs, probabilities):
+    """ measurement_statistics_observable: observables on basis states. """
+    check_measurement_statistics_observable(op, state, pairs, probabilities)
```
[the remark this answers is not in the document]
sarsid: fixed

qutip/qip/circuit.py
```diff
@@ -385,35 +546,47 @@ def add_circuit(self, qc, start=0):
...
+                elif gate.name in self.user_gates:
+                    self.add_gate(
+                        gate.name, targets=gate.targets,
+                        arg_value=gate.arg_value)
+            else:
+                measurement = circuit_op
+                self.add_measurement(
+                                measurement.name,
+                                targets=[measurement.targets[0] + start],
+                                classical_store=measurement.classical_store)
```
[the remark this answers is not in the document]
sarsid: I am a bit confused. How will we incorporate the shift in targets and controls if we are doing that !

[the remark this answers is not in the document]
Canoming: > I am a bit confused. How will we incorporate the shift in targets and controls if we are doing that !

We don't need to characterize the gate before copy it. All we need is to change the values of targets and controls. It's very simple if the default values are set to `[]`:
```
self.add_gate(gate.name, [target + start for target in gate.targets],
        [control + start for control in gate.controls], gate.arg_value,
        gate.arg_label)
```
By the way, I'm also considering provide more flexibilities in `add_circuit`. We can allow permutation of qubits instead of simply shift them, as in [my module](https://github.com/Canoming/VariationalQCircuit/blob/6447a8fbae09b30e469fb5072ae5dbfec50b2a44/variational_circuit/vcirc/base.py#L224).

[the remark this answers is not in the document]
BoxiLi: Or we copy it first and then change.
```python
new_gate = deepcopy(gate)
new_gate.targets = [t+start for t in new_gate.targets]
new_gate.controls= [t+start for t in new_gate.controls]
```
For `Measurement`, we don't have `controls` though, so only targets.

qutip/qip/circuit.py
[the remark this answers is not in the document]
BoxiLi: I see.

qutip/qip/circuit.py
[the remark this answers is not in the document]
Canoming: It's also possible. But still, even for gates, if the default is `None`, we also need to check whether `[t+start for t in ...]` is valid. I'll suggest to change the default of targets and controls to empty lists. It looses nothing to store `targets` and `controls` as empty lists, even we don't need them. It works fine with measurements.

The only issue is that, it may require lots of dirty works. We have to change those `if gate.controls is None` everywhere. And there could be many (?) guys using this kind of codes in their projects.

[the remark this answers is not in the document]
jakelishman: Just be careful with setting default _arguments_ to mutable types like "empty list".  It's fine for set parameters to be things like that, but they should _always_ be constructed fresh for each instance, otherwise it's quite easy to introduce a problem like this:
```ipython
In [1]: class A:
   ...:     def __init__(self, x=[]):
   ...:         self.x = x
   ...: a = A()
   ...: b = A()
   ...: a.x.append(1)
   ...: a.x, b.x
Out[1]: ([1], [1])
```

Much better to do:
```ipython
In [2]: class B:
   ...:     def __init__(self, x=None):
   ...:         self.x = x if x is not None else []
   ...: a = B()
   ...: b = B()
   ...: a.x.append(1)
   ...: a.x, b.x
Out[2]: ([1], [])
```
or something like this.

By the way, an alternative for the copying could be to give `Gate` and `Measurement` a `copy()` method and a `shift(n_qubits)` method, where each instance then can have its own logic for how this is implemented (e.g. `[el.copy().shift(n) for el in circuit_elements]`).

I'm glad I saw this conversation, though, because I'll absolutely have to check that the new data structures can support deep copy operations - being low-level C extension types, we  store non-reference-counted pointers to memory for speed, and that's anathema to black-box copy methods in a GC environment.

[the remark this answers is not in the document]
sarsid: Excellent catch @jakelishman, I was already making that error with the `cbits` parameter, and although it was probably not an immediate issue ! 

Re: changing the `add_circuit` function, given that it is slightly tangential to the purposes of this PR, maybe we should keep it for a subsequent PR or even package it with the cleanup PR @BoxiLi  @Canoming, what do you think ?

sarsid:
Some updates: 

1. I have added specifying kets as `ops` in `measure` and `measurement_statistics`. I have also added a check for whether the sum of all the measurement operators sums to identity which should be the case with any POVMs. New tests for the same have been added as well ! 
2. I have removed `self.gates` as well as `self.gates_and_measurements`. The replacement is the more "general" (and less of a mouthful) `self.circuit_ops` which contains both. I still allow `propagators` function to be allowed in this case (because it is useful internally in `run`). However, now if a circuit with measurement needs to call `resolve_gates` or `adjacent_gates`, it needs to add measurements after the fact. 

@BoxiLi, any opinions ? I have also addressed all the other minor issues as well !

ajgpitch:
If there are lot of input checks (which is good), then maybe include a `skip_checks=False` kwarg. Maybe you did (sorry didn't check code).

qutip/qip/circuit.py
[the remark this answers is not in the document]
Canoming: @jakelishman Thanks for the comment. The code is still simple even if the default is `None`. 
```python
if new gate.targets is not None:
    new_gate.targets = [t+start for t in new_gate.targets]
```
So, I guess we can just leave the arguments like that.

@sarsid @BoxiLi  I think it's a good idea to leave it to the cleanup PR. It has nothing to do with the performance or new functionality, unless someone is interested in writing the `copy()` functions. That sounds like a hard job.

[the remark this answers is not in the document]
jakelishman: Oh yeah, you don't need to worry for stuff like that - it's fine for `new_gate.targets` to be `[]` and possibly even better, since it's then uniform - I was just going for caution when _setting_ that default in the function signature.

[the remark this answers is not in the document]
BoxiLi: Agreed with all of you, personally I think leaving the default to be `None` is a safer choice and indeed let's leave it to a cleanup PR.

BoxiLi:
Nice work!

> I have removed self.gates as well as self.gates_and_measurements. The replacement is the more "general" (and less of a mouthful) self.circuit_ops which contains both.

I like this merge of to attributes. However, we shouldn't change the attribute name `gates`, at least not in this PR here. It's an open attribute API and as you see we used it a lot in tests. It's very likely that others are using it too. 

Personally, I would keep the attribute `gates` since even for measurement operations, people still usually say "measurement gates". `circuit_ops` is certainly more accurate, but might be hard to remember and thus makes the learning curve steeper.

sarsid:
> Nice work!
> 
> > I have removed self.gates as well as self.gates_and_measurements. The replacement is the more "general" (and less of a mouthful) self.circuit_ops which contains both.
> 
> I like this merge of to attributes. However, we shouldn't change the attribute name `gates`, at least not in this PR here. It's an open attribute API and as you see we used it a lot in tests. It's very likely that others are using it too.
> 
> Personally, I would keep the attribute `gates` since even for measurement operations, people still usually say "measurement gates". `circuit_ops` is certainly more accurate, but might be hard to remember and thus makes the learning curve steeper.

Rolled back to self.gates

BoxiLi:
The PR is in good shape for merging I believe.

I'd like to discuss the parameters API a bit. The parameters used now for measurement function is `ops` and `state`. I guess @hodgestar coined this interface in his PR. It fully made sense for observable and state. However, if we now generalize it to the measurement from the perspective of experimentalists. The parameter `state` becomes a more prominent one whereas `ops` becomes secondary because it only offers a basis for the measurement. This become clearer if one list all the possible use scenarios where:

- `state=Qobj, ops=None`: measure the `state` in default computational basis
- `state=Qobj, ops=ket`: Probability of measure `state` and get `ket`, (together with the post-measurement state)
- `state=Qobj, ops=Qobj`: Measure the `state` with one POVM operator specified by `ops`
- `state=Qobj, ops=list of ket`
- `state=Qobj, ops=list of Qobj`

More generally, we can even provide the user with the optional parameter `targets`, where the given `ops` is only for those targets qubits. We can add identity to the measurement operator for the rest of the qubits automatically for the user.

sarsid:
Updates:
- I have switched the order of `state` and `op`/`ops`. 
- There is now an optional `targets` arguments on each of the measurement functions to enable applying `ops` on particular
`qubits` although this is inherently delicate, especially when dealing with non-qubit type dimensions. Right now, it's restricted to `qubit` style  expansions (in terms of dimensions). This indicated in the docstring. I will add some more information on how to use this in the user guide but I think it works best for more simpler use cases !

sarsid:
Again, I am not sure why some unrelated tests are failing !

BoxiLi:
These failing tests are different from those random failing we've been seeing before. I also have this in my scheduler PR, but it is completely unrelated to the PR. It starts to appear yesterday, but nothing was merged in the last three weeks. Does anyone have a clue?

jakelishman:
`scipy` 1.5 has changed some of its private attributes for matrix multiplication (to a _much_ better name!), and so matrix-matrix multiplication is completely broken for us at the moment: see scipy/scipy@53fac7a.

There's also been some changes to how Hermitian eigenvalues and vectors are calculated (which actually is good news for us in general), which may have had a bit of a knock-on for some of the `zheevr` tests, but I'll check that later.

The new data layer types will fix the matrix multiplication issue permanently, because we'll not be duplicating/reusing large tracts of `scipy` private code - we have our own Cython versions that operate faster on more optimised types.  In the meantime, #1298 is a patch to catch the renames.

jakelishman:
@sarsid, @BoxiLi: the scipy issues are all fixed and merged to master.

BoxiLi:
> @sarsid, @BoxiLi: the scipy issues are all fixed and merged to master.

Many thanks!

sarsid:
@quantshah @BoxiLi  @hodgestar I have wrapped up the two measurement functions in the `measure` and `measurement_statistics` function based on whether `ops` is a  `list of Qobjs` or a `Qobj`. Do we want the individual functions (especially POVM) to be more granular.  Moreover, the docstring for the wrapper function is really clunky (given the case-wise output types). Any ideas on how to make that simpler ?(one idea is to not specify return types and only have parameters given it's bound to be clunky!)

ps. I'll fix the tests all at once after we decide on the api

BoxiLi:
The only concern I have for this API is that now `measure(zero_state, plus_state)` and `measure(zero_state, [plus_state])` will get completely different behavior. The first is observable measurement while the later the usual one. This might be confusing and was also my concern before. Did you come to some conclusion last Thursday? @quantshah @hodgestar @sarsid

qutip/measurement.py
```diff
@@ -169,3 +325,143 @@ def measure(op, state):
...
+            The ket or density matrix specifying the state to measure.
+    ops : Qobj or list of Qobjs
+          - measurement observable
+          or
+          - list of measurement operators M_i or kets
+            Either:
+            1. specifying a POVM s.t. E_i = dagger(M_i) * M_i
+            2. projection operators if ops correspond to
+               projectors (s.t. E_i = dagger(M_i) = M_i)
+            3. kets (transformed to projectors)
```
BoxiLi, line 442: Please highlight the difference also here or in the function documentation:
- if `ops` is list, it is measurement povm or projective
- if `ops` is qobj, it is measurement observerable

Probably also give an extra emphasis in the `rst` doc.

qutip/measurement.py
```diff
@@ -169,3 +325,143 @@ def measure(op, state):
...
+    ----------
+    state : Qobj
+            The ket or density matrix specifying the state to measure.
+    ops : list of Qobjs
+          list of measurement operators M_i or kets
+          Either:
+          1. specifying a POVM s.t. E_i = dagger(M_i) * M_i
+          2. projection operators if ops correspond to
+             projectors (s.t. E_i = dagger(M_i) = M_i)
+          3. kets (transformed to projectors)
```
BoxiLi, line 349: Same here

qutip/measurement.py
```diff
@@ -169,3 +325,143 @@ def measure(op, state):
...
+         The result of the measurement (one of the eigenvalues of op).
+     state : Qobj
+         The new state (a ket if a ket was given, otherwise a density
+         matrix).
+     case : ops is a list of measurement operators (list of Qobjs)
+     index : float
+         The resultant index of the measurement.
+     state : Qobj
+         The new state (a ket if a ket was given, otherwise a density
+         matrix).
```
BoxiLi, line 461: Is this a good way to do it? @nathanshammah @ajgpitch @quantshah @hodgestar

nathanshammah: * please add a line descriptor under each variable, e.g., case (which is repeated twice on line 450 and 456).
* using :class:`qutip.Qobj` would make it clickable and take one to the API. Or even just :class:`.Qobj`.

BoxiLi:
I suggest we merge it after the tests and the codeclimate issues are (partially) fixed. We can make an update PR for doc if necessary.

sarsid:
Addressed all the above concerns, made measurement documentation to be as Alex indicated in email!

BoxiLi:
I find it's good to go. If there is no other big suggestions, I'll merge. Nice work!

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
+        else:
+            collapsed_states.append(None)
+
+    return collapsed_states, probabilities
+
+
+def measurement_statistics_povm(state, ops, targets=None):
+    '''
+    Returns measurement statistics (resultant states and probabilities)
+    for a measurements specified by a set of positive operator valued
```
nathanshammah, line 149:
```suggestion
    for a measurement specified by a set of positive operator valued
```

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
+    measurements on a specified ket or density matrix.
+
+    Parameters
+    ----------
+    state : :class:.Qobj
```
nathanshammah, line 154:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
+            The ket or density matrix specifying the state to measure.
+    ops : list of :class:.Qobj
```
nathanshammah, line 156:
```suggestion
    ops : list of :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
+          3. kets (transformed to projectors)
+    targets : list of ints, optional
+              Specifies a list of target "qubit" indices on which to apply
+              the measurement using qutip.qip.gates.expand_operator to expand
+              ops into full dimension.
+
+
+    Returns
+    -------
+    collapsed_states : List of :class:.Qobj
```
nathanshammah, line 171:
```suggestion
    collapsed_states : List of :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
     """
     Return the measurement eigenvalues, eigenstates (or projectors) and
     measurement probabilities for the given state and measurement operator.
 
     Parameters
     ----------
-    op : Qobj
-        The measurement operator.
-    state : Qobj
+    state : :class:.Qobj
```
nathanshammah, line 208:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
         The ket or density matrix specifying the state to measure.
+    op : :class:.Qobj
```
nathanshammah, line 210:
```suggestion
    op : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -35,55 +35,205 @@
...
+              the measurement using :func:qutip.qip.gates.expand_operator
+              to expand op into full dimension.
+
 
     Returns
     -------
     eigenvalues: List of floats
         The list of eigenvalues of the measurement operator.
-    eigenstates_or_projectors: List of Qobj
+    eigenstates_or_projectors: List of :class:.Qobj
```
nathanshammah, line 222:
```suggestion
    eigenstates_or_projectors: List of :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -106,25 +256,31 @@ def measure(op, state):
 
     Parameters
     ----------
-    op : Qobj
-        The measurement operator.
-    state : Qobj
+    state : :class:.Qobj
```
nathanshammah, line 259:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -106,25 +256,31 @@ def measure(op, state):
...
         The ket or density matrix specifying the state to measure.
+    op : :class:.Qobj
```
nathanshammah, line 261:
```suggestion
    op : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -106,25 +256,31 @@ def measure(op, state):
...
+              the measurement using :func:qutip.qip.gates.expand_operator
+              to expand op into full dimension.
+
 
     Returns
     -------
     measured_value : float
         The result of the measurement (one of the eigenvalues of op).
-    state : Qobj
+    state : :class:.Qobj
```
nathanshammah, line 273:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+    Perform a measurement specified by list of POVMs.
+
+    This function simulates a POVM measurement. The measurement
+    collapses the state to one of the resultant states of the measurement
+    and returns the index of the operator corresponding to the collapsed
+    state as well as the collapsed state.
+
+    Parameters
+    ----------
+    state : :class:.Qobj
```
nathanshammah, line 341:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+    ops : list of :class:.Qobj
+          list of measurement operators M_i or kets
+          Either:
+          1. specifying a POVM s.t. E_i = dagger(M_i) * M_i
+          2. projection operators if ops correspond to
+             projectors (s.t. E_i = dagger(M_i) = M_i)
+          3. kets (transformed to projectors)
+    targets : list of ints, optional
+              Specifies a list of target "qubit" indices on which to apply
+              the measurement using :func:qutip.qip.gates.expand_operator
```
nathanshammah, line 352:
```suggestion
              the measurement using :func:`qutip.qip.gates.expand_operator`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+              to expand ops into full dimension.
+
+    Returns
+    -------
+    index : float
+        The resultant index of the measurement.
+    state : :class:.Qobj
```
nathanshammah, line 359:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+
+
+def measurement_statistics(state, ops, targets=None):
+    """
+    A dispatch method that provides measurement statistics
+    handling both observable style measurements
+    and projector style measurements(POVMs and PVMs).
+
+    For return signatures, please check:
+    - :func:qutip.measurment.measurement_statistics_observable
```
nathanshammah, line 378:
```suggestion
    - :func:`qutip.measurement.measurement_statistics_observable`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+      for observable measurements.
+    - :func:qutip.measurement.measurement_statistics_povm for povm measurements.
```
nathanshammah, line 380:
```suggestion
    - :func:`qutip.measurement.measurement_statistics_povm` for povm measurements.
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+
+    Parameters
+    ----------
+    state : :class:.Qobj
```
nathanshammah, line 384:
```suggestion
    state : :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+            The ket or density matrix specifying the state to measure.
+    ops : :class:.Qobj or list of :class:.Qobj
```
nathanshammah, line 386:
```suggestion
    ops : :class:`.Qobj` or list of :class:`.Qobj`
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+
+
+def measure(state, ops, targets=None):
+    """
+    A dispatch method that provides measurement results
+    handling both observable style measurements
+    and projector style measurements(POVMs and PVMs).
+
+    For return signatures, please check:
+    - :func:qutip.measurment.measure_observable for observable measurements.
```
nathanshammah, line 414:
```suggestion
    - :func:`qutip.measurement.measure_observable` for observable measurements.
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+    - :func:qutip.measurement.measure_povm for povm measurements.
```
nathanshammah, line 415:
```suggestion
    - :func:`qutip.measurement.measure_povm` for povm measurements.
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+
+    Parameters
+    ----------
+    state : :class:.Qobj
+            The ket or density matrix specifying the state to measure.
+    ops : :class:.Qobj or list of :class:.Qobj
+          - measurement observable (:class:.Qobj)
```
nathanshammah, line 422:
```suggestion
    state : :class:`.Qobj`
            The ket or density matrix specifying the state to measure.
    ops : :class:`.Qobj` or list of :class:`.Qobj`
          - measurement observable (:class:`.Qobj`)
```

qutip/measurement.py
```diff
@@ -169,3 +325,115 @@ def measure(op, state):
...
+          or
+          - list of measurement operators M_i or kets (list of :class:.Qobj)
```
nathanshammah, line 424:
```suggestion
          - list of measurement operators M_i or kets (list of :class:`.Qobj`)
```

sarsid:
> Looks fine to me, great work! I added some suggestions that you @sarsid should be able to commit directly. You need to add ticks like "`" and "`" just after :class: and :func: to be picked up by Sphinx. This is a general feature of restructured text.

Thanks for the fixes!

nathanshammah:
Great work @sarsid. (turns out that everytime I was writing `, it did not render here.)

jakelishman:
If useful, you can escape backticks in markdown with a \\, so you can write \\\` to get one.  If you need a single backtick in an inline code block, you can use two backticks to start the code block, so to get `` ` `` you type \`\` \` \`\`.

PR #1275, opened 2020-05-25
Update steadystate.py

NS2LPS:
**Description**
Minor bug fix in steadystate 

**Changelog**
Fix the calculation of the weight to be the actual mean over the rows of the max of the abs of the liouvillian (now it is just the maximum of the abs of all the elements).

Fix the direct dense solver to add weight*T to the first row of the Liouvillian instead of replacing the first row.

### DIFF
--- a/qutip/steadystate.py
+++ b/qutip/steadystate.py
@@ -266,7 +266,7 @@ def steadystate(A, c_op_list=[], method='direct', solver=None, **kwargs):
 
     # Set weight parameter to avg abs val in L if not set explicitly
     if 'weight' not in kwargs.keys():
-        ss_args['weight'] = np.mean(np.abs(A.data.data.max()))
+        ss_args['weight'] = np.mean(np.abs(A.data).max(0))
         ss_args['info']['weight'] = ss_args['weight']
 
     if ss_args['method'] == 'direct':
@@ -477,7 +477,7 @@ def _steadystate_direct_dense(L, ss_args):
     b[0] = ss_args['weight']
 
     L = L.data.todense()
-    L[0, :] = np.diag(ss_args['weight']*np.ones(n)).reshape((1, n ** 2))
+    L[0, :] += np.diag(ss_args['weight']*np.ones(n)).reshape((1, n ** 2))
     _dense_start = time.time()
     v = np.linalg.solve(L, b)
     _dense_end = time.time()


fminga:
Thank for the remark. 
@NS2LPS Replacing the first row of the Liouvillian is a standard way to ensure the algorithm produces a steady state with trace one. Since this condition is, in theory, redundant, adding the weighted first-row or substituting the first line should be equivalent. Often the algorithm is unstable, so one weights the trace 1 condition. Maybe how much of the first row is kept can make the algorithm even more stable?

hodgestar:
@NS2LPS Apologies for the slow reviews. Would you be up for updating this PR and then we can review it properly? It would be good to somehow add tests to these cases, but I'm not sure how easy that is to do.

NS2LPS:
I will update my contribution as soon as I come back from holidays. Best, Jérôme

 -------- Message original --------
 Objet : Re: [qutip/qutip] Update steadystate.py (#1275)
 De : Simon Cross 
 À : qutip/qutip 
 Cc : NS2 group at LPS  ,Mention 

@NS2LPS Apologies for the slow reviews. Would you be up for updating this PR and then we can review it properly? It would be good to somehow add tests to these cases, but I'm not sure how easy that is to do.

hodgestar:
> I will update my contribution as soon as I come back from holidays. Best, Jérôme

Thanks! Enjoy your holidays.

hodgestar:
@NS2LPS Just checking whether you're still keen to write a test for this and finish off the PR?

NS2LPS:
Sorry, my holidays were not that long... Here are some tests of the proposed modifications.

1 ) Computation of the weight
The proposed modification may lead to a large change of the weight in some circumstances.
For example :
```
H = sigmaz()
c_op_list = [1e-8*sigmap()]
L = liouvillian(H, c_op_list)
n = H.shape[0]
b = np.zeros(n ** 2)
print(L.data.todense())

[[ 0.e+00+0.j  0.e+00+0.j  0.e+00+0.j  1.e-16+0.j]
 [ 0.e+00+0.j -5.e-17+2.j  0.e+00+0.j  0.e+00+0.j]
 [ 0.e+00+0.j  0.e+00+0.j -5.e-17-2.j  0.e+00+0.j]
 [ 0.e+00+0.j  0.e+00+0.j  0.e+00+0.j -1.e-16+0.j]]
```

The elements of the Liouvillian have modulus close to unity, which should be reflected in the value of the weight.

```
# Current version
weight = np.mean(np.abs(L.data.data.max()))
print('Weight',weight)
b[0] = weight
Ldense = L.data.todense()
Ldense[0, :] = +np.diag(weight*np.ones(n)).reshape((1, n ** 2))
rho1 = np.linalg.solve(Ldense, b)

Weight 1.0000000000000001e-16

# Proposed modified version
weight = np.mean(np.abs(L.data).max(0)).real # added .real to get a float 
print('Weight',weight)
b[0] = weight
Ldense = L.data.todense()
Ldense[0, :] = +np.diag(weight*np.ones(n)).reshape((1, n ** 2))
rho2 = np.linalg.solve(Ldense, b)

Weight 1.0
```

The current version leads to a very small weight because the imaginary part is discarded. Also, the mean is not doing anything. The proposed modification computes the modulus before evaluating the maximum for each line and then takes the mean. This may change the final steady state in some circumstances, but I could not rapidly find a situation where this is actually important. In this example, the two steady states are the same even though the weights are very different.
```
np.linalg.norm(rho1-rho2)
>> 1.1102230246251565e-16
```


2 ) Construction of the matrix to obtain the steady state
As mentioned by @fminga, this should not change anything in most situations. But the proposed modification will make the matrix identical to the one computed by the algorithm when the sparse option is chosen. Here is an example, where the proposed modification performs slightly better than the current version. 
```
H = identity(2)
c_op_list = [sigmam(), 1e-8*sigmap()]
L = liouvillian(H, c_op_list)
weight = 1.0
n = 2
b = np.zeros(n ** 2)
b[0] = weight
```

The current version gives:
```
Ldense = L.data.todense()
Ldense[0, :] = np.diag(weight*np.ones(n)).reshape((1, n ** 2))
np.linalg.solve(Ldense, b)
>> array([0.+0.j, 0.-0.j, 0.-0.j, 1.+0.j])
```

The proposed modification gives:
```
Ldense = L.data.todense()
Ldense[0, :] += np.diag(weight*np.ones(n)).reshape((1, n ** 2))
np.linalg.solve(Ldense, b)
>> array([1.e-16+0.j, 0.e+00-0.j, 0.e+00-0.j, 1.e+00+0.j])
```
This is a better result than the one obtained with the current version.

NS2LPS:
For the moment, I don't have better arguments than :

- the current calculation of the weight does not make too much sense and does not correspond to its description
``# Set weight parameter to avg abs val in L if not set explicitly``

- the first row of the matrix used to find the steady state does not match the one used in sparse mode

I believe that my modifications cure these two problems. But I would understand that you don't want to implement them since I could not demonstrate that they solve actual problems. I could try to come with a more convincing example if I can find one, but this may take some time.

PS : In my last example, the first population is expected to be 1e-16 because of the 1e-8*sigmap() jump. But I agree
that both versions give the same results within machine precision.

hodgestar:
I've reimplemented this in #1802 rather than attempting to resolve conflicts in this PR and added tests there.

PR #1282, opened 2020-06-03
Core package

jakelishman:
This took longer than anticipated because of a bunch of knock-on nasty Cython problems, and I struggled to get OpenMP running locally on macOS ([but I've got it now](https://www.binhbar.com/posts/2020/06/compiling-openmp-libraries-on-macos/)).

Tag: @Ericgig, @ajgpitch.

---

Separate out what will become the "core" of qutip into its own subpackage.  This package is intended to have most of its public symbols re-imported into the global package namespace.

There should be no substantive changes to any actual worker code in this large commit, just what is necessary to carve out a major entangled part of the core library.

Some files currently in core may be moved out in a future version (e.g. graph, metrics, semidefinite), but right now they are too entangled with `Qobj` or other true-core parts, and to remove them now would require too much substantive change to the code, which would be difficult to review in this commit.

Major changes required as part of this:
 - module-level imports (e.g. `from qutip.qobj import Qobj`) are replaced with simple ones from the namespace (e.g.  `from qutip import Qobj`).  While inside the `core` package, imports are kept as module-specific, largely to support the mess of circular imports present in qutip at the more.
 - .pxi files are no longer recommended as the way to share code between Cython files, so these have been converted into proper Cython modules and added to the distribution.
 - collection of Cython extension components in `setup.py` is consolidated to support them no longer all existing in the same place.  As the data layer is fleshed out, likely there will be even more Cython locations present.
 - the test for OpenMP in `__init__.py` had to be made more indirect via importlib to avoid causing a full import of `qutip.core`, which would in turn cause Qobj and other parts to be imported.  These components required `__init__.py` to have discovered whether OpenMP existed already before they were imported, causing false negatives.


Ericgig:
Good work. 
Removing pxi is great. This was really old cython.

I would have put solvers in core since the core usage of qutip is evolution of quantum systems. But I am thinking more in what would stay in the main package and what would become spin-off (qip, control). 

The OpenMP check seems too complicated for something that should be simple. I would like to remove the need of this flag. But for now, if you set it in the `__init__` of `core` instead of `qutip`, could you then not use the `try` as before?

jakelishman:
Solvers: absolutely the solvers should always be packaged with the main `qutip`.  I had sort of envisaged that we'd put them in a `solve` package, which would be exported into the main namespace just like `core` is.  This way is more just for greater separation, so that the full core isn't entangled with the solvers themselves - I think separating them like this will help us ensure that we specify the data layer API much more cleanly if the solvers aren't referenced within it. (i.e. there's no reference to any solver within `core`, but the API is built sufficiently cleanly that they can still access all the internals).

OpenMP check: I think there's two neater ways:
1. perhaps `settings` should just be in `core` too?  I avoided putting the check in `core/__init__.py` because I'm really trying to avoid bi-directional dependencies between the packages, but moving `settings.py` into `core` would solve that.
2. alternatively, we could just write `settings.py` as part of `setup.py` - determine whether we built against OpenMP/MKL at compile-time, and then just write it in?  We can still choose whether or not to use it in various runtime situations.

Ericgig:
I prefer 2 in for openmp, it is fixed at install. It could be in `setup.py` or in a cython file with conditional compilation.

The mkl flag depends on numpy's blas python interface. Cython use of blas/lapack is independent of it. 

(Yes, since this week, I did not know this existed before)

jakelishman:
Doing option 2 is a bit more involved, but I agree that I think it's the right way of doing it.

For now, I could do option 1 (or just leave as-is) and we can do option 2 properly later?  It likely needs to be part of a bigger design decision with how we handle all compile-time/run-time settings and options.

Ericgig:
You leave it as is for now.
I am happy to merge this like it is and move more stuff around later.
Or do you want to split `solver` also now?

jakelishman:
Let's leave it for now. We don't get any benefit from it right now unlike the enforced separation of the core, and we might change our plans in the future anyway.

PR #1288, opened 2020-06-15
Remove eigh usage on mac 

Ericgig:
**Description**
On mac, with OpenBlas, `eigh` can cause segmentation fault, see #1197. For 4.5.1, we simply skipped the tests causing problems ( #1257). This replace the calls of `numpy.linalg.eigh` to `numpy.linalg.eig`, and lapack calls of `ZHEEVR` to `ZGEEV` when at risk of segfault. This will result in slower `Qobj.eigenstates` and `brmesolve`. 

Tests where adapted, the issues were the phase of the eigenvectors and the lower numerical precision of `eig`.

While the flag to use `eig` instead of `eigh` is `qutip.settings.eigh_unsafe`, it must be set before importing `qutip.sparse` to have any effect. Thus it is not a true settings, (like `has_openmp`).

**Changelog**
Replace calls of `eigh` to `eig` when at risk of segfault.

### DIFF
--- a/qutip/sparse.py
+++ b/qutip/sparse.py
@@ -60,8 +60,9 @@ if eigh_unsafe:
         idx = np.argsort(val)
         if eigvals:
             return (val[idx[eigvals[0]:eigvals[1]+1]],
-                    vec[:,idx[eigvals[0]:eigvals[1]+1]])
-        return val[idx], vec[:,idx]
+                    vec[:, idx[eigvals[0]:eigvals[1]+1]])
+        return val[idx], vec[:, idx]
+    
     def eigvalsh(a, UPLO="L", eigvals=[]):
         val = la.eigvals(a)
         val = np.sort(np.real(val))
--- a/qutip/sparse.py
+++ b/qutip/sparse.py
@@ -54,15 +54,31 @@ from qutip.cy.spconvert import (arr_coo2fast, zcsr_reshape)
 from qutip.settings import debug, eigh_unsafe
 
 if eigh_unsafe:
+    def _orthogonalize(vec, other):
+        cross = np.sum(np.conj(other) * vec)
+        vec -= cross * other
+        norm = np.sum(np.conj(vec) * vec)**0.5
+        vec /= norm
+
     def eigh(mat, eigvals=[]):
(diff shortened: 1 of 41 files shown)


nonhermitian:
FYI, One has to be a bit careful as there is no internal orthonormalization step in eig or the lapack equivalent. This is trouble for degenerate systems. 

Ran into for the same issue in Qiskit.

qutip/tests/test_measurement.py
```diff
@@ -143,64 +156,37 @@ def test_measurement_statistics_input_errors():
...
-    np.random.seed(seed)
-    measurements = []
-    for _ in expected_measurements:
-        value, new_state = measure(op, state)
-        measurements.append((value, new_state))
-    assert_(measurements == expected_measurements)
+def check_measure(op, state):
+    """ Check that all measure results are in measurement_statistics which
+    was previously tested.
+    This remove the depandance on the prng.
```
nathanshammah, line 162:
```suggestion
    This remove the dependance on the prng.
```

qutip/sparse.py
```diff
@@ -51,7 +51,43 @@
...
+        val, vec = la.eig(mat)
+        val = np.real(val)
+        idx = np.argsort(val)
+        val = val[idx]
+        vec = vec[:, idx]
+        if eigvals:
+            val = val[eigvals[0]:eigvals[1]+1]
+            vec = vec[:, eigvals[0]:eigvals[1]+1]
+        same_eigv = 0
+        for i in range(1, len(val)):
```
Ericgig, line 73: I do the orthonormalization raise by Paul Nation in this loop.

Ericgig:
May not be needed anymore with scipy 1.5? To test more before merging (#1299 )

jakelishman:
I now get segfaults on Mac with scipy 1.5.  I think this PR is still needed.

jakelishman:
While this isn't an alternative for this PR, scipy 1.5 made new LAPACK drivers available for `scipy.linalg.eigh`, one or more of which might not have the same problems in OpenBLAS on Mac.  Still, we can't require scipy 1.5 when that only released like 5 days ago, but it's interesting going forwards.

PR #1289, opened 2020-06-16
Settings reset

Ericgig:
**Description**

- Add the `reset` function to `qutip.settings` that is described in the documentation but was never implemented. 
- Add `save`, `load` to `qutip.settings`, allowing to save in alternate files.
- The list of configuration that can be saved in qutiprc is generated when loading the module.
- Ungraded the `configrc` functionalities to allow other filenames and support more than a few hard coded `qutip.settings`'s keys.
- Reading `qutiprc` will raise warning instead of errors when unknown keys. So if we remove keys in v5, old file will not cause crashes when importing qutip for the few that use qutiprc. 

**Changelog**
Add `reset` to `qutip.settings`

### DIFF
--- a/qutip/settings.py
+++ b/qutip/settings.py
@@ -97,12 +97,13 @@ def _valid_config(key):
         return True
     return False
 
+
 _environment_keys = ["ipython", 'has_mkl', 'has_openmp',
                      'mkl_lib', 'fortran', 'num_cpus']
 __self = locals().copy()  # Not ideal, making an object would be better
 __all_out = [key for key in __self if _valid_config(key)]
 __all = [key for key in __all_out if key not in _environment_keys]
-__default = {key:__self[key] for key in __all}
+__default = {key: __self[key] for key in __all}
 __section = "qutip"
 del _valid_config
 __self = locals()
@@ -150,7 +151,7 @@ def reset():
         num_cpus = int(os.environ['QUTIP_NUM_PROCESSES'])
     elif 'cpus' in info:
         import qutip.hardware_info
-        info =  qutip.hardware_info.hardware_info()
+        info = qutip.hardware_info.hardware_info()
         num_cpus = info['cpus']
     else:
         try:
--- a/qutip/settings.py
+++ b/qutip/settings.py
@@ -88,26 +88,27 @@ except:
 
 
 def _valid_config(key):
-    environment_keys = ["ipython", 'has_mkl', 'has_openmp',
-                        'mkl_lib', 'fortran', 'num_cpus']
-    if key in environment_keys:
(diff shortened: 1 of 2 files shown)


qutip/settings.py
```diff
@@ -48,6 +48,7 @@
 # number of cpus (set at qutip import)
 num_cpus = 0
 # flag indicating if fortran module is installed
```
ajgpitch, line 50: We used to have a Fortran version of the MC solver -- long since dead. So we could remove this if we wanted, but no real need to.

ajgpitch (approved):
This looks good to go. @Ericgig please go ahead and merge.

jakelishman:
@Ericgig Eric, is this PR ready for merge?

Ericgig:
Yes

jakelishman:
@Ericgig, Eric: to check, there's nothing in this PR that wasn't also implemented on `dev.major`, right?  (See #1337.)  I'm merging up `master` into `dev.major`, and this PR causes a huge merge conflict, but I'm fairly sure it's just because you had to implement the same thing in two completely separate ways.

Ericgig:
@jakelishman, Equivalent change but with different implementation have been made in `dev.,major`. Here setting variable are directly in the module, but in v5 they are in a `Settings` object. Adding the object was a big change for major.
Maybe cherry-picky commits from master to `dev-major` would be easier.

jakelishman:
Perfect, that's what I thought thanks.

`git cherry-pick` isn't a good choice for merging, because it creates new commit objects.  Doing that means we never explicitly told git how the two implementations should be picked going forwards, and so any additional changes on `dev.major` _also_ could never be merged in, because they'd have different commits as their parents.  Logically it's a merge operation and this is a "true" merge conflict, so the right thing to do is to tell git how to solve the conflict.

I've already got a fixed merge all ready to go - once we've got #1440 merged I'll make the PR up, so `dev.major` is once again completely ahead of `master`.

jakelishman:
tbh, the only real use-case for `git cherry-pick` is to backport single bug-fix commits from `master` onto an already existing release branch without merging in additional feature commits.  Since `dev.major` is intended to _include_ all of `master` (and `master` will eventually point to the `dev.major` tip), we shouldn't cherry pick from one to the other, but merge all changes.  Having both #1289 and #1337 is effectively the same as having cherry-picked the feature onto both branches already, which is why the merge is now a bit tricky.  That's not a problem at all (with how we developed it, it's the correct thing to do), it just means that when I joined them back together, I just merged this bit with the (logical - I did it all in one go) strategy `--ours` to tell `git` to keep the `dev.major` version.

PR #1296, opened 2020-06-22
Partial new data layer with basic dispatch and types CSR and Dense

jakelishman:
Add the `data.CSR` and `data.Dense` types, and begin the port of the current operations of a mixture of `fast_csr_matrix` and `CSR_Matrix` into simple ones on the new `CSR` type.  This is significantly faster at initialisation and interaction between Python and Cython.

Also add the base multiple dispatcher type which will be used in the future to implement data layer multiple dispatch.

@qutip/data-layer-devs.


jakelishman:
Ooo, that's an annoying error in 7732470.  Looks like gcc (which I'm using on Mac) does something different to the default clang.  Should be a straightforward solution.

jakelishman:
(I wasn't able to "assign" @tehruhn for some reason)

Ericgig:
A few comments, question.


1. Why are file separated by function and not by datatype? So adding a new datatype will result in modification im a lot of file making support (PR evaluation) harder.
2. `__repr__` return a very clean output in `Qobj`. `CSR`'s is quite simple. This is because data layers are never meant to be printed directly?
3. For solver states, we mostly use Fortran ordered array, not C ordered. Will you support both?
4. In matmul, you use `malloc`, `calloc`, `free`. It is better to use the python PyDataMem_NEW. I never had any issue, but there can be some on windows: https://github.com/numpy/numpy/issues/8253
5. `mat_mul` can be used inplace, but no `add_csr`. 
6. It would be good to have inplace multiplication with a scalar instead of always making a copy.

Ericgig:
One more.
For dense matrix operation, do you intend to write the code or use blas/lapack? Using the libraries is less work and leave advanced user the ability to link optimized version for there systems, or gpu accelerated one if they want, with no more effort on our side.

jakelishman:
1. Adding a new data type is a large undertaking which will likely be very infrequent, so it's perhaps not so sensible to optimise in favour of making this especially simple.  More likely is that new operations will be added to the data layer more frequently (e.g. you found that you wanted specific things for `CQobjEvo`), and this format makes _those_ PRs much easier to review.  But mostly, because organising by file type is ok when all operations are like `matmul(CSR, CSR) -> CSR`, but which file should the operation `matmul(CSR, Dense) -> COO` go into?  Much easier to find what you're looking for if all `matmul` functions are together, and then setting up the dispatcher will be the last item in the `matmul` file.

2. Exactly - it's more of an internal type.  Currently `Qobj` converts the underlying data object to dense `ndarray` and prints it for its `repr` method - I don't have any plans to change that, so `Qobj` will still be full and user-facing.  My `CSR` repr actually just tells you the same information that `scipy.sparse.csr_matrix` does, just in a slightly more Python-looking form.

3. Yes, in some form or another.  Perhaps we could talk about the best way to do that in a future meeting?

4. `PyDataMem_NEW` isn't actually a Python routine, it's a numpy one, but that's largely irrelevant.  The Python equivalent is `cpython.mem.PyMem_Malloc` and family.  The reason I don't use the latter in general is because sometimes `PyMem_Malloc` allocates into Python-reserved stack-space, and if we subsequently pass the pointer to numpy, it will try to free it and cause a segfault.  The reason for using `malloc` and `free` here is mostly just because there was a comment in the code I copied it from saying that raw `malloc` and `free` were slightly faster for allocating heap space to be released within the same function, so I just did what they did.  I can change it - it's not important.  When allocating space that _may_ be passed to numpy, I always use `PyDataMem_NEW` (or friends).

5. That's a mistake - I meant to take out that ability in `matmul_csr` because it's just asking for trouble (the user won't know how much space to allocate, and we're basically just asking for a segfault).  I did `matmul_csr` first, and I hadn't decided yet exactly what I was doing.

6. Yeah, I'm going to add additional `imul`, `ineg` and `idiv` routines as dispatched operations.

7. LAPACK/BLAS as far as we can.  I've no pretense to being able to write faster code than hardware-specific optimised stuff for matrices.  I haven't actually written most of the numerical code in this PR either - it's just ported from current `qutip`, with the variable names changed to be more descriptive.

Ericgig:
Great, keep the good work.
For 5, in-place multiplication and addition are not for the user, but for other part of the core's cython code. `cqobjevo` use in-place addition. If the state can be a data layer, in-place `matmul`, even if only faster for the dense case, will probably be appreciated. 
`expect` will probably use `mv_csr`, or the any data equivalent, so it would be useful to add it to the .pxd.

jakelishman:
Ok, yeah, I definitely see the use-case here.  I'll leave it removed from `matmul` for the time being because I hadn't really organised that code to allow safe calling of the function with an `out` parameter, but I'll look to re-instate it, similarly with `add`.

For `add`: I envisage a nice possibility for `QobjEvo` using `CSR` backing along the vein of `CQobjEvoTdMatched`: on instantiation we add together all the matrices, then we `memset` all zeroes along the `data` array and store it as `self._structure`.  Then each time we `__call__` the object, we simply do `out = self._structure.copy()`, and use that as the output matrix, because we guarantee we'll always have enough space and the correct structure.  There's a couple of minor kinks in the logic of `add_csr` that we might have to iron out in order to avoid an additional matrix addition, but the savings in memory allocation could be good.

You're absolutely right about the `pxd` files - I haven't taken enough care to make sure they've got all the definitions in yet.  Basically everything that isn't prefixed with an underscore in the `.pyx` files _should_ be in the `.pxd` files.  I'll add them in tomorrow.

Ericgig:
I could not compile the PR with Cython 0.29.14, upgrading it to 0.29.21 worked. Please increase the cython requirement in setup.py and requirements.txt.

jakelishman:
Oh yeah, I forgot about that - it's a bug in Cython with C-tuple handling.

Ericgig:
@jakelishman 
A little late but in test_crs:
```
@pytest.fixture(params=[
    pytest.param((1, 5), id='ket'),
    pytest.param((5, 1), id='bra'),
...
```
Shouldn't `bra` and `ket` be inverted?

jakelishman:
Whoops!  Yeah, absolutely.  I'll fix it in the other PR.  It only affects the names of tests at least.

PR #1298, opened 2020-06-25
Support scipy >= 1.5 in fast_csr_matrix matmul and legacy.ptrace

jakelishman:
Scipy 1.5 changed the methods of performing the matrix multiplication to
avoid copying the `indptr` matrix multiple times.  This is actually not an
issue at all for us, because we only allow int32 indices (so if the
indices were to upcast, we just throw and exception), but the underlying
private Cython call structure also changed, and our `fast_csr_matrix` was
a direct clone.

We add try/catch blocks to switch on the correct types (the overhead
from a failed lookup is trivial compared to the matrix multiplication
time), because we have to support more than just the most recent version
of scipy.  We could look up the version of scipy installed, but the
"try it and see" approach is more typical Python.

This problem should be eliminated by the move to the new data types, as
we will stop relying on private scipy code.

See: commit scipy/scipy@53fac7a

### DIFF
--- a/qutip/fastsparse.py
+++ b/qutip/fastsparse.py
@@ -180,10 +180,10 @@ class fast_csr_matrix(csr_matrix):
         Do the sparse matrix mult returning fast_csr_matrix only
         when other is also fast_csr_matrix.
         """
-        M, K1 = self.shape
-        K2, N = other.shape
+        M, _ = self.shape
+        _, N = other.shape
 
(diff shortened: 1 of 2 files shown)


jakelishman:
This second commit fixes the test error in `test_ptrace.py` - the commit message explains what I've done.

The errors in `brtools` are due to a much more difficult fix to implement.  See #1299.

Issue #1299, opened 2020-06-25
Scipy 1.5 changes to linalg.eigh break brtools zheevr tests


jakelishman:
This is causing test failures in all CI builds against scipy 1.5 currently.  The other failing tests on scipy 1.5 are due to changes to the handling of `csr` matmul operations, which is fixed in patch #1298.

As far as I can see, there is no way to recover the exact same behaviour as existed in scipy < 1.5 (there's optimisations of the `lwork` input to the LAPACK routine), and consequently we are not getting the exact same set of eigenvectors for problem dimensions > 32.

This does not mean that `brtools` is suddenly doing something wrong.  It just means that the test assumed we would have the exact same output as `scipy`, and we don't any more.  

This is an annoying problem to fix, I think.  Two possible methods:

1. attempt to "trick" scipy into doing the same pre-1.5 behaviour.  Possibly this would involve monkey-patching out the ability of `eigh` to optimise the `lwork` parameters (i.e. stubbing out `scipy.linalg.lapack._compute_lwork` to always return `lwork=18*nrows`, `lrwork=24*nrows` and `liwork=10*nrows` to match `qutip.cy.brtools.ZHEEVR`).
2. do some linear algebra work to determine that the eigenvectors we have returned are equivalent to the ones scipy has returned.

2 is a better solution, but harder to implement, especially in the presence of degenerate or near-degenerate eigenvalues (of which there are likely a lot around 0).  1 is very unsatisfying, but might be enough.

In general, this problem will hopefully be solved a bit more thoroughly when `brmesolve` is updated to use the new data layer types, and we standardise the calculation of eigenvalues and -vectors across QuTiP.

Hopefully someone has a better idea than me?  @Ericgig, @nonhermitian.

Ericgig:
In `brtools`'s tests, we could check that the results `eigenvector` are mathematically right, without matching `scipy`'s result. Should be quite simple and we would more resilient to changes in scipy. 

Did anybody check if it solves the segfault issues on Mac yet?

jakelishman:
Certainly the `zheevr` segfaults on my machine with scipy < 1.5 and _doesn't_ with scipy 1.5, but that's hardly a perfect test - segfaults are pretty intermittent at the best of times.

[...]

PR #1300, opened 2020-06-25
Disable brtools zheevr test

jakelishman:
Temporarily disables the `brtools` test `test_zheevr` when using `scipy` version >= 1.5.  See #1299 for more information and discussion.

This does not _need_ to be merged if we solve the more general problem quickly, but will fix the problem in the interim if we need it.  Build on top of #1298 just to show that taken together they "fix" all the tests.


jakelishman:
Deprecated by #1301.

PR #1301, opened 2020-06-25
Fix brtools zheevr test

### LINKED ISSUE #1299: Scipy 1.5 changes to linalg.eigh break brtools zheevr tests

jakelishman:
The new test explicitly checks the eigenvalue equation for each generated eigenvalue/vector pair, and tests (via the determinant) that all eigenvectors are linearly independent, as they must be since they are eigenvectors of a Hermitian matrix.

The previous test relied on the implementation of `zheevr` producing _exactly_ the same result as `scipy.linalg.eigh`, which was very fragile, and indeed broken by scipy 1.5.

Fixes #1299.
Deprecates PR #1300, since this actually turned out to be a much easier fix than I had feared.  Intermittent segfaults on Mac persist, unfortunately, so this still requires #1288.

### DIFF
--- a/qutip/tests/test_brtools.py
+++ b/qutip/tests/test_brtools.py
@@ -49,11 +49,18 @@ def test_zheevr():
     """
     for dimension in range(2, 100):
         H = qutip.rand_herm(dimension, 1/dimension)
-        our_evals = np.zeros(dimension, dtype=np.float64)
-        our_evecs = _test_zheevr(H.full('F'), our_evals)
-        scipy_evals, scipy_evecs = scipy.linalg.eigh(H.full())
-        np.testing.assert_allclose(scipy_evals, our_evals, atol=1e-12)
-        np.testing.assert_allclose(scipy_evecs, our_evecs, atol=1e-12)
+        Hf = H.full()
+        evals = np.zeros(dimension, dtype=np.float64)
+        # This routine modifies its arguments inplace, so we must make a copy.
+        evecs = _test_zheevr(Hf.copy(order='F'), evals).T
+        # Assert linear independence of all the eigenvectors.
+        assert abs(scipy.linalg.det(evecs)) > 1e-12
+        for value, vector in zip(evals, evecs):
+            # Assert the eigenvector satisfies the eigenvalue equation.
+            unit = vector / scipy.linalg.norm(vector)
+            test_value = np.conj(unit.T) @ Hf @ unit
+            assert abs(test_value.imag) < 1e-12
+            assert abs(test_value - value) < 1e-12
 
 
 @pytest.mark.parametrize("operator", [


Ericgig:
Thank you for your quick fix.

jakelishman:
No problem - test failures are a big deal, because they prevent everyone else from seeing what might be going wrong with their own code, and lots of people are actively working on QuTiP at the moment.

> Merging once tests pass
"if", I think - I have a mac, so I still get the segfaults!

Ericgig:
If only the mac one fail I will merge it anyway and update #1288.

jakelishman:
Assuming this works, should we to backport the changes here and in #1298 to QuTiP 4.4 and 4.5? Possibly @nathanshammah has an opinion on this too.

jakelishman:
Maybe we should require that at least one of the test cases requires a version of scipy pre-1.5 so we don't cause any regressions for older versions of scipy as Travis builds start to reliably pull 1.5?  Perhaps we can add a requirement for scipy=1.4.1 to the Linux py3.7, MKL/OMP one?

Ericgig:
I agree with having a test pre 1.5.
These changes should be released quickly, but probably just to 4.5.

Issue #1246, opened 2020-05-03
"pip3 install quitp" fails unless cython is already installed


jevonlongdell:
**Describe the bug**
It looks like the pypi package doesn't have cython as a dependency?

**To Reproduce**
Do a "pip3 install cython" on a fresh ubuntu 20.04 install (i had already installed numpy and scipy)

** This is what happens when I try it**
jevon@bb8:~$ pip3 install qutip
Collecting qutip
  Downloading qutip-4.5.0.tar.gz (3.5 MB)
     |████████████████████████████████| 3.5 MB 509 kB/s 
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-69anjthq/qutip/setup.py'"'"'; __file__='"'"'/tmp/pip-install-69anjthq/qutip/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-69anjthq/qutip/pip-egg-info
         cwd: /tmp/pip-install-69anjthq/qutip/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-69anjthq/qutip/setup.py", line 52, in <module>
        from Cython.Build import cythonize
    ModuleNotFoundError: No module named 'Cython'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.

** Installing cython fixes it **

jevon@bb8:~$ pip3 install cython
Collecting cython
  Downloading Cython-0.29.17-cp38-cp38-manylinux1_x86_64.whl (2.0 MB)
     |████████████████████████████████| 2.0 MB 721 kB/s 
Installing collected packages: cython
Successfully installed cython-0.29.17
jevon@bb8:~$ pip3 install qutip
Collecting qutip
  Using cached qutip-4.5.0.tar.gz (3.5 MB)
Requirement already satisfied: cython>=0.21 in ./.local/lib/python3.8/site-packages (from qutip) (0.29.17)
Requirement already satisfied: numpy>=1.12 in ./.local/lib/python3.8/site-packages (from qutip) (1.18.3)
[...]

Issue #1174, opened 2020-02-04
Installing as external dependency fails if Cython is missing


thangleiter:
Installing `qutip` as an external dependency fails if `cython` is not installed since it's imported without checks in [`setup.py`.](https://github.com/qutip/qutip/blob/0f412cd4bf983a3b497d54406103d0209b553a7a/setup.py#L52-53)

From what I can tell, this was an issue before (#426) and was subsequently fixed (#489) but then again broken (https://github.com/qutip/qutip/commit/82e209512927a0d6086de31364004fdb0da8c59a).

Sidestepping this would be great since for instance readthedocs to my knowledge doesn't allow running custom commands (like `pip install cython`) before installing dependencies when building a documentation that requires `qutip`.

BoxiLi:
Closed by #1309

PR #1309, opened 2020-07-06
Add pyproject.toml to specify build-system as standardized in PEP 518

terrorfisch:
**Description**
Add `pyproject.toml` to specify build-system as standardized in PEP 518. This enables building qutip in an emtpy environment i.e. without `setuptools`, `Cython`, `numpy` and `scipy` being pre-installed.

If you have a new minimal debian install you sould be able to install this version with
```
# apt-get install python3-pip
$ pip3 install $PATH_TO_QUTIP
```
I only veryfied this in WSL so please test :)

**Related issues or PRs**
This solves #1246 and #1174 and makes CI for packages that depend on qutip a lot easier.

**Changelog**
Add `pyproject.toml` so qutip can be built without prerequisites

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel", "Cython", "numpy", "scipy"]


BoxiLi:
Thanks for the PR @terrorfisch!

I'm not very familiar with this so I'm probably making stupid mistakes. When I try locally `pip3 install \qutip`, also on WSL actually, it still chocks at the line `from Cython.Build import cythonize` in the `setup.py`. And I don't really understand why `pyproject.toml` can help come across that. Do you understand the reason behind why it should work?

BoxiLi:
Ok, my bad, I specified the wrong folder. It works for me now, although it takes a bit longer to solve the dependency. It might be very useful for downstream packages since these issues kept popping up.

BoxiLi:
Thanks for the help @terrorfisch !

gschaffner:
@Ericgig It looks like #1246 and #1174 are fixed but didn't get closed automatically with this PR.

BoxiLi:
Thanks @gschaffner

PR #1313, opened 2020-07-07
fix the wrong dtype in the lattice module on windows

BoxiLi:
**Description**
`f16` does not exist on Windows. Use `np.longdouble` instead, which is `f16` on Mac/Linux and `f8` on Windows.

On windows one gets
```python
>>> np.dtype([('f', np.longdouble)])         
dtype([('f', '<f8')])
```
On Ubuntu
```python
>>> np.dtype([('f', np.longdouble)])
dtype([('f', '<f16')])
```

**Changelog**
fix the dtype error in the lattice module

### DIFF
--- a/qutip/lattice.py
+++ b/qutip/lattice.py
@@ -851,7 +851,7 @@ class Lattice1d():
         if self.period_bnd_cond_x == 0:
             raise Exception("The lattice is not periodic.")
         (knxA, qH_ks, val_kns, vec_kns, vec_xs) = self._k_space_calculations()
-        dtype = [('eigen_value', '<f16'), ('eigen_vector', Qobj)]
+        dtype = [('eigen_value', np.longdouble), ('eigen_vector', Qobj)]
         values = list()
         for i in range(self.num_cell):
             for j in range(self._length_of_unit_cell):


BoxiLi:
@Ericgig I tested it on Windows. All tests pass.

jakelishman:
Is there any particular reason you need `long double` at all?  `'f16'` is quad-precision, which is a very non-standard type at the best of times, and no real surprise it's not supported on Windows.

BoxiLi:
I did really look into the details of the lattice module. From a quick look, I don't think there is a particular reason, maybe just to be compatible with all choice of float number? Probably @Ericgig knows more since it was last year's GSoC project.

jakelishman:
That's fair.  The C standard doesn't actually really specify anything for `long double` (it just has to be at least the same size as `double`), which is why `np.longdouble` is weird between different compilation targets.  _Technically_ `np.double` doesn't have to be 64-bit, IEEE-754 double-precision, but I doubt there's any machine around today that runs Python where it isn't.

Looking at the file, it's actually completely unnecessary to use anything other than `np.double` (i.e. `np.float64`) here - the `np.longdouble` part is constructing a new array, but the values it then stores in that field come from an array which has dtype `float`, which is 64-bit in Python.  It largely shouldn't matter, though if some people use the eigenvalue array directly for computation, they may find stuff is slower than they expect - the quad-precision arithmetic is not always hardware-supported, but may use software emulation.

BoxiLi:
I remember we also encountered this os dependency before. `np.int` is also not the same on Windows and Linux, which has its root in C long.

jakelishman:
Yeah, using `np.int`, `np.long` or `np.longlong` when it _matters_ what size they are is a recipe for disaster.  C99 guarantees that `stdint.h` will be available, so you can always safely use `np.int8`, `np.int16`, `np.int32` and `np.int64` if you know the size you need.  If you just need "at least" whatever, then in C `int` is >= 16-bit, `long` is >= 32-bit and `long long ` is >= 64-bit.

Ericgig:
Don't overthink anything in lattice, it was rough to get anything usable from that GSoC. We only checked that the functions worked, but were lenient about the style and code-structure.

PR #1322, opened 2020-07-15
make sure the specified solver in Processor.run_state is used (not always mesolve)

julianiacoponi:
**Description**
Spotted the `solver` in `Processor.run_state` was set but not used, so it would always use `mesolve` even if `mcsolve` was specified, so I fixed it.

**Related issues or PRs**
Not sure if this fixes any issues - I doubt it? - I guess no one had tried to use the `mcsolve` option here yet.

**Changelog**
Ensured `solver` kwarg of `Processor.run_state` correctly sets the solver to `mesolve` or `mcsolve` (`mesolve` was hardcoded in before).

### DIFF
--- a/qutip/qip/device/processor.py
+++ b/qutip/qip/device/processor.py
@@ -703,7 +703,7 @@ class Processor(object):
         elif solver == "mcsolve":
             solver = mcsolve
 
-        evo_result = mesolve(
+        evo_result = solver(
             H=noisy_qobjevo, rho0=init_state,
             tlist=noisy_qobjevo.tlist, **kwargs)
         return evo_result


julianiacoponi:
@BoxiLi I guess
https://github.com/qutip/qutip/blob/4f9b04b149734a7d81cb25ff25a844e4ee41a6ea/qutip/tests/test_processor.py#L333-L345
didn't successfully add a check for this?

I'll try get the unit tests running locally so I can add to this, but feel free to fix this up on my behalf if it's obvious to you :)

BoxiLi:
Yes, you are right. Thanks for spotting this! That was a stupid mistake. I added the `mcsolve` option along with a few other small stuffs, didn't pay enough attention to it. The test just tests if the things run fine... Actually I don't really have a straightforward idea of how to test if a specific solver is used. Free free to propose any idea you have!

julianiacoponi:
I can probably do it with [`unittest.mock.assert_called_once_with`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with), although I'm not sure what the etiquette is on using multiple different unittesting frameworks is, as I see this repo is `pytest`ed mostly. I guess `unittest` is part of the standard library so it should be okay?

I will add a commit with the unittest updated when I get the chance, though feel free to merge this 1-line fix as it might as well be in master now that it's been noticed.

BoxiLi:
>  although I'm not sure what the etiquette is on using multiple different unittesting frameworks is, as I see this repo is pytested mostly.

This is because we are moving from `nose` to `pytest`. Not made intentionally.

PR #1324, opened 2020-07-22
moving lattice to qutip/qutip-lattice

Ericgig:
**Description**
For the new release, lattice support in lattice should be moved to it's own module.
This is removing these files from the next major version of qutip.
https://github.com/qutip/qutip-lattice replacing them and ready for a v0.1.0 release.
* The documentation is still inside qutip's documentation.

@sbisw002 do you want to be a contributor on qutip/qutip-lattice, your commit history is lost by the move.

@nathanshammah, I set the _authors_ of the new repository as Saumya, you, me and Clement, and I kept the same support (numfocus and unitaryfund). Are you fine with that?


nathanshammah:
sounds good to me. thanks for taking the lead on this.

sbisw002:
Hi. Eric. Nathan.

This looks wonderful. Please keep me as a contributor.

I still have to help post the examples. Please help me with that. I need to
know what exactly is needed for that.

Also, I have some ideas about spin-fermion transformations that should be
compatible with both the Dicke bases and the lattice module that I shall
tell you guys about soon. And also about some cool unitary evolutions.

Saumya

PR #1328, opened 2020-07-27
Replace fast_csr_matrix in core

jakelishman:
Not a complete switch-over to `CSR` from `fast_csr_matrix`, but this goes a really long way to achieving that.  `Qobj` is done completely, but `QobjEvo` is pretty difficult to disentangle, and that really slowed me down a lot, and I'm not at all sure I've got it right there yet.

OpenMP is ignored for now as low-priority.  We'll need to get it working before a QuTiP 5.0 release, but I'm not working on it at the moment.

`CQobjEvo` has been absolutely stripped bare.  All its variants are gone, since the data layer types will handle this instead.  `QobjEvo` is basically managed by shim at the moment, and needs a huge tidy up to bring it up to speed with the new data layer, and keep it maintainable.

The data layer types have been taught to pickle themselves safely and automatically, and so consequently `QobjEvo` is magically pickle-able (aside from the OpenMP types).

This PR includes all the commits from #1296 as well, but it's still a pretty giant PR as it stands right now.

The tests won't pass yet, because I haven't finished converting everything.


jakelishman:
Closing and reopening this so the "files changed" doesn't include changes already merged to `dev.major`.

PR #1336, opened 2020-08-10
Coefficient object

Ericgig:
Add a `Coefficient` object to encapsulate the coefficients of the QobjEvo.
`Coefficient`s replace qobjevo's codegen. Their advantages are
- Coefficient are always 'compiled'. String are compiled to cython code at initialization.
- Have cython decorator: taking the conjugate does no longer require new compilation.
- String coefficient are saved to reuse with basic pattern matching: "cos(a*t)" and "cos( w1 *t )" will use the same compiled object. This object is compiled once and kept between session.
- Cleaner compilation using `setuptools` instead of a monkey patched `pyximport`.
- `QobjEvo` is always compiled.
- `ObjeEvo` does not need to keep track of type making the code simpler.

There are shims to make `Pulse` and qip tests work since they used `QobjEvo.tlist` and `QobjEvo.ops.coeff`.
@BoxiLi, the new QobjEvo can be added even if coefficient's tlist are different, but you cannot get back the tlist and the array from them. Also for _step_func_coeff, before the first time, `Coefficient` return the first value of the array, not 0 as expected in pulse.py.


jakelishman:
For future posterity for others, there are some other comments at jakelishman/qutip#2.  I looked through this PR and it looks good to me.

qutip/core/cy/coefficient.pyx
```diff
@@ -0,0 +1,389 @@
...
+        self.args = args
+
+    cdef complex _call(self, double t) except *:
+        return self.func(t, self.args)
+
+    def copy(self):
+        return StrFunctionCoefficient(self.base, self.args.copy())
+
+    def __reduce__(self):
+        return (StrFunctionCoefficient, (self.base, self.args))
```
jakelishman, line 136: This class has `autopickle` set, but also defines a `__reduce__` method.

qutip/solve/stochastic.py
```diff
@@ -824,7 +824,7 @@ def _prespostdag(op):
 
     for op in sso.c_ops:
         LH -= op._cdc() * sso.dt * 0.5
-        sso.pp += op.apply(_prespostdag)._f_norm2() * sso.dt
+        sso.pp += spre(op) * spost(op.dag()) * sso.dt
```
jakelishman, line 827: could be `sprepost(op, op.dag())`?

qutip/solve/stochastic.py
```diff
@@ -939,7 +939,7 @@ def _prespostdag(op):
 
     sso.sops = [[spre(op._cdc()) + spost(op._cdc()),
                  spre(op._cdc()),
-                 op.apply(_prespostdag)._f_norm2()] for op in sso.sc_ops]
+                 spre(op) * spost(op.dag())] for op in sso.sc_ops]
```
jakelishman, line 942: could be `sprepost(op, op.dag())`?  Also, the `_f_norm2()` has gone.

Ericgig: `apply`+`_f_norm2` do not create any of the cross terms coming from `sprepost`.
I removed the `_f_norm2` method.

BoxiLi:
> There are shims to make Pulse and qip tests work since they used QobjEvo.tlist and QobjEvo.ops.coeff.
@BoxiLi, the new QobjEvo can be added even if coefficient's tlist are different, but you cannot get back the tlist and the array from them. Also for _step_func_coeff, before the first time, Coefficient return the first value of the array, not 0 as expected in pulse.py.

Thanks @Ericgig ! This is very useful!

PR #1338, opened 2020-08-12
Data-layer type creation, conversion and dispatch

jakelishman:
Data-layer type creation, conversion and dispatch
=================================================

_Note: this is currently a work-in-progress, and not all functionality may exist as written while the PR is still in draft mode.  I will add to the descriptions as I go._

## Background

Previously in #1332, the old `fast_csr_matrix` type was replaced by the new, custom `CSR` type as the data backing for `Qobj`, and all internal QuTiP data representations.  This produced some speed-ups in some places due to improved algorithms and better cache usage in places, but its principle advantage was the massive reduction in overhead for function calls between Python and C space, which largely affected small objects.

The full aim, however, is to have QuTiP 5 support many different data representations as the backing of `Qobj`, and use the most suitable representation for the given data. This will not require every single QuTiP function to have an exponential number of versions for every possible combination of inputs, but only to have specialisations for the most common data combinations.  This concept is the "data layer".

All code examples in this PR are prefixed with

```python
>>> from qutip.core import data
```

## Specification

The core to achieving this is fast, fully specified inter-conversion between all known data types, and efficient multiple-dispatch for mathematical operations.  There are then four principle components of the data-layer:

1. a creation routine which returns an appropriate data-layer type given some arbitrary Python object (`data.create`)
2. a routine which can perform the conversion from any data-layer type to any other data-layer type (`data.to`)
3. completely specialised mathematical operations (e.g. `data.add_csr_dense_dense(CSR, Dense) -> Dense`)
4. an object which provides multiple dispatch operations on its input arguments to use an exact specialisation (defined in item 3) if known, or uses the conversion routine (item 2) to convert the inputs into ones matching a specialisation if not: `data.Dispatcher`.  The exported mathematical functions will all be instances of this type.

The minimum work needed to define a new data-layer type is to provide `data.to` with two conversion functions; one into the new type from a current data-layer type, and one which converts the new type _into_ a current data-layer type.  Once this is done, every single QuTiP component will be able to use the new data-layer type, although until specialisations are given which use it, it will always be achieved by conversion to another type, and conversion back.  In this way, a new type can be added incrementally, with only the most common operations needing to be defined to get good efficiency.

**Important caveat:** the data layer operates only on _exact_ types; subclasses of defined types will be treated as completely different types.  This is to do with keeping the computational complexity of multiple-dispatch operations as O(1) (i.e. I don't know how to do multiple dispatch in constant time allowing inheritance).

### `data.to`: conversion between types

```python
>>> matrix = data.dense.identity(5)
>>> matrix
Dense(shape=(5, 5), fortran=True)
>>> data.to(data.CSR, matrix)
CSR(shape=(5, 5), nnz=5)
```

```python
>>> data.to[data.CSR, data.Dense]
<converter to CSR from Dense>
```

```python
>>> data.to[data.Dense]
<converter to Dense>
```

```python
>>> class NewDataType:
...     # [...]
>>> def new_from_dense(matrix: data.Dense) -> NewDataType:
...     # [...]
>>> def dense_from_new(matrix: NewDataType) -> data.Dense:
...     # [...]
>>> data.to.add_conversions([
...     (NewDataType, data.Dense, new_from_dense),
...     (data.Dense, NewDataType, dense_from_new),
... ])
>>> data.to[data.CSR, NewDataType]
<converter to CSR from NewDataType>
```

#### Basic usage

Convert data into a different type.  This object is the knowledge source for every allowable data-layer type in QuTiP, and provides the conversions between all of them.

The base use is to call this object as a function with signature
```text
    (type, data) -> converted_data
```
where `type` is a type object (such as `data.CSR`, or that obtained by calling `type(matrix)`) and `data` is data in a data-layer type.  If you want to create a data-layer type from non-data-layer data, use `create` instead.

You can get individual converters by using the key-lookup syntax.  For example, the item
```python
    to[CSR, Dense]
```
is a callable which accepts arguments of type `Dense` and returns the equivalent item of type `CSR`.  You can also get a generic converter to a particular data type if only one type is specified, so
```python
    to[Dense]
```
is a callable which accepts all known (at the time of the lookup) data-layer types, and converts them to `Dense`.  See the "Efficiency notes" section below for more detail.

Internally, the conversion process may go through several steps if new data-layer types have been defined with few conversions specified between them and the pre-existing converters.  The first-class QuTiP data types `Dense` and `CSR` will typically have the fastest connectivity.

#### Adding new types

You can add new data-layer types by calling the `add_conversions` method of this object, and then rebuilding all of the mathematical dispatchers.  See the docstring of that method for more information.

#### Implementation details

Not all conversions have to be specified for a new type; it is enough to have just one to and from a known type to a new type.  The rest of the conversion graph is built up by graph traversal over known types (the graph is reconstructed whenever `add_conversions` is called), where the approximate cost of each function is used as the weight of an "edge" joining two data-layer type "vertices".  The shortest path conversion function is constructed and stored (as the interal type `data.convert._converter`) for each pair of types.  We willingly sacrifice memory efficiency for speed-efficiency here, since we expect there to be few data-layer types, but for the calls to happen millions of times.

The converters returned by single-key access (e.g. `data.to[data.Dense]`) are constructed individually on a call to `__getitem__`, and are instances of the private type `data.convert._partial_converter`, which internally stores a reference to every "full" converter, and dispatches to the correct one when called.

The entire `data.to` object and all subsidiary `_converter` and `_partial_converter` objects are `pickle`-able. 

#### Efficiency notes

From an efficiency perspective, there is very little benefit to using the key-lookup syntax.  Internally, `to(to_type, data)` effectively calls `to[to_type, type(data)]`, so storing the object elides the creation of a single tuple and a dict lookup, but the cost of this is generally less than 500ns.  Using the one-argument lookup (e.g. `to[Dense]`) is no more efficient than the general call at all, but can be used in cases where a single callable is required and is more efficient, concise and descriptive than `functools.partial`.

### `data.Dispatcher`: arbitrary multiple-dispatch operations

```python
>>> import scipy.sparse
>>> import numpy as np
>>> a = data.CSR(scipy.sparse.csr_matrix(np.random.rand(5, 5)))
>>> b = data.Dense(np.random.rand(5, 5))
>>> data.add(a, b)
Dense(shape=(5, 5), fortran=True)
>>> data.add(a, b, out=data.CSR)
CSR(shape=(5, 5), nnz=25)
```

```python
>>> data.add[data.CSR, data.Dense]
<indirect specialisation (CSR, Dense, Dense) of add>
>>> data.add[data.CSR, data.CSR, data.CSR]
<direct specialisation (CSR, CSR, CSR) of add>
```

#### Basic usage

A `Dispatcher` provides a single mathematical function for _all_ combinations of types known by `data.to`, regardless of whether the particular specialisation has been defined for the input data types.  In the first example above, the operator `data.add` currently only knows two specialisations; it knows how to add `CSR + CSR -> CSR` and `Dense + Dense -> Dense` directly, but it is still able to produce the correct result when asked to do `CSR + Dense -> CSR` and similar.  The type of the output can be, but does not need to be, specified.  The `Dispatcher` will choose a suitable output type if one is not given.

For example, the objects `data.add`, `data.pow` and `data.matmul` are some examples of dispatchers in the data layer.  Respectively, these have the signatures
```python
data.add(left: Data, right: Data, scale: complex = 1) -> Data
data.pow(matrix: Data, n: integer) -> Data
data.matmul(left: Data, right: Data) -> Data
```
These are callable functions, so the base use is to call them.

Just like `data.to`, key-lookup syntax can be used to get a single callable object representing a single specialisation.  The callable object has an attribute `direct` which is `True` if no type conversions would need to take place, and `False` is at least one would have to happen.  Just like in the regular call, you can either specify or not specify the type of the output, but the types of the inputs must always be given.
```python
>>> data.pow[data.CSR]
<direct specialisation (CSR, CSR) of pow>
>>> data.pow[data.CSR].direct
True
>>> data.pow[data.CSR, data.Dense].direct
False
```
The returned object is callable with the same signature as the dispatcher (except the `out` keyword argument is no longer there), and requires that the inputs match the types stated.

#### Adding new specialisations

New specialisations can be added to a pre-existing dispatcher with the `Dispatcher.add_specialisations` method.  This is very similar in form to `data.to.add_conversions`; it takes lists of tuples, where the first elements of the tuple define the types in the specialisation, and the last is the specialised function itself.

For example, a user might need to multiply `Dense @ CSR` frequently and get a `Dense` output.  Currently, there is no direct specialisation for this:
```python
>>> data.matmul[Dense, CSR, Dense]
<indirect specialisation (Dense, CSR, Dense) of matmul>
```
The user may then choose to define their own specialisation to handle this case efficiently:
```python
>>> def matmul_1(left: Dense, right: CSR) -> Dense:
...     # [...]
...     return out
```
They would give this to `data.matmul` by calling
```python
>>> data.matmul.add_specialisations([
...     (Dense, CSR, Dense, matmul_1),
... ])
```
Now we find
```python
>>> data.matmul[Dense, CSR, Dense]
<direct specialisation (Dense, CSR, Dense) of matmul>
```

Additionally, the whole lookup table will be rebuilt taking this new specialisation into account, which means the indirect specialisation `matmul(Dense, CSR) -> CSR` will now make use of this new method, because it has a low conversion weight.

#### Adding new types

Now let's say the user wants to add a new `NewDataType` type all across QuTiP.  The only action they _must_ take is to tell `data.to` about this new type.  Let's say they define it like this:
```python
>>> class NewDataType:
...     # [...]
>>> def new_from_dense(matrix: data.Dense) -> NewDataType:
...     # [...]
>>> def dense_from_new(matrix: NewDataType) -> data.Dense:
...     # [...]
>>> data.to.add_conversions([
...     (NewDataType, data.Dense, new_from_dense),
...     (data.Dense, NewDataType, dense_from_new),
... ])
```
As we saw in the previous section, this is enough to define all conversions in `data.to`.  What's more, this is _also_ enough to define all operations in the data layer as well:
```python
>>> data.matmul[NewDataType, data.CSR]
<indirect specialisation (NewDataType, CSR, CSR) of matmul>
```
All of the data layer will now work seamlessly with the new type, even though this is actually achieved by conversion to and from a known data type.  There was no need to call anything other than `data.to.add_conversions`.  Internally, this is achieved by `data.Dispatcher.__init__` storing a reference to itself in `data.to`, and `data.to` calling `rebuild_lookup` as part of `add_conversions`.

Now the user only needs to add in the specialisations that they actually need for the bottle-neck parts of their application, and leave the dispatcher to handle all other minor components by automatic conversion.  As in the previous subsection, they do this by calling `add_specialisations` on the relevant operations.

#### Creating a new dispatcher

In most user-defined functions which operate on `Qobj.data` it will be completely sufficient for them to simply call `data.to(desired_type, input_data)` on entry to the function, and then they can guarantee that they are always working with the type of data they support.

However, in some cases they may want to support dispatched operations in the same way that we do within the library code.  For this reason, the data layer exports `Dispatcher` as a public symbol.  The minimal amount of work that needs to be done is to call the initialiser, and then call `add_specialisations`.  For example, let's say the user has defined two specialisations for their simple new function `add_square`:
```python
>>> def add_square_csr(left, right):
...     return data.add_csr(left, data.matmul_csr(right, right))
...
>>> def add_square_dense(left, right):
...     return data.add_dense(left, data.matmul_dense(right, right))
...
```
(Ignore for now that this would be better achieved by just using the dispatchers `data.add` and `data.matmul` directly.)  Now they create the dispatcher simply by doing
```python
>>> add_square = data.Dispatcher(add_square_csr, inputs=('left', 'right'), name='add_square', out=True)
>>> add_square.add_specialisations([
...     (data.CSR, data.CSR, data.CSR, add_square_csr),
...     (data.Dense, data.Dense, data.Dense, add_square_dense),
... ])
```
This is enough for `Dispatcher` to have extracted the signature and satisfied all of the specialisations.  Note that the `inputs` argument does not provide the signature, it tells the dispatcher which arguments are data-layer types it should dispatch on, e.g. for `data.pow` as defined above `inputs = ('matrix',)`, but the signature is `(matrix, n) -> out`.  See that the specialisations are now complete:
```python
>>> add_square
<dispatcher: add_square(left, right)>
>>> add_square[data.Dense, data.CSR, data.CSR]
<indirect specialisation (Dense, CSR, CSR) of add_square>
```

In the initialisation, the function `add_square_csr` is passed as an example from which `Dispatcher` extracts the call signature, the module name and the docstring (if it exists).  It is not actually added as a specialisation until `add_square.add_specialisations` is called afterwards.

If desired, the user can set or override the docstring for the resulting dispatcher by directly writing to the `__doc__` attribute of the object.  We _always_ do this within the library.

_Note:_ within the Cython components of the library, we manually construct the signature and pass it into `Dispatcher.__init__` because Cython-compiled functions do not embed their signature in a manner in which `inspect.signature` can extract it (even with the `embedsignature` directive).  We also use this to cut out some arguments in the call signatures which would not work with the dispatch mechanism (like `out` parameters).

#### Other features

In combination with `data.to`, this now allows QuTiP to handle _any_ backing data store for `Qobj`, even if literally zero mathematical functions are defined for the type.

The `Dispatcher` can operate on a function with any call signature (except ones which use `*args` or `**kwargs`), even if not all of the arguments are data-layer types.  At definition, the creator of the `Dispatcher` says which input arguments are meant to be dispatched on, and whether the output should be dispatched on, and all other arguments are passed through like normal.

#### Implementation details

The backing specialisations can be found in `Dispatcher._specialisations`, and the complete lookup table is in `Dispatcher._lookup`.  These are marked as private, because messing around with them will almost certainly cause the dispatcher to stop working.

Only one specialisation needs to be defined for a dispatcher to work with _all_ data types known by `data.to`.  We achieve this because `data.to` guarantees that all possible conversions between data types will exist, so `data.Dispatcher` can always convert its inputs into those which will match one of its known specialisations.

Within the initialisation of the data layer, we use a "magic" `_defer` keyword argument to `add_specialisations` to break a circular dependency.  This is because the "type" modules `data.csr` and `data.dense` depend on some mathematical modules (e.g. `add` and `matmul`) to provide the `__add__` and similar methods on the types.  For ease of development we want the dispatchers to be defined in the same modules that all the specialisations are (though this is not at all necessary), but the dispatchers require `data.to` to be populated with the types before specialisations can be added.  The `_defer` keyword here just defers the building of the lookup table until an explicit call to `Dispatcher.rebuild_lookup()`, breaking the cycle.  The user will never need to do this, because by the time they receive the `Dispatcher` object, `data.to` is already initialised to a minimum degree.

#### Efficiency notes

The specialisations returned by the `__getitem__` lookups are not significantly faster than just calling the dispatcher directly, because the bulk of the heavy lifting is done when `add_specialisations` or `rebuild_lookup` is called.  On call, the generic signature `(*args, **kwargs)` has to be bound to the actual signature of the underlying operation, regardless of whether the specialisation has already been found.  At the Cython level there is short-circuit access to the call machinery in the specialisations themselves, but this cannot be safely exposed outside of the `Dispatcher` class itself.


qutip/core/data/convert.pyx
```diff
@@ -0,0 +1,285 @@
...
+to be exported names, but are the backing machinery of the functions `data.to`
+and `data.create`, which are built up as the last objects in the `__init__.py`
+initialisation of the `data` module.
+"""
+
+# This module is compiled by Cython because it's the core of the entire
+# dispatch table, and having it compiled to a C extension saves about 1µs per
+# call.  This is not much at all, and there's very little which benefits from
+# Cython compiliation, but such core functionality is called millions of times
+# even in a simple interactive QuTiP session, and it all adds up.
```
Ericgig, line 15: Didn't you talk about reducing our reliance on cython when proposing your project...

jakelishman: This is at compile-time only, not at run-time.  The user doesn't need to have Cython installed.  I thought we'd talked about reducing Cython usage only in the context of string coefficients?

At any rate, the entire data layer is implemented in Cython, and there's no escaping that while keeping up the speed gains.  New data-types don't need to be specified using Cython at all, so there's no run-time requirement on it.

jakelishman: As in, the user never has to recompile this object - you can add new data types at runtime from within Python space, no problem at all.

Ericgig: I have no issue with it. I was commenting on how it seems you got to like cython during the last few months.

jakelishman: Oh yeah!  Cython's definitely got a very steep initial learning curve, and its documentation is a bit lacking in places, but I'm definitely a big fan for the really highly used components.  One thing I am careful about is implementing too much in Cython - it makes it much harder for downstream users to monkeypatch our functionality to suit their own needs.

Ericgig:
You say this will be used a lot, but `CSR` and `Dense` are both quite complete and few functions will require this. As I understand it, it is mostly for other specialized and user defined data types. 

Will dispatch be rebuild when adding a new data layer?
ie. CSC ptrace dispatcher: Does it realize `CSC_ptrace` does not exist and use `data.to` at call. Or does it build a `CSC_ptrace` remembering the call series `data.to[data.CSR, data.CRC]`, `CSR_ptrace`, `data.to[data.CRC, data.CSR]`?

jakelishman:
> You say this will be used a lot, but `CSR` and `Dense` are both quite complete and few functions will require this. As I understand it, it is mostly for other specialized and user defined data types.

If we take the function `add` as an example, it takes two arguments and returns a data-layer type, so that's three arguments it will dispatch on.  Even if the _only_ types are `Dense` and `CSR`, that's still 8 possible specialisations that might be called for.  QuTiP core is unlikely to use many of them (who will use `Dense + Dense -> CSR`?), but the code has to be able to handle them.  Typically the user will never call `data.to` directly, but the dispatcher will do it for them when a conversion is needed.  `CSR` and `Dense` are the first-class types, so we'll in general aim that these guys never need to be converted within core library code, but as soon as any other data type (even `CSC`) is defined, it'll need to handle it.

> Will dispatch be rebuild when adding a new data layer?

Yes.  It will either be a manual call to a function `qutip.rebuild_dispatch()`, or more likely, the call to `to.add_conversions()` will trigger it automatically for all dispatcher objects, even user-defined ones.  The dispatcher will handle everything for the user.

Let's say you define a new `CSC` type, and you supply `data.to` with converters between `CSC` and `CSR` _only_, and you specify _no_ mathematical specialisations for it.  First off, `data.to` builds a compound converter between `CSC` and `Dense` by finding the shortest path (it'll do `CSC -> CSR -> Dense` and `Dense -> CSR -> CSC`).  Next, `data.to` rebuilds all the dispatch tables known by QuTiP, including `data.ptrace`.

Currently `data.ptrace` has only one specialisation, and that's `ptrace(CSR, spec) -> CSR`.  So now a call `data.ptrace(CSC, [0])` internally is equivalent to defining and calling
```python
def ptrace_csc_csr(matrix: CSC, spec) -> CSR:
    csr = data.to(CSR, matrix)
    return data.ptrace_csr(matrix, spec)
```
This will all happen automatically, with no necessary intervention from the user or the developer.  Note that the dispatcher by default will output a `CSR` type, _not_ `CSC`.  This is because outputting `CSR` involves fewer conversions, and consequently has less "weight".  The dispatcher thought process is
1. are all input and output dispatch types specified? [no - the single dispatched input is, but the output type is not]
2. find all known specialisations which match the types which _are_ specified [there aren't any]
3. find the specialisation which requires the least conversion "weight" to satisfy [it finds`CSR -> CSR`, since in this case it's the only one that exists]
4. use `data.to` to convert inputs into the types matching the specialisation [calls `data.to(CSR, csc_matrix)`]
5. call the specialised function
6. call `data.to` on the output to produce the required output type [none was specified, so it just returns what it got]

If you instead decide that you care that you get a `CSC` object, you would call `data.ptrace(csc_matrix, spec, out=CSC)`.  Steps 2 and 6 are now different - first, specialisations which output `CSC` directly will likely have a lower conversion weight and so will be selected first (there aren't any, so in practice in this example nothing changes).  In step 6, the output was specified to be `CSC`, and it knows it called a specialisation which returns `CSR`, so it will return `data.to(CSC, output)`.

All this will happen completely internally, and automatically.  You won't need to define `ptrace` for `CSC` matrices for the data layer to find the partial trace of `CSC` data, or any other mathematical function.  If you do, the dispatcher will automatically use it when given `CSC` data, but will continue to use the `CSR` specialised version when passed `CSR` data.  This will work for every single mathematical function defined on the data layer, no matter how many new data-layer types are added, and how many mathematical operations there are.

jakelishman:
Oh, and I haven't exactly decided how the dispatchers will store their "constructed" methods yet.  It'll depend a little on if there are major speed implications one way or the other (I'm especially concerned with the speed of binding to arbitrary function signatures), and whether they will require a huge amount of memory.  I suspect that memory won't ever be too much concern - at their hearts, the dispatchers will mostly need a couple of `dict` attributes and a whole bunch of references to pre-existing objects.  All that stuff is pretty light-weight.

ajgpitch:
This all sounds most ingenious too me. Quite beyond what we expected. I really like the flexibility and easy in which a new type can be added.
It seems to all make sense. I have a few questions, but I will save them for when we meet in a bit.

jakelishman:
@Ericgig: #1337 broke `mcsolve` seed reuse - it seems to be deliberate, since all the seed code has been explicitly cut.  Is this something that `mcsolve` will no longer support?  If so, we need to remove the tests of seeding, because they're failing the build, and if not we need to reinstate it quickly so we can have passing tests.

jakelishman:
Oooh, Linux-only segfault.  I'll have to hook up a Linux VM to try and debug that.

Ericgig:
Seed will be moved but not removed. Please skip the test for now.
I was surprised how easily the test where passing. If all the solver test where skipped, there may be a few fails.
I will add the KeyError in __setitem__.

jakelishman:
Yeah, I didn't notice before because I always run the tests locally before I push big things like that, and all the solver tests run locally, since it was just a packaging mix-up.

As long as changing `__setitem__` means that `SolverOptions(seeds=[1, 2, 3])` throws an error (for now), then it's all good.

jakelishman:
I'm nervous because I haven't been able to work out why the segfaults in 9d67492 appeared, and I haven't been able to reproduce them at all.  Running with AddressSanitizer on, I get absolutely no problems on Mac at all, and I set up a quick Linux VM and couldn't catch anything either.  I had a real go at getting AddressSanitizer running on Travis on my own repo, but I wasn't successful, so couldn't get it that way either.

It seems to have disappeared for now, so it _could_ just have been a one-off mess up in the environment, but that's not usually how computers work.  Let's just keep an eye on it, and let me know if anyone can find a repro for it.

Also note the coverage is back up to the _more_ reasonable (but still way too low) 73% now all the tests are enabled...

Other than that, all the dispatchers are now in, but not yet used.  This PR is ready for review, and I'll open a new one to make the switch over, because it'll involve changing a large number of files.

jakelishman:
For future reference, `data.create` is still to do properly, but it's a low priority for now.

qutip/core/data/tidyup.pyx
```diff
@@ -39,3 +44,53 @@ cpdef CSR tidyup_csr(CSR matrix, double tol, bint inplace=True):
...
+        The absolute tolerance to use to determine whether a real or imaginary
+        part should be truncated to zero.
+
+    inplace : bool, optional (True)
+        Whether to do the operation in-place.  The output matrix will always be
+        returned, even if this argument is `True`; it will just be the same
+        Python object as was input.
+    """
+tidyup.add_specialisations([
+    (CSR, tidyup_csr),
```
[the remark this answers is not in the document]
jakelishman: Well spotted!  The new PR I'll send soon makes some steps to addressing this, but I didn't want to overburden this one.

We do need to add a lot more documentation and make the `auto_tidyup` option _much_ more visible, because on small sizes, it actually adds a lot of overhead to mathematical operations.  With regards to the dispatchers, this problem is perhaps solved a little better by having the default method of resolving unary operations (like `tidyup`) when no output type is specified being to maintain the same type, even if that means converting the output back.

I imagine there's loads of little tweaks and usage papercuts to find and solve that a beta period will really help uncover.

jakelishman:
Yeah, `create` isn't fully done - it's basically the last resort for turning user input into a data-layer type, so it's not very important that it's efficient.  We'll say in the documentation that it's preferable to create a `Dense` or `CSR` instance and pass that it, but have `data.create` as the backup to allow everything to "just work", albeit slowly.

It will probably have an `add_creators` method, which will be similar to `add_specialisations` and `add_conversions`, but it'll only take an output type, a function and a priority.  Then it'll probably just be a loop of `try: create(input); except: pass` until a creator succeeds.

About the `*_csr` still used in `Qobj`: yes, these are deliberately left as-is in this PR, so as not to bulk up the files changed.  There'll be a new PR to add those, and others to add more specialisations between `Dense` and `CSR`.

Right now I'm writing a completely orthogonal PR updating the `pytest`, `coverage` and Coveralls support for `dev.major`.  I'm not yet backporting that to the 4.X branch because the `sparse_routines.pxi` file complicates matters for the coverage, apparently...

qutip/core/data/tidyup.pyx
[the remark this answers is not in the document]
jakelishman: By "makes some steps to addressing this", I actually just mean that I added a `tidyup_dense` method.

This PR actually misrepresents all that's required to make a new type at the moment; currently you also have to specify `adjoint`, `conj` and `transpose` methods, but I'll likely remove that requirement.  At any rate, once I've written the proper documentation for how to add a new type, there'll be a full guide on how to do it, with everything that needs to be done.

I think I'm going to move the relevant methods onto the `data.Data` ABC, and now I've solved the circular dispatch import problem, I'll likely attach the full dispatchers in as the mathematical methods, and the `adjoint`, `transpose`, etc methods.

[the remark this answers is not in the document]
jakelishman: See also: #1349.

PR #1341, opened 2020-08-12
Remove fastsparse.py file

jakelishman:
This file was accidentally left in in #1332, even though all its functionality is replaced.  PR now made onto the correct branch.


jakelishman:
The tests report errors because of the cancelled PR #1340 which was made to the wrong branch.

PR #1342, opened 2020-08-14
Quality of life improvement to QobjEvo

Ericgig:
**Description**
- Add product between `Qobj`, `QobjEvo` and `Coefficient`:
 `QobjEvo([op, "t"]) === op * coefficient("t")`
- `tensor` accept `QobjEvo`
- Overload `&` for tensor product for `Qobj` and `QobjEvo`:
 `tensor(A, B) === A & B`
- Add `__matmul__` to `QobjEvo`.
- Remove call to `QobjEvo.compile` outside `qobjEvo.py`. (Always compiled)
- Add `dims` and `shape` as read-only properties of `QobjEvo`.
- Reworked `QobjEvo`'s tests. 
- Found a bug in `liouvillian` with constant `QobjEvo` and `Qobj`.


Ericgig:
@jakelishman can I merge this?

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+        elif form == "mixed_tlist":
+            coeff_1 = coefficient(self.cplx_coeff.array(self.tlist,
+                                                        self.args),
+                                  tlist=self.tlist)
+            coeff_2 = coefficient(self.real_coeff.array(self.tlistlog,
+                                                        self.args),
+                                  tlist=self.tlistlog)
+            obj = QobjEvo([self.cte_qobj,
+                           [self.cplx_qobj, coeff_1],
+                           [self.real_qobj, coeff_2]])
```
jakelishman, line 221: This `if .. elif .. elif` isn't really how the parametrisation is meant to be used.  You could parametrise over a tuple of `form, input_args, input_kwargs` parameters or something (you can override the default fixture using `mark.parametrize`).  The advantage of this is that `pytest` will print out the values of the input parameters on a test failure, so we'd see them right there in the test log, whereas here we'd just get the string, and it takes more work to find out what's gone on.

Second, there's no assertions in this test; I get that it's in theory just testing that there are no errors on construction, but it should probably also assert that the output has the correct number of ops, the coefficients are what we expect, etc etc.

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+        self.qobjevos[form] = obj
+
+    def test_call(self, form):
+        t = self._rand_t()
+        assert isinstance(self.qobjevos[form](t), Qobj)
+
+        for _ in range(10):
```
jakelishman, line 228: You can use `pytest.mark.repeat(10)` as a decorator on the test, then we get stats on if all runs failed, or just one.

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+                _assert_qobj_almost_eq(self.qobjevos[form](t), expected)
+
+    def test_product_coeff(self):
+        coeff_1 = coefficient("1")
+        coeff_t = coefficient("t")
+        created = self.cte_qobj * coeff_1
+        _assert_qobjevo_equivalent(self.qobjevos['cte'], created)
+
+        created = created * coeff_t
+        for _ in range(10):
```
jakelishman, line 249: same comment about `mark.repeat`

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+        _assert_qobj_almost_eq(as_qevo, as_qobj)
+
+    def test_tidyup(self):
+        "QobjEvo tidyup"
+        obj = self.qobjevos["no_args"]
+        obj *= 1e-10 * np.random.random()
+        obj.tidyup(atol=1e-8)
+        t = self._rand_t()
+        # check that the Qobj are cleaned
+        assert_equal(np.max(obj(t).full()), 0.)
```
jakelishman, line 367: use `np.testing.assert_allclose(obj(t).full(), 0)` so we get proper reporting on a failure (`assert_equal` should never ever be used any more - in `pytest`, the bare `assert x == y` is much clearer)

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+        pickled = pickle.dumps(self.qobjevos[form], -1)
+        recreated = pickle.loads(pickled)
+        _assert_qobjevo_equivalent(recreated, self.qobjevos[form])
+
+    @pytest.mark.parametrize('superop', [lindblad_dissipator,
+                                         partial(lindblad_dissipator, chi=0.5),
+                                         sprepost,
+                                         lambda O1, O2: liouvillian(O1, [O2])],
+                             ids=['lindblad', 'lindblad_chi',
+                                  'sprepost', 'liouvillian'])
```
jakelishman, line 382: you can write
```python
pytest.mark.parametrize('superop', [
  pytest.param(lindblad_dissipator, id='lindblad'),
  pytest.param(partial(lindblad_dissipator, chi=0.5), id='lindblad_chi'),
  ...
])
```
to make it easier to read the ids, and to make changes/add lines later.

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+        _assert_qobjevo_equivalent(transposed, obj.trans())
+
+    def test_mul_vec(self, form):
+        "QobjEvo mul_vec"
+        t = self._rand_t()
+        N = self.N
+        vec = np.arange(N)*.5+.5j
+        op = self.qobjevos[form]
+        assert np.allclose(op(t).full() @ vec,
+                           op.mul_vec(t, vec))
```
jakelishman, line 434: this is one scenario where `pytest`'s bare assert isn't as good as `np.testing.assert_allclose` - currently `pytest` doesn't treat `numpy` arrays specially in the introspection

qutip/tests/core/test_qobjevo.py
```diff
@@ -42,111 +42,442 @@
...
+
+    def test_expect_psi(self, form):
+        "QobjEvo expect psi"
+        t = self._rand_t()
+        N = self.N
+        vec = _data.dense.fast_from_numpy(np.arange(N)*.5 + .5j)
+        op = self.qobjevos[form]
+        _expect = _data.expect_csr_dense
+        Qo1 = op(t)
+        assert np.allclose(_expect(Qo1.data, vec), op.expect(t, vec))
```
jakelishman, line 455: all lines with `_expect` in them will clash with the new PR that I'm shortly going to make which puts in the dispatchers properly.  If you're ready to merge this soon, then leave them as they are right now and I'll rebase the new PR on top of this once it's merged.  Otherwise, this is just a warning about why there's a huge merge conflict.

qutip/core/qobjevo.py
```diff
@@ -586,32 +632,45 @@ def __neg__(self):
...
-        res.compile()
+        res._compile()
         return res
 
+    def __and__(self, other):
+        """
+        Syntax shortcut for tensor:
+        A & B ==> tensor(A, B)
+        """
+        return qutip.tensor(self, other)
```
jakelishman, line 643: if you import `tensor` as a trailing import at the end, you don't need to prefix it with `qutip` and that helps the core package keep its separability from the rest of QuTiP (though I'd still like to break more of the circular imports) - like is done in `qobj.py`

I'm sure there's a nicer way to handle these circular imports anyway, I just haven't thought it through properly yet.  I _used_ to have to be much stricter about it back when I wrote F#...

qutip/core/qobjevo.py
```diff
@@ -756,6 +827,12 @@ def expect(self, t, state, herm=0):
         return exp.real if herm else exp
 
     def mul_vec(self, t, vec):
+        """
+        Product of the operator quantum object at time t
+        with the given vector state.
+        """
+        # TODO: mostly used in test to compare with the cqobjevo version.
+        # __mul__ sufficient? remove?
```
jakelishman, line 835: I don't think `CQobjEvo` even has a `mul_vec` implementation any more?

jakelishman (commented):
there's a few places I didn't flag that'll have merge conflicts with the new dispatching PR, but that's not a particular problem - whichever of us gets merged second can rebase over the first one.

Glad to see the short-hand tensor notation go in, too.

qutip/tests/core/test_qobjevo.py
Ericgig: I don't like `mark.repeat`, it would create 10 lines of tests for each type of `QobjEvo`. It clogs the pytest outputs more than give extra useful information...

qutip/tests/core/test_qobjevo.py
Ericgig: When will your PR be ready? I would like to merge this one quickly mostly for the liouvillian fix and auto-compilation of QobjEvo which affect solvers.

qutip/core/qobjevo.py
Ericgig: There are a few circular import created by this `&` and `Coefficient * Qobj` product... 
The only way I see to remove all these would be to monkey patch this `__and__` method in tensor.py.

qutip/tests/core/test_qobjevo.py
jakelishman: I wrote the `mark.repeat` code back when I was first starting to learn `pytest` - it's in our `conftest.py`.  We could probably change it to only list failing lines when reporting the tests.  I'd argue though that it does add information, because you find out whether it's a systematic problem with the test (all repeats fail) or if it's a one-off, and `pytest` collates the failures down at the end, where extra repeats don't show at all, so there's no textual cost in the main report.

qutip/tests/core/test_qobjevo.py
jakelishman: yeah, merge yours first

qutip/core/qobjevo.py
jakelishman: Yeah, I suppose it's not an easy problem to solve right now.  I have a (slightly irrational) dislike of circular imports and of subpackages not having a strict order (e.g. `qutip.core` should ideally be fully initialised before any other part of `qutip` is accessed), but it's not a big deal right now.

PR #1347, opened 2020-08-24
Fix coverage reporting on Travis

### LINKED ISSUE #1339: Coveralls does not handle Cython files correctly
In #1332 there is a huge drop in reported test coverage by Coveralls.  This is because large tracts of the most-tested Python files were either severely reduced in length, or converted to Cython files, and it seems that Coveralls does not correctly handle Cython; it reports the existence of a `x.py` file for every `x.pyx` file that exists, with the `.py` file having 7 lines, all of which are apparently untested.

The new data-layer types _are_ tested albeit not as completely as they will be once they're finished, but Coveralls has reported a huge drop in coverage.  I don't know if any coverage estimators exist for Cython (I would expect not due to their compiled nature), but it's perhaps something we should consider.

Also, Coveralls counts the test files as part of its coverage indicator.  By my rough count, there are 42,223 SLOC in `qutip` (i.e. excluding comments and whitespace but unfortunately including docstrings), of which 11,259 are in `qutip/tests`.  Aside from a couple of skipped tests, 100% of the lines in `qutip/tests` will be run, so if we produce a modified coverage report, having 61% of the SLOC covered in _all_ of `qutip` is equivalent to
```python
>>> (0.61*42_223 - 1.00*11_259) / (42_223 - 11_259)
0.4681898
```
**46.8% true coverage**.

We probably could really do with a concerted effort to add more tests of all the little nuances of the code.

jakelishman:
In #1339 I pointed out that the test files were being incorrectly counted in the coverage statistics when running on Travis.  This transpired to be because `coverage.py` wasn't able to find our `.coveragerc` file correctly when `pytest` was called in the manner it is on Travis.

This PR then does a couple of things:
1. correctly locates the `.coveragerc` file
2. excludes the Cython bootstrapping code from being part of the coverage
3. correctly locates the pytest configuration file `pytest.ini` (although now we have a `pyproject.toml`, we could move that configuration there)
4. completely reorganises `.travis.yml` to remove dead configs, properly share common setups, and generally make it a bit easier to read and see what's going on
5. fixes file reporting in Coveralls (look at the "tree" tab in the Coveralls web report compared to the current `master`); you can now click on the files and get proper line-by-line detail on what was covered and what wasn't, e.g. [here's the report for `mcsolve.py`](https://coveralls.io/builds/32982136/source?filename=qutip/mcsolve.py)

We _don't_ enable Cython line-tracing and coverage analysis.  This is actually [quite easily possible](https://cython.readthedocs.io/en/latest/src/tutorial/profiling_tutorial.html#enabling-coverage-analysis), but enabling full tracing including on `nogil` functions absolutely tanks performance to the degree where the coverage Travis run would take well in excess of two hours to complete (I think Travis actually just kills things that run longer than 2 hours).

I might return to the Cython elements to see how much of them when _can_ reasonably do coverage analysis on.

My prediction about actual test coverage dropping once the test files are correctly omitted, which I mentioned in #1339, came true though not _quite_ as cataclysmically (`coverage.py` uses statements, not SLOC as its metric); coverage drops from ~71% to ~63%.  Technically the true percentage is higher because the Cython code is generally very well covered, but arguing between 63% and 71% when we should be aiming to get to >95% is missing the point.

Fixes #1339.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,190 +1,91 @@
 # After changing this file, check it on:
 #   http://lint.travis-ci.org/
 
-_stage_generic_linux: &stage_generic_linux
-  os: linux
-  language: python
-  # change dir
-  before_script:
-    - QUTIP_DOWNLOAD=$(pwd -P)
-    - mkdir qutip_testing
-    - cd qutip_testing
-  # command to run tests
-  script:
-    - python -m qutip.about
-    - QUTIP_INSTALL=$(dirname $(python -c 'import qutip; print(qutip.__file__)'))
-    - pytest --verbosity=1 --cov=qutip --cov-config="${QUTIP_DOWNLOAD}/.coveragerc" --rootdir="${QUTIP_INSTALL}/qutip/tests" -c "${QUTIP_INSTALL}/qutip/tests/pytest.ini" --pyargs qutip
-  install:
-    - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
-    - bash miniconda.sh -b -p $HOME/miniconda
-    - export PATH="$HOME/miniconda/bin:$PATH"
-    - hash -r
-    - conda config --set always_yes yes --set changeps1 no
-    - conda info -a
-    - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
-    - source activate test-environment
-    - conda install mkl blas=*=mkl numpy scipy pytest pytest-cov cython coveralls
-    - python setup.py install
+# These just set the default, we'll override them for the Mac entries in the job
+# matrix.  Travis likes us to specify the defaults.
+os: linux
+dist: xenial
+language: python
 
-_stage_linux_36: &stage_linux_36
-  <<: *stage_generic_linux
-  name: "Python 3.6"
-  dist: trusty
-  python: 3.6
-  script:
-    - python -m qutip.about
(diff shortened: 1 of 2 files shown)


jakelishman:
Ok, that Coveralls report is clearly broken, but you can see in the test log that the coverage reporting is correctly fixed (see the Linux, Python 3.7 report). I suspect that my directory hopping has caused the generated `.coverage` file to get lost, which I'll fix tomorrow.

jakelishman:
That last commit updates the versions of macOS and XCode that we test against.  XCode 12 is comparatively very slow (both compile and run), so we should likely try to figure out why that's the case and improve it.  We made the same commit on `dev.major` in #1348.

.travis.yml
```diff
@@ -1,184 +1,92 @@
...
-      <<: *stage_osx_37
-    - stage: test
-      <<: *stage_osx_38
+    - name: "Linux, Python 3.8: regular"
+      python: 3.8
+      install:
+        - conda install numpy scipy matplotlib pytest cython
+        - python setup.py install
+
+    - name: "Linux, Python 3.7: OpenMP, MKL, scipy<1.5 (coverage)"
```
Ericgig, line 54: The name says scipy<1.5 but the installation does not set that limit.

jakelishman: Thanks for spotting it!  I'll have made the exact same mistake in #1348 too (already merged to `dev.major`) so I'll send another quick PR on that branch to fix it there as well.

PR #1348, opened 2020-08-25
Fix test coverage on `dev.major`

jakelishman:
This is a completely equivalent PR to #1347, but applied to `dev.major`.  See the comments and discussion in that PR for more details.

Addresses #1339.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,10 @@
 [run]
-include = */qutip/*
-omit = */tests/*
+source = qutip
+omit =
+    # QuTiP test files
+    */qutip/tests/*
+
+[report]
+exclude_lines =
+    # Skip Python wrappers which help load in C extension modules.
+    __bootstrap__()
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,190 +1,91 @@
 # After changing this file, check it on:
 #   http://lint.travis-ci.org/
 
-_stage_generic_linux: &stage_generic_linux
-  os: linux
-  language: python
-  # change dir
-  before_script:
-    - QUTIP_DOWNLOAD=$(pwd -P)
-    - mkdir qutip_testing
-    - cd qutip_testing
-  # command to run tests
-  script:
-    - python -m qutip.about
-    - QUTIP_INSTALL=$(dirname $(python -c 'import qutip; print(qutip.__file__)'))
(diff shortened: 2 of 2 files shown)


jakelishman:
The test failure is unrelated; the swap to the new `CSR` type has meant that a few test tolerances have become much more fragile when considering numerical precision; since we're now using our own mathematical operations rather than scipy's, we're a bit more likely to have disagreements with them.

This may also have to do with the new version of `Qobj` being more consistent about applying the `auto_tidyup` rule.  This is one rule I really think we should look at removing or seriously reworking; it's basically just "numerical imprecision as a feature, not a bug".

Ericgig:
While you're at it, could you update the mac test to a more recent version of mac. xcode12 is now out but we are still testing on xcode10 (https://docs.travis-ci.com/user/reference/osx).  There were segfault with xcode11, thus we skipped it, but I hope it's solved with the `eigh` patch.

jakelishman:
Sure.  Shall I backport that change to `master` too?

Ericgig:
If the tests pass with xcode12 we should activate it on `master`. But if it breaks them, we should find a solution for `dev.major` but let it go for v4.

Having windows test would help a lot, but if mac change a lot between versions, maybe go with 2 linux, 2 mac, 1 windows. 
We can have python 3.6 on mac or remove it. No cython should be run on windows.

Ericgig:
xcode12's test quite slow, almost twice the time of other tests...

jakelishman:
I think that's a good thing that we've tested that - we need to work out why performance is so bad on mac with Python 3.8.

jakelishman:
I've just spotted a bug in `isherm` on the new data layer, which is my fault for not getting the tests for that written yet.  That may also be having some impact on the test fragility.

Ericgig:
xcode10 with python 3.8 is ~21min. 
`python setup.py install` took ~203s with xcode10 but 385s with xcode12... Compilation seems to be the issue.

jakelishman:
I mean, it's still a problem we can try and figure out.  Unfortunately I generally don't actually use the compilers bundled with XCode, and I don't use macOS 10.15 because of the screwy updates they made to dev tools...

jakelishman:
> I've just spotted a bug in `isherm` on the new data layer, which is my fault for not getting the tests for that written yet. That may also be having some impact on the test fragility.

Actually, I _didn't_ introduce a bug in `isherm`!  It's just also present in the release version of QuTiP, but _masked_ by the `auto_tidyup`.

In "unrelated" news: I discovered a dumb bug in `tidyup_dense`.

Issue #1350, opened 2020-08-25
Incorrect results from zcsr_isherm


jakelishman:
In certain circumstances, `zcsr_isherm` can give a `False` negative.  This is usually masked by the `auto_tidyup` option being turned on.

Reproduce:
```python
>>> import numpy as np
>>> import scipy.sparse
>>> import qutip
>>> x = scipy.sparse.csr_matrix(np.array([[1, 1e-15j], [0, 1]]))
>>> bool(qutip.cy.spmath.zcsr_isherm(x, 1e-12))
False
```
Up to the tolerance specified (`1e-12`), the given matrix _is_ Hermitian, so it should return `True`.

This happens when a small non-zero value (lower than the tolerance) is compared to a zero.  This is because the structure test in `zcsr_isherm` assumes that for Hermicity to hold, there _must_ always be an explicitly stored corresponding element for each entry, however clearly this isn't always the case.

It hasn't appeared before because we always have `auto_tidyup` turned on with a sufficiently large tolerance that these values would be removed by almost all operations which could have introduced this disparity (mostly matrix multiplication).

(I'll fix this on both the `master` branch and on `dev.major`, this is just for posterity)

PR #1351, opened 2020-08-25
Use dispatchers to allow Qobj to use any backing data

jakelishman:
Finally, `Qobj` can use any backing data store.  This fairly small and simple PR is effectively the culmination of all the work on `dev.major` over the last few months, and effectively is just like flicking a switch to turn on all the capabilities that we've built up through #1282, #1296, #1332, #1338 and #1340.

There is still a lot of work to be done, but I'm trying to transition to smaller, easier-to-check PRs to make review easier.

Possibly incomplete to do list:
 - write a proper `data.create`
 - possibly add the dispatchers in as `data.Data` mathematical methods (e.g. `__matmul__`)
 - add options for controlling default output types from the dispatchers
 - tests for `Qobj` with both types
 - tests for creation and property routines of the data layer
 - tests for the dispatch operation
 - tests for conversion operations
 - more specialisations to be written for `Dense` and `CSR`/`Dense`
 - more ergonomic selection of _method_, not just output type (for example, you should be able to specify you want the `Dense` specialisation of `eigs` even if you pass it a `CSR`)
 - more ergonomic dispatchers for matrix creation (e.g. how exactly will `qutip.basis` function, and how will users add specialisations?)
 - documentation, both user-facing and developer-facing
 - fix algorithms in `add_csr` and `matmul_csr` to use `csr.Accumulator` (should provide a speedup and remove some sorts)
 - fix `isherm_csr` (see #1350 - `isherm_csr` uses the same algorithm, so has the same problem)

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1709,6 +1709,8 @@ class Qobj:
                         copy=False)
         # We use the condition from John Watrous' lecture notes,
         # Tr_1(J(Phi)) = identity_2.
+        # See: https://cs.uwaterloo.ca/~watrous/LectureNotes.html,
+        # Theory of Quantum Information (Fall 2011), theorem 5.4.
         tr_oper = qobj.ptrace([0])
         return np.allclose(tr_oper.full(), np.eye(tr_oper.shape[0]),
                            atol=settings.core['atol'])
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1054,13 +1054,35 @@ class Qobj:
         return out
 
     @_tidyup
-    def ptrace(self, sel):
+    def ptrace(self, sel, dtype=None):
         """
-        Partial trace of the quantum object.
+        Take the partial trace of the quantum object leaving the selected
+        subspaces.  In other words, trace out all subspaces which are _not_
+        passed.
+
+        This is typically a function which acts on operators; bras and kets
+        will be promoted to density matrices before the operation takes place
+        since the partial trace is inherently undefined on pure states.
+
+        For operators which are currently being represented as states in the
+        superoperator formalism (i.e. the object has type `operator-ket` or
+        `operator-bra`), the partial trace is applied as if the operator were
+        in the conventional form.  This means that for any operator `x`,
+            operator_to_vector(x).ptrace(0) == operator_to_vector(x.ptrace(0))
+        and similar for `operator-bra`.
+
+        The story is different for full superoperators.  In the formalism that
+        QuTiP uses, if an operator has dimensions (`dims`) of
+        `[[2, 3], [2, 3]]` then it can be represented as a state on a Hilbert
+        space of dimensions `[2, 3, 2, 3]`, and a superoperator would be an
+        operator which acts on this joint space.  This function performs the
+        partial trace on superoperators by letting the selected components
+        refer to elements of the _joint space_, and then returns a regular
+        operator (of type `oper`).
 
         Parameters
         ----------
-        sel : int/list
+        sel : int or iterable of int
             An ``int`` or ``list`` of components to keep after partial trace.
             The selected subspaces will _not_ be reordered, no matter order
             they are supplied to `ptrace`.
@@ -1071,7 +1093,6 @@ class Qobj:
             Quantum object representing partial trace with selected components
             remaining.
         """
-        # TODO: reorganise ptrace functions into proper data layer bits.
         try:
             sel = sorted(sel)
         except TypeError:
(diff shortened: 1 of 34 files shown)


jakelishman:
@Ericgig if you want to merge #1342 first, I'll rebase this PR on top of it once it's in.

Ericgig:
@jakelishman I merged #1342, you can rebase.
Will this PR be merge ready once the merge conflict resolved?

jakelishman:
I suppose the answer is "yes", but there's still that huge to do list at the top that needs to be handled (including some really really important bits like the tests).

jakelishman:
That's the rebase done.  Those tests _should_ pass now; they do on my machine.

Ericgig:
Which warnings? `pyximport`'s? This is still used by `brmesolve`.
I forgot about the build directory... Will see to it later. Thank you for remindering me.

jakelishman:
```python
jake@tauros$ python
Python 3.7.7 (default, May  6 2020, 04:59:01)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import qutip
/Users/jake/.anaconda3/anaconda3/envs/qutip-dev/lib/python3.7/site-packages/setuptools/distutils_patch.py:26: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  "Distutils was imported before Setuptools. This usage is discouraged "
>>>
```

It was caused by the original coefficients PR.  Could be interplay between `pyximport` and the later import of `setuptools`, but I'm not sure.  I don't think you _need_ `setuptools` (which is also causing the `build` directories) - I think you can use importlib to import the result of `cythonize` directly?

qutip/tests/core/test_qobjevo.py
```diff
@@ -176,42 +175,43 @@ def _rand_t(self):
...
+        pytest.param('func',
+                     [cte_qobj, [cplx_qobj, cplx_coeff.func]],
+                     {'args': args}, id='func'),
+        pytest.param('string',
+                     [cte_qobj, [cplx_qobj, cplx_coeff.string]],
+                     {'args': args}, id='string'),
+        pytest.param('spline',
+                     [[cplx_qobj, cplx_coeff.spline(tlist, args)], cte_qobj],
+                     {}, id='spline'),
+        pytest.param('array', [cte_qobj,
```
Ericgig, line 189: If you prefer this spacing, at least be consistent.

jakelishman: oh, that was a mistake. I didn't even mean to change the first bit

jakelishman: or more, I didn't mean to commit it - I just fiddle around with it while I'm reading it

jakelishman: I'll revert it and fix the commit

qutip/core/qobj.py
```diff
@@ -833,25 +850,25 @@ def diag(self):
...
+        # TODO: add a `diagonal` method to the data layer?
+        out = _data.to(_data.CSR, self.data).as_scipy().diagonal()
         if np.any(np.imag(out) > settings.core['atol']) or not self.isherm:
             return out
         else:
             return np.real(out)
 
     @_tidyup
-    def expm(self, method='dense'):
+    def expm(self, dtype=_data.Dense):
```
Ericgig, line 861: Why ask for the data type here, instead of the string? This force the user to know where to find the data layer object, which I don't like... But the string are messy and cannot be extended as easily to new types... I don't like it but it has it's advantages.

qutip/core/superop_reps.py
```diff
@@ -59,7 +59,13 @@
 from .states import basis
 
 
-_SINGLE_QUBIT_PAULI_BASIS = (identity(2), sigmax(), sigmay(), sigmaz())
+# TODO: revisit when creation routines have dispatching.
+_SINGLE_QUBIT_PAULI_BASIS = (
+    identity(2).to(_data.CSR),
```
Ericgig, line 64: Wouldn't it be better to have these as `Dense` per default?

qutip/core/qobj.py
jakelishman: Yeah, I don't like this either.  It's a similar problem with `eigs`, but that's even less obvious.  I'm going to think about it a bit more, and hopefully I'll have some more ideas to suggest - it's one of my to do bullet points.

qutip/core/superop_reps.py
jakelishman: Possibly?  They're only used in the next function down in `superop_reps`, and I didn't test enough of what it was doing to know if a sparse or dense representation was better for that function.  It looks sort of sparse, maybe?

jakelishman:
I'm happy to go either way on whether to merge this now or later.  I'm happy to keep adding to it, but I'm also mindful that I don't want this to turn into another +15000/-15000 PR.

I possibly need to look into that test failure a little more - it's temperamental, and I'm not sure if it's something where it's just a case of a too strict tolerance, or if there is an actually numerical precision problem.

Ericgig:
I prefer smaller PRs, this is big enough.
Continue in another one please.
I will put the test to a to check list and review it when working on `brmesolve`.

jakelishman:
It's ~caused by the swap to the data-layer - probably it's just minor floating-point uncertainties due to slightly changed algorithms, but I'm not 100% sure on that.

PR #1355, opened 2020-08-31
Some minor fixes

### LINKED ISSUE #1354: class ``_file_list()`` in ``qobjevo.py`` is rebundant
**Describe the bug**
As described in the title, ``file_`` in ``_file_list().add(file_)`` already has the extension ``'.pyx'`` (this is done by _import_str() in qobjevo_codegen.py), it is unnecessary to add yet another suffix.

**To Reproduce**
This can be verified by add a line printing out ``file_`` in ``_file_list().clean()`` 
then run any test that can trigger it (for example ``test_QobjEvo_with_state()``) and observe the output.

**Expected behavior**
Should only have a single ``'.pyx'`` in ``file_``

**Your Environment**
```
Current admin team: Alexander Pitchford, Paul D. Nation, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, and Boxi Li
Project Manager: Franco Nori.

QuTiP Version:      4.6.0.dev0+8e181cc4
Numpy Version:      1.19.1
Scipy Version:      1.5.0
Cython Version:     0.29.21
Matplotlib Version: 3.3.1
Python Version:     3.8.5
Number of CPUs:     8
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True
Installation path:  C:\Users\berry\PycharmProjects\qutip\qutip
==============================================================================
==============================================================================
```

**Additional context**
I discovered this when dealing with some test failures in windows, it seems that there are multiple problems in ``qobjevo.py`` (actually, what I really wanted to figure out in the first place was that some compiled modules would be deleted before being imported). Anyway, I will update this issue whenever I make any progress.

berryboy2012:
increased ``maxiter`` at ``test_driven_cavity_power_gmres()`` to 10000 in tests/test_steadystate.py
resolved mutable default value at ``parallel_map()`` in parallel.py
resolved mutable default value at various places in qobjevo.py
cleaned up now defunct class ``_file_list`` in qobjevo.py (fixes #1354)

**Description**
Fixes #1354, increases ``maxiter`` at ``test_driven_cavity_power_gmres()`` to ``10000`` in ``tests/test_steadystate.py``.

**Related issues or PRs**
fix #1354

**Changelog**
Increased ``maxiter`` at ``test_driven_cavity_power_gmres()`` to 10000 in tests/test_steadystate.py.
Resolved mutable default value at ``parallel_map()`` in parallel.py.
Resolved mutable default value at various places in qobjevo.py.
Cleaned up now defunct class ``_file_list`` in qobjevo.py (fixes #1354).

### DIFF
--- a/qutip/parallel.py
+++ b/qutip/parallel.py
@@ -193,7 +193,7 @@ def serial_map(task, values, task_args=tuple(), task_kwargs={}, **kwargs):
     return results
 
 
-def parallel_map(task, values, task_args=tuple(), task_kwargs={}, **kwargs):
+def parallel_map(task, values, task_args=tuple(), task_kwargs=None, **kwargs):
     """
     Parallel execution of a mapping of `values` to the function `task`. This
     is functionally equivalent to::
@@ -222,6 +222,8 @@ def parallel_map(task, values, task_args=tuple(), task_kwargs={}, **kwargs):
         each value in ``values``.
 
     """
+    if task_kwargs is None:
+        task_kwargs = dict()
     os.environ['QUTIP_IN_PARALLEL'] = 'TRUE'
     kw = _default_kwargs()
     if 'num_cpus' in kwargs:
--- a/qutip/qobjevo.py
+++ b/qutip/qobjevo.py
@@ -110,35 +110,6 @@ str_env = {
     "spe": scipy.special}
 
 
-class _file_list:
-    """
-    Contain temp a list .pyx to clean
-    """
-    def __init__(self):
-        self.files = []
-
-    def add(self, file_):
-        self.files += [file_ + ".pyx"]
-
-    def clean(self):
-        to_del = []
-        for i, file_ in enumerate(self.files):
-            try:
-                os.remove(file_)
-                to_del.append(i)
-            except Exception:
-                if not os.path.isfile(file_):
-                    to_del.append(i)
-
-        for i in to_del[::-1]:
-            del self.files[i]
-
-    def __del__(self):
-        self.clean()
-
-coeff_files = _file_list()
-
-
 class _StrWrapper:
     def __init__(self, code):
         self.code = "_out = " + code
@@ -444,8 +415,14 @@ class QobjEvo:
         Return the time-dependent quantum object as a list
     """
 
-    def __init__(self, Q_object=[], args={}, copy=True,
-                 tlist=None, state0=None, e_ops=[]):
+    def __init__(self, Q_object=None, args=None, copy=True,
+                 tlist=None, state0=None, e_ops=None):
+        if e_ops is None:
+            e_ops = []
+        if args is None:
+            args = {}
+        if Q_object is None:
+            Q_object = []
         if isinstance(Q_object, QobjEvo):
             if copy:
                 self._inplace_copy(Q_object)
@@ -646,12 +623,14 @@ class QobjEvo:
 
     def __del__(self):
         for file_ in self.coeff_files:
-            try:
-                os.remove(file_)
-            except:
-                pass
+            # Using os module to manipulate filesystem is not thread-safe,
+            # seems fine at the moment though. For the same reason,
+            # do not ignore exceptions here.
+            os.remove(file_)
 
-    def __call__(self, t, data=False, state=None, args={}):
+    def __call__(self, t, data=False, state=None, args=None):
+        if args is None:
+            args = {}
         try:
             t = float(t)
         except Exception as e:
@@ -1524,7 +1503,6 @@ class QobjEvo:
                         get_coeff, file_ = _compile_str_single(
                                                                 part.coeff,
                                                                 self.args)
-                        coeff_files.add(file_)
                         self.coeff_files.append(file_)
                         funclist.append(get_coeff)
                     else:
@@ -1542,7 +1520,6 @@ class QobjEvo:
                                                         self.args,
                                                         self.dynamics_args,
                                                         self.tlist)
-                coeff_files.add(file_)
                 self.coeff_files.append(file_)
                 self.coeff_get = _UnitedStrCaller(funclist, self.args,
                                                   self.dynamics_args,
@@ -1557,7 +1534,6 @@ class QobjEvo:
                                                         self.args,
                                                         self.dynamics_args,
                                                         self.tlist)
-                    coeff_files.add(file_)
                     self.coeff_files.append(file_)
                     self.compiled_qobjevo.set_factor(obj=self.coeff_get)
                     self.compiled += "cyfactor"
@@ -1568,7 +1544,6 @@ class QobjEvo:
                                                         self.args,
                                                         self.dynamics_args,
                                                         self.tlist)
-                    coeff_files.add(file_)
                     self.coeff_files.append(file_)
                     funclist = [part.get_coeff for part in self.ops]
                     self.coeff_get = _UnitedStrCaller(funclist, self.args,
@@ -1608,7 +1583,6 @@ class QobjEvo:
             else:
                 pass
 
-            coeff_files.clean()
             if code:
                 return Code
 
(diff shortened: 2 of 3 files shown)


qutip/parallel.py
```diff
@@ -222,6 +222,8 @@ def parallel_map(task, values, task_args=tuple(), task_kwargs={}, **kwargs):
         each value in ``values``.
 
     """
+    if task_kwargs is None:
+        task_kwargs = dict()
```
jakelishman, line 226: this can also be spelt `task_kwargs = task_kwargs or {}`

jakelishman, line 226: this can also be spelt `task_kwargs = task_kwargs or {}`

qutip/qobjevo.py
```diff
@@ -646,12 +623,14 @@ def _check_old_with_state(self):
...
-            try:
-                os.remove(file_)
-            except:
-                pass
-
-    def __call__(self, t, data=False, state=None, args={}):
+            # Using os module to manipulate filesystem is not thread-safe,
+            # seems fine at the moment though. For the same reason,
+            # do not ignore exceptions here.
+            os.remove(file_)
```
jakelishman, line 629: File-system manipulation is inherently thread-unsafe since you're mutating global state accessible from many locations, so I'm not sure what you mean by this.  The `__del__` destructor of a Python object will only be called by one thread, with safety maintained by the GIL, so provided we enforce that these files are internal to the `QobjEvo` instance (which we do), I don't see a threading issue with this.

Also, I don't understand what you mean by "for the same reason, do not ignore exceptions here".  The major possible exception is that the file does not exist already (perhaps the user cleaned it up manually?).  In this case, ignoring the exception will guarantee that no later file will be deleted.  It should not be an unhandled exception for a file to fail to be deleted, even if it's a permission or other access issue; at worst we've just littered the user's directory with a little bit of detritus, whereas an unhandled exception crashes their script.

qutip/tests/test_steadystate.py
```diff
@@ -487,7 +487,7 @@ def test_driven_cavity_power_gmres():
     c_ops = [np.sqrt(Gamma) * a]
     M = build_preconditioner(H, c_ops, method='power')
     rho_ss = steadystate(H, c_ops, method='power-gmres', M=M, mtol=1e-1,
-                         use_precond=1)
+                         use_precond=1, maxiter=10000)
```
jakelishman, line 490: Why did you make this change?

jakelishman (commented):
Thanks for your interest in contributing to QuTiP!

As it stands, we are writing a major overhaul of how time-dependence coefficients are handled within the library as part of the QuTiP 5.0 release cycle, and all of the code in `QobjEvo` that you've touched here is either removed or significantly changed.  Sorry about that - we're trying to make the development process a bit more transparent now (new dev docs, more public discussion etc).

Still, we know there are current problems for Windows users (it gets neglected because the developers working on the core typically use linux or mac), particularly in the multiprocessing components.  If you are able to help isolate the bug in #1202 (I see you've seen that), commenting on that issue or otherwise letting us know would be very helpful.

If I may offer a couple of points of advice:
1. a PR should generally achieve only one goal (i.e. "fix dangerous default arguments" _or_ "remove `qobjevo._file_list`" or "fix [steadystate] test"), so that relevant bits can be merged separately when ready.  It's ok for a PR to be quite large if the goal is quite large, but it's more difficult to merge unrelated changes all bundled up in one.
2. generally it's best to discuss removals and new feature designs with us before starting work, so we can help guide you towards something we'd be happy to merge.  This is less of a concern when submitted bug fixes for a clearly identifiable problem.

Tagging @Ericgig.

qutip/parallel.py
berryboy2012: This is auto-corrected by PyCharm, I'm fine with it unless shortening code is the way to go.

qutip/tests/test_steadystate.py
berryboy2012: This test cannot get passed on my machine because ``maxiter`` is too small.

qutip/qobjevo.py
berryboy2012: First of all, ``os.remove()`` uses relative path at the moment, and ``working path`` is a global state affecting all threads, if any thread modifies current working path, all ``os.remove()`` related operations will be sabotaged. I think it is best to use absolute path with ``pathlib``. 
Since we do enforce that those files are internal, it shouldn't be the case where those files were deleted before we delete a ``QobjEvo`` (it is very unlikely that users can remove the file before our code does and not causing a problem in importing corresponding module i.e. ``ImportError``). If files were deleted before destructor of ``QobjEvo`` kicks in, either there are some bugs in our code or users deleting those files will get into trouble at some point. To sum up, I don't think covering up potential bugs in our code by ignoring impossible but 'harmless' edge cases.

qutip/tests/test_steadystate.py
jakelishman: It passes on every other test on Travis without this and we set the default for general usage to `maxiter=1000`, so we need to work out why it is different on your machine.

qutip/parallel.py
jakelishman: It's less of a bother here, but in other functions you've changed there are 3 or 4 of them in a row which makes things more tedious to read.

Please try not to include automated changes in completely unrelated files when you're making commits (and PRs, to a lesser extent) - it's much easier in the review process and the commit history if these are separated.

qutip/qobjevo.py
jakelishman: The point about absolute paths is good - I assumed that the files were all absolute already.  This change here does not change any behaviour to do with that, though.

If you are most concerned about ensuring you clean things up, you should not be throwing unhandled exceptions until all your cleanup code has had a chance to run.  If you are particularly concerned about this, then you could run the entire loop with a `try: pass; except OSError` on each step and re-raise afterwards with a merged error.

There _are_ several possibilities for an `OSError` to be raised when trying to delete a file which do show that something is a bit wrong, but are not errors in the rest of the programme.  In a long-running script, the user may already have tidied up the files.  This is not impossible and will not generally cause an error at runtime; `importlib` caches imported modules, so will not search for the file, and the file will certainly have been imported before the `__del__` method is called.  If the user tries to `importlib.reload` (on a completely private module they "officially" have no way of knowing about), _then_ they will get the `ModuleNotFoundError` which is a fine way to handle it.  If the permissions on the directory are changed while the script is running, the file may also fail to be deleted, which is again weird but not an error and should not raise - this can happen when running on multi-process clusters which may dispatch to a different node.  These are unlikely edge cases, but they are _not_ bugs.  Yes, somewhere in QuTiP deleting the file before the `__del__` method is likely a mistake, but in this case there is no harm done by the file failing to be deleted (and if it causes issues, we will get an `ImportError` elsewhere), whereas there is harm done by raising too early when cleaning a list of files where only one failed because the user had opened it in a text editor (thanks Windows) or raising because you're running code on a different cluster node which doesn't have write permissions to the work directory.

We're not looking to punish users.  Methods should not raise unless we cannot continue safely and unambiguously.

The file handling in `QobjEvo` is not perfect, you're right.  Right now, it's essentially temporary code; the whole time-dependent coefficient handling is changed entirely in the next major branch, and none of this code exists any more.  We'd be happy to merge bug-fixes for an identifiable problem in the short-term, but there's little point in changing working code that is scheduled for deletion.

berryboy2012:
Thanks for your attention! However, I am quite busy recently, and default branch works fine for me, so I will close this PR ATM. I will try putting some effort into it once the reworked ``qobjevo.py`` is pushed to default branch.

jakelishman:
No problem, thank you very much for your interest in contributing!  If you do happen to come across anything more about the multiprocessing hangs on Windows, we'd be super interested in hearing about it.

PR #1365, opened 2020-10-03
Integrate the pulse scheduler into Processor

BoxiLi:
**Description**
Integrate the scheduler into Processor. Users can choose to first schedule the circuit/control pulses before running the simulation.

**Changelog**
Supports Processor with the scheduler.


BoxiLi:
Replaced by #1444 and #1445

PR #1380, opened 2020-11-13
Fixed invalid string literals and unclosed files

ElieGouzien:
**Description**
Minor changes to correct invalid string literals and close files opened by `hardware_info.py`.

This removes warnings that currently appear when importing quTiP with non-standard warnings settings (for instance when testing a code that import quTiP).

**Changelog**
Give a short description of the PR in a few words. This will be shown in the QuTiP change log after the PR gets merged.
For example: 
Fixed invalid string literals and proporly close opened files.

### DIFF
--- a/qutip/_mkl/utilities.py
+++ b/qutip/_mkl/utilities.py
@@ -53,7 +53,7 @@ def _set_mkl():
         if plat == 'darwin':
             lib = '/libmkl_rt.dylib'
         elif plat == 'win32':
-            lib = '\\mkl_rt.dll'
+            lib = r'\mkl_rt.dll'
         elif plat in ['linux2', 'linux']:
             lib = '/libmkl_rt.so'
         else:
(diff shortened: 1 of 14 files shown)


jakelishman:
All looks good to me.  Thanks a lot for changing all of that - this is all clearly good stuff and should be merged.

The Travis errors are unrelated - looks like (my...) PR changing `.travis.yml` (#1347) didn't set the Python version correctly in `conda`, and since Python 3.9 is now released, builds are all breaking as we don't distribute binaries compatible with 3.9 yet.

ElieGouzien:
I've rebased on top of master, now that #1381 has been merged.

However there is still an error, but I don't really understand the link with my changes.

jakelishman:
There's no link, just bad luck.  This won't prevent merge - the tests clearly show your change is fine.

Unfortunately our test suite isn't perfect, and sometimes it tries to integrate an insanely stiff system or something like that by accident.

ElieGouzien:
OK, good news.

Anyway, I forced pushed because I realized I've forgotten an invalid string. Maybe I will have more luck this time.

PR #1384, opened 2020-11-16
Fix bug in permute.dimensions_csr

### LINKED ISSUE #1382: Bug in Qobj.permute() in dev.major
Unexpected behaviour in ``Qobj.permute()`` in dev.major branch.

**To Reproduce**
E.g. take any permutation of identity matrices for 'large' system size (e.g. `n>=16`), which should still result in the identity matrix:
```python
>>> import numpy as np
>>> from qutip import tensor, qeye
>>> n = 16
>>> np.isclose(tensor([qeye(2)] * n).permute(list(range(n)))[0,0], 1.0)
False
```

**Expected behavior**
In this example, the permutation should do nothing and full matrix should be diag(1, 1, 1,...), but it is giving diag(0, 1, 1, ...). This occurs for me at `n>=16`. Note, the error is of course not confined to trivial permutations of the identity operator, but this seemed the cleanest case to present.

**Your Environment**
QuTiP Version:      5.0.0b1.dev0+b862fe9
Numpy Version:      1.19.2
Scipy Version:      1.5.2
Cython Version:     0.29.21
Matplotlib Version: 3.3.2
Python Version:     3.7.9
Number of CPUs:     4
BLAS Info:          INTEL MKL
INTEL MKL Ext:      True

**Additional context**
This affects the `expand` functions (e.g. `expand_operator`, `gate_expand_2toN`).
I do not get this error in v4.5.

jakelishman:
First remove the completely unnecessary triplicated calls to `_Indexer.single()` by caching the results.  An otherwise sensible use of the sparse code would get ruined by vastly over-using the call, which is exactly what the special-case is trying to avoid.

Second, fix a buggy line: in `_dimensions_csr_sparse` there used to be the lines (229--230) that's hard to spot in the diff because of the release of the GIL in this commit:
```python
    if n:
        out.row_index[index.single(row) + 1] = len
```
where the assignment should have been `n`, not `len`.  `len` is the longest row seen before now, so the first row seen would also be zeroed out by this function.  Also, a matrix with some rows having more elements in that others would set the wrong values in later short rows, probably causing a segfault.

Release the GIL when we don't need it for iteration - that's often a free speedup in Cython code.

Fixes #1382.

### DIFF
--- a/qutip/core/data/permute.pyx
+++ b/qutip/core/data/permute.pyx
@@ -262,7 +262,8 @@ cpdef CSR dimensions_csr(CSR matrix, object dimensions, object order):
         return _indices_csr_rowonly(matrix, index.all())
     if matrix.shape[0] != matrix.shape[1]:
         raise ValueError("dimensional permute requires square operators")
-    if (matrix.shape[0] * matrix.shape[1]) // csr.nnz(matrix) > 0:
+    cdef double row_density = (<double> csr.nnz(matrix)) / (<double> matrix.shape[0])
+    if row_density >= 0.5:
         permutation = index.all()
         return _indices_csr_full(matrix, permutation, permutation)
     return _dimensions_csr_sparse(matrix, index)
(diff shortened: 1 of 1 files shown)


jakelishman:
The tests will probably fail horrendously on Linux until #1383 is merged.

PR #1394, opened 2020-11-29
Remove some unused imports

jakobjakobson13:
The unused import statements have been found via an automated code analysis.

### DIFF
--- a/qutip/_mkl/spmv.py
+++ b/qutip/_mkl/spmv.py
@@ -31,8 +31,7 @@
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 import numpy as np
-import ctypes
-from ctypes import POINTER,c_int,c_char,c_double, byref
+from ctypes import POINTER, c_int, c_char, byref
 from numpy import ctypeslib
 import qutip.settings as qset
 zcsrgemv = qset.mkl_lib.mkl_cspblas_zcsrgemv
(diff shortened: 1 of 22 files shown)


nathanshammah:
Thanks @jakobjakobson13. A few modules' tests fail, on all operating systems, https://travis-ci.org/github/qutip/qutip/jobs/746708587. I wonder why.

By the way, what software did you use, `autoflake`?

jakobjakobson13:
> Thanks @jakobjakobson13. A few modules' tests fail, on all operating systems, https://travis-ci.org/github/qutip/qutip/jobs/746708587. I wonder why.
> 
I think I went a bit to far. Perhaps I will close this pull request and cut it down into smaller chunks.

> By the way, what software did you use, `autoflake`?
>
No, `deepcode.ai` but I don't know really know how good or bad it performs in contrast to other static code checkers.

jakelishman:
It would maybe be good to keep everything in one PR that can be squashed into a single commit at the end - this is only a single "logical" change at the end of the day.  Nice to get these things tidied up.

You probably need to run all the tests locally before just pushing to check if everything works; QuTiP has quite a few code generation steps and indirect evaluation that can't be detected with static analysis, which is why you're seeing a lot of test failures (though that's _mostly_ `.pyx` files).  Something seems wrong with the static analyser if it's not spotting that `Qobj` is used _everywhere_, though, and I notice in several commits it has a lot of false positives for removal.  I can't imagine any situation where a static analyser should remove an in-library import from an `__init__.py` file, but it's done that in a couple of places.

Similarly, in `cy/pyxbuilder.py` I can see that it's removed a line `import pyximport`, but I can see that that import _is_ used - it's so close it's even within the minimal diff!

PR #1405, opened 2020-12-11
Local multiplication routine, targeted at Dense Qobj

jsmarsha11:
**Description**
Added one new public function `local_multiply_dense` in a new file qutip/qip/operations/local_operations.py, and tests at tests/test_local_operations.py.

This can dramatically improve the speed for performing certain matrix multiplications, taking advantage the new Dense data layer. In particular, for cases where the density matrix/state has few zero's, and we want to multiply a local operator (e.g. one acting on only a small subset of the full system), this routine is expected to be faster.

**What is does**
Using some standard tools, such as NumPy's einsum one can compute the local matrix operation directly, i.e. applying a one or two-local gate on a Qobj of many qubits. This is only really useful if one uses the Dense data layer, since otherwise there are conversion overheads which kill any speed-up (at the end of the day, we need NumPy arrays).

The code has two 'backends', one using einsum as mentioned, and another using what I call a 'vectorization' approach (where the density matrix or vector is reshaped to a vector over the target qubits).

Note: It will work on qudits in general (does not have to be qubits).

**Performance**
I find, in systems where the state is dense (few zeros),  one can get a speed-up of between 2 and up to around 10x faster (depends on the size and system etc). Of course, sometimes using the standard built in sparse routines are faster, so it really depends what one is doing. Attached is a wall-clock time comparison between the standard sparse way for a 1-local multiplication on a vector (using `gate_expand_1toN`), and the added code (dense local). N being the number of qubits. Here I made things as hard as possible for the sparse routine (all elements non-zero), so this is probably the limit of the speed-up.

1_local_vector_adaptive_backend

**Other notes**
1. I am a contractor for NASA, and they required a slight change to the license for the code I added. This is why I made two new files for this work specifically (instead of putting it in e.g. gates.py), which now includes a Copyright NASA sentence. I will talk privately with someone about this.

2. ~~I actually think that we could make a wrapper function `local_multiply` which can additionally take a backend 'sparse', so that it is easier to compare the above code, to one using the default sparse way (e.g. using the `expand_operator` type functions). I didn't do this yet as thought i'd see what thoughts are on this -- I can easily add this.~~ **Update (20/12/20)**: I just added some basic functionality for this. Will need to add more tests and maybe re-work it a bit still.

3. Currently the input local operator is expected to be square, and we do not take output dimension arguments. This means it keeps the system dimensions the same. This can be relaxed easily, but we probably need an additional output dimension list, which needs to be consistent with the input operator. Again, this is easy to do, but was not sure if it is needed. If I get time I might add this anyway.

4. There are some examples of the code in the docstring to see how to run it (and also in the tests of course). I also have offline a script to perform certain speed comparisons (vs the sparse routine), I can share some more results of that. I figured this type of script shouldn't go into the project, but I can also add this.

**Changelog**
Added a local multiplication routine, to speed up certain dense matrix operations.


nathanshammah:
Thank you @jsmarsha11!

jakelishman:
Sorry I didn't get round to commenting on this PR sooner.  Since it's now come up in #1579 around the subject of sorting out licensing in general in QuTiP, let me also note some stuff about that here.  #1579 proposes to remove all individual-file licence text, because it is in general redundant, and makes it harder for us to centrally manage our licence.  There is some concern (mostly mine), that it will be a little difficult for us to manage additional code which does not sign over copyright to the QuTiP admin team; if we accept it, we will have to keep the notice in perpetuity.  This will pose a bit of a Theseus' ship problem for us if we ever want to re-organise the file that contains this code; after how many modifications is the code no longer NASA's, but QuTiP's?  As long as it's not _ours_, we can't separate it out into different files, or mix it with other code (say additional implementations).  I'm definitely not a lawyer, but I'm worried that will cause us code organisation problems in the future, and while the licence would give us permission to improve the code, it's the reorganisation I'm worried about.

Secondly, about the code itself: this sort of local operation is something I absolutely would like to include, but I'm not certain that the form it's in right now is the best, general way to do it.  I think this is part of much larger discussion about how to handle tensor networks - for example, at the moment, `targets` is fine for states and operators, but it won't work for superoperators.  I'm not sure that this is solvable in an ergonomic way with QuTiP's current dimensions specifiers, but it's something we're definitely going to look into after the initial release of 5.0.

Also, there are already several disparate functions in QuTiP that provide sort-of similar functionality, and I think we need to have a proper go at tidying them all up into one cohesive interface.  This includes `subsystem_apply` (most similar to this), but also `ptrace` and `partial_tranpose`, which can be seen as special cases of this exact same type of tensor contraction (not to mention `tensor_contract`!).

The other thing from a code perspective is that this doesn't use the data-layer in the intended manner.  A lot of this may be because our documentation of the data-layer isn't readily available yet (we weren't expecting interest this early!).  It does individual detection on two different types, rather than creating a `Dispatcher` to handle the multiple dispatch over potentially different data types of the two inputs.  I would imagine that a better form of organisation for this sort of routine, which will be able to handle arbitrary tensor-network operations, will end up being made up of two components:
1. a `Dispatcher` version of something akin to `einsum`, with a couple of limitations
2. a high-level wrapper function to handle the specific case of local multiplication, which examines the dimensions and target specifiers to produce the input to the `einsum`-like function, then calls the `Dispatcher` with this information.

I think this form would likely be preferable for several reasons:
1. it will allow arbitrary mixing of different data types (`CSR` complete space and `Dense` operator, or both `CSR`, or whatever), with the allowed conversions respecting the global dispatcher rules, not having this function use a different, special configuration
2. it will mean that the function can also be implemented by plug-in data types (TF/CuPy/etc), installed separately to QuTiP
3. it will be more general, and easier to maintain; the current `subsystem_apply`, `partial_transpose`, `ptrace`, `tensor_contract` and this can all become special cases of this backing `einsum`-like routine.
4. after we've implemented new dimensions objects, we'll be able to use a more standardised description of how to refer to individual subspaces of a Hilbert space, which will provide a more consistent UX across the library.  This is particularly important if we want this to work well with superoperators.

Really, thank you very much for making the PR - I really am excited to see people with an interest in the new data layer!  I'm sorry that I took quite so long to respond properly to it, and that when I have, I've ended up being somewhat against the implementaton as it is.  I would be really interested in pursuing this, likely after the initial release of 5.0, but right now I think we might need to get our ducks in a row about licensing primarily, and then about how we're going to handle tensor networks in general.  In order to prevent headaches down the line (and again, very very sorry), would it be ok if we leave this unmerged, and don't attempt to modify it in any way?  I don't want to get into a situation where we decide we can't accept non-QuTiP-licenced code, and we want to use/build on your code or implement something similar, but our hands are tied because anything we do will involve our knowledge of IP we don't control.

jsmarsha11:
Hi @jakelishman,
Thanks for the update and your detailed comments! No worries about the time, I follow the project so see everyone is busy on implementing the fundamentals for v5.

I see two main points to consider:
1) the code will need to be rewritten to make a more consistent use of the data layer with the rest of the project.
2) the license issue.

For 1, indeed, I was not so familiar with the intended use of the data layer and using a `Dispatcher` as these are all new capabilities (mostly this was collecting together some old `einsum` code I had and packaging it together to work with `Qobj`). If we think this type of thing will indeed be useful (I personally still do given the potential speed ups), I am happy to try to implement things in the way you mention when I get the time to go through this exercise (most likely will require your input also). If eventually we can converge on a design we can try to merge it into a later version than 5.0.

For 2, yeah I totally understand all the points you mention. I will write to you about this issue.

So overall I see this progressing as follows:
I can close this PR and will start a fresh branch implementing this in the intended manner. I expect this could take a while as I figure out how the `Dispatcher` etc. works, but also I will be doing this in my free time (I don't see this as a problem since you will release V5 without this and we can see about merging it in later).

Thanks, and feel free to let me know if you have any problems with the above plan or any other questions.
Otherwise I can let you know when I have updates.

PR #1407, opened 2020-12-16
Core update for solver in v5

Ericgig:
**Description**
First part of rework of solver for dev.major. It allows QobjEvo to support any data types for both itself and the state. And reuse the QobjEvoFunc of #1123 so function can be used in any solver. 

- Update `cQobjEvo` to use `Data` object instead of `CSR` matrices only.
- Split `cQobjEvo.matmul` and `expect`: used to accept only `Dense` state, now accept any data type and use Jake's dispatcher. But added `matmul_dense` and `expect_dense` to do operation inplace when possible. Dispatche on `inplace` out parameter is not possible and such operation are mostly limited to `Dense`.
- Compiling string coefficients will now create the build directory in `.qutip` folder instead of current active directory.
- array coefficients addition will add array when possible.
- `scipy.special` had a cython interface... added it to string coefficient's cimport as `cython_special`.
- Added an inplace addition for `Dense` object and inplace multiplication with a scalar for both `Dense` and `CSR`.
- Added `Dense` specialization for `expect`.
- Added `project_dense`.
- Added `QobjEvoFunc`, a wrapper for function `f(t: float, args: dict) -> Qobj` to be used like `QobjEvo` in solver.
- Added the method `to` to QobjEvo to control data types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,15 @@ qutip/cy/*.c
 *.dat
 qutip/core/*.h
 
+qutip/core/*/*.html
+qutip/solve/*.html
+qutip/solver/*.html
+qutip/solver/nonmarkov/*.html
+qutip/solve/nonmarkov/*.html
+qutip/solver/ode/*.html
+qutip/solver/ode/*.c
+
+
 *.qo
 
 benchmark/benchmark_data.js
--- a/.gitignore
+++ b/.gitignore
@@ -30,12 +30,8 @@ qutip/cy/*.c
 *.dat
 qutip/core/*.h
 
-qutip/core/*/*.html
-qutip/solve/*.html
-qutip/solver/*.html
-qutip/solver/nonmarkov/*.html
-qutip/solve/nonmarkov/*.html
-qutip/solver/ode/*.html
+
+qutip/**/*.html
 qutip/solver/ode/*.c
 
 
--- a/qutip/core/cy/cqobjevo.pxd
+++ b/qutip/core/cy/cqobjevo.pxd
@@ -37,15 +37,23 @@ cimport numpy as cnp
 from qutip.core.data.base cimport idxint
 from qutip.core.data cimport CSR, Dense, Data
 
+
+cpdef enum LTYPE:
+    MIXED_TYPE   = 0
+    CSR_TYPE     = 1
+    Dense_TYPE   = 2
+
+
 cdef class CQobjEvo:
     cdef readonly (idxint, idxint) shape
-    cdef object dims
+    cdef readonly object dims
     cdef str type
+    cdef LTYPE layer_type
     cdef str superrep
-    cdef bint issuper
+    cdef readonly bint issuper
     cdef size_t n_ops
 
-    cdef CSR constant
+    cdef Data constant
     cdef list ops
     cdef list coeff
     cdef object coefficients
@@ -56,7 +64,8 @@ cdef class CQobjEvo:
     cdef list dynamic_arguments
     cdef dict args
     cdef object op
-    cpdef dyn_args(self, double t, Data matrix)
 
-    cpdef Dense matmul(self, double t, Dense matrix, Dense out=*)
+    cpdef Data matmul(self, double t, Data matrix)
+    cpdef Dense matmul_dense(self, double t, Dense matrix, Dense out=*)
     cpdef double complex expect(self, double t, Data matrix) except *
+    cpdef double complex expect_dense(self, double t, Dense matrix) except *
--- a/qutip/core/cy/cqobjevo.pxd
+++ b/qutip/core/cy/cqobjevo.pxd
@@ -37,18 +37,10 @@ cimport numpy as cnp
 from qutip.core.data.base cimport idxint
 from qutip.core.data cimport CSR, Dense, Data
 
-
-cpdef enum LTYPE:
-    MIXED_TYPE   = 0
-    CSR_TYPE     = 1
-    Dense_TYPE   = 2
-
-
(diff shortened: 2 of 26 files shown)


qutip/tests/core/data/test_mathematics.py
```diff
@@ -770,3 +771,22 @@ def op_numpy(self, matrix):
...
+            return np.outer(np.conj(matrix), matrix)
+        else:
+            return np.outer(matrix, np.conj(matrix))
+
+    shapes = [
+        (pytest.param((1, 1), id="scalar"),),
+        (pytest.param((1, 100), id="bra"),),
+        (pytest.param((100, 1), id="ket"),),
+    ]
+
```
jakelishman, line 788: If I remember correctly, you should also supply some `bad_shapes` to test as well, making sure that you can project an operator?

qutip/core/cy/cqobjevo.pxd
```diff
@@ -37,15 +37,23 @@ cimport numpy as cnp
...
+cpdef enum LTYPE:
+    MIXED_TYPE   = 0
+    CSR_TYPE     = 1
+    Dense_TYPE   = 2
+
+
 cdef class CQobjEvo:
     cdef readonly (idxint, idxint) shape
-    cdef object dims
+    cdef readonly object dims
```
jakelishman, line 49: This gives a false sense of safety - while `readonly` prevents you from reassigning the `list`, since the type is mutable you can still just change it anyway and mess stuff up.  That said, it's not like we can _really_ stop anyone in Python from doing whatever the hell the like, if they really put their mind to it.

jakelishman (commented):
I've only looked at this PR so far - if there are changes in this one, all the others will have to change too, so I'll hold off until then.

My major comments here are on how to handle the dispatch at the C level.  I never had time to write design documents, but I think it is something we ought to talk about.  At the least, we should probably have the dispatching mechanisms inside the compile units of those operations (so like the C-level `matmul` dispatcher should be in `matmul.pyx`).

That said, it's still harder than I suggested - for example, you kind of want a partial dispatcher for `matmul`, where the second argument and output are fixed to `Dense`.  It might be worth us profiling to see if that's something that's absolutely required, or if we can just use a generic method without any measurable impact on speed.

.gitignore
```diff
@@ -30,6 +30,15 @@ qutip/cy/*.c
...
 
+qutip/core/*/*.html
+qutip/solve/*.html
+qutip/solver/*.html
+qutip/solver/nonmarkov/*.html
+qutip/solve/nonmarkov/*.html
+qutip/solver/ode/*.html
+qutip/solver/ode/*.c
+
+
```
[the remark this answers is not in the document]
Ericgig: Never crossed wildcard directories before... good to know.

qutip/core/cy/cqobjevo.pxd
Ericgig: They can change them at the Qobj and QobjEvo level anyway. This could be solved later when we use more than list for dims.

qutip/core/qobjevo.py
```diff
@@ -691,6 +727,49 @@ def tidyup(self, atol=1e-12):
...
+        """
+        res = self.copy()
+        cte_res = function(res.cte, *args, **kw_args)
+        if not isinstance(cte_res, Qobj):
+            raise TypeError("The function must return a Qobj")
+        res.cte = cte_res
+        for op in res.ops:
+            op.qobj = function(op.qobj, *args, **kw_args)
+        res._compile()
+        return res
```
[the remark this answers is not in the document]
Ericgig: I like the `linear_map` name,  lot clearer than `apply`. I will not add `coefficient_mapping` now, but it's on the table if it there is interest.

Ericgig:
I made some benchmarks timing different way to call matmul and contouring the dispatcher is worth it. 
For 10x10, CRS x Dense, it's 4x slowdown for matmul and 2x for expect.
For 1000 x 1000, it's 10% and 5%, reasonable but still felt.

The easiest would be having `matmul(self, Dense in, Dense out)` method for Data layer. There is no need for all pairs of layer types to be supported by this, states in scipy's solver are always Dense. But it's a little late for that

Not having that, since 99% of users will use our data type and we only really need it for `matmul` and `expect`, we can so a manual disptach for just those.

The way I did it was quite ugly, I wanted to limit the calls to `isinstance` which are not proper cython so did it once moving the type. I did not think of `type`...
`type(op) is CSR` is 15x faster than `isinstance(op, CSR)` and about as fast as comparing for enums, making have `layer_type` useless. Renamed it to `matmul_data_dense_dense` in matmul.pyx.
Still short of a c dispatcher for function pointers, but I am not attacking this.

qutip/core/cy/cqobjevo.pxd
jakelishman: Ok, I'm happy with this then.  `readonly` is correct in principle, and the new `dims` will be immutable (I'm pretty sure).

qutip/core/qobjevo.py
```diff
@@ -737,13 +736,15 @@ def permute(self, order):
             op.qobj = op.qobj.permute(order)
         return res
 
-    # function to apply custom transformations
-    def apply(self, function, *args, **kw_args):
+    # function to linear_map custom transformations
+    def linear_map(self, op_mapping):
         """
         Apply function to each Qobj contribution.
```
jakelishman, line 742: "function" -> "mapping" to match parameter name change?

jakelishman (approved):
LGTM other than that one comment about a word in a docstring.

PR #1408, opened 2020-12-16
CSC data layer

Ericgig:
**Description**
Add CSC data-layer. For our use, CSC is almost always slower than CSR...
I build it for testing possible use in AHS.

I reused the CSR code as much as possible. Specializations where the *free* transformation `CSC -> transpose(CSR)` could be used have been added.

**Related issues or PRs**
Build on top of #1407.


qutip/core/data/csc.pyx
```diff
@@ -0,0 +1,434 @@
...
+    out.row_index = matrix.col_index
+    out.size = matrix.size
+    out.shape = (matrix.shape[1], matrix.shape[0])
+    if copy:
+        out._deallocate = False
+        return out.copy()
+    else:
+        out._deallocate = True
+        matrix._deallocate = False
+        return out
```
jakelishman, line 409: This looks like it'll segfault - it's very difficult to manage this sort of data ownership like this, which is why I deferred to `numpy` as soon as it got even a little complicated.  In this case, there's a bug if `matrix._deallocate` is `False` on entry to the function and `copy=False`, then `out._deallocate` will get set to `True`, but that isn't safe since `matrix` doesn't own its data.  You also _can't_ fix this with

```python
else:
    out._deallocate = matrix._deallocate
```

because if `out` goes out of scope before `matrix`, then access to `matrix` will segfault.  We can't allow any segfaulting operation to be callable from Python space (and even leaving it in C space is very questionable).

I can think of two ways to make this safe, but I'm not convinced by either of them:

1. add a reference to the underlying object into the returned object to ensure they never go out of scope together, but this is basically reinventing Python and numpy's garbage collection and ownership semantics, and I have no doubts that we'll get it wrong 100 times.
2. force the creation of the `scipy` view of the input, and then use our fast `scipy.sparse.csr_matrix`/`...csc_matrix` constructors to produce one for the output and then fake an unpickle from that.  That lets numpy handle the data ownership safely, but it'll be difficult to follow the internal logic.

If I had to pick, I'd go with option 2, but I'm not totally certain that's the best way.

jakelishman (commented):
this isn't a proper review, just a single comment I had on a flick through

qutip/core/data/csc.pyx
Ericgig: `as_tr_csr` and `from_tr_csr` are to be used in pair when using CSR specializations for CSC. So instead of needing them to work in all cased we can have them fail when miss-used to signal us to rework the affected specializations.
I added a check in `to_scipy` to refuse to create scipy matrix if `_scipy` does not exist it does not own data and `from_tr_csr` will refuse to steal ownership if the target does not have it.

I made them generic ValueError, maybe BufferError is more appropriate...

qutip/core/data/csc.pyx
```diff
@@ -395,7 +399,7 @@ cpdef CSC from_tr_csr(CSR matrix, bint copy=True):
     If copy is False, steal data ownership from the CSR.
     """
     if not matrix._deallocate and not copy:
-        raise ValueError("Input matrix should own its data.")
+        raise ValueError("Input matrix should own its data if not copying.")
```
jakelishman: This now fails in the case that the user passes in a matrix which is already scipy-backed, which prevents people from using this function in Python-space code.  We'd designed the data layer with the idea that you can extend our builtin specialities from Python space,  but since a lot of matrices creating in Python space _will_ be scipy-backed, this generally breaks that.

If you always create a scipy view onto the base matrix, you can safely do this without copying by storing scipy views onto the same numpy arrays, and that avoids the need to reinvent refcounting elsewhere.  I don't think the data layer should be raising exceptions for anything other than a completely impossible user error, and from Python space (which this function is accessible to) a developer should never have to know whether a C object owns its data or not.

qutip/core/data/csc.pyx
```diff
@@ -185,6 +186,9 @@ cdef class CSC(base.Data):
         # be collected while we're alive.
         if self._scipy is not None:
             return self._scipy
+        if not self._deallocate:
+            # Does not own the data, thus cannot give it to scipy.
+            raise ValueError("Can't pass data ownership to scipy")
```
jakelishman, line 191: We can't throw an exception in `as_scipy()` - it violates the core principle that Python space should be able to modify data-layer objects, which prevents users from being able to use the objects unless they compile their own code.

You can create `ndarray`s that don't own their data, but you need to set the `base` property of the `ndarray` to the relevant owner in Python space to ensure that the references don't go stale.  To do this, we'd need to add a `base` property to all the data-layer objects to store the ownership, and we're suddenly re-inventing numpy's refcounting, which we certainly will get wrong in the future.

As soon as we have views onto data, we should let numpy handle it.  The `as_scipy()` routines aren't very expensive at all, really, and having the only alternative to deallocation being "defer to numpy" rather than "defer to another data-layer object" means that we completely avoid having to do refcounting ourselves, and we avoid throwing any exceptions in the data layer.

jakelishman: That said, there's clearly a problem with my `CSR._deallocate` here - it's probably redundant, because the only situation you _should_ be able to get a `CSR` that doesn't own its data is to have a `scipy` backing.  I think the reason it exists is to do with the `__init__` method possibly raising an exception, but I'm not 100% certain.

The objects are all designed under the assumption that the only other possible owner of data is numpy, but I think we should make that more rigorous.  I really don't think we should be handling any ownership concerns ourselves beyond "I personally allocated this pointer, so I will also deallocate it" - you can see how easy it is to make small mistakes, whereas numpy is battle-tested.

jakelishman: If I've done anything similar in currently existing code, we should change it as well - I don't _think_ I did, but if you do spot anything let me know.

Ericgig: I am not sure I am following you here.
If a CRS layer has `_deallocate = False` and `_scipy = None`, wouldn't `as_scipy()` break the reference counting anyway?
So the exception is not the issue, but the user having access to those object.

I never intended user using these function and had them `cpdef` for my own debugging...
I hid them and let the user access only to the copy version. The error should them only appear when we mess up in development but never to the user.

jakelishman: Yeah, that's why I mean there's a redundancy in `CSR` - it shouldn't be possible to obtain an object with `_deallocate = False` and `_scipy = None`, so having a system where such an invalid object can exist was bad programming on my part.

I think the functions could be accessible to Python space - it's a useful operation that can safely exist with the scipy views.  I suppose it doesn't matter so much if it's `cdef` for now to hide it and we make it visible later with the correct safety procedures. I don't really like the idea of committing something that's a bit fragile, even if it's only intended for devs - I don't like `csr.fast_from_scipy` for a similar reason, really - but if it's compile-only it's probably ok. I should change the unpickle methods to put the proper error checking in, though, since they're absolutely required to be Python accessible.

jakelishman: I'm worried that this reference stealing feels deeply thread-unsafe, but that's probably not a major concern - I put a little thought into it in the rest of the data layer, but not enough to be sure that all (non-stateful) operations are thread-safe.  Obviously the scatter/gather pattern in `CSR` is thread unsafe, but that's by definition a stateful operation.

hodgestar:
Closing at @Ericgig's request -- this turned out to be slower than CSR in (almost) all cases.

PR #1409, opened 2020-12-16
Evolver class and utilities

Ericgig:
**Description**
Make and upgrade a few class commonly used by solvers.

All solver's core is a loop that progress between time in `tlist` and take a snapshot of the state at these time. The work of this loop have been moved to 2 classes: `Result` and `Evolver`. 

`Result` was previously just a container of the solver output. It now does the snapshot: compute the expectation values, save the state, compute trajectories average, etc. 

`Evolver` is qutip's ODE solver interface: the states are `Data` object and the function is `QovjEvo`(or `QobjEvoFunc`). There are multiple type of evolver wrapping different scipy's ode solver and a few of our own. The wrapper around scipy's zvode that was previously used in qutip's is `EvolverScipyZvode`, but we now have our own cython solver based on [Verner's method](http://people.math.sfu.ca/~jverner/) that accept any Data type as state. Different evolver based on different ODE method are interchangeable and have their dispatcher `evolver_collection`, allowing to easily add new methods. 

With this, time-dependent system are managed by `QobjEvo`, ODE by `Evolver`, therefore the solvers can concentrate more closely on the physic.

**Evolver:**
Method supported with this PR: `adams`, `bdf`, `dop853`, `lsoda`, `vern7`, `vern9`, `diag`. The 4 first are from `scipy.integrate.ode`, the 3 last are ours. `vern7` and `vern9` are runge-kutta methods of 7th and 9th order using [verner's most efficient coefficient](http://people.math.sfu.ca/~jverner/) with dense output. `diag` diagonalize a constant system and solve the ODE in that basis.
The minimum that must be given for a new evolver are the `prepare`, `set_state`, `get_state` and `step` methods.

`evolver_collection` work as a dispatcher over evolvers. Solvers obtain their evolver instance using `evolver_collection[method, rhs]` with both `method` and `rhs` taken from the options. `evolver_collection` keep a list of description, used options and capacities for each member. It can list the available method matching desired condition. The `rhs` options is not used yet in this PR, it is for future Adaptative Hilbert Space and other modification to the ODE that affect the matrix-state product.

##### `SolverQEvo`, `Feedback`:
Wrapper around `QobjEvo` that handle feedback arguments and provide `ndarray` interface.

##### qutip/solver/ode/...
Code qutip's evolver, mostly for Verner's method. 

##### progress_bar:
Added a progress bar passed on `tqdm`. `update` no longer need the iteration number.

##### parallel map:
- Added a `loky` based parallel map, hopefully will help windows users.
- Added a `reduce_func` input. Presently the map function is saving all results and the average, std, etc. are only computed at the end. There was no real advantage in saving only average values since all trajectories were in memory at the end of the map. But users expected lower memory usage when asking for only the results average.
- Added `timeout` options to the map, when passed, no error is raised but obtained results are returned. This allow to ask for 30s worth of trajectories instead of fixing a number.

**Related issues or PRs**
Build on top of the #1407


qutip/solver/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .evolver import *
```
quantshah, line 1: Should we not go for absolute imports and explicitly writing down all imports if not too many? This is the convention for most other imports in QuTiP. 
```
from qutip.solver.evolver import ... 
```

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
+
+from ..evolver import Evolver, evolver_collection
+from .verner7efficient import vern7
+from .verner9efficient import vern9
+from .wrapper import QtOdeFuncWrapper
+import numpy as np
+from ...core import data as _data
```
quantshah, line 7: Same comment about relative vs absolute imports.

qutip/solver/__init__.py
jakelishman: In dev.major we swapped to preferring relative imports everywhere in qutip when you're staying within the same package, so this is the consistent choice.  Absolute imports disguise circular import issues, because it's much less obvious when you're leaving a package, and particularly with us rearranging a lot of internal modules, absolute imports make it much harder to modify this sort of thing. The absolute imports remaining should only be where legacy code is circular, in cython `cimport`, or when accessing a different package. In the latter case, you shouldn't be specifying the internal structure though, you should only be importing from the public API, because everything else is liable to change.

It's also a lot of typing for a questionably negative outcome, especially if it ends up in user code - a user should never have to know what file we defined `tensor` in, they should just be able to do `qutip.tensor`. There's no benefit to making them type `from qutip.core.tensor import tensor`, and we make no guarantees that we won't rearrange the internal structure at some point, so if they do type that, it may break in a later version. It's the same thing between separate packages.

qutip/solver/ode/evolver.py
jakelishman: The last line should _probably_ be `from qutip import data as _data` since it's a separate package and exported to the public interface, but I'm fairly sure I've also been a bit lax elsewhere with that one too. The others are within the same package, and it's easy to see that at a glance.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+    # Solve an ODE for func.
+    # Calculate the required expectation values or invoke callback
+    # function at each time step.
+    #
+    # Use verner method implimented in cython
+    #
+    _error_msg = ("ODE integration error: Try to increase "
+                  "the allowed number of substeps by increasing "
+                  "the nsteps parameter in the Options class.")
+    description = "qutip implementation of verner most efficient runge-kutta"
```
quantshah, line 23: "Qutip implementation of Verner's most efficient Runge-Kutta."

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+    used_options = ['atol', 'rtol', 'nsteps', 'first_step', 'max_step',
+                    'min_step', 'interpolate']
+
+    def prepare(self):
```
quantshah, line 27: Missing docstring.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class EvolverVern(Evolver):
+    # -------------------------------------------------------------------------
+    # Solve an ODE for func.
+    # Calculate the required expectation values or invoke callback
+    # function at each time step.
+    #
+    # Use verner method implimented in cython
+    #
```
quantshah, line 19: This should be a multiline docstring imo since it is a public class: https://www.python.org/dev/peps/pep-0257/

```
A class to use the Verner method [] for solving ...
```

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+    def prepare(self):
+        func = QtOdeFuncWrapper(self.system)
+        opt = {key: self.options.ode[key]
+               for key in self.used_options
+               if key in self.options.ode}
+        ode = vern7 if self.options.ode['method'] == 'vern7' else vern9
+        self._ode_solver = ode(func, **opt)
+        self.name = "qutip " + self.options.ode['method']
+
+    def get_state(self, copy=True):
```
quantshah, line 36: Missing docstring.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+        state = self._ode_solver.y
+        return self._ode_solver.t, state.copy() if copy else state
+
+    def set_state(self, t, state):
```
quantshah, line 40: Missing docstring.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+
+class EvolverDiag(Evolver):
+    # -------------------------------------------------------------------------
+    # Solve an ODE for func.
+    # Calculate the required expectation values or invoke callback
+    # function at each time step.
+    #
+    # Diagonalize the Hamiltonian and
+    # This should be used for constant Hamiltonian evolution (sesolve, mcsolve)
+    #
```
quantshah, line 83: Should be a docstring for the class.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+    description = "Diagonalize a constant system"
+    used_options = []
+
+    def __init__(self, system, options, args, feedback_args):
```
quantshah, line 87: In input arguments should be documented - system, options, args, feedback_args.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+        if not system.const:
+            raise ValueError("Hamiltonian system must be constant to use "
+                             "diagonalized method")
+        self.system = system
+        self.options = options
+        self._dt = 0.
+        self._expH = None
+        self.prepare()
+
+    def prepare(self):
```
quantshah, line 97: Docstring, even if one liner, should be present for automatic documentation of the API. It looks a bit weird otherwise with the documentation.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+    def backstep(self, t):
+        return self.step(t)
+
+    def get_state(self, copy=False):
+        y = self.U @ self._y
+        return self._t, _data.dense.fast_from_numpy(y)
+
+    def set_state(self, t, state0):
+        self._t = t
+        self._y = (self.Uinv @ state0.to_array())
```
quantshah, line 128: I think these are self explanatory by still a single line docstring would be good to keep the docs looking nice.

qutip/solver/ode/verner7efficient.pyx
```diff
@@ -0,0 +1,484 @@
+#cython: language_level=3
+# Verner 7 Efficient
```
quantshah, line 2: In many other qutip submodules, a single line docstring is used to say what the code does. E.g.,

```
"""
This module provides an efficient method for Verner ... described in ...
"""
```

qutip/solver/ode/verner9efficient.pyx
```diff
@@ -0,0 +1,850 @@
+#cython: language_level=3
+# Verner 9 Efficient
+# http://people.math.sfu.ca/~jverner/
+import numpy as np
+cimport numpy as cnp
+from qutip.solver.ode.explicit_rk cimport Explicit_RungeKutta
+from .wrapper cimport QtOdeData, QtOdeFuncWrapper
+from .wrapper import qtodedata
```
quantshah, line 8: Why not use absolute imports? 

```
from qutip.solver.ode.wrapper ...
```

qutip/solver/ode/wrapper.pyx
```diff
@@ -0,0 +1,164 @@
+#cython: language_level=3
+
+import numpy as np
+from .._solverqevo cimport SolverQEvo
+from ...core cimport data as _data
+from ...core.data.norm cimport frobenius_dense
+from ...core.data.norm import frobenius
```
quantshah, line 7: Why not absolute imports?

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
+__all__ = ['SolverOptions',
+           'SolverResultsOptions', 'SolverOdeOptions',
+           'McOptions']
+
+from ..optionsclass import optionsclass
```
quantshah, line 5: I would suggest absolute imports.

qutip/solver/result.py
```diff
@@ -0,0 +1,586 @@
+
+
+import numpy as np
+from ..core import Qobj, QobjEvo, spre, issuper
```
quantshah, line 4: I would suggest absolute imports.

qutip/solver/result.py
```diff
@@ -0,0 +1,586 @@
...
+        return self.average_final_state
+
+    @property
+    def num_traj(self):
+        return len(self.trajectories)
+
+
+class MultiTrajResultAveraged:
+    """
+    Contain result of simulations with multiple trajectories.
```
quantshah, line 393: I think the initialisation parameter `c_ops` should be documented in the class definition or the __init__.

qutip/solver/result.py
```diff
@@ -0,0 +1,586 @@
...
+        self._sum_states = None
+        self._sum_last_states = None
+        self._sum_expect = None
+        self._sum2_expect = None
+        self._to_dm = True # MCsolve
+        self.num_c_ops = num_c_ops
+        self._num = 0
+        self._collapse = []
+
+    def add(self, one_traj):
```
quantshah, line 406: Missing docstring.

qutip/solver/ode/verner9efficient.pyx
jakelishman: To me this is the poster child for relative imports.  As far as I see, absolute style only has negatives: it would be much less clear how the module `wrapper` relates to this one (i.e. part of the same package), it adds maintenance cost if any part of the path ever changes (and lots of internal paths do change in `dev.major`), and it's generally just harder to scan the lines because there's more fluff.  PEP 8 suggests
> explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

QuTiP's certainly got a complex package layout (this bit's nested 4 deep).

I agree with you on stuff like `from ...core` though - that should be `from qutip import x` (`core` is physical organisation only - it's transparent to the user).

When I separated out the core package, which modified every `import` statement, I generally tried to stick to "absolute when crossing an API boundary, relative within the same API implementation", which is why loads of `dev.major` swapped to relative imports.  I'm sure I didn't get it perfect, though.

qutip/solver/ode/verner7efficient.pyx
```diff
@@ -0,0 +1,484 @@
...
+        cdef int i, j
+        cdef double t = self.t_front
+
+        for i in range(self.rk_step):
+            (<QtOdeData> self.k[i]).zero()
+
+        self.f.call(self.k[0], t, self._y_prev)
+
+        self._y_temp.copy(self._y_prev)
+        self._y_temp.inplace_add((<QtOdeData> self.k[0]), dt * self.a[1][0])
```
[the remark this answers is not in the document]
hodgestar: Slightly more radical suggestion -- could we just use the SciPy Runge Kutta solver and specify our own parameters?

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Returns options class to be used as options in evolution solvers.
+
+    The default can be changed by::
+
+        qutip.settings.solver['order'] = 10
+
+    Options
+    -------
+    """
```
hodgestar, line 30: The documentation for the options seems to be missing here.

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@optionsclass("ode", SolverOptions)
+class SolverOdeOptions:
+    """
+    Class of options for evolution solvers such as :func:`qutip.mesolve` and
+    :func:`qutip.mcsolve`. Options can be specified either as arguments to the
+    constructor::
+
+        opts = SolverOptions(order=10, ...)
```
hodgestar, line 47: `SolverOptions` here should be `SolverOdeOptions` but the rest of the opening parts of the docstrings for the two clases should also say something different.

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@optionsclass("results", SolverOptions)
+class SolverResultsOptions:
+    """
+    Class of options for evolution solvers such as :func:`qutip.mesolve` and
+    :func:`qutip.mcsolve`. Options can be specified either as arguments to the
+    constructor::
+
+        opts = SolverOptions(order=10, ...)
```
hodgestar, line 117: Same comment here.

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
...
+    Options
+    -------
+    store_final_state : bool {False, True}
+        Whether or not to store the final state of the evolution in the
+        result class.
+    store_states : bool {False, True}
+        Whether or not to store the state vectors or density matrices in the
+        result class, even if expectation values operators are given. If no
+        expectation are provided, then states are stored by default and this
+        option has no effect.
```
hodgestar, line 139: `normalize_ket` should also be documented.

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
...
+        # "average_expect": True,
+        # average expectation values
+        # "average_states": False,
+        # average mcsolver density matricies assuming steady state evolution
+        # "steady_state_average": False,
+    }
+
+
+@optionsclass("mcsolve", SolverOptions)
+class McOptions:
```
hodgestar, line 164: Perhaps this should be called `McsolveOptions` to make it more clear what the class if for. I initially thought it was a generic options class of some sort.

qutip/solver/options.py
```diff
@@ -0,0 +1,215 @@
...
+    Options
+    -------
+
+    norm_tol : float {1e-4}
+        Tolerance used when finding wavefunction norm in mcsolve.
+    norm_t_tol : float {1e-6}
+        Tolerance used when finding wavefunction time in mcsolve.
+    norm_steps : int {5}
+        Max. number of steps used to find wavefunction norm to within norm_tol
+        in mcsolve.
```
hodgestar, line 192: Some parameter documentation missing here too.

qutip/solver/parallel.py
```diff
@@ -0,0 +1,289 @@
...
+
+def get_map(options):
+    if "parallel" in options['map']:
+        return parallel_map
+    elif "serial" in options['map']:
+        return serial_map
+    elif "loky" in options['map']:
+        return loky_pmap
+    else:
+        raise InputError("")
```
hodgestar, line 289: This should have an error message and should use `ValueError`.

qutip/tests/solver/test_evolver.py
```diff
@@ -0,0 +1,168 @@
+
+from qutip.solver.evolver import *
+from qutip.solver.options import SolverOptions
+import qutip as qt
+import pytest
+import numpy as np
+from qutip.core.qobjevofunc import QobjEvoFunc
+
+def id_from_pair(pair):
+    #raise Exception(str(pair) + "   " + str(pair[0]) + "    " + str(type( pair[0])))
```
hodgestar, line 10: Debugging code that could probably be deleted?

qutip/tests/solver/test_parallel.py
```diff
@@ -0,0 +1,103 @@
...
+
+from qutip.solver.parallel import parallel_map, serial_map, loky_pmap
+try:
+    import loky
+except ModuleNotFoundError:
+    loky = False
+
+def _func1(x):
+    return x**2
+
```
hodgestar, line 46: We should add a test for passing an unknown map and the error raised.

setup.py
```diff
@@ -194,6 +198,12 @@ def write_version_py(filename='qutip/version.py'):
         '_steadystate',
         '_stochastic',
     ],
+    'solver': [
+        '_feedback',
+        '_solverqevo',
+    ],
+    'solver.ode': [file.split("/")[-1][:-4]
+                   for file in glob("qutip/solver/*/*.pyx")],
```
hodgestar, line 206: I suspect these won't work on Windows because the path separator there is not `/`. If that's important, we should fix it.

qutip/solver/ode/wrapper.pyx
```diff
@@ -0,0 +1,164 @@
...
+    cdef object f
+
+    def __init__(self, object f):
+        self.f = f
+
+    cpdef void call(self, QtOdeData out, double t, QtOdeData y):
+        out.copy(QtOdeData(self.f(t, y.raw())))
+
+
+cdef class QtOdeData:
```
hodgestar, line 30: I don't understand what these `Data` classes are for and ideally it would be nice to make them go away somehow and use more ordinary things like Qobjs somehow.

hodgestar (commented):
Hello!

The new solvers will be really cool, and looking forward to having them in.

I am still quite new to QuTiP, and very new to this part of QuTiP, so hopefully if some of my comments don't make sense you will correct me.

Overall I felt that some bits got a bit more complex than they really need to be (e.g. the Runge-Kutta solver code) and it would be nice to find ways to make some of that complex code not need to be there.

Looking forward to hearing your thoughts and feedback & chatting more in the PR.

qutip/solver/ode/wrapper.pyx
jakelishman: it looks like they're meant to be dispatched operations over the core `qutip.core.data` types (`CSR` and `Dense`) at the C level.  I assume it's being done here to avoid dynamic allocations in tight loops, but it really just doesn't feel right at all.

`copy` is already implemented on data types, and in addition `raw`, `data` and `set_data` would all be meaningless if the `QtOdeData` type was dropped.  The two numpy-like types can simply be dropped and replaced with conversions up to `Dense` - there shouldn't be any meaningful data kicking around inside QuTiP which _isn't_ in our `Data` formats anyway.

`iadd`, `imul` and `izero` could just be defined on the `Data` types themselves - these data types are meant to help us in QuTiP, not cause us to write huge amounts of boilerplate to wrap them for every specific instance.  It's nice if it's easy to implement a new data-layer type, but that should be a secondary concern.

While we want to avoid Python-space calls while in C mode, I think it's fine to "hard-code" C-level dispatchers that operate _purely_ on single types - the cost of having (defined in `data.norm`,
```cython
cdef double frobenius_(Data data):
    dtype = type(data)
    if dtype == CSR:
        return frobenius_csr(data)
    if dtype == Dense:
        return frobenius_dense(data)
    return np.norm(data.to_array())
```
isn't that high.  It's not ideal, but short of dropping all the way to C++ to get more performant run-time type-dispatch, I'm not sure there's a better way to do it in Cython.

qutip/solver/ode/evolver.py
Ericgig: There is 2 version of Verner method for each order, 'most efficient' and 'most robust', name given by the author.
See http://people.math.sfu.ca/~jverner/.

qutip/core/cy/cqobjevo.pxd
```diff
@@ -52,6 +52,7 @@ cdef class CQobjEvo:
 
     cdef void _factor(self, double t) except *
 
+    # To remove when safe
```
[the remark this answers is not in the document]
Ericgig: `QobjEvo` still have methods that where used for feedback support, but this will be moved to `SolverQEvo`.
There are still calls them the in `[se|me|mc]solve` function.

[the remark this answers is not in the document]
jakelishman: I'm not sure I see why there should be a special data object for handling time evolution within solvers at all?  Even with feedback, it does seem to me that the solvers themselves should be handling it, not the data - only the solvers should know which feedback ought to be generated.

I'm not really sure why we need more data classes than `Qobj` and `QobjEvo`.  The solvers should be specifying the behaviour, rather than making a new data class each time - mixing behaviour and data too much is why the existing data classes never seem to fit quite right.  Really with the advent of `Coefficient`, even `QobjEvo` isn't completely necessary for speed - all the vdispatched time-dependence stuff is in `Coefficient` - but since we already have it there's not as much of a problem keeping it.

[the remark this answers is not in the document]
jakelishman: Like, as far as a `Coefficient` is concerned, I wouldn't expect it to know anything about whether an argument is generated by a solver or set by a user as a constant.  Say I had a coefficient that was the cos of a frequency multiplied by an expectation value: it could be implemented something like

```python
time_dependence = qutip.coefficient('cos(expect_z * frequency * t)'
H = [qutip.sigmax(), [qutip.sigmay()]]
state = qutip.basis(2, 0)
times = np.linspace(0, 10, 101
feedbacks = [ExpectationFeedback(qutip.sigmaz(), argname='expect_z')]
qutip.sesolve(H, state, times, feedback=feedbacks)
```

Then `Coefficient` only knows about `args`, and the solver knows which feedbacks it should generate.

qutip/solver/_solverqevo.pyx
```diff
@@ -0,0 +1,116 @@
...
+        self.num_calls += 1
+        return self.base.matmul(t, vec)
+
+    cdef _data.Dense mul_dense(self, double t, _data.Dense vec, _data.Dense out):
+        if self.has_dynamic_args:
+            self.apply_feedback(t, vec)
+        self.num_calls += 1
+        return self.base.matmul_dense(t, vec, out)
+
+    def set_feedback(self, dict feedback, dict args, bint issuper, bint norm):
```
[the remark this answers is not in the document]
Ericgig: The old name was `dynamic_arguments`, but I feel `feedback` to be more descriptive. Which name you prefer?

In previous version the function of a time-dependant system could be f(t, state, args), where the solver would feed the state as a 1D colunm-stacked numpy array. This was causing confusion so I added reserved args key: if `"state_qobj"` was a key in the `args` dict, it would take the value of the state as a Qobj, etc. So "dynamic_args", but now that `QobjEvo` is user facing and independent from the [...] solve function, it need to move out.

I did not plan to make `Feedback` user facing and use it to hide the internal state of the solver. If by making their own, you mean `feedback=[QobjFeedback(...)]`, then you need to remember that call instead of a keyword. Feels the same to me. 
If you means having the user create their feedback function, that is what I am trying to avoid. I want the user to have access to a `Qobj` and not a reshaped data object. From that `Qobj`, the user can extract the needed information using a custom function.

If we want to keep it as simple as possible, we can remove `ExpectFeedback` (which is only a shortcut to computing expectation value from the `Qobj`) and have feedback being a string of the key from args to overwrite by the solver with the state. `mcsolve` would have 2 such input, one for the state and a second for the collapse.

qutip/solver/_solverqevo.pyx
```diff
@@ -0,0 +1,116 @@
...
+cdef class SolverQEvo:
+    def __init__(self, base, options, dict args, dict feedback):
+        self.base_py = base
+        self.base = base.compiled_qobjevo
+        self.set_feedback(feedback, args, base.cte.issuper,
+                          options.ode['feedback_normalize'])
+        self.num_calls = 0
+
+    def jac_np_vec(self, t, vec):
+        return self.jac_data(t).to_array()
```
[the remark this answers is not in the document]
Ericgig: Never yet, but since we have often stiff system, it would be nice if we also had radau method in the `Evolver` which would use it.

qutip/solver/_solverqevo.pyx
```diff
@@ -0,0 +1,116 @@
...
+
+    cdef _data.Data jac_data(self, double t):
+        if self.has_dynamic_args:
+            raise NotImplementedError("jacobian not available with feedback")
+        self.num_calls += 1
+        return self.base.call(t, data=True)
+
+    cpdef list get_coeff(self, double t, vec=None):
+        # TODO: Is this needed? I used it in debug but does it have any other use?
```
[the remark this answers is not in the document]
Ericgig: So +1 for no other use, erasing.

qutip/solver/_solverqevo.pyx
```diff
@@ -0,0 +1,116 @@
...
+        for dargs in self.dynamic_arguments:
+            if isinstance(dargs, CollapseFeedback):
+                dargs.set_collapse(collapse)
+
+    cpdef void apply_feedback(self, double t, _data.Data matrix) except *:
+        cdef Feedback feedback
+        for dyn_args in self.dynamic_arguments:
+            feedback = <Feedback> dyn_args
+            val = feedback._call(t, matrix)
+            self.args[feedback.key] = val
```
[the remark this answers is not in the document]
Ericgig: `args[key] ` is how the user get access to the feedback data.

qutip/solver/_solverqevo.pyx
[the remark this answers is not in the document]
jakelishman: I'm with Eric in that I don't really see a sensible way for a user to define their own feedback function.  The solvers have very different internals and mechanisms of action, so there's no really meaningful `Feedback` object a user could create that could be reused between different solvers.  That would be the job of `Coefficient` (which already exists) - it takes stuff the solvers have generated, and then can apply arbitrary actions on them (albeit right now it's limited to stuff that can be a Python one-liner if you want it compiled).

I'd imagine you actually call them with a slight broadening of the `args` dictionary to mean what variables should be in scope for time-dependences and how to generate them: the `args` dictionary passed to solvers like `mesolve` could be
```python
args = {
    variable_name: <method_of_generating>,
    ...
}
```

So the key is always a variable name, and the value tells you how to generate it, which is one of:
 - constant (value is purely numeric - matches existing behaviour)
 - expectation: value is a 2-tuple `('expect', operator)` where `operator` is some `Qobj` (or even `QobjEvo`)
 - state: value is a tuple beginning with the string `'state'`, and possibly a second element that represents the type/form of the state? `('state', 'dm')`
 - collapse: value is a 1-tuple `('collapse',)`

There's always going to need to be some way of passing on the information to the solvers that they should generate something, but I think I think "magic values" is a better approach to "magic keys" in the `args` dictionary (the latter is the case now), because 
 - "magic keys" don't have an obvious value you should give them when passing the dictionary
 - "magic keys" prevent the user from having meaningful variable names in their coefficients - it's not clear what `expect_0` will be filled with (`e_ops[0]`, but you can't tell what operator that is just by looking at the coefficient), but `expect_sigmaz` is semantically clear no matter what
 
 For example
 ```python
 args = {
     'trap_frequency': 2*np.pi * 20e3,
     'expect_z': ('expect', qutip.sigmaz()),
     'current_state': ('state', 'dm'),
     'which_collapse': ('collapse',),
}
result = qutip.mcsolve(H, rho0, times, args=args)
```

That way `args` means "scope" more (but changing the name is probably just annoying to people).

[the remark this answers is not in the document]
Comment (jakelishman) [reply qutip/solver/_solverqevo.pyx:78]: ^ Though with this way, I still don't really see a need for `SolverQEvo` at all.  It would all just be handled by regular `QobjEvo`, and the solvers would just have one extra function call to add stuff to the `scope` Python dictionary before calling `QobjEvo(t, args)`.

File: qutip/solver/ode/evolver.py [hunk 36]
```diff
@@ -0,0 +1,141 @@
...
+
+    def backstep(self, t):
+        self._ode_solver.integrate(t)
+        return self.get_state()
+
+    def step(self, t):
+        """ Evolve to t, must be `set` before. """
+        self._ode_solver.integrate(t)
+        if not self._ode_solver.successful():
+            raise Exception(self._error_msg)
```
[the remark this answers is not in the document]
Ericgig: Yes, scipy raise a warning containing more information about why it failed. It's not readily available but we should be able to fish it.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+        return self._ode_solver.t, state.copy() if copy else state
+
+    def set_state(self, t, state):
+        self._ode_solver.set_initial_value(state, t)
+
+    def backstep(self, t):
+        self._ode_solver.integrate(t)
+        return self.get_state()
+
+    def step(self, t):
```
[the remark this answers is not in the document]
Ericgig: I started by using that, (and `backstep` was also included in it), but for the implementation each ODE solver manage stepping and non consecutive step differently, making `step` quite complex. We could join then with `.step(self, t, step=False)` and move the implementation in `_one_step`, `_step`, `_backstep`.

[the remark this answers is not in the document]
Ericgig: Scipy uses `integrate` instead of `step`, we could match them.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+    description = "qutip implementation of verner most efficient runge-kutta"
+    used_options = ['atol', 'rtol', 'nsteps', 'first_step', 'max_step',
+                    'min_step', 'interpolate']
+
+    def prepare(self):
+        func = QtOdeFuncWrapper(self.system)
+        opt = {key: self.options.ode[key]
+               for key in self.used_options
+               if key in self.options.ode}
+        ode = vern7 if self.options.ode['method'] == 'vern7' else vern9
```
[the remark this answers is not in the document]
Ericgig: The evolver_collection will complain when asking for `Evolver` not existing.

qutip/solver/ode/evolver.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def get_state(self, copy=True):
+        state = self._ode_solver.y
+        return self._ode_solver.t, state.copy() if copy else state
+
+    def set_state(self, t, state):
+        self._ode_solver.set_initial_value(state, t)
+
+    def backstep(self, t):
+        self._ode_solver.integrate(t)
```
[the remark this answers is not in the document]
Ericgig: The `t` is smaller than the last one.

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,283 @@
...
+cnp.import_array()
+
+
+cdef class Explicit_RungeKutta:
+    def __cinit__(Explicit_RungeKutta self):
+        self.rk_step = 1
+        self.rk_extra_step = 1
+        self.order = 1
+        self.denseout_order = 1
+        self.can_interpolate = False
```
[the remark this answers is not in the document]
Ericgig: This is for other Runge-Kutta method, if we decide to add the traditional 4th order, it would have `can_interpolate=False`. If 
`Explicit_RungeKutta` object is used directly, it implement euler method and has `can_interpolate=False`.

qutip/solver/evolver.py
```diff
@@ -0,0 +1,628 @@
...
+        if not self._ode_solver.successful():
+            raise Exception(self._error_msg)
+        return self.get_state()
+
+    def get_state(self, copy=False):
+        t = self._ode_solver.t
+        if self._mat_state:
+            state = _data.column_unstack_dense(
+                _data.dense.fast_from_numpy(self._ode_solver.
+                                            _y.view(np.complex)),
```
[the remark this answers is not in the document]
Ericgig: Here should it be `_data.dense.Dense(array)` instead of `fast_from_numpy(array)`?

qutip/solver/ode/wrapper.pyx
Ericgig: I left the support for general python function beside QobjEvo and numpy array beside Data. Do we want to keep that? Or stick with something that work only inside Qutip?  If we drop more general support, we can remove these wrapper.

And yes, I spent a lot of time trying to work around using Data and the Dispatcher. I went with light wrapper around Data since I felt force to use the dispatcher otherwise... Now that "hard-code" C-level dispatchers are fine, I will update with that.

qutip/ui/progressbar.py
```diff
@@ -142,3 +153,39 @@ def update(self, n):
...
         print("\r", "Total run time: %s" % self.time_elapsed())
+
+
+class TqdmProgressBar(BaseProgressBar):
+    """
+    A progress bar using tqdm module
+    """
+
+    def __init__(self, iterations=0, chunk_size=10):
+        from tqdm import tqdm
```
quantshah, line 164: We could use the conditional import of tqdm as `from tqdm.auto import tqdm´ (https://github.com/tqdm/tqdm/blob/master/tqdm/auto.py). This seems to be well suited for switching between notebooks or command line code.

qutip/solver/options.py
ajgpitch: Seems like a good idea

setup.py
ajgpitch: I think it probably will work, but I think it would be better to use `os.path.split` and `os.path.join`

qutip/solver/__init__.py
ajgpitch: What about the use of the `*`, should we not be specific about which names we are importing?

ajgpitch:
A great piece of work @Ericgig. This has been wanting for a long time. Such an abstraction of the ODE solver should allow for easier development of other types of solver, and make the physics clearer in the code. I should also mean that we could add in new methods for solving ODEs more easily. Thanks for taking on this massive challenge.

ajgpitch:
It seems like the main design debate is over the need for `SolverQEvo` vs use of `Coefficient` class.

Ericgig:
Split in smaller pieces.

PR #1414, opened 2021-01-04
Fix ptrace for unusual index selections

### LINKED ISSUE #1239: ptrace called on non-composite object fails
**Describe the bug**
calling ptrace(0) should simply return the object. Instead it fails with an error

**To Reproduce**
import qutip as qp
qp.sigmax().ptrace(0)

**Expected behavior**
expect it to return the existing object

**Your Environment**
QuTiP Version:      4.5.0
Numpy Version:      1.18.1
Scipy Version:      1.4.1
Cython Version:     0.29.16
Matplotlib Version: 3.2.1
Python Version:     3.7.6
Number of CPUs:     8
BLAS Info:          Generic
OPENMP Installed:   False
INTEL MKL Ext:      False
Installation path:  C:\Users\Lior Ella\scoop\apps\miniconda3\4.5.12\envs\tfenv\lib\site-packages\qutip

### LINKED ISSUE #1325: ptrace inconsistent between sparse and dense
`Qobj.ptrace` is inconsistent between `sparse=True` and `sparse=False` arguments if the selection arguments are not in order:

```python
>>> import qutip
>>> a, b, c = qutip.rand_ket(2), qutip.rand_ket(3), qutip.rand_ket(4)
>>> sparse = qutip.tensor(a, b, c).ptrace([2, 1], sparse=True)
>>> dense = qutip.tensor(a, b, c).ptrace([2, 1], sparse=False)
>>> sparse.dims == dense.dims
False
>>> sparse.dims
[[4, 3], [4, 3]]
>>> dense.dims
[[3, 4], [3, 4]]
```

I'm fairly sure the only problem is that the dimensions come out incorrectly - for separable states, we find
```python
>>> ptraced = qutip.tensor(a, b, c).ptrace([2, 1])
>>> built = a.proj().tr() * qutip.tensor(b.proj(), c.proj())
>>> np.allclose(ptraced.full(), built.full())
True
```
which is also consistent with `_ptrace_dense` (which sorts `sel`).

Pretty sure the solution is just to sort `sel` on input, but wanted to check with @Ericgig.

### LINKED ISSUE #1413: Taking the partial trace of a quantum object over its own basis generates an error now
I apologize for not being that familiar with the technical terminology anymore as I have been out of quantum optics research for over 3 years, but I hope that I'll be able to explain the issue anyway.

**Describe the issue**
Taking partial trace of a quantum object over its basis does not return the quantum object itself anymore but produces a type error.

**To Reproduce**
Since version 4.5.0 running
```python
psi = qt.thermal_dm(2,1)
psi.ptrace(0)
```

Produces the following error

```
<ipython-input-7-87c31fb793ec> in <module>
----> 1 psi.ptrace(0)

~/opt/anaconda3/lib/python3.8/site-packages/qutip/qobj.py in ptrace(self, sel, sparse)
   1353             return q.tidyup() if settings.auto_tidyup else q
   1354         else:
-> 1355             return _ptrace_dense(self, sel)
   1356 
   1357     def permute(self, order):

~/opt/anaconda3/lib/python3.8/site-packages/qutip/qobj.py in _ptrace_dense(Q, sel)
   2203         rhomat = vmat.dot(vmat.conj().T)
   2204     else:
-> 2205         rhomat = np.trace(Q.full()
   2206                           .reshape(rd + rd)
   2207                           .transpose(qtrace + [nd + q for q in qtrace] +

TypeError: 'numpy.float64' object cannot be interpreted as an integer
```

Whereas in versions older than 4.5.0 (verified in 4.3.2 and 4.4.1) running
```python
psi = qt.thermal_dm(2,1)
psi.ptrace(0)
```
produced the Qobj psi as the output. 

```
Quantum object: dims = [[2], [2]], shape = (2, 2), type = oper, isherm = True

(0.667, 0.0,
0.0, 0.333)
```

Is this change on purpose? I've found it very useful in older versions to loop over taking the partial trace of quantum objects whether it was necessary or not and I believe that unless there is any specific reason for the change other people may prefer the older output generation. Curious what other people think, thanks!

jakelishman:
1. _Non-compound Hilbert spaces_: For state `A` on a simple non-compound Hilbert space, we should have `A.ptrace(0) == A` always.  This has appeared as something useful in a few users' scripts (see #1239, #1413).  This patch fixes the underlying bug in `_ptrace_dense` which caused this to fail (even though the logic should have supported it), and adds fast-paths to the dense and sparse partial trace functions to avoid doing work if all subspaces are being kept.

   Fix #1239
   Fix #1413

2. _Unsorted selections_: the tracing logic already ignores the order of the selection terms, this just fixes sparse ptrace's `dims` output.

   Fix #1325

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -2209,17 +2209,18 @@ def _ptrace_dense(Q, sel):
         vmat = (Q.full()
                 .reshape(rd)
                 .transpose(sel + qtrace)
-                .reshape([np.prod(dkeep), np.prod(dtrace)]))
+                .reshape([np.prod(dkeep, dtype=np.int32),
+                          np.prod(dtrace, dtype=np.int32)]))
         rhomat = vmat.dot(vmat.conj().T)
     else:
         rhomat = np.trace(Q.full()
                           .reshape(rd + rd)
                           .transpose(qtrace + [nd + q for q in qtrace] +
                                      sel + [nd + q for q in sel])
-                          .reshape([np.prod(dtrace),
-                                    np.prod(dtrace),
-                                    np.prod(dkeep),
-                                    np.prod(dkeep)]))
+                          .reshape([np.prod(dtrace, dtype=np.int32),
+                                    np.prod(dtrace, dtype=np.int32),
+                                    np.prod(dkeep, dtype=np.int32),
+                                    np.prod(dkeep, dtype=np.int32)]))
     return Qobj(rhomat, dims=[dkeep, dkeep])
 
 
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -2204,6 +2204,9 @@ def _ptrace_dense(Q, sel):
     dkeep = (rd[sel]).tolist()
     qtrace = list(set(np.arange(nd)) - set(sel))
     dtrace = (rd[qtrace]).tolist()
+    if not dtrace:
+        # If we are keeping all dimensions, no need to construct an ndarray.
+        return Q.copy()
     rd = list(rd)
     if isket(Q):
         vmat = (Q.full()
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -1329,6 +1329,8 @@ class Qobj(object):
         ----------
         sel : int/list
             An ``int`` or ``list`` of components to keep after partial trace.
+            The order is unimportant; no transposition will be done and the
+            spaces will remain in the same order in the output.
 
         Returns
         -------
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -2203,9 +2203,14 @@ def _ptrace_dense(Q, sel):
     else:
         sel = np.asarray(sel)
     sel = list(np.sort(sel))
+    for x in sel:
+        if not 0 <= x < len(rd):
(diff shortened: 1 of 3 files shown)


qutip/qobj.py
```diff
@@ -2216,10 +2219,10 @@ def _ptrace_dense(Q, sel):
                           .reshape(rd + rd)
                           .transpose(qtrace + [nd + q for q in qtrace] +
                                      sel + [nd + q for q in sel])
-                          .reshape([np.prod(dtrace),
-                                    np.prod(dtrace),
-                                    np.prod(dkeep),
-                                    np.prod(dkeep)]))
+                          .reshape([np.prod(dtrace, dtype=np.int32),
```
Ericgig, line 2222: Thank you for the quite patch.
Could you please add the `dtype` to the ket case also (line 2215).

jakelishman: Oh yeah, good point.  Fixed that, and force-updated the PR.

Ericgig:
The pass through made me remember, did we ever fixed #1325?

jakelishman:
No, I guess we never did, at least not on `master`.  I've updated this PR to do both.

PR #1419, opened 2021-01-20
Fast qdiags

jakelishman:
Adds custom `csr.diags` and swaps `qdiags` over to use it to avoid using the `scipy` version with its large overhead.  Constructing a tridiagonal matrix of dimension 100 gets approximately a 7x speed up (320µs to 45µs on my machine) and it's better for small dimensions - the `scipy` overhead has a constant component something around 280µs compared to a constant ~20µs in the new `csr` one.  Not a big deal since it's hardly ever going to be a bottleneck, but it's just part of my work speeding up general QuTiP operations.

### DIFF
--- a/qutip/tests/core/data/test_csr.py
+++ b/qutip/tests/core/data/test_csr.py
@@ -269,3 +269,40 @@ class TestFactoryMethods:
         assert base.shape == (dimension, dimension)
         assert sci.nnz == dimension
         assert (sci - scipy_test).nnz == 0
+
+    @pytest.mark.parametrize(['diagonals', 'offsets', 'shape'], [
+        pytest.param([2j, 3, 5, 9], None, None, id='main diagonal'),
+        pytest.param([1], None, None, id='1x1'),
+        pytest.param([[0.2j, 0.3]], None, None, id='main diagonal list'),
+        pytest.param([0.2j, 0.3], 2, None, id='superdiagonal'),
+        pytest.param([0.2j, 0.3], -2, None, id='subdiagonal'),
+        pytest.param([[0.2, 0.3, 0.4], [0.1, 0.9]], [-2, 3], None,
+                     id='two diagonals'),
+        pytest.param([1, 2, 3], 0, (3, 5), id='main wide'),
+        pytest.param([1, 2, 3], 0, (5, 3), id='main tall'),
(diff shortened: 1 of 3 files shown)


qutip/tests/core/data/test_csr.py
```diff
@@ -269,3 +269,40 @@ def test_identity(self, dimension, scale):
...
+        pytest.param([[1, 2, 3, 4], [4, 5, 4j, 1j]], [-1, -2], (8, 4),
+                     id='two tall sub'),
+        pytest.param([[1, 2, 3], [4, 5, 6], [1, 2]], [1, -1, -2], (4, 4),
+                     id='out of order'),
+        pytest.param([[1, 2, 3], [4, 5, 6], [1, 2]], [1, 1, -2], (4, 4),
+                     id='sum duplicates'),
+    ])
+    def test_diags(self, diagonals, offsets, shape):
+        base = csr.diags(diagonals, offsets, shape)
+        # Build numpy version test.
```
Ericgig, line 296: Why not use `scipy.sparse.diags` to compare? 
There are a few differences with scipy, the most apparent being that you sum duplicate offsets instead of raising an error. Since you added that feature, do you plan to use it somewhere?

jakelishman: I'd written it expecting to write the same sort of thing for `Dense` in the future, so comparing to `numpy` seemed simpler - mostly I actually just didn't think to use `scipy.sparse.diags` as a comparison, since in the rest of the data layer tests I always tested against numpy arrays.  Probably it's better to keep it like this - it's easy to see that the test behaviour is identical to the spec, whereas `scipy.sparse.diags` differs a little (diagonals can be longer than necessary), and I don't know how they handle explicit zeros.

On the second part, I actually just thought `scipy` summed duplicates...  I didn't have anything in mind, I thought I was making a drop-in replacement on that bit.  Whoops.  Should have checked.  There's no speed cost if you don't use the summation behaviour, so it _could_ stay unless there's a definitive negative to it?  I certainly can't think of a use-case for it and if there really isn't one, then perhaps it's better to remove it to catch typos?

Ericgig: I don't see any negative, I would be happy merge as is.
Have you started the dense version? Should I wait for it to merge?

jakelishman: I haven't started it - perhaps the best strategy is to merge this and then use your `Dense` version from #1420?

PR #1420, opened 2021-01-20
Add type choice in Qobj creation function.

Ericgig:
**Description**
Qobj support Dense representation, but there are still no easy way to create dense matrix. So this add a keyword only `dtype` options to most function of qutip.states, qutip.operators and qutip.random_object.

To make them, I made 2 new dispatched function: `diag` and `one_element`. `diag` has the same use than @jakelishman #1419 function, but I fall back on scipy quickly... So I will switch to his version as soon as it is merged. 

To make the `dtype` more user friendly than seeking the type in `qutip.core.data`, I changed `to` to accept the string representation of the name (case insensitive). Only `to` accept string, `dispatch` is not really user facing so there is no reason to touch it.

I also updated the `create` to use `add_creators` and make `Dense` out of `numpy array`. Jake, I am not sure of what were your plans for this. So if I am messing your plans, please tell me. 

Tests are not all created yet and it will create a conflict with #1419 and #1407.

**Related issues or PRs**
#1419

### DIFF
--- a/qutip/core/data/__init__.py
+++ b/qutip/core/data/__init__.py
@@ -14,6 +14,7 @@ from .expm import *
 from .inner import *
 from .kron import *
 from .matmul import *
+from .make import *
 from .mul import *
 from .pow import *
 from .project import *
--- a/qutip/core/data/__init__.py
+++ b/qutip/core/data/__init__.py
@@ -34,5 +34,20 @@ to.add_conversions([
     (Dense, CSR, dense.from_csr, 1),
     (CSR, Dense, csr.from_dense, 1.4),
 ])
+to.register_aliases(['csr', 'CSR'], CSR)
+to.register_aliases(['Dense', 'dense'], Dense)
+
+
+from . import _creator_utils
+import numpy as np
+create.add_creators([
+    (_creator_utils.is_data, _creator_utils.data_copy, 100),
+    (_creator_utils.isspmatrix_csr, CSR, 80),
+    (_creator_utils.is_nparray, Dense, 80),
+    (_creator_utils.issparse, CSR, 20),
+    (_creator_utils.true, Dense, -np.inf),
+])
+del _creator_utils
+del np
 
 from .dispatch import Dispatcher
--- a/qutip/core/data/_creator_utils.py
+++ b/qutip/core/data/_creator_utils.py
@@ -1,0 +1,27 @@
+"""
+Define functions to use as Data creator for `create` in `convert.py`.
+"""
+
+from scipy.sparse import isspmatrix_csr, issparse
+import numpy as np
+from .csr import CSR
+from .base import Data
+from .dense import Dense
+
+__all__ = ['is_data', 'is_nparray', 'data_copy', 'isspmatrix_csr', 'issparse']
+
+
+def is_data(arg):
+    return isinstance(arg, Data)
+
+
+def is_nparray(arg):
+    return isinstance(arg, np.ndarray)
+
+
+def true(arg):
+    return True
+
+
+def data_copy(arg, shape, copy=True):
+    return arg.copy() if copy else arg
--- a/qutip/core/data/_creator_utils.py
+++ b/qutip/core/data/_creator_utils.py
@@ -24,4 +24,8 @@ def true(arg):
 
 
 def data_copy(arg, shape, copy=True):
+    if shape is not None and shape != arg.shape:
+        raise ValueError("".join([
+            "shapes do not match: ", str(shape), " and ", str(arg.shape),
+        ]))
     return arg.copy() if copy else arg
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -265,23 +265,41 @@ cdef class _to:
             self._str2type[dtype.__name__.lower()] = dtype
 
     def parse(self, dtype):
-        """Convert case-insensitive string name of the data-layer to the type
-        itself.
+        """
+        Return a data-layer type object given its name or the type itself.
 
         Parameters
         ----------
         dtype : type, str
-            string to read or type itself.
+            Either the (case-insensitive) name of a data-layer type or a type
+            itself.
+
+        Returns
+        -------
+        type
+            A data-layer type.
+
+        Raises
+        ------
+        TypeError
+            If ``dtype`` is neither a string nor a type.
+        ValueError
+            If ``dtype`` is a name, but no data-layer type of that name is
+            registered, or if ``dtype`` is a type, but not a known data-layer
+            type.
         """
         if type(dtype) is type:
             if dtype not in self.dtypes:
-                raise TypeError("Unsupported data type: " + dtype)
+                raise ValueError(
+                    "Type is not a data-layer type: " + repr(dtype))
             return dtype
         elif type(dtype) is str:
             if dtype.lower() not in self._str2type:
-                raise TypeError("Type name is not known: " + dtype)
+                raise ValueError(
+                    "Type name is not known to the data-layer: " + repr(dtype))
             return self._str2type[dtype.lower()]
-        raise TypeError("Invalid data-layer type: " + dtype)
+        raise TypeError(
+            "Invalid dtype is neither a type nor a type name: " + repr(dtype))
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -283,7 +283,7 @@ cdef class _to:
             aliases = [aliases]
         for alias in aliases:
             if type(alias) is not str:
-                raise TypeError("The alias must be a str")
+                raise TypeError("The alias must be a str : " + repr(alias))
             self._str2type[alias] = layer_type
 
     def parse(self, dtype):
@@ -374,11 +374,12 @@ cdef class _create:
                 Function determining if the given object can be converted to a
                 data-layer type using this creator.
 
-            creator function: callable (object, shape) -> Data
-                The creator function. It should take an object and a shape
-                and return a data-layer type instance. The object may be
-                any object for which the condition function returned ``True``
-                when tested.
+            creator function: callable (object, shape, copy=True) -> Data
+                The creator function. It should take an object and a shape and
+                return a data-layer type instance. The object may be any object
+                for which the condition function returned ``True`` when tested.
+                The ``object`` and ``shape`` parameters are passed positionally,
+                and the ``copy`` parameter is passed by keyword.
 
             priority : real, optional (0)
                 The priority associated with this creator. Higher priority
@@ -402,9 +403,12 @@ cdef class _create:
         """
         for condition, creator, *priority in creators:
             if not callable(condition):
-                raise TypeError(str(condition) + " is not a callable")
+                raise TypeError(repr(condition) + " is not a callable")
             if not callable(create):
-                raise TypeError(str(create) + " is not a callable")
+                raise TypeError(repr(create) + " is not a callable")
+            if len(priority) >= 2:
+                raise ValueError("Too many values to unpack for a creator, " +
+                                 "expected 2 or 3, got "+ str(2+len(priority)))
             priority = float(priority[0] if priority else 0)
             self._creators.append((condition, creator, priority))
         self._creators.sort(key=lambda creator: creator[2], reverse=True)
@@ -424,8 +428,9 @@ cdef class _create:
         """
         for condition, create, _ in self._creators:
             if condition(arg):
-                return create(arg, shape, copy)
-        raise TypeError("arg cannot be converted to qutip data format")
+                return create(arg, shape, copy=copy)
+        raise TypeError(f"arg `{repr(arg)}` cannot be converted to "
+                        "qutip data format")
 
 
 to = _to()
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -334,34 +334,47 @@ cdef class _create:
     def add_creators(self, creators):
         """
         Add creation functions to make a data-layer object from an arbitrary
-        python object.
+        Python object.
 
         Parameters
         ----------
         creators : iterable of (condition, creator, [priority])
-            An iterable of 2- or 3-tuples describing all the new data layer
-            creation function.
+            An iterable of 2- or 3-tuples describing the new data layer
+            creation functions.
             Each element can individually be a 2- or 3-tuple; they do not need
             to be all one or the other.
 
             Elements
             ........
             condition : callable (object) -> Data
-                function determining if that object can be made to a data layer
-                using this creator.
+                function determining if the given object can be converted to a
+                data-layer type using this creator.
 
             creator function: callable (object, shape) -> Data
-                The creator function.  This should take an object and a shape
-                and output a data-layer object. The object can be anything that
-                The condition function returned `True` when tested.
+                The creator function. It should take an object and a shape
+                and return a data-layer type instance. The object may be
+                any object for which the condition function returned ``True``
+                when tested.
 
             priority : positive real, optional (1)
-                The priority associated with this creation. Higher priority
+                The priority associated with this creator. Higher priority
                 conditions will be tested first and the first valid creator
-                (condition(object) == True) will handle the creation.
-                Priority 100 is used for already data-layer entry.
-                80 for object that have a direct data-layer equivalent
-                (scipy.sparse.csr_matrix, numpy.ndarray)
+                (i.e. for which ``condition(object) == True``) will handle
+                the creation.
+
+        Notes
+        -----
+            Default creators are added with the following priorities:
+
+                * Objects that are instances of data-layer types are
+                  converted using ``.copy`` with priority ``100``.
+                * Objects that are ``numpy.ndarray`` instances or for
+                  which ``scipy.sparse.isspmatrix_csr`` is ``True`` are
+                  converted using an internal CSR converter with
+                  priority ``80``.
+                * Objects for which ``scipy.sparse.issparse`` is ``True``
+                  are converted using an internal CSR converter with
+                  priority ``10``.
         """
         _creators = []
         for creator in creators:
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -147,6 +147,7 @@ cdef class _to:
     cdef dict _direct_convert
     cdef dict _convert
     cdef readonly dict weight
+    cdef readonly dict _str2type
 
     def __init__(self):
         self._direct_convert = {}
@@ -154,6 +155,7 @@ cdef class _to:
         self.dtypes = set()
         self.weight = {}
         self.dispatchers = []
+        self._str2type = {}
 
     def add_conversions(self, converters):
         """
@@ -259,32 +261,46 @@ cdef class _to:
                     _converter(convert[::-1], to_t, from_t)
         for dispatcher in self.dispatchers:
             dispatcher.rebuild_lookup()
+        for dtype in self.dtypes:
+            self._str2type[dtype.__name__.lower()] = dtype
+
+    def parse(self, dtype):
+        """Convert case-insensitive string name of the data-layer to the type
+        itself.
+
+        Parameters
+        ----------
+        dtype : type, str
+            string to read or type itself.
+        """
+        if type(dtype) is type:
+            if dtype not in self.dtypes:
+                raise TypeError("Unsupported data type: " + dtype)
+            return dtype
+        elif type(dtype) is str:
+            if dtype.lower() not in self._str2type:
+                raise TypeError("Type name is not known: " + dtype)
+            return self._str2type[dtype.lower()]
+        raise TypeError("Invalid data-layer type: " + dtype)
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
     def __getitem__(self, arg):
-        if isinstance(arg, type):
+        if type(arg) is not tuple:
             arg = (arg,)
-        if not isinstance(arg, tuple) or not arg or len(arg) > 2:
+        if not arg or len(arg) > 2:
             raise KeyError(arg)
-        to_t = arg[0]
-        if to_t not in self.dtypes:
-            raise TypeError("to_type is not known: " + str(to_t))
+        to_t = self.parse(arg[0])
         if len(arg) == 1:
             converters = {
                 from_t: self._convert[to_t, from_t] for from_t in self.dtypes
             }
             return _partial_converter(converters, to_t)
-        from_t = arg[1]
-        if from_t not in self.dtypes:
-            raise TypeError("from_type is not known: " + str(from_t))
+        from_t = self.parse(arg[1])
         return self._convert[to_t, from_t]
 
     def __call__(self, to_type, data):
-        if not isinstance(to_type, type):
-            raise TypeError(repr(to_type) + " is not a type object")
-        if to_type not in self.dtypes:
-            raise ValueError("unknown output type: " + to_type.__name__)
+        to_type = self.parse(to_type)
         from_type = type(data)
         if from_type not in self.dtypes:
             raise TypeError("unknown input type: " + from_type.__name__)
@@ -294,28 +310,93 @@ cdef class _to:
 
 
 cdef class _create:
+    cdef readonly list _creators
+
     def __init__(self):
-        pass
+        self._creators = []
 
     def add_creators(self, creators):
-        pass
+        """
+        Add creation functions to make a data-layer object from an arbitrary
+        python object.
+
+        Parameters
+        ----------
+        creators : iterable of (condition, creator, [priority])
+            An iterable of 2- or 3-tuples describing all the new data layer
+            creation function.
+            Each element can individually be a 2- or 3-tuple; they do not need
+            to be all one or the other.
+
+            Elements
+            ........
+            condition : callable (object) -> Data
+                function determining if that object can be made to a data layer
+                using this creator.
+
+            creator function: callable (object, shape) -> Data
+                The creator function.  This should take an object and a shape
+                and output a data-layer object. The object can be anything that
+                The condition function returned `True` when tested.
+
+            priority : positive real, optional (1)
+                The priority associated with this creation. Higher priority
+                conditions will be tested first and the first valid creator
+                (condition(object) == True) will handle the creation.
+                Priority 100 is used for already data-layer entry.
+                80 for object that have a direct data-layer equivalent
+                (scipy.sparse.csr_matrix, numpy.ndarray)
+        """
+        _creators = []
+        for creator in creators:
+            if len(creator) == 2:
+                condition, create = creator
+                priority = 10
+            elif len(creator) == 3:
+                condition, create, priority = creator
+            if not callable(condition):
+                raise TypeError(str(condition) + " is not a callable")
+            if not callable(create):
+                raise TypeError(str(create) + " is not a callable")
+            _creators.append((condition, create, priority))
+        self._creators = sorted(_creators,
+                                key=lambda creator: creator[2],
+                                reverse=True)
 
     def __call__(self, arg, shape=None):
-        from qutip.core.data import CSR, csr, dense
-        import numpy as np
-        import scipy.sparse
-        if isinstance(arg, CSR):
-            return arg.copy()
-        if scipy.sparse.issparse(arg):
-            return CSR(arg.tocsr(), shape=shape)
-        # Promote 1D lists and arguments to kets, not bras by default.
-        arr = np.array(arg, dtype=np.complex128)
-        if arr.ndim == 1:
-            arr = arr[:, np.newaxis]
-        if arr.ndim != 2:
-            raise TypeError("input has incorrect dimensions: " + str(arr.shape))
-        return csr.from_dense(dense.fast_from_numpy(arr))
-
+        for condition, create, _ in self._creators:
+            if condition(arg):
+                return create(arg, shape)
+        # Fallback on numpy if we don't have any matching creator method
+        return _data.Dense(np.array(arg, dtype=np.complex128),
+                           shape=shape, copy=False)
 
 to = _to()
 create = _create()
+
+import qutip.core.data as _data
+import numpy as np
+import scipy.sparse
+
+create.add_creators([
+    (
+        lambda arg: isinstance(arg, _data.base.Data),
+        lambda arg, shape: arg.copy(),
+        100
+    ),
+    (
+        lambda arg: scipy.sparse.isspmatrix_csr(arg),
+        lambda arg, shape: _data.CSR(arg, shape=shape),
+        80
+    ),
+    (
+        lambda arg: isinstance(arg, np.ndarray),
+        lambda arg, shape: _data.Dense(arg, shape=shape, copy=False),
+        80
+    ),
+    (
+        lambda arg: scipy.sparse.issparse(arg),
+        lambda arg, shape: _data.CSR(arg.tocsr(), shape=shape),
+        10
+    ),
+])
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -319,9 +319,7 @@ cdef class _to:
 
     def __call__(self, to_type, data):
         to_type = self.parse(to_type)
-        from_type = type(data)
-        if from_type not in self.dtypes:
-            raise TypeError("unknown input type: " + from_type.__name__)
+        from_type = self.parse(type(data))
         if to_type == from_type:
             return data
         return self._convert[to_type, from_type](data)
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -261,8 +261,30 @@ cdef class _to:
                     _converter(convert[::-1], to_t, from_t)
         for dispatcher in self.dispatchers:
             dispatcher.rebuild_lookup()
-        for dtype in self.dtypes:
-            self._str2type[dtype.__name__.lower()] = dtype
+
+    def register_aliases(self, aliases, layer_type):
+        """
+        Register a user frendly name for a data-layer type to be recognized by
+        the :method:`parse` method.
+
+        Parameters
+        ----------
+        aliases : str or list of str
+            Name of list of names to be understood to represent the layer_type.
+
+        layer_type : type
+            Data-layer type, must have been registered with
+            :method:`add_conversions` first.
+        """
+        if layer_type not in self.dtypes:
+            raise ValueError(
+                "Type is not a data-layer type: " + repr(layer_type))
+        if isinstance(aliases, str):
+            aliases = [aliases]
+        for alias in aliases:
+            if type(alias) is not str:
+                raise TypeError("The alias must be a str")
+            self._str2type[alias] = layer_type
 
     def parse(self, dtype):
         """
@@ -271,8 +293,7 @@ cdef class _to:
         Parameters
         ----------
         dtype : type, str
-            Either the (case-insensitive) name of a data-layer type or a type
-            itself.
+            Either the name of a data-layer type or a type itself.
 
         Returns
         -------
@@ -294,10 +315,13 @@ cdef class _to:
                     "Type is not a data-layer type: " + repr(dtype))
             return dtype
         elif type(dtype) is str:
-            if dtype.lower() not in self._str2type:
+            try:
+                return self._str2type[dtype]
+            except KeyError:
                 raise ValueError(
-                    "Type name is not known to the data-layer: " + repr(dtype))
-            return self._str2type[dtype.lower()]
+                    "Type name is not known to the data-layer: " + repr(dtype)
+                    ) from None
+
         raise TypeError(
             "Invalid dtype is neither a type nor a type name: " + repr(dtype))
 
@@ -346,8 +370,8 @@ cdef class _create:
 
             Elements
             ........
-            condition : callable (object) -> Data
-                function determining if the given object can be converted to a
+            condition : callable (object) -> bool
+                Function determining if the given object can be converted to a
                 data-layer type using this creator.
 
             creator function: callable (object, shape) -> Data
@@ -356,7 +380,7 @@ cdef class _create:
                 any object for which the condition function returned ``True``
                 when tested.
 
-            priority : positive real, optional (1)
+            priority : real, optional (0)
                 The priority associated with this creator. Higher priority
                 conditions will be tested first and the first valid creator
                 (i.e. for which ``condition(object) == True``) will handle
@@ -364,68 +388,45 @@ cdef class _create:
 
         Notes
         -----
-            Default creators are added with the following priorities:
+        Default creators are added with the following priorities:
 
-                * Objects that are instances of data-layer types are
-                  converted using ``.copy`` with priority ``100``.
-                * Objects that are ``numpy.ndarray`` instances or for
-                  which ``scipy.sparse.isspmatrix_csr`` is ``True`` are
-                  converted using an internal CSR converter with
-                  priority ``80``.
-                * Objects for which ``scipy.sparse.issparse`` is ``True``
-                  are converted using an internal CSR converter with
-                  priority ``10``.
+        * Objects that are instances of data-layer types are
+          converted using ``.copy`` with priority 100.
+        * Objects that have a direct equivalent such as ``numpy.ndarray``
+          or ``scipy.sparse.csr_matrix`` are converted with priority 80.
+        * Objects for which ``scipy.sparse.issparse`` is ``True``
+          are converted using an internal CSR converter with
+          priority 20.
+        * If no condition are meet, ``numpy.array`` is used to try convert
+          the input to an array (priority -inf).
         """
-        _creators = []
-        for creator in creators:
-            if len(creator) == 2:
-                condition, create = creator
-                priority = 10
-            elif len(creator) == 3:
-                condition, create, priority = creator
+        for condition, creator, *priority in creators:
             if not callable(condition):
                 raise TypeError(str(condition) + " is not a callable")
             if not callable(create):
                 raise TypeError(str(create) + " is not a callable")
-            _creators.append((condition, create, priority))
-        self._creators = sorted(_creators,
-                                key=lambda creator: creator[2],
-                                reverse=True)
+            priority = float(priority[0] if priority else 0)
+            self._creators.append((condition, creator, priority))
+        self._creators.sort(key=lambda creator: creator[2], reverse=True)
 
-    def __call__(self, arg, shape=None):
+    def __call__(self, arg, shape=None, copy=True):
+        """
+        Build a :class:`qutip.data.Data` object from arg.
+
+        Parameters
+        ----------
+        arg : object
+            Object to be converted to `qutip.data.Data`. Anything that can be
+            converted to a numpy array are valid input.
+
+        shape : tuple, optional
+            The shape of the output as (``rows``, ``columns``).
+        """
         for condition, create, _ in self._creators:
             if condition(arg):
-                return create(arg, shape)
-        # Fallback on numpy if we don't have any matching creator method
-        return _data.Dense(np.array(arg, dtype=np.complex128),
-                           shape=shape, copy=False)
+                return create(arg, shape, copy)
+        raise TypeError("arg cannot be converted to qutip data format")
+
 
 to = _to()
 create = _create()
-
-import qutip.core.data as _data
-import numpy as np
-import scipy.sparse
-
-create.add_creators([
-    (
-        lambda arg: isinstance(arg, _data.base.Data),
-        lambda arg, shape: arg.copy(),
-        100
-    ),
-    (
-        lambda arg: scipy.sparse.isspmatrix_csr(arg),
-        lambda arg, shape: _data.CSR(arg, shape=shape),
-        80
-    ),
-    (
-        lambda arg: isinstance(arg, np.ndarray),
-        lambda arg, shape: _data.Dense(arg, shape=shape, copy=False),
-        80
-    ),
-    (
-        lambda arg: scipy.sparse.issparse(arg),
-        lambda arg, shape: _data.CSR(arg.tocsr(), shape=shape),
-        10
-    ),
-])
--- a/qutip/core/data/dense.pyx
+++ b/qutip/core/data/dense.pyx
@@ -44,8 +44,8 @@ class OrderEfficiencyWarning(EfficiencyWarning):
 cdef class Dense(base.Data):
     def __init__(self, data, shape=None, copy=True):
         base = np.array(data, dtype=np.complex128, order='K', copy=copy)
-        # Ensure that the array is continuous.
-        # Non continuous array with copy=False would otherwise slip through
+        # Ensure that the array is contiguous.
+        # Non contiguous array with copy=False would otherwise slip through
         if not (cnp.PyArray_IS_C_CONTIGUOUS(base) or
                 cnp.PyArray_IS_F_CONTIGUOUS(base)):
             base = base.copy()
--- a/qutip/core/data/dense.pyx
+++ b/qutip/core/data/dense.pyx
@@ -44,8 +44,15 @@ class OrderEfficiencyWarning(EfficiencyWarning):
 cdef class Dense(base.Data):
     def __init__(self, data, shape=None, copy=True):
         base = np.array(data, dtype=np.complex128, order='K', copy=copy)
+        # Ensure that the array is continuous.
+        # Non continuous array with copy=False would otherwise slip through
+        if not (cnp.PyArray_IS_C_CONTIGUOUS(base) or
+                cnp.PyArray_IS_F_CONTIGUOUS(base)):
+            base = base.copy()
         if shape is None:
             shape = base.shape
+            if len(shape) == 0:
+                shape = (1, 1)
             # Promote to a ket by default if passed 1D data.
             if len(shape) == 1:
                 shape = (shape[0], 1)
@@ -345,4 +352,110 @@ cpdef Dense from_csr(CSR matrix, bint fortran=False):
         for ptr_in in range(matrix.row_index[row], matrix.row_index[row + 1]):
             out.data[ptr_out + matrix.col_index[ptr_in]*col_stride] = matrix.data[ptr_in]
         ptr_out += row_stride
+    return out
+
+
+cdef inline base.idxint _diagonal_length(
+    base.idxint offset, base.idxint n_rows, base.idxint n_cols,
+) nogil:
+    if offset > 0:
+        return n_rows if offset <= n_cols - n_rows else n_cols - offset
+    return n_cols if offset > n_cols - n_rows else n_rows + offset
+
+
+@cython.wraparound(True)
+def diags(diagonals, offsets=None, shape=None):
+    """
+    Construct a matrix from diagonals and their offsets.  Using this
+    function in single-argument form produces a square matrix with the given
+    values on the main diagonal.
+    With lists of diagonals and offsets, the matrix will be the smallest
+    possible square matrix if shape is not given, but in all cases the
+    diagonals must fit exactly with no extra or missing elements. Duplicated
+    diagonals will be summed together in the output.
+
+    Parameters
+    ----------
+    diagonals : sequence of array_like of complex or array_like of complex
+        The entries (including zeros) that should be placed on the diagonals in
+        the output matrix.  Each entry must have enough entries in it to fill
+        the relevant diagonal and no more.
+    offsets : sequence of integer or integer, optional
+        The indices of the diagonals.  `offsets[i]` is the location of the
+        values `diagonals[i]`.  An offset of 0 is the main diagonal, positive
+        values are above the main diagonal and negative ones are below the main
+        diagonal.
+    shape : tuple, optional
+        The shape of the output as (``rows``, ``columns``).  The result does
+        not need to be square, but the diagonals must be of the correct length
+        to fit in exactly.
+    """
+    cdef base.idxint n_rows, n_cols, offset
+    try:
+        diagonals = list(diagonals)
+        if diagonals and np.isscalar(diagonals[0]):
+            # Catch the case where we're being called as (for example)
+            #   diags([1, 2, 3], 0)
+            # with a single diagonal and offset.
+            diagonals = [diagonals]
+    except TypeError:
+        raise TypeError("diagonals must be a list of arrays of complex") from None
+    if offsets is None:
+        if len(diagonals) == 0:
+            offsets = []
+        elif len(diagonals) == 1:
+            offsets = [0]
+        else:
+            raise TypeError("offsets must be supplied if passing more than one diagonal")
+    offsets = np.atleast_1d(offsets)
+    if offsets.ndim > 1:
+        raise ValueError("offsets must be a 1D array of integers")
+    if len(diagonals) != len(offsets):
+        raise ValueError("number of diagonals does not match number of offsets")
+    if len(diagonals) == 0:
+        if shape is None:
+            raise ValueError("cannot construct matrix with no diagonals without a shape")
+        else:
+            n_rows, n_cols = shape
+        return zeros(n_rows, n_cols)
+    order = np.argsort(offsets)
+    diagonals_ = []
+    offsets_ = []
+    prev, cur = None, None
+    for i in order:
+        cur = offsets[i]
+        if cur == prev:
+            diagonals_[-1] += np.asarray(diagonals[i], dtype=np.complex128)
+        else:
+            offsets_.append(cur)
+            diagonals_.append(np.asarray(diagonals[i], dtype=np.complex128))
+        prev = cur
+    if shape is None:
+        n_rows = n_cols = abs(offsets_[0]) + len(diagonals_[0])
+    else:
+        try:
+            n_rows, n_cols = shape
+        except (TypeError, ValueError):
+            raise TypeError("shape must be a 2-tuple of positive integers")
+        if n_rows < 0 or n_cols < 0:
+            raise ValueError("shape must be a 2-tuple of positive integers")
+    for i in range(len(diagonals_)):
+        offset = offsets_[i]
+        if len(diagonals_[i]) != _diagonal_length(offset, n_rows, n_cols):
+            raise ValueError("given diagonals do not have the correct lengths")
+    if n_rows == 0 and n_cols == 0:
+        raise ValueError("can't produce a 0x0 matrix")
+
+    out = zeros(n_rows, n_cols, fortran=True)
+
+    cdef size_t diag_idx, idx, n_diagonals = len(diagonals_)
+
+    for diag_idx in range(n_diagonals):
+        offset = offsets_[diag_idx]
+        if offset <= 0:
+            for idx in range(_diagonal_length(offset, n_rows, n_cols)):
+                out.data[idx*(n_rows+1) - offset] = diagonals_[diag_idx][idx]
+        else:
+            for idx in range(_diagonal_length(offset, n_rows, n_cols)):
+                out.data[idx*(n_rows+1) + offset*n_rows] = diagonals_[diag_idx][idx]
     return out
--- a/qutip/core/data/dispatch.pyx
+++ b/qutip/core/data/dispatch.pyx
@@ -552,8 +552,9 @@ cdef class Dispatcher:
         callable object which requires that the dispatched arguments match
         those specified in `types`.
         """
-        if isinstance(types, type):
+        if type(types) is not tuple:
             types = (types,)
+        types = tuple(_to.parse(arg) for arg in types)
         try:
             return self._lookup[types]
         except KeyError:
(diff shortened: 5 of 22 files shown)


qutip/core/data/convert.pyx
```diff
@@ -259,10 +261,31 @@ cdef class _to:
...
         for dispatcher in self.dispatchers:
             dispatcher.rebuild_lookup()
+        for dtype in self.dtypes:
+            self._str2type[dtype.__name__.lower()] = dtype
+
+    def str2type(self, arg):
+        if isinstance(arg, type):
+            return arg
+        if isinstance(arg, str):
+            if not arg.lower() in self._str2type:
```
jakelishman, line 271: `x not in y` is supposedly better/faster than `not x in y` according to the Python docs, but I don't really know any detail on that

qutip/core/data/convert.pyx
```diff
@@ -294,28 +318,98 @@ cdef class _to:
 
 
 cdef class _create:
+    cdef readonly list _creators
+
```
jakelishman, line 322: Yeah, this is something pretty much exactly like what I had in mind.  Looks good to me for now - probably we'll revise it slightly once we've nailed down how the `Dispatcher` options will look like but I'd much rather have this in place and build from it than keep saying "tomorrow and tomorrow and tomorrow".

qutip/core/data/dense.pyx
```diff
@@ -44,8 +44,15 @@ class OrderEfficiencyWarning(EfficiencyWarning):
 cdef class Dense(base.Data):
     def __init__(self, data, shape=None, copy=True):
         base = np.array(data, dtype=np.complex128, order='K', copy=copy)
+        # Ensure that the array is continuous.
+        # Non continuous array with copy=False would otherwise slip through
+        if not (cnp.PyArray_IS_C_CONTIGUOUS(base) or
+                cnp.PyArray_IS_F_CONTIGUOUS(base)):
+            base = base.copy()
```
jakelishman, line 51: Nice catch, thanks.

qutip/core/operators.py
```diff
@@ -64,6 +65,10 @@ def jmat(j, which=None):
         Which operator to return 'x','y','z','+','-'.
         If no args given, then output is ['x','y','z']
 
+    dtype : qutip.data.Data or str
```
jakelishman, line 68: I suppose the type is `type` not `qutip.data.Data`

qutip/core/operators.py
```diff
@@ -50,9 +50,10 @@
 from . import data as _data
 from .qobj import Qobj
 from .dimensions import flatten
+from .data import to
 
 
-def jmat(j, which=None):
+def jmat(j, which=None, *, dtype=None):
```
jakelishman, line 56: Good choice to use `*`.

qutip/core/operators.py
```diff
@@ -102,166 +107,198 @@ def jmat(j, which=None):
...
-        ptr = np.concatenate((np.arange(j + 1, dtype=np.int32),
-                              np.arange(j, N, dtype=np.int32)))
-    return _data.csr.CSR((data, ind, ptr), shape=(N, N))
+    # else:
+    #     j = int(j)
+    #     ind = np.concatenate((np.arange(j, dtype=np.int32),
+    #                           np.arange(j + 1, N, dtype=np.int32)))
+    #     ptr = np.concatenate((np.arange(j + 1, dtype=np.int32),
+    #                           np.arange(j, N, dtype=np.int32)))
+    # return _data.csr.CSR((data, ind, ptr), shape=(N, N))
```
jakelishman, line 172: can delete the commented out code - source control saves it anyway

qutip/core/operators.py
```diff
@@ -64,6 +65,10 @@ def jmat(j, which=None):
         Which operator to return 'x','y','z','+','-'.
         If no args given, then output is ['x','y','z']
 
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 70: The idea of the default here is one thing that we'll probably revisit (I envisage being able to define a set of Dispatcher rules on a per-function basis in the exact same way you do your global settings), but I'm entirely happy to have it work like this to get the ball rolling.

qutip/core/operators.py
```diff
@@ -369,6 +406,10 @@ def destroy(N, offset=0):
         The lowest number state that is included in the finite number state
         representation of the operator.
 
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 411: defaults to `CSR` not `Dense`

qutip/core/operators.py
```diff
@@ -407,6 +453,14 @@ def create(N, offset=0):
     N : int
         Dimension of Hilbert space.
 
+    offset : int (default 0)
+        The lowest number state that is included in the finite number state
+        representation of the operator.
+
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 462: also defaults to `CSR`

qutip/core/operators.py
```diff
@@ -426,6 +480,17 @@ def create(N, offset=0):
...
+    data = np.sqrt(np.arange(offset+1, N+offset, dtype=complex))
+    dtype = to.str2type(dtype or _data.CSR)
+    out = _data.diag[dtype](data, -1)
+    return Qobj(out,
+                dims=[[N], [N]],
+                type='oper',
+                isherm=False,
+                isunitary=False,
+                copy=False)
     return destroy(N, offset=offset).dag()
```
jakelishman, line 494: unreachable code

qutip/core/operators.py
```diff
@@ -470,20 +535,25 @@ def qzero(dimensions):
...
     qzero : qobj
         Zero operator Qobj.
 
     """
+    dtype = to.str2type(dtype or _data.CSR)
     size, dimensions = _implicit_tensor_dimensions(dimensions)
     # A sparse matrix with no data is equal to a zero matrix.
     type_ = 'super' if isinstance(dimensions[0][0], list) else 'oper'
-    return Qobj(_data.csr.zeros(size, size), dims=dimensions, type=type_,
+    return Qobj(_data.zeros[dtype](size, size), dims=dimensions, type=type_,
```
jakelishman, line 552: That's a nice way of doing things!

qutip/core/operators.py
```diff
@@ -495,6 +565,10 @@ def qeye(dimensions):
         the new Qobj are set to this list.  This can produce either `oper` or
         `super` depending on the passed `dimensions`.
 
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 570: wrong default

qutip/core/operators.py
```diff
@@ -542,16 +617,20 @@ def position(N, offset=0):
         The lowest number state that is included in the finite number state
         representation of the operator.
 
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 622: wrong default

qutip/core/operators.py
```diff
@@ -586,6 +669,10 @@ def num(N, offset=0):
...
```
jakelishman, line 674: wrong default

qutip/core/qobjevo.py
```diff
@@ -891,7 +892,7 @@ def mul(self, t, mat):
             return out
 
     def _compile(self, code=False, matched=False, dense=False):
-        self.tidyup()
+        # self.tidyup()
```
jakelishman, line 895: delete?

qutip/core/states.py
```diff
@@ -148,12 +152,15 @@ def basis(dimensions, n=None, offset=None):
...
-    sci.indptr[location+1:] = 1
+
+    dtype = to.str2type(dtype or _data.Dense)
+    data = _data.one_element[dtype]((size, 1), (location, 0), 1)
+    # data = _data.csr.empty(size, 1, 1)
+    # sci = data.as_scipy(full=True)
+    # sci.data[0] = 1
+    # sci.indices[0] = 0
+    # sci.indptr[:location+1] = 0
+    # sci.indptr[location+1:] = 1
```
jakelishman, line 163: delete?

qutip/core/states.py
```diff
@@ -321,6 +345,10 @@ def fock_dm(dimensions, n=None, offset=None):
         The lowest number state that is included in the finite number state
         representation of the state in the relevant dimension.
 
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 350: wrong default

qutip/core/states.py
```diff
@@ -434,27 +470,31 @@ def thermal_dm(N, n, method='operator'):
...
     else:
         i = np.arange(N)
         if method == 'operator':
             beta = np.log(1.0 / n + 1.0)
             diags = np.exp(-beta * i)
             diags = diags / np.sum(diags)
             # populates diagonal terms using truncated operator expression
-            rm = sp.spdiags(diags, 0, N, N, format='csr')
+            # rm = sp.spdiags(diags, 0, N, N, format='csr')
+
```
jakelishman, line 482: delete?

qutip/core/states.py
```diff
@@ -434,27 +470,31 @@ def thermal_dm(N, n, method='operator'):
...
         elif method == 'analytic':
             # populates diagonal terms using analytic values
-            rm = sp.spdiags((1.0 + n) ** (-1.0) * (n / (1.0 + n)) ** (i),
-                            0, N, N, format='csr')
+            # rm = sp.spdiags((1.0 + n) ** (-1.0) * (n / (1.0 + n)) ** (i),
+            #                 0, N, N, format='csr')
```
jakelishman, line 486: delete?

qutip/core/states.py
```diff
@@ -395,6 +427,10 @@ def thermal_dm(N, n, method='operator'):
         ``string`` that sets the method used to generate the
         thermal state probabilities
 
+    dtype : qutip.data.Data or str
+        Storage representation. Any data-layer known to `qutip.data.to` is
+        accepted. Default `qutip.data.Dense` (numpy.ndarray like).
```
jakelishman, line 432: wrong default

qutip/random_objects.py
```diff
@@ -79,21 +86,21 @@ def rand_jacobi_rotation(A, seed=None):
     angle = 2*np.random.random()*np.pi
     a = np.sqrt(0.5) * np.exp(-1j*angle)
     b = np.conj(a)
-    i = np.int(np.floor(np.random.random() * n))
+    i = np.random.randint(n)
```
jakelishman, line 89: At some point we should maybe think about moving to the new (`numpy >= 1.17`) random methods, but I think we've got a few more numpy versions before they completely remove them.  I certainly don't want to do that right now.

jakelishman (commented):
I like this a lot in general.  I really need to find time to write the proper spec for `Dispatcher` rules and overrides and how the user will interact with it, but I think you're exactly right to have a `dtype` argument like this attached to all creation functions.  I like the string touch too - similar to numpy.  Also like numpy, perhaps we ought to promote `Dense` and `CSR` into the main `qutip` namespace, so you can also do `dtype=qutip.Dense`, like numpy's `dtype=np.complex128'?

I think the function `to.str2type` should probably be internal to `to` - maybe have `to.__getitem__` and all other entry points to `to` automatically call it so that we ensure there's a single source of truth for what constitutes a valid `dtype` to any given function.

Really, I'm happy that you're working on it and pushing it forwards while I'm not finding the time to do it.  I'd much rather have your stuff merged in now and ready to go, and _if_ we decide to revise it a little in the future, we've already got this to work from.

Glad to see `test_random` and `test_rand` (or whatever they were called) merged into one file, too.

qutip/core/data/convert.pyx
```diff
@@ -294,28 +318,98 @@ cdef class _to:
...
+                raise TypeError(str(create) + " is not a callable")
+            _creators.append((condition, create, priority))
+        self._creators = sorted(_creators,
+                                key=lambda creator: creator[2],
+                                reverse=True)
+
+    def __call__(self, arg, shape=None):
+        for condition, create, _ in self._creators:
+            if condition(arg):
+                return create(arg, shape)
```
jakelishman, line 379: What should happen if it drops off the end of the `for` loop?

qutip/core/data/convert.pyx
```diff
@@ -294,28 +318,98 @@ cdef class _to:
...
+    ),
+    (
+        lambda arg: scipy.sparse.issparse(arg),
+        lambda arg, shape: _data.CSR(arg.tocsr(), shape=shape),
+        10
+    ),
+    (
+        lambda arg: True,
+        lambda arg, shape: _data.Dense(np.array(arg, dtype=np.complex128),
+                                       shape=shape, copy=False),
```
jakelishman, line 412: This looks like it's asking for trouble.  Might be better to handle the "fall-through" case explicitly in `create`.

qutip/core/data/make.py
```diff
@@ -0,0 +1,108 @@
...
+from .csr import CSR
+from . import csr, dense, CSR, Dense
+from numbers import Number
+
+__all__ = ['diag',
+           'one_element_csr', 'one_element_dense', 'one_element']
+
+
+def diag_signature(diagonals, offsets=None, shape=None):
+    """
```
jakelishman, line 14: I can't remember where or how, but I'm pretty sure somewhere in one of the other files I found a "cleaner" way of doing this without leaking a function name, perhaps using `inspect`?

qutip/core/operators.py
```diff
@@ -50,9 +52,10 @@
 from . import data as _data
 from .qobj import Qobj
 from .dimensions import flatten
+from .data import to
```
jakelishman, line 55: Personally I'd like to keep "QuTiP house style" to be `from . import data as _data`, and then use `_data.to` exclusively to be hyper clear what's being converted, but open to other views.

qutip/core/operators.py
```diff
@@ -102,166 +109,181 @@ def jmat(j, which=None):
...
     raise ValueError('invalid spin operator: ' + which)
 
 
-def _jplus(j):
+def _jplus(j, *, dtype=None):
     """
     Internal functions for generating the data representing the J-plus
     operator.
     """
+    dtype = to.str2type(dtype or _data.CSR)
```
jakelishman, line 144: I'd have `str2type` be private to `data.to`, and just have `__getitem__` call it.  The fewer public functions our objects have, the easier they are to use.

It also means that we can guaranteed that the interface is uniform _everywhere_, and isn't dependent on us writing additional boilerplate within each individual function.

jakelishman: This comment recurs on the other ~20 lines it appears within this file.

qutip/core/operators.py
```diff
@@ -420,8 +461,15 @@ def create(N, offset=0):
...
-    return qdiags([data], [-1])
+    data = np.sqrt(np.arange(offset+1, N+offset, dtype=complex))
+    dtype = to.str2type(dtype or _data.CSR)
+    out = _data.diag[dtype](data, -1)
+    return Qobj(out,
+                dims=[[N], [N]],
+                type='oper',
+                isherm=False,
+                isunitary=False,
+                copy=False)
```
jakelishman, line 472: Why not defer to `qdiags`?

qutip/core/operators.py
```diff
@@ -632,9 +714,10 @@ def squeeze(N, z, offset=0):
      [ 0.00000000+0.j -0.30142443+0.j  0.00000000+0.j  0.95349007+0.j]]
 
     """
+    dtype = to.str2type(dtype or _data.Dense)
     asq = destroy(N, offset=offset) ** 2
     op = 0.5*np.conj(z)*asq - 0.5*z*asq.dag()
-    return op.expm()
+    return op.expm().to(dtype)
```
jakelishman, line 720: In theory the right way to do this is to pass the out `dtype` to `expm` and let the data-layer do its thing in the normal manner, rather than having an explicit conversion step.

qutip/core/states.py
```diff
@@ -51,7 +51,7 @@
 from .qobj import Qobj
 from .operators import jmat, displace
 from .tensor import tensor
-
+from .data import to
```
jakelishman, line 54: Same comment about potential style - having looked through, I think that if `to.str2type` is private, then the import won't even be needed anyway.

qutip/core/states.py
```diff
@@ -436,27 +464,28 @@ def thermal_dm(N, n, method='operator'):
...
             # populates diagonal terms using analytic values
-            rm = sp.spdiags((1.0 + n) ** (-1.0) * (n / (1.0 + n)) ** (i),
-                            0, N, N, format='csr')
+            diags = (1.0 + n) ** (-1.0) * (n / (1.0 + n)) ** (i)
         else:
             raise ValueError(
                 "'method' keyword argument must be 'operator' or 'analytic'"
             )
+        dtype = to.str2type(dtype or _data.CSR)
+        rm = _data.diag[dtype](diags, 0, (N,N))
```
jakelishman, line 484: missing a space `(N, N)`

qutip/core/states.py
```diff
@@ -926,20 +985,22 @@ def enr_fock(dims, excitations, state):
 
     """
     nstates, state2idx, _ = enr_state_dictionaries(dims, excitations)
-    data = sp.lil_matrix((nstates, 1), dtype=np.complex)
+    # data = sp.lil_matrix((nstates, 1), dtype=np.complex)
```
jakelishman, line 988: delete comment

qutip/core/states.py
```diff
@@ -926,20 +985,22 @@ def enr_fock(dims, excitations, state):
...
+    dtype = to.str2type(dtype or _data.CSR)
     try:
-        data[state2idx[tuple(state)], 0] = 1
+        data =_data.one_element[dtype]((nstates, 1),
+                                       (state2idx[tuple(state)], 0), 1)
+        # data[state2idx[tuple(state)], 0] = 1
```
jakelishman, line 993: delete comment

qutip/core/states.py
```diff
@@ -977,23 +1042,32 @@ def enr_thermal_dm(dims, excitations, n):
     diags = [np.prod((n / (n + 1)) ** np.array(state))
              for idx, state in idx2state.items()]
     diags /= np.sum(diags)
-    data = sp.spdiags(diags, 0, nstates, nstates, format='csr')
+    # data = sp.spdiags(diags, 0, nstates, nstates, format='csr')
```
jakelishman, line 1045: delete comment

qutip/core/superop_reps.py
```diff
@@ -232,7 +233,7 @@ def _choi_to_chi(q_oper):
     """
     nq = _nq(q_oper.dims)
     B = _superpauli_basis(nq).data
-    return Qobj(B.adjoint() @ q_oper.data @ B,
+    return Qobj(matmul(matmul(B.adjoint(), q_oper.data), B),
```
jakelishman, line 236: Pretty sure this operation is called `inner_op` or something on the data layer and is optimised.

qutip/core/superop_reps.py
```diff
@@ -250,7 +251,7 @@ def _chi_to_choi(q_oper):
     B = _superpauli_basis(nq).data
     # The Chi matrix has tr(chi) == d², so we need to divide out
     # by that to get back to the Choi form.
-    return Qobj((B @ q_oper.data @ B.adjoint()) / q_oper.shape[0],
+    return Qobj(matmul(matmul(B, q_oper.data), B.adjoint()) / q_oper.shape[0],
```
jakelishman, line 254: `inner_op`

qutip/random_objects.py
```diff
@@ -53,47 +57,50 @@
...
+        j = np.random.randint(n)
     data = np.hstack((np.array([a, -b, a, b], dtype=complex),
                       np.ones(n - 2, dtype=complex)))
     diag = np.delete(np.arange(n), [i, j])
     rows = np.hstack(([i, i, j, j], diag))
     cols = np.hstack(([i, j, i, j], diag))
     R = sp.coo_matrix((data, (rows, cols)), shape=(n, n), dtype=complex)
     R = _data.create(R.tocsr())
-    return R @ A @ R.adjoint()
+    return _data.matmul(_data.matmul(R, A), R.adjoint())
```
jakelishman, line 100: `inner_op`

qutip/random_objects.py
```diff
@@ -368,24 +429,33 @@ def rand_ket_haar(N=2, dims=None, seed=None):
         _check_dims(dims, N, 1)
     else:
         dims = [[N], [1]]
-    U = rand_unitary_haar(N, seed=seed, dims=[dims[0], dims[0]])
-    return U @ basis(dims[0], [0]*len(dims[0]))
+    U = rand_unitary_haar(N, seed=seed, dims=[dims[0], dims[0]], dtype=dtype)
```
jakelishman, line 432: This doesn't need to have the same out dtype, right?

qutip/tests/core/test_operators.py
```diff
@@ -31,112 +31,139 @@
...
-    assert_equal(np.allclose(spinhalf[0].full(), paulix), True)
-    assert_equal(np.allclose(spinhalf[1].full(), pauliy), True)
-    assert_equal(np.allclose(spinhalf[2].full(), pauliz), True)
-    assert_equal(np.allclose(jmat(1 / 2., '+').full(), sigmap), True)
-    assert_equal(np.allclose(jmat(1 / 2., '-').full(), sigmam), True)
+    assert np.allclose(sigmax().full(), paulix*2)
+    assert np.allclose(sigmay().full(), pauliy*2)
+    assert np.allclose(sigmaz().full(), pauliz*2)
+    assert np.allclose(sigmap().full(), sigma_p)
+    assert np.allclose(sigmam().full(), sigma_m)
```
jakelishman, line 82: Should these be `np.testing.assert_allclose`?

qutip/core/operators.py
jakelishman: Maybe we'd keep the approximate functionality, but call it `to.parse` and have it work only on scalar inputs?  Then it _can_ be used externally, but in general I would expect `Dispatcher.__getitem__` to call it internally.

qutip/tests/core/test_states.py
```diff
@@ -60,69 +72,109 @@ def test_implicit_tensor_basis_like(to_test, size, n):
...
+     id="phase_basis_phi0"),
+    pytest.param(spin_state, qutip.spin_Jz, (3, 2), (3,), 2,
+     id="spin_state"),
+    pytest.param(zero_ket, qutip.qeye, (10,), (10,), 0,
+     id="zero_ket"),
+])
+def test_div_basis(base, operator, args, opargs, eigenval, dtype):
+    state = base(*args, dtype=dtype[0])
+    oper = operator(*opargs)
+    assert np.allclose((oper * state).full(), state.full() * eigenval)
```
jakelishman, line 109: `np.testing.assert_allclose`?

qutip/tests/core/test_superoper.py
```diff
@@ -130,7 +130,7 @@ def testMatrixVecMat(self):
         M = _data.create(np.random.rand(10, 10))
         V = qutip.stack_columns(M)
         M2 = qutip.unstack_columns(V)
-        assert _data.csr.nnz(M - M2) == 0
+        assert qutip.Qobj(M - M2).norm('max') < 1e-8
```
jakelishman, line 133: `np.testing.assert_allclose` maybe better for debugging?

qutip/core/superop_reps.py
Ericgig: Doesn't inner_op return a scalar?

qutip/core/operators.py
Ericgig: I preferred not to add any extra step to the `Dispatcher`, keeping the string `dtype` as a user format only, having any user provided `dtype` converted as pure type immediately and keeping all internal with only that format.
(For me anything that works with Data-layer directly is internal.)
This is why, I made `str2type` available to the python layer.
I will change it to `to.parse` and leave the `Dispatcher` untouched for now.

jakelishman: But this is an extra step to Dispatcher - it's just that the complexity of the function is now spread around in many many places, which leaves us open to forgetting to do it at some point.  Putting this test in `Dispatcher` doesn't add any additional runtime, and it makes a single source of truth for "what is a data-layer type".  That's much easier to maintain and extend.

Ericgig: `expm` takes a `dtype` input, under this logic should it be parsed by the `Dispatcher`? 
If yes, I am not sure I will get all place where I should parse in `Dispatcher`...
I will do the change in `Dispatcher.__getitem__`, `Dispatcher.__call__` , but I am pretty sure I am missing some places...

Ericgig:
@jakelishman, I believe I addressed all your comment (expect the fortran option).
But I expect that it will cause issue when `self._pass_on_dtype` is True: `dtype` will be parsed when called as:
`_data.expm(data, dtype)` but not `_data.expm[type(data)](data, dtype)`.
But I don't know where to fix this, `_constructed_specialisation.__call__`, `_bind.bind`? Or parse in `_data.expm_...` and parse `dtype` in `Dispatcher.__call__` only when referring to the output type?

qutip/core/data/dense.pyx
```diff
@@ -44,8 +44,15 @@ class OrderEfficiencyWarning(EfficiencyWarning):
...
         if shape is None:
             shape = base.shape
+            if len(shape) == 0:
+                shape = (1, 1)
```
jakelishman, line 55: So we're making the `Dense` constructor implicitly have numpy's `atleast_2d` behaviour?  Seems like a pretty good choice to me.

qutip/core/data/dense.pyx
jakelishman, line 51: So I looked again at the `numpy.array` documentation, and it looks like this test shouldn't be necessary - instead we should set `order='A'` which is the behaviour we want.

qutip/core/data/dispatch.pyx
```diff
@@ -552,8 +552,9 @@ cdef class Dispatcher:
         callable object which requires that the dispatched arguments match
         those specified in `types`.
         """
-        if isinstance(types, type):
+        if isinstance(types, (type, str)):
             types = (types,)
```
jakelishman, line 556: Similar comment to before about having a single source of truth - this should be just `if not isinstance(types, tuple)` now to allow `to.parse` to be the only place that defines what's acceptable (I think it's fine to assert everywhere that `tuple` will _never_ be an acceptable type).

qutip/core/data/dispatch.pyx
```diff
@@ -552,8 +552,9 @@ cdef class Dispatcher:
...
+        types = tuple(_to.parse(arg) for arg in types)
```
jakelishman, line 557: This doesn't "smell" right to be to be creating a second temporary object and using Python iteration, but let's accept it for now.

I'll look into it more to see if what else we can squeeze out of this function in terms of performance, but right now let's just go for it.

qutip/core/operators.py
```diff
@@ -52,7 +54,56 @@
...
+            - k>0 kth upper diagonal
+            - k<0 kth lower diagonal
+    dims : list, optional
+        Dimensions for operator
+
+    shape : list, tuple, optional
+        Shape of operator.  If omitted, a square operator large enough
+        to contain the diagonals is generated.
+
+    dtype : type or str
```
jakelishman, line 78: In theory we should use an alias like numpy's `array_like` for these so we can just define what's valid in one place, but let's not worry about it right now.

jakelishman (commented):
Looks mostly good to me - a couple of comments about `if` tests that can be removed now only.  I think the code looks much easier to maintain and easier to extend with `to.parse` used as the single source of truth like this.

I'm not sure I understand your point about `_pass_on_dtype`?  `expm` doesn't take a `dtype` argument, so that Boolean is never `True` for it.  If I remember correctly, there's very little that actually uses that option right now - it's there to support the case where a data-layer function includes dispatch of its own, because otherwise `Dispatcher` "eats" the `dtype` kwarg and wouldn't pass it on to an internal function that wants it.

jakelishman:
By the way, are you waiting on me for any other PR reviews at the moment?   I think I'm up-to-date, but not certain.

qutip/core/data/convert.pyx
```diff
@@ -259,14 +261,26 @@ cdef class _to:
                     _converter(convert[::-1], to_t, from_t)
         for dispatcher in self.dispatchers:
             dispatcher.rebuild_lookup()
+        for dtype in self.dtypes:
+            self._str2type[dtype.__name__.lower()] = dtype
+
+    def parse(self, dtype):
+        if isinstance(dtype, type):
```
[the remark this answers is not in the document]
jakelishman: Also, second thought - since this is the fast-path, I think it might be a good idea to have the first test be `if type(dtype) is type` instead.  We know that the only thing we store in our `dict` are immediate `type` instances and never subclasses, so we know for certain it'll fail if the passed argument is a subclass and `type(x) is type` is the fastest possible test.

hodgestar:
It would be good if our naming can help avoid confusion between the concept of a `dtype` (which is the type of elements within an array) and our `data backend type` (which is how the array itself is represented). I'm not sure what a good convention is, but ensuring the docstrings for the backend type have specific examples of valid inputs would definitely help (e.g. `dense` or `csr`).

hodgestar:
This seems like a good idea to me though! It might also help with the upcoming TensorFlow backend (where we'll probably need to mostly create TF backend QObjs in order to get reasonable performance).

jakelishman:
> It would be good if our naming can help avoid confusion between the concept of a `dtype` (which is the type of elements within an array) and our `data backend type` (which is how the array itself is represented). I'm not sure what a good convention is, but ensuring the docstrings for the backend type have specific examples of valid inputs would definitely help (e.g. `dense` or `csr`).

We actually deliberately called this `dtype` to make it familiar to Numpy syntax. The concept of the Numpy `dtype` is mostly irrelevant when using QuTiP, because all data is invariably `double complex`, and this `dtype` is the larger data type of `Qobj`.  I suppose it could be `fmt` instead (as a common abbreviation of `format`)? Or even the full word `format` I suppose.

Agree on the docstrings - we should have an alias like Numpy's `array_like` that we define clearly in documentation in one place.

qutip/core/data/dense.pyx
Ericgig: I did not understand the documentation the same way and upon testing:
```
>>> np.array(np.ones(4)[::2], order='A', copy=False).flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
...
```
So it seems `copy=False` as priority.

jakelishman: Oh, you're right.  I misunderstood.  In that case, this test is right.

qutip/core/data/dispatch.pyx
Ericgig: I could do:
```
        if type(types) is not tuple:)
            try:
                return self._lookup[(_to.parse(types),)]
            except KeyError:
                raise TypeError("specialisation not known for types: " + str(types)) from None
        types = tuple(_to.parse(arg) for arg in types)
        try:
            return self._lookup[types]
        except KeyError:
            raise TypeError("specialisation not known for types: " + str(types)) from None
```
But it also feels wrong...

jakelishman: Yeah, I couldn't think of anything super neat off the top of my head either - I just wanted to bring it up.  I wouldn't worry about it right now because it's likely not much of an issue.  Just in case we revisit this in the future, the _potential_ problems are
 - `for x in y` Python iteration through a type that Cython may not be able to infer is `tuple` (should check the annotated Cython output / generated C++)
 - this requires allocation of a new `tuple`, in a tight, commonly called function, when we're sure at this point that we already have a tuple of the correct size created
 
Let's just leave it as-is for now - even looking back at my fairly recent Cython code I see parts where I could have done it better, so it's likely I'm just fretting over premature optimisation.

Ericgig:
@jakelishman @hodgestar

I would like to merge this PR soon.
Is the only missing part is a term for `dtype : type or str`?

Numpy use `data-type` which point to https://numpy.org/doc/stable/reference/arrays.dtypes.html
The closest equivalent we have would be `qutip.core.Data`.
Here would `data-type` works if we have a page in the documentation pointing to it?
Otherwise all I can think is `layer-type` or `qdata-type` which feels wrong. I would prefer to keep the explicit
```
    dtype : type or str
        Storage representation. Any data type known to `qutip.data.to` is
        accepted.
```
over creating a new term for users to learn.

qutip/core/data/convert.pyx
```diff
@@ -259,32 +261,64 @@ cdef class _to:
                     _converter(convert[::-1], to_t, from_t)
         for dispatcher in self.dispatchers:
             dispatcher.rebuild_lookup()
+        for dtype in self.dtypes:
+            self._str2type[dtype.__name__.lower()] = dtype
+
+    def parse(self, dtype):
+        """
```
hodgestar, line 268: @Ericgig I updated the docstring and the error messages. Let me know what you think.

qutip/tests/core/data/test_convert.py
```diff
@@ -0,0 +1,71 @@
...
+])
+def test_converters(from_, base, to_, dtype):
+    converter = to[to_, from_]
+    assert isinstance(converter(base), dtype)
+    converter = to[to_]
+    assert isinstance(converter(base), dtype)
+    assert isinstance(to(to_, base), dtype)
+
+
+def test_parse():
```
hodgestar, line 47: @Ericgig I added these tests -- let me know if I left any cases we should check out.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+    cdef readonly list _creators
+
     def __init__(self):
-        pass
+        self._creators = []
 
     def add_creators(self, creators):
-        pass
+        """
+        Add creation functions to make a data-layer object from an arbitrary
```
hodgestar, line 336: @Ericgig I update the docstring. Let me know what you think.

@Ericgig @jakelishman Minor complaint -- this function is called `add_creators` but it completely overrides the list of creators each time, unless I am missing something? Perhaps not something we need to fix in this PR, but I imagine we will need to fix it for the two GSoC projects that are starting soon so that they can add their own creators?

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+        self._creators = sorted(_creators,
+                                key=lambda creator: creator[2],
+                                reverse=True)
+
+    def __call__(self, arg, shape=None):
+        for condition, create, _ in self._creators:
+            if condition(arg):
+                return create(arg, shape)
+        # Fallback on numpy if we don't have any matching creator method
+        return _data.Dense(np.array(arg, dtype=np.complex128),
```
hodgestar, line 400: I wonder if we should rather add this as a default creator below and just raise an error here if no creator is found? That way the default creator can be overridden by setting something with higher priority.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+                raise ValueError(
+                    "Type is not a data-layer type: " + repr(dtype))
+            return dtype
+        elif type(dtype) is str:
+            if dtype.lower() not in self._str2type:
+                raise ValueError(
+                    "Type name is not known to the data-layer: " + repr(dtype))
+            return self._str2type[dtype.lower()]
+        raise TypeError(
+            "Invalid dtype is neither a type nor a type name: " + repr(dtype))
```
jakelishman, line 302: In such inner-loop code we should be careful not to repeat method calls or `in` checks:
```python
In [1]: %load_ext Cython

In [2]: %%cython
   ...: cdef dict _KNOWN = {'dense': object, 'csr': object}
   ...: def current(dtype):
   ...:     if dtype.lower() not in _KNOWN:
   ...:         raise ValueError
   ...:     return _KNOWN[dtype.lower()]
   ...:
   ...: def alternate(dtype):
   ...:     try:
   ...:         return _KNOWN[dtype.lower()]
   ...:     except KeyError:
   ...:         raise ValueError from None
   ...:

In [3]: %timeit current('CSR')
   ...: %timeit alternate('CSR')
259 ns ± 3.57 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
159 ns ± 0.394 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: 1 - 159/259
Out[4]: 0.3861003861003861
```
So a nearly 40% speedup by not duplicating calls.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+
+                * Objects that are instances of data-layer types are
+                  converted using ``.copy`` with priority ``100``.
+                * Objects that are ``numpy.ndarray`` instances or for
+                  which ``scipy.sparse.isspmatrix_csr`` is ``True`` are
+                  converted using an internal CSR converter with
+                  priority ``80``.
+                * Objects for which ``scipy.sparse.issparse`` is ``True``
+                  are converted using an internal CSR converter with
+                  priority ``10``.
```
jakelishman, line 377: This looks like it's one, then two levels too indented - the first paragraph doesn't need an indent at all, and if I remember right, rst lists shouldn't be indented either.

qutip/tests/core/data/test_convert.py
Ericgig: We could read type for to.dtype to automatically add new types instead of fixing it to "CRS", "Dense".
I would be one of the only test that test error message... Not sure if I like it.
Testing error is fine but the exact error message?

qutip/core/data/convert.pyx
Ericgig: Do we expect user to know this function exist?
If so, do we want to use the `data-layer` term?
data-layer is a new term for users that we use internally, not sure if we describe it anywhere.

If we don't, It's great.

qutip/core/data/convert.pyx
Ericgig: I created this function in this PR and yes allowing to add new creator for the new data-type elsewhere is worth it (and expected behaviour from the name). So I should fix it before merging, well spotted.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+            elif len(creator) == 3:
+                condition, create, priority = creator
+            if not callable(condition):
+                raise TypeError(str(condition) + " is not a callable")
+            if not callable(create):
+                raise TypeError(str(create) + " is not a callable")
+            _creators.append((condition, create, priority))
+        self._creators = sorted(_creators,
+                                key=lambda creator: creator[2],
+                                reverse=True)
```
jakelishman, line 393: This `add_creators` function seems to completely overwrite all known creators.  It can always trust that its internal list is sorted, so it should probably just insert into the list, rather than replacing it.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+
+    def __call__(self, arg, shape=None):
+        for condition, create, _ in self._creators:
+            if condition(arg):
+                return create(arg, shape)
+        # Fallback on numpy if we don't have any matching creator method
+        return _data.Dense(np.array(arg, dtype=np.complex128),
+                           shape=shape, copy=False)
```
jakelishman, line 401: We should document this somewhere.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,171 @@ cdef class _to:
...
+        lambda arg: isinstance(arg, np.ndarray),
+        lambda arg, shape: _data.Dense(arg, shape=shape, copy=False),
+        80
+    ),
+    (
+        lambda arg: scipy.sparse.issparse(arg),
+        lambda arg, shape: _data.CSR(arg.tocsr(), shape=shape),
+        10
+    ),
+])
```
jakelishman, line 431: Using lambdas here makes this object un-pickleable.  I think we should probably avoid that in the core library - I don't have a particular use-case in mind, though, so it's probably not necessary to change right now.

qutip/tests/core/data/test_convert.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
+import pytest
+from scipy import sparse
+from qutip.core.data import to, create, CSR, Dense, dense, csr
```
jakelishman, line 4: The data layer uses pretty common names - I had generally intended that it would always take qualified access, so `from qutip import data`, then `data.to`, `data.create`, etc, because it's not at all obvious that a function called `to` should be from `qutip` and not any other library a user might have.  We should probably write in that style in the tests, since they are used as a source of usage examples.

qutip/tests/core/data/test_convert.py
jakelishman: I think all of the `assert to.parse(...) is ...` should be separate parametrised cases of a single test, so they're all guaranteed to run - we get more information on test failures that way, and better reporting from pytest.  As it is, if there's a string casing bug, the exception tests wouldn't run.

We usually (or try to) do something like
```python
@pytest.parametrize(['input', 'type'], [('dense', Dense), ('Dense', Dense), ('DeNsE', Dense), ...])
def test_parse_correct_values(input, type):
    assert data.to.parse(input) is type

@pytest.parametrize(['input', 'exc_class', 'exc_message'], [
    (5, TypeError, "Invalid dtype is neither a type nor a type name: 5"),
    ...
])
def test_parse_raises(input, exc_class, exc_message):
    with pytest.raises(exc_class) as exc:
        data.to.parse(input)
    assert str(exc.value) == exc_message
```

qutip/tests/core/data/test_csr.py
```diff
@@ -306,3 +306,25 @@ def test_diags(self, diagonals, offsets, shape):
...
+        pytest.param((1,1), (0,0), None, id='minimal'),
+        pytest.param((10,10), (5,5), 1.j, id='on diagonal'),
+        pytest.param((10,10), (1,5), 1., id='upper'),
+        pytest.param((10,10), (5,1), 2., id='lower'),
+        pytest.param((10,1), (5,0), None, id='column'),
+        pytest.param((1,10), (0,5), -5j, id='row'),
+        pytest.param((10,2), (5,1), 1+2j, id='tall'),
+        pytest.param((2,10), (1,5), 10, id='wide'),
+    ])
+    def test_one_element(self, shape, position, value):
```
jakelishman, line 320: We should test that this fails safely with a Python exception if it's asked to place an element outside the shape.

qutip/tests/core/data/test_csr.py
```diff
@@ -306,3 +306,25 @@ def test_diags(self, diagonals, offsets, shape):
...
+        test = np.zeros(shape, dtype=np.complex128)
+        if value is None:
+            base = data.one_element_csr(shape, position)
+            test[position] = 1.0+0.0j
+        else:
+            base = data.one_element_csr(shape, position, value)
+            test[position] = value
+        assert isinstance(base, data.CSR)
+        assert base.shape == shape
+        np.testing.assert_allclose(base.to_array(), test, rtol=1e-10)
```
jakelishman, line 330: This should probably be an exact test, not a relative one.

qutip/core/data/convert.pyx
Ericgig: Do we want to make `Qobj` from anything that can be made an array?
In v4, we never directly raised any error: for unknown type it fell-back to `scipy.csr_matrix` and let it fail.
Here I do the same, but with numpy instead of scipy.

This here should not be the default but the fallback, so any creator will overwrite it.

qutip/tests/core/data/test_dense.py
```diff
@@ -215,3 +215,62 @@ def test_identity(self, dimension, scale):
...
+        test = np.zeros(shape, dtype=np.complex128)
+        if value is None:
+            base = data.one_element_dense(shape, position)
+            test[position] = 1.0+0.0j
+        else:
+            base = data.one_element_dense(shape, position, value)
+            test[position] = value
+        assert isinstance(base, data.Dense)
+        assert base.shape == shape
+        assert np.allclose(base.to_array(), test, atol=1e-10)
```
jakelishman, line 276: Same comments about testing failure conditions and exact tests.

qutip/tests/core/test_operators.py
```diff
@@ -31,112 +31,139 @@
...
-import pytest
-
+import scipy.sparse as sp
+from functools import partial
 import qutip
-from qutip import (jmat, basis, destroy, create, displace, qzero, qeye,
-                    num, squeeze, charge, tunneling)
+from qutip import Qobj
+from qutip.core.operators import *
+from qutip.core.data import to
```
jakelishman, line 42:
```python
from qutip import data as _data

_data.to
```

qutip/tests/core/test_states.py
```diff
@@ -33,21 +33,33 @@
...
 import pytest
 import numpy as np
-from numpy.testing import assert_, run_module_suite
+import scipy.sparse as sp
 import qutip
-from qutip import (expect, destroy, coherent, coherent_dm, thermal_dm,
-                   fock_dm, triplet_states)
+from qutip import Qobj
+from qutip.core.states import *
+from qutip.core.data import to
```
jakelishman, line 40:
```python
from qutip import data as _data
```

qutip/tests/core/test_states.py
```diff
@@ -33,21 +33,33 @@
 
 import pytest
 import numpy as np
-from numpy.testing import assert_, run_module_suite
+import scipy.sparse as sp
 import qutip
-from qutip import (expect, destroy, coherent, coherent_dm, thermal_dm,
-                   fock_dm, triplet_states)
+from qutip import Qobj
+from qutip.core.states import *
```
jakelishman, line 39: I don't think we should be using star-imports at all, even in tests - they're very unclear for the reader of the file.

qutip/tests/core/test_superoper.py
```diff
@@ -65,7 +65,7 @@ def testOperatorVector(self):
         N = 3
         rho1 = qutip.rand_dm(N)
         rho2 = qutip.vector_to_operator(qutip.operator_to_vector(rho1))
-        assert (rho1 - rho2).norm('max') < 1e-8
+        np.testing.assert_allclose(rho1.full(), rho2.full(), 1e-8)
```
jakelishman, line 68: It would have been better to keep these changes out of this unrelated PR.  This stuff is being fixed on `master` already in a relevant PR, and it's going to cause merge conflicts now.  I can deal with it, though, so it's probably not worth the effort of removing.

qutip/tests/test_random.py
```diff
@@ -31,65 +31,233 @@
...
-def test_rand_super_bcsz_cptp():
-    """
-    Random Qobjs: Tests that BCSZ-random superoperators are CPTP.
-    """
-    S = rand_super_bcsz(5)
-    assert_(S.iscptp)
-
-def check_func_dims(func, args, kwargs, dims):
+from qutip.random_objects import *
+from qutip.core.data import to
```
jakelishman, line 39: avoid star-imports, don't import `to` directly

jakelishman: Are the changes in this file related to this PR?

jakelishman (commented):
It's a bit hard to review now because I think there are lots of changes made that aren't related to the headline of the PR, so it's difficult to pull out the important bits - I'm not sure I spotted everywhere the new creation operations are used.  I've done my best, though.

qutip/core/data/convert.pyx
jakelishman: I think we don't generally expect users to know it exists, but it'll show up if they do `help(data.to)` or look in the API documentation, so we should make sure its docstring is clear.

I think if we commit to the aliases just being strings, it'll be fine to call the type `type or str` all the way through the codebase.

qutip/core/data/convert.pyx
jakelishman: Ah, just saw this comment about `add_creators` - I've got the same comment in my review as well.

qutip/core/data/convert.pyx
jakelishman: As it stands, I don't think it matters a great deal whether this is a creator with predicate `lambda arg: True` and priority `-inf`, or written in like this.  That would change if we had a method to _remove_ creators as well, but I can't think of any neat way to do that at all (we probably _really_ don't want to make the backing list public), so I don't really see a difference.

I think we probably do want to accept anything array-like - it's very convenient (and I'm pretty sure used in the test base) to be able to do `Qobj([[1, 0], [0, 1]])` and such like.

qutip/core/data/convert.pyx
```diff
@@ -259,63 +261,158 @@ cdef class _to:
...
+
+    def parse(self, dtype):
+        """
+        Return a data-layer type object given its name or the type itself.
+
+        Parameters
+        ----------
+        dtype : type, str
+            Either the (case-insensitive) name of a data-layer type or a type
+            itself.
```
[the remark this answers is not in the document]
Ericgig: Make sense, those 4 alias are all we really need.

qutip/tests/core/test_superoper.py
jakelishman: PS: I'd like to merge up `master` to `dev.major` again once your outstanding PRs are in.  Ideally we'll do that before the GSoC students start doing any real work, so the docs are up.

qutip/tests/test_random.py
Ericgig: I only needed 1 new test, but having 2 files for these tests was bugging me...
I will split this in a new PR.

jakelishman: Oh, I wouldn't go to the effort of splitting it - could you just tag which tests are really new?

Ericgig: `test_random_type` is the only new test.

But I also rewrote them with `pytest.fixture`.
So while it's the same thing that is tested, the mechanic is changed.
`test_random_seeds(rand_func_check)` was 6 tests in the now deleted test_rand.py...

I did do too much when adding 1 test and I thinks is worth to actually push that to master, not dev.major.
(There will be a conflict when merging master->dev.major)
It won't be hard to split the PR, I just have to checkout this one file in a new branch and erase the new test.

qutip/core/data/dense.pyx
```diff
@@ -346,3 +353,108 @@ cpdef Dense from_csr(CSR matrix, bint fortran=False):
...
+    """
+    Construct a matrix from diagonals and their offsets.  Using this
+    function in single-argument form produces a square matrix with the given
+    values on the main diagonal.
+    With lists of diagonals and offsets, the matrix will be the smallest
+    possible square matrix if shape is not given, but in all cases the
+    diagonals must fit exactly with no extra or missing elements. Duplicated
+    diagonals will be summed together in the output.
+    Parameters
+    ----------
```
jakelishman, line 377: Missing a blank line

qutip/core/data/make.py
```diff
@@ -0,0 +1,100 @@
...
+    Construct a matrix from diagonals and their offsets.  Using this
+    function in single-argument form produces a square matrix with the given
+    values on the main diagonal.
+    With lists of diagonals and offsets, the matrix will be the smallest
+    possible square matrix if shape is not given, but in all cases the
+    diagonals must fit exactly with no extra or missing elements. Duplicated
+    diagonals will be summed together in the output.
+    Parameters
+    ----------
```
jakelishman, line 17: Missing blank line before the section header.

jakelishman (commented):
Looking solid - I just flicked through again and saw a couple of missing blank lines in some docstrings, and I think there are a couple of outstanding changes from the last review.

+1 on the alias system and the bad-path tests of `one_element`

Ericgig:
I reset the `test_random`, `test_operator` and `test_states` and only added test relating to `dtype`. The others change I made will go in other PR. 

I also changed `add_creator` to take only one new creator at a time and add it to the list. I re-sort each time as the list. Creators are added in `core.data.__init__`, including the fallback to numpy. I had some circular import when import data in `convert.pyx` and it fell more at it's place after `add_conversions` is called.

Could you take another look and if it's good I will clean the commit history.

qutip/core/data/__init__.py
```diff
@@ -34,5 +35,20 @@
     (Dense, CSR, dense.from_csr, 1),
     (CSR, Dense, csr.from_dense, 1.4),
 ])
+to.register_alias('csr', CSR)
+to.register_alias('CSR', CSR)
+to.register_alias('Dense', Dense)
+to.register_alias('dense', Dense)
+
+
+import qutip.core.data.creator_util as creator_util
```
jakelishman, line 44: Can just be `from . import creator_util` - but let's also call it `_creator_utils` (leading underscore and plural) instead to make it clear that it's completely private.  With the functions living inside `data.create`, things like Sphinx might still find it.

qutip/core/data/convert.pyx
```diff
@@ -260,61 +262,159 @@ cdef class _to:
...
 cdef class _create:
+    cdef readonly list _creators
+
     def __init__(self):
-        pass
+        self._creators = []
 
-    def add_creators(self, creators):
-        pass
+    def add_creator(self, condition, creator, priority=0):
```
jakelishman, line 355: What was the reason for making it only take a single one at a time?  This way seems inconsistent with `data.to.add_conversions`, and it introduces a quadratic dependency (complexity goes to O(n^2 log n)).  Not that the complexity will matter, really, but I'm not sure I see the reason, and it isn't consistent with `data.to`.

qutip/core/data/creator_util.py
```diff
@@ -0,0 +1,41 @@
...
+    return CSR(arg.tocsr(), shape=shape)
+
+
+def true(arg):
+    return True
+
+
+def numpy_array(arg, shape):
+    array = np.array(arg, dtype=np.complex128)
+    return Dense(array, shape=shape, copy=False)
```
jakelishman, line 41: I think the `np2dense`, `sparse2csr` and `numpy_array` functions are all unnecessary - the `CSR` and `Dense` constructors have all this behaviour already (and logically all of these _should_ copy the data).

Ericgig: I though `CSR` only accepted csr input and `Dense`, array...
`Dense`'s copy default is `True` but `CSR`'s is `False`, I will add a copy option to the creator function.

qutip/core/data/convert.pyx
Ericgig: `add_conversions` force rebuilding the dispatch table. So it's worth doing at once. 
But here the time cost is insignificant and the function cleaner that way.
Doing it one at a time feels more natural, and with the dispatch table, we are forced to do it at once in `add_conversions`.

qutip/core/data/creator_util.py
jakelishman: Oh, let's change `CSR.__init__` to `copy=True` - I think it's much safer if that's always the default.  Yeah, having a `copy` parameter on the `create` function is definitely the right call.  I think we'll need to change `Qobj.__init__` a little to cope, but that's ok - I think I knew we'd have to revisit that at some point anyway.

qutip/core/data/convert.pyx
jakelishman: I'm not sure about it feeling more natural (iterables are very Pythonic), and not certain how it would be significantly more complicated in a loop, but ok.  My concern is less about necessity and more about consistency.  I'd imagine that if someone wants to implement a new type, they probably want to add conversions from several different types of object (just like we do), so as a user I'd expect the function to take several at once like with `add_conversions`.  How easy/difficult it is for us internally is secondary to consistent UX for me.  (In a similar vein, I might even make the argument that `register_alias` should be `register_aliases`, and allow either a string or an iterable as its `alias` argument.)

Ericgig: Yes, `register_aliases(([CSR', 'csr'], CSR)` feels natural. Reverting.

jakelishman (approved):
Only super minor stuff from me now.  I've given it my tick on the assumption that the last couple of points of error-checking (and the suggested commits to remove old info) get included.  I think Simon (@hodgestar) is still be looking though, so wait for his as well.  I know I'm still on the hook to review the immutable coefficient one - I haven't forgotten, I promise.

hodgestar (commented):
Left some small requests for exception improvements, but otherwise looks good to me!

qutip/core/data/convert.pyx
```diff
@@ -260,61 +262,170 @@ cdef class _to:
...
+            ........
+            condition : callable (object) -> bool
+                Function determining if the given object can be converted to a
+                data-layer type using this creator.
+
+            creator function: callable (object, shape) -> Data
+                The creator function. It should take an object and a shape
+                and return a data-layer type instance. The object may be
+                any object for which the condition function returned ``True``
+                when tested.
```
jakelishman, line 381:
```suggestion
            creator function: callable (object, shape, copy=True) -> Data
                The creator function. It should take an object and a shape and
                return a data-layer type instance. The object may be any object
                for which the condition function returned ``True`` when tested.
                The ``object`` and ``shape`` parameters are passed positionally,
                and the ``copy`` parameter is passed by keyword.
```

qutip/core/data/convert.pyx
```diff
@@ -260,61 +262,170 @@ cdef class _to:
...
+        arg : object
+            Object to be converted to `qutip.data.Data`. Anything that can be
+            converted to a numpy array are valid input.
+
+        shape : tuple, optional
+            The shape of the output as (``rows``, ``columns``).
+        """
+        for condition, create, _ in self._creators:
+            if condition(arg):
+                return create(arg, shape, copy)
```
jakelishman, line 427:
```suggestion
                return create(arg, shape, copy=copy)
```

Since we're relying on the `Dense` and `CSR` constructors, let's pass `copy` as a keyword.  I think `arg` and `shape` are fine to pass positionally, but having `copy`, which is usually an optional argument, be keyword gives user-defined functions a bit more flexibility.

jakelishman (commented):
Oh sorry, one more comment about passing conventions in the creators.  But that's it from me, promise.

qutip/core/data/convert.pyx
```diff
@@ -260,61 +262,159 @@ cdef class _to:
...
-        if isinstance(arg, type):
+        if type(arg) is not tuple:
             arg = (arg,)
-        if not isinstance(arg, tuple) or not arg or len(arg) > 2:
+        if not arg or len(arg) > 2:
             raise KeyError(arg)
-        to_t = arg[0]
-        if to_t not in self.dtypes:
-            raise TypeError("to_type is not known: " + str(to_t))
+        to_t = self.parse(arg[0])
```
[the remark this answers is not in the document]
jakelishman: I'm not sure - I think it's quite neat to have the exact same errors raised regardless of the call method.  The `__getitem__` is more used as an overload, so the syntaxes `to(arg, type)` or `to[in_type, out_type]` are really just the same thing.

Particularly with the alias system, I think it is quite nice to have the `TypeError` and `ValueError` separation maintained; the former is _always_ going to be because of programming error, but the latter may appear for typo'd aliases, or if you've accidentally forgotten to import a plugin you wanted.  That seems kind of helpful to me as a user, whereas `KeyError` kind of seems not quite right - the fact that the dispatch system truly is handled by a key-lookup is kind of an implementation detail.  Maybe I'm not the best person to ask, though - I know too much about the internals, so what's "expected" to me might not be what a user thinks.

qutip/core/data/_creator_utils.py
```diff
@@ -24,4 +24,8 @@ def true(arg):
 
 
 def data_copy(arg, shape, copy=True):
+    if shape and shape != arg.shape:
```
jakelishman, line 27: Probably `shape is not None` to help catch cases of a user doing something super dumb like `shape=()` or `shape=0` by accident.
```suggestion
    if shape is not None and shape != arg.shape:
```

hodgestar:
@Ericgig Thank you for making all the small changes. Shout when you'd like a re-re-review.

Ericgig:
@hodgestar I think I included all your comments and improvement already. 
I rebased today's 1 line commits into one and it seems to have properly kept everyone contribution.

hodgestar:
@Ericgig Not sure if it's directly related to this PR, but the Travis build failed on `test_diag_liou_mult` with the following tolerance error:
```
>           np.testing.assert_allclose(target, calculated, atol=1e-12)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=1e-12
E           
E           Mismatched elements: 2 / 6724 (0.0297%)
E           Max absolute difference: 3.36667153e-12
E           Max relative difference: 1.85567862e-08
```
Not sure if we should relax the tolerance slightly now (maybe 1e-11?), or in a separate PR later, or if there is something real to fix.

jakelishman:
@hodgestar: already fixed in `master`: #1474

qutip/core/data/convert.pyx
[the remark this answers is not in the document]
hodgestar: I feel like `__getitem__` is kind of expected to mostly raise `KeyError` since that's what the interface looks like from the outside, but we can always fix it if anyone complains. :)

hodgestar (approved):

hodgestar:
> @hodgestar: already fixed in `master`: #1474

And I even approved that PR! Doh. :)

A good reminder that we need to merge master into `dev.major` once the solver PRs are done though.

hodgestar:
@Ericgig Yay! Let's merge! :D

Issue #846, opened 2018-03-31
mesolve gives num_collapse =0


TejasAvinashShetty:
mesolve returns num_collapse =0, even when I pass collapse operators to it like this 
 ``` python
from qutip import *
import numpy as np
import matplotlib.pyplot as plt
from math import *

times = np.linspace(0.0, 10.0, 200)
psi0 = tensor(fock(2,0), fock(10, 5))
a = tensor(qeye(2), destroy(10))
sm = tensor(destroy(2), qeye(10))
H = 2 * np.pi * a.dag() * a + 2 * np.pi * sm.dag() * sm + 2 * np.pi * 0.25 * (sm * a.dag() + sm.dag() * a)


result = mesolve(H, psi0, times, [np.sqrt(0.1)*a], [a.dag()*a, sm.dag()*sm])
print("result.num_collapse",result.num_collapse)
# I am getting the no of collapse operators as zero

print("len(result.expect[0]): {}".format(len(result.expect[0])))

fig1 = plt.figure()
ax1 = fig1.add_subplot(1, 1, 1)

ax1.set_title("Final evolution")
ax1.set_xlabel("Time")
ax1.set_ylabel("P")
ax1.plot(times, result.expect[0], 'b*', linewidth=2)
ax1.plot(times, result.expect[1], 'g+', linewidth=2)

plt.show()
``` 
(Some of the code is from @ajgpitch ).

PFA the test run of the qutip.testing module as [qutip_testing.pdf](https://github.com/qutip/qutip/files/1865397/qutip_testing.pdf) so that anyone interested could see if there is anything wrong with my installation. Also please let me know if there is any other piece of information that needs to be known.

>Computer OS: MacOS 10.13.3 High Sierra.

-- 
Thanking You,

Kind regards,

Tejas Shetty

Email ID: tejasshetty.1808@gmail.com

**Comment by TheIdealis** [comment]
I would guess that this behavior was intended and that storing the number of collapse operators is intended for the results of of `mcsolve`. Nonetheless, you can add it easily in the `mesolve` function beneath:
``` python
res = _mesolve_const(H, rho0, tlist, c_ops,
                     e_ops, args, options,
                     progress_bar)
res.num_collapse = len(c_ops)
```
or the whole conditional statement.

nathanshammah:
Try setting default for `num_collapse = None`.

jakelishman:
Fixed by #1424.

PR #1424, opened 2021-01-28
Correctly set num_collapse in mesolve output

jakelishman:
In the result from `qutip.mesolve(..., c_ops=[...])`, the attribute `num_collapse` would always be 0 no matter how many `c_ops` were actually passed.  This was correct in `mcsolve`, but missing in `mesolve`.  This is likely because the result is mostly created within `_generic_ode_solve()`, but at this point the `c_ops` list has already been consumed in the creation of the Liouvillian.  There is no need to set `col_times` and `col_which` (compare mcsolve) because mesolve tracks the full density matrix, rather than following the quantum trajectory approach.

Noticed by @nathanshammah during @quantshah's talk on Twitch.

@Ericgig: to check - I assume this wasn't by design?

### DIFF
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -265,6 +265,7 @@ def mesolve(H, rho0, tlist, c_ops=None, e_ops=None, args=None, options=None,
 
     res = _generic_ode_solve(func, ode_args, rho0, tlist, e_ops, options,
                              progress_bar, dims=rho0.dims)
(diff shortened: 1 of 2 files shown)


Ericgig:
No, it's not intended, thank you for catching.

jakelishman:
@nathanshammah: turns out you already knew about this! See: #846.

PR #1427, opened 2021-01-30
Improve the error message of circuit plotting

### LINKED ISSUE #1426: Ambiguous error message for circuit plotting
**Describe the issue**
We use subprocess for the LaTeX plotting procedure. But the current error message for a failed subprocess is very ambiguous, which makes it very hard to debug.

For illustration purpose, I deliberately change the following line so that the input pdf file has a wrong name.
https://github.com/qutip/qutip/blob/403e466becbc5d36947ca15b9211d884df83f91a/qutip/qip/circuit_latex.py#L172
I changed it to e.g. `in_file = "pdf-that-does-not-exist.pdf"`. Running the code, the output error message is
```
CalledProcessError: Command '('magick', '-density', '100', 'aqcirc.pdf', 'qcirc.png')' returned non-zero exit status 1.
```
This is not very useful since the exit status 1 can have many reasons. It can also be that an additional dependency is missing (which is the reason I started to debug at the first place).
What I would like to see is the real system error message:
```
magick: unable to open image 'pdf-that-does-not-exist.pdf': No such file or directory
```

**Possible solution**
I didn't find a nice solution for this. Simply changing the parameter of `subprocess.run` to `stderr=subprocess.PIPE` doesn't work for me. I have to catch the exception and print the stderr message:
```
def _run_command(*args, **kwargs):
    try:
        functools.partial(subprocess.run, check=True, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE)(*args, **kwargs)
    except subprocess.CalledProcessError as err:
        raise RuntimeError(err.stderr)
```
which gives me the output
```
RuntimeError: b"magick: unable to open image 'aqcirc.pdf': No such file or directory @ error/blob.c/OpenBlob/3496.\r\n"
```
It's ok but still weird. I guess I overlooked some feature of `subprocess`.

@jakelishman I'm wondering if you have a better solution?

BoxiLi:
Catch and raise the error message of the subprocess.

**Description**
See https://github.com/qutip/qutip/issues/1426#issuecomment-770080582

**Related issues or PRs**
This should fix #1426 

**Changelog**
Improve the error message of circuit plotting

### DIFF
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -33,6 +33,7 @@
 import collections
 import functools
 import os
+import sys
 import shutil
 import subprocess
 import tempfile
@@ -48,9 +49,16 @@ _latex_template = r"""
 """
 
 
-_run_command = functools.partial(subprocess.run, check=True,
-                                 stdout=subprocess.DEVNULL,
-                                 stderr=subprocess.DEVNULL)
+def _run_command(command, *args, **kwargs):
+    try:
+        return subprocess.run(
+            command, *args,
+            check=True, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE,
+            **kwargs,
+        )
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(e.stderr.decode(sys.stderr.encoding)) from None
+
 _run_command.__doc__ = \
     """
     Run a command with stdout and stderr explicitly thrown away, raising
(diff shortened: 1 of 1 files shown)


BoxiLi:
I took your improved solution :) Like the `raise ... from None` syntax.

jakelishman:
The docstring can come live in its usual place, but otherwise that looks good to me! I'm always in favour of improving error meesages.

jakelishman (approved):

PR #1429, opened 2021-01-31
Build wheels and distribute to PyPI (pip) from GitHub Actions

jakelishman:
This patch overhauls how QuTiP is packaged and distributed.  Once this patch is merged to `master` (not just `dev.major`), there will be an option on QuTiP's GitHub Actions tab for people who have write permission on qutip/qutip to build all wheels and optionally release the build to PyPI, making it available on `pip` as a binary release.

The major changes are in the files `pyproject.toml`, `setup.py`, `setup.cfg` and `.github/workflows/build.yml`, while `MANIFEST.in` and a new file `VERSION` were also touched.  The patch also contains two commits that remove unused OpenMP code and remove a particularly problematic C struct from some inner Python code; the struct and the presence of the dead code caused portability problems, and they in no way change the behaviour of the package.

### Distribution changes:
  - there is a GitHub Action which will build all the wheels for CPython 3.6+ on Linux (`manylinux1` x86 and x86_64), macOS and Windows (32- and 64-bit).  This action is triggered manually, and the wheels will be available for download afterwards.  Optionally, the action will push to PyPI, making the version immediately available on `pip` by binary release.
  - `.cpp` files are now distributed with the wheels but not added to source control.  This is a step towards a full Cython-free delivery of the QuTiP source (binary releases never require Cython), but right now I think our `setuptools` machinery doesn't quite handle that.

### Changes to `setup.py`
  - all package metadata is moved out into `setup.cfg`, which is easier to read and change, and modern `setuptools`' preferred way of doing things
  - build requirements are moved into `pyproject.toml` including listing `setuptools` as the build method, in accordance with PEP 517.  Any PEP 517-complaint installation and build process will now _enforce_ the presence of `numpy`, `scipy` and `cython` (and `setuptools`) before attempting the build, so no more need for attempted import guards.
  - packages are now discovered rather than explicitly stated, using standard `setuptools` machinery
  - Cython extension modules are now discovered rather than explicitly stated, using globbing
  - version information is now stored in a single-line file `VERSION`.  This is mostly for CI reasons; it's much easier to parse and override at the CI level when it's a single file, rather than attempting  `sed s/.../.../ setup.py` or something crazy like that.  The validity of the version string is tested by regex.
  - local version information is now _always_ appended, even if `git` isn't present - the local version is just then "nogit" rather than the git shorthash
  - `setup.py` gained a new `--release` flag or environment variable `CI_QUTIP_RELEASE`; the only effect right now is to suppress the local identifier on the version information, replacing the old `ISRELEASED` boolean flag in the code

### Notes on the wheels

Currently I build only for CPython.  I haven't tried to build pypy wheels, but we _may_ support it - Cython does in theory, but I think complex numbers and raw pointers aren't handled completely smoothly, and we make heavy use of both.  We get `manylinux1` builds for Linux, which is the oldest `manylinux` spec.  This can be updated in the future, but I'm not sure I see a need; all our heavy mathematical lifting is either done by custom code (which is mostly immune) or BLAS/LAPACK (which we link to dynamically), so I don't think there's much to be gained.  I don't know if the macOS images will run on the new M1 chips, but if not, there's not much that can be done there until GitHub Actions adds the necessary cross-compilation headers and libraries to their CI.

### Examples

You can see the result of this upload on the testing PyPI server: https://test.pypi.org/project/qutip-jakelishman (version 5.0.0a1 failed to upload due to previous testing, and 5.0.0a2 has a broken OpenMP configuration so won't run).  I ran the CI action from my fork, with a temporary commit that changed the server to the testing archive (and the package name).  Here's a screenshot of what the deployment screen will look like:

GitHub Actions screen when triggering the build and deploy action

### Further action needed

The GitHub Action won't become available until this is merged, _and_ `dev.major` is merged into `master`.  This is because GitHub only makes workflows on the default branch available.

~We also need to generate a PyPI secret token with write access to the `qutip` package, and add it to the qutip/qutip repository secrets with the name `PYPI_TOKEN` (as defined in `.github/workflows/build.yml`).~ *edit: Done 2021-02-15.*


jakelishman:
Oh wait, that OpenMP stuff might not have been entirely unused - that's what I get for not re-testing after making non-trivial changes.  I'll fix that, but the principles of how the distribution works should stay the same.

jakelishman:
Note: in theory you can install QuTiP from the test PyPI server right now by doing
```
pip install -i https://test.pypi.org/simple qutip-jakelishman
```
but since it's on the "wrong" server, you first have to manually install the dependencies yourself (`numpy` and `scipy`).  When it's released properly on PyPI it'll handle the dependencies correctly - this is just an artifact of being on the testing server.

jakelishman:
Ok, I'm not sure that enforcing numpy >= 1.20 at compile time is sufficient.  I don't know if it's a Cython thing, a numpy thing or an us thing, but I'll figure it out later when I've got more time.  The distribution changes here are still good here, and if it ends up that for the initial release of 5.0 we only support numpy 1.20+, that's unfortunate but not really the end of the world.

hodgestar (approved):
These changes look good to me!

jakelishman:
Thanks!  I'll work out how to handle this numpy 1.20 C-API change properly before this is ready to merge, I think - turns out my reading of the numpy release notes was wrong, or (possibly) Cython is generating some compile-time-constant code it should be deciding at runtime (as of the latest numpy).  Hopefully it's just a configuration change on our end that I can work out somewhere.

jakelishman:
Ok, it turns out I got the numpy release notes the wrong way round; we have to build our binaries with `1.16.6 <= numpy < 1.20` to maintain ABI compatibility no matter the numpy version.  I have updated the compiler infrastructure to take this into account - this is in some ways a temporary restriction until two or three years in the future when we can drop support for all numpy versions less than 1.20.

@hodgestar, @nathanshammah: perhaps you could just glance over this again and make sure everything looks right to you?  Simon, I know you already looked over it once, so if you've not got time again, no problem - barely anything has changed since you last saw it.

I wrote this PR against `dev.major`, but I suppose in theory I could backport it to the 4.x branch if that's absolutely required.  I think some parts of the setup changed for `dev.major`, which is why I didn't originally try to make it to the current version.

hodgestar:
Happy with the numpy version range changes since my last look!

BoxiLi:
Will this also restrict all the downstream packages to only use NumPy <1.20 or is it just a restriction when compiling `qutip`?

jakelishman:
@BoxiLi purely a build requirement.  Installation allows any numpy >= 1.12.

setup.cfg
```diff
@@ -0,0 +1,42 @@
...
+packages = find:
+include_package_data = True
+zip_safe = False
+install_requires =
+    numpy>=1.12
+    scipy>=1.0
+setup_requires =
+    numpy>=1.16.6,<1.20
+    scipy>=1.0
+    cython>=0.29.20
```
jakelishman, line 35: @BoxiLi: the different install vs setup limits are here

jakelishman: @Ericgig this is a different PR to the merge, but it's clearer to see the difference in requirements in the new packaging methods (note that I'll likely be rebasing/redoing this PR if #1453 is merged anyway, or backporting it to master)

jakelishman: (also note the numpy `install_requires` version in this PR should be bumped up to 1.16.6 - whoops!)

hodgestar:
Still looks good to me.

We could put some of the code in `setup.py` into functions so that it's a bit more clear what is temporary state and how the code is organized, but we don't have to fix everything in one PR.

nathanshammah:
I'll add a review by the end of next week. Sorry for the delay.

nathanshammah (commented):
Hi @jakelishman, thanks for the endeavor and the care in documenting the steps and issues. 

- "GitHub only makes workflows on the default branch available.": sorry what do you mean by this? 
- A related question: Line 6 - 9 in  `.github/workflows/build.yml`  identify on which branch to check the condition to trigger the action? How the current syntax forbids the action from being triggered from a fork?
- Is the setup config change necessary? It looks a bit cumbersome
- regarding testpypi not going to look into the correct place for dependencies, if that is an issue you referenced, I think that  ```
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.python.org/simple/ qutip ``` should fix that, as it tells it to look for anything else not on the testpypi but on pypi.

jakelishman:
1. inside the "Actions" tab at the top of the GitHub page, you will only see workflows that are present in the code in the default git branch.  `dev.major` isn't QuTiP's default branch, so the workflows won't appear immediately after this is merged, only when the `.github/workflows/build.yml` file also exists in `master`.

2. nothing prevents you from running the workflow on a fork, and there's no way and no reason to prevent this - it's also how I tested it.  If you did run it on a fork, you'd be using your own GitHub Actions credits (it doesn't cost us anything because we're an open-source project).  Only qutip/qutip knows the secret to publish to PyPI, so no fork can succeed on that.  The branch check here serves two purposes: the first is that I needed some "tickbox" to let the admin decide at runtime whether they needed to publish the package or just build the wheels for personal download; the second is that we release off release branches, not master, so there needed to be an input to say _which_ branch should be built.

3. it's not 100% necessary, but it's the right thing to do.  It decouples the project data from the setup code, which makes it much easier to modify both, and allows other tools to access the data.  There are other PEPs in the works as well that will move all this data into `pyproject.toml`, rather than just `setup.cfg`.

4. oh, that's useful.  I didn't really know or look into it, because it was only a temporary thing anyway.

I might actually retire this PR and make it against `master` instead.  It's functionally the same as this one, and most of the diff is the same, except that its `setup.py` is better organised, like Simon suggested.

jakelishman:
I think it might be a good idea to obsolete this pull request, and merge in #1465 instead, so we can use in for the next 4.6.0 release.

jakelishman:
Obsoleting in favour of #1465.

PR #1430, opened 2021-02-01
Fix consumable iterator in test parametrisation

jakelishman:
If
```python
    qutip.testing.run()
```
was called twice in the same Python session, then cases where a consumable iterator (such as `itertools.permutations` or `itertools.chain`) was used in the test parametrisation would fail on subsequent runs.  This is because the iterator is consumed when the module is loaded, and cannot be re-run after.

This was not caught in CI because naturally CI only runs the testing suite once in a session.


hodgestar (approved):
Good catch!

PR #1432, opened 2021-02-03
Remove use of deprecated numpy type aliases

jakelishman:
Numpy 1.20 officially deprecated use of `np.int` and other things like `np.complex`.  These were just thin aliases to Python types anyway, so that replacement is safe.  Swap `float` and `complex` to `np.float64` and `np.complex128`; these still match Python precision (and standard double-precision floats), but more importantly we assume at all points in Cython code that we're dealing with double-precision arithmetic.  Anything else would be a larger problem for us.

In cases where `np.int` was used as a type check, we actually care more about checking for Integral typing, rather than specifically the int class; super weird constructs like
```python
isinstance(np.int64(1), int) == False
```
so taking indices out of an ndarry may lead to incorrect results when this sort of test is used.  Better to use the abstract
```python
isinstance(..., numbers.Integral)
```

See: https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

### DIFF
--- a/qutip/_mkl/spmv.py
+++ b/qutip/_mkl/spmv.py
@@ -49,9 +49,9 @@ def mkl_spmv(A, x):
 
      # Allocate output, using same conventions as input
      if x.ndim == 1:
-        y = np.empty(m,dtype=np.complex,order='C')
+        y = np.empty(m,dtype=np.complex128,order='C')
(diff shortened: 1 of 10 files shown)


hodgestar (approved):
Looks good!

Issue #1433, opened 2021-02-03
QuTiP incompatible with numpy 1.20 due to Qobj.__array__


jakelishman:
QuTiP is currently incompatible with numpy 1.20.  Tests will fail to even collect with an error such as
```
AttributeError: 'numpy.ndarray' object has no attribute 'dag'
```
and a lot of functionality will break - anything that requires `Qobj.eigenstates()` for example.

This is because `Qobj` defines `__array__`, one of numpy's "array interface" functions, intended for classes that can be safely converted _implicitly_ into an `ndarray`.  This isn't really the case for `Qobj` - it loses all sorts of information when you do that, which is why we've maintained the separate `Qobj.full()` for explicitly getting the dense matrix representation of a `Qobj`.  This is not to mention that numpy ufuncs probably should not be able to implicitly convert `Qobj` - I'd strongly argue that `np.sin(qutip.basis(2, 1))` should be `TypeError`, _not_ `array([[0.        +0.j], [0.84147098+0.j]])` (like it is right now).  We actually already removed this "functionality" in `dev.major`.

The most pressing incompatibility is that a few points in QuTiP put a few `Qobj` into a `np.array(dtype=object)`.  In numpy 1.20, this no longer produces a 1D array of `Qobj`, but a 3D array of complex.  This breaks `Qobj.eigenstates`, and prevents test collection due to it being present in `states.py::qutrit_basis()`, which is called during parametrisation.

There are a two possible ways to solve this, and we ought to release a fix with one of them in a patch ASAP:
  1. remove `Qobj.__array__`
  2. remove all use of `Qobj` in `np.array`

I'm personally in favour of "explicit is better than implicit" in this case, i.e. removing `Qobj.__array__` and relying on `Qobj.full()`.  As another example along this vein, note `scipy.sparse` matrices don't implement this either, and they're arguably closer to being safely coerced to `ndarray` than we are.

### Related issues

#938: feature request for implementing `__array__`.  This only asks for `np.array(qobj)` as a convenience, acknowledging the availability of `Qobj.full()`.
#1017: includes a comment on buggy behaviour caused by `__array__`.  Note that the solution given there (`np.asarray(..., dtype=object)`) will no longer work with numpy 1.20.

hodgestar:
I agree that performing a ufunc like `np.sin(...)` on a `QObj` and getting an `np.array` back is not that useful since presumably one would like a QObj back (otherwise why not just call `.full()` and work with the resulting numpy array directly).

[...]

PR #1434, opened 2021-02-03
Remove Qobj.__array__ for numpy 1.20 compatibility

jakelishman:
Proposed solution to #1433.  See that issue/commit logs for more details on this PR.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -208,7 +208,6 @@ class Qobj(object):
         Returns normalized quantum object.
 
     """
-    __array_priority__ = 100  # sets Qobj priority above numpy arrays
 
     def __init__(self, inpt=None, dims=[[], []], shape=[],
                  type=None, isherm=None, copy=True,
@@ -1086,12 +1085,6 @@ class Qobj(object):
             return self.data.toarray(order=order).squeeze()
         else:
             return self.data.toarray(order=order)
-
-    def __array__(self, *arg, **kwarg):
-        """Numpy array from Qobj
-        For compatibility with np.array
-        """
-        return self.full()
 
     def diag(self):
         """Diagonal elements of quantum object.


hodgestar (approved):
The change looks good to me.

Did any of the QuTiP tests break? If no, that is quite encouraging because it suggests that a large codebase using QObj might not be affected at all.

I think we need to document somewhere (perhaps in the release notes of documentation on Qobj) what a use should do if they were relying on `__array__` (i.e. call `.full()` explicitly themselves).

jakelishman:
Actually I'm going to close this for now pending further discussion on #1433.  The failing tests are using the implicit array coercion, which we may or may not want, but regardless, there's a bigger decision to be made.

ajgpitch:
I think Michael Goerz suggested adding this as it is somehow helpful for
his krotov package

jakelishman:
He did - he's the original feature requester, but given super super weird interactions with ufuncs (see #1433), it's not so clear to me that it was _necessarily_ the right idea to implement especially now it's causing problems with how numpy arrays are used in the library.  There always was a work-around; allowing `np.array(qobj)` was more to save him from writing `x = x.full() if isinstance(x, Qobj) else np.asarray(x)`.

That said, I came round to the idea more as I read more about the control we can have over the rest of the numpy infrastructure.  We can allow `np.array` on `Qobj` (satisfying Michael's original request), and with a little bit of extra code we can prevent all the super weird ufunc/implicit casting behaviour, because `Qobj` shouldn't be able to implicitly interoperate with raw arrays.  I know in my personal code I'd rather use `Qobj.full()`, but I no longer think it's _wrong_ to allow `np.array(qobj)` given that other people like it, even with these numpy 1.20 problems.  It does make it difficult to put `Qobj` in arrays, but numpy 1.20 has ruled out being able to have `np.array(qobj)` and `np.array([qobj1, qobj2], dtype=object)`, so that's not really our fault.

ajgpitch:
> but numpy 1.20 has ruled out being able to have np.array(qobj) and np.array([qobj1, qobj2], dtype=object)
Sorry, @jakelishman please clarify. Numpy 1.2 does not allow for arrays of objects? Seems strange. I some recollection doing this sometime, I am hoping it's not in qutip.control :/

ajgpitch:
Maybe @goerz wants to comment on these ideas?

jakelishman:
Best to move further discussion to #1433, which is the main issue for this, since this PR is closed pending a decision over there.

PR #1436, opened 2021-02-05
Catch integer overflow segfault in zcsr_kron

### LINKED ISSUE #1406: Segmentation fault while trying to make large tensor product
**Describe the bug**

When I try to make a tensor profuct of a large number of qubits, I get a segmentation fault.

**To Reproduce**

```python
from qutip import *; tensor([basis(2,0) for i in range(31)])
```

The terminal out put is
```
Fatal Python error: Segmentation fault

Current thread 0x000000011b71be00 (most recent call first):
  File "/Users/evanphilip/opt/miniforge3/lib/python3.8/site-packages/qutip/tensor.py", line 116 in tensor
  File "<stdin>", line 1 in <module>
zsh: segmentation fault  python3 -q -X faulthandler
```

**Your Environment**

Current admin team: Alexander Pitchford, Paul D. Nation, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, and Boxi Li
Project Manager: Franco Nori.

QuTiP Version:      4.5.2
Numpy Version:      1.19.2
Scipy Version:      1.5.2
Cython Version:     None
Matplotlib Version: 3.3.2
Python Version:     3.8.6
Number of CPUs:     4
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True
Installation path:  /Users/evanphilip/opt/miniforge3/lib/python3.8/site-packages/qutip
==============================================================================
==============================================================================

jakelishman:
The function _safe_multiply was meant to raise a Python exception if the
integer arithmetic overflowed.  However, neither it nor several of its
call sites had the full exception-handling mechanisms enabled.

In addition, extremely large tensor products such as
    tensor([basis(2, 0) for _ in [None]*31])
would segfault when using 32-bit integer arithmetic, because the size
calculation was not checked with _safe_multiply.

Fix #1406: turns the segfault into an exception.

### DIFF
--- a/qutip/cy/spmath.pyx
+++ b/qutip/cy/spmath.pyx
@@ -356,9 +356,9 @@ def zcsr_kron(object A, object B):
     cdef int rowsB = B.shape[0]
     cdef int colsB = B.shape[1]
 
-    cdef int out_nnz = _safe_multiply(dataA.shape[0], dataB.shape[0])
-    cdef int rows_out = rowsA * rowsB
-    cdef int cols_out = colsA * colsB
+    cdef int out_nnz = _safe_multiply(indptrA[rowsA], indptrB[rowsB])
+    cdef int rows_out = _safe_multiply(rowsA, rowsB)
+    cdef int cols_out = _safe_multiply(colsA, colsB)
 
     cdef CSR_Matrix out
     init_CSR(&out, out_nnz, rows_out, cols_out)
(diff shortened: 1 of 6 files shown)


jakelishman:
oh man, I don't know what I did locally to have my tests work but everything fail on Travis...

jakelishman:
Ok, the problem was a dumb change I made in the name of "neatness", and didn't test locally.  For some reason I decided to change my original `cdef int _safe_multiply(int a, int b) except -1` into `except 0`, completely eschewing 60 proud years of C tradition _and_ forgetting that 0 is a normal, valid return value and can't be the error condition.

qutip/cy/spmath.pyx
```diff
@@ -695,8 +695,9 @@ cdef bint _zcsr_isherm_full(
         free_CSR(&transpose)
 
 
+# Exception signal value must not be something we will ever want to return.
 @cython.overflowcheck(True)
-cdef _safe_multiply(int A, int B):
+cdef int _safe_multiply(int A, int B) except -1:
```
Ericgig, line 700: Using -1 means we can never use it for negative values. Since it's used to compute sizes it's not an issue, but why not use `except? -1`. It should not affect performance and since it's not clear otherwise that this function would not accept negative values, it feels safer.
Otherwise everything seems fine.

jakelishman: Great point - I'd got so used to using `except *` with objects and complex numbers I forgot `except?` existed.  This is absolutely the right way to do it.

PR #1441, opened 2021-02-16
Fix dims output from Qobj.eigenstates

jakelishman:
When called on a superoperator, the output from `Qobj.eigenstates()` would set the `dims` of the eigenstates to `[in_dims[0], [1, 1]]` rather than the correct operator-ket form `[in_dims[0], [1]]`.  This largely didn't have any knock-on effects because nothing ever used the second element, but it was still a little inconsistent with the rest of the library.


jakelishman:
@BoxiLi this is the single change I rebased out from the numpy fix

BoxiLi (approved):

PR #1445, opened 2021-02-17
Improve the noise module

BoxiLi:
**Description**
The internal processing of noise is simplified. instead of adding Pulse object to the pulse list, Noisy dynamics are added to a dummy pulse object called systematic noise.

Update the test. The test was not correctly collected by `pytest`, because the name was `Test` with capital T.

**Changelog**
Improve noise processing in qutip.qip

### DIFF
--- a/qutip/qip/noise.py
+++ b/qutip/qip/noise.py
@@ -45,22 +45,20 @@ def process_noise(pulses, noise_list, dims, t1=None, t2=None,
     noisy_pulses: list of :class:`qutip.qip.Pulse`
         The noisy pulses, including the system noise.
     """
-    if device_noise:  # first pulse independent noise
-        systematic_noise = Pulse(None, None, label="systematic_noise")
-        if (t1 is not None) or (t2 is not None):
-            RelaxationNoise(t1, t2).get_noisy_dynamics(dims, systematic_noise)
-
-        for noise in noise_list:
-            if isinstance(noise, (DecoherenceNoise, RelaxationNoise)):
-                noise.get_noisy_dynamics(dims, systematic_noise)
-
-    # then pulse-dependent noise
     noisy_pulses = deepcopy(pulses)
+    systematic_noise = Pulse(None, None, label="systematic_noise")
+
+    if (t1 is not None) or (t2 is not None):
+        noise_list.append(RelaxationNoise(t1, t2))
+
     for noise in noise_list:
-        if isinstance(noise, ControlAmpNoise):
-            noise.get_noisy_dynamics(noisy_pulses)
-        elif isinstance(noise, UserNoise):
-            noise.get_noisy_dynamics(noisy_pulses, systematic_noise, dims)
+        if isinstance(noise, (DecoherenceNoise, RelaxationNoise)) \
+                and not device_noise:
+            pass
+        else:
+            noisy_pulses, systematic_noise = noise._apply_noise(
+                dims=dims, pulses=noisy_pulses,
+                systematic_noise=systematic_noise)
 
     if device_noise:
         return noisy_pulses + [systematic_noise]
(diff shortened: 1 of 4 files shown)


qutip/qip/noise.py
```diff
@@ -338,9 +274,9 @@ def get_noisy_dynamics(self, dims):
                     T2_eff = 1./(1./t2-1./2./t1)
                 else:
                     T2_eff = t2
-                op = 1/np.sqrt(2*T2_eff) * sigmaz()
-                lindblad_noise.add_lindblad_noise(op, qu_ind, coeff=True)
-        return lindblad_noise
+                op = 1/np.sqrt(2*T2_eff) * 2 * num(dims[qu_ind])
```
jakelishman, line 277: `sigmaz()` and `2 * num(2)` aren't equivalent operators - I assume the constant offset doesn't matter so much, but `num` will also "swap" the effects on the two basis states?  compare `[[1, 0], [0, -1]]` to `[[0, 0], [0, 2]]`.  I don't know if that's a problem for you, though.

qutip/qip/noise.py
```diff
@@ -504,7 +433,7 @@ class UserNoise(Noise):
     def __init__(self):
         pass
 
-    def get_noisy_dynamics(self, pulses, dims):
+    def get_noisy_dynamics(self, pulses, systematic_noise, dims):
```
jakelishman, line 436: This doesn't match the signature of other `get_noisy_dynamics` calls in the inheritance chain.  If there's parsing to be done on all calls, you could always have `get_noisy_dynamics` to be defined in the base class, and ask subclasses to overide a related method `_get_noisy_dynamics`, which would have much stronger guarantees about exactly what will be passed in its arguments (or have them pass in that function to the `UserNoise` constructor so things are composed not overridden, but I just hate OO programming so ignore me).

qutip/qip/noise.py
```diff
@@ -43,83 +43,29 @@ def process_noise(pulses, noise_list, dims, t1=None, t2=None,
...
-def process_device_noise(noise_list, dims, t1=None, t2=None):
-    """
-    Apply pulse independent noise to the input list of pulses.
-    It does not modify the input
-    pulse, but return a new one containing the noise.
-    Pulse dependent noise will be ignored.
+    if device_noise:  # first pulse independent noise
+        systematic_noise = Pulse(None, None, label="systematic_noise")
+        if (t1 is not None) or (t2 is not None):
+            RelaxationNoise(t1, t2).get_noisy_dynamics(dims, systematic_noise)
```
jakelishman, line 51: Now I hate OO style overall, but this seems to break its general principles to me - shouldn't all subclasses of `Noise` have compatible call signatures?  Otherwise they're not polymorphic to each other at all, and the `Noise` class isn't correctly defining its interface.  Subclassing is meant to _avoid_ behaviour like
```
if isinstance(x, Subclass1):
    ...
elif isinstance(x, Subclass2):
    ...
elif isinstance(x, Subclass3):
    ...
```
by allowing objects to change their behaviour, but not their interface.

qutip/qip/noise.py
BoxiLi: You are right that it swaps the energy level. It will probably cause the state to rotate in a different direction? In my case, the measured T2 time is not changed since the tests pass. If one looks closer, there will be difference.

Since in qip the cavity is modelled with the number operator, I think `num` should be the more correct choice here.

qutip/qip/noise.py
BoxiLi: Yes, you are right, I should unify the signature of get_noisy_dynamics to be `def get_noisy_dynamics(self, pulses, systematic_noise, dims)`

qutip/qip/noise.py
jakelishman: ah ah, I was thinking about the second one as like an amplitude-damping term, where really it's a dephasing, right?  This is a long way out of my area of expertise, but if I remember correctly, the derivation of dephasing Kraus operators considers the effect of phase rotation by a normally-distributed angle, right?  If that's the case, then this is identical - the distribution is symmetric around the mean rotation of 0, so it doesn't matter which sign you define the operator with

PR #1447, opened 2021-02-19
fix the circuit latex generation for measurement and swap gate

BoxiLi:
**Description**
Wrong plot 
circuit_wrong

Correct plot now
circuit_correct

**Changelog**
Fix broken gate plotting

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -30,7 +30,6 @@
 #    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
-import copy
 from collections.abc import Iterable
 from itertools import product
 import numbers
@@ -1655,10 +1654,8 @@ class QubitCircuit:
                                 if _swap_processing:
                                     col.append(r" \qswap \qw")
                                     continue
-                                temp = copy.copy(gate.targets)
-                                temp.remove(n)
-                                second_targets = temp[0]
-                                distance = n - second_targets
+                                distance = abs(
+                                    gate.targets[1] - gate.targets[0])
                                 col.append(r" \qswap \qwx[%d] \qw" % distance)
                                 _swap_processing = True
 
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -30,7 +30,7 @@
 #    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
-
+import copy
 from collections.abc import Iterable
 from itertools import product
 import numbers
@@ -1645,13 +1645,22 @@ class QubitCircuit:
             if isinstance(op, Gate):
                 gate = op
                 col = []
+                _swap_processing = False
                 for n in range(self.N+self.num_cbits):
 
                     if gate.targets and n in gate.targets:
 
                         if len(gate.targets) > 1:
                             if gate.name == "SWAP":
-                                col.append(r" \qswap \qwx ")
+                                if _swap_processing:
+                                    col.append(r" \qswap \qw")
+                                    continue
+                                temp = copy.copy(gate.targets)
+                                temp.remove(n)
+                                second_targets = temp[0]
+                                distance = n - second_targets
+                                col.append(r" \qswap \qwx[%d] \qw" % distance)
+                                _swap_processing = True
 
                             elif ((self.reverse_states and
                                    n == max(gate.targets)) or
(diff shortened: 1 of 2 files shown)


jakelishman (commented):
This stateful looping and branching looks very weird to me, but I think it's because the whole loop looks inside-out - is there a reason the `for` loop is outside the `if gate.targets and n in gate.targets` etc?  Seems to me that the whole thing should first decide the kind of gate, then loop over only the necessary qubits - that'd be both more efficient and easier to read.

At any rate, that's much bigger task than is sensible now.  I don't see anything actually wrong in the code.

BoxiLi:
> Seems to me that the whole thing should first decide the kind of gate, then loop over only the necessary qubits - that'd be both more efficient and easier to read.

I feel the same. Indeed it is just something I encountered when writing the doc. I guess this structure to make use of similarity among those gates. Many have one control and one target. Those gates work well. SWAP just happens to not be one of this... Also, ISWAP doesn't work very well if it acts on qubits not next to each other. The gates need to be classified better. Maybe completely rewrite it at some point.

jakelishman:
we shouldn't really be blasé about codeclimate issues (I don't accept them on bits of `qutip.core` at all 😉), particularly since this is about function complexity, but really the complexity problem is from before, and it's a minor loophole that has meant it only got flagged now

BoxiLi:
It is a bit funny that the code climate says that the old complexity issue is fixed and there is a new complexity issue.

Somehow the complexity restriction is a bit too tight (~20 now, it used to be 10). Any algorithm that is really doing some hard work will exceed it. I don't think splitting is always a good choice...

jakelishman:
I think the file moved?  CodeClimate tracks by file, but doesn't detect renames.

Not sure I agree on the complexity issues...  If a function is so complex that CodeClimate is complaining about it, it's almost certainly trying to do too much in the body without separating it.  CodeClimate really heavily penalises `if` statements as it's difficult for a human to maintain a complex branching algorithm in their heads without making a mistake.  I don't know about you, but in the places in `qutip` where the `if/elif/else` branches are nested four+ deep, I'm completely lost and I have to have 2 views opened onto the same file just to keep referring between.

In this bit that it's complaining about, the `if` it's complaining about is nested 5 deep (not to mention the 2 for loops it's inside), and the second-level `if` body is so long I can't fit it all on my screen (50+ lines!).  When it gets to the `else` I can't even see what that `else` refers to!

If we look at functions CodeClimate complains about in `qip.circuit`, the complexity is mostly because the monolithic `Gate` class contains behaviour and data about _every_ possible gate.  There's no encapsulation anywhere, so every function that takes a `Gate` has to know all the different possible gates.  The `Gate` class is not actually providing any benefit at all - everything is purely done by global lookup table (in the ok cases) or 50-branch `if/elif` (in the bad cases) based on the name.  `QubitCircuit` shouldn't need the million `_gate_resolved` functions - if there's a `Gate` class, instances of that (whether done by subclass or composition) should be supplying it (though the lookup table is a good choice under the circumstances).

Given how `circuit.py` currently organises its data, it feels like you have to introduce some seemingly arbitrary splits to keep CodeClimate happy, but I'd argue the problem is the data structure, and CodeClimate is right to complain about the complexity.  In stuff like compilers, there's a reason code compilers do parsing->verification->optimisation, with several passes in the last one, and transform things internally into a very strict intermediate representation at every step.

`Qobj._repr_latex_` was like this - the `master` version is claimed to be like "78" complexity, but I'd be prepared to bet CodeClimate would have the `dev.major` version less than 10.  The change is that the `master` version basically reimplements the same algorithm 4 times with complex, split up for loops (not to mention that for some reason the alignment and new-line characters are considered part of formatting a number?).  The `dev.major` version first decides if it's going to truncate the rows and/or columns, then it just calls a formatter on each row it decided it wants rendering with a special value to say when it should output a truncation character, and then joins the rows together.

Obviously it's not _always_ the data structure - e.g. in `propagator` the branching to choose an evaluation method is fine, but once it's chosen, it should delegate to specialised methods.  That way, the maintainer can verify the branching paths with a nice top-down overview without 50 lines of specialised logic inbetween each one, then you can verify each specialised path independently.

jakelishman:
actually, just realised my fork is configured to do CodeClimate on `dev.major`.  For say `qobj.py`, the `master` version has [maintainability F](https://codeclimate.com/github/qutip/qutip/qutip/qobj.py), while `dev.major` has [maintainability A](https://codeclimate.com/github/jakelishman/qutip/qutip/core/qobj.py) with the only issues being TODOs and like 5 pep8s.

BoxiLi:
Oh, I completely agree with the argument nested loops/if/else. I was more talking about the logistic depth requirement of code climate. I just realized that is not the climate issue here.

But, indeed `circuit.py` is quite messy. And the Gate class is more or less just a place holder. A refactoring to define each gate as a subclass should make things much easier. Also for this latex method. I like it very much what is going on in `dev.major`!

Tbh the millions of `_gate_resolved` functions are actually already after a short refactoring last year by a brave volunteer. It used to be 400-line functions. That was horrible. Just like the `adjacent_gates` method in `spinchain.py` (another F). I haven't dared to touch it...

> For say qobj.py, the master version has maintainability F, while `dev.major` has maintainability A with the only issues being TODOs and like 5 pep8s.

That's amazing! When I joined QuTiP, the overall maintainability was D, as I remember. Now it is B. I'm keen to see it becomes A after `dev.major` got merged :)

PR #1448, opened 2021-02-19
QuTiP 4.5.3

jakelishman:
Boxi and I decided to hold off on 4.6.0 until after the next admin meeting - it was too rushed to finish off all the features, and there are deprecations to be discussed.

This is a proposed patch release on the 4.5.X branch that purely adds the fixes for NumPy 1.20.  The base of this PR is the code state at 4.5.2, so this PR is a complete list of changes from 4.5.2 to 4.5.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ _stage_linux_38: &stage_linux_38
     - conda info -a
     - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
     - source activate test-environment
-    - conda install numpy scipy pytest pytest-cov cython coveralls
+    - conda install 'numpy>=1.16.6,<1.20' scipy pytest pytest-cov cython coveralls
     - python setup.py install
 
 
@@ -74,8 +74,7 @@ _stage_linux_no_cython: &stage_linux_no_cython
     - conda info -a
     - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
     - source activate test-environment
-    - conda install -c conda-forge nomkl blas=*=openblas numpy scipy pytest pytest-cov
-    - conda install -c conda-forge cython --freeze-installed
+    - conda install blas=*=openblas 'numpy>=1.16.6,<1.20' scipy pytest pytest-cov cython
     - conda list
     - python setup.py install
     - conda uninstall cython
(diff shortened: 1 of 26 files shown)


jakelishman:
Just fixing the tests - the lattice module got deleted in a separate feature on `master` before the Numpy 1.20 changes.  My mistake for not running them before making the PR.

jakelishman:
The test failure on Python 3.6 appears to be due to some exceptionally weird behaviour by `conda` in the Travis setup - if you look closely, you can see that when it runs `conda uninstall cython`, it _downgrades_ `numpy` to 1.14.2.  This would not be a problem in a `conda-forge` release build, because the `numpy` API dependency is pinned in the feedstock, but it is a potential problem on `pip` source builds.  It's possible this super odd behaviour of `conda` is the cause of people's complaints on the email lists about failed imports.

I will push a change to the requirements here to try and enforce `numpy >= 1.16.6`, which ought to solve the dependency issues.

BoxiLi:
> you can see that when it runs conda uninstall cython, it downgrades numpy to 1.14.2

This is really strange.

> This would not be a problem in a conda-forge release build, because the numpy API dependency is pinned in the feedstock, but it is a potential problem on pip source builds.

I was setting up a GitHub Action test for `qutip_qip` https://github.com/qutip/qutip-qip/pull/6. `pip installation` works well there, but it could be that I don't use any run-time compiling.

BoxiLi (approved):

jakelishman:
Yeah, `qutip-qip` is a much easier build and installation process because you don't need to be concerned about compiler infrastructure, ABI compatibility or stuff like that.  In addition in that particular test run, there were three individual calls to change the installed packages, which meant plenty of space for stuff to go wrong.

In this case, I think it was a strange interaction between the old `.travis.yml` specifically installing `numpy` and all other dependencies from `conda-forge`, without the channel being in the defaults list.  Since `conda uninstall` doesn't specify an additional channel, the dependency calculator considers `defaults` to be higher priority, so it sets various packages to arbitrary allowed ones from `defaults`, which with `openblas` installed like this meant numpy `1.14`.

That's my guess at least - it's a pretty weird set up circumstances, but the new `.travis.yml` file that's already in `master` is far simpler about the build process.

Ericgig (approved):

PR #1453, opened 2021-02-24
Merge master into dev.major

jakelishman:
Brings `dev.major` up-to-date with the current state of master (f995f6f6).

Includes nontrivial additional changes to resolve merge conflicts in some nonobvious files, due to the files moving place:

|new file                  | old file |
|--------------------------|------------------------|
|core/qobj.py              | qobj.py   (`eigenstates`, `__array_ufunc__`) |
|core/data/kron.pyx        | cy/spmath.pyx   (`kron`) |
|core/data/ptrace.pyx      | qobj.py   (`ptrace`) |
|core/data/properties.pyx  | cy/spmath.pyx   (`isherm`) |
|tests/test_device.py      | tests/test_cavityqed.py |
|tests/test_device.py      | tests/test_spinchain.py |

The merging of PRs gh-1289 and gh-1337 (overhauling the settings and reset handling on master and dev.major respectively) is handled by taking only the results of gh-1337; in terms of the "setting reset" functionality, these two PRs were equivalent but a huge merge-conflict because they achieved the same thing in two completely separate manners.


pyproject.toml
```diff
@@ -1,2 +1,8 @@
 [build-system]
-requires = ["setuptools", "wheel", "Cython", "numpy", "scipy"]
+requires = [
+    "setuptools",
+    "wheel",
+    "cython>=0.29.20",
+    "numpy>=1.16.6,<1.20",
```
Ericgig, line 6: Doesn't this exclude numpy  1.20?

jakelishman: Yes, but only from the system doing the building.  Let's say numpy 1.20 is installed, then a user does `pip install qutip` - `pip` will recognise that the numpy requirement is not correctly satisfied for the build, but it is for installation, so it'll temporarily install `numpy==1.19.2` (probably) into a venv, do the build within that, then drop `qutip` into the main environment

jakelishman: cf https://github.com/qutip/qutip/pull/1429/files#r582891252

qutip/core/metrics.py
```diff
@@ -75,9 +75,7 @@ def fidelity(A, B):
     --------
     >>> x = fock_dm(5,3)
     >>> y = coherent_dm(5,1)
-    >>> fidelity(x,y)
-    0.24104350624628332
-
+    >>> np.testing.assert_almost_equal(fidelity(x,y), 0.24104350624628332)
```
Ericgig, line 78: Having doctest like this for only a few functions does not feel right...  
We should think about it when working on the docs.

Ericgig (approved):

qutip/core/metrics.py
jakelishman: Yeah, it feels like maybe someone got things mixed up when they wrote this doctest - surely the purpose of doctests are meant to check the existing examples make sense, rather than obfuscating the example?  We'll have to work out how to do approximate numeric comparisons, I suppose.

PR #1461, opened 2021-03-11
Core update for solvers.

Ericgig:
**Description**
As proposed by @hodgestar, I am splitting #1409 into smaller PRs. This is part 1/3, but contain some change in core that would come later in the Solver PR:

- Add manual cython dispatched function. Used by vern ODE method.

- Add `to` to `QobjEvoFunc` and some check. Used by brmesolve rework.

- Add a version to string Coefficient, a 'try_parsing' options, partial arguments update and basic cleaning function.
This comes mostly from talk with @jakelishman.
When updating coefficient with new argument, it no longer require all coefficient to be given, but will simply update those available. 
Since those change must overwrite old version of the coefficient, a version tag is added to the coefficient folder and a function to erase the old version. It use it's own `COEFF_VERSION`, not qutip version since I don't expect most qutip update will not touch `coefficient.pyx` so we need not to force user to recompile. But it raise the risk of forgetting to change it when working on them. 
Lastly, I added an option to disable the string coefficient parsing since there seems to have some reservation in the last meeting.
But there is a weakness when using that form if unused args are passed: the `key = args['key']` code could be generated for those extra `args` but will not affect the hash name. 

**Related issues or PRs**
Will replace part of #1409.

### DIFF
--- a/qutip/core/__init__.py
+++ b/qutip/core/__init__.py
@@ -3,6 +3,7 @@ from .coefficient import *
 from .interpolate import *
 from .qobj import *
 from .qobjevo import *
+from .qobjevofunc import *
 from .expect import *
 from .tensor import *
 from .states import *
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -442,11 +442,6 @@ def compile_code(code, file_name, parsed, c_opt):
             raise Exception("Could not compile") from e
         finally:
             sys.argv = oldargs
-        try:
-            libfile = glob.glob(file_name + "*")[0]
-            shutil.move(libfile, os.path.join(root, libfile))
-        except Exception:
-            warn("File")
     finally:
         os.chdir(pwd)
     return try_import(file_name, parsed)
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -9,6 +9,7 @@ import glob
 import importlib
 import shutil
 import numbers
+from contextlib import contextmanager
 from collections import defaultdict
 from setuptools import setup, Extension
 try:
@@ -321,14 +322,13 @@ def coeff_from_str(base, args, args_ctypes, compile_opt):
     hash_ = hashlib.sha256(bytes(parsed, encoding='utf8'))
     file_name = "qtcoeff_" + hash_.hexdigest()[:30]
     # See if it already exist, if not write and cythonize it
-    coeff, file_name = try_import(file_name, parsed)
+    coeff = try_import(file_name, parsed)
     if coeff is None or compile_opt['recompile']:
         code = make_cy_code(parsed, variables, constants,
                             raw, compile_opt)
         coeff = compile_code(code, file_name, parsed, compile_opt)
     keys = [key for _, key, _ in variables]
     const = [fromstr(val) for _, val, _ in constants]
-    print(keys, variables)
     return coeff(base, keys, const, args)
 
 
@@ -342,31 +342,14 @@ def try_import(file_name, parsed_in):
         mod = importlib.import_module(file_name)
     except ModuleNotFoundError:
         # Coefficient does not exist, to compile as file_name
-        return None, file_name
+        return None
 
     if mod.parsed_code == parsed_in:
         # Coefficient found!
-        return mod.StrCoefficient, file_name
-
-    # hash collision!
-    # Increment filename
-    tries = -1
-    file_name_ = file_name
-    while mod.parsed_code != parsed_in and tries < 10:
-        tries += 1
-        file_name_ = file_name + str(tries)
-        try:
-            mod = importlib.import_module(file_name_)
-        except ModuleNotFoundError:
-            # Coefficient does not exist, to compile as file_name_
-            return None, file_name_
-
-    if mod.parsed_code != parsed_in:
-        # At least 10 coeff with the same hash!? Giveup...
+        return mod.StrCoefficient
+    else:
         raise ValueError("string hash collision, change the string "
                          "or clean files in qutip.settings.install['tmproot']")
-    else:
-        return mod.StrCoefficient, file_name
 
 
 def make_cy_code(code, variables, constants, raw, compile_opt):
@@ -436,34 +419,37 @@ cdef class StrCoefficient(Coefficient):
 
 
 def compile_code(code, file_name, parsed, c_opt):
-    root = get_root()
     pwd = os.getcwd()
-    os.chdir(root)
-    [os.remove(file) for file in glob.glob(file_name + "*")]
-    full_file_name = os.path.join(root, file_name)
-    file_ = open(full_file_name + ".pyx", "w")
-    file_.writelines(code)
-    file_.close()
-    oldargs = sys.argv
+    root = get_root()
     try:
-        sys.argv = ["setup.py", "build_ext", "--inplace"]
-        coeff_file = Extension(file_name,
-                               sources=[full_file_name + ".pyx"],
-                               extra_compile_args=c_opt['compiler_flags'].split(),
-                               extra_link_args=c_opt['link_flags'].split(),
-                               include_dirs=[np.get_include()],
-                               language='c++')
-        setup(ext_modules=cythonize(coeff_file, force=c_opt['recompile']))
-    except Exception as e:
-        raise Exception("Could not compile") from e
-    try:
-        libfile = glob.glob(file_name + "*")[0]
-        shutil.move(libfile, os.path.join(root, libfile))
-    except Exception:
-        warn("File")
-    sys.argv = oldargs
-    os.chdir(pwd)
-    return try_import(file_name, parsed)[0]
+        os.chdir(root)
+        [os.remove(file) for file in glob.glob(file_name + "*")]
+        full_file_name = os.path.join(root, file_name)
+        file_ = open(full_file_name + ".pyx", "w")
+        file_.writelines(code)
+        file_.close()
+        oldargs = sys.argv
+        try:
+            sys.argv = ["setup.py", "build_ext", "--inplace"]
+            coeff_file = Extension(file_name,
+                                   sources=[full_file_name + ".pyx"],
+                                   extra_compile_args=c_opt['compiler_flags'].split(),
+                                   extra_link_args=c_opt['link_flags'].split(),
+                                   include_dirs=[np.get_include()],
+                                   language='c++')
+            setup(ext_modules=cythonize(coeff_file, force=c_opt['recompile']))
+        except Exception as e:
+            raise Exception("Could not compile") from e
+        finally:
+            sys.argv = oldargs
+        try:
+            libfile = glob.glob(file_name + "*")[0]
+            shutil.move(libfile, os.path.join(root, libfile))
+        except Exception:
+            warn("File")
+    finally:
+        os.chdir(pwd)
+    return try_import(file_name, parsed)
 
 
 # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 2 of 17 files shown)


qutip/tests/core/test_coefficient.py
```diff
@@ -46,6 +52,10 @@ def g(t, args):
     return np.cos(args["w"] * t * np.pi)
 
 
+def h(t, args):
+    return args["a"] + args["b"] +t
```
hodgestar, line 56: Nitpick: Missing space between `+` and `t`.

qutip/tests/core/test_coefficient.py
```diff
@@ -116,10 +126,25 @@ def test_CoeffCallArgs(base, kwargs, tol):
...
     assert np.allclose(coeff(t, {"w": w}), val, rtol=tol)
 
 
+@pytest.mark.parametrize(['base', 'tol'], [
+    pytest.param(h, 1e-10, id="func"),
+    pytest.param("a + b + t", 1e-10, id="string")
+])
+def test_CoeffCallArguments(base, tol):
+    # Partial args update
+    t = np.random.rand() * 0.9 + 0.05
```
hodgestar, line 139: We should just us fixed test cases here instead of random numbers, I think. If we use random numbers the test might just pass or fail randomly which is bad, and it doesn't really buy us anything over picking some specific numbers I don't think. If we want to test a range of different numbers, we should explicitly test the numbers we want to.

As an aside, I'd like to add hypothesis to qutip tests at some point, especially stuff in the core, but that's a bit of a rabbit hole, so I'll leave it until after I've started and version 5 has settled down a bit more.

hodgestar (commented):
Thanks for breaking these out -- it was much easier to review this bit by itself.

I've left a few comments. I think Jake should weigh in too on these core things. One big picture concern is that quite a few changes have been made to the core here, but there are comparatively few tests that have been added or changed. We probably want to test these core things quite thoroughly if we can, since all of the rest of QuTiP relies on them.

qutip/core/cy/cqobjevo.pyx
```diff
@@ -128,12 +128,15 @@ cdef class CQobjEvo:
...
-        if isinstance(matrix, Dense):
-            return self.matmul_dense(t, matrix)
+        if type(matrix) is Dense and (out is None or type(out) is Dense):
+            return self.matmul_dense(t, matrix, out)
         self._factor(t)
-        out = _data.matmul(self.constant, matrix)
+        if out is None:
+            out = _data.matmul(self.constant, matrix)
+        else:
+            out = _data.add(out, _data.matmul(self.constant, matrix))
```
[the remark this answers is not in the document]
Ericgig: It to not allocate memory when not needed.
`out = add(out, matmul(mat, vec))` (iadd is only available to Dense) would need to allocate 2 vectors for the operation, but none. In some case, I have seen this allocation being up to 30% of the time of the operation. 

We can work in-place for `Dense`'s `matmul`, but not `CSR`. So this method allow to skip any memory allocation for `Dense` and have a matching behaviour for other data type.

qutip/core/coefficient.py
```diff
@@ -273,30 +321,52 @@ def coeff_from_str(base, args, args_ctypes, compile_opt):
...
+
+    if mod.parsed_code == parsed_in:
+        # Coefficient found!
+        return mod.StrCoefficient, file_name
+
+    # hash collision!
+    # Increment filename
+    tries = -1
+    file_name_ = file_name
+    while mod.parsed_code != parsed_in and tries < 10:
```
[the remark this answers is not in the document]
Ericgig: Good point, if the hash is not sufficient and we don't support infinite hash collision like `dict`, we should just quickly fail instead of trying to postpone. Expecting about 10000 coefficients used per users, 30 hex digit should be more than enough.

qutip/core/coefficient.py
```diff
@@ -389,7 +463,7 @@ def compile_code(code, file_name, parsed, c_opt):
         warn("File")
     sys.argv = oldargs
     os.chdir(pwd)
```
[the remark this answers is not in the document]
Ericgig: Let's just fix it now.

qutip/core/cy/cqobjevo.pyx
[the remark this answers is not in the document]
jakelishman: I'm also going to have a think about re-using allocation space within the `CSR` type for `add` and `matmul`.  I've got some general ideas with a two-part swapping buffer that's half fleshed out in my head at the moment.

qutip/core/cy/cqobjevo.pyx
```diff
@@ -216,19 +219,22 @@ cdef class CQobjFunc(CQobjEvo):
     def call(self, double t, int data=0):
         return self.base(t, data=data)
```
jakelishman, line 220: All these calls happen with Python semantics - is that intended?

qutip/installsettings.py
```diff
@@ -58,18 +58,12 @@ class InstallSettings:
         _logger = None
 
     try:
-        qutip_conf_dir = os.path.join(os.path.expanduser("~"), '.qutip')
-        if not os.path.exists(qutip_conf_dir):
-            os.mkdir(qutip_conf_dir)
-        tmproot = os.path.join(qutip_conf_dir, 'coeffs')
+        tmproot = os.path.join(os.path.expanduser("~"), '.qutip')
```
jakelishman, line 61: This is safe on Windows, right?

qutip/tests/core/test_coefficient.py
jakelishman: Nathan asked me about this around the time I first introduced pytest, but I didn't know how to do it at the time.  I was basically learning how to do _any_ unit testing (no formal training strikes again!).  Seems like a good idea to me in general - random testing has bitten us a few times, although I did make some efforts to try and make sure that pytest reported _which_ random values were used so it could be reproduced.

jakelishman (dismissed):
LGTM in general - one comment about some questionable error handling, and a couple of questions about stuff I wasn't sure.

qutip/core/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .interpolate import *
 from .qobj import *
 from .qobjevo import *
+from .qobjevofunc import *
```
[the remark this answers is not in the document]
Ericgig: I am thinking of removing `QobjEvoFunc` in the close future. But otherwise it can be as visible as `QobjEvo`.

qutip/core/cy/cqobjevo.pyx
Ericgig: Could be made cpdef, but `self.base` is a python object so it probably have only a small impact.
And it won't survive the merger of QobjEvo and cQobjEvo.

qutip/core/coefficient.py
```diff
@@ -363,32 +419,36 @@ def optstr(self):
...
+            setup(ext_modules=cythonize(coeff_file, force=c_opt['recompile']))
+        except Exception as e:
+            raise Exception("Could not compile") from e
+        finally:
+            sys.argv = oldargs
+        try:
+            libfile = glob.glob(file_name + "*")[0]
+            shutil.move(libfile, os.path.join(root, libfile))
+        except Exception:
+            warn("File")
```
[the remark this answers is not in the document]
Ericgig: Yeah.. and we are moving from root to root... The whole block does nothing.

Ericgig:
@hodgestar Could you tell me if I have answered comment and approve this?

qutip/core/cy/cqobjevo.pyx
jakelishman: Yeah, the `self.base` was the bit I was asking about.  As long as it's expected that it's all Python semantics, that's fine.  I forgot which bits were meant to be fastcall and which weren't.

qutip/core/__init__.py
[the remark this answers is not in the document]
Ericgig: I wrote my plans for `QobjEvoFunc` in #1468.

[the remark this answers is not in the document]
jakelishman: Ok, I'm almost always in favour of simplifying public interfaces (Y)

qutip/installsettings.py
hodgestar: https://docs.python.org/3/library/os.path.html#os.path.expanduser is documented as supporting Windows.

qutip/core/cy/cqobjevo.pyx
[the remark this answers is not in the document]
hodgestar: My point is that in this line:
```python
out = _data.add(out, _data.matmul(self.constant, matrix))
```
the `_data.matmul(self.constant, matrix)` part appears to already creates a temporary object, so we appear to not have saved any temporary objects over just returning `out` and allowing the caller to do `caller_out = _data.add(caller_out, out)` themselves (note that I don't propose having the caller call `_data.add` directly -- rather that they call it via some friendlier interface).

Do you agree? Or did I miss something about how this works?

qutip/tests/core/test_coefficient.py
```diff
@@ -34,8 +34,15 @@
...
+from functools import partial
 from qutip.core.coefficient import (coefficient, norm, conj, shift,
-                                    reduce, CompilationOptions)
+                                    reduce, CompilationOptions,
+                                    clean_compiled_coefficient
+                                   )
+
+
+# Ensure the latest version is tested
+clean_compiled_coefficient(True)
```
hodgestar, line 45: We shouldn't call clean up on import (otherwise any code that imports this test file will clear the cache). Instead we could add it to a `setup_function` function (see https://docs.pytest.org/en/stable/xunit_setup.html). This would also mean that the cache would be cleared between each test which would improve the isolation of the tests from each other.

hodgestar (approved):
@Ericgig Thank you for addressing and replying to my previous comments. I left two more, but I think the overall shape of this PR is good, so I'm approving it. Happy for you to merge it once you're happy.

jakelishman:
That test failure actually looks like a serious problem - there's a `NaN` that has crept into an output after a `copy` operation.  In this case I think it's important that we really track down the root cause of how that happened, so we can be sure it can't happen again.

Ericgig:
I could not replicate the bug locally, but I have a good guess.
I think it's fixed in changing `t>` to `t>=` in inter.pyx.

Ericgig:
I rechecked the bug and it's really the case bad handling the limit of the range.
At `t=tlist[-1]`, it effectively did `y[len(y)-1] + 0*y[len(y)]` with boundary unchecked.

Ericgig:
@jakelishman If you are happy with the bug fix, can I merge this and the next PR and prepare the next batch?

jakelishman:
Oh, sorry I missed this.  The problem you found sounds pretty likely to be the cause.  Let's keep it in mind, just in case.  This looks fine to me to merge.

Let me know if you're waiting on me on other PRs, or poke me (via GitHub "request review" or by email) if in the future I'm not replying after a few days...  I haven't quite figured out what I want my GitHub notification settings to be yet in order to prioritise the most important things.

jakelishman (approved):

qutip/tests/core/test_coefficient.py
```diff
@@ -46,6 +53,23 @@ def g(t, args):
     return np.cos(args["w"] * t * np.pi)
 
 
+def h(t, args):
+    return args["a"] + args["b"] + t
+
+
+def _assert_eq_over_interval(coeff1, coeff2, rtol=1e-12, inside=False):
+    "assert coeff1 == coeff2"
+    "inside refer to the range covered by tlistlog: [0.01, 1]"
```
hodgestar, line 62: Belated review: Probably these lines should be
```python
def _assert_eq_over_interval(...):
    """ assert coeff1 == coeff2
         inside refer to the range covered by tlistlog: [0.01, 1]
    """
```
or
```python
def _assert_eq_over_interval(...):
    """ assert coeff1 == coeff2 """
    # inside refer to the range covered by tlistlog: [0.01, 1]
```
As things currently stand the docstring is `"assert coeff1 == coeff2"` and then the `"inside ..."` is just a statement on a line by itself.

hodgestar:
@Ericgig Thanks again for splitting this out. It was a big help. Apologies for being away for the last few commits.

Issue #1460, opened 2021-03-10
correlation_2op_1t() and spectrum() giving the physically incorrect result unless solver='es'


gsteele13:
**Describe the bug**
See discussion here:

https://groups.google.com/g/qutip/c/2lod1DuFZiE/m/y8Bj6dKHDAAJ?pli=1

To my surprise, my PhD student had this problem again today, I am a bit surprised nothing has been done about this. 

At least making `es` the default solver for correlation and spectrum, updating the docs to give a warning to only use the es solver (for now?), and maybe longer term figuring out why the other solvers are not handling it properly

**To Reproduce**

https://github.com/gsteele13/gary-exploring-qutip/blob/master/Correlations.ipynb

```
import numpy as np
import pylab as plt
from qutip import *

N = 5
w = 1 * 2 * np.pi              # 1 Hz oscillator frequency
Q = 40
kappa = w / Q

# Set up some operators
a = destroy(N)
n = num(N)
x = (a + a.dag())/np.sqrt(2)
p = -1j * (a - a.dag())/np.sqrt(2)
H = w * a.dag() * a

c_ops = []

psi0 = fock(N,0)
tlist = np.linspace(0, 10, 500)

corr = correlation_2op_1t(H, psi0, tlist, c_ops, x, x, solver='me')

wlist1, spec1 = spectrum_correlation_fft(tlist, corr)

plt.figure(figsize=(14,8))
plt.plot(tlist, np.real(corr))
plt.plot(tlist, np.imag(corr))
plt.show()

plt.figure(figsize=(14,8))
plt.plot(wlist1, np.abs(spec1))
plt.show()
```

**Expected behavior**
Correlation should have an imaginary component and quantum PSD should not be symmetric about f = 0

**Your Environment**
Software	Version
QuTiP	4.4.1
Numpy	1.17.4
SciPy	1.4.1
matplotlib	3.3.4
Cython	0.29.14
Number of CPUs	4
BLAS Info	INTEL MKL
IPython	7.9.0
Python	3.7.5 (default, Oct 25 2019, 10:52:18) [Clang 4.0.1 (tags/RELEASE_401/final)]
OS	posix [darwin]
Wed Mar 10 08:51:15 2021 CET

**Additional context**

nwlambert:
hi gary,  thanks for making this an issue, i think it fell through the cracks on the google groups.  

[...]

Issue #1437, opened 2021-02-08
Turning off `store_state` in mesolve


BoxiLi:
**Describe the issue**
In `mesolve`, all the intermediate result for each `t` in the `tlist` will be saved. If the given `tlist` is very long, this can be a bottleneck for time and memory. In the `Option` class, there is an option `store_states`, but it does not work as I expected. Setting it to False does not turn this off.

https://github.com/qutip/qutip/blob/7a73b136cb74f832cf19f15e93c7268244674bdb/qutip/mesolve.py#L461-L463

According to the code above, there is no way to completely turn this off. One has to either save expectation values or save the intermediate states.

**To Reproduce**
```
from qutip import *
import numpy as np

options = Options()
options.store_states = False
print(options.store_states)
result = mesolve(sigmax(), basis(2,0), np.linspace(0,1,10), options=options)
print(options.store_states)
```
gives
```
False
True
```

@Ericgig @jakelishman

Ericgig:
This is expected. I don't see any reason to compute the state at a time if you don't want the state or and expectation value... If you only want the last state use tlist=[0, t_last]. Also `tlist` does not have to be a linear space, if you want to save only the few last values you can do: `tlist=[0,990,1000]`, etc.

If you need to use an array coefficient and this is your restriction on `tlist`, use `qutip.Cubic_spline` (or build the `QobjEvo` before the solver, but I don't remember if it is safe in v4).

jakelishman:
We probably shouldn't mutate the given `Options` object, but the call as written here doesn't make a huge amount of sense - why would you want to run the integrator but not have any output stored at all?  You'd just be spinning your processor.

That said, if we're going to do something different to what the user requested we should at least emit a warning.  There's also interplay with `store_final_state` here too; it's questionable technique to do `store_states=False, e_ops=None; store_final_state=True` with a `tlist` with more than two elements, but probably people who aren't familiar with `nsteps` will do something like that, and it seems a bit picky to prevent/warn on that when the intention seems clear.  I think that the current behaviour doesn't take this case into account when falling back - it'll store all states, which in the 4.x branch is a fairly major time sink (but much much faster in 5.0).

What behaviour would you expect for your use case?  I'd probably do something like:
1. when we change the `Options`, we have to clone it first to prevent leaking our mutations
2. if we're not going to save anything, we should warn the user but then do it anyway - maybe their time-dependence functions leak state, and they're just doing something weird with that
[...]

Issue #1238, opened 2020-04-27
qutip.Result capturing output from e_ops as custom callback function


marekyggdrasil:
**Is your feature request related to a problem? Please describe.**

It is very convenient to take measurements in QuTip by passing list of observables in the `e_ops` argument of `mesolve` or `sesolve`, a list of expectation values at each time step of all those operators will be provided in returned `qutip.Result` object.

It becomes much less elegant once we implement custom measurements and pass a callback function (for example to work with time-dependent Hamiltonians etc).

According to the [documentation](http://qutip.org/docs/4.1/apidoc/functions.html#module-qutip.mesolve)

> An instance of the class qutip.Result, which contains either an array result.expect of expectation values for the times specified by tlist, (...), or nothing if a callback function was given in place of operators for which to calculate the expectation values.

So to work with callback function, we need to manage the storage by ourselves, for example by defining a global list, as it is done in [this example](https://nbviewer.jupyter.org/github/jrjohansson/qutip-lectures/blob/master/Lecture-8-Adiabatic-quantum-computing.ipynb#Evolve-the-system-in-time).

**Describe the solution you'd like**

Currently when we pass a list of operators such as

```python
e_ops = [sigmax(), sigmay(), sigmaz()]
result = mesolve(H, rho0, e_ops=e_ops)
```

We will be able to access expectation value of `sigmax()` in `result.expect[0]`, of `sigmay()` in `result.expect[1]` and of course `sigmaz()` in `result.expect[2]`.

Why not capture the custom measurement in similar way?

```
def customMeasurement(tau, psi):
    exp_x = expect(sigmax(), psi)
    exp_y = expect(sigmaz(), psi)
    exp_z = expect(sigmay(), psi)
    energy = expect(H, psi)
    fidelity = np.abs(psi_ref.overlap(psi))**2.
    return [exp_x, exp_y, exp_z, energy, fidelity]
```

Now it is possible to still get expectation values as before, but in addition we can take more custom measurements (in this example energy and fidelity of state) and access them in exactly the same way for the convenience. The only thing user needs to keep track of is order of operators returned from the callback function.

So now it is as before but in addition `result.expect[3]` contains energy and `result.expect[4]` contains fidelity and no custom managed global external lists required as it was in the [example](https://nbviewer.jupyter.org/github/jrjohansson/qutip-lectures/blob/master/Lecture-8-Adiabatic-quantum-computing.ipynb#Evolve-the-system-in-time)

**Describe alternatives you've considered**

[...]

PR #1092, opened 2019-10-05
Use loky-based parallel_map

goerz:
[WIP]

PR #1462, opened 2021-03-11
Support class for solvers

Ericgig:
**Description**
Splitting #1409, part 2/3.
Contain the non Integrator/Evoler part of the PR.

##### Options
Add move some options around and add a class for ODE options. 
parallel_map and progress_bar will become options, not arguments to the `solve` function.
`store_states` default is None so that `store_states=False` is not overwritten when no e_ops given. (#1437)

##### Result
With this version, the `Result` object is responsible to know if states are stored and to compute expectation values, in the method `add(t, state: Qobj)`. The solver will be responsible to create the `Qobj` for the state.
(#1460): The type of the expectation value [float, complex] is set by the `expect` function, `QobjEvo.expect` or user provided expect function. 
(#1238): Multiple callback function can be given as expectation: `e_ops=expect_func`, `e_ops=[expect_func1, expect_func2]`, are both valid.
`mcsolve` (and probably stochactic) use `MultiTrajResult` or `MultiTrajResultAveraged`. They have both `average_state` and `run_states` instead of `states` which content change depending on options. They can also compute the `photocurrent` from the collapse. (making photocurrent_sesolve useless).

##### parallel
Add a `reduce_func` which allows to treat a results without waiting for all task to be done. Presently even when using `average_states`, all states are saved and the average is only computed at the end, which is not the behaviour expected by the users. 
(#1092 ) Add a loky based parallel map. It does not pickle object the same way as multiprocessing.

##### progress bar
Add a bar based on `tqdm`, which is a standard progress bar in python.

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -190,6 +190,10 @@ class CompilationOptions:
     options = {
         # use cython for compiling string coefficient
         "use_cython": _use_cython,
+        # try to parse the string so qutip recognise similar string as one
+        # compiled coefficient:
+        # "a*t", "a * t", "b*t" would all use one compiled version
+        "try_parse": True,
         # In compiled Coefficient, are int kept as int?
         # None indicate to look for list subscription
         "accept_int": None,
@@ -303,6 +307,7 @@ def coeff_from_str(base, args, args_ctypes, compile_opt):
         coeff = compile_code(code, file_name, parsed, compile_opt)
     keys = [key for _, key, _ in variables]
     const = [fromstr(val) for _, val, _ in constants]
+    print(keys, variables)
     return coeff(base, keys, const, args)
 
 
@@ -334,7 +339,13 @@ def try_import(file_name, parsed_in):
         except ModuleNotFoundError:
             # Coefficient does not exist, to compile as file_name_
             return None, file_name_
-    return mod.StrCoefficient, file_name_
+
+    if mod.parsed_code != parsed_in:
+        # At least 10 coeff with the same hash!?
+        # Giveup
+        raise ValueError("string hash collision, change the string or ")
+    else:
+        return mod.StrCoefficient, file_name
 
 
 def make_cy_code(code, variables, constants, raw, compile_opt):
@@ -592,32 +603,38 @@ def parse(code, args, compile_opt):
     return code, variables, ordered_constants
 
 
-def use_hinted_type(var_tuple, ncode, args_ctypes):
-    name, key, type_ = var_tuple
-    if key in args_ctypes:
-        code = code.replace(cte, cte[0] + name, 1)
+def use_hinted_type(variables, code, args_ctypes):
+    variables_manually_typed = []
+    for i, (name, key, type_) in enumerate(variables):
+        if key in args_ctypes:
+            new_name = "self._custom_" + args_ctypes[key] + str(i)
+            code = code.replace(name, new_name)
+            variables_manually_typed.append((new_name, key, args_ctypes[key]))
+        else:
+            variables_manually_typed.append((name, key, type_))
+    return code, variables_manually_typed
 
 
 def try_parse(code, args, args_ctypes, compile_opt):
     """
     Try to parse and verify that the result is still usable.
     """
+    if not compile_opt['try_parse']:
+        variables = [("self." + name, name, "object") for name in args]
+        code, variables = use_hinted_type(variables, code, args_ctypes)
+        return code, variables, [], True
     ncode, variables, constants = parse(code, args, compile_opt)
     if compile_opt['no_types']:
         # Fallback to all object
         variables = [(f, s, "object") for f, s, _ in variables]
         constants = [(f, s, "object") for f, s, _ in constants]
-    variables_manually_typed = []
-    for i, (name, key, type_) in enumerate(variables):
-        if key in args_ctypes:
-            new_name = "self._custom_" + args_ctypes[key] + str(i)
-            ncode = ncode.replace(name, new_name)
-            variables_manually_typed.append((new_name, key, args_ctypes[key]))
-        else:
-            variables_manually_typed.append((name, key, type_))
-    variables = variables_manually_typed
-    if (compile_opt['extra_import'] or
-        test_parsed(ncode, variables, constants, args)):
+    ncode, variables = use_hinted_type(variables, ncode, args_ctypes)
+    if (
+        compile_opt['extra_import']
+        and not compile_opt['extra_import'].isspace()
+    ):
+        return ncode + ";", variables, constants, False
+    if test_parsed(ncode, variables, constants, args):
             return ncode, variables, constants, False
     else:
         warn("Could not find c types", StringParsingWarning)
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -28,7 +28,8 @@ from .cy.coefficient import (InterpolateCoefficient, InterCoefficient,
                              Coefficient)
 
 
-__all__ = ["coefficient", "CompilationOptions", "Coefficient"]
+__all__ = ["coefficient", "CompilationOptions", "Coefficient",
+           "clean_compiled_coefficient"]
 
 
 class StringParsingWarning(Warning):
@@ -236,6 +237,26 @@ def get_root():
     return root
 
 
+def clean_compiled_coefficient(all=False):
+    """
+    Remove previouly compiled string Coefficient.
+
+    Parameter:
+    ----------
+    all: bool
+        If not `all` will remove only previous version.
+    """
+    import glob
+    import shutil
+    tmproot = qset.install['tmproot']
+    root = os.path.join(tmproot, 'qutip_coeffs_{}'.format(COEFF_VERSION))
+    folders = glob.glob(os.path.join(tmproot, 'qutip_coeffs_') + "*")
+    for folder in folders:
+        if all or folder != root:
+            shutil.rmtree(folder)
+
+
+
 def proj(x):
     if np.isfinite(x):
         return (x)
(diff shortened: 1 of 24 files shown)


Ericgig:
The third part could be 1409 or a new PR, which ever is the easiest for you.
If you prefer a new PR, feel free to close 1409.

qutip/core/coefficient.py
```diff
@@ -341,9 +362,9 @@ def try_import(file_name, parsed_in):
             return None, file_name_
 
     if mod.parsed_code != parsed_in:
-        # At least 10 coeff with the same hash!?
-        # Giveup
-        raise ValueError("string hash collision, change the string or ")
+        # At least 10 coeff with the same hash!? Giveup...
+        raise ValueError("string hash collision, change the string "
+                         "or clean files in qutip.settings.install['tmproot']")
```
jakelishman, line 367: Did we change our minds about allowing hash collisions at all at some point?

qutip/core/coefficient.py
```diff
@@ -630,12 +655,11 @@ def try_parse(code, args, args_ctypes, compile_opt):
         constants = [(f, s, "object") for f, s, _ in constants]
     ncode, variables = use_hinted_type(variables, ncode, args_ctypes)
     if (
-        compile_opt['extra_import']
-        and not compile_opt['extra_import'].isspace()
+        (compile_opt['extra_import']
+        and not compile_opt['extra_import'].isspace())
+        or test_parsed(ncode, variables, constants, args)
```
jakelishman, line 660: It's difficult to read the indentation for the grouping here.  Also, looks like the `isspace()` call should be more like `if compile_opt['extra_import'].strip()` to account for multiple whitespace?  That could also be done on assignment, rather than on use.

jakelishman (approved):

jakelishman:
There's a new failure on an `mesolve` test in the Xcode 12 branch based on the relative tolerance.  It's _probably_ fine, but maybe just check.  Let's merge the fix for that in with this so we don't end up with the `diag_liou_mult` problem again (that I still haven't sorted out...).  It's probably just a case of being a little more lenient with the test tolerance.

qutip/core/coefficient.py
Ericgig: `isspace` return True for any numbers and mixes of tabs, newlines and spaces.

qutip/core/coefficient.py
Ericgig: This was changed in the core branch and not merged here. No new changes to coefficient.

qutip/core/coefficient.py
jakelishman: Ok cool. Still, if you strip it on assignment, then you don't have to remember to test it every time it's used later.

qutip/solver/options.py
```diff
@@ -0,0 +1,258 @@
...
+
+    The default can be changed by::
+
+        qutip.settings.solver['progress_bar'] = 'enhanced'
+
+    Options
+    -------
+    progress_bar : str {'text', 'enhanced', 'tqdm', ''}
+        How to present the solver progress.
+        'tqdm' use the python module of the same name and raise an error if
```
hodgestar, line 32:
```suggestion
        'tqdm' uses the python module of the same name and raise an error if
```

qutip/solver/options.py
```diff
@@ -0,0 +1,258 @@
...
+        not installed.
+        True will result in 'text'.
+        Empty string or False will disable the bar.
```
hodgestar, line 35: Nitpick: Could we put the explanations in the same order as the strings in the type?

What does `enhanced` do?

qutip/solver/options.py
```diff
@@ -0,0 +1,258 @@
...
+    SolverOptions constructor::
+
+        opts = SolverOptions(method=bdf, ...)
+
+    or by changing the class attributes after creation::
+
+        opts = SolverOptions()
+        opts.ode['method'] = 'bdf'
+
+    Returns options class to be used as options in evolution solvers.
```
hodgestar, line 63: The start of the docstring seems to be a copy and paste of the text from the `SolverOptions` docstring. This one should say something different. :)

qutip/solver/options.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+@optionsclass("results", SolverOptions)
+class SolverResultsOptions:
+    """
+    Class of options for Results of evolution solvers such as
+    :func:`qutip.mesolve` and :func:`qutip.mcsolve`.
+    Options can be specified when constructing SolverOptions
+
+        opts = SolverOptions(store_final_state=True, ...)
```
hodgestar, line 147:
```suggestion
        opts = SolverResultOptions(store_final_state=True, ...)
```

qutip/solver/options.py
```diff
@@ -0,0 +1,258 @@
...
+
+    or by changing the class attributes after creation::
+
+        opts = SolverOptions()
```
hodgestar, line 151:
```suggestion
        opts = SolverResultOptions()
```

qutip/solver/options.py
```diff
@@ -0,0 +1,258 @@
...
+        result class.
+
+    store_states : bool {False, True, None}
+        Whether or not to store the state vectors or density matrices.
+        On `None` the states will be saved if no expectation operators are
+        given.
+
+    normalize_output : str {"", "ket", "all"}
+        normalize output state to hide ODE numerical errors.
+        On "ket", only 'ket' output are normalized.
```
hodgestar, line 173: What do `""` and `"all"` do?

qutip/solver/parallel.py
```diff
@@ -0,0 +1,286 @@
...
+#    HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#    SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+#    LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#    DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#    THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+"""
+This function provides functions for parallel execution of loops and function
```
hodgestar, line 34:
```suggestion
This module provides functions for parallel execution of loops and function
```

qutip/solver/parallel.py
```diff
@@ -0,0 +1,286 @@
...
+mappings, using the builtin Python module multiprocessing.
```
hodgestar, line 35:
```suggestion
mappings, using the builtin Python module multiprocessing or the loky parallel execution library. 
```

qutip/tests/solver/test_parallel.py
```diff
@@ -0,0 +1,108 @@
...
+import time
+import pytest
+
+from qutip.solver.parallel import parallel_map, serial_map, loky_pmap
+"""
+try:
+    import loky
+except ModuleNotFoundError:
+    loky = False
+"""
```
hodgestar, line 44: This commented out try/except import an be deleted now.

qutip/tests/solver/test_parallel.py
```diff
@@ -0,0 +1,108 @@
...
+                         [parallel_map, loky_pmap, serial_map],
+                         ids=['parallel_map', 'loky_pmap', 'serial_map'])
+@pytest.mark.parametrize('num_cpus',
+                         [1, 2],
+                         ids=['1', '2'])
+def test_map(map, num_cpus):
+    if map is loky_pmap:
+        loky = pytest.importorskip("loky")
+    # if map is loky_pmap and not loky:
+    #     pytest.skip("module loky not available")
```
hodgestar, line 67: This commented out skip can be deleted now.

qutip/tests/solver/test_parallel.py
```diff
@@ -0,0 +1,108 @@
...
+def _func2(x, a, b, c, d=0, e=0, f=0):
+    assert d > 0
+    assert e > 0
+    assert f > 0
+    time.sleep(np.random.rand() * 0.1)  # random delay
+    return x**2
+
+@pytest.mark.parametrize('map',
+                         [parallel_map, loky_pmap, serial_map],
+                         ids=['parallel_map', 'loky_pmap', 'serial_map'])
```
hodgestar, line 59: We should use the same form of calling `parametrize` for `test_map` and `test_map_accumulator`. No point in having different styles in the same file.

qutip/tests/solver/test_parallel.py
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    x = np.arange(10)
+    y1 = [_func1(xx) for xx in x]
+
+    y2 = map(_func2, x, args, kwargs, map_kw=map_kw)
+    assert ((np.array(y1) == np.array(y2)).all())
+
+
+@pytest.mark.parametrize('map',[
```
hodgestar, line 84:
```suggestion
@pytest.mark.parametrize('map', [
```

hodgestar (approved):
I happy with the changes here overall and have left a few comments on things that I think should be tidied up.

qutip/tests/solver/test_parallel.py
jakelishman: Personally I like the `pytest.param` style best because it keeps all the information line-local, though in this case I _think_ it's also possible to pass an id function `ids=lambda f: f.__name__`

qutip/solver/options.py
Ericgig: No, the user always create a `SolverOptions` object which contain a `SolverResultOptions`. The creation is flat, everything is in  `SolverOptions.__init__`. The nested view is on modification after creation and printing:
```
opt = SolverOptions(store_final_state=True, method='bdf')
```
is equivalent to
```
opt = SolverOptions()
opt.results["store_final_state"] = True
opt.ode["method"] = "bdf"
```
I like this interface but it still feels rough...

qutip/solver/options.py
Ericgig: Change it look:
text:
```
10.0%. Run time:   0.00s. Est. time left: 00:00:00:00
20.0%. Run time:   0.00s. Est. time left: 00:00:00:00
```
enhanced:
`[*******  30%               ] Elapsed 13.43s / Remaining 00:00:00:31`
Not sure how to describe them...

PR #1465, opened 2021-03-17
Build wheels and distribute to PyPI (pip) from GitHub Actions

jakelishman:
This patch obsoletes #1429, by applying the packaging changes to the current version of `master` rather than needing the new structure on `dev.major`.  This patch should (mostly) apply cleanly to `dev.major` as well, so merging this up will be easy - the new automated systems in `setup.py` also mean that the distribution can handle both the current `master` and the `dev.major` package layouts with no changes.

The rest of this text is very similar to the original message in #1429.

---

This patch overhauls how QuTiP is packaged and distributed.  Once this patch is merged, there will be an option on QuTiP's GitHub Actions tab for people who have write permission on qutip/qutip to build all wheels and optionally release the build to PyPI, making it available on `pip` as a binary release.

The major changes are in the files `pyproject.toml`, `setup.py`, `setup.cfg` and `.github/workflows/build.yml`, while `MANIFEST.in` and a new file `VERSION` were also touched.

### Distribution changes:
  - there is a GitHub Action which will build all the wheels for CPython 3.6+ on Linux (`manylinux1` x86 and x86_64), macOS and Windows (32- and 64-bit).  This action is triggered manually, and the wheels will be available for download afterwards.  Optionally, the action will push to PyPI, making the version immediately available on `pip` by binary release.
  - `.cpp` files are now distributed with the wheels but not added to source control.  This is a step towards a full Cython-free delivery of the QuTiP source (binary releases never require Cython), but right now I think our `setuptools` machinery doesn't quite handle that.

### Changes to `setup.py`
  - all package metadata is moved out into `setup.cfg`, which is easier to read and change, and modern `setuptools`' preferred way of doing things
  - all directory-tree access is done in a completely platform-agnostic manner, rather than just hoping that Unix-style paths will work on Windows
  - the file is tidied and structured into a small script, with separated components
  - build requirements are moved into `pyproject.toml` including listing `setuptools` as the build method, in accordance with PEP 517.  Any PEP 517-complaint installation and build process will now _enforce_ the presence of `numpy`, `scipy` and `cython` (and `setuptools`) before attempting the build, so no more need for attempted import guards.
  - packages are now discovered rather than explicitly stated, using standard `setuptools` machinery
  - Cython extension modules are now discovered rather than explicitly stated, using globbing
  - version information is now stored in a single-line file `VERSION`.  This is mostly for CI reasons; it's much easier to parse and override at the CI level when it's a single file, rather than attempting  `sed s/.../.../ setup.py` or something crazy like that.  The validity of the version string is tested by regex.
  - local version information is now _always_ appended, even if `git` isn't present - the local version is just then "nogit" rather than the git shorthash
  - `setup.py` gained a new `--release` flag or environment variable `CI_QUTIP_RELEASE`; the only effect right now is to suppress the local identifier on the version information, replacing the old `ISRELEASED` boolean flag in the code

### Notes on the wheels

Currently I build only for CPython.  I haven't tried to build pypy wheels, but we _may_ support it - Cython does in theory, but I think complex numbers and raw pointers aren't handled completely smoothly, and we make heavy use of both.  We get `manylinux1` builds for Linux, which is the oldest `manylinux` spec.  This can be updated in the future, but I'm not sure I see a need; all our heavy mathematical lifting is either done by custom code (which is mostly immune) or BLAS/LAPACK (which we link to dynamically), so I don't think there's much to be gained.  I don't know if the macOS images will run on the new M1 chips, but if not, there's not much that can be done there until GitHub Actions adds the necessary cross-compilation headers and libraries to their CI.

### Examples

You can see the result of this upload on the testing PyPI server: https://test.pypi.org/project/qutip-jakelishman (version 4.6.0a1 is the current state of `master`).  I ran the CI action from my fork, with a temporary commit that changed the server to the testing archive (and the package name).  Here's a screenshot of what the deployment screen will look like:

GitHub Actions screen when triggering the build and deploy action


.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
+name: Build wheels and deploy
+
+on:
+  workflow_dispatch:
+    inputs:
+      confirm_ref:
+        description: "Enter branch/tag name to deploy to PyPI, or leave blank to only build wheels."
```
hodgestar, line 7: To me this reads as "I can type the name of the branch I want to deploy to PyPI here", but the intention seems to be that one should type the name of this branch to confirm. Maybe we can change the start to `Enter the name of this tag or branch to deploy ..."?

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Install cibuildwheel
+        run: |
+          # cibuildwheel does the heavy lifting for us. Originally tested on
+          # 1.8.0, but should be fine at least up to any major new release.
+          python -m pip install 'cibuildwheel<2.0'
+
+      - name: Install Visual C++ for Python 2.7
+        if: runner.os == 'Windows'
+        run: |
+          choco install vcpython27 -f -y
```
hodgestar, line 56: What do we need Python 2.7 for here?

.gitignore
```diff
@@ -7,6 +7,7 @@
 
 # Packages
 *.egg
+.eggs
```
hodgestar, line 10: What makes `.eggs`?

qutip/cy/brtools.pyx
```diff
@@ -403,7 +403,7 @@ cdef np.ndarray[complex, ndim=1, mode='c'] vec_to_fockbasis(double complex * eig
     PyDataMem_FREE(temp1)
     cdef np.ndarray[complex, ndim=1, mode='c'] out = \
                 np.PyArray_SimpleNewFromData(1, &nrows2, np.NPY_COMPLEX128, fock_vec)
-    PyArray_ENABLEFLAGS(out, np.NPY_OWNDATA)
+    PyArray_ENABLEFLAGS(out, np.NPY_ARRAY_OWNDATA)
```
hodgestar, line 406: Could you elaborate a little on these changes? Did the numpy constant name change at some point or what happened?

hodgestar (approved):
I left some small suggestions and questions, but otherwise this looks great! Thank you @jakelishman.

hodgestar:
The CI failure on ` macOS, Python 3.8: XCode 12` is because the tests took too long. Not sure if this is just a Travis issue or a pre-existing problem or something new.

.github/workflows/build.yml
jakelishman: Yeah, this could be better.  I struggled with this wording, especially because there's also a minor space constraint - if you look in the picture in the PR text above, you can see that this text gets wrapped into really short lines, so I didn't want to write something that would end up 5 lines long and have no-one read it.

Perhaps?
> Confirm branch name to deploy to PyPI, leave blank to only build wheels.

This is a wordsmithing problem, so I bet @ajgpitch will enjoy it too.

.github/workflows/build.yml
jakelishman: It's part of the cibuildwheel default action - I think it's because the lowest common denominator for Windows CI runners is assuming Python 2.7.  This is just bootstrapping cibuildwheel; once it starts running, I think it pulls in some Docker images that have a proper Python 3.x environment all ready to go, and does the build inside them.

.github/workflows/build.yml
hodgestar: `Confirm branch name to deploy to PyPI, leave blank to only build wheels.` looks good to me. Or perhaps `Enter branch name to confirm deploying to PyPI, leave blank to only build wheels`?

Maybe anything other than blank or the correct branch name should fail the build? That is very much a nice to have though.

.gitignore
jakelishman: It's a funny one - setuptools uses it as part of the PEP 517 procedure, but only when you call `python setup.py ...` directly, and it needs to install a `setup_requires` requirement.  If you use a "regular" PEP517 builder via pip, it'll install the temporary build things in a venv in a temporary directory (which sucks a little because it means Cython invariably detects a change in the numpy version you're building against, and rebuilds everything).

The thing is, that if you run `python setup.py ...` directly, the first lines of the files import all the third-party imports required in `setup_requires`, so it should just report a Python error.  However, if you have a version of numpy, scipy or cython that violates the _version_ requirements, `setup.py` will download suitable versions once it hits `setuptools.setup()`.  The problem is that the damage is already done for us at that point - Cython will have cythonised and compiled against a bad version of Numpy, or using an old version of Cython.

Still, it's possible for the directory to be created, and we will never want it committed to source control.

qutip/cy/brtools.pyx
jakelishman: This is a deprecated C-API call as of Numpy 1.7 (years ago).  It added the `NPY_ARRAY_` namespace to tidy up their API.  The old ones are unlikely to be removed until Numpy 2.0 it seems, so it's not particularly urgent.

This is actually a commit I left in by accident in the rebase.  Originally I also wanted to specify the `NPY_DEPRECATED_APIVERSION=NPY_1_16_API_VERSION` macro in the compilation to keep us writing modern C-API code against Numpy.  Unfortunately, Cython 0.29 still generates some pre-1.7 API code when you use the `cdef ndarray[int, ndim=2]` form (instead of the more modern typed memoryview), so I had to remove it and I must have still picked this commit in the rebase.

I can force-push remove it, or I could just leave it in since it's in its own commit anyway, so anyone checking the git log will see it as a separate entity at least (caa5e32).

jakelishman:
The macOS Xcode 12 runners have been terminally slow for some time now.  Looking at the setup time, it took nearly 9 minutes to build QuTiP, which is crazy compared to Linux and Windows build times (~2 minutes), or even Xcode 11 (~3 minutes).

We noticed this starting several months ago, but the problem is that I don't have a machine I can test it on to see if it's a Travis problem or an Xcode 12 one. I'm unwilling to change my personal macOS version beyond 10.14 in part _because_ there are changes to how C compilation and debugging are handled - Apple are making it harder and harder to attach debuggers and set up sane compilers that aren't Xcode, and it took me long enough to get an actually fully featured compiler running on 10.14, let alone the next versions.  Their vendored version of clang doesn't even support OpenMP!

.github/workflows/build.yml
jakelishman: Yeah, it's probably better to hard fail if something's typed but it's the wrong thing.  Unfortunately, there's no way to do pre-processing on these values, so I'll have to change the "canary" job that just exists to provide a visible tick saying "will attempt to deploy" to explicitly kill the whole run if the string is non-empty and unequal to the branch name.

jakelishman:
@hodgestar how's the `setup.py` file look to you, Simon?  That's the biggest change from the other version of this PR - I tidied it all up into separated functions as I think you mentioned in that review.

.gitignore
hodgestar: Tx for the explanation -- let's keep it.

qutip/cy/brtools.pyx
hodgestar: Happy for it to stay in.

hodgestar:
> The macOS Xcode 12 runners have been terminally slow for some time now. Looking at the setup time, it took nearly 9 minutes to build QuTiP, which is crazy compared to Linux and Windows build times (~2 minutes), or even Xcode 11 (~3 minutes).

If it's a known issue, let's leave it for when we switch CI to GitHub Actions (assuming that's the choice we end up making).

hodgestar:
> @hodgestar how's the `setup.py` file look to you, Simon? That's the biggest change from the other version of this PR - I tidied it all up into separated functions as I think you mentioned in that review.

I noticed! It looks great. :D

jakelishman:
> If it's a known issue, let's leave it for when we switch CI to GitHub Actions (assuming that's the choice we end up making).

It's very very difficult to tell, because Travis seem to be aggressively winding down the processing power available to us on Macs - I can't tell if there's a problem in this code, if Travis are over-stuffing their Xcode 12 machines with too many VMs, or even if it's some weird emulation (I think Azure _emulates_ PPC architectures if you ask for them!).

jakelishman:
So I re-ran the failed macOS Xcode 12 test, and it succeeded no problems this time round, taking half the previous amount of time to build the library.  I suspect Travis' Macs are just being intermittently overloaded.

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: "Built wheels will be deployed at the end."
+        run: |
+          echo "Built wheels will be deployed at the end."
+
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
```
nathanshammah, line 29: is there a reason why this is not ubuntu-latest?

.github/workflows/build.yml
nathanshammah: Can a check on `github.repository` or `github.repository_owner` work as well? See related PR for Mitiq https://github.com/unitaryfund/mitiq/pull/597

nathanshammah: I think that making the trigger based on a label ("PyPI release") would be nice.

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
+name: Build wheels and deploy
```
nathanshammah, line 1:
```suggestion
name: Build wheels and deploy (to PyPI, if active)
```

setup.cfg
```diff
@@ -0,0 +1,42 @@
...
+name = qutip
+url = http://qutip.org
+description = QuTiP: The Quantum Toolbox in Python
+long_description = file: README.md
+long_description_content_type = text/markdown; variant=GFM
+keywords = quantum, physics, dynamics
+license = BSD 3-Clause License
+license_file = LICENSE.txt
+classifiers =
+    Development Status :: 2 - Pre-Alpha
```
nathanshammah, line 11: This is unrelated to this PR, but, are we still in  Pre-Alpha? That sounds strange given it's 4.5 etc. I suggest opening an issue to choose a dev status more adherent to reality.

nathanshammah (approved):
Thanks @jakelishman. I added some suggestions on the way this can be triggered. E.g., with a label. Another option is to have the creation of GH release be the trigger event, see the [workflow](https://github.com/unitaryfund/mitiq/blob/master/.github/workflows/publish-pypi.yml) for Mitiq.

.github/workflows/build.yml
jakelishman: The trigger is a manual click action - you have to press a button on GitHub so there's no danger of "accidentally" running it on a fork.

.github/workflows/build.yml
jakelishman: I just got it from the `cibuildwheel` default action - it makes no difference at all, because all the actual build is done within a Docker image.  I suspect 18.04 was the last LTS release of Ubuntu.

.github/workflows/build.yml
jakelishman: Don't really mind, but the "PyPI" name appears elsewhere in the admin-facing components already.

setup.cfg
jakelishman: This only gets changed when we make a release - the "pre-alpha" status refers to the state of the master branch.

.github/workflows/build.yml
jakelishman: Or were you suggesting we do it based on a label?  In this, the only way to get a build is to type in a particular branch name, which implies that you have to have produced a special branch/git tag for the release, so that might be quite similar to what you were asking?

.github/workflows/build.yml
jakelishman: @hodgestar: on further thought, it's possible I'm wrong on this and it may only be needed if you want to build Python 2.7 wheels too.  I should check if it works without this line before I merge in unnecessary stuff.

qutip/cy/brtools.pyx
jakelishman: Yeah, not ideal of me to mess up the PR history like that, but at least from the git user's perspective it doesn't matter so much.

.github/workflows/build.yml
hodgestar: I quite like having to do a manual release rather than using a tag because it allows one to rerun the build if a problem occurs and to try things out without filling the github repository with tags named things like "test-457".

jakelishman: ^ This is the reason I went for this method in the end. I think when I originally proposed automatic deployment I talked about it being tag-based, but I think this is cleaner.

You can also force-push a tag that had a failed release attached to it, though - that's not doing any harm, because a tag isn't _truly_ public until the release is made successfully.

jakelishman: > 
> 
> Can a check on `github.repository` or `github.repository_owner` work as well? See related PR for Mitiq [unitaryfund/mitiq#597](https://github.com/unitaryfund/mitiq/pull/597)

So this file uses a slightly different mechanism - in mitiq you're tying the release action to GitHub's "make release" button, whereas here I tie it to you manually running this action.  One major advantage to this method is that admins can build a full set of wheels _without_ making a public release, which is a bigger deal for us because our wheels involve very nontrivial compilation, so there's rather more chance of something going wrong.

BoxiLi:
About tests being very slow, I saw the following notice when I open the details:
```
Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.
```
I think we are still on `travis-ci.org`? Is this related to us? Maybe "switch CI to GitHub Actions" is more pressing than we thought.

jakelishman:
> I think we are still on `travis-ci.org`? Is this related to us? Maybe "switch CI to GitHub Actions" is more pressing than we thought.

Yeah, it's one of the reasons that issue has made a fairly huge resurgence recently.  We're absolutely running on borrowed time on Travis.

.github/workflows/build.yml
jakelishman: I turned this into `ubuntu-latest` because I suppose `ubuntu-latest` will always work, whereas `ubuntu-18.04` will (eventually) go out-of-date.

.github/workflows/build.yml
jakelishman: I tested it, and this isn't necessary for what we want to build.  I've removed it.

jakelishman:
Last review re-request, hopefully.  I think everything's correct and ready now.

1. I had forgotten to build the sdist here, so it would have been missing from the pip package.  That's fixed now.
2. It's harder than you might expect to quickly fail all jobs in a GitHub Actions workflow right now.  It seems that they expect you to try and continue for as long as possible for jobs on different runners, so I had to make everything depend on the confirmation verification to get the fail-fast behaviour.
3. Hopefully neatened up the language we've been talking about.
4. Removed installation unnecessary VC++ tools for Python 2.7 on Windows.

.github/workflows/build.yml
```diff
@@ -0,0 +1,168 @@
...
+    env:
+      GITHUB_REF: ${{ github.ref }}
+      CONFIRM_REF: ${{ github.event.inputs.confirm_ref }}
+    steps:
+      - name: Compare confirmation to current reference
+        shell: bash
+        run: |
+          [[ -z $CONFIRM_REF || $GITHUB_REF =~ ^refs/(heads|tags)/$CONFIRM_REF$ ]]
+          if [[ -z $CONFIRM_REF ]]; then
+            echo "Only building wheels."
```
hodgestar, line 35:
```suggestion
            echo "Build only. No files will be uploaded to PyPI."
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,168 @@
...
+          else
+            echo "Built wheels will be deployed at the end."
```
hodgestar, line 37:
```suggestion
            echo "Full deploy. Built wheels and source distributions will be uploaded to PyPI."
```

hodgestar (approved):
I made two small suggestions to the echo'ed messages to clarify what "deploy" means and that we're uploading more than just wheels, but happy for this to be merged.

jakelishman:
Ok, it's taken about 2 months of us deciding on what exactly we want, but I think we're better off for it.  Everything's passing, so I'm merging this now.

PR #1469, opened 2021-03-25
BUG: Options for coherent not checked correctly

madphysicist:
**Description**
Since the code is checking `or offset ? 0`, the method is effectively being ignored in most cases. Pretty sure that's not intended behavior. Changing to correct behavior in `coherent` makes `coherent_dm` pretty trivial.

In theory, this introduces nothing that isn't explicitly documented. However, since there don't appear to be any tests that were failing because of this, I'll try to add some.

**Changelog**
Fixed method name check in `coherent` and `coherent_dm`.


qutip/states.py
```diff
@@ -218,14 +218,19 @@ def coherent(N, alpha, offset=0, method='operator'):
     but would in that case give more accurate coefficients.
 
     """
-    if method == "operator" and offset == 0:
+    if offset < 0:
+        raise ValueError('Offset must me non-negative')
```
jakelishman, line 222: me -> be

qutip/states.py
```diff
@@ -291,17 +296,8 @@ def coherent_dm(N, alpha, offset=0, method='operator'):
...
-        return psi * psi.dag()
-
-    elif method == "analytic":
-        psi = coherent(N, alpha, offset=offset, method='analytic')
-        return psi * psi.dag()
-
-    else:
-        raise TypeError(
-            "The method option can only take values 'operator' or 'analytic'")
+    psi = coherent(N, alpha, offset=offset)
```
jakelishman, line 299: Shouldn't this pass on the `method` value?

jakelishman (changes requested):
Thanks for spotting this.  I think the "error" was actually more to do with the default handling, which you have fixed.

However, I think you might have missed a bit of the logic in your solution - the "operator" method is _only_ valid for a zero offset, so it should be an error to ask for
```python
coherent(5, 0.25j, offset=5, method="operator")
```
but yours will give the same results as if you had set `offset=0`.  That should be quite simple to fix, and your new handling of the defaults makes it much easier to detect a user error.  It's not exactly a ringing endorsement of our testing that we didn't catch the logic slip in CI, though...  Good extra tests are always welcome.

qutip/states.py
madphysicist: Yes :(. I should probably have put WIP in the title given that I just threw something together without checking it. I'll fix everything later tonight.

jakelishman:
@madphysicist: would you like to finish this up?  If you don't have time, I'll make the last changes and get it merged (keeping your commits and your contribution record) for inclusion in QuTiP 4.7.

madphysicist:
@jakelishman Sorry about letting this slip through the cracks. I'll finish it up myself ASAP.

jakelishman:
Oh, don't worry about rushing - there's no hurry.  We don't expect to release 4.7 for _at least_ a month, probably two.  I was just mentioning it now so you'd have time if you were still interested.

hodgestar:
@madphysicist Just re-pinging. As before, it's perfectly okay if you've run out of time a bit -- just let us know.

qutip/states.py
hodgestar, line 222:
```suggestion
        raise ValueError('Offset must be non-negative')
```

qutip/states.py
hodgestar, line 299:
```suggestion
    psi = coherent(N, alpha, offset=offset, method=method)
```

hodgestar (commented):
Just applying the changes @jakelishman suggested.

hodgestar:
I think we just need to resolve the conflicts in `qutip/tests/test_states.py` and then perhaps we can merge.

qutip/tests/test_states.py
```diff
@@ -236,4 +252,4 @@ def test_bell_state():
     for state1, state2 in combinations(states, 2):
         assert state1.overlap(state2) == pytest.approx(0.0)
 
-    assert qutip.singlet_state() == qutip.bell_state('11')
+    assert qutip.singlet_state() == qutip.bell_state('11')
```
hodgestar, line 255:
```suggestion
    assert qutip.singlet_state() == qutip.bell_state('11')

```

qutip/tests/test_states.py
```diff
@@ -123,6 +123,22 @@ def test_CoherentState():
         qutip.coherent(N, alpha, method="other")
     assert str(e.value) == ("The method option can only take "
                             "values 'operator' or 'analytic'")
+    with pytest.raises(ValueError) as e:
+        qutip.coherent(N, -1)
+    assert str(e.value) == ("XXX")
```
hodgestar, line 128: TODO: Fix this error message.

qutip/tests/test_states.py
```diff
@@ -123,6 +123,22 @@ def test_CoherentState():
...
+    N = 10
+    rho = qutip.coherent_dm(N, 1)
+    assert rho.tr() == pytest.approx(1.0)
+    with pytest.raises(ValueError) as e:
+        qutip.coherent_dm(N, 1, method="other")
+    assert str(e.value) == ("The method option can only take "
+                            "values 'operator' or 'analytic'")
+    with pytest.raises(ValueError) as e:
+        qutip.coherent_dm(N, -1)
+    assert str(e.value) == ("XXX")
```
hodgestar, line 141: TODO: Fix this error message.

hodgestar (changes requested):
I resolved the merge conflict, but the tests might need some cleaning up since they were re-written on master to use pytest more.

qutip/tests/test_states.py
hodgestar, line 128:
```suggestion
    with pytest.raises(ValueError) as e:
        qutip.coherent(N, alpha, -1)
    assert str(e.value) == ("Offset must be non-negative")
```

qutip/tests/test_states.py
hodgestar, line 141:
```suggestion
    with pytest.raises(ValueError) as e:
        qutip.coherent_dm(N, 1, offset=-1)
    assert str(e.value) == ("Offset must be non-negative")
```

hodgestar (approved):

Ericgig:
@hodgestar  If an offset is passed with `method="operator"`, it is quietly ignored...

hodgestar:
@Ericgig Should we just raise an error if `offset != 0` is passed with `method="operator"`?

Ericgig:
That would be good.

PR #1471, opened 2021-03-26
Fix OpenMP calibration and tidyup __init__

### LINKED ISSUE #1470: OpenMP broken in current master, OpenMP tests worrying
OpenMP handling is broken in the current `master`.  It is likely caused by changes made in #1289.  Trying to `import qutip` with OpenMP built produces errors
```python
In [1]: import qutip
Calibrating OPENMP threshold...
<ipython-input-1-e01d5608f34f> in <module>
----> 1 import qutip

~/code/qutip/qutip/qutip/__init__.py in <module>
    277         print('Calibrating OPENMP threshold...')
    278         thrsh = calculate_openmp_thresh()
--> 279         qutip.configrc.write_rc_key(rc_file, 'openmp_thresh', thrsh)
    280
    281 # Load the config file

~/code/qutip/qutip/qutip/configrc.py in write_rc_key(key, value, section, rc_file)
    169     config.read(rc_file)
    170     if not config.has_section(section):
--> 171         config.add_section(section)
    172     config.set(section, key, str(value))
    173

~/.anaconda3/envs/qutip-dev/lib/python3.8/configparser.py in add_section(self, section)
   1205         RawConfigParser.add_section by validating if the section name is
   1206         a string."""
-> 1207         self._validate_value_types(section=section)
   1208         super().add_section(section)
   1209

~/.anaconda3/envs/qutip-dev/lib/python3.8/configparser.py in _validate_value_types(self, section, option, value)
   1178         """
   1179         if not isinstance(section, str):
-> 1180             raise TypeError("section names must be strings")
   1181         if not isinstance(option, str):
   1182             raise TypeError("option keys must be strings")

TypeError: section names must be strings
```

The issue is probably due to `__init__.py` being out-of-date with regards to the change to how the `qutiprc` file is written.  The immediate fix should be very simple, however it highlights a larger problem with our testing - this should have been caught by the OpenMP tests.

In this case, I believe that the tests don't see the issue because there is a test in `__init__.py` that doesn't trigger the OpenMP calibration if only there is only 1 CPU.  Presumably, we only ever get allocated single-CPU VMs on Travis, so it hasn't got caught.  That makes me worry a lot that the OpenMP code isn't even being tested correctly.

jakelishman:
This fixes the calls to the `qutip.configrc` module in `qutip/__init__.py` when setting up OpenMP.

Also, this tidies up a lot of very old code out of `__init__.py` that was testing for ancient versions of SciPy and Numpy.  Since we have hard requirements at installation time for those, it's fine for us not to test their versions at initialisation.  We still should check the version of Cython because it's an optional dependency, so isn't always constrained by the package manager.

Some of the changes deliberately make `qutip/__init__.py` less forgiving of exceptions, especially when it comes to loading up `pyximport`.  This is deliberate, even though it may make a couple more errors surface that have previously gone undetected.  As it stands right now, the file will causes several errors to be silently ignored, even if they weren't the error we were testing for.  That masks problems for a short while, but they usually resurface during use when they're much harder to debug.  We should be solving the underlying problems, not sweeping them under the rug.  In particular, we've frequently had complaints about `pyximport` failing, or trying to import OpenMP modules that have previously been compiled but aren't currently active, and things like that.  This patch won't fix all those problems, but it hopefully will make some of them easier to debug when they do appear.

There is more explanation to the logic behind each change in the commit messages.

Fix #1470.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,6 +72,7 @@ jobs:
         - python setup.py develop --with-openmp
       before_script:
         - QUTIP_INSTALL="${QUTIP_DOWNLOAD}/qutip"
+        - export QUTIP_NUM_PROCESSES=2
       script:
         - python -c "import qutip; qutip.about()"
         - pytest --verbosity=1 --cov=qutip --cov-config="${QUTIP_DOWNLOAD}/.coveragerc" --rootdir="${QUTIP_INSTALL}/tests" -c "${QUTIP_INSTALL}/tests/pytest.ini" --pyargs qutip
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -30,9 +30,7 @@
 #    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
-from __future__ import division, print_function, absolute_import
 import os
-import sys
 import warnings
 
 import qutip.settings
@@ -112,6 +110,8 @@ if qutip.settings.num_cpus == 0:
             qutip.settings.num_cpus = multiprocessing.cpu_count()
         except NotImplementedError:
             qutip.settings.num_cpus = 1
+
+del multiprocessing
 
 
 # Find MKL library if it exists
@@ -239,4 +239,4 @@ if has_rc:
 # -----------------------------------------------------------------------------
 # Clean name space
 #
-del os, sys, multiprocessing, warnings
+del os, warnings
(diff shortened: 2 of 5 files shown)


hodgestar (approved):
I'm not an expert on this code, but removing all the old version testing looks good and I didn't see anything obviously wrong.

jakelishman:
@nwlambert Neill, does this fix everything for you? I'd like to have a couple of people test it if possible, since I toughened up the exception throwing, which might have caused more previously hidden problems to surface.

nwlambert:
Yup, it installed ok, the openmp threshold check error is gone, and all tests passed ok~!

jakelishman:
Ok, I'm going to merge this now so it any potential issues it may turn up will have more exposure through other people working on the code.  I wouldn't think that the changes I've made are particularly contentious, just pruning out some stuff that's been obsoleted by the passage of time.

PR #1478, opened 2021-03-29
Issue#1460

### LINKED ISSUE #1460: correlation_2op_1t() and spectrum() giving the physically incorrect result unless solver='es'

AGaliciaMartinez:
**Description**
The function mesolve was not including the complex part of the expectation value when the initial state was a not Hermitian. This led to an error in `correlation_2op_1t()` which uses an initial state that is not Hermitian to compute the correlation.
I also fixed an error in the documentation where `tlist` was referenced instead of `taulist` (`tlist` is not an argument in `correlation_2op_1t()`) and included a test for `correlation_2op_1t()`.

**Related issues or PRs**
Fix #1460

**Changelog**
Fixed error where complex expectation value for non Hermitian initial state was not being included in results when using mesolve.
Fix doxumentation error in `correlation_2op_1t()`.
Added test for `correlation_2op_1t()`.

### DIFF
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -537,6 +537,7 @@ def _generic_ode_solve(func, ode_args, rho0, tlist, e_ops, opt,
 
     if opt.store_final_state:
         cdata = get_curr_state_data(r)
-        output.final_state = Qobj(cdata, dims=dims, isherm=True)
+        output.final_state = Qobj(cdata, dims=dims,
+                                  isherm=rho0.isherm or None)
 
     return output
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -510,18 +510,20 @@ def _generic_ode_solve(func, ode_args, rho0, tlist, e_ops, opt,
 
         if opt.store_states or expt_callback:
             cdata = get_curr_state_data(r)
+            fdata = dense2D_to_fastcsr_fmode(cdata, size, size)
+
+            # Try to guess if there is a fast path for rho_t
+            if issuper(rho0) or not rho0.isherm:
+                rho_t = Qobj(fdata, dims=dims)
+            else:
+                rho_t = Qobj(fdata, dims=dims, fast="mc-dm")
 
         if opt.store_states:
-            if issuper(rho0):
-                fdata = dense2D_to_fastcsr_fmode(cdata, size, size)
-                output.states.append(Qobj(fdata, dims=dims))
-            else:
-                fdata = dense2D_to_fastcsr_fmode(cdata, size, size)
-                output.states.append(Qobj(fdata, dims=dims, fast="mc-dm"))
+            output.states.append(rho_t)
 
         if expt_callback:
             # use callback method
-            output.expect.append(e_ops(t, Qobj(cdata, dims=dims)))
+            output.expect.append(e_ops(t, rho_t))
 
         for m in range(n_expt_op):
             output.expect[m][t_idx] = expect_rho_vec(e_ops_data[m], r.y,
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -525,7 +525,8 @@ def _generic_ode_solve(func, ode_args, rho0, tlist, e_ops, opt,
 
         for m in range(n_expt_op):
             output.expect[m][t_idx] = expect_rho_vec(e_ops_data[m], r.y,
-                                                     e_ops[m].isherm and rho0.isherm)
+                                                     e_ops[m].isherm
+                                                     and rho0.isherm)
 
         if t_idx < n_tsteps - 1:
             r.integrate(r.t + dt[t_idx])
--- a/qutip/mesolve.py
+++ b/qutip/mesolve.py
@@ -525,7 +525,7 @@ def _generic_ode_solve(func, ode_args, rho0, tlist, e_ops, opt,
 
         for m in range(n_expt_op):
             output.expect[m][t_idx] = expect_rho_vec(e_ops_data[m], r.y,
-                                                     e_ops[m].isherm)
+                                                     e_ops[m].isherm and rho0.isherm)
 
         if t_idx < n_tsteps - 1:
             r.integrate(r.t + dt[t_idx])
--- a/qutip/tests/test_correlation.py
+++ b/qutip/tests/test_correlation.py
@@ -269,7 +269,12 @@ def test_hamiltonian_order_unimportant():
                          ids=["hermitian", "nonhermitian"])
 @pytest.mark.parametrize("w", [1, 2])
 @pytest.mark.parametrize("gamma", [1, 10])
-def test_correlation_2op_1t(solver, state, is_e_op_hermitian, w, gamma):
+def test_correlation_2op_1t_known_cases(solver,
+                                        state,
+                                        is_e_op_hermitian,
+                                        w,
+                                        gamma,
+                                       ):
     """This test compares the output correlation_2op_1 solution to an analytical
     solution."""
 
--- a/qutip/tests/test_correlation.py
+++ b/qutip/tests/test_correlation.py
@@ -58,6 +58,7 @@ def test_correlation_solver_equivalence(solver, start, legacy):
     system.
     """
     a = qutip.destroy(_equivalence_dimension)
+    x = ( a +a.dag() )/np.sqrt(2)
     H = a.dag() * a
     G1 = 0.75
     n_th = 2
@@ -74,8 +75,8 @@ def test_correlation_solver_equivalence(solver, start, legacy):
     # broken, whereas if only one fails, then it implies that only that one is
     # broken.  We test that all solvers are equivalent by transitive equality
     # to the "me" solver.
-    base = correlation(H, start, None, times, c_ops, a.dag(), a, solver="me")
-    cmp = correlation(H, start, None, times, c_ops, a.dag(), a, solver=solver)
+    base = correlation(H, start, None, times, c_ops, x, x, solver="me")
+    cmp = correlation(H, start, None, times, c_ops, x, x, solver=solver)
     np.testing.assert_allclose(base, cmp, atol=tol)
 
 
--- a/qutip/tests/test_correlation.py
+++ b/qutip/tests/test_correlation.py
@@ -76,8 +76,8 @@ def test_correlation_solver_equivalence(solver, start, legacy):
     # broken, whereas if only one fails, then it implies that only that one is
     # broken.  We test that all solvers are equivalent by transitive equality
     # to the "me" solver.
-    base = correlation(H, start, None, times, c_ops, a, a.dag(), solver="me")
-    cmp = correlation(H, start, None, times, c_ops, a, a.dag(), solver=solver)
+    base = correlation(H, start, None, times, c_ops, a.dag(), a, solver="me")
+    cmp = correlation(H, start, None, times, c_ops, a.dag(), a, solver=solver)
     np.testing.assert_allclose(base, cmp, atol=tol)
 
 
--- a/qutip/tests/test_correlation.py
+++ b/qutip/tests/test_correlation.py
@@ -75,8 +75,8 @@ def test_correlation_solver_equivalence(solver, start, legacy):
     # broken, whereas if only one fails, then it implies that only that one is
     # broken.  We test that all solvers are equivalent by transitive equality
     # to the "me" solver.
-    base = correlation(H, start, None, times, c_ops, x, x, solver="me")
-    cmp = correlation(H, start, None, times, c_ops, x, x, solver=solver)
+    base = correlation(H, start, None, times, c_ops, a, a.dag(), solver="me")
+    cmp = correlation(H, start, None, times, c_ops, a, a.dag(), solver=solver)
     np.testing.assert_allclose(base, cmp, atol=tol)
 
 
--- a/qutip/tests/test_correlation.py
+++ b/qutip/tests/test_correlation.py
@@ -255,26 +255,21 @@ def test_hamiltonian_order_unimportant():
     np.testing.assert_allclose(forwards, backwards, atol=1e-6)
 
 
-# Paramters for the test
-solver_list = ['me', 'es']
-start_list = [ 'coherent_dm', 'fock', None]
-e_op_list = ['non-hermitian', 'hermitian']
-w_list = [1,2] # Fast and slow oscillations
-gamma_list = [1,10] # Small and high damping
-args = list(product(solver_list, start_list, e_op_list, w_list, gamma_list))
-
-# We want to mark mc as slow
-solver_list = ['mc']
-start_list = [ 'fock']
-e_op_list = ['non-hermitian', 'hermitian']
-w_list = [1,2] # Fast and slow oscillations
-gamma_list = [1,10] # Small and high damping
-args_slow = list(product(solver_list, start_list, e_op_list, w_list, gamma_list))
-args += [pytest.param(*arg, marks = pytest.mark.slow) for arg in args_slow]
-
-
-@pytest.mark.parametrize(["solver", "start", "e_op", "w", "gamma"], args)
-def test_correlation_2op_1t(solver, start, e_op, w, gamma):
+@pytest.mark.parametrize(['solver', 'state'], [
+    pytest.param('me', _equivalence_fock, id="me-ket"),
+    pytest.param('me', _equivalence_coherent, id="me-dm"),
+    pytest.param('me', None, id="me-steady"),
+    pytest.param('es', _equivalence_fock, id="es-ket"),
+    pytest.param('es', _equivalence_coherent, id="es-dm"),
+    pytest.param('es', None, id="es-steady"),
+    pytest.param('mc', _equivalence_fock, id="mc-ket",
+                 marks=[pytest.mark.slow]),
+])
+@pytest.mark.parametrize("is_e_op_hermitian", [True, False],
+                         ids=["hermitian", "nonhermitian"])
+@pytest.mark.parametrize("w", [1, 2])
+@pytest.mark.parametrize("gamma", [1, 10])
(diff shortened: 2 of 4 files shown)


qutip/mesolve.py
```diff
@@ -510,22 +510,25 @@ def get_curr_state_data(r):
...
-                output.states.append(Qobj(fdata, dims=dims, fast="mc-dm"))
+                rho_t = Qobj(fdata, dims=dims, fast="mc-dm")
+
+        if opt.store_states:
+            output.states.append(rho_t)
 
         if expt_callback:
             # use callback method
-            output.expect.append(e_ops(t, Qobj(cdata, dims=dims)))
+            output.expect.append(e_ops(t, rho_t))
```
nwlambert, line 526: if store_states=false, and exp_callback = true, rho_t is not defined to use here?

nwlambert (commented):
thanks asier, looks like an elegant fix.  just added a quick comment.

qutip/mesolve.py
AGaliciaMartinez: Thanks for reviewing the PR Neill!. I think that if `expt_callback = True` and `store_states=False` the function should still work since it creates `rho_t` when any of those are True. However, I have just realized that I did not include any test for this case. I will include it as soon as possible.

qutip/tests/test_mesolve.py
```diff
@@ -920,5 +920,41 @@ def f(t, args):
...
+
+    options = Options()
+    options.store_final_state = True
+    options.store_states = True
+    result = mesolve(H, rho0, tlist, e_ops=[x], options=options)
+
+    imag_part = np.abs(np.imag(result.expect[0][-1]))
+    # Since we used an initial state that is not Hermitian, the expectation of
+    # x must be imaginary for t>0.
+    msg = 'Mesolve is not working properly with a non Hermitian density ' +
```
nwlambert, line 945: looks like the tests are failing cos of this linebreak, I guess wrapping in parentheses should fix it?

AGaliciaMartinez: Yes, indeed. It should be fixed in the last commit.

qutip/tests/test_mesolve.py
```diff
@@ -920,5 +920,53 @@ def f(t, args):
...
+    not Hermitian.
+    See Issue #1460
+    """
+    N = 2
+    a = destroy(N)
+    n = num(N)
+    x = (a + a.dag())/np.sqrt(2)
+    H = a.dag() * a
+
+    rho0 = x*fock_dm(N, 0)
```
jakelishman, line 934: It doesn't matter _too_ much because this is a test and not full code, but this is a pretty confusing setup.  Why create `n` but not use it?

It's also not _immediately_ clear to me that `rho0` is non-Hermitian - I can tell that it is after a second or so of thought, but maybe a comment above it wouldn't hurt.

qutip/tests/test_mesolve.py
```diff
@@ -920,5 +920,53 @@ def f(t, args):
...
+    options.store_states = True
+
+    result = mesolve(H, rho0, tlist, e_ops=[x], options=options)
+
+    imag_part = np.abs(np.imag(result.expect[0][-1]))
+    # Since we used an initial state that is not Hermitian, the expectation of
+    # x must be imaginary for t>0.
+    msg = ('Mesolve is not working properly with a non Hermitian density ' +
+           ' matrix as input. Check computation of '
+          )
```
jakelishman, line 948: Just for the future: these messages aren't actually necessary, though I get that you were just copying the style of the rest of the file (which _is_ a good thing to do in general!).

A failed assertion already means that something isn't working properly, and `pytest` will tell use exactly which assertion failed and show the code context, so the strings don't actually help too much.

jakelishman (commented):
Thank you for looking at this.  The changes within `mesolve` look sensible to me - I can't see any performance regressions, and the intermediate `Qobj` creation only sets Hermicity if it's known, and leaves it unset if `rho0` wasn't Hermitian, which is good.

You've followed the testing style of the file you were changing, which is the right way to approach a larger codebase.  We're currently half-way through transitioning a load of tests to a newer `pytest` style, so if you work with other tests, you may well see a very different style in them - in general, you can just stick to the same style that's in whatever file you're editing at the time.

Given that this PR is to fix a particular bug in `correlation`, please could you also add a reproduction for the bug in #1460 in order to ensure that it truly is fixed (I see no reason it shouldn't be from this), and to make sure that we catch any possible regressions in the future?

jakelishman:
The current test failure is unrelated and not your fault.  It's everyone's favourite `dnorm` again, and it's not because of any _new_ problems with it, it's just because we recently re-enabled the `dnorm` tests (see #1463) that had been lying dormant due to a problem in how they were written when we swapped to `pytest` as our runner.

qutip/tests/test_mesolve.py
AGaliciaMartinez: Thank you for reviewing the code! I changed it in the last commit to be more explicit with a comment and removed the unnecessary `n` variable.

qutip/tests/test_mesolve.py
AGaliciaMartinez: Thank you for the comment! I will take that into account in the future when writing tests.

AGaliciaMartinez:
> Given that this PR is to fix a particular bug in `correlation`, please could you also add a reproduction for the bug in #1460 in order to ensure that it truly is fixed (I see no reason it shouldn't be from this), and to make sure that we catch any possible regressions in the future?

I modified test_correlation.py to catch this error. It was previously using a non hermitian expectation operator which is why it did not raise an error. I changed it to use an hermitian operator. This means it would fail the test without the change in mesolve. 

However, I wonder now if the test should check for both Hermitian and non-Hermitian expectation operators. I did not add both tests because I did not want to make it slower. Should I make it to cover both cases?

qutip/tests/test_correlation.py
```diff
@@ -74,8 +75,8 @@ def test_correlation_solver_equivalence(solver, start, legacy):
     # broken, whereas if only one fails, then it implies that only that one is
     # broken.  We test that all solvers are equivalent by transitive equality
     # to the "me" solver.
-    base = correlation(H, start, None, times, c_ops, a.dag(), a, solver="me")
-    cmp = correlation(H, start, None, times, c_ops, a.dag(), a, solver=solver)
+    base = correlation(H, start, None, times, c_ops, x, x, solver="me")
+    cmp = correlation(H, start, None, times, c_ops, x, x, solver=solver)
```
jakelishman, line 79: This isn't _really_ a sufficient regression test for this bug.  This test is called "test correlation solver equivalence", and its purpose is just to test that all the solvers give compatible answers for the particular inputs; it doesn't actually test that the answer is correct.

It would be better to leave this test as it was, and put in a new test whose purpose is specifically to catch the error you're trying to fix.  As it is, you've removed a test of the behaviour where the operators were non-Hermitian, and we still need to test that case as well.

The example given in #1460 has a simple analytic solution, so you can write a test that explicitly compares each of the solvers' ability to reproduce that - use `np.testing.assert_allclose` to compare approximate floating-point equality.  You'll have to see what tolerance is sensible, but I'd guess an absolute tolerance of around 1e-7 should be good.

I'll not say the analytic solution here in case you want to work it out yourself, but if you don't want to, then just reply to this comment and I'll tell you it.

jakelishman: Oh, I just saw your comment about having modified the tests.

Yes, we should include both tests - _within reason_ the speed of tests is the secondary concern, after making sure they test everything we need them to test (note: we're not perfect at this yet).  In this case, calling any of the solvers with time-independent operators takes next-to-no time, so there's really no need to worry about it at all.

AGaliciaMartinez:
I added a test that compares `correlation_2op_1t` to its analytical solution. I was trying to match the `c_ops` of this new test to the ones appearing in `test_correlation_solver_equivalence`, which are:
`    c_ops = [np.sqrt(G1 * (n_th+1)) * a,
                    np.sqrt(G1 * n_th) * a.dag()]
`
However, the analytical solution I obtained for this case did not work unless I used a very small value for `n_th = 0.1`. I was wondering whether the analytical solution I got was wrong or whether this actually makes sense. I was thinking that the difference probably comes from the cut we use for the number of states (`N = 20`).

The new test for `correlation_2op_1t` should be very robust as it tests for a lot of different scenarios, including the one presented in  #1460. I was wondering whether these were too many cases.

qutip/tests/test_correlation.py
AGaliciaMartinez: > I'll not say the analytic solution here in case you want to work it out yourself, but if you don't want to, then just reply to this comment and I'll tell you it.

Thank you. Was good practice! :smile:

qutip/tests/test_correlation.py
```diff
@@ -74,8 +76,8 @@ def test_correlation_solver_equivalence(solver, start, legacy):
     # broken, whereas if only one fails, then it implies that only that one is
     # broken.  We test that all solvers are equivalent by transitive equality
     # to the "me" solver.
-    base = correlation(H, start, None, times, c_ops, a.dag(), a, solver="me")
-    cmp = correlation(H, start, None, times, c_ops, a.dag(), a, solver=solver)
+    base = correlation(H, start, None, times, c_ops, a, a.dag(), solver="me")
+    cmp = correlation(H, start, None, times, c_ops, a, a.dag(), solver=solver)
```
jakelishman, line 80: You've put these back the wrong way round.

qutip/tests/test_correlation.py
```diff
@@ -251,3 +253,64 @@ def test_hamiltonian_order_unimportant():
...
+start_list = [ 'fock']
+e_op_list = ['non-hermitian', 'hermitian']
+w_list = [1,2] # Fast and slow oscillations
+gamma_list = [1,10] # Small and high damping
+args_slow = list(product(solver_list, start_list, e_op_list, w_list, gamma_list))
+args += [pytest.param(*arg, marks = pytest.mark.slow) for arg in args_slow]
+
+
+@pytest.mark.filterwarnings("ignore::FutureWarning")
+@pytest.mark.parametrize(["solver", "start", "e_op", "w", "gamma"], args)
```
jakelishman, line 277: First: what's issuing a `FutureWarning` in this test?  We probably should be using whatever it is.

Second: you can actually just use several calls to `pytest.mark.parametrize` to achieve this.  For example:
```python
@pytest.mark.parametrize(['solver', 'state'], [
    pytest.param('me', _equivalence_fock, id="me-ket"),
    pytest.param('me', _equivalence_dm, id="me-dm"),
    pytest.param('me', None, id="me-steady"),
    pytest.param('es', _equivalence_fock, id="es-ket"),
    pytest.param('es', _equivalence_dm, id="es-dm"),
    pytest.param('es', None, id="es-steady"),
    pytest.param('mc', _equivalence_fock, id="mc-steady", marks=[pytest.mark.slow]),
])
@pytest.mark.parametrize("hermitian", [True, False], ids=["hermitian", "nonhermitian"])
@pytest.mark.parametrize("w", [1, 2])
@pytest.mark.parametrize("gamma", [1, 10])
def test_known_case(...):
```

qutip/tests/test_correlation.py
```diff
@@ -251,3 +253,64 @@ def test_hamiltonian_order_unimportant():
...
+def test_correlation_2op_1t(solver, start, e_op, w, gamma):
```
jakelishman, line 278: The name should be a bit more descriptive than this!  Maybe `test_2op_1t_known_cases`?

qutip/tests/test_correlation.py
```diff
@@ -251,3 +253,64 @@ def test_hamiltonian_order_unimportant():
...
+    a_op = x if e_op=="hermitian" else a
+    b_op = x if e_op=="hermitian" else a.dag()
+    c_ops = [np.sqrt(gamma) * a]
+
+    if start=='fock':
+        psi0 = _equivalence_fock
+    elif start=='coherent_dm':
+        psi0 = _equivalence_coherent
+    else:
+        psi0 = start
```
jakelishman, line 296: You can just pass these as the function arguments - see above how I used `pytest.param` to set a non-standard id for them.  Good thinking on trying to keep the tests looking nice, though.

qutip/tests/test_correlation.py
```diff
@@ -251,3 +253,64 @@ def test_hamiltonian_order_unimportant():
     backwards = qutip.correlation_2op_2t(H[::-1], start, times, times, [sp],
                                          sp.dag(), sp)
     np.testing.assert_allclose(forwards, backwards, atol=1e-6)
+
+
+# Paramters for the test
+solver_list = ['me', 'es']
+start_list = [ 'coherent_dm', 'fock', None]
```
jakelishman, line 260: I'm not sure you need to test the case with `None` - since you know your Hamiltonian and your collapse operator, your steady state will always just be `|0><0|`, right?  It might be better to parametrise over 2 different values of the Fock state (say 1 and 5), and parametrise over `dm=[True, False]` (tied to the solver in use) instead?

qutip/tests/test_correlation.py
AGaliciaMartinez: > First: what's issuing a FutureWarning in this test? We probably shouldn't be using whatever it is.

The `FutureWarning` should not be there. I copied from another test that uses legacy code that does issue this `FutureWarning`. I will remove it.

> Second: you can actually just use several calls to pytest.mark.parametrize to achieve this. For example:

I did not know about this! Thank you, I will implement it in this way.

qutip/tests/test_correlation.py
AGaliciaMartinez: I do actually think that the `None` case should be tested. Accepting `None` as the initial state is supposed to be a "feature" in which case we should test that it behaves properly.

Apart from `None`, the possible initial states could be a ket or a dm. Both cases are already covered with `coherent_dm` and `fock` which were chosen to be that way just because they already appeared in the file. I would not change these cases since I believe they already test thoroughly enough `correlation_2op_1t`, although if you still think it is necessary I have no problem with changing them.

nwlambert:
> I added a test that compares `correlation_2op_1t` to its analytical solution. I was trying to match the `c_ops` of this new test to the ones appearing in `test_correlation_solver_equivalence`, which are:
> ` c_ops = [np.sqrt(G1 * (n_th+1)) * a, np.sqrt(G1 * n_th) * a.dag()]`
> However, the analytical solution I obtained for this case did not work unless I used a very small value for `n_th = 0.1`. I was wondering whether the analytical solution I got was wrong or whether this actually makes sense. I was thinking that the difference probably comes from the cut we use for the number of states (`N = 20`).

I did a quick test and it seemed ok upto about n_th = 1 for N=20 and an atol of 2e-5.  
as an aside, I guess some minimal condition here is that the thermal occupation of the  cut-off state should be smaller than the tolerance, which i guess you can get out of the analytic expression for the thermal state. maybe.

jakelishman:
In theory the `correlation_2op_1t` function is only valid when you pass in the steady state solution of the Liouvillian.  However, we implement it just by calculating `<A(t) B(0)>` - it's not _immediately_ clear to me that that our output is necessarily correct when you don't pass in a steadystate (especially if `taulist` doesn't start from 0).  I _think_ it's fine in all cases, but perhaps Neill you can check me on that front?

I'd agree with Neill that truncation of the thermal state basis is likely the reason you were failing to get that to work exactly right.  The tolerance calculation will actually be a bit more involved because QuTiP returns a normed dm. That will cause all the values to shift by relatively the same amount, and then you've got to factor in the solver tolerance on top of those errors, but the principle is correct for sure.

qutip/tests/test_correlation.py
jakelishman: You're right that we need to test the steady-state functionality of `correlation_2op_1t`.  I'd thought it was tested somewhere else in the file already.  I'm fine with the test cases as-is.

jakelishman:
Ok, I'm going to merge this now because I'm gearing up for the 4.6 release and I want to include this bugfix.

PR #1480, opened 2021-03-31
Issue DeprecationWarning for eseries, essolve and ode2es

### LINKED ISSUE #1281: Consider deprecation of eseries and essolve
As part of the creation of the `core` package and the data layer, I'm hitting basically every single part of QuTiP, so I'm flagging parts we might want to consider for removal.

`eseries` is completely eclipsed by `QobjEvo`, so I think we can definitely get rid of that.  `essolve` is rather specific and not very used, I suspect.  We _could_ keep `essolve` without too much effort, but we should change the internals to use `QobjEvo` (though I also wouldn't be averse to removing it too).

@ajgpitch, @Ericgig, @nathanshammah.

jakelishman:
These are to be removed in QuTiP 5.0.  We use `FutureWarning` rather than `DeprecationWarning` because a lot (most?) QuTiP users run interactively in notebooks, which generally filter out the latter.  *edit*: changed to `DeprecationWarning` after review.

The "es" solver options in correlation gain warning filters, because they aren't being removed in 5.0; the small parts of functionality they need is just being folded into private helpers within the module.

The legacy correlation API had its deprecation warnings issued in 2014...  But nonetheless, we shouldn't actually remove functionality until a major release.

See fe5ef7d, d9ea79a for more details around the removals that have already been merged into `dev.major`.

Fix #1281

### DIFF
--- a/qutip/correlation.py
+++ b/qutip/correlation.py
@@ -1152,14 +1152,13 @@ def _correlation_es_2t(H, state0, tlist, taulist, c_ops, a_op, b_op, c_op):
     L = liouvillian(H, c_ops)
 
     corr_mat = np.zeros([np.size(tlist), np.size(taulist)], dtype=complex)
-    solES_t = ode2es(L, rho0)
 
+    solES_t = ode2es(L, rho0)
     # evaluate the correlation function
     for t_idx in range(len(tlist)):
         rho_t = esval(solES_t, [tlist[t_idx]])
         solES_tau = ode2es(L, c_op * rho_t * a_op)
         corr_mat[t_idx, :] = esval(expect(b_op, solES_tau), taulist)
-
     return corr_mat
 
 
@@ -1182,19 +1181,15 @@ def _spectrum_es(H, wlist, c_ops, a_op, b_op):
 
     # eseries solution for (b * rho0)(t)
     es = ode2es(L, b_op * rho0)
-
     # correlation
     corr_es = expect(a_op, es)
-
     # covariance
     cov_es = corr_es - a_op_ss * b_op_ss
-    # tidy up covariance (to combine, e.g., zero-frequency components that cancel)
+    # tidy up covariance (to combine, e.g., zero-frequency components that
+    # cancel)
     cov_es.tidyup()
-
     # spectrum
-    spectrum = esspec(cov_es, wlist)
-
-    return spectrum
+    return esspec(cov_es, wlist)
 
 
 # Monte Carlo solvers
--- a/qutip/eseries.py
+++ b/qutip/eseries.py
@@ -37,6 +37,8 @@ import numpy as np
 import scipy.sparse as sp
 from qutip.qobj import Qobj
 
+import warnings
+
 
 class eseries():
     """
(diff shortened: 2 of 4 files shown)


jakelishman:
So this test has had a failure in `test_diag_liou_mult` again, despite the merging of #1474.  I still believe the fix in that issue is correct in principle, but I'll investigate in more detail to determine the tightest allowable tolerances in another issue.

hodgestar:
If we set the appropriate stacklevel argument, then DeprecationWarnings should show up in notebooks (which don't filter out DeprecationWarnings from the `__main__` level).

FutureWarnings always show up, which will be the wrong thing if QuTiP is used inside, e.g., another application or package.

I haven't actually tried this, so hopefully Jupyter's behaviour matches the docs.

jakelishman:
Ahh, you're right.  I never managed to get it to work right before now, but I just tried again and managed to get `DeprecationWarning` to show up in all of `python warn.py`, direct calls from the IPython terminal and from Jupyter notebooks.  I think perhaps I learned about stacklevel _after_ the last time I tried to get that to work.  My reading of the docs before that had always just ended up with me assuming that somebody using a library interactively was an "end-user", so `FutureWarning` was appropriate.

I'll change this PR when I've got time - I needed to fix up a few warning controls in tests before it was ready anyway.

jakelishman:
@Ericgig @hodgestar, Eric and Simon: unless there's anything else that springs to mind, I'll pencil in the 4.6 release for later this week/weekend after this is merged to `master`?

Simon in particular: there's a couple of accessible places in `correlation` that make deprecated `ode2es` calls.  That particular functionality of the solvers isn't being removed in 5.0 (though maybe it should - `mesolve` is faster and better in almost every way), so they aren't themselves deprecated.  Right now they don't trigger `DeprecationWarning` popups if a user calls them (since the stacklevel is 2).  Despite that, should I _also_ wrap them in a warning filter to catch it?  I don't really know best practices here.

I kept postponing doing the 4.6 release because I'd keep thinking of little bits and bobs that I thought should go in, but at this point Boxi's (@BoxiLi) new pulse scheduler stuff is in and ready, the packaging and distribution methods are updated, and we'll have issued deprecation warnings for the things that we know are being removed.  Anything beyond this can start to form the basis of a 4.7 release in a few months' time, with the Floquet changes (they might be waiting on me at the moment) and possibly some new stuff out of GSoC.  Possibly that'll also include some deprecation warnings and other bits and bobs to do with packages getting moved out of qutip/qutip, and the changes to how the data-layer is going to work.

Ericgig:
Go for it. I am curious to see how it will go with the new distribution method. If it works well it will encourage us to make more frequent release.

BoxiLi:
Sounds good, @jakelishman shall I merge the last qip PR or do you still plan on making further comments?

jakelishman:
Oh, sorry, I thought it was already merged.  I've swapped my thing to an approval.

hodgestar:
@jakelishman I would vote for not wrapping them right now. We should stop using `ode2es` from the correlation functions (or continue to use it but no longer expose it to users). The DeprecationWarnings will hopefully help remind us to do that.

jakelishman:
The change to "not" using `ode2es` in the correlation functions is already in `dev.major` (it basically absorbs the few working parts of `ode2es` into a private function in `qutip.correlation`), so my plan had just been to leave the file as-is until that's merged to `master`.  My logic in that just feels a bit questionable though, hence my asking.

hodgestar:
@jakelishman Leaving it as is sounds fine to me. Deprecation warnings are filtered out by default when running Python scripts these days (unless one runs with `-X dev`) so that and the notebook filtering probably avoids the majority of cases where someone will be annoyed or confused by the warnings.

hodgestar (approved):

PR #1481, opened 2021-04-01
Merge.qobjevo

Ericgig:
**Description**
Merge `QobjEvo`, `cQobjEvo`, `QobjEvoFunc` into one cython class. (#1468)

This remove the need to have both keep 2 objects (`QobjEvo`, `cQobjEvo`) by solver and the need to support 3 classes.

It support any format for time-dependant system accepted by `mesolve`.
It keeps mostly the same interface as the present `QobjEvo`, the biggest changes being:
- `QobjEvo.mul_vec`, `QobjEvo.solver_set_args`,  `QobjEvo._cdc` removed.
- `QobjEvo.mul` renamed to `QobjEvo.matmul` and only accept `Qobj` input (from `Qobj`, `Data`, `np.array`)
- `QobjEvo.expect` only accept `Qobj` input (from `Qobj`, `Data`, `np.array`)
- `cQobjEvo.matmul_dense` and `cQobjEvo.expect_dense` removed, use the `_data` version instead, they have optimization for `Dense` included.
- Some of `QobjEvo`'s attributes are no longer accessible: `cte`, `ops`. Instead of `A.cte.dims`, simply use `A.dims`. Same for `shape`, `issuper` and `isoper`.
- `shift` renamed to `_insert_time_shift`. It is meant for use in correlation only so I renamed to a longer name. 

Internally, some of the logic have been moved to an `Element` class (pair `Qobj`, `Coefficient`) to support function format.
`matmul` is there to support the bloch redfield tensor. Function based `QobjEvo` reuse calls to the function: 
With `cdc = QobjEvo(f).dag() * QobjEvo(f)`, `cdc(t)` will call `f` only once. This remove the need of special code of `QobjEvoFunc` in superoperator. 

##### Leftover issues
The class keep a copy of `args` only for the `__call__(self, t, args)` method, since the coefficient have their own copy of `args`. It could be removed if we remove that functionality.
While coefficient with different `args` can be mixed, function cannot.


qutip/tests/core/test_qobjevo.py
```diff
@@ -121,376 +149,234 @@ def _assert_qobjevo_different(obj1, obj2):
...
+        _assert_qobjevo_equivalent(copy, qevo)
+        assert copy is not qevo
+
+    @pytest.mark.parametrize('bin_op', [
+        pytest.param(lambda a, b: a + b, id="add"),
+        pytest.param(lambda a, b: a - b, id="sub"),
+        pytest.param(lambda a, b: a * b, id="mul"),
+        pytest.param(lambda a, b: a @ b, id="matmul"),
+        pytest.param(lambda a, b: a & b, id="tensor"),
+    ])
```
jakelishman, line 192: If you find yourself reusing a parametrisation a lot, you can encode it up top in a fixture, e.g.
```python
@pytest.fixture(params=[
    pytest.param(lambda a, b: a + b, id="add"),
    ....
])
def bin_op(request):
    return request.param
```

qutip/tests/core/test_qobjevo.py
```diff
@@ -121,376 +149,234 @@ def _assert_qobjevo_different(obj1, obj2):
...
+        qevo = QobjEvo(*getattr(pseudo_qevo, coeff_type)())
+        assert isinstance(qevo(0), Qobj)
+        assert qevo.isoper
+        assert not qevo.const
+        assert not qevo.issuper
+        _assert_qobjevo_equivalent(pseudo_qevo, qevo)
+
+    @pytest.mark.parametrize('coeff_type',
+                             ['func_coeff', 'string',
+                              'spline', 'array', 'logarray'])
```
jakelishman, line 168: Why re-parametrise `coeff_type` with the same values that are already defined?

qutip/tests/core/test_qobjevo.py
```diff
@@ -65,51 +92,52 @@ def _cplx(t, args):
...
-    pytest.param("no_args", id="no_args"),
-    pytest.param("mixed_tlist", id="mixed_tlist"),
-    pytest.param("qobjevofunc", id="qobjevofunc"),
+    pytest.param(real_qevo, id="real"),
+    pytest.param(herm_qevo, id="hermitian"),
+    pytest.param(cplx_qevo, id="complex"),
 ])
-def extra_form(request):
+def pseudo_qevo(request):
     return request.param
```
jakelishman, line 124: Not sure if you intended this, but you'll use the same three random pseudo-QobjEvo objects all the way through - the randomness only happens once.

jakelishman (commented):
I'm happy to see the back of `CQobjEvo` and its million variants.  This looks like a much neater way of handling things, and the unification of `QobjEvo` and `CQobjEvo` will help keep the codebase rather tidier, both in lines of code and in mental load.

One thing I just thought of - does the new `QobjEvo` pickle itself fine?  I don't remember seeing a `__reduce__` method.

qutip/core/__init__.py
```diff
@@ -2,8 +2,7 @@
 from .coefficient import *
 from .interpolate import *
 from .qobj import *
-from .qobjevo import *
-from .qobjevofunc import *
+from .cy.qobjevo import *
```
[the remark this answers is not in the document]
Ericgig: Agreed, but outside the scope of this PR.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+                                 "`f(t: double, args: dict) -> Qobj`")
+            self.dims = qobj.dims
+            self.shape = qobj.shape
+            self.funcQEvo = set([Q_object])
+            self.elements.append(_FuncElement(Q_object, qobj, args))
+
+        else:
+            raise TypeError("Format not understood")
+
+    def __call__(self, t, args={}):
```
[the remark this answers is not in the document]
Ericgig: I am thinking of removing the option and have `__call__` accept only `t`. Since each coefficient has a copy of args, `QobjEvo` keep a copy only for this functionality. It usage should be rare enough that asking the user to update args with arguments seems reasonable.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+        res = self.copy()
+        res += other
+        return res
+
+    def __radd__(self, other):
+        if not isinstance(other, (Qobj, QobjEvo, numbers.Number)):
+            return NotImplemented
+        res = self.copy()
+        res += other
+        return res
```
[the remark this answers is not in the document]
Ericgig: Yes, in cython's docs, to support both pre and post 3.0, they suggest have both `__op__(left, right)` and `__rop__(self, other)` or use `c_api_binop_methods=True`. I used to former since the `__rop__` methods already existed in `QobjEvo`. We can choose to use any of the method for Qutip. `c_api_binop_methods=True` may be safer since less code and tests are needed.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+                    coeffs[i] = coeffs[i] + element.coefficient
+                    break
+            else:
+                qobjs.append(element.qobj)
+                coeffs.append(element.coefficient)
+        for qobj, coeff in zip(qobjs, coeffs):
+            cleaned_elements.append(_EvoElement(qobj, coeff))
+        return cleaned_elements
+
+    def compress(self):
```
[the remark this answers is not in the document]
Ericgig: It is used by some user. But it's probably not needed in `add`.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+        coeff_elements = self._compress_merge_qobj(coeff_elements)
+        cleaned_elements += coeff_elements + func_elements
+
+        self.elements = cleaned_elements
+
+    ###########################################################################
+    # properties                                                              #
+    ###########################################################################
+    @property
+    def num_obj(self):
```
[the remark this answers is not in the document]
Ericgig: I removed `self.elements` from python interface. I don't see any need for easy access to it.
`num_obj` is used in tests. But yes I forgot to rename it.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+        expect : float or complex
+            `state.adjoint() @ self @ state` if `state` is a ket.
+            `trace(self @ matrix)` is `state` is an operator or operator-ket.
+        """
+        if not isinstance(t, (int, float)):
+            raise TypeError("the time need to be a real scalar")
+        if not isinstance(state, Qobj):
+            raise TypeError("A Qobj state is expected")
+        if not (
+            (self.dims[1] == state.dims[0]) or
```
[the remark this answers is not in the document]
Ericgig: This is the same test as in `expect.py`... adding a test here but we should not forget to correct it there also.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+                self.args = (<QobjEvo> Q_object).args
+                self.elements = (<QobjEvo> Q_object).elements
+                self.funcQEvo = (<QobjEvo> Q_object).funcQEvo
+            if args:
+                self.arguments(args)
+            return
+
+        self.elements = []
+        self.funcQEvo = set()
+        self.func_qobj = {}
```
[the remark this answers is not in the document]
Ericgig: When I added `self.func_qobj` it made `self.funcQEvo` redundant.
It is a dict of `{func: funct(t)}`so that the like of `op * op.dag()` only call each function once.

qutip/core/superoperator.py
```diff
@@ -80,42 +80,38 @@ def liouvillian(H, c_ops=[], data_only=False, chi=None):
...
         else:
             raise TypeError("Invalid type for Hamiltonian.")
+        L += sum(lindblad_dissipator(c_op, chi=chi_)
+                 for c_op, chi_ in zip(c_ops, chi))
+        return L
+
+    if H.isoper:
+        op_dims = H.dims
+        op_shape = H.shape
+    elif h.issuper:
```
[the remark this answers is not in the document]
Ericgig: This is wrong... Adding a test to catch it in the future.

qutip/core/superoperator.py
```diff
@@ -240,10 +172,10 @@ def lindblad_dissipator(a, b=None, data_only=False, chi=None):
     D : qobj, QobjEvo
         Lindblad dissipator superoperator.
     """
+    if a.issuper:
+        return a
```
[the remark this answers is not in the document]
Ericgig: `liouvillian` and `mesolve` accept `H` and `c_ops` being already superoperators, only `lindblad_dissipator` did not... We can keep it that way.

qutip/tests/solve/test_stochastic_me.py
```diff
@@ -288,7 +293,7 @@ def f_dargs(a, args):
     return args["expect_op_3"] - 1
 
 
-@pytest.mark.xfail(reason="not working yet")
+@pytest.mark.skip(reason="not working yet")
```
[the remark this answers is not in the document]
Ericgig: The test can take 10 sec to fail and the implementation closer to non existent than broken in the present version. For sanity, I would prefer skip it totally until it is fixed, which should come not that far.

qutip/tests/core/test_qobjevo.py
Ericgig: To remove one type: `func_call`.  `skip` sound wrong since we would be skipping a test that should not exist. If there is a feature in pytest that skip a param without listing it in the test list this is what should be used.

qutip/tests/core/test_qobjevo.py
Ericgig: Yes this is intended. Recreating the Qobj would only take time without any effect on the validity of the tests.

qutip/core/__init__.py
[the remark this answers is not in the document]
jakelishman: I brought it up only because you were adding a "new" Cython file - this one could just go into `qutip/core`?

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: The way I think about `args` at different scoping levels, I like having it available to pass to `__call__` here, but I'm not too beaten up if you think it would be much better to remove it.  It'd probably be good to have benchmarks to know if it's actually worthwhile doing, though, since we're changing the API.

Regardless, as long as it's in here, a default value of `{}` is unsafe - it's always unsafe in Python to have a default parameter be mutable.

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: Ok, sounds reasonable to me.  As it stands right now, I wouldn't at all be surprised if Cython 3.0 is a breaking change for us, but it's a good idea to start getting ready.

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: Oh, good catch then - that's absolutely a nasty bug waiting to happen (though the circumstances in which it would trigger are particularly arcane...)

qutip/core/superoperator.py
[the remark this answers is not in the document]
jakelishman: `mesolve` absolutely should accept `H` as a Liouvillian, but it seems misleading to have `liouvillian` accept a superoperator?  Do you know if there's a reason we have `liouvillian` do that?

qutip/tests/solve/test_stochastic_me.py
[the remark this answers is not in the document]
jakelishman: 10s isn't a huge amount of time in the scheme of things, really.  Would a successful test (once the feature is complete) take 10s as well, or will it be significantly shorter?  If it's the same, we should probably leave it as `xfail`.  If it'll be shorter, then `skip` is fine, though let's change the message to say "not implemented yet" to be clearer.

qutip/tests/core/test_qobjevo.py
jakelishman: I also see now that you did this elsewhere in the file.

qutip/tests/core/test_qobjevo.py
jakelishman: Ah, I missed that one of them was missing.  I thought it was the same list.  How come we don't need to test that one in this case?

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
Ericgig: The issue I see is if you have coefficient with different value of the same key in args, they would lose there unique values  with this...  `Coeffcient.__call__` also accept an input args but it overwrite the dict and does not come back to the previous state. I will keep the function by making a copy, this would ensure that the original args are untouched, but it may be slower.

qutip/core/superoperator.py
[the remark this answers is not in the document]
Ericgig: I would guess its so `mesolve` can always just call `liouvillian(H, c_ops)`, it was done in `25303b6` 8 years ago. Now is the time to change it if we want to. It was not properly tested (the `h` leftover passed...).
I am bothered with having `liouvillian` and `lindblad_dissipator`not matching.

qutip/tests/core/test_qobjevo.py
Ericgig: This is for the creation of `QobjEvo` with `Qobj * Coefficient`. Function returning a Qobj cannot be created this way.

Ericgig:
About pickling. `QobjEvo` doesn't contain any cython only data and cython autopickling is working fine.
There is a test for it. (But not for windows, there is still a risk there).
But function used in `linear_map` or in coefficient etc. may cause problem.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+            raise TypeError("A Qobj state is expected")
+        if not (
+            (self.dims[1] == state.dims[0]) or
+            (self.issuper and self.dims[1] == state.dims)
+        ):
+            raise ValueError("incompatible dimensions " + str(self.dims) +
+                             ", " + str(state.dims))
+        out = self.expect_data(t, state.data)
+        if fabs(out.imag) < 1e-15:
+            return out.real
```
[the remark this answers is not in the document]
Ericgig: QobjEvo cannot detect their hermiticity without looking into Coefficient... The options I see are:
- Have an hermitian flag manually set at creation (and trust it blindly).
- `QobjEvo.expect` always return a complex. 
- Check for if the expectation value is effectively real. Here I should have used a relative test: `abs(out.imag/out.real) < 1e-15`, still not ideal since `0` will always be complex.
- Compute the hermiticity of op(t) each time. (But this would be too slow)

qutip/core/cy/coefficient.pyx
```diff
@@ -24,11 +24,14 @@ cdef class Coefficient:
...
-        """update args and return the
+    def __call__(self, double t, dict args=None):
+        """Update args and return the coefficient value at `t`.
         """
+        cdef Coefficient cp
         if args:
-            self.arguments(args)
+            cp = self.copy()
+            cp.arguments(args)
+            return cp._call(t)
```
hodgestar, line 34: It's odd that we have to create a new `Coefficient` to run a method on this instance. The question is what is `self.args`? Is it the fixed arguments for this Coefficient object? Or is it the default arguments only? And what are the args to `__call__`? Are they all the arguments, or just additional ones?

One approach would be to change `_call` to take `args` as a paramter and do `all_args = self.args.copy(); all_args.update(args)` (with better variable names, of course). Or to pass the full args to `_call` and do the argument updating here.

Thoughts?

qutip/core/qobj.py
```diff
@@ -1802,11 +1812,3 @@ def ptrace(Q, sel):
...
         raise TypeError("Input is not a quantum object")
     return Q.ptrace(sel)
-
-
-# TRAILING IMPORTS
-# We do a few imports here to avoid circular dependencies.
-from qutip.core.superop_reps import to_choi
-from qutip.core.superoperator import vector_to_operator, operator_to_vector
-from qutip.core.tensor import tensor_swap, tensor
-from qutip.core import metrics as mts
```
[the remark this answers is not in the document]
hodgestar: We can probably tell CodeClimate to be quiet if needed. :)

Separating things to avoid the circular dependencies would be awesome, but +1 on moving the imports to where they're used.

qutip/core/superoperator.py
```diff
@@ -52,24 +52,22 @@ def _map_over_compound_operators(f):
...
-    and a ``list`` of collapse operators. Like liouvillian, but with an
-    experimental implementation which avoids creating extra Qobj instances,
-    which can be advantageous for large systems.
+    and a ``list`` of collapse operators.
 
     Parameters
     ----------
     H : Qobj or QobjEvo
-        System Hamiltonian.
+        System Hamiltonian or Hamiltonian composant of a Liouvillian
```
hodgestar, line 70:
```suggestion
        System Hamiltonian or Hamiltonian composed of a Liouvillian
```

Ericgig: No, the Liouvillian is build from the Hamiltonian not the other war around.
If `composant` is confusing (or am I forcing a French word here?), how about `term` or `part`

jakelishman: "composant" isn't an English word (though it almost feels like it should be...) - you probably mean "component"

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: So my main point about the kwarg default needs to be `args=None` I think is fixed, so I'm fine on this thread.

For the other bits: I think my actual concern is more about how `args` is handled _everywhere_.  Currently, you can pass an `args` dictionary at various points, which feels like you're defining `args` within that scope.  However, what really happens is that an underlying `Coefficient` object gets mutated (no matter how many references may exist to it), and the values are never restored afterwards.  Now that I think about it, this is already causing behaviour that's incorrect:
```python
>>> my_td = qutip.coefficient('w + 0*t', args={'w': 1})
>>> my_obj = qutip.QobjEvo([[qutip.qeye(2), my_td]])
>>> my_obj(0) == qutip.qeye(2)
True
>>> # now let's do some evolution
>>> H = [[qutip.sigmax(), my_td]]
>>> res = qutip.mesolve(H, qutip.basis(2, 0).proj(), [0, 1], args={'w': 2})
>>> # now let's look at my_obj again
>>> my_obj(0) == qutip.qeye(2)
False
>>> my_obj(0) == 2 * qutip.qeye(2)
True
```

I think we really very much need to avoid implicit mutation like this.  Perhaps we need to discuss/design this a bit more, because this unexpected mutation absolutely will cause people problems, since we're really billing `Coefficient` as compile once, use many times.

To hook it back to the point you bring up, though: having `__call__` mutate any accessed `Coefficient` seems quite clearly wrong (hence you putting in the `copy`), but to me it's equally wrong to have `mesolve` (and anything else) do the same thing as well.  This is beyond the scope of this PR for sure, but it's something I think will really bite us and users if we leave it as-is.

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: Yeah, best leave `compress` for users, but having it called on every operation will be a quadratic slowdown that we'll definitely not remember in the future.

qutip/core/cy/coefficient.pyx
jakelishman: As I see it, there are two problems at play - one of call speed (`Coefficient`'s primary purpose is C-level fast calling) and one of mutability.  We attempt to avoid reads to an `args` dictionary within each call, but instead read it once at the most recent time they're set, and bind them to named variables within the C struct (created by our codegen).  Then call speed is faster from them, since the typical use-case will involve many many calls to the `Coefficient` via something like `mesolve` before any `args` are changed.

Directly calling `Coefficient` while passing `args` is not going to happen very much at all, but still ought to be supported just for logical consistency within the library - there are various points at which you can define the `args` in functions that indirectly call `Coefficient`, so it's only natural that direct calls should work too, and much easier for the user if we handle the state-dependent boilerplate.

That said, I mentioned in a different comment that the current implementation really has mutating-state problems (see https://github.com/qutip/qutip/pull/1481#discussion_r609744805).  This `copy` means that we're expending more GC cycles to collect stuff (and hence we shouldn't be using this form in inner loops, but we already don't), but it does maintain the expected scoping rules.  If you have another reference to this same coefficient, _this_ way won't allow a single apparently-pure function call to have side effects.

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: I wouldn't remove `args` from the signature of any of these points, I'd make it so that _nothing_ except a user's direct call to `Coefficient.arguments()` can mutate an object that they have a reference to.  This becomes a little more complicated when dealing with `QobjEvo` - a user's call to `QobjEvo.arguments()` seems ok to mutate held `Coefficient` objects, but passing the `QobjEvo` to `mesolve` shouldn't risk any mutation at all.

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: Yeah, I get all the problems with determining a hermicity for `QobjEvo`. I think anything we do to make it "knowable" will be pretty useless at best, and endlessly frustrating for users at worst - it'll be hard to do right even in things like `mesolve`.

I'm not sure I can see how any of the `Qobj` properties can work on `QobjEvo`, to be honest (e.g. `isherm`, `istp`, `isunitary` etc).  We probably shouldn't define them.

For now, I guess my suggestion is to a) lookup the tolerance in a settings object somewhere and b) swap it to a relative tolerance (with a little care when the real part is zero).

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
jakelishman: Ok cool, that seems sensible.

qutip/core/superoperator.py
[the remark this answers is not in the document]
jakelishman: I'd have `liovillian` and `lindblad_dissipator` match by having them both reject being passed superoperators, but I'm not particularly strongly inclined in either direction.  I agree they should match, for sure.

qutip/solve/_stochastic.pyx
```diff
@@ -1916,35 +1916,36 @@ cdef class SMESolver(StochasticSolver):
                                         np.ndarray[complex, ndim=1] guess) except *:
         # np.ndarray to memoryview is OK but not the reverse
         # scipy function only take np array, not memoryview
-        spout, check = sp.linalg.bicgstab(self.imp(t, data=1).as_scipy(),
-                                          dvec, x0=guess, tol=self.tol, atol=1e-12)
+        spout, check = sp.linalg.bicgstab(self.imp(t).data.as_scipy(),
```
jakelishman, line 1919: This still has the unsafe assumption that `data` will always be `CSR` I think?

qutip/tests/core/test_qobjevo.py
jakelishman: Ok cool, sounds sensible.

qutip/core/cy/coefficient.pyx
Ericgig: If you have a `Coefficient`, say `sin(w*t)`, it store it's value of `w` so it can be called with `coeff(t)`.
But you can call it with different value of `w` by passing it in an `args` dictionary.
This only applied to that single call and leave the object `args` untouched.

I feel this feature will be mostly used in tests and maybe some debugging by users trying to see if their `QobjEvo` are behaving as expected...
So speed should not matter much here and making a copy is the simplest way to ensure it does not to have any side effect.
If we have a need for efficiency here, we can move the logic in `_call`.

jakelishman:
After second review, this generally looks pretty solid to me, and (bar the one or two minor outstanding comments) I'd be happy to merge it.

I think we possibly need to take a deeper look at where and when it's acceptable for us to mutate the `args` stored in a `Coefficient`, particularly when we don't know how many references are being held to a given object, but that's not in the scope of this PR.

qutip/solve/_stochastic.pyx
Ericgig: I made the conversion when creating `self.imp` in `stochastic.py`. So unless someone try to use this without using the python interface it should be safe.

qutip/core/options.py
```diff
@@ -21,6 +21,10 @@ class CoreOptions:
     atol : float {1e-12}
         general absolute tolerance
 
+    atol : float {1e-12}
```
jakelishman, line 24: rtol

jakelishman (dismissed):

qutip/core/superoperator.py
```diff
@@ -60,7 +60,7 @@ def out(qobj):
     return out
 
 
-def liouvillian(H, c_ops=[], data_only=False, chi=None):
+def liouvillian(H=None, c_ops=None, data_only=False, chi=None):
```
hodgestar, line 63: We should update the docstring to explain what happens when `H` and `c_ops` are `None`.

hodgestar: And probably what `chi` and `data_only` are while we're here.

qutip/core/cy/coefficient.pyx
hodgestar: Maybe then a better interface would be the following:
```python
def replace(self, args=None):
    cp = self.copy()
    if args:
        cp.arguments(args)
    return cp

def __call__(self, t):  # note lack of args
      ...
```
Then users can either call `.arguments(...)` if they need to update the arguments, or `.replace(...)` if they need a new Coefficient instance with update arguments and `__call__` is always fast because the arguments have already been updated before.

Thoughts? Also not necessary for this to happen in this PR perhaps -- it can just be an idea for the future.

hodgestar: P.S. `.replace(...)` is inspired by `datetime.datetime.replace`.

qutip/core/cy/qobjevo.pxd
```diff
@@ -0,0 +1,23 @@
+#cython: language_level=3
+
+from qutip.core.data cimport CSR, Dense, Data
+from qutip.core.data.base cimport idxint
+
+cdef class QobjEvo:
+    cdef:
+        list elements
+        set funcQEvo
```
hodgestar, line 9: Unrelated to this PR, but why is `funcQEvo` a set?

qutip/core/cy/qobjevo.pxd
```diff
@@ -0,0 +1,23 @@
...
+        readonly list dims
+        readonly (idxint, idxint) shape
+        int _issuper
+        int _isoper
+        double _shift_dt
+        dict func_qobj
```
hodgestar, line 15: And why is `func_qobj` a dict?

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1101 @@
...
+    cpdef void call(self, double t, dict function_qobj):
+        pass
+
+    def __mul__(left, right):
+        cdef _CteElement base
+        cdef object factor
+        if type(left) is _CteElement:
+            base = left
+            factor = right
+        if type(right) is _CteElement:
```
hodgestar, line 95: Could this be an `elif`?

qutip/core/cy/qobjevo.pxd
jakelishman: I think it actually just needs removing from the definition file - it already was removed from the implementation.

hodgestar (commented):
Apologies for the delayed and slightly sporadic comments. I'm not an expert on this yet, so I'm happy for this to be merged with Jake's approval. My questions were mostly just me getting to grips with things or small ideas for improvements.

qutip/core/superoperator.py
Ericgig: I can't really explain the `chi`...
There are a few thing I don't understand with `lindblad_dissipator` (not trace preserving when `b` is given).
Now that `Qobj` creation is fast, `data_only` could be removed, but adding a description for now.

qutip/core/cy/coefficient.pyx
jakelishman: I like it, and personally I'd go all the way and have `Coefficient` be a completely immutable type. I'd remove `Coefficient.arguments` and only have the suggested `Coefficient.replace` - I'm always really put off by having to deal with mutability in any form, because you can almost always end up with funny cases where stuff can mutate from underneath you (like I demonstrated in that comment I linked above).  In stuff I write for myself, I keep everything immutable unless it _absolutely_ has to be mutable for performance.  Maybe it's just because I prefer it, but I do find that immutability-by-default eliminates a whole class of side-effect-dependent bugs that OOP often struggles with.

I think that the way we now manage compiled Cython coefficients means that the copy operation should be very fast, and this is never going to be inner-loop code anyway, so there's likely little-to-no measurable performance gain from allowing the mutation if we do it right.

Ericgig: A few things I see:
- There is two situation that `Coefficient.arguments` could be in a loop: feedback and control.
For feedback, the user should expect an impact on speed: time to create a `Qobj` for the state is similar to copying a `Coefficient`.
For control, they want to be able to step through the evolution changing the `args` at each timestep.
- Also array based `Coefficient` will be the slowest to `replace` with the copy the arrays but have not `args` to update.
- Copying string array is slow presently: it uses `pickle.loads(pickle.dumps(self))`
- I am not the best with names but `replace` feels wrong. What are we replacing?

Looking at timing, with a proper `copy` for string coefficient and `replace` returning a copy only for mutable ones (`return self` for array based ones), this should be mostly unnoticeable even in evolution with feedback. (If we see a performance issue there later, we could always remake `arguments` as a `cdef` method.)

jakelishman: Array coefficients won't need to copy the arrays to replace the arguments; as long as we maintain that the arrays are never mutated after the coefficient is created, the only references to them will be held inside `Coefficient`, so there's no problem there.  If copying of string arrays is slow enough to be bottleneck, we can look into speeding it up - the internal C callable has already been Cythonised and compiled, so at its heart the only work that needs to be done is to allocate a new Python object and copy a few references.

If we move to a model where `Coefficient` is immutable, then `replace` is replacing one or more scope variables - it's got similarities with the `_replace` method on named tuples:
```python
import collections
Record = collections.namedtuple('Record', ['name', 'age'])
jake = Record('jake', 26)
twin = jake._replace(name='john')
```
and a handful of other standard library types.  I always preferred `with`, personally, but we can't use that because it's a keyword :(

Ericgig: I am sold on making `Coefficient` immutable, and `QobjEvo`'s `_BaseElement` with it. 
The copying of string coefficient is 50x the copying of function coefficient, so it is worth to rework it, but we need to rework the codegen anyway to remove the arguments function.

With the logic of `replace`, it feels like it should do more that just act on the `args` and be able to update the `tlist` or the like...

jakelishman: Oh yeah, good idea - it can be very like the `namedtuple` one then.  The signature becomes
```python
def replace(self, *, arguments=None, tlist=None, ...):
    ...
```
so the calls look like
```python
coeff = qutip.coefficient(...)
coeff.replace(arguments={...})
```

Ericgig:
I made the change for immutable `Coefficient` and `_Elements`. 
`replace` is about 200ns slower than the previous inplace `arguments`.
With `_Elements` _immutable_, `QobjEvo.copy` is a shallow copy of the `elements` list. 
But the `_BaseElement._call` is not returning the (data, coeff) pair, it stores it to save on the creation of a python object.
This it may not safe with threading: if 2 copies of a `QobjEvo` are called at the same time with different `t`, both could return the same result. I am not sure how good is the GIL in this case.
I don't expect it to be an issue, in the rare case threading would be used the options of deepcopy is available `QobjEvo.copy(deepcopy=True)`.

ps. I won't be active for Qutip next week to respond to your review/comments.

jakelishman:
I've lost the plot a bit with a whole load of changes to `Coefficient` now merged into this PR.  Is there a chance we can split that out into its own PR?

I would expect `QobjEvo.__call__` to be thread-safe/re-entrant. The fact that coefficients seemingly don't support this immediately is worrying to me.  A simple call like that changing "global" state of the object is dangerous; it has a habit of leaking out even in single-threaded applications, just like the bug I described previously with a call to `mesolve` modifying an existing `QobjEvo` in place.  In this case I don't see an immediate bug, but it's very non-obvious behaviour and it would be easy to accidentally introduce one in it in the future (or maybe I just can't think of one now).

Having `_BaseElement` be mutable makes all time-dependent operations of `QobjEvo` unresolvably thread-unsafe without copying on every operation.  That said, looking again, I'm not sure I understand `_BaseElement`: it seems like `_EvoElement` and `_FuncElement` do entirely different things on `call`?  What's the purpose of it, and if we put all speed considerations aside what would the signature and behaviour of `_BaseElement._call` be?

Could you write some docstrings on all the elements to explain their use-cases as well, so we've got it all stored for the future?

Ericgig:
I can move a part of the changes in coefficient to another PR, but not all of them.
I removed `shift` and `norm` from `Coefficient`, they become unused only with the new `QobjEvo`.
I will move the immutable `Coefficient` changes to another PR.

The signature of `_BaseElement._call` would be `cdef (double complex, Qobj) _call(self, double t, dict function_qobj)`.
`_EvoElement` would return `(coefficient(t), qobj)` with `qobj` fixed.
`_FuncElement` would return `(1., func(t, args))` with `func` a function based time dependant system.
I also keep a copy of the `Data` instead of reading it from the `Qobj` each time, since it's a python object.

For example, in `expect` I do:
```
element._call(t, ...)
out += expect(element.data, state) * element.coeff
```
If another call to `_call` arrive between the commands, the output would be wrong.
But this call all be solved by splitting `_call`...
`out += expect(element.data(t, ...), state) * element.coeff(t, ...)`

Ericgig:
I moved the Coefficient changes to #1507, all should be thead-safe and _[...]Element are better documented.
#1507 should be merged before this.
The commit history somewhat a mess... I suggest squash and merge.

jakelishman (commented):
I have some questions about clarifying the docstrings, mostly.  About the commit history: this is a pretty big PR.  After #1507 is merged, if you've got time/are able to, it might be good if you could interactive rebase the commit history on top of it, and merge/tidy up the commits to tell a sensible story.  It'd be super super difficult to work out what changes were being made and why a year in the future, if all we've got is one +2000/-3000 commit.

qutip/core/cy/qobjevo.pyx
```diff
@@ -11,28 +11,32 @@ from .. import Qobj
...
+
 
 # Changing settings.core['rtol'] will not automatically update HERM_RTOL.
 def set_herm_tol(rtol):
+    global HERM_RTOL
     HERM_RTOL = rtol
 
+
+def get_herm_tol():
+    return HERM_RTOL
```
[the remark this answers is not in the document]
Ericgig: Reading a value from a nested python object that almost never change in a function that can be inside the inner-loop in solver feel a waste. 
I would like options values to have properly-like behaviours and make the needed change when updated but I should probably wait for that to be implemented before doing this...

qutip/core/cy/_element.pyx
```diff
@@ -0,0 +1,341 @@
...
+                _data.matmul(self.data(t), state, self.coeff(t))
+            )
+
+    def arguments(_ProdElement self, new_args, fElem_safe):
+        return _ProdElement(
+            self.left.arguments(new_args, fElem_safe),
+            self.right.arguments(new_args, fElem_safe),
+            self.transform.copy(),
+            self.conj
+        )
```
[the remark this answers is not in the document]
Ericgig: It act as `replace`, I just can't like the name... 
And there is no use to replace anything other than `args`.

qutip/core/cy/_element.pyx
```diff
@@ -0,0 +1,341 @@
...
+                            self.transform.copy(), self.conj)
+
+    def __matmul__(left, right):
+        return _ProdElement(left, right, [])
+
+    def linear_map(self, function, conjugate=False):
+        return _ProdElement(
+            self.left, self.right,
+            self.transform + [function],
+            self.conj ^ conjugate
```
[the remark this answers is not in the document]
Ericgig: Changed the type from `bint` to `bool`.

qutip/core/cy/_element.pyx
```diff
@@ -0,0 +1,341 @@
...
+        )
+
+
+cdef class _ProdElement(_BaseElement):
+    """
+    Product of `_FuncElement`/`_MapElement` with other `_BaseElement`
+    Include a stack of linear transformation from the result of the product.
+    op = QobjEvo(f) * qobj1 ->
+        _ProdElement(_FuncElement(f, args), _CteElement(qobj1))
+    """
```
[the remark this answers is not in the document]
Ericgig: There is no need to use for `_ProdElement` in those cases: `[qobj1, coeff1] @ [qobj2, coeff2] = [qobj1 @ qobj2, coeff1 * coeff2]`. With function, we can only do that product after we know the time at which it is needed, therefore we need to stack to operation.

qutip/core/cy/_element.pyx
```diff
@@ -0,0 +1,341 @@
...
+# qevo.elements = [_FuncElement(H0, args)]
+# qevo.dag().elements = [_MapElement(_FuncElement(H0, args), [dag])]
+# (qevo * qevo).elements = [_ProdElement(_FuncElement(...), _FuncElement(...))]
+#------------------------------------------------------------------------------
+cdef class _FuncElement(_BaseElement):
+    """
+    Block containing varying Qobj.
+    Used with QobjEvo build from a function with the signature:
+        func(t, args) -> Qobj
+    Each `_FuncElement` contain an immutable pair (func, args).
```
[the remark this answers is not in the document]
Ericgig: Well... yes? 
`_FuncElement` is the only element containing a function returning a `Qobj`.
If you are asking about the immutable part, they are all immutable, but only this one contain a `(func, args)` pair.

qutip/core/cy/_element.pyx
```diff
@@ -0,0 +1,341 @@
...
+        # Here, `f2` is still one instance
+        for old, new in fElem_safe:
+            if old is self:
+                return new
+        new = _FuncElement(self.func, {**self.args, **new_args})
+        fElem_safe.append((self, new))
+        return new
+
+
+cdef class _MapElement(_BaseElement):
```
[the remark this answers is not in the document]
Ericgig: It does not derive from `_FuncElement` but contain one. There is little to reuse from `_FuncElement` that is not already in `_BaseElement`.

hodgestar:
@Ericgig Is this the next PR to review in the solver chain? If so, would you merge in dev.major and I can start reviewing?

jakelishman:
I've just made #1574 to move our testing to Travis.  This will cause a merge conflict in `setup.py` for you when it's merged, but I think that will be the only conflict - you should be able to rebase on top cleanly.  The reason for the conflict is because the "new" (`master`) `setup.py` discovers all the Cython modules to be compiled, and doesn't need to have them specified, so you shouldn't actually need to make any changes to it to have it work correctly.

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
hodgestar: Can we resolve this thread now that `Coefficients` are immutable?

qutip/core/cy/coefficient.pyx
hodgestar: This all got implemented in #1507 -- thanks!

Ericgig:
@hodgestar, I closed many comments that are no longer relevant. I hope it makes the review easier.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,835 @@
...
+        if isinstance(Q_object, QobjEvo):
+            self.dims = Q_object.dims.copy()
+            self.shape = Q_object.shape
+            self._shift_dt = (<QobjEvo> Q_object)._shift_dt
+            self._issuper = (<QobjEvo> Q_object)._issuper
+            self._isoper = (<QobjEvo> Q_object)._isoper
+            if copy or args:
+                self.elements = (<QobjEvo> Q_object).elements.copy()
+            else:
+                self.elements = (<QobjEvo> Q_object).elements
```
hodgestar, line 214: We have to copy the elements here otherwise we'll have to two QobjEvo's modifying the same list of elements resulting in confusing bugs. Perhaps there is some benefit that I'm not seeing, but I think we should remove the `copy=True` argument entirely and always copy the list of elements.

hodgestar: Ah, I see we use `copy` later on to flag copying `Qobj`s. We can keep it for that, I think.

qutip/qip/device/processor.py
```diff
@@ -533,16 +532,22 @@ def get_qobjevo(self, args=None, noisy=False):
 
         qu_list = []
         c_ops = []
+        t_lists = []
```
hodgestar, line 535: As a QuTiP user, I would be pretty horrified to discover that `self.tlist` was set by calling `.get_qobjevo(...)` so let's find another way to do this? E.g. just add a helper method that returns the full tlist and call that alongside `.get_objevo` before calling the solver?

The original code was definitely odd here -- merging the tlists in all the coefficients didn't make a lot of sense.

qutip/qip/device/processor.py
```diff
@@ -700,7 +705,7 @@ def run_state(self, init_state=None, analytical=False, states=None,
 
         evo_result = solver(
             H=noisy_qobjevo, rho0=init_state,
-            tlist=noisy_qobjevo.tlist, **kwargs)
+            tlist=self.tlist, **kwargs)
```
hodgestar, line 708: See comment in `.get_qobjevo` about calculating the full tlist in its own method rather than storing it on self.

qutip/solve/mcsolve.py
```diff
@@ -276,7 +276,7 @@ def make_system(self, H, c_ops, tlist=None, args={}, options=None):
         ss.col_args = var
         for c in c_ops:
             cevo = QobjEvo(c, args, tlist=tlist)
-            cdc = cevo._cdc()
+            cdc = cevo.dag() * cevo
```
hodgestar, line 279: Given the discussions happened around `__mul__` vs `__matmul__` should we use `cevo.dag() @ cevo` here?

qutip/solve/mcsolve.py
```diff
@@ -648,7 +648,7 @@ def make_diag_system(self, H, c_ops):
         for c in c_ops:
             c_diag = Qobj(Ud @ c.full() @ U, dims=c.dims)
             cevo = QobjEvo(c_diag)
-            cdc = cevo._cdc()
+            cdc = cevo * cevo.dag()
```
hodgestar, line 651: Same comment about using `@` here.

qutip/tests/solve/test_stochastic_me.py
```diff
@@ -262,13 +262,18 @@ def test_general_stochastic():
...
     def d2(t, vec):
         out = []
+        vec_d = _data.Dense(vec)
         for op in sc_opsM:
-            out.append(op.mul_vec(t,vec)-op.expect(t,vec)*vec)
-        return np.stack(out)
+
+            out.append(
+                _data.add(op.matmul_data(t,vec_d),
+                          vec_d, -op.expect_data(t,vec_d))
```
hodgestar, line 274: It might be nice to write `scale=-op.expect_data(t,vec_d)` for those less familiar with the scale option to `.add`.

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,833 @@
...
+            else:
+                qobjs.append(element.qobj(0))
+                coeffs.append(element._coefficient)
+        for qobj, coeff in zip(qobjs, coeffs):
+            cleaned_elements.append(_EvoElement(qobj, coeff))
+        return cleaned_elements
+
+    def compress(self):
+        """
+        Look for redundance in the QobjEvo components, then merge them.
```
hodgestar, line 634: Could we extend this document with a list of what optimizations currently occur?

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,1114 @@
...
+            expect_func = _data.expect
+
+        for element in self.elements:
+            part = (<_BaseElement> element)
+            part.call(t, self.func_qobj)
+            out += part.coeff * expect_func(part.data, state)
+        return out
+
+    cdef double complex _expect_dense(QobjEvo self, double t, Dense state):
+        "For Dense state, `column_stack_dense` can be done inplace."
```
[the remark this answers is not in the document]
hodgestar: @Ericgig Could you update this comment to reflect what `_expect_dense` does? It's not doing column_stack_dense in all cases.

hodgestar (changes requested):
@Ericgig I left some questions and suggestions, but everything is reviewed now! :D

qutip/core/superoperator.py
```diff
@@ -44,177 +44,95 @@
...
+    data = -1j * _data.kron(spI, H.data)
+    data = _data.add(data, _data.kron(H.data.transpose(), spI), scale=1j)
+
+    for c_op, chi_ in zip(c_ops, chi):
+        c = c_op.data
+        cd = c.adjoint()
+        cdc = _data.matmul(cd, c)
+        data = _data.add(data, _data.kron(c.conj(), c), np.exp(1j*chi_))
+        data = _data.add(data, _data.kron(spI, cdc), -0.5)
+        data = _data.add(data, _data.kron(cdc.transpose(), spI), -0.5)
```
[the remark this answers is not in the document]
Ericgig: It's 10x faster for small system (2, 2) and still 2x for larger (100, 100). There are less temporary objects.

qutip/qip/device/processor.py
Ericgig: I made a `tlist` method. Adding `QobjEvo` with different `tlist` is not allowed in v4, thus the `merge_qobjevo`, but I don't understand why the solver `tlist` is the merged pulse `tlist`.

qutip/core/tensor.py
```diff
@@ -78,38 +82,32 @@ def tensor(*args):
...
             return right.linear_map(partial(tensor, left))
         if isinstance(right, Qobj):
-            return left.linear_map(lambda op: tensor(op, right))
-        id_like_left = qeye(left.dims[0])
-        id_like_right = qeye(right.dims[0])
-        left = left.linear_map(lambda op: tensor(op, id_like_right))
-        right = right.linear_map(partial(tensor, id_like_left))
-        return left * right
+            return left.linear_map(_reverse_partial_tensor(right))
+        left_t = left.linear_map(_reverse_partial_tensor(qeye(right.dims[1])))
```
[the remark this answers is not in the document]
Ericgig: It was wrong, it was left that needed to be changed. Added a test.

qutip/core/superoperator.py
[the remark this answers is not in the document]
hodgestar: Woot. I also wonder a bit if some of the temporary objects can be removed in the slow path, but let's leave this as is for now and just make sure we have tests for both parts.

qutip/qip/device/processor.py
hodgestar: The solver `tlist` being the merged pulse `tlist` is odd.

I'm guessing it's a heuristic so that the user doesn't need to supply one, but I'm not sure it's a good heuristic. Not all QobjEvos even have tlists.

A better heuristic would maybe be a tlist that includes the start and end of each pulse, any tlist values from each pulse (for step function pulses that use tlists) and enough times in between those to make the solver happy.

@BoxiLi Do you know why `.run_state` uses this merged tlist?

I'm happy for us to keep the old behaviour in this PR -- hopefully @BoxiLi can open an appropriate issue for us if the tlist needs to change in v5.

qutip/core/superoperator.py
```diff
@@ -235,15 +153,17 @@ def lindblad_dissipator(a, b=None, data_only=False, chi=None):
...
     Returns
     -------
     D : qobj, QobjEvo
         Lindblad dissipator superoperator.
     """
+    from .cy.qobjevo import QobjEvo
     if b is None:
         b = a
-        if isinstance(a, QobjEvoFunc):
-            return a._lindblad_dissipator(chi)
```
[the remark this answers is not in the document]
hodgestar: Thanks!

qutip/core/superoperator.py
```diff
@@ -44,177 +44,95 @@
...
 def _map_over_compound_operators(f):
     """
     Convert a function which takes Qobj into one that can also take compound
     operators like QobjEvo, and applies itself over all the components.
     """
     @functools.wraps(f)
     def out(qobj):
-        if isinstance(qobj, QobjEvoBase):
-            return qobj.linear_map(f)
+        from .cy.qobjevo import QobjEvo
```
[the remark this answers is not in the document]
hodgestar: Thanks!

qutip/qip/device/processor.py
```diff
@@ -542,13 +541,32 @@ def get_qobjevo(self, args=None, noisy=False):
...
 
         if noisy:
             return final_qu, c_ops
         else:
             return final_qu, []
 
+    def tlist(self, noisy):
+        """
+        Get the merged tlist of all pulses.
+
```
hodgestar, line 555:
```suggestion
        Return the merged tlist of all the pulses.
        
        Parameters
        ---------------
        noisy: bool, optional
            If noise should be included. Default is False.
```

qutip/qip/device/processor.py
hodgestar: Leaving this unresolved so that @BoxiLi can see it easily, but we can consider it resolved as far as this PR goes.

hodgestar:
@Ericgig Thanks for the changes! The only outstanding comments now are:

* Update comment on _expect_dense.
* Apply my docstring suggestion for the new tlist function (if you like the suggestion)
* Resolve merge conflicts (not sure how those crept in -- maybe the license change -- hopefully nothing serious)

And then I think we can merge! :D

qutip/qip/device/processor.py
```diff
@@ -542,13 +541,36 @@ def get_qobjevo(self, args=None, noisy=False):
...
+        if not noisy:
+            dynamics = self.pulses
+        else:
+            dynamics = self.get_noisy_pulses(
+                device_noise=True, drift=True)
+
+        t_lists = []
+        for pulse in dynamics:
+            tlist = pulse.get_full_tlist()
+            if tlist is not None:
```
hodgestar, line 570: Can `get_full_tlist` ever return anything other than a list? If `None` can be returned, perhapds `Drift.get_full_tlist` should return `None` instead?

qutip/core/cy/qobjevo.pyx
```diff
@@ -0,0 +1,839 @@
...
+            if state.shape[1] != 1:
+                state = column_stack_dense(state, inplace=state.fortran)
+            for element in self.elements:
+                part = (<_BaseElement> element)
+                coeff = part.coeff(t)
+                part_data = part.data(t)
+                out += coeff * expect_super_data_dense(part_data, state)
+            if state.fortran:
+                # `state` was reshaped inplace, restore it's original shape
+                column_unstack_dense(state, nrow, inplace=state.fortran)
```
hodgestar, line 788: Am I understanding that we mutate `state` (if `state.fortran`) and then mutate it back here? What happens if one of the functions calls raises an exception? Will state be left in a modified state? Should we fix that somehow?

Ericgig: We re not left with a modified state because of `cython`'s error handling. error are ignored and only caught back when in python:

```
>>> qevo = qt.QobjEvo(qt.spre(qt.qeye(4)))
>>> super_state = qt.rand_dm(3, dtype='dense').data.reorder()
>>> qevo.expect_data(0, super_state)

~/algo/qutip/qutip/core/data/expect.pyx in qutip.core.data.expect.expect_super_csr_dense()

~/algo/qutip/qutip/core/data/expect.pyx in qutip.core.data.expect._check_shape_super()

ValueError: incompatible shapes (16, 16), (9, 1)

Exception ignored in: 'qutip.core.data.expect.expect_super_data_dense'
  File "qutip/core/data/expect.pyx", line 209, in qutip.core.data.expect.expect_super_csr_dense
  File "qutip/core/data/expect.pyx", line 37, in qutip.core.data.expect._check_shape_super
ValueError: incompatible shapes (16, 16), (9, 1)

0j

>>> np.isfortran(super_state.as_ndarray()), super_state.shape
(True, (3, 3))
```

As you can see the shape is still `(3,3)` while being dense fortran data.
(And `qevo.expect_data` returned `0j` with the error.)

hodgestar: Woot. This Cython error handling stuff is crazy. :)

hodgestar (approved):
Approved! At last! Phew.

qutip/core/cy/qobjevo.pyx
jakelishman: That's not safe error handling! Ignoring the exception doesn't mean we should continue - when we're running in C mode, we have a responsibility to ensure that _no_ exceptions are ever raised. This is how segfaults happen.

qutip/core/cy/qobjevo.pyx
hodgestar: @Ericgig Let's sort this out before merging.

hodgestar (changes requested):
Let's sort out the exception handling issue before merging.

qutip/core/cy/qobjevo.pyx
jakelishman: Just to add: the fact that you could make an "Exception ignored in" message appear using only Python-space commands means that's something's wrong.  Given that the internal function calls already acquire the GIL to raise exceptions, there shouldn't be too much worry about possibly needing to do that here as well.  The cost of checking the exception state after the function should be trivial compared to the cost of calculating even a single expectation value, and even if it isn't, Python-accessible segfaults aren't a worthwhile tradeoff for that extra bit of speed.

I think you'll have to check the error condition after each one in the loop.  Things like `column_stack_dense` can mutate the shape of a `Dense` in-place, so `QobjEvo` can't be certain its shape is always accurate.  Those mutations at least won't cause segfaults, but they're still Python-accessible errors that cannot be avoided, so we have to check and raise.

hodgestar: Previously `expect_dense` had `except *` so perhaps we can fix this by adding that back and then by putting the clean up in a finally block?

Perhaps we can also add a test for a coefficient call that raises (which was the case I was originally worried about).

Ericgig:
@hodgestar 
I went with `try`. It works better than I though in cython, with about 8ns cost, or about 2% for a (16,16) operator, and quickly become insignificant.

qutip/core/cy/qobjevo.pyx
```diff
@@ -5,6 +5,7 @@ import numpy as np
 import numbers
 import itertools
 from functools import partial
+from cpython.exc cimport PyErr_CheckSignals
```
hodgestar, line 8: I don't see PyErr_CheckSignals used anywhere explicitly. Is this import still needed?

hodgestar:
> I went with `try`. It works better than I though in cython, with about 8ns cost, or about 2% for a (16,16) operator, and quickly become insignificant.

Thanks @Ericgig for checking the timing. 2% sounds great for 16x16, and as you say it will quickly become insignificant.

The fix looks good to me. I just left one question about a possible debugging import that got left behind.

qutip/core/cy/qobjevo.pyx
Ericgig: Sorry, it's not needed.

hodgestar (approved):
Approved!

@Ericgig Will leave the honour of doing the merge to you! :D

PR #1483, opened 2021-04-04
Fix PEP8

lebinyu:
**Description**
Fix PEP8 format issue

**Changelog**
Fixed existing PEP8 issue of '*.py' in qutip/qutip

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -129,6 +129,7 @@ class Bloch():
         Positions of +z and -z labels respectively.
 
     """
+
     def __init__(self, fig=None, axes=None, view=None, figsize=None,
                  background=False):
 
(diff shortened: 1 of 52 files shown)


BoxiLi:
Hi @lebinyu,

It is helpful in general to have a clean code following PEP8 indeed. However, I'm not sure if such a large scale of correction in one PR will cause some troubles. 
- There might be others working on some of these files and this PR will very likely cause merge conflicts.
- On `dev.major`, we are developing qutip 5.0 and from time to time we may merge the master branch into that. These kinds of massive style correction may lead to problems there. 

I would recommend splitting the commit and probably also the PR. Each time a few related files within one commit. Better focus on modules like `qip` and `control` etc, where the code is up-to-date (For qip, better in https://github.com/qutip/qutip-qip).  The core qutip part (solvers, Cython code) is undergoing intensive development in `dev.major`.

PR #1484, opened 2021-04-05
Change sys.platform to sysconfig.get_platform in setup.py

jakelishman:
The latter had the potential to cause problems when moving to cross-compilation for macosx-arm64 on conda-forge.  See https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/ for more details.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ import warnings
 
 # Required third-party imports, must be specified in pyproject.toml.
 from setuptools import setup, Extension
-from distutils import sysconfig
+import distutils.sysconfig
 import numpy as np
 from Cython.Build import cythonize
 from Cython.Distutils import build_ext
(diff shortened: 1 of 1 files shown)


hodgestar:
It looks like `sysconfig.get_platform` only exists in Python 3.8 and above. :/

jakelishman:
Oh no, it's because we shadow `sysconfig` by importing it from `distutils` instead of the full thing.  Ugh, now I have to work out if we specifically need the `distutils` version, or if we can use anything.  I have an annoying feeling that the `distutils` thing is to make a silly little monkeypatch we needed.

jakelishman:
We _do_ need `distutils.sysconfig` at one point - we're monkeypatching some compiler flags that `distutils` decides it's a good idea to set, so we modify the state of the module.  For some reason distutils decides that we really want `-Wstrict-prototypes` appended to our C++ compilations, where it's not a valid option.

Still, we can just separate out the namespaces to fix it.

hodgestar (approved):

PR #1487, opened 2021-04-06
Updating tests in meetrics, superops,and tensor modules

MrRobot2211:
**Checklist**

- [ x] Please read [Contributing to QuTiP Development](https://github.com/qutip/qutip-doc/blob/master/CONTRIBUTING.md)
- [] Please add tests to cover your changes if applicable.
- [ x] If the behavior of the code has changed or new feature has been added, please also update the [documentation](https://github.com/qutip/qutip-doc) and the [notebook](https://github.com/qutip/qutip-notebooks). Feel free to ask if you are not sure.

**Description**
Updating tests in the metrics,superop_reps and tensor modules to  pytests.

**Changelog**
- Updating unittests to pytests
-  Rechecking environment dependent expected failure cases


jakelishman:
I know you've marked this as draft, but let me point out that you've not really aligned your `master` with ours after the merge (one of the reasons it's not a good idea to develop using `master`).  You've merged in the changes from our branch, but really you needed to overwrite your local copy with ours; I squashed the previous PR down into just one commit to keep the commit log cleaner and easier to interpret.  In this sort of triangular workflow where you're proposing PRs that might not be merged as-is, you need to treat our repository as the single source of truth - you should never change your `master` branch except by doing `git pull` to our repository, so you can't get out of sync.

Since you've now developed on the top of an incorrect tip, it'll be a bit difficult for you to untangle if you're not very familiar with `git`.  You _can_ do it with rebasing or cherry-picking techniques to keep what you've done, but in this particular case (and there's no way you could have known this), I was actually just about to publish PR #1490 that makes all these same changes in `test_metrics.py`, but also goes a lot further in rewriting the tests in a more pytest-y style.

It might be better for you if you reset your repository to match ours exactly, and then start again using the proper branching structure for the superoperator and tensor tests you wanted to look at.  Perhaps you could have a look at #1490 as well to see how I go about parametrising tests and organising everything to keep it tidy (not saying that my code is perfect by a long shot, and the concept of "random" tests is questionable overall, but it's what we've got at the moment).  Remember as well that readibility is a big concern.  Using an automated tool to enforce pep8 compliance often won't get you the best results - our real style requirement is that it's easy to read, so linebreaks should be in sensible places (that still abide by pep8).

MrRobot2211:
Hi Jake okey will re align and pull then and focus the other modules. Let's see if I am getting this right, is it correct style to add parameter to fixtures when they are used all along the test module, but in very specific special cases just leave the variable assignment before the test definition, like here https://github.com/jakelishman/qutip/blob/862d0de844d9d5b39f0301d0ea4c47c37d995499/qutip/tests/test_metrics.py#L126 ?

jakelishman:
In general the biggest thing you should consider is readability, and to be fair, that's a little subjective.  You're welcome to comment on my PR if you think I've done anything that's unreadable.

You can scope fixtures at whatever level is appropriate.  In this particular case, I scoped the `dimension` fixture at the module level, because there's loads of places that needed to test varying dimensions, and it's very convenient to define it just once.  In the few cases where we needed to override that (e.g. in `dnorm`), I overrode it by providing a definition in a tighter scope - either at class level or per-function (via `pytest.mark.parametrize` directly).  The correct scoping just depends a lot on the type of tests you're writing - if you're going to have to overwrite something constantly then module scoping isn't good, but if you have lots of different functions to test, and every test for a given function will want the same parametrisation, then class scoping might be a good choice.

If you only need the parametrisation once, then it's best to put it right next to the place it's used (like in the case of mine that you've highlighted).  It means that if you've opened that file to read the tests, then everything you need to understand `Test_fidelity.test_known_cases` will be on your screen if you just scroll to it.  That makes it much easier for the next person who has to come along to edit your code.  I used a couple of temporary variables in that case just to aid readability - my cases wouldn't have nicely fit on one line if I'd shoved the `qutip.basis(2, 0)` stuff inside the `pytest.param` constructors, and it would be hard to read.  This way you can easily verify that the code is correct bit-by-bit; you can see clearly that the names are descriptive and match exactly what they say, and that means you can trust them when they then appear in the parametrisation.

I won't pretend to be perfect at any part of coding, and particularly testing.  I rewrote all the testing files starting from those beginning with "a" up to (now) "metrics" while I was first getting to know the QuTiP codebase, and hopefully you can see that my early efforts weren't that great, and they get a bit better and easier to read as time went on.  (The later files haven't been converted yet.)  The main things are to always be thinking about readability both of the code and the error messages that are coming out, and making sure that your tests are really testing that your functions have your behaviour you want them to have.  Another thing we really don't have enough of in QuTiP yet is tests for "negative" behaviour, i.e. testing that functions _fail_ when they're passed incorrect inputs - consequently, quite a lot of QuTiP actually will just silently do the wrong thing if it's passed nonsense.

jakelishman:
Also, sorry for stepping on your toes on the `test_metrics` thing - that really wasn't intentional, and I was only looking at it more because there are a few historic problems with dnorm testing that I'd remembered later.  I didn't mean to cut you off from what you were doing, and had you had more work done on it, I'd have worked with you to merge the two together rather than suggesting you throw yours out.

MrRobot2211:
Hi Jake no problem. I was just asking about that particular line because I faced the same issue of readability versus properly scoping and could not get to a satisfactory solution. I was geting the variables from a global function that created a dictionary and it was not clean at all.

Issue #1330, opened 2020-07-27
QubitCircuit.resolve_gates() errors on "FREDKIN" gate.


sarsid:
**Describe the bug**
When applying `resolve_gates()` method to `QubitCircuit` containing the FREDKIN gate, the following error is produced: 

ValueError: Gate RY requires one target

**To Reproduce**

```python
from qutip.qip.circuit import QubitCircuit
A = QubitCircuit(3)
A.add_gate("FREDKIN", targets=[0, 1], controls=[2])
A.resolve_gates()
```

It might be a good idea to check this for other gates as well as look into the internal of `resolve_gates` to clean/modify it a bit. 

Note: As requested by @BoxiLi, first mentioned by me in #1318

purva-thakre:
@BoxiLi I am a bit confused with the wording of this issue. Is the goal to find what other gates might be creating `ValueError` when `resolve_gates` acts on them ? Then once these gates are identified, figure out the source of error in code of `resolve_gates` ?

sarsid:
Yep, the `resolve_gates()` function is quite messy right now and works on a case by case basis. In the FREDKIN case, it was giving this error. The issue would atleast involve resolving the FREDKIN gate properly as a starting point.

purva-thakre:
Hi @sarsid I am having trouble reproducing this issue. When I copied above code, I get `<qutip.qip.circuit.QubitCircuit object at 0x7fe142c414f0>` and not the `value` error referenced by you. Is there something else I should be doing ?

BoxiLi:
Just tried this, I still the same error. So the `resolve_gates()` function works for you? What are the gates you get in the end? Are they correct?

BoxiLi:
Better also check the qutip version you are using

purva-thakre:
@BoxiLi I have the latest version installed. 

I do not get any gates after applying `resolve_gates()`. My issue is I fail to get the same error type i.e. `ValueError`. When I ran the code as it is, I do not get any gates besides an indicator that the code ran successfully. So, I used `print(A.resolve_gates())` which gave me `qutip.qip.circuit.QubitCircuit object at 0x7fb64044eaf0`. This is most definitely not the expected output of `resolve_gates()` as specified [here](https://github.com/qutip/qutip/blob/8681e995c09986a4355ba16d586dd2d2f1f49657/qutip/qip/circuit.py#L1205)

BoxiLi:
[...]

PR #1489, opened 2021-04-07
Fixed bugs in _gate_FREDKIN function

BOBO1997:
**Description**
Modified the function `_gate_FREDKIN` in `qutip/qutip/qip/circuit.py`.

As reported in [issue #1330](https://github.com/qutip/qutip/issues/1330), there was a bug when unrolling Fredkin gate into basis gates. This bug derives from the use of a list with length >1 when making a new single-qubit gate in the function `_gate_FREDKIN`.
In this PR, I fixed this problem by manually unrolling this part.
For the details and the correctness verification, please refer to the attached pdf: [issue_1330.pdf](https://github.com/qutip/qutip/files/6272589/issue_1330.pdf)

Thank you very much!

**Related issues or PRs**
Fixed the bug reported at [issue #1330](https://github.com/qutip/qutip/issues/1330).

**Changelog**
Fixed errors in _gate_FREDKIN function.


BoxiLi:
Thanks, @BOBO1997. Could you also add a quick test for this?

BOBO1997:
Thanks a lot for the quick review!
I added test function `testFREDKINdecompose` into `qutip/tests/test_qubitcircuit.py`.
After I added this `testFREDKINdecompose` function, I found the test fails with the existing decomposition.
Therefore I totally changed the way of decomposition of FREDKIN gate, which eventually passed the quick test.

BOBO1997:
Apart from this issue, I would like to confirm the behaviors of some features in `qutip.qip.QubitCircuit`.

First, for `qutip.qip.QubitCircuit.add_gate()`, there is a optional `list` type attribute called index.
I saw this is used to insert gates in the middle of the gate list (`qutip.qip.QubitCircuit.gates`).
The corresponding code is shown below.
```
            for position in index:
                num_mes = (sum(isinstance(op, Measurement) for op
                               in self.gates[:position]))
                gate = Gate(name, targets=targets, controls=controls,
                            arg_value=arg_value, arg_label=arg_label,
                            classical_controls=classical_controls,
                            control_value=control_value)
                self.gates.insert(position, gate)
```
In this code, if `index` contains two or more elements, then the insert positions would be shifted one by one.
Is this behavior expected?
If not, I would like to modify this part in this PR, since modification is easy.

Second, as I mentioned in issue #1330, we can use a syntax sugar like:
```
qc = QubitCircuit(6)
qc.add_gates("RX", targets=[0,2,4,5], arg_value=np.pi/2)
```
when we want to apply the same single-qubit gates on the different qubits.
Personally, I think this syntax is good in terms of code readability.
If this looks good for the core development members, I would also add this syntax in this PR.

Best!

qutip/tests/test_qubitcircuit.py
```diff
@@ -148,6 +148,24 @@ def testSNOTdecompose(self):
...
+        qc1 = QubitCircuit(3)
+        print("OK")
+        qc1.add_gate("FREDKIN", targets=[0, 1], controls=[2])
+        print("OK")
+        U1 = gates.gate_sequence_product(qc1.propagators())
+        print("OK")
+        qc2 = qc1.resolve_gates()
+        print("OK")
+        U2 = gates.gate_sequence_product(qc2.propagators())
+        print("OK")
```
BoxiLi, line 165: There is no need to print ok here. If something goes wrong, pytest will locate which line the error comes from and summarize it in the result.

BoxiLi (changes requested):
Looks good to me, apart from a minor change request in the test.

BOBO1997:
I just fixed this point. Thanks for pointing it out!

BoxiLi (approved):

BoxiLi:
@BOBO1997  Thanks!

BoxiLi:
@BOBO1997 Sorry I forgot about your proposal above.

- What is the proposed behaviour of gate insertion? I guess you mean that, in the current code, if one inserts more than one gates, the length of the list is modified during the loop. This makes the definition of `position` tricky. Is this what you mean?

- About the syntax sugar, this sounds nice indeed. However, there might be a rather big refactor in `QubiCircuit` in the near future. I had a proposed refactor of `QubiCircuit` https://github.com/qutip/qutip-qip/discussions/15. Although no one is working on that yet. I feel like this change will be much easier to implement after that refactor?

In any case, they should be in a different PR as they don't relate to this topic of FREDKIN.

jakelishman:
One thing to consider when you're adding syntactic sugar is making sure that you don't accidentally reuse the same syntax to mean two completely different things.  Here, `targets=[0, 1]` means something very different if the gate is two-qubit compared to if the gate is single-qubit.  Pushing it further, what would `targets=[0, 1, 2, 3]` mean for a two-qubit gate?  What about a two-qubit gate with one target and one control?  What if a target is duplicated?  These aren't unsolvable problems by any means, but they're something to consider: does the helper syntax still help once the use-cases get complicated?

The for-loop syntax alternative is easy to read, easy to write and presumably (though I don't work on `qutip.qip`) doesn't prevent any meaningful speedups, since this doesn't involve reparsing the circuit or anything like that.

BoxiLi:
> One thing to consider when you're adding syntactic sugar is making sure that you don't accidentally reuse the same syntax to mean two completely different things.

Thanks @jakelishman A more elegant way is to define a meta gate like `All(X)` in ProjectQ, which acts like `Tensor`. But it is true that this syntax sugar is not as simple as it looks like.

BOBO1997:
@BoxiLi Exactly. In the current code, the positions in `index` would not indicate the original positions during the insertion loop. (e.g. When we want to insert a gate `A` to `gates=[B0,B1,B2,B3,B4,B5]` in `index=[0,2,3]`, then the output gate list would be `gates=[A,B0,A,A,B1,B2,B3,B4,B5]`, not `gates=[A,B0,B1,A,B2,A,B3,B4,B5]`.)
I just want to check whether this behavior is correct or not.

Actually, since this issue is not related to the PR for fixing FREDKIN, I should have created a new issue/PR. I'm sorry for this.
I created a new issue page #1497.

Issue #880, opened 2018-06-12
BUG: dnorm tests sporadically failing


ajgpitch:
The dnorm tests are sporadically (quite regularly) failing e.g.

```
======================================================================
ERROR: Metrics: checks that special cases for dnorm agree with SDP solutions.
----------------------------------------------------------------------
cvxpy.error.SolverError: Solver 'CVXOPT' failed. Try another solver or solve with verbose=True for more information. Try recentering the problem data around 0 and rescaling to reduce the dynamic range.
```

I suggest we turn them off by not installing cvxpy for any of the test configurations.
We can turn them on again if someone can make an attempt to make the dnorm tests more reliable @cgranade 

I will raise a PR for this now.

ajgpitch:
This has been the final straw. We have tried to keep dnorm supported. See #872 and the suggestion in #874

Issue #874, opened 2018-06-04
ENH: Remove dnorm cvxpy dependence


ajgpitch:
The `metrics.dnorm` function uses the cvxpy library to calculate the diamond norm. Twice now it has caused many test failures out of the blue #484 and #872.
It would seem like the developers do not prioritise maintaining a conda package. As this is our recommended platform, then it is a risk having this cvxpy dependence.

Perhaps there is a method using scipy that we could employ?

@cgranade perhaps could provide some context?

nathanshammah:
This seems closed by #880?

ajgpitch:
I think that we don't currently have a method for calculating the diamond norm. I am assuming that we would like to maintain one. We could either try again with cvxpy, or try find another method.

nathanshammah:
A recent Python package using QuTiP as a dependency, `qptomographer`, addresses the diamond norm distance. See: https://qptomographer.readthedocs.io/en/latest/figures_of_merit/#the-diamond-norm-distance. Also, [ArXiv:1808.00358](https://arxiv.org/abs/1808.00358).

phfaist:
Our diamond norm calculation in [QPtomographer](https://github.com/Tomographer/QPtomographer/) is written in C++ in the file [QPtomographer/cxx/diamond_norm_scs.h](https://github.com/Tomographer/QPtomographer/blob/master/cxx/diamond_norm_scs.h) (see [QPtomographer/test/test_diamond_norm.cxx](https://github.com/Tomographer/QPtomographer/blob/master/test/test_diamond_norm.cxx) for an example usage) and uses the C interface of [the SCS solver](https://github.com/cvxgrp/scs). The way it stands, it also depends on [Eigen C++ header library](http://eigen.tuxfamily.org/), but it should be easy to remove this dependency as it is only used to specify the input matrices. Other minor dependencies are easily removed.

I'm not sure what underlying mechanism you use in QuTiP, but it should be relatively straightforward to reformat the whole class into a C implementation with only SCS as a dependency. Unfortunately the python `scs` package does not install C headers, so that dependency might turn out to be something to worry about. Perhaps SCS itself can be integrated in the QuTiP source?

I'd be happy if our code is useful for you, and let me know if you'd like any clarifications about our implementation.

nathanshammah:
[...]

PR #1490, opened 2021-04-07
Allow rerunning of dnorm tests, refactor test_metrics.py

jakelishman:
dnorm tests have historically been very flaky, and for as long as we're using random tests (which we possibly shouldn't be), we always run the risk of temperamental failures.  In particular, the semidefinite solvers used in dnorm fail every now and again on random states; we historically haven't treated this as a bug, since it's pretty similar to an integration failing to converge because the system was too stiff.  It means that the user has to adjust some settings and try again.

To prevent this sort of error from failing our test suite, however, we can mark the tests as being allowed to rerun twice on a failure.  We had a relatively small sample to judge the test failure rate by, but my very very approximate guess is that we'd have a failure about 1 in 10 runs on Travis.  We counter this with two strategies:

1. parametrise the tests in pytest style, so each random repetition is a separate instance, and known to pytest (some of this was done in earlier commits)
2. allow 2 reruns of every dnorm test

The two of these together should remove all random failures; since all random loops have been moved into pytest-handled repeats, there are far fewer calls to dnorm handled within each test. This means less chance that an individual test run will fail, but a slightly higher chance the entire suite will.  Now with the test re-running a single failure point is tried again up to twice more, to determine if it was just a one-off, or if the test is truly broken.

The reruns are handled by a pytest plugin pytest-rerunfailures.  It's quite nontrivial behaviour to hook this in to the pytest mechanism, so better to use an external dependency to do it than to vendor in a custom version.

The dnorm problems have resurfaced a little after lying dormant for a long time, because #1463 reactivated them.  Before that, they'd not actually been running because none of the test runners had cvxpy installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,8 @@ _mac_generic_setup: &mac_generic_setup
   os: osx
   language: generic
   install:
-    - conda install nomkl blas=*=openblas numpy scipy matplotlib cython pytest
+    - conda install nomkl blas=*=openblas numpy scipy matplotlib cython
+    - pip install pytest pytest-rerunfailures
     - python setup.py install
 
 jobs:
@@ -54,7 +55,8 @@ jobs:
     - name: "Linux, Python 3.8: regular"
       python: 3.8
       install:
-        - conda install numpy scipy matplotlib pytest cython
+        - conda install numpy scipy matplotlib cython
+        - pip install pytest pytest-rerunfailures
         - python setup.py install
 
     - name: "Linux, Python 3.7: OpenMP, MKL, scipy<1.5 (coverage)"
@@ -68,7 +70,8 @@ jobs:
       # This is why we use `setup.py develop` here, and why we don't change
       # directories.
       install:
-        - conda install mkl blas=*=mkl numpy 'scipy<1.5' matplotlib cython pytest pytest-cov coveralls cvxopt
+        - conda install mkl blas=*=mkl numpy 'scipy<1.5' matplotlib cython cvxopt
+        - pip install pytest pytest-rerunfailures pytest-cov coveralls
         - conda install -c conda-forge cvxpy
         - python setup.py develop --with-openmp
       before_script:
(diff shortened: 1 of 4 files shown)


hodgestar:
@jakelishman I'm somewhat against re-running flaky tests all the time. It's inelegant and hides other possible flakiness. An alternative approach is to set the random seed before generating the random states using a small fixture, e.g.:

```python
import numpy
import random as rand

@pytest.fixture
def random():
    rand.seed(0)
    numpy.random.seed(0)
```
(from https://github.com/pytest-dev/pytest/issues/667#issuecomment-112206152)

potentially we could also run the tests with a few different random seeds (to cover more cases).

If we add hypothesis (https://hypothesis.readthedocs.io/en/latest/) one day, we could generate a broader set of cases to solve and filter out bad cases.

jakelishman:
I completely agree with you that random tests really aren't the right way to go in general, but right now our problem is that basically QuTiP's entire testing suite is based on randomised tests, and that's an awful lot of technical debt to overcome any time soon :(. 

The `dnorm` tests have historically been a huge problem, even though basically nobody's worked on `dnorm` for a long long time now (for context, see in particular #880 and #874).  We don't actually even know if the failures in `dnorm` are deterministic and reproducible, although it turns out that I had a brief look into this when I joined as a GSoC student about a year ago...  I don't remember doing it at all!

I know this is super inelegant and not the perfect method, but it's kind of a nasty trade-off - I don't have time to really dig into `dnorm` (a topic I know almost nothing about) to work out an appropriate solution to temperamental errors, we do want `dnorm` to be tested since we're shipping it, and the slight flakiness keeps causing us test suite failures about 1 in 10 times.  I'm somewhat opposed to fixing the random seed because having the variance over several test runs does actually help us cover more ground (eventually) - fixing it to a "known good" seed actually seems to me to be more of a mask of potential problems than this.

Of note: the pytest plugin I'm using reports the number of reruns it performed in the analysis at the bottom - they show up in a yellow alongside the "skips" and "xfails" - so it's not completely hidden.

I'd love to have a property-based testing suite like hypothesis up and running - that definitely seems like the proper solution to this in the end.  In the meantime, I'm open to swapping to a fixed random seed if you think that's much better - testing is one place where I'm really aware of my lack of formal experience in software engineering, so I'm basically just learning as I go.

hodgestar (approved):

hodgestar:
@jakelishman I've approved this PR -- it's definitely worth getting rid of test flakiness right away.

Re random seed: We wouldn't want to set the random seed for all tests, just those like the dnorm ones that are causing issues. But let's leave that for the future.

jakelishman:
Ok, I'll merge it for now, but I'll try and return to look at `dnorm` in more detail in the future, and particularly its tests; I'm also not a huge fan of having a secondary test dependence, but for the most part that's not really an issue.

Issue #1495, opened 2021-04-11
Scipy<1.5 hermitian eigenvalue claculation causing wrong results, should update requirements and testsuite


MrRobot2211:
**Describe the bug**
Currently qutip is requiring scipy > 1.12  and in the CI checks scipy <1.5 is being checked. There was a bug in scipy 1.4 See [.Stackoverflow error](https://stackoverflow.com/questions/54314529/mkl-error-parameter-12-for-large-matrices-with-scipy-linalg-eigvalsh-in-an) , [bug report ](scipy/scipy#8205)  that has been fixed  by this [solution](scipy/scipy#11304) in the 1.5 release.  This is related to the `linalg.eigh() API` which performs eigenvalue calculation, it causes error of this form  `Intel MKL ERROR: Parameter 12 was incorrect on entry to ZHBRDB`,  which mostly the get silenced, and  instead of terminating it just outputs an array of zeros except for the last entry.   This issue occurs exclusively when using the API with high dimensional tensors, which are probably not being handled in the tests, but will inevitably cause miscalculations for the users as eigenvalue finding of hermitics  is pervasive.

**To Reproduce**
Just as an illustration you cun run  

`pytest qutip/tests/test_superop_reps.py -s -k SuperChoiChi`

from the codes in #1491  in an environment like the generated for `coverage` tests in the CI

The terminal out put is
```
qutip/tests/test_superop_reps.py ..
Intel MKL ERROR: Parameter 12 was incorrect on entry to ZHBRDB.
F

================================================================= FAILURES ==================================================================
_________________________________________________ TestSuperopReps.test_SuperChoiChiSuper[8] _________________________________________________

self = <qutip.tests.test_superop_reps.TestSuperopReps object at 0x7f01d2d5c6d0>, dimension = 8

    @pytest.mark.parametrize('dimension', [2, 4, 8])
    def test_SuperChoiChiSuper(self, dimension):
        """
        Superoperator: Converting two-qubit superoperator through
        Choi and chi representations goes back to right superoperator.
        """
        superoperator = super_tensor(rand_super(dimension), rand_super(dimension))
    
        choi_matrix = to_choi(superoperator)
        chi_matrix = to_chi(choi_matrix)
        test_supe = to_super(chi_matrix)
    
        # Assert both that the result is close to expected, and has the right
        # type.
        #print(test_supe.dims)
        #print(superoperator.dims)
>       assert (test_supe - superoperator).norm(maxiter=1e6) < 1e-5
[...]

PR #1498, opened 2021-04-13
Scipy<1.5 and mkl eigenh problem solved

MrRobot2211:
**Description**
Setting eigh_unsafe on sipy<1.5 and  MKL

**Related issues or PRs**
#1495

**Changelog**
- Changed init to set eigh not safe in case scipy<1.5 and mkl backend
- Added the correct pytest breaking test from issue #1491 to check the solution

### DIFF
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -63,10 +63,16 @@ else:
     os.environ['KMP_DUPLICATE_LIB_OK'] = 'True'
 
 import platform
+import scipy
+import mkl
+from packaging import version
 from qutip.utilities import _blas_info
-qutip.settings.eigh_unsafe = (_blas_info() == "OPENBLAS" and
-                              platform.system() == 'Darwin')
-del platform, _blas_info
+qutip.settings.eigh_unsafe = ((_blas_info() == "OPENBLAS" and
+                              platform.system() == 'Darwin') or 
+                              (version.parse(scipy.__version__) < version.parse("1.5") 
+                              and (_blas_info() == 'INTEL MKL') )
+                              )
+del platform, _blas_info, scipy
 # -----------------------------------------------------------------------------
 # setup the cython environment
 #
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -67,12 +67,13 @@ import scipy
 import mkl
 from packaging import version
 from qutip.utilities import _blas_info
+is_old_scipy = version.parse(scipy.__version__) < version.parse("1.5")
 qutip.settings.eigh_unsafe = ((_blas_info() == "OPENBLAS" and
-                              platform.system() == 'Darwin') or 
-                              (version.parse(scipy.__version__) < version.parse("1.5") 
-                              and (_blas_info() == 'INTEL MKL') )
+                              platform.system() == 'Darwin') or
+                              (is_old_scipy
+                              and (_blas_info() == 'INTEL MKL'))
                               )
-del platform, _blas_info, scipy
+del platform, _blas_info, scipy, version, is_old_scipy
 # -----------------------------------------------------------------------------
 # setup the cython environment
 #
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -64,7 +64,6 @@ else:
 
 import platform
 import scipy
-import mkl
 from packaging import version
 from qutip.utilities import _blas_info
 is_old_scipy = version.parse(scipy.__version__) < version.parse("1.5")
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -64,15 +64,24 @@ else:
 
 import platform
 import scipy
-from packaging import version
+from packaging import version as pac_version
 from qutip.utilities import _blas_info
-is_old_scipy = version.parse(scipy.__version__) < version.parse("1.5")
-qutip.settings.eigh_unsafe = ((_blas_info() == "OPENBLAS" and
-                              platform.system() == 'Darwin') or
+
+is_old_scipy = pac_version.parse(scipy.__version__) < pac_version.parse("1.5")
+qutip.settings.eigh_unsafe = (
+                              # macOS OpenBLAS eigh is unstable, see #1288
+                              (_blas_info() == "OPENBLAS" and
+                               platform.system() == 'Darwin')
+                              or
+                              # The combination of
+                              # scipy<1.5 and MKL causes wrong results
+                              # when calling eigh for big matrices.
+                              # See #1495, #1491 and #1498.
                               (is_old_scipy
-                              and (_blas_info() == 'INTEL MKL'))
-                              )
-del platform, _blas_info, scipy, version, is_old_scipy
+                               and (_blas_info() == 'INTEL MKL'))
+                               )
+
+del platform, _blas_info, scipy, pac_version, is_old_scipy
 # -----------------------------------------------------------------------------
 # setup the cython environment
 #
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -69,16 +69,16 @@ from qutip.utilities import _blas_info
 
 is_old_scipy = pac_version.parse(scipy.__version__) < pac_version.parse("1.5")
 qutip.settings.eigh_unsafe = (
-                              # macOS OpenBLAS eigh is unstable, see #1288
-                              (_blas_info() == "OPENBLAS" and
-                               platform.system() == 'Darwin')
-                              or
-                              # The combination of
-                              # scipy<1.5 and MKL causes wrong results
-                              # when calling eigh for big matrices.
-                              # See #1495, #1491 and #1498.
-                              (is_old_scipy
-                               and (_blas_info() == 'INTEL MKL'))
+        # macOS OpenBLAS eigh is unstable, see #1288
+        (_blas_info() == "OPENBLAS" and
+         platform.system() == 'Darwin')
+        or
+        # The combination of
+        # scipy<1.5 and MKL causes wrong results
+        # when calling eigh for big matrices.
+        # See #1495, #1491 and #1498.
+        (is_old_scipy and
+         (_blas_info() == 'INTEL MKL'))
                                )
 
 del platform, _blas_info, scipy, pac_version, is_old_scipy
--- a/qutip/__init__.py
+++ b/qutip/__init__.py
@@ -69,17 +69,12 @@ from qutip.utilities import _blas_info
 
 is_old_scipy = pac_version.parse(scipy.__version__) < pac_version.parse("1.5")
 qutip.settings.eigh_unsafe = (
-        # macOS OpenBLAS eigh is unstable, see #1288
-        (_blas_info() == "OPENBLAS" and
-         platform.system() == 'Darwin')
-        or
-        # The combination of
-        # scipy<1.5 and MKL causes wrong results
-        # when calling eigh for big matrices.
-        # See #1495, #1491 and #1498.
-        (is_old_scipy and
-         (_blas_info() == 'INTEL MKL'))
-                               )
+    # macOS OpenBLAS eigh is unstable, see #1288
+    (_blas_info() == "OPENBLAS" and platform.system() == 'Darwin')
+    # The combination of scipy<1.5 and MKL causes wrong results when calling
+    # eigh for big matrices.  See #1495, #1491 and #1498.
+    or (is_old_scipy and (_blas_info() == 'INTEL MKL'))
+)
 
 del platform, _blas_info, scipy, pac_version, is_old_scipy
 # -----------------------------------------------------------------------------
--- a/qutip/tests/test_superop_reps.py
+++ b/qutip/tests/test_superop_reps.py
@@ -79,6 +79,8 @@ class TestSuperopReps:
 
         # Assert both that the result is close to expected, and has the right
         # type.
+        
+
         assert (test_supe - superoperator).norm() < 1e-5
         assert choi_matrix.type == "super" and choi_matrix.superrep == "choi"
         assert chi_matrix.type == "super" and chi_matrix.superrep == "chi"
@@ -367,4 +369,5 @@ def test_chi_known():
         [0, 0, 0, 0],
         [0, 0, 0, 0]
     ])
+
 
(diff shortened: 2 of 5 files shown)


qutip/tests/test_superop_reps.py
```diff
@@ -58,295 +59,311 @@
 tol = 1e-9
 
 
-class TestSuperopReps(object):
+
+class TestSuperopReps:
```
jakelishman, line 63: Is there a reason that you left one function inside this class and moved all the others outside it?  In this case the class isn't needed, because it would just include everything in the file and the filename is enough of a distinction between different testing units, but if we're going to keep it, then all the logically grouped functions should be inside it.

jakelishman (commented):
Thank you for finding this.   In the future, it might have been a bit more organised to fix the the `eigh` issue and add the extra dimension test in one PR, and make whatever changes you wanted to do to the other `superop_reps` tests in a different one to keep the history easier to read, but I'm not so bothered about this.  It's an ideal - it's important to think about how someone a year in the future is going to understand the repository history - but not one that we enforce.

jakelishman:
Also, how does this PR fit relative to #1491?  They appear to make many conflicting changes with each other.

MrRobot2211:
> Also, how does this PR fit relative to #1491?  They appear to make many conflicting changes with each other.

Hi Jake with respect to this. The apparent conflicting changes are only in the test file, if I am not moatakeb. Ideally I would like this one to get approved, and then I will rebase #1491 from master.

qutip/tests/test_superop_reps.py
MrRobot2211: The main issue with this one is that if functions go inside the class  `pytest` tries to run them as tests and fails. This is necesary since I am only making  the failing test pytest compatible. But there might be a workaround I do not know of.

qutip/__init__.py
```diff
@@ -63,10 +63,16 @@
     os.environ['KMP_DUPLICATE_LIB_OK'] = 'True'
 
 import platform
+import scipy
+from packaging import version
```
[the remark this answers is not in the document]
MrRobot2211: I am adding an alias then.Here I was uncomfortable adding a scipy dependency just to get the version. But I did not find a way out of it.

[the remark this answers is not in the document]
jakelishman: We already depend on scipy (otherwise why would you need the version check?).  The problems are that you've used `packaging` without specifying it as a dependency, and that the import could shadow the name `version`, which is a name we want exported to refer to the `qutip.version` module.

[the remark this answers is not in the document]
MrRobot2211: Sorry I meant uncomfortable importing the whole package as it was not being importes anywhere in the file.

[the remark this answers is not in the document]
jakelishman: `scipy` is well-designed, and just importing `scipy` is a very very light import.  You'd need to import every individual subpackage of `scipy` to cause performance issues (though actually not really, because `qutip` already imports a lot of the subpackges at initialisation, which we probably shouldn't).

qutip/tests/test_superop_reps.py
jakelishman: What do you mean?  `pytest` is the only tool that runs tests in QuTiP.  Also, all the tests run whether or not they're inside the class.

What do you mean by test failures?  I cannot accept a PR that introduces test failures, not matter what your future plans are, and by that I don't just mean hiding tests that don't pass, I mean you mustn't cause performance or logic regressions, and running all the tests ensures that.  A PR is a complete unit of work, it has to be self-contained and make a usable change.  Remember that this is a tool that thousands of people depend on, and any change you make that breaks things can affect users and certainly affects other developers trying to work on the library at the same time - there's a difference in working with a team and working just for yourself.

MrRobot2211: Remember for example this PR #1463 , there we made only some of the tests pytest compatible thus runnable. If we had put  some of the extra test inside a class they would have ran and failed just because they were not written in a pytest amicable format. 
Before this current PR none of the superop_metric tests was being ran, as pytest was not being imported, only unittest.

qutip/__init__.py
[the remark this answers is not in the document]
MrRobot2211: Iresolved the shadowing and I added packaging to `requrements.txt`, though as you mention it is the de facto stdlib. Should it be added somewhere else?

qutip/tests/test_superop_reps.py
jakelishman: You might want to check you're running the tests correctly.  You can see in the CI logs on GitHub that they are being run, and if you pass the `-v` option when running a single file locally (e.g. `pytest -v test_superop_reps.py`) you'll see clearly that they're all being run locally as well, without any issues.  The only reason that there were issues in #1463 was because it used the old nose-style `yield` format, which `pytest` doesn't support (it uses its more powerful parametrisation instead).

The point about not introducing breaking changes still stands, though - if you thought you were introducing breaking changes, then the correct thing to do would have been to write the test in the same style as the rest of the tests so that they all continued to run, then have a different PR to swap them all to `pytest`.  That way we never compromise our testing, even temporarily.

MrRobot2211: Sorry you are totaly right. The tests are still running if put inside the class. Sorry about that.

qutip/__init__.py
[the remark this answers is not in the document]
jakelishman: `requirements.txt` doesn't actually have all that much to do with our specified requirements - it's usually just a convenience file (and one that we've historically forgotten to update).  The actual package requirements are in `setup.cfg`.

`packaging` is not required to have a functioning version of Python; it's not part of the Python standard library.  However, it's _such_ a common dependency (transitively - it's used by `setuptools`), that you already had it installed.  A brand-new, clean Python installation won't, though, so you have to specify it.

qutip/tests/test_superop_reps.py
jakelishman: That's quite ok, no need to apologise!

qutip/__init__.py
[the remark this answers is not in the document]
MrRobot2211: I chose 18.1 as the minimal packaging release version based on the heuristic that it has to have at least an year of life. In any case I do not hold any strong opinions on this.

We should check the minimal version on the CI now that it has become a dependency right?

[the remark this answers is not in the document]
jakelishman: The aim of version requirements is only to mark the minimum version that includes all the features we need, not to arbitrarily limit people.  As far as I'm aware, `packaging` has always had `packaging.version.parse`, so there's not need to specify a minimum version.  (Also my heuristic was we can't depend on anything younger than two years old at a minimum, but packaging 18.1 does satisfy that.)

Technically yeah, it needs to be included in the CI as well, though really it doesn't matter too much - it'll automatically be installed because our build process depends on `setuptools`.

jakelishman:
Including a dimension of 8 makes the tests very slow to run.  There's no real need to expand the dimensions of the test to such massive sizes - testing the stability of `eigh` should be done using a minimal test in a more specific location, such as in our general eigensystem tests.

In this case, you're constructing several 4096x4096 matrices using a sparse format, but they're all 100% dense so you spend ~30 seconds just handling sparse overhead that's nothing to do with the function call.  You need to think about what a test is actually trying to test, and how long it _should_ take.  When your changes are increasing the runtime by a factor of ~100, that's something you need to consider - are you actually increasing our test coverage, or just spinning the CI for longer?  What is your change meant to catch?  Could it be done faster in a different place?  Take your time when you're changing the code, and really think about what you're trying to achieve.

qutip/__init__.py
```diff
@@ -63,10 +63,25 @@
...
+                              (_blas_info() == "OPENBLAS" and
+                               platform.system() == 'Darwin')
+                              or
+                              # The combination of
+                              # scipy<1.5 and MKL causes wrong results
+                              # when calling eigh for big matrices.
+                              # See #1495, #1491 and #1498.
+                              (is_old_scipy
+                               and (_blas_info() == 'INTEL MKL'))
+                               )
```
jakelishman, line 82: Please also fix the formatting of this.  The purpose of the style guidelines is to make code readable - just because something technically passes an automated test doesn't mean it's in a good format.  You can do both at the same time.

MrRobot2211: How do you mean? The `or `  line ? I usually se `SQL` queries written like this. BUt if you want I can change it,

MrRobot2211:
> Including a dimension of 8 makes the tests very slow to run. There's no real need to expand the dimensions of the test to such massive sizes - testing the stability of `eigh` should be done using a minimal test in a more specific location, such as in our general eigensystem tests.
> 
> In this case, you're constructing several 4096x4096 matrices using a sparse format, but they're all 100% dense so you spend ~30 seconds just handling sparse overhead that's nothing to do with the function call. You need to think about what a test is actually trying to test, and how long it _should_ take. When your changes are increasing the runtime by a factor of ~100, that's something you need to consider - are you actually increasing our test coverage, or just spinning the CI for longer? What is your change meant to catch? Could it be done faster in a different place? Take your time when you're changing the code, and really think about what you're trying to achieve.

Good point I did not realize it was taking this long on the last mac test. To be clear by eigensystem you are referring to this one https://github.com/qutip/qutip/blob/master/qutip/tests/test_eigenstates.py ?

MrRobot2211:
The current test implementation is taking 3 seconds locally. We'll see...

qutip/__init__.py
jakelishman: Sorry, it's just a minor nitpick, but the spacing - right now it's all crammed into ~35 characters at the far right, when you can space it more legibly as
```python
qutip.settings.eigh_unsafe = (
    # macOS ...
    (_blas_info() == "OPENBLAS" and platform.system() == "Darwin")
    # scipy<1.5 and MKL has bugs...
    or (is_old_scipy and _blas_info() == "INTEL MKL")
)
```

That's all I meant, sorry.

setup.cfg
```diff
@@ -33,6 +33,7 @@ setup_requires =
     numpy>=1.16.6,<1.20
     scipy>=1.0
     cython>=0.29.20
+    packaging
```
jakelishman, line 36: You need to specify this in both `setup_requires` and `install_requires` - the former is about the build process, the latter is about what you need at installation.

jakelishman (commented):
You're right that the test is in a much better place here.  I'm not certain what your point was about putting the eigenvalues tests in with the eigensystem - eigenvalues are just part of an eigensystem, so it seems pretty relevant?  At any rate, the first step to finding eigenvectors in any numerical method is to find the eigenvalues first, so they're always super linked, and this file is fine.

About the test timing: you can use `rand_herm(2000, density=0.01)` to get a much more sparse density matrix to work with.  On my machine, that reduced the test runtime by one-half, so maybe it'll help?  I don't know if the problem still appears with those matrices, though.  The potential failure from not testing this is very large, so if it costs us ~20 seconds to test it on CI, we're just going to have to eat it unfortunately.  But this is the reason we (ought to) work to have fast tests everywhere else - it gives us the time budget to spend when we need it.

Thanks a lot for putting in the time to sort this out, and working it into a minimal test for us.  Sorry for being rather short/exacting earlier - I stressed myself out a bit at work, and I should have taken a bit of time off.  Thanks for the work you've been putting in tracking this down, I really do appreciate it.

MrRobot2211:
Hi, thak you for the feedback it is really helpful. Using lower density matrices is reducing  my time by 10%, and without the fix it produces error; it is an excellent finding.

MrRobot2211:
This is weird it is showing an apparently unrelated failure while saving and loading pickles.

And the CI times seem longer.

MrRobot2211:
Looks like we are right at the edge of maximum permited time, and it fluctuates due to all the random matrix generators. Maybe we could limit the number of repetitions for every test to 5?

jakelishman:
I've just pushed a couple of commits to fix up a little bit of the formatting (it was the squashed lines I meant, rather than the particular level of indent or location of the keyword `or`) and to slightly fix the tolerance of the eigenvalue test.  The individual uncertainty in any eigenvalue is allowed to be ~1e-12, so when you sum them all up to compare them, you have to propagrate the uncertainty through.

Don't worry about the random test failure you got in a pickle test one time - that wasn't your fault and we know about it and have fixed it for the next major release of QuTiP already.

Also don't worry about the speed of the CI tests.  You can't reliably benchmark by comparing CI runs because you never know what else was running on the physical hardware at the same time.  Travis are ramping down the amount of processing power available to us while they're transition business model.  I'll be moving our testing infrastructure elsewhere when I've got a bit more time to sort it out, and have decided what we actually want/need/can get.  In the mean time, we just live with the occasional test failure if a Mac Xcode12 machine is overloaded when we get to it.  We can rerun them if we need to; it's not worth compromising on our test coverage.

jakelishman:
Also, this added test runs in less than 1 second on my machine.  Even an overloaded CI machine isn't going to take excessively long on it.

jakelishman (approved):

MrRobot2211:
I have a couplle of questions to understande better the difference between using `assert_all_close`, `assert` and `assert_equal`. 
In this case you changed `assert_all_close` to `assert` because of the time difference right? 
Why are the other tests in the same file using `assert_equal` instead of `assert` ?

jakelishman:
In general in pytest we use the bare keyword `assert` for almost everything.  All the functions in `np.testing` are from the "before times", when test tooling wasn't as good as it is nowadays - they were a big improvement on available warning messages back before 2015.  Nowadays, though, `pytest` does introspection on assertion failures, so it has even better message reporting than the `np.testing` functions, and it lets you write much clearer code.  For example, `assert_equal(x, y)` is much better spelled in pytest as `assert x == y`, because it's much faster to read and pytest does better error reporting for it.  Similarly, testing for exceptions is better with
```python
with pytest.raises(ValueError):
    call_that_raises(1, "hello, world")
```
as opposed to the numpy form `np.testing.assert_raises(ValueError, call_that_raises, 1, "hello, world")` - the pytest one is much easier to read, and the error message is better.

The only numpy one we still use is `np.testing.assert_allclose` when we're comparing several numpy arrays.  That's just because `pytest` doesn't have a good function that's aware of multidimensional arrays (at least not one that I know).

In this case, I changed `np.testing.assert_allclose(x, y, atol=tol)` to `assert abs(x - y) < tol`.  I removed `np.testing.assert_allclose`, because that gives the mistaken impression that we're comparing arrays (we're not, we're comparing two scalars).  I could have changed it to `assert x == pytest.approx(y, tol=tol)`, except in this case the test file hasn't been converted to `pytest` style yet, and I wanted to keep the total change of this PR as small as possible.

PR #1499, opened 2021-04-16
Adds hyperlinks to docstrings of Qobj and Qobjevo class

purva-thakre:
**Description**
Adds hyperlinks for functions, classes, attributes etc. wherever possible. 

- Deleted methods table. See [this comment](https://github.com/qutip/qutip/pull/1499#issuecomment-822679427). 
- Every function, class etc is linked by default method of ``:obj:`.some_func` ``
- Functions not in API doc are referenced as  'func_not_in_API'. 
- Added `TypeError` and `ValueError` etc. in docstrings when an error of either type is raised by a function. Most of these might not be needed as [this link](https://numpydoc.readthedocs.io/en/latest/format.html#raises) advises to provide information about only the obvious errors. I read the link after making the changes. If needed, the new additions can be removed from the docstrings. 

**Related issues or PRs**
[One of the fixes for issue 71 in documentation](https://github.com/qutip/qutip-doc/issues/71). 

**Changelog**
Add hyperlinks in documentation and errors in docstrings of functions in qobj and qobevo

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -1101,7 +1101,7 @@ class Qobj(object):
         Returns
         -------
         data : array
-            Array of complex data from quantum object's :obj:`.data` attribute.
+            Array of complex data from quantum object's `data` attribute.
         """
         if squeeze:
             return self.data.toarray(order=order).squeeze()
@@ -1541,8 +1541,7 @@ class Qobj(object):
 
         Raises
         ------
-        ValueError : Must be a Hermitian operator to remove negative
-        eigenvalues
+        ValueError : Must be a Hermitian operator
             When input operator is not Hermitian.
 
         ValueError : Method not recognized
@@ -1737,7 +1736,7 @@ class Qobj(object):
             Array of eigenvalues for operator.
 
         eigvecs : array
-            Array of quantum operators representing the oprator eigenkets.
+            Array of quantum operators representing the operator eigenkets.
             Order of eigenkets is determined by order of eigenvalues.
 
         Notes
@@ -1868,7 +1867,7 @@ class Qobj(object):
         return out
 
     def extract_states(self, states_inds, normalize=False):
-        """:obj:`.Qobj` with states in ``states_inds`` only.
+        """:obj:`.Qobj` with states in `states_inds` only.
 
         Parameters
         ----------
@@ -1886,12 +1885,12 @@ class Qobj(object):
         -------
         q : :obj:`.Qobj`
             A new instance of :obj:`.Qobj` that contains only the states
-            corresponding to the indices in ``states_inds``.
+            corresponding to the indices in `states_inds`.
 
         Raises
         -------
         TypeError : Can only eliminate states from operators or state vectors
-            If input is not ket, bra or an operator.
+            If input is not :obj:`.ket`, :obj:`.bra` or an :obj:`.operators`.
 
         Notes
         -----
@@ -1911,7 +1910,7 @@ class Qobj(object):
         return q.unit() if normalize else q
 
     def eliminate_states(self, states_inds, normalize=False):
-        """Creates a new quantum object with states in ``states_inds`` eliminated.
+        """Creates a new quantum object with states in `states_inds` eliminated.
 
         Parameters
         ----------
@@ -1929,7 +1928,7 @@ class Qobj(object):
         -------
         :obj:`.Qobj`
             A new instance of :obj:`.Qobj` that contains only the states
-            corresponding to indices that are **not** in ``states_inds``.
+            corresponding to indices that are **not** in `states_inds`.
 
         Notes
         -----
@@ -2172,8 +2171,7 @@ class Qobj(object):
 
         Raises
         ------
-        TypeError : Unrecognized format for specification of time-dependent
-        Qobj
+        TypeError : Unrecognized format for specification of time-dep Qobj
             When type of `qobj_list` cannot be recognized - :obj:`.Qobj` or
             list etc.
 
@@ -2345,7 +2343,7 @@ def dims(inpt):
     Raises
     -------
     TypeError : Input is not a quantum object
-        When input cannot be identified as :obj:`.Qobj`. 
+        When input cannot be identified as :obj:`.Qobj`.
 
     Notes
     -----
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -100,15 +100,16 @@ class Qobj(object):
     inpt : array_like
         Data for vector/matrix representation of the quantum object.
     dims : list
-        Dimensions of object used for tensor products.
+        Dimensions of object used for :obj:`.tensor` products via :obj:`.dims`.
     shape : list
-        Shape of underlying data structure (matrix shape).
+        Shape of underlying data structure (matrix shape) via
+        :obj:`.dims_to_tensor_shape`.
     copy : bool
         Flag specifying whether :class:`Qobj` should get a copy of the
         input data, or use the original.
     fast : bool
         Flag for fast :class:`Qobj` creation when running ODE solvers like
-        :func:`qutip.mesolve` and :func:`qutip.mcsolve`.
+        :func:`~qutip.mesolve` and :func:`~qutip.mcsolve`.
         This parameter is used internally only.
 
 
@@ -117,26 +118,26 @@ class Qobj(object):
     data : array_like
         Sparse matrix characterizing the quantum object.
     dims : list
-        List of dimensions keeping track of the tensor structure.
+        List of dimensions keeping track of the :obj:`.tensor` structure.
     shape : list
         Shape of the underlying `data` array.
     type : str
         Type of quantum object: :func:`~qutip.states.bra`,
-        :func:`~qutip.states.ket`, 'oper' (:func:`~qutip.qobj.isoper`),
-        'operator-ket' (:func:`~qutip.qobj.isoperket`),
-        'operator-bra' (:func:`~qutip.qobj.isoperbra`), or 'super'
-        (:func:`~qutip.qobj.issuper`).
+        :func:`~qutip.states.ket`, 'oper' (:obj:`.isoper`),
+        'operator-ket' (:obj:`.isoperket`),
+        'operator-bra' (:obj:`.isoperbra`), or 'super'
+        (:obj:`.issuper`).
     superrep : str
         Representation used if ``type`` is 'super'
-        (:func:`~qutip.qobj.issuper`). One of 'super'
+        (:obj:`.issuper`). One of 'super'
         (:func:`~qutip.superoperator.liouvillian` form) or 'choi'
         (Choi matrix with tr = dimension via :mod:`~qutip.superop_reps`).
     isherm : bool
         Indicates if quantum object represents Hermitian operator via
-        :func:`~qutip.qobj.isherm`.
+        :obj:`.isherm`.
     isunitary : bool
-        Indictaes if quantum object represents unitary operator via
-        :func:`~qutip.qobj.check_isunitary`.
+        Indicates if quantum object represents unitary operator via
+        :meth:`check_isunitary`.
     iscp : bool
         Indicates if the quantum object represents a map, and if that map is
         completely positive (CP).
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -104,10 +104,11 @@ class Qobj(object):
     shape : list
         Shape of underlying data structure (matrix shape).
     copy : bool
-        Flag specifying whether Qobj should get a copy of the
+        Flag specifying whether :class:`Qobj` should get a copy of the
         input data, or use the original.
     fast : bool
-        Flag for fast qobj creation when running ode solvers.
+        Flag for fast :class:`Qobj` creation when running ODE solvers like
+        :func:`qutip.mesolve` and :func:`qutip.mcsolve`.
         This parameter is used internally only.
 
 
@@ -885,7 +886,7 @@ class Qobj(object):
         return s
 
     def dag(self):
-        """Adjoint operator of quantum object.
+        """Adjoint operator of quantum object (:class:`Qobj`).
         """
         out = Qobj()
         out.data = zcsr_adjoint(self.data)
@@ -912,7 +913,7 @@ class Qobj(object):
         return J_dual
 
     def conj(self):
-        """Conjugate operator of quantum object.
+        """Conjugate operator of quantum object (:class:`Qobj`).
         """
         out = Qobj()
         out.data = self.data.conj()
@@ -920,7 +921,7 @@ class Qobj(object):
         return out
 
     def norm(self, norm=None, sparse=False, tol=0, maxiter=100000):
-        """Norm of a quantum object.
+        """Norm of a quantum object (:class:`Qobj`).
 
         Default norm is L2-norm for kets and trace-norm for operators.
         Other ket and operator norms may be specified using the `norm` and
@@ -1051,7 +1052,7 @@ class Qobj(object):
         Returns
         -------
         data : array
-            Array of complex data from quantum objects `data` attribute.
+            Array of complex data from quantum object's ``data`` attribute.
         """
         if squeeze:
             return self.data.toarray(order=order).squeeze()
@@ -1065,7 +1066,7 @@ class Qobj(object):
         return self.full()
 
     def diag(self):
-        """Diagonal elements of quantum object.
+        """Diagonal elements of quantum object (:class:`Qobj`).
 
         Returns
         -------
@@ -1120,7 +1121,7 @@ class Qobj(object):
         return out.tidyup() if settings.auto_tidyup else out
 
     def check_herm(self):
-        """Check if the quantum object is hermitian.
+        """Check if the quantum object (:class:`Qobj`) is hermitian.
 
         Returns
         -------
@@ -1522,7 +1523,7 @@ class Qobj(object):
         Parameters
         -----------
         bra : :class:`qutip.Qobj`
-            Quantum object of type 'bra' or 'ket'
+            Quantum object of type 'bra' 
 
         ket : :class:`qutip.Qobj`
             Quantum object of type 'ket'.
@@ -1778,7 +1779,7 @@ class Qobj(object):
         return out
 
     def extract_states(self, states_inds, normalize=False):
-        """Qobj with states in state_inds only.
+        """:class:`Qobj` with states in ``states_inds`` only.
 
         Parameters
         ----------
@@ -1796,7 +1797,7 @@ class Qobj(object):
         -------
         q : :class:`qutip.Qobj`
             A new instance of :class:`qutip.Qobj` that contains only the states
-            corresponding to the indices in `state_inds`.
+            corresponding to the indices in `states_inds`.
 
         Notes
         -----
@@ -1816,7 +1817,7 @@ class Qobj(object):
         return q.unit() if normalize else q
 
     def eliminate_states(self, states_inds, normalize=False):
-        """Creates a new quantum object with states in state_inds eliminated.
+        """Creates a new quantum object with states in ``states_inds`` eliminated.
 
         Parameters
         ----------
@@ -1824,7 +1825,7 @@ class Qobj(object):
             The states that should be removed.
 
         normalize : True / False
-            Weather or not the new Qobj instance should be normalized (default
+            Weather or not the new :class:`Qobj` instance should be normalized (default
             is False). For Qobjs that represents density matrices or state
             vectors normalized should probably be set to True, but for Qobjs
             that represents operators in for example an Hamiltonian, normalize
@@ -1832,9 +1833,9 @@ class Qobj(object):
 
         Returns
         -------
-        q : :class:`qutip.Qobj`
-            A new instance of :class:`qutip.Qobj` that contains only the states
-            corresponding to indices that are **not** in `state_inds`.
+        q : :class:`Qobj`
+            A new instance of :class:`Qobj` that contains only the states
+            corresponding to indices that are **not** in ``states_inds``.
 
         Notes
         -----
@@ -1964,7 +1965,7 @@ class Qobj(object):
 
     def check_isunitary(self):
         """
-        Checks whether :class:`QobjEvo`  is a unitary matrix
+        Checks whether :class:`Qobj` is a unitary matrix
         """
         if self.isoper:
             eye_data = fast_identity(self.shape[0])
@@ -2062,11 +2063,11 @@ class Qobj(object):
             The time for which to evaluate the time-dependent Qobj instance.
         args : dictionary
             A dictionary with parameter values required to evaluate the
-            time-dependent Qobj intance.
+            time-dependent Qobj instance.
 
         Returns
         -------
-        output : :class:`qutip.Qobj`
+        output : :class:`Qobj`
             A Qobj instance that represents the value of qobj_list at time t.
 
         """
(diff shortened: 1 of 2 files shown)


hodgestar:
@purva-thakre Thank you for opening this! Let us know when it's ready and we'll review. Please ask you have any questions or get stuck with anything.

purva-thakre:
@hodgestar I do not know how to link any of the methods listed in a table like [ here](http://qutip.org/docs/latest/apidoc/classes.html#qobj). I tried to link [`expm()`](https://github.com/qutip/qutip/blob/67c46870749b6e5925feed106a493eac9124bcc3/qutip/qobj.py#L177) using `:meth:expm()`. When I preview the html build, entire `:meth:expm()` shows up in the table instead of a link to expm() method. 

image


I think this has to do with how `:` are used by[ python domain](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-python-domain) and[ toctree](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree) (for API documentation - https://github.com/qutip/qutip-doc/blob/master/apidoc/apidoc.rst#api-documentation). 

Appreciate any pointers about this.

BoxiLi:
Have you tried
```
:meth:`Qobj.expm`
```
?
First, there shouldn't be any `()` in the reference to a function/method. And second, only with :meth:\`expm\` I think Sphinx will try to look for a function `expm` in the public API, which does not exist. Because this one is a method under the class `Qobj`.

purva-thakre:
@BoxiLi Yes, I have not used parentheses for functions not in the public API. These changes are still local and haven't been pushed yet. 

My issue is with the `Methods` table for `Qobj` class. When I try the same edit for the code lines rendered as a table, there's no hyperlinks in the table. The entire code line shows up as it is. See screenshot below : 
image

I think this is due to how Sphinx tables are defined and a simialr manner to define python object references as well.

jakelishman:
Well for one that table of methods just shouldn't exist - it can all be deleted.  It's a holdover from a previous method of doing API documentation, but now since we use numpydoc, the equivalent gets done automatically for us right below where that table appears in the docs.  You'll not succeed at any sort of formatting in that table, because all the headers are parsed as literal strings, and (mostly) aren't subject to rst formatting at all.

Second, should you want it elsewhere, the minimal magic incantation is `` :meth:`expm` `` if you're inside the object you're referring to.  If you're outside it, use a qualified `` :meth:`Qobj.copy` `` (can also be used to refer to other objects' methods from inside an object).

purva-thakre:
Thanks ! I have removed the table. 

I wanted to check if I should link to functions not in public API ? I could create references to such functions via their permalinks in Github's repo. I would have to add [external links](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages) for each function or link the function's location by [cross referencing the file](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-documents).

purva-thakre:
Thanks ! I have removed the table. 

Couple of questions : 

- What's the preferred method to link functions not in public API ? 
I could create a sphinx reference label to link to these functions either by using Github's permalink or a path to such functions in the directory. You can see this when I try to link `check_isunitary` which is not in the API but is defined in `Qobj` file in the linked branch. 
- Qobj vs qobj link : 
There are problems with linking functions for `Qobj` class. For example, if I want to link to [`isoperbra`](http://qutip.org/docs/latest/apidoc/functions.html#qutip.qobj.isoper) then the only link that will work is `qutip.qobj.isoperbra` not `qutip.Qobj.isoperbra`. If latter is preferred then I will have to create `ref` labels for each of these functions as stated above. More examples can be observed in the `superrep` attributes of `Qobj` class of linked branch.

BoxiLi:
> What's the preferred method to link functions that are not public APIs?

I don't think we should link to non-public API in the doc. Non-public API should not (at least no encouraged) be used outside of QuTiP because we may change it without issuing a deprecation warning. Besides, [`check_unitary`](http://qutip.org/docs/latest/apidoc/classes.html?highlight=check_isunitary#qutip.Qobj.check_isunitary) is a public API, it is a class method under `Qobj`, rather than a function, see bellow.

> Qobj vs qobj link

`qutip.qobj.isoperbra` and `qutip.Qobj.isoperbra` are different. The lower case `qobj` is a submodule in QuTiP (qobj.py) while `Qobj` is a class. `isoperbra` is not a class method under `Qobj`. It is a function defined under the submodule `qobj`. So only `qutip.qobj.isoperbra` is the correct path.

To make the life simpler, I would recommend to use the shortcut ``:func:`.isoperbra` `` for functions and ``:meth:`.Qobj.class_method` ``. Or even simpler: ``:obj:`.isoperbra` ``. Sphinx will automatically look for the correct match. As long as there are no two functions with the same name, we are safe. This isalso because `qutip.qobj.isoperbra` will be a wrong path in qutip 5.0. The file is moved.

nathanshammah:
@purva-thakre, is this still a draft PR?

purva-thakre:
@nathanshammah  Yes, I still have to make a couple of changes to this.

qutip/qobj.py
```diff
@@ -978,15 +952,15 @@ def conj(self):
...
+        Other ket and operator norms may be specified using the :obj:`.norm`
+        and argument.
 
         Parameters
         ----------
         norm : str
-            Which norm to use for ket/bra vectors: L2 'l2', max norm 'max',
-            or for operators: trace 'tr', Frobius 'fro', one 'one', or max
+            Which norm to use for ket/bra vectors: L2, max norm 'max',
+            or for operators: trace `tr', Frobius 'fro', one 'one', or max
```
purva-thakre, line 963: Change ` to '

qutip/qobj.py
```diff
@@ -978,16 +952,18 @@ def conj(self):
     def norm(self, norm=None, sparse=False, tol=0, maxiter=100000):
         """Norm of a quantum object.
 
-        Default norm is L2-norm for kets and trace-norm for operators.
-        Other ket and operator norms may be specified using the `norm` and
-        argument.
+        Default norm is `L2-norm` for kets and `trace-norm` for operators.
```
jakelishman, line 955: Why are these back-ticks added?  For one, single backticks just produce italic text (not monospace), and two, L2-norm and trace-norm are just words, so don't need any formatting?

qutip/qobj.py
```diff
@@ -978,16 +952,18 @@ def conj(self):
...
+        Other ket and operator norms may be specified using the :obj:`.norm`
+        and argument.
```
jakelishman, line 957: No need for the word "and"

qutip/qobj.py
```diff
@@ -978,16 +952,18 @@ def conj(self):
...
 
         Parameters
         ----------
         norm : str
-            Which norm to use for ket/bra vectors: L2 'l2', max norm 'max',
-            or for operators: trace 'tr', Frobius 'fro', one 'one', or max
-            'max'.
+            Norm to use for :obj:`.ket` or :obj:`.bra` vectors: L2,
```
jakelishman, line 962: don't link `ket` and `bra`

qutip/qobj.py
```diff
@@ -978,16 +952,18 @@ def conj(self):
...
+            max norm 'max',
+
+            Norm for :obj:`.operators` : trace 'tr', Frobius 'fro', one 'one',
```
jakelishman, line 965: don't link operators, it's just a word

qutip/qobj.py
```diff
@@ -1041,13 +1025,18 @@ def proj(self):
 
         Parameters
         ----------
-        Q : :class:`qutip.Qobj`
-            Input bra or ket vector
+        Q : :obj:`.Qobj`
+            Input :obj:`.bra` or :obj:`.ket` vector
```
jakelishman, line 1029: don't link bra and ket

qutip/qobj.py
```diff
@@ -953,6 +915,12 @@ def dag(self):
     def dual_chan(self):
         """Dual channel of quantum object representing a completely positive
         map.
+
+        Raises
+        -------
+        ValueError : Dual channels are only implemented for CP maps.
+            If the input is not a completely positive map.
```
jakelishman, line 922: you don't need to copy in the error message - it just adds in an extra location for us to forget to update it if it changes

qutip/qobj.py
```diff
@@ -1003,9 +979,17 @@ def norm(self, norm=None, sparse=False, tol=0, maxiter=100000):
...
+        float
             The requested norm of the operator or state quantum object.
 
+        Raises
+        -------
+        ValueError : For matrices, norm must be 'tr', 'fro', 'one', or 'max'
+            When norm chosen is not from default methods of operators.
+
+        ValueError : For vectors, norm must be 'l2', or 'max'
+            When norm chosen is not from default methods of bra/kets.
```
jakelishman, line 991: you don't need to have a "Raises" section for "obvious" errors like "passed a completely invalid parameter"

qutip/qobj.py
```diff
@@ -1041,13 +1025,18 @@ def proj(self):
...
         -------
-        P : :class:`qutip.Qobj`
-            Projection operator.
+        :obj:`.Qobj`
+            Projection operator of a state vector.
+
+        Raises
+        -------
+        TypeError : Projector can only be formed from a bra or ket
+            If the input is not a ket or bra.
```
jakelishman, line 1039: unnecsesary section

qutip/qobj.py
```diff
@@ -1065,7 +1054,7 @@ def tr(self):
 
         Returns
         -------
-        trace : float
+        float
```
jakelishman, line 1057: ?

qutip/qobj.py
```diff
@@ -1081,6 +1070,11 @@ def purity(self):
             For a pure state, the purity is 1.
             For a mixed state of dimension `d`, 1/d<=purity<1.
 
+        Raises
+        -------
+        TypeError : Purity is defined on a density matrix or state
+            If the input is a superoperator.
+
         """
```
jakelishman, line 1078: not necessary

qutip/qobj.py
```diff
@@ -1107,7 +1101,7 @@ def full(self, order='C', squeeze=False):
         Returns
         -------
         data : array
-            Array of complex data from quantum objects `data` attribute.
+            Array of complex data from quantum object's `data` attribute.
```
jakelishman, line 1104: should be two backticks around "data"

qutip/qobj.py
```diff
@@ -1137,31 +1131,32 @@ def diag(self):
...
 
         Raises
         ------
-        TypeError
-            Quantum operator is not square.
+        TypeError : Invalid operand for matrix exponential
+            When dimensions do not match i.e. input is not a square operator.
+
+        ValueError : Method must be 'dense' or 'sparse'
+            When available deafult method is not chosen.
```
jakelishman, line 1155: don't copy error messages, also the ValueError is an "obvious" error because the allowed values are specified

qutip/qobj.py
```diff
@@ -1202,13 +1197,14 @@ def sqrtm(self, sparse=False, tol=0, maxiter=100000):
...
+        :obj:`.Qobj`
             Matrix square root of operator.
 
         Raises
         ------
-        TypeError
-            Quantum object is not square.
+        TypeError : Invalid operand for matrix square root
+            When input is not a square operator i.e. rows and columns are not
+            equal in size.
```
jakelishman, line 1207: not sure why you changed this - the original is very clear

qutip/qobj.py
```diff
@@ -1234,48 +1230,50 @@ def sqrtm(self, sparse=False, tol=0, maxiter=100000):
...
+        :obj:`.Qobj`
+            Matrix cosine of an operator.
 
         Raises
         ------
-        TypeError
-            Quantum object is not square.
+        TypeError : Invalid operand for matrix square root.
+            When input is not a square operator, number of rows and columns
+            do not match.
```
jakelishman, line 1246: same comment

qutip/qobj.py
```diff
@@ -1284,19 +1282,20 @@ def sinm(self):
...
+        :obj:`.Qobj`
             Matrix inverse of operator.
 
         Raises
         ------
-        TypeError
-            Quantum object is not square.
+        TypeError : Invalid operand for matrix inverse
+            When input is not square, number of rows and columns
+            do not match.
```
jakelishman, line 1298: same comment

qutip/qobj.py
```diff
@@ -1311,14 +1310,14 @@ def unit(self, inplace=False,
...
-        Uses norm from Qobj.norm().
+        Uses norm from :obj:`.Qobj.norm()`.
 
         Parameters
         ----------
         inplace : bool
             Do an in-place normalization
         norm : str
-            Requested norm for states / operators.
+            Requested :obj:`.norm` for states / operators.
```
jakelishman, line 1320: just a word

qutip/qobj.py
```diff
@@ -1234,48 +1230,50 @@ def sqrtm(self, sparse=False, tol=0, maxiter=100000):
...
-        TypeError
-            Quantum object is not square.
+        TypeError : Invalid operand for matrix square root
+            When input is not a square operator, number of rows and columns
+            do not match.
 
         Notes
         -----
-        Uses the Q.expm() method.
+        Uses the Q. :obj:`.expm()` method.
```
jakelishman, line 1276: this looks very weird in the rendered version.  You can just say "Uses the `` :obj:`~.expm` `` method" - it's clear from context that it refers to `Qobj`, and avoids the weird split between the Q and the monospace font.

qutip/qobj.py
```diff
@@ -1328,10 +1327,15 @@ def unit(self, inplace=False,
...
-        oper : :class:`qutip.Qobj`
+        oper : :obj:`.Qobj`
             Normalized quantum object if not in-place,
             else None.
 
+        Raises
+        -------
+        Exception : inplace kwarg must be bool
+            Inplace must be True or False.
+
```
jakelishman, line 1338: unnecsesary

qutip/qobj.py
```diff
@@ -1408,11 +1412,11 @@ def tidyup(self, atol=settings.auto_tidyup_atol):
         ----------
         atol : float
             Absolute tolerance used by tidyup. Default is set
-            via qutip global settings parameters.
+            via qutip global settings parameters in :obj:`.Options`.
```
jakelishman, line 1415: The link here isn't correct.  The linked options are the solver options, whereas the global atol settings are in `qutip.settings`.

qutip/qobj.py
```diff
@@ -1448,14 +1452,21 @@ def transform(self, inpt, inverse=False, sparse=True):
...
+        :obj:`.Qobj`
             Operator in new basis.
 
+        Raises
+        ------
+        TypeError : Invalid size of ket list for basis transformation
+            When input is a list or array and the dimensions do not match.
+
+        TypeError : Invalid operand for basis transformation
+            When input is not a proper quantum object, array or operator.
```
jakelishman, line 1464: This one isn't necessary

qutip/qobj.py
```diff
@@ -1525,8 +1536,16 @@ def trunc_neg(self, method="clip"):
...
+        :obj:`.Qobj`
+            A valid density operator as a quantum object.
+
+        Raises
+        ------
+        ValueError : Must be a Hermitian operator
+            When input operator is not Hermitian.
+
+        ValueError : Method not recognized
+            If method other than 'clip' or 'sgs' is specified.
```
jakelishman, line 1548: this one's not necessary

qutip/qobj.py
```diff
@@ -1577,21 +1596,29 @@ def matrix_element(self, bra, ket):
 
         Parameters
         -----------
-        bra : :class:`qutip.Qobj`
-            Quantum object of type 'bra' or 'ket'
+        bra : :obj:`.Qobj`
+            Quantum object of type :obj:`.bra`
```
jakelishman, line 1600: don't link `bra`, and you've also changed the description.  It's allowed to pass a "ket" in this argument.

qutip/qobj.py
```diff
@@ -1577,21 +1596,29 @@ def matrix_element(self, bra, ket):
...
 
-        ket : :class:`qutip.Qobj`
-            Quantum object of type 'ket'.
+        ket : :obj:`.Qobj`
+            Quantum object of type :obj:`.ket`.
```
jakelishman, line 1603: don't link `ket`.

qutip/qobj.py
```diff
@@ -1577,21 +1596,29 @@ def matrix_element(self, bra, ket):
...
+        complex
             Complex valued matrix element.
 
+        Raises
+        -------
+        TypeError : Can only get matrix elements for an operator
+            When input is not a valid operator.
+
+        TypeError : Can only calculate matrix elements for bra  and ket vectors
+            When input is not a valid ket or bra vector.
```
jakelishman, line 1616: no necessary

qutip/qobj.py
```diff
@@ -1577,21 +1596,29 @@ def matrix_element(self, bra, ket):
...
+
         Notes
         -----
-        It is slightly more computationally efficient to use a ket
-        vector for the 'bra' input.
+        It is slightly more computationally efficient to use a `ket`
+        vector for the `bra` input.
```
jakelishman, line 1621: the "ket" should be left with no formatting (it's just a word), and `bra` should have 2 backticks to make it monospaced rather than italic

qutip/qobj.py
```diff
@@ -1687,7 +1714,7 @@ def eigenstates(self, sparse=False, sort='low', eigvals=0,
             Use sparse Eigensolver
 
         sort : str
-            Sort eigenvalues (and vectors) 'low' to high, or 'high' to low.
+            Sort eigenvalues (and vectors) 'low' to 'high', or 'high' to 'low'.
```
jakelishman, line 1717: This change has removed the meaning.  I'm not saying the original was very clear, but the intent was that a value of 'low' refers to a low-to-high sort and 'high' means high-to-low.

qutip/qobj.py
```diff
@@ -1750,7 +1777,7 @@ def eigenenergies(self, sparse=False, sort='low',
         sparse : bool
             Use sparse Eigensolver
         sort : str
-            Sort eigenvalues 'low' to high, or 'high' to low.
+            Sort eigenvalues 'low' to 'high', or 'high' to 'low'.
```
jakelishman, line 1780: same comment

qutip/qobj.py
```diff
@@ -1794,9 +1821,15 @@ def groundstate(self, sparse=False, tol=0, maxiter=100000, safe=True):
...
-        eigvec : :class:`qutip.Qobj`
+        eigvec : :obj:`.Qobj`
             Eigenket for the ground state of quantum operator.
 
+        Raises
+        -------
+        WARNING : Ground state may be degenerate.
+            Use Q. :obj:`.eigenstates()`
+
+
```
jakelishman, line 1832: there's no need to mention warnings at all, but if you were to, they'd go in a "Warns" section, not "Raises".

qutip/qobj.py
```diff
@@ -1834,25 +1867,30 @@ def trans(self):
         return out
 
     def extract_states(self, states_inds, normalize=False):
-        """Qobj with states in state_inds only.
+        """:obj:`.Qobj` with states in `states_inds` only.
```
jakelishman, line 1870: broken formatting of states_inds

qutip/qobj.py
```diff
@@ -1834,25 +1867,30 @@ def trans(self):
...
+            normalize should be ``False``.
 
         Returns
         -------
-        q : :class:`qutip.Qobj`
-            A new instance of :class:`qutip.Qobj` that contains only the states
-            corresponding to the indices in `state_inds`.
+        q : :obj:`.Qobj`
+            A new instance of :obj:`.Qobj` that contains only the states
+            corresponding to the indices in `states_inds`.
```
jakelishman, line 1888: broken formatting

qutip/qobj.py
```diff
@@ -1834,25 +1867,30 @@ def trans(self):
...
         states_inds : list of integer
             The states that should be kept.
 
         normalize : True / False
-            Weather or not the new Qobj instance should be normalized (default
-            is False). For Qobjs that represents density matrices or state
-            vectors normalized should probably be set to True, but for Qobjs
-            that represents operators in for example an Hamiltonian, normalize
-            should be False.
+            Weather or not the new :class:`Qobj` instance should be normalized
```
jakelishman, line 1878: should be "Whether"

jakelishman (changes requested):
Thank you for making the PR - I saw lots of fixed typos, and places where added links definitely helped the documentation.

However, I only got part way through, but really there's also a lot of changes that aren't appropriate here, sorry.  Most of `Qobj` and `QobjEvo` already do link when they should (though it's not quite 100%).  A few points:

- Don't put a link on a word if it doesn't _specifically_ refer to the object you're linking to.  For example, the words "ket", "bra" and "operator" should almost _never_ have links - they're just words, and they're not referring to any function or module.  The word "tensor" usually just means "tensor", it shouldn't be a link to the module.  The reason not to do this is that it makes it seem like we're giving a special meaning to common words, and it dilutes the use of "real" links.
- "Raises" sections should be used very sparingly.  There's no need to point out that passing an incorrect value or an incorrect type will cause a `ValueError` or `TypeError`, for example.  The reason is that this just dilutes the available information - it's _always_ the case that an incorrect value will raise `ValueError`, so the programmer doesn't get any new information from this.  It's more appropriate to use the "Parameters" section to document what's allowed; a programmer can assume that if they violate that, a suitable exception will be raised.  The "Raises" section should only be for very non-obvious errors, or errors that even a programmer who has read the "Parameters" section might reasonably make accidentally.
- Don't put the error message in the description of a "Raises" section.  The message is intended to inform the user when it appears, but when you're writing a "Raises" section, that's the purpose of the description below the type of the exception; there's no additional information gained from having the message, and it clutters up the docs.
- Take care not to change the meaning of sentences when you're adding in additional formatting - I saw a few places where changes in the formatting accidentally deleted a word, or hid some extra meaning.
- In RST, single backticks (`` ` ``) produce italics, not monospaced text.
- There's no need to change `:class:`/`:func:`/`:meth:` directives into `:obj:` directives - either style is totally fine.  Don't worry about it now that you've done it, though, it's fine to leave them in the new style (because either is fine).

Do you have the docs building correctly? It's worth checking to make sure that the changes you made are showing up the way you expected.  In the worst case scenario, GitHub now builds them for you whenever you push changes.  Click on the "Checks" tab at the top of the PR, and on the right of the screen there's an "Artifacts" button - you can download the file there, and open "index.html" in your web browser to get the full set of rendered docs.

purva-thakre:
> Don't put a link on a word if it doesn't specifically refer to the object you're linking to

Yeah, I was worried if linking words like ket, bra were needed or not. I'll remove them. 

>  The "Raises" section should only be for very non-obvious errors
>  Don't put the error message in the description of a "Raises" section

No problem. I will remove errors created due to incorrect parameters and other obvious errors + error messages. 

>  I saw a few places where changes in the formatting accidentally deleted a word, or hid some extra meaning

I think the accidental deletes might have been due to getting caught in some cut/copy/paste flow. 

> Do you have the docs building correctly?

Yes, I do. I was a bit confused about how to try to format to functions not in API doc. So, I still tried to link a `ref` to them so that they are formatted similar to hyperlinks. I think a couple of these appear as italics.

qutip/qobj.py
purva-thakre: Ok. I will put it back to what it is supposed to be. I was a bit confused about this because in some cases, variables / methods appear in double quotes, some appear in italics and some cases use double backticks for monospaced text.

PR #1501, opened 2021-04-17
Fix errors in gate definitions

gecrooks:
**Description**
Fix errors in gate definitions

**Changelog**
Correct expansion of berkeley and swapalpha gates
Correct definition of CZ gate
Add CPHASE gate to list of control gates


BoxiLi:
Thanks @gecrooks ! Looks good. Can't imagine how these typos have been in the code for so long. Gates like berkeley are not used by a lot.

PR #1493, opened 2021-04-10
Fix improperly formed docstrings

jakelishman:
This isn't a complete "fix" of every piece of bad formatting in the API documentation that we're building, but it does fix every warning that Sphinx produces in both the `html` and `latexpdf` targets.

It's quite a big patch, but there's no meaningful changes, only formatting ones - I haven't attempted to actually change the meaning of any of the writing, only to reformat it.  If we want to make any changes to any of the docstrings, we can do that in separate patches.  My intent is to merge this in as the very final patch in the 4.6 release, since it's useful for doing the documentation release.

See also qutip/qutip-doc#136

PR #1505, opened 2021-04-21
Fix more warnings in documentation build

jakelishman:
There were a few invalid escape sequences and scipy imports that somehow escaped being fixed #1493, so this fixes those.

Also, in #1480 we issued deprecation warnings for user-facing uses of `eseries`, `essolve` and `ode2es`, however we are not deprecating the correlation and spectrum methods (in QuTiP 5.0 the necessary components are just absorbed into private functions inside `correlation.py`), so these should not generate warnings.  Simon (@hodgestar) and I briefly discussed whether we needed a special warnings filter inside those functions, but we thought we didn't need it because default Python filters would prevent `DeprecationWarning` from being shown, since it is triggered by library code and not code in `__main__`.

The problem is that not everything uses the default filters; both `pytest` and the documentation build do not suppress all `DeprecationWarning` messages.  This was a particular problem in `pytest`, where approximately 68,000 warnings (!) were generated.  Since this is _not_ deprecated functionality, nor is it using deprecated functionality, so there shouldn't be any warnings.  This PR removes those as well.

The end result is that a completely clean documentation build now should generate 0 warnings.  My eventual intent is to create an automatic documentation build process, which treats warnings as errors to prevent us from breaking it in PRs.  This cannot happen until the merger of the documentation into this repository, however, because otherwise new features/feature removals could get into an impossible situation where the two PRs to `qutip` and `qutip-doc` would not be aware of each other, so would return failed CI.

Tests still generate several warnings, but that's for a different PR.

### DIFF
--- a/qutip/correlation.py
+++ b/qutip/correlation.py
@@ -39,6 +39,7 @@ __all__ = ['correlation_2op_1t', 'correlation_2op_2t', 'correlation_3op_1t',
 
 from re import sub
 from warnings import warn
+import warnings
 import types
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


hodgestar (approved):
Looks good! Thanks for sorting out the 68k deprecation warnings and sorry my suggestion didn't work out.

jakelishman:
Haha, it's no problem - I didn't spot it was a problem in that PR either.

I'm marginally worried that the test suite was even able to generate 68,000 warnings - it seems like the code was hit an excessive number of times.  I'm fairly sure that `correlation.py` hasn't been meaningfully touched in years (there are 8-year-old deprecation warnings in it...), but it might be time for one of us to review it.

PR #1506, opened 2021-04-22
Fix test warnings

jakelishman:
This PR fixes every warning reported in `pytest`, and then makes `pytest` treat warnings as errors to hopefully make sure we stick to this state of affairs.  (Or at least it fixes all the ones that appear on my particular machine...)

The reason this is desirable is that it means we cannot ignore deprecation warnings in dependencies or even in our own code.  It also helps us not let certain parts of the codebase get lost in the sands of time, or at least compels us to do something about potential problems before they actually become catastrophic.  This is _somewhat_ of a trial, though I hope we can stick to it; warnings usually indicate that something's wrong.

Tests can still test run code that generates warnings, but you now have to mark it explicitly in `pytest`.  This doesn't mean you should just "blank out" warnings by filtering them away.  You should first consider if what you're doing is the right way to do it (fix the warning), and then only if it absolutely is and you're testing something that is _supposed_ to warn, then you can use `pytest.warns` to _assert_ that it warns.

Of note: a `warnings.catch_warnings` block _inside_ a test (i.e. in test code, not module code) does not work when `pytest` is treating warnings as errors.

There are a few classes of fixes in this PR:

- filtering "expected" warnings from `pyximport` (hoping to remove QuTiP's dependence on this in 5.0 anyway)
- removal of the old `np.matrix` type in favour of `ndarray`, with some changes to accommodate the different arithmetic semantics
- convert an `__init__` method in a test into just class attributes to allow `pytest` to collect it
- bug fixes for behaviour that was raising warnings we hadn't been paying attention to

I'll be mightily surprised if all 5 CI runs pass immediately with warnings enabled; I haven't tested MKL on my laptop, and a quick glance at old logs suggests that SciPy 1.4 might be issuing a few more.

### DIFF
--- a/qutip/cy/pyxbuilder.py
+++ b/qutip/cy/pyxbuilder.py
@@ -32,9 +32,19 @@
 ###############################################################################
 import sys
 import os
+import warnings
 
 import numpy as np
-import pyximport
+
+with warnings.catch_warnings():
+    # TODO: pyximport loads the imp module, which is deprecated in favour of
+    # importlib and slated for removal in Python 3.12.  Our intent is to
+    # remove all usage of pyximport in QuTiP 5.0 with new Coefficient objects,
+    # before Python 3.12 is released.
+    warnings.filterwarnings(
+        "ignore", category=DeprecationWarning, module="pyximport",
+    )
+    import pyximport
 
 old_get_distutils_extension = pyximport.pyximport.get_distutils_extension
 
--- a/qutip/steadystate.py
+++ b/qutip/steadystate.py
@@ -472,8 +472,8 @@ def _steadystate_direct_sparse(L, ss_args):
 
 def _steadystate_direct_dense(L, ss_args):
     """
-    Direct solver that use numpy dense matrices. Suitable for
-    small system, with a few states.
+    Direct solver that uses numpy arrays. Suitable for small systems with few
+    states.
     """
     if settings.debug:
         logger.debug('Starting direct dense solver.')
@@ -483,14 +483,14 @@ def _steadystate_direct_dense(L, ss_args):
     b = np.zeros(n ** 2)
     b[0] = ss_args['weight']
 
-    L = L.data.todense()
+    L = L.full()
     L[0, :] = np.diag(ss_args['weight']*np.ones(n)).reshape((1, n ** 2))
     _dense_start = time.time()
     v = np.linalg.solve(L, b)
     _dense_end = time.time()
     ss_args['info']['solution_time'] = _dense_end-_dense_start
     if ss_args['return_info']:
-        ss_args['info']['residual_norm'] = la.norm(b - L*v, np.inf)
+        ss_args['info']['residual_norm'] = la.norm(b - L@v, np.inf)
     data = vec2mat(v)
     data = 0.5 * (data + data.conj().T)
 
--- a/qutip/steadystate.py
+++ b/qutip/steadystate.py
@@ -39,24 +39,23 @@ collapse operators.
 __all__ = ['steadystate', 'steady', 'build_preconditioner',
            'pseudo_inverse']
 
-import warnings
+import functools
 import time
-import scipy
+import warnings
+
+from packaging.version import parse as _parse_version
 import numpy as np
 from numpy.linalg import svd
-from scipy import prod
+import scipy
 import scipy.sparse as sp
 import scipy.linalg as la
-from scipy.sparse.linalg import (use_solver, splu, spilu, spsolve, eigs,
-                                 LinearOperator, gmres, lgmres, bicgstab)
+from scipy.sparse.linalg import (
+    use_solver, splu, spilu, eigs, LinearOperator, gmres, lgmres, bicgstab,
+)
 from qutip.qobj import Qobj, issuper, isoper
 
 from qutip.superoperator import liouvillian, vec2mat, spre
-from qutip.sparse import sp_permute, sp_bandwidth, sp_reshape, sp_profile
-
-from qutip.superoperator import liouvillian, vec2mat
-from qutip.sparse import (sp_permute, sp_bandwidth, sp_reshape,
-                          sp_profile)
+from qutip.sparse import sp_permute, sp_bandwidth, sp_profile
 from qutip.cy.spmath import zcsr_kron
 from qutip.graph import weighted_bipartite_matching
 from qutip import (mat2vec, tensor, identity, operator_to_vector)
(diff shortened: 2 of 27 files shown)


jakelishman:
Not a huge amount to be done about that CodeClimate failure to be honest - I'm not going to go through and change that function into something more sensible, because the whole type is _completely_ rewritten in 5.0 anyway.  If I remember correctly, that whole class is copied from SciPy too, with some small adjustments to the initialisation to speed it up.  The failure isn't anything new, it's just being flagged because my diff hit a return statement in a function with about 25 returns.

qutip/cy/pyxbuilder.py
```diff
@@ -32,9 +32,19 @@
 ###############################################################################
 import sys
 import os
+import warnings
 
 import numpy as np
-import pyximport
+
+with warnings.catch_warnings():
+    # pyximport loads the imp module, which is deprecated in favour of
```
hodgestar, line 40: Sometimes I find it useful to start such comments with "TODO:" so that one can easily search the code for things we wanted to sort out but might have forgotten about.

qutip/steadystate.py
```diff
@@ -72,6 +73,11 @@
     from qutip._mkl.spsolve import (mkl_splu, mkl_spsolve)
 
 
+_GMRES_CALLBACK_TYPE = {}
```
hodgestar, line 76: Nitpicking: Could we make `_GMRES_CALLBACK_TYPE = None` and then set `= "legacy"` if needed and still construct the `extra` dictionary in the function that uses it? It just feels a bit odd to have a mutable global variable with a slightly misleading name (i.e. it's not just the callback type -- it's a dict of extra options).

qutip/tests/test_device.py
```diff
@@ -70,7 +69,7 @@
 
 
 device_lists = [
-    pytest.param(DispersiveCavityQED, {"g":0.1}, id = "DispersiveCavityQED"),
+    pytest.param(DispersiveCavityQED, {"g": 0.02}, id="DispersiveCavityQED"),
```
hodgestar, line 72: What warning did `g = 0.1` trigger?

qutip/tests/test_superoper.py
```diff
@@ -218,14 +218,13 @@ class TestSuper_td:
...
 
-    def __init__(self):
-        N = 3
-        self.t1 = QobjEvo([qeye(N)*(1.+0.1j),[create(N)*(1.-0.1j),f]])
-        self.t2 = QobjEvo([destroy(N)*(1.-0.2j)])
-        self.t3 = QobjEvo([[destroy(N)*create(N)*(1.+0.2j),f]])
-        self.q1 = qeye(N)*(1.+0.3j)
-        self.q2 = destroy(N)*(1.-0.3j)
-        self.q3 = destroy(N)*create(N)*(1.+0.4j)
+    N = 3
```
hodgestar, line 221: It might be nice to initialize these for each test using `setup_method` (https://docs.pytest.org/en/stable/xunit_setup.html#method-and-function-level-setup-teardown) or similar, but having them here is no worse than things were before, although it does hide the problem a bit by avoiding the warning.

hodgestar (approved):
Left some small comments, but looks good.

qutip/tests/test_device.py
BoxiLi: It is a warning saying that `dispersive approximation may not be valid`, related to whether the physics model is accurate or not.

hodgestar:
> Not a huge amount to be done about that CodeClimate failure to be honest

These complexity checking tools were all written with business software application logic in mind and don't cater well to cases where a single conceptual unit can be quite complex. Perhaps when we move to GitHub actions we can run our own linting tools directly there and get rid of CodeClimate.

qutip/steadystate.py
jakelishman: I actually renamed it because I thought it was clearer what it was this way round - it wasn't something a user could pass in, so it actually only _did_ have one possible kwarg in it. When I first read the file I assumed that it could be influenced by a user.

Agree that the global state isn't ideal - I could instead redefine `gmres` at the top to wrap the scipy implementation and do the version check there?

qutip/tests/test_device.py
jakelishman: Yeah, Boxi's right. The calls were wrapped in a `warnings.catch_warnings()` block inside the test, but I think that when you turn warnings into errors, pytest hooks itself in a little too deeply for that to work.

qutip/tests/test_superoper.py
jakelishman: Yeah, I'll just change the `__init__` into a `setup_method`. That's clearly what was intended by the test anyway.

This was less of a warning and more of a testing failure - pytest refuses to collect tests that have class initialisers.

jakelishman:
Honestly, I've found very very few places on QuTiP where the code complexity measure is actually bad. In my mind, it's exactly correct to complain about the function I touched - it _is_ nearly impossible to read. It would be much easier to scan for funny business if it were turned more formally into a single-dispatch function.

jakelishman:
I'm a fan of the CodeClimate complexity tests, but I wouldn't be averse to removing its complaints about "TODO" comments - I'm not sure what we really gain from that in this sort of open-source development.  In a business where people are working full-time then I imagine it can be worth preventing a merge to finalise some details, but in our very decentralised development, really in some cases (especially in the next major release), we do need to merge in parts of code we know to be imperfect.

That said, we probably should make an effort to clear out some of the TODOs and FIXMEs that are in the codebase already.

Pushed fix-up changes since last review:
 - wrapped the steadystate SciPy linalg functions so they're simplified at the point of call, and unknown keywords in old SciPy versions are just silently "eaten" (this is the same behaviour that the file used to have, just formalised a little more).
 - added `TODO` in a comment in `qutip.cy.pyxbuilder`
 - changed the static class attributes in `test_superoper.py` into a `setup_method` function

qutip/steadystate.py
```diff
@@ -72,6 +71,36 @@
...
+# Similarly, SciPy < 1.1 does not recognise the `atol` keyword.
+#
+# Respective checks can be removed when SciPy version requirements are raised.
+
+if _parse_version(scipy.__version__) < _parse_version("1.1"):
+    gmres = _eat_kwargs(gmres, ['atol', 'callback_type'])
+    lgmres = _eat_kwargs(lgmres, ['atol'])
+    bicgstab = _eat_kwargs(bicgstab, ['atol'])
+elif _parse_version(scipy.__version__) < _parse_version("1.4"):
+    gmres = _eat_kwargs(gmres, ['callback_type'])
```
jakelishman, line 101: @hodgestar: new approach to management of deprecated kwargs in scipy that isn't quite as mutable-global-state-ish (though really I suppose it's still mutable state... We are writing Python, after all.).

hodgestar (approved):
New approach looks okay to me.

I think we should silence the CodeClimate warning if we can (even if it's just with `# noqa` or similar).

jakelishman:
Why though? It _is_ a badly written function and it should be changed. It's a true error, and I wouldn't be confident that there isn't an edge-case bug in it, because it's got so many branches I can't separate them easily in my head when reading it. There's loads of ways to write the same functionality where it's decomposed into smaller bits that are all much easier to check for correctness individually (`mul_dense`, `mul_scalar`, `mul_sparse`, etc) - way easier to verify the top-level dispatching logic into those bits, and then way easier to verify each individual component within them.

Also, as long as we have the "TODO" error filter turned on, it would be wrong to silence the error that comes from adding a new one, right? If we wanted to skip it, we should turn it off globally (which I'm not against for the TODO thing). If it's in, this way at least lets us use codeclimate properly to see when certain errors were introduced, or when issues in funny-looking functions might have been affected.

At any rate, the many-return function is completely removed in `dev.major` since we completely rewrote the backing types.

hodgestar:
@jakelishman Sorry -- I forgot we added the TODO. I was speaking from a bigger picture perspective -- having CodeClimate fail all the time means that we might not notice if more failures are added accidentally, which would reduce the value of running CodeClimate in CI.

jakelishman:
Oh yeah, I totally agree with that perspective - it's in part why I've gone on my little crusade over the past couple of weeks to remove all the warnings and errors from all the build processes.

What do you think a sensible strategy is for handling bug fixes that touches code that issues complaints?  CodeClimate will typically restate its complaint if you change a line that contributes to it (e.g. here I changed a "return" statement in a function that has too many).  I guess there's either "maintainers state whether any particular PR is acceptable with a failing warning"; "write a [skip ci] commit message" (or whatever it's called); or "any PR that touches complained-about code must fix the error".

Without any experience, to my mind it seems like points 1 and 2 are basically the same, but point 1 gives the power to override to maintainers, and point 2 to contributors - seems like version 2 works well in a business where "everyone" is trusted to modify the code, but in the distribued opensource world, maybe not?  Point 3 seems ideal in principle, but it seems like it would make it very hard to have first-time contributors do anything with older parts of the library - there'd be a huge onus to completely overhaul functionality, even just to fix small stuff.

I really don't know what recommended practices are here - I was roughly working under the assumption that we'd have maintainers be allowed to override complexity failures on occasion, but we've never really discussed it as far as I know, and I do recognise the problems with it.

jakelishman:
Also: I don't think CodeClimate actually supports skipping the tests when we call it as an action like this.  We'd have to set it up to run on CI runner under our control (e.g. on GitHub Actions), and then configure a method of skipping tests within that.

jakelishman:
I'll merge this now (it's going to have a couple of merge conflicts with other PRs currently in preparation), but I think there's a bit of discussion that's worth having about our processes for CodeClimate errors.

PR #1507, opened 2021-04-22
Immutable Coefficient

Ericgig:
Remove `Coefficient.argument` which modify the object inplace for `Coefficient.replace` which return a modified copy.
`Copy` of string coefficient is optimized.
Initially from #1481, more discussion there.

Some circular import in unrelated files have been moved inside function as they are causing problem in tests.

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -103,8 +103,10 @@ def coefficient(base, *, tlist=None, args={}, args_ctypes={},
         return coeff_from_str(base, args, args_ctypes, compile_opt)
 
     elif callable(base):
-        # TODO add tests?
-        return FunctionCoefficient(base, args.copy())
+        op = FunctionCoefficient(base, args.copy())
+        if not isinstance(op(0), numbers.Number):
+            raise TypeError("The coefficient function must return a number")
+        return op
     else:
         raise ValueError("coefficient format not understood")
 
@@ -216,7 +218,7 @@ class CompilationOptions:
 
 
 # Version number of the Coefficient
-COEFF_VERSION = "1.0"
+COEFF_VERSION = "1.1"
 
 
 def get_root():
@@ -358,22 +360,29 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
     """
     cdef_cte = ""
     init_cte = ""
+    copy_cte = ""
     for i, (name, val, ctype) in enumerate(constants):
         cdef_cte += "        {} {}\n".format(ctype, name[5:])
+        copy_cte += "        out.{} = {}\n".format(name[5:], name)
         init_cte += "        {} = cte[{}]\n".format(name, i)
     cdef_var = ""
     init_var = ""
+    init_arg = ""
     args_var = ""
     call_var = ""
+    copy_var = ""
     for i, (name, val, ctype) in enumerate(variables):
         cdef_var += "        str key{}\n".format(i)
         cdef_var += "        {} {}\n".format(ctype, name[5:])
+        copy_var += "        out.key{} = self.key{}\n".format(i, i)
+        copy_var += "        out.{} = {}\n".format(name[5:], name)
         if not raw:
             init_var += "        self.key{} = var[{}]\n".format(i, i)
         else:
             init_var += "        self.key{} = '{}'\n".format(i, val)
-        args_var += "        if self.key{} in args:\n".format(i)
-        args_var += "            {} = args[self.key{}]\n".format(name, i)
+        init_arg += "        {} = args[self.key{}]\n".format(name, i)
+        args_var += "            if self.key{} in arguments:\n".format(i)
+        args_var += "                out.{} = arguments[self.key{}]\n".format(name[5:], i)
         if raw:
             call_var += "        cdef {} {} = {}\n".format(ctype, val, name)
 
@@ -401,10 +410,21 @@ cdef class StrCoefficient(Coefficient):
 
     def __init__(self, base, var, cte, args):
         self.codeString = base
-{}{}        self.arguments(args)
+{}{}{}
 
-    cpdef void arguments(self, dict args) except *:
-{}        pass
+    cpdef Coefficient copy(self):
+        cdef StrCoefficient out = StrCoefficient.__new__(StrCoefficient)
+        out.codeString = self.codeString
+{}{}
+        return out
+
+    def replace(self, *, dict arguments=None, tlist=None):
+        cdef StrCoefficient out
+        if arguments:
+            out = self.copy()
+{}
+            return out
+        return self
 
     @cython.initializedcheck(False)
     @cython.cdivision(True)
@@ -413,8 +433,12 @@ cdef class StrCoefficient(Coefficient):
 
     def optstr(self):
         return self.codeString
-""".format(compile_opt['extra_import'], code, cdef_cte, cdef_var,
-           init_cte, init_var, args_var, call_var, code)
+""".format(compile_opt['extra_import'], code,
+           cdef_cte, cdef_var,
+           init_cte, init_var, init_arg,
+           copy_cte, copy_var,
+           args_var,
+           call_var, code)
     return code
 
 
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -129,18 +129,6 @@ def shift(coeff, _t0=0):
     return ShiftCoefficient(coeff, _t0)
 
 
-def reduce(coeff, args):
-    """ Reduce decorated string coefficient to 1 object:
-    c = coefficient("t")
-    optimize(c + conj(c)) => coefficient("t+conj(t)")
-    """
-    reduced = coeff.optstr()
-    if reduced:
-        return coefficient(reduced, args=args)
-    else:
-        return coeff
-
-
 # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 # %%%%%%%%%      Everything under this is for string compilation      %%%%%%%%%
 # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -430,9 +418,6 @@ cdef class StrCoefficient(Coefficient):
     @cython.cdivision(True)
     cdef complex _call(self, double t) except *:
 {call_var}        return {code}
-
-    def optstr(self):
-        return self.codeString
 """
     return code
 
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -386,7 +386,7 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
         if raw:
             call_var += "        cdef {} {} = {}\n".format(ctype, val, name)
 
-    code = """#cython: language_level=3
+    code = f"""#cython: language_level=3
 # This file is generated automatically by QuTiP.
 
 import numpy as np
@@ -398,47 +398,42 @@ from qutip.core.cy.math cimport erf, zerf
 from qutip.core.cy.complex_math cimport *
 from qutip.core.data cimport Data
 cdef double pi = 3.14159265358979323
-{}
+{compile_opt['extra_import']}
 
-parsed_code = "{}"
+parsed_code = "{code}"
 
 @cython.auto_pickle(True)
 cdef class StrCoefficient(Coefficient):
     cdef:
         str codeString
-{}{}
+{cdef_cte}{cdef_var}
 
     def __init__(self, base, var, cte, args):
         self.codeString = base
-{}{}{}
+{init_cte}{init_var}{init_arg}
 
     cpdef Coefficient copy(self):
         cdef StrCoefficient out = StrCoefficient.__new__(StrCoefficient)
         out.codeString = self.codeString
-{}{}
+{copy_cte}{copy_var}
         return out
 
     def replace(self, *, dict arguments=None, tlist=None):
         cdef StrCoefficient out
         if arguments:
             out = self.copy()
-{}
+{args_var}
             return out
         return self
 
     @cython.initializedcheck(False)
     @cython.cdivision(True)
     cdef complex _call(self, double t) except *:
-{}        return {}
+{call_var}        return {code}
 
     def optstr(self):
         return self.codeString
-""".format(compile_opt['extra_import'], code,
-           cdef_cte, cdef_var,
-           init_cte, init_var, init_arg,
-           copy_cte, copy_var,
-           args_var,
-           call_var, code)
+"""
     return code
 
 
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -356,7 +356,7 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
     cdef_var = ""
     init_var = ""
     init_arg = ""
-    args_var = ""
+    replace_var = ""
     call_var = ""
     copy_var = ""
     for i, (name, val, ctype) in enumerate(variables):
@@ -369,8 +369,9 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
         else:
             init_var += "        self.key{} = '{}'\n".format(i, val)
         init_arg += "        {} = args[self.key{}]\n".format(name, i)
-        args_var += "            if self.key{} in arguments:\n".format(i)
-        args_var += "                out.{} = arguments[self.key{}]\n".format(name[5:], i)
+        replace_var += "            if self.key{} in arguments:\n".format(i)
+        replace_var += ("                out.{}"
+                        " = arguments[self.key{}]\n".format(name[5:], i))
         if raw:
             call_var += "        cdef {} {} = {}\n".format(ctype, val, name)
 
@@ -392,6 +393,9 @@ parsed_code = "{code}"
 
 @cython.auto_pickle(True)
 cdef class StrCoefficient(Coefficient):
+    \"\"\"
+    String compiled as a :obj:`Coefficient` using cython.
+    \"\"\"
     cdef:
         str codeString
 {cdef_cte}{cdef_var}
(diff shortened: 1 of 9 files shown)


qutip/core/coefficient.py
```diff
@@ -401,10 +410,21 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
...
+{}{}
+        return out
+
+    def replace(self, *, dict arguments=None, tlist=None):
+        cdef StrCoefficient out
+        if arguments:
+            out = self.copy()
+{}
+            return out
+        return self
```
jakelishman, line 427: This string is pretty hard to read - it might be better if it's converted into an f-string with interpolation, rather than having ~12 arguments in a format call afterwards.

qutip/core/cy/coefficient.pyx
```diff
@@ -12,23 +12,23 @@ cimport cython
...
-        self.args.update(args)
+    def replace(self, *, arguments=None, tlist=None):
+        """Return the Coefficient with args or tlist changed. """
+        # Cases where tlist or args are supported are managed in those classes
+        return self
 
     def __call__(self, double t, dict args={}):
-        """update args and return the
-        """
+        """Update args and return the coefficient value at `t`. """
```
jakelishman, line 29: "Update" might be a bit misleading now, because it makes it seem like the coefficient is mutable.

qutip/core/cy/coefficient.pyx
```diff
@@ -164,6 +179,12 @@ cdef class InterpolateCoefficient(Coefficient):
     def __reduce__(self):
         return InterpolateCoefficient, (self.spline,)
 
+    cpdef Coefficient copy(self):
+        return InterpolateCoefficient(self.spline)
+
+    def replace(self, *, arguments=None, tlist=None):
+        return self
```
jakelishman, line 186: This seems somewhat misleading; shouldn't we make a new spline object if the `tlist` is updated?

jakelishman: Also, what's the difference between `InterpolateCoefficient` and `InterCoefficient`?  Could we have some class docstrings on all the `Coefficient` objects to make it clear too?

qutip/core/cy/coefficient.pyx
```diff
@@ -253,6 +283,15 @@ cdef class StepCoefficient(Coefficient):
...
 
+    cpdef Coefficient copy(self):
+        return StepCoefficient(self.coeff_np, self.tlist_np, self.cte)
+
+    def replace(self, *, arguments=None, tlist=None):
+        if tlist:
+            return InterCoefficient(self.coeff_np, tlist)
+        else:
+            return self.copy()
+
```
jakelishman, line 294: Is `InterCoefficient` a mistake here, or is there a reason for the replacement?

jakelishman (commented):
Similar to the `_BaseElement` stuff in the other PR, could we get some docstrings on all the different types of `Coefficient` to explain exactly what they're for and why?

qutip/core/cy/coefficient.pyx
Ericgig: Cannot be done easily, the original coefficient array is not stored in the Cubic_Spline object.
`Inter` and `Interpolate` are 2 implementations of cubic spline.
We should deprecate one of them with v5, now that we can.
`Cubic_Spline` is more memory efficient, `InterCoefficient` is faster.
`Cubic_Spline` support setting second derivative at boundaries, `InterCoefficient` support non-uniform `tlist`.

qutip/core/cy/coefficient.pyx
Ericgig: It is a mistake.

qutip/core/cy/coefficient.pyx
```diff
@@ -275,113 +536,218 @@ cdef class SumCoefficient(Coefficient):
...
+    Result of `Coefficient._cdc()` or `qutip.coefficent.norm(Coefficient)`.
+
+    Methods
+    -------
+    conj():
+        Conjugate of the `Coefficient`.
+    copy():
+        Create a copy of the `Coefficient`.
+    replace(arguments, tlist):
+        Create a new `Coefficient` with updated arguments and/or tlist.
```
jakelishman, line 673:
```suggestion
```

qutip/core/cy/coefficient.pyx
```diff
@@ -275,113 +536,218 @@ cdef class SumCoefficient(Coefficient):
...
+    Result of `Coefficient._shift()` or `qutip.coefficent.shift(Coefficient)`.
+
+    Methods
+    -------
+    conj():
+        Conjugate of the `Coefficient`.
+    copy():
+        Create a copy of the `Coefficient`.
+    replace(arguments, tlist):
+        Create a new `Coefficient` with updated arguments and/or tlist.
```
jakelishman, line 720:
```suggestion
```

qutip/core/qobj.py
```diff
@@ -625,6 +625,7 @@ def __call__(self, other):
         or by vectorization and devectorization, as
         appropriate.
         """
+        from qutip.core.superoperator import vector_to_operator, operator_to_vector
```
jakelishman, line 628: All changes in this file and a couple of others seem unrelated - might be better to skip them, since I think Shahnawaz wanted to work on decoupling the superoperator.py/superop_reps.py files from qobj.py a bit more at the logic level.

qutip/core/coefficient.py
```diff
@@ -127,18 +129,6 @@ def shift(coeff, _t0=0):
...
-    c = coefficient("t")
-    optimize(c + conj(c)) => coefficient("t+conj(t)")
-    """
-    reduced = coeff.optstr()
-    if reduced:
-        return coefficient(reduced, args=args)
-    else:
-        return coeff
-
-
```
jakelishman, line 141: It's probably good this is going - because of the scoping rules of `Coefficient`, these optimisations would have produced incorrect results if the `args` lists weren't guaranteed to be identical.  As it stands, you could have something like
```python
base = qutip.coefficient('A * exp(w * t)', args={'A': 0, 'w': 0})
series = sum(base.replace(args={'A': A, 'w': w}) for A, w in [(1, 2*np.pi), (2, 4*np.pi), ...])
```
and then
```python
reduce(series, <anything>)
```
wouldn't work correctly.  That type of coefficient reuse (though not the summation) is something I do quite heavily in my own research code, because it lets me have a single compilation at the start of the programme, then I use the same compilation unit for several different variables in the same Hamiltonian without recompiling anythin.  Gives me a big speed boost, and a really nice feature of the new coefficient system!

jakelishman (commented):
A few recurring points I just wanted to flag:
 - you need two backticks (` `` `` `) to make monospaced text in docstrings unless it's part of a Sphinx directive (`` :obj:`Qobj` ``)
 - a noun is "built from" rather than "build from" (don't you just love English?)
 - lists of methods aren't necessary in class docstrings, because Sphinx collects all the methods right underneath for us
 - I think `Coefficient.replace` might want to have the logical signature `(self, **kwargs)` instead of `(self, args=None, tlist=None)` (with suitable error handling), and each subclass's version is `(self, args=None)` or `(self, args=None, tlist=None)` or `(self, **kwargs)` as appropriate.  It might be a bit more maintainable in the face of any new coefficient types, and it means that we aren't in the slightly confusing position of documenting most subclasses with a "this argument is not used" note.

If you want to accept several commit suggestions, you can go to the "Files changed" tab, then you get the option to "add commit to batch", which lets you commit loads of things in one go, rather than one-by-one.  I don't know why the default view doesn't allow that.

qutip/core/cy/coefficient.pyx
```diff
@@ -164,27 +302,72 @@ cdef class InterpolateCoefficient(Coefficient):
...
+    """
     cdef int n_t, cte
     cdef double dt
     cdef double[::1] tlist
     cdef complex[::1] coeff_arr, second_derr
     cdef object tlist_np, coeff_np, second_np
 
-    def __init__(self, coeff_arr, tlist, second=None, cte=None):
+    def __init__(self, coeff_arr, tlist, _second=None, _cte=None):
```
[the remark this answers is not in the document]
Ericgig: They are use by `__reduce__` or computed from `coeff_arr`.

qutip/core/cy/coefficient.pyx
```diff
@@ -164,27 +302,72 @@ cdef class InterpolateCoefficient(Coefficient):
...
+        ----------
+        arguments : dict
+            Not used
+
+        tlist : np.array
+            New array of times for the array coefficients.
+        """
+        if tlist is not None:
+            return InterpolateCoefficient(
+                Cubic_Spline(tlist[0], tlist[1],
```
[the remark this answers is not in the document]
Ericgig: Supposed to be first and last.

hodgestar (changes requested):
Left some questions and comments, but they're all fairly minor. Looking close to being able to merge!

qutip/core/cy/coefficient.pyx
```diff
@@ -12,27 +13,40 @@ cimport cython
...
-    def __call__(self, double t, dict args={}):
-        """update args and return the
+        Parameters
+        ----------
+        **kwargs : dict
+            Values to change in the :obj:`Coefficient`
         """
-        if args:
-            self.arguments(args)
+        return self
```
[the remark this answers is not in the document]
jakelishman: I'd disagree on this - to me it's expected that `replace()` is a no-op, equivalent to `copy()`, and I'd be surprised if we threw an exception on it.  It's convenient for programmatic calling over a range of input `kwargs` - you could easily imagine it'd be useful in parametrised tests, for example.

qutip/core/coefficient.py
```diff
@@ -389,32 +386,39 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
...
-{}        pass
+{init_cte}{init_var}{init_arg}
+
+    cpdef Coefficient copy(self):
+        cdef StrCoefficient out = StrCoefficient.__new__(StrCoefficient)
+        out.codeString = self.codeString
+{copy_cte}{copy_var}
+        return out
+
+    def replace(self, *, dict arguments=None, tlist=None):
```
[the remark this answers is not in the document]
jakelishman: I asked a similar thing about the signature of `replace`, since quite a few coefficients take in `tlist` and ignore it.  I was wondering if it would be better to have the signature be `replace(**kwargs)` in all cases, and have each implementation ignore all kwargs it doesn't care about.  Then we could document only the needed kwargs in the docstring, but we'd still have the desirable behaviour that you could indiscriminately call `replace` on a higher-order object (say `QobjEvo`), and nothing would need to worry about what accepts what.  It's also potentially more extensible, because if we add a new coefficient type, it wouldn't be such a huge effort if it has additional parametrisation.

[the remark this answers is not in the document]
Ericgig: I made it so the signature always include `**kwargs` and the variable that can be directly replaced. So for function coefficient, it is `replace(arguments=None, **kwargs)`.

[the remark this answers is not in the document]
jakelishman: The docstrings should still say which keyword arguments are valid, though, so the user knows.  Otherwise there'd be no way to know what was valid.

qutip/core/cy/coefficient.pyx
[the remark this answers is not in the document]
hodgestar: I agree about the `.replace()` case (i.e. the one with no arguments) but having `.replace(tlist=[...])` silently do nothing seems odd. I think this is part of a wider conversation about what something like:
```python
tlist1 = np.logspace(-5, 0, 100)
tlist2 = np.logspace(-5, 0, 20)
c1 = coefficient(np.exp(-1j * tlist1, tlist=tlist1))
c2 = coefficient(np.exp(-1j * tlist2, tlist=tlist2))
c3 = c1 + c2
c3.replace(tlist=np.logspace(-5, 0, 50))
```
should do.

I'll resolve this item though and start a separate comment conversation about the bigger picture.

hodgestar:
Update from Google Meet discussion:

@jakelishman @Ericgig and I had a quick video call and decided that `.replace(...)` should be renamed something like `.replace_arguments(...)` and accept either a dictionary of arguments (as a positional only argument) or keyword arguments (with the new values for the arguments). In either case, the given argument keys should be update to their new values and any other keys should retain their previous values.

For the other attributes `.replace(...)` was supporting replace:

* the shift value can just be updated by wrapping the coefficient with `ShiftCoefficient` or updating the `t0` of an existing shift coefficient.
* the `tlist` value really doesn't make that much sense to update without also updating the associated function values and there are many ways this can go wrong (e.g. sub-expressions with different tlists, supplying a new tlist with the incorrect length).

qutip/core/cy/coefficient.pyx
```diff
@@ -275,113 +450,192 @@ cdef class SumCoefficient(Coefficient):
...
+            self.base.replace_arguments(_args, **kwargs)
+        )
+
 
 @cython.auto_pickle(True)
 cdef class NormCoefficient(Coefficient):
+    """
+    Norm of a :obj:`Coefficient`.
+    Used as a shortcut of conj(coeff) * coeff
+    Result of ``Coefficient._cdc()`` or ``qutip.coefficent.norm(Coefficient)``.
```
hodgestar, line 563:
```suggestion
    The L2 :func:`norm` of a :obj:`Coefficient`. A shortcut for ``conj(coeff) * coeff``.
    
    :obj:`NormCoefficient` is returned by ``qutip.coefficent.norm(Coefficient)``.
```

qutip/core/cy/coefficient.pyx
```diff
@@ -275,113 +450,192 @@ cdef class SumCoefficient(Coefficient):
...
 
-    cpdef void arguments(self, dict args) except *:
-        self._t0 = args["_t0"]
-        self.base.arguments(args)
+    def replace_arguments(self, _args=None, **kwargs):
+        """
+        Return a :obj:`Coefficient` with args changed for :obj:`Coefficient`
+        built from 'str' or a python function. Or a the :obj:`Coefficient`
+        itself if the :obj:`Coefficient` do not use arguments. New arguments
+        can be passed as a dict or as keywords.
```
hodgestar, line 617:
```suggestion
        Replace the arguments (``args``) of a coefficient.
        
        Returns a new :obj:`Coefficient` if the coefficient has arguments, or the original coefficient if it does not.
        Arguments to replace may be supplied either in a dictionary as the first position argument, or passed as
        keywords, or as a combination of the two. Arguments not replaced retain their previous values.
```

qutip/core/cy/coefficient.pyx
```diff
@@ -275,113 +450,192 @@ cdef class SumCoefficient(Coefficient):
...
         return NormCoefficient(self.base.copy())
 
 
 @cython.auto_pickle(True)
 cdef class ShiftCoefficient(Coefficient):
+    """
+    Introduce a time shift in the :obj:`Coefficient`.
+    Used intenally in correlation.
+    Result of ``Coefficient._shift()`` or
+    ``qutip.coefficent.shift(Coefficient)``.
```
hodgestar, line 603:
```suggestion
    Introduce a time shift into the :obj:`Coefficient`.
    
    Used internally within qutip when calculating correlations.
    
    :obj:ShiftCoefficient is returned by ``qutip.coefficent.shift(Coefficient)``.
```

hodgestar (approved):
I've left a slew of docstring suggestions, but the code looks good to me. Thank you for implementing this idea to make coefficients immutable and sorting out the replacement of arguments!

qutip/core/cy/coefficient.pyx
```diff
@@ -261,6 +432,10 @@ cdef class StepCoefficient(Coefficient):
 
 @cython.auto_pickle(True)
 cdef class SumCoefficient(Coefficient):
+    """
+    :obj:`Coefficient` built from the sum of 2 other Coefficients.
+    Result of :obj:`Coefficient` + :obj:`Coefficient`.
```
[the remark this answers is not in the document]
Ericgig: Sometime your replaced ```:obj:`Coefficient` ``` with coefficient, here you replaced `Coefficients` with ```:obj:`Coefficient`s```...
If not always using :obj: is acceptable then why not simply use `coefficients`. 
The example will result in an error, it should be valid code or clearly pseudo code.

qutip/core/cy/coefficient.pyx
```diff
@@ -135,14 +237,46 @@ def coeff(t, args):
...
 
     def __reduce__(self):
         return (StrFunctionCoefficient, (self.base, self.args))
 
+    def replace_arguments(self, _args=None, **kwargs):
+        """
+        Return a :obj:`Coefficient` with args changed for :obj:`Coefficient`
+        built from 'str' or a python function. Or a the :obj:`Coefficient`
+        itself if the :obj:`Coefficient` do not use arguments. New arguments
+        can be passed as a dict or as keywords.
```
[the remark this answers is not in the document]
Ericgig: `Returns a new :obj:`Coefficient` if the coefficient has arguments, or the original coefficient if it does not.` This kind of coefficient has arguments. Do we need this line?
We could merge the 2 first lines as:
```Returns a new :obj:`Coefficient` with replaced arguments (``args``).```

File: qutip/core/cy/coefficient.pyx (hunk 16 above)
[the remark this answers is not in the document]
Comment (hodgestar) [reply qutip/core/cy/coefficient.pyx]: I did use ``:obj:`Coefficients`s`` somewhere, but I can't see it here. I'm also a bit unsure how this ended up as `t * t` and `t` -- I don't remember typing that, but perhaps I typo'ed something without noticing. Either way, we should definitely make the code work -- my intention was to show an actual expression that would return a ``SumCoefficinet``.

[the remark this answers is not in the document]
Comment (Ericgig) [reply qutip/core/cy/coefficient.pyx]: I edited it before committing. `w1 * t` would fail since `w1` is not defined, but string with only t in it would work.

File: qutip/core/cy/coefficient.pyx [hunk 18]
```diff
@@ -12,30 +13,69 @@ cimport cython
...
+    composing time-dependant operator in list format for :obj:`QobjEvo`.
 
-    cpdef void arguments(self, dict args) except *:
-        self.args.update(args)
+    :obj:`Coefficient` are immutable.
+    """
+    def __init__(self):
+        raise NotImplementedError("Only sub-classes should be initiated.")
+
+    def replace_arguments(self, _args=None, **kwargs):
```
jakelishman, line 30: We can't change it until Python 3.8 is our lowest supported version, but an interesting note: in recent versions, we can actually do this pattern without even preventing the use of a variable name:
```python
>>> def f(x, /, **kwargs):
...     return x, kwargs
>>> f(1)
(1, {})
>>> f(x=1)
(None, {'x': 1})
>>> f(1, x=2):
(1, {'x': 2})
```
Of course it would be a logical error for a user to actually do that in this case, but still an interesting corner of the language!

jakelishman (approved):
🎉

qutip/core/cy/coefficient.pyx
jakelishman: (edited post to correct the signature...)

hodgestar:
\o/

PR #1509, opened 2021-04-24
Changes signature of rand_ket

### LINKED ISSUE #1504: Rand_ket does not work as intended
**Describe the bug**
qutip.random_objects.rand_ket does not work as intended
rand_ket(N=0, density=1, dims=None, seed=None) should be to figure out what N is, given dims is assigned when N is set to be 0 or None.

**To Reproduce**

```python
from qutip.random_objects import rand_ket
rand_ket(None, dims=[[2],[1]])
```
The terminal out put is
```
NameError                                 Traceback (most recent call last)
<ipython-input-61-18d662b8b1db> in <module>
----> 1 rand_ket(None, dims=[[2],[1]])

~\Anaconda3\envs\quspin_env\lib\site-packages\qutip\random_objects.py in rand_ket(N, density, dims, seed)
    326         _check_dims(dims, N, 1)
    327     elif dims:
--> 328         N = prod(dims[0])
    329         _check_dims(dims, N, 1)
    330     else:

NameError: name 'prod' is not defined
```

**Expected behavior**
rand_ket should give you the same answer as rand_ket(2, dims=[[2],[1]])

**Your Environment**
QuTiP Version:      4.6.0
Numpy Version:      1.19.2
Scipy Version:      1.6.2
Cython Version:     None
Matplotlib Version: 3.3.4
Python Version:     3.7.10
Number of CPUs:     8
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True

purva-thakre:
**Description**
Fixes #1504 and changes signature of `rand_ket` as [requested](https://github.com/qutip/qutip/issues/1504#issuecomment-823455134). 

Added tests to check behavior of `rand_ket` by checking shape of output when -  both N and dims are specified,  only dims is specified and finally only N is specified. 

**Changelog**
Fixes typo in random objects and changes signature of rand_ket

### DIFF
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -322,8 +322,7 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     Raises
     -------
     ValueError
-        Specify either the number of rows of operator (N) or
-        dimensions of quantum object (dims).
+        If neither `N` or `dims` are not specified.
 
     """
     if seed is not None:
@@ -373,8 +372,7 @@ def rand_ket_haar(N=None, dims=None, seed=None):
     Raises
     -------
     ValueError
-        Specify either the number of rows of operator (N) or
-        dimensions of quantum object (dims).
+        If neither `N` or `dims` are not specified.
     """
     if N is None and dims is None:
         raise ValueError('Specify either the number of rows of operator'
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -306,7 +306,7 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     Parameters
     ----------
     N : int
-        Number of rows for output quantum operator.
+        Number of rows for output quantum vector.
         If None or 0, N is deduced from dims.
     density : float
         Density between [0,1] of output ket state.
@@ -317,7 +317,7 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     Returns
     -------
     oper : qobj
-        Nx1 ket state quantum operator.
+        Nx1 ket quantum state vector.
 
     Raises
     -------
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -369,6 +369,12 @@ def rand_ket_haar(N=None, dims=None, seed=None):
     -------
     psi : Qobj
         A random state vector drawn from the Haar measure.
+
+    Raises
+    -------
+    ValueError
+        Specify either the number of rows of operator (N) or
+        dimensions of quantum object (dims).
     """
     if N is None and dims is None:
         raise ValueError('Specify either the number of rows of operator'
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -322,7 +322,7 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     Raises
     -------
     ValueError
-        If neither `N` or `dims` are not specified.
+        If neither `N` or `dims` are specified.
 
     """
     if seed is not None:
@@ -372,7 +372,7 @@ def rand_ket_haar(N=None, dims=None, seed=None):
     Raises
     -------
     ValueError
-        If neither `N` or `dims` are not specified.
+        If neither `N` or `dims` are specified.
     """
     if N is None and dims is None:
         raise ValueError('Specify either the number of rows of state vector'
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -330,7 +330,7 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
         np.random.seed(seed=seed)
     if N is None and dims is None:
         raise ValueError('Specify either the number of rows of operator'
-        '(N) or dimensions of quantum object (dims)')
+                         '(N) or dimensions of quantum object (dims)')
     elif N is not None and dims:
         _check_dims(dims, N, 1)
     elif dims:
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -325,7 +325,7 @@ def rand_ket(N=0, density=1, dims=None, seed=None):
     if N and dims:
         _check_dims(dims, N, 1)
     elif dims:
-        N = prod(dims[0])
+        N = np.prod(dims[0])
         _check_dims(dims, N, 1)
     else:
         dims = [[N],[1]]
@@ -364,7 +364,7 @@ def rand_ket_haar(N=2, dims=None, seed=None):
     if N and dims:
         _check_dims(dims, N, 1)
     elif dims:
-        N = prod(dims[0])
+        N = np.prod(dims[0])
         _check_dims(dims, N, 1)
     else:
         dims = [[N],[1]]
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -319,10 +319,19 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     oper : qobj
         Nx1 ket state quantum operator.
 
+    Raises
+    -------
+    ValueError
+        Specify either the number of rows of operator (N) or
+        dimensions of quantum object (dims).
+
     """
     if seed is not None:
         np.random.seed(seed=seed)
-    if N is not None and dims:
+    if N is None and dims is None:
+        raise ValueError('Specify either the number of rows of operator'
+        '(N) or dimensions of quantum object (dims)')
+    elif N is not None and dims:
         _check_dims(dims, N, 1)
     elif dims:
         N = np.prod(dims[0])
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -330,13 +330,13 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     if N is None and dims is None:
         raise ValueError('Specify either the number of rows of state vector'
                          '(N) or dimensions of quantum object (dims)')
-    elif N is not None and dims:
+    if N is not None and dims:
         _check_dims(dims, N, 1)
     elif dims:
         N = np.prod(dims[0])
         _check_dims(dims, N, 1)
     else:
-        dims = [[N],[1]]
+        dims = [[N], [1]]
     X = sp.rand(N, 1, density, format='csr')
     while X.nnz == 0:
         # ensure that the ket is not all zeros.
@@ -377,13 +377,13 @@ def rand_ket_haar(N=None, dims=None, seed=None):
     if N is None and dims is None:
         raise ValueError('Specify either the number of rows of state vector'
                          '(N) or dimensions of quantum object (dims)')
-    elif N and dims:
+    if N and dims:
         _check_dims(dims, N, 1)
     elif dims:
         N = np.prod(dims[0])
         _check_dims(dims, N, 1)
     else:
-        dims = [[N],[1]]
+        dims = [[N], [1]]
     psi = rand_unitary_haar(N, seed=seed) * basis(N, 0)
     psi.dims = dims
     return psi
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -322,13 +322,13 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     """
     if seed is not None:
         np.random.seed(seed=seed)
-    if N is not None and dims is not None:
+    if N is not None and dims:
         _check_dims(dims, N, 1)
     elif dims:
         N = np.prod(dims[0])
         _check_dims(dims, N, 1)
-    #else:
-        #dims = [[N],[1]]
+    else:
+        dims = [[N],[1]]
     X = sp.rand(N, 1, density, format='csr')
     while X.nnz == 0:
         # ensure that the ket is not all zeros.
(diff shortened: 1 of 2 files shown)


qutip/random_objects.py
```diff
@@ -322,10 +322,10 @@ def rand_ket(N=0, density=1, dims=None, seed=None):
...
         np.random.seed(seed=seed)
-    if N and dims:
+    if N is not None and dims:
         _check_dims(dims, N, 1)
     elif dims:
-        N = prod(dims[0])
+        N = np.prod(dims[0])
         _check_dims(dims, N, 1)
     else:
         dims = [[N],[1]]
```
hodgestar, line 331: If `dims` and `N` are both `None` this will set `dims` to `[[None], [1]]` which probably is not what we want. I suggest we raise a `ValueError` instead and add a test for this case.

hodgestar (changes requested):
Thanks for this PR! I've requested an addition to handle an error cases, but the rest looks good to me.

qutip/random_objects.py
purva-thakre: Thanks for pointing this out. I should have caught it before !

purva-thakre:
Made a change to `test_mesolve.py ` because a test was failing.

nathanshammah (approved):
LGTM

qutip/tests/test_random.py
```diff
@@ -91,5 +91,26 @@ def test_rand_super_dims():
...
+    test_both_N_dim = rand_ket(N=7,dims=[[7], [1]])
+    assert(np.shape(test_both_N_dim)==(7,1))
+
+    # when dims is specified and N has to be determined
+    test_only_dim = rand_ket(dims=[[7], [1]])
+    assert(np.shape(test_only_dim)==(7,1))
+
+    # when N is specified but dims has to be defined using default
+    test_only_N = rand_ket(N=7)
+    assert(np.shape(test_only_N)==(7,1))
```
jakelishman, line 111: I suggest you actually merge these cases into the test above, but a couple of additional points too:
1. can you add some cases that test `N` is calculated correctly including implicit tensor structure in `dims`, because that's what this is meant to be most useful for, right?
2. `np.shape` is for NumPy arrays, not `Qobj` (the fact that it works is kind of unintended side-effect).  `Qobj` already provides `Qobj.shape` to access this information directly.  
3. a very minor quibble, but `assert` isn't a function, so it doesn't need the brackets

qutip/tests/test_mesolve.py
```diff
@@ -893,7 +893,7 @@ def test_array_str_py_coeff(self):
             tlist=tlist, args={"_step_func_coeff": 1})
         result = mesolve(qu, rho0=rho0, tlist=tlist)
         assert(qu.type == "mixed_callable")
-        assert_allclose(
+        np.testing.assert_allclose(
```
jakelishman, line 896: It might be better to remove this from the PR - it's not related to anything else you're doing here (nor to the test failure).  It's ok if you have a seemingly random test failure - one of the admins can check it over, and (as in this case) re-run the test if necessary.

qutip/tests/test_random.py
```diff
@@ -91,5 +91,26 @@ def test_rand_super_dims():
         check_func_dims(func, (7, ), {}, [[[7], [7]]] * 2)
         check_func_dims(func, (6, ), {'dims': [[[2, 3], [2, 3]], [[2, 3], [2, 3]]]}, [[[2, 3], [2, 3]], [[2, 3], [2, 3]]])
 
+def test_rand_ket():
+    """
+    Random Qobjs: Tests that N and dims are determined as needed.
+    """
```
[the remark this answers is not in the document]
purva-thakre: Yeah, I saw that function but it was also testing `rand_ket_haar`, I thought it would be better to define a separate function for `rand_ket` only. Since you are requesting a change to `rand_ket_haar` as well, I will just move the extra cases there.

purva-thakre:
> can you add some cases that test N is calculated correctly including implicit tensor structure in dims, because that's what this is meant to be most useful for, right?

@jakelishman I have added 2 more functions to test N and tensor shape. I was a bit confused about how to check for tensor structure in dims. So, I check if the shape of output matches.

qutip/random_objects.py
```diff
@@ -319,13 +319,22 @@ def rand_ket(N=0, density=1, dims=None, seed=None):
     oper : qobj
         Nx1 ket state quantum operator.
 
+    Raises
+    -------
+    ValueError
+        Specify either the number of rows of operator (N) or
+        dimensions of quantum object (dims).
```
jakelishman, line 326: For the "raises" section of the documentation, the description should be the condition that the exception is raised, rather than a description of how to fix it (see [the numpydoc documentation](https://numpydoc.readthedocs.io/en/latest/format.html#raises)).  For example,
```
Raises
------
ValueError
    If neither ``N`` or ``dims`` are specified.
```

qutip/tests/test_random.py
```diff
@@ -71,13 +72,34 @@ def check_func_dims(func, args, kwargs, dims):
     # TODO: promote this out of test_random, as it's generically useful
     #       in writing tests.
     resdims = func(*args, **kwargs).dims
-    assert_(resdims == dims, "Checking {}; epected dimensions of {}, got {}.".format(func.__name__, dims, resdims))
+    assert_(resdims == dims, "Checking {}; expected dimensions of {}, got {}.".format(func.__name__, dims, resdims))
+
+def check_func_shape(func, args, kwargs, shape):
+    resdims = func(*args, **kwargs).dims
+    assert_(dims_to_tensor_shape(resdims)==shape,"Checking {}; expected shape of {}, got {}.".format(func.__name__, shape, dims_to_tensor_shape(resdims)))
```
jakelishman, line 79: I don't understand the purpose of this test?  `dims_to_tensor_shape` isn't related to what you're trying to do here; it's for resolving some reorderings that superoperators need.  You're already testing the dimensions in `check_func_dims`, and `dims_to_tensor_shape` in this case will always just return the dimensions as a "flattened" tuple.

Was there another reason for including it?

qutip/tests/test_random.py
```diff
@@ -71,13 +72,34 @@ def check_func_dims(func, args, kwargs, dims):
     # TODO: promote this out of test_random, as it's generically useful
     #       in writing tests.
     resdims = func(*args, **kwargs).dims
-    assert_(resdims == dims, "Checking {}; epected dimensions of {}, got {}.".format(func.__name__, dims, resdims))
+    assert_(resdims == dims, "Checking {}; expected dimensions of {}, got {}.".format(func.__name__, dims, resdims))
```
jakelishman, line 75: You wouldn't have known this necessarily, but now we use `pytest`, the better way to write this test is just to write
```python
assert resdims == dims
```
(Note that's a bare `assert` keyword, rather than numpy's slightly odd `assert_`.)  Pytest will show us the values of all the function parameters (`func`, `args`, `kwargs`) and both sides of the `==` (`resdims` and `dims`) if the test fails, which means we get the best possible error message without actually needing to write one at all.

qutip/random_objects.py
```diff
@@ -319,13 +319,22 @@ def rand_ket(N=0, density=1, dims=None, seed=None):
...
+
     """
     if seed is not None:
         np.random.seed(seed=seed)
-    if N and dims:
+    if N is None and dims is None:
+        raise ValueError('Specify either the number of rows of operator'
+                         '(N) or dimensions of quantum object (dims)')
```
jakelishman, line 333: I was going to question why this refers to an operator, but I just saw the docstring of this function has it wrong as well.  Could you change this exception (and the one in `rand_ket_haar`), and the docstring of this function to refer to a "state vector" instead of an "operator"?

qutip/tests/test_random.py
```diff
@@ -71,13 +72,34 @@ def check_func_dims(func, args, kwargs, dims):
...
+
+def check_func_shape(func, args, kwargs, shape):
+    resdims = func(*args, **kwargs).dims
+    assert_(dims_to_tensor_shape(resdims)==shape,"Checking {}; expected shape of {}, got {}.".format(func.__name__, shape, dims_to_tensor_shape(resdims)))
+
+def check_func_N(func, args, kwargs, N):
+    new_state_shape=func(*args, **kwargs).shape
+    assert_(new_state_shape[0]==N,"Checking {}; expected N of {}, got {}.".format(func.__name__, N, new_state_shape[0]))
```
jakelishman, line 83: If you're going to test the shape (a good idea!) it might be better to check the _entire_ shape, i.e. do
```python
assert new_state_shape == (N, 1)
```
since we know it exactly.

jakelishman (commented):
Thanks for making the change to `rand_ket_haar` as well.  The tests you've chosen look like a sensible mixture and seem to cover every case.  I had a question about one of the additional "checks" you added in to the tests - in the comments below.

qutip/tests/test_random.py
purva-thakre: For some reason, I remember thinking that I should test the tuple as well. Unfortunately, I forgot that exact reason ! Like you said, the tensor structure is actually tested in `check_func_dims`. I think that's why I was confused about ways to test tensor structure. I will make sure to remove this check.

qutip/tests/test_random.py
purva-thakre: Would you prefer I remove the error message `Checking {}; expected dimensions of {}, got {}.` ? 
Like you said, pytest will try to give best possible error message.

jakelishman: I'm happy if you don't want to bother right now (since you were just matching the style of the rest of the code), but that's up to you.

qutip/tests/test_random.py
```diff
@@ -58,26 +58,48 @@ def test_rand_dm_ginibre_rank():
...
     # TODO: promote this out of test_random, as it's generically useful
     #       in writing tests.
     resdims = func(*args, **kwargs).dims
-    assert_(resdims == dims, "Checking {}; epected dimensions of {}, got {}.".format(func.__name__, dims, resdims))
+    assert resdims == dims
+
+def check_func_N(func, args, kwargs,expected_shape):
+    # Here, expected_shape is supposed to be of form N X 1. When assigning
+    # this to a tuple, value 1 cannot be used.
+    number_of_rows , column_number = expected_shape
```
jakelishman, line 79: I'm not quite sure what you meant by the second sentence in the comment.  Also, what's the reason for `number_of_rows` and `column_number`?  They're not used at all (and it's kind of odd that you called one `number_of_<x>` and the other `<x>_ number`!).

purva-thakre: Maybe I ought to have chosen different variable names. The issue is that if I use `N,1 == expected_shape`, I end up with 
`SyntaxError : Can't assign to a literal`. I cannot use tuple `(N,1)` in function definition directly. So, I tried to use the method defined [here](https://stackoverflow.com/a/20035625/10241324). 

By default, `column_number == 1` whereas `number_of_rows` is determined by N or dims.

jakelishman: But I'm not sure why you want to assign anything at all?  What are you trying to achieve?

purva-thakre: It's to do with how `expected_shape` is input in `check_func_N`. 

What I have done is for `expected_shape` to be input as a tuple - (`number_of_rows`,`column_number`). Here, I can input `expected_shape==(N,1)` in `check_func_N` and then at the step where both shapes are checked `new_state_shape==expected_shape`, `(N,1)==(N,1)`. 

If I do not do this extra step then the step fails due to `(N,1)==((N,1),1)`. In addition, if I don't unpack the tuple using  (`number_of_rows`,`column_number`) then the function won't accept `expected_shape==(N,1)`. 

I could change the input for `expected_shape` to be just N instead of `(N,1)`. This could help avoid above step.

jakelishman: Huh?  I really don't understand what you're saying at all.  The unpacking has _no_ effect on what `expected_shape` is allowed to be, and no effect on the rest of the code.  This would work completely fine:
```python
def check_function_shape(func, args, kwargs, expected_shape):
    test_shape = func(*args, **kwargs).shape
    assert test_shape == expected_shape

check_function_shape(rand_ket, (6,), {}, (6, 1))
```

purva-thakre: Yeah, what you have does work. Unfortunately, I cannot replicate the issue that was causing a failure `(N,1)==((N,1),1)`. 
I should spend time to document all errors properly 😅

The new changes accepts `N` and checks the shape as `(N,1)`.

jakelishman: Ok, it's good that you understand what went wrong.  There's no need to have both lines of
```python
assert test_shape == (N, 1)
assert test_shape[0] == N
```
because the second line can _never_ fail if the first line passes (and `test_shape` is a tuple), but that's ok.

jakelishman:
I've just pushed a little change to correct some minor points of style in `random_objects.py` (that weren't your fault - they were there before), and to correct the docstring to say that only `N=None` triggers the deduction, not `N=0`.

(I also just finished the conversion of the tests into a parametrised pytest style - I'm trying to keep get these gradually updated as we look at the files, but I didn't want to ask you to do a _whole_ bunch more work, when you've already done a lot to get this ready, thanks!)

jakelishman (approved):

jakelishman:
Whoops, I meant to squash that into one commit rather than merge in all the history of the review.  No big deal, though.

PR #1513, opened 2021-04-29
Graft qutip/qutip-doc onto qutip/qutip

jakelishman:
This PR is perhaps not quite as insane as it looks.  It "simply" writes the entire history of the `qutip/qutip-doc` repository into `qutip/qutip` as if it has always been there within the subdirectory `doc`.  The commit objects of `qutip/qutip-doc` are all _technically_ completely rewritten (all the hashes change when you change the tree structure), but I kept all the metadata of them so they'll appear the same, and you can traverse the history just like you could before.

Aside from the complete rewriting of history to place it inside a `doc` subdirectory, `git` sees this PR as an actual merge commit, because I melded the two repositories together.  All commits will appear separately in `git log`, as you'd expect (although by default `git log --graph` might only show you one branch).  I rewrote history to make using `git log` and `git blame` a little less confusing; if I had just made a single commit renaming everything into the subdirectory, there might have been weird name conflicts and a rename to navigate when trying to look through the history.

This PR also logically changes a few links, so there are 3 commits tacked on the top to make the necessary changes within the `doc` folder and within the main QuTiP README.  These include moving the docs GitHub Actions workflow up into the root of the repository (and suitably modifying it), and updating several links within the documentation.  The documentation build action [has succeeded on my branch](https://github.com/jakelishman/qutip/runs/2469049269).

I have excluded the `docs` repository from the CodeClimate analysis, since it's not really part of the code we want tested.  I'm not absolutely certain that this is the correct decision (there's plenty of ancient code within `docs`), so I'm open to changing it.  As part of this exclusion, I also had to specify the exclusion of the `tests` and (autogenerated) `dist` directories.  This is not actually new behaviour; when no exclusions are set, [CodeClimate implicitly excludes these folders](https://docs.codeclimate.com/docs/excluding-files-and-folders#auto-generated-file-and-folder-exclusions).  I'm not intending to change behaviour (aside from the obvious) within this PR, so I codified those defaults into our file for now.  If we want to begin counting the tests in the complexity analysis (and I'm not wildly against that), we can remove the line in a different PR.

I suggest that reviewers use the "Commit filter" in the "Files changed" tab to show only the changes that come after merge commit b3fa49c (the one which actually grafts in `qutip/qutip-doc` - it's near the very bottom of the list), unless you enjoy looking through a binary tonne of commits.  Also, I'm laughing at the diff-stat line `+14,785 | -1`.  It's the `-1` that makes it art.


hodgestar (approved):
The last three commits look good to me.

jakelishman:
Oh, a thought: perhaps I should modify the commit messages of PRs in the old qutip-doc to make them link correctly. For example, instigate a rewrite rule of #xxx goes to qutip/qutip-doc#xxx?

hodgestar:
> Oh, a thought: perhaps I should modify the commit messages of PRs in the old qutip-doc to make them link correctly. For example, instigate a rewrite rule of #xxx goes to qutip/qutip-doc#xxx?

That sounds great, but I don't think it's required if it turns into a rabbithole.

Ericgig:
It would be nice to also change the link to the documentation in the pull request template.

jakelishman:
Ok, I've updated all PR links.  It was no harder than basically running `sed s_#_qutip/qutip-doc#_g` on the commit messages - I checked that nothing untoward got changed as well.

I've also updated the PR template - thanks Eric, I never would have thought of that.

jakelishman:
Shahnawaz - I think you had some thoughts about merging in the documentation?  It'd be good to know what you think about this strategy as well.  I haven't attempted to change any structure of the documentation in this PR; it's building reliably (evidenced by the GH Action), and we can make further overhauls no matter which repo it's based in.

quantshah (approved):

quantshah:
@jakelishman Thanks Jake for this effort. I had some opinion about redoing the IPython calls but that can be tackled in a different PR. Feel free to merge this.

jakelishman:
Yeah, sounds good.  I agree that it's better not to tack any behavioural changes onto this PR, because it would just confuse matters.  I'll hold off merging til Nathan's had a look through as well, since I can see that he's just started.

nathanshammah (commented):
Great! Just checking that this PR will keep the commits of the original authors in qutip-qip (as it seems).

Some contextual questions not really addressable in the PR:

- What about the issues open there on qutip-doc?
- What happens to qutip-doc, we archive it?

This is a major step for the project, reducing clumsiness and stepping functions to  updating the documentation at the time of code contributions.

jakelishman:
Yeah, this is a "true" merge of the qutip-doc repository with this one.  Technically all the commits have been rewritten (you'll find that the hashes won't match between the repositories), but they'll all be credited to the same authors on GitHub.  Everyone will actually get the credit "twice" - once for the archived qutip/qutip-doc repo, and once for the new commit object in here.  If anybody was in the habit of signing their commits with a private key, that will have been lost, but the only people that's likely to have affected is me and the GitHub merge bot.  It doesn't mean anything in the greater scheme of things, because I've signed the merge commit.  (signing is mostly just a way of verifying that the author really was who it says it was)

Answers:
- I'll migrate all open issues on qutip-doc into the qutip/qutip repository, and tag them with a "docs" tag
- We'll archive it, posting a message on the front page of it (or the README if GitHub doesn't have a separate mechanism).  We need to leave the repository up and available on the web, because old documentation refers to it in links.

PR #1546, opened 2021-05-06
Add release number to 'About this Documentation'

hodgestar:
**Description**
When I released QuTiP 4.6.1 I forgot to update the version number in the documentation configuration, `conf.py`. This PR updates `conf.py` to read the version and release from the `VERSION` file in an attempt to remove the possibility of such mistakes in future releases.

I've also deleted the `doc/sphinxext/requirements.txt` which appears to be no longer necessary.

**Changelog**
Read documentation version and release from the VERSION file.


doc/conf.py
```diff
@@ -68,11 +69,12 @@
...
 # built documents.
-#
+qutip_root = pathlib.Path(__file__).absolute().parent.parent
+qutip_release = qutip_root.joinpath("VERSION").read_text().strip()
 # The short X.Y version.
-version = '4.7'
+version = ".".join(qutip_release.split(".")[:2])
 # The full version, including alpha/beta/rc tags.
-release = '4.7.0.dev0'
+release = qutip_release
```
jakelishman, line 77: I always feel a little weird about walking the filesystem, though I'm not really sure why.  We have to have the `qutip` package available on the path to build the docs, so we could also parse this version out of `qutip.__version__` - this might be a little more robust in case we ever change the method we assign the version.

There's also `packaging.version.parse` that can be used to parse the version string, but I'm not sure we actually derive much benefit from that in this case.

doc/frontmatter.rst
```diff
@@ -45,6 +45,8 @@ This document contains a user guide and automatically generated API documentatio
 
 :version: |version|
 
+:release: |release|
+
```
jakelishman, line 49: Probably only need the release, right?  All the information's in that anyway.

doc/conf.py
hodgestar: The downside to `packaging.version.parse(qutip.__version__)` is that it reads the installed version of qutip, not the version from the source tree the documentation is being built from. So, e.g., right now on my local copy of this branch
```python
>>> packaging.version.parse(qutip.__version__)
<Version('4.6.1')>
```
even though
```console
$ cat VERSION 
4.7.0.dev
```
because I was impatient and just did `pip install qutip` rather than `pip install .`.

The other direction I've gone is to use something like bumpversion to just update all the versions manually with a single command (that avoids both difficulties, but adds a new dependency and a config file).

jakelishman: To some degree this is intentional, though - the reason being that the documentation build involves importing `qutip` and running a whole bunch of code from it.  The API documentation at the end is autogenerated by reading the importable version of QuTiP rather than what's in the same folder.

Perhaps we should loudly complain if `qutip.__version__` and `VERSION` don't evaluate to the same thing when they're parsed and canonicalised?

jakelishman: i.e. if you try to build the documentation with `pip install qutip`, your documentation will not actually be correct!

hodgestar: I thought about this a bit more. There are really 3 versions to worry about:

* `../VERSION` (the canonical version of the source we're building the documentation from)
* `../qutip/version.py` (the generated package version file in the source we're building from, currently this appears to be a bit bogus on master -- it's `5.0.0b1`)
* `qutip.__version__` (the version of the qutip package we've imported while building)

My latest views (subject to possible updates):

* Since we're building from a particular set of source files,  `../VERSION` is the canonical version. I.e. this is analogous to reading from `VERSION` in `setup.py`.
* We should raise a warning if `qutip.__version__` and the canonical version don't match. This should catch mismatches in HTML documentation builds in CI, but still allow developers to not worry too much locally.

Thoughts?

jakelishman: `qutip/version.py` shouldn't be in source control at all - it's autogenerated and forcefully overwritten any time `setup.py` executes - so let's just get rid of it.

The rest I all agree on.  I think raising a warning on mismatch is better for local development (as you did, it's nice to allow a little laziness when you're just trying to check separate things), but let's make sure that that warning becomes an error when it's running on CI.

We can't actually _directly_ compare the strings in `VERSION` and `qutip.__version__`; the latter has local versioning (`+<githash>`) applied while you're in a dev build (which only updates when you actually run `setup.py`, so it may not even be the correct hash), and the the former is not necessarily in canonical form (e.g. it's allowed to be, say, `4.6.0alpha2` instead of `4.6.0.a2`).  So to compare, we'll have to parse the version, canonicalise it and discard local versioning, but I think that's fine.

hodgestar: After some discussion in Discord, @jakelishman and I agreed that the right thing to do for the sanity check is really to warn if the documentation source and qutip package source are not from the same source folder.

PR #1553, opened 2021-05-12
Re-implement entropy_relative.

hodgestar:
**Description**
Re-implement entropy_relative.

This function was implemented in March 2012, but then removed a month later. After some digging, it appears that the reason the function was removed is that it assumed the eigenvectors of the density matrices were identical (and in the same order).

After some reading of Nielsen & Chuang, I have re-implemented it to address the issue.

**Still todo**
 - [x] Update the documentation.
 - [ ] Add an example to an example notebook.
 - [x] Add some tests for specific cases where we know what the answer should be.
 - [x] Stretch goal: Vectorize the calculation of P_ij and the rest of the loop as much as we can (but let's make sure the answers are correct first).

**Changelog**
Re-implement entropy_relative which returns the quantum relative entropy between two density matrices.

### DIFF
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -228,6 +228,7 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     """
     Calculates the relative entropy S(rho||sigma) between two density
     matrices.
+
     Parameters
     ----------
     rho : :class:`qutip.Qobj`
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -252,6 +252,8 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
     """
     if not rho.isoper or not sigma.isoper:
         raise TypeError("Inputs must be density matrices.")
+    if rho.shape != sigma.shape or rho.dims != sigma.dims:
+        raise ValueError("Inputs must have the same shape and dims.")
     if base == 2:
         log_base = log2
     elif base == e:
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -232,9 +232,11 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
     Parameters
     ----------
     rho : :class:`qutip.Qobj`
-        First density matrix.
+        First density matrix (or ket which will be converted to a density
+        matrix).
     sigma : :class:`qutip.Qobj`
-        Second density matrix.
+        Second density matrix (or ket which will be converted to a density
+        matrix).
     base : {e,2}
         Base of logarithm. Defaults to e.
     sparse : bool
@@ -253,6 +255,10 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
     Section 11.3.1, pg. 511 for a detailed explanation of quantum relative
     entropy.
     """
+    if rho.isket:
+        rho = ket2dm(rho)
+    if sigma.isket:
+        sigma = ket2dm(sigma)
     if not rho.isoper or not sigma.isoper:
         raise TypeError("Inputs must be density matrices.")
     if rho.dims != sigma.dims:
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -35,7 +35,7 @@ __all__ = ['entropy_vn', 'entropy_linear', 'entropy_mutual', 'negativity',
            'concurrence', 'entropy_conditional', 'entangling_power',
            'entropy_relative']
 
-from numpy import conj, e, inf, inner, real, sort, sqrt, vdot
+from numpy import conj, e, inf, imag, inner, real, sort, sqrt
 from numpy.lib.scimath import log, log2
 from qutip.qobj import ptrace
 from qutip.states import ket2dm
@@ -276,7 +276,11 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     # S is +inf if the kernel of sigma (i.e. svecs[svals == 0]) has non-trivial
     # intersection with the support of rho (i.e. rvecs[rvals != 0]).
     rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
+    if any(imag(rvals) >= tol):
+        raise ValueError("Input rho has non-real eigenvalues.")
     svals, svecs = sp_eigs(sigma.data, sigma.isherm, vecs=True, sparse=sparse)
+    if any(imag(svals) >= tol):
+        raise ValueError("Input sigma has non-real eigenvalues.")
     nzrvals = rvals[abs(rvals) >= tol]
     # Calculate S
     S = sum(nzrvals * log_base(nzrvals))
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -231,12 +231,15 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
 
     Parameters
     ----------
-    rho : qobj
+    rho : :class:`qutip.Qobj`
         First density matrix.
-    sigma : qobj
+    sigma : :class:`qutip.Qobj`
         Second density matrix.
     base : {e,2}
-        Base of logarithm.
+        Base of logarithm. Defaults to e.
+    sparse : bool
+        Flag to use sparse solver when determining the eigenvectors
+        of the density matrices. Defaults to False.
 
     Returns
     -------
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -283,18 +283,17 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     if any(imag(svals) >= tol):
         raise ValueError("Input sigma has non-real eigenvalues.")
     svals = real(svals)
+    # Calculate inner products of eigenvectors and return +inf if kernel
+    # of sigma overlaps with support of rho.
+    P = abs(inner(rvecs, conj(svecs))) ** 2
+    if (rvals >= tol) @ (P >= tol) @ (svals < tol):
+        return inf
+    # Avoid -inf from log(0) -- these terms will be multiplied by zero later
+    # anyway
+    svals[abs(svals) < tol] = 1
     nzrvals = rvals[abs(rvals) >= tol]
     # Calculate S
-    S = sum(nzrvals * log_base(nzrvals))
-    P = abs(inner(rvecs, conj(svecs))) ** 2
-    for i in range(len(rvals)):
-        for j in range(len(svals)):
-            if abs(svals[j]) < tol and not (
-                    abs(rvals[i]) < tol or abs(P[i, j]) < tol):
-                # kernel of sigma intersects support of rho
-                return inf
-            if svals[j] >= tol:
-                S -= rvals[i] * P[i, j] * log_base(svals[j])
+    S = nzrvals @ conj(log_base(nzrvals)) - rvals @ P @ log_base(svals)
     # the relative entropy is guaranteed to be >= 0, so we clamp the
     # calculated value to 0 to avoid small violations of the lower bound.
     return max(0, S)
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -224,11 +224,10 @@ def entropy_mutual(rho, selA, selB, base=e, sparse=False):
     return out
 
 
-def entropy_relative(rho, sigma, base=e, sparse=False):
+def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     """
     Calculates the relative entropy S(rho||sigma) between two density
     matrices.
-
     Parameters
     ----------
     rho : :class:`qutip.Qobj`
@@ -242,6 +241,9 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
     sparse : bool
         Flag to use sparse solver when determining the eigenvectors
         of the density matrices. Defaults to False.
+    tol : float
+        Tolerance to use to detect 0 eigenvalues or dot producted between
+        eigenvectors. Defaults to 1e-12.
 
     Returns
     -------
@@ -275,7 +277,7 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
     # intersection with the support of rho (i.e. rvecs[rvals != 0]).
     rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
     svals, svecs = sp_eigs(sigma.data, sigma.isherm, vecs=True, sparse=sparse)
-    nzrvals = rvals[rvals != 0]
+    nzrvals = rvals[abs(rvals) >= tol]
     # Calculate S
     S = sum(nzrvals * log_base(nzrvals))
     for i in range(len(rvals)):
@@ -284,7 +286,8 @@ def entropy_relative(rho, sigma, base=e, sparse=False):
                 dot(rvecs[i], svecs[j].conjugate()) *
                 dot(svecs[j], rvecs[i].conjugate())
             )
-            if svals[j] == 0 and (rvals[i] * P_ij != 0):
+            if abs(svals[j]) < tol and not (
+                    abs(rvals[i]) < tol or abs(P_ij) < tol):
                 # kernel of sigma intersects support of rho
                 return inf
             if svals[j] != 0:
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -249,7 +249,25 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     Returns
     -------
     rel_ent : float
-        Value of relative entropy.
+        Value of relative entropy. Guaranteed to be greater than zero
+        and should equal zero only when rho and sigma are identical.
+
+    Examples
+    --------
+
+    First we define two density matrices:
+
+    >>> rho = qutip.ket2dm(qutip.ket("00"))
+    >>> sigma = rho + qutip.ket2dm(qutip.ket("01"))
+    >>> sigma = sigma.unit()
+
+    Then we calculate their relative entropy using base 2 (i.e. ``log2``)
+    and base e (i.e. ``log``).
+
+    >>> qutip.entropy_relative(rho, sigma, base=2)
+    1.0
+    >>> qutip.entropy_relative(rho, sigma)
+    0.6931471805599453
 
     References
     ----------
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -277,11 +277,11 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     # S is +inf if the kernel of sigma (i.e. svecs[svals == 0]) has non-trivial
     # intersection with the support of rho (i.e. rvecs[rvals != 0]).
     rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
-    if any(imag(rvals) >= tol):
+    if any(abs(imag(rvals)) >= tol):
         raise ValueError("Input rho has non-real eigenvalues.")
     rvals = real(rvals)
     svals, svecs = sp_eigs(sigma.data, sigma.isherm, vecs=True, sparse=sparse)
-    if any(imag(svals) >= tol):
+    if any(abs(imag(svals)) >= tol):
         raise ValueError("Input sigma has non-real eigenvalues.")
     svals = real(svals)
     # Calculate inner products of eigenvectors and return +inf if kernel
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -294,7 +294,7 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     svals[abs(svals) < tol] = 1
     nzrvals = rvals[abs(rvals) >= tol]
     # Calculate S
-    S = nzrvals @ conj(log_base(nzrvals)) - rvals @ P @ log_base(svals)
+    S = nzrvals @ log_base(nzrvals) - rvals @ P @ log_base(svals)
     # the relative entropy is guaranteed to be >= 0, so we clamp the
     # calculated value to 0 to avoid small violations of the lower bound.
     return max(0, S)
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -292,7 +292,9 @@ def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
                 return inf
             if svals[j] != 0:
                 S -= rvals[i] * P_ij * log_base(svals[j])
-    return real(S)
+    # the relative entropy is guaranteed to be >= 0, so we clamp the
+    # calculated value to 0 to avoid small violations of the lower bound.
+    return max(0, real(S))
 
 
 def entropy_conditional(rho, selB, base=e, sparse=False):
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -35,7 +35,7 @@ __all__ = ['entropy_vn', 'entropy_linear', 'entropy_mutual', 'negativity',
            'concurrence', 'entropy_conditional', 'entangling_power',
            'entropy_relative']
 
-from numpy import dot, e, inf, real, sort, sqrt, vdot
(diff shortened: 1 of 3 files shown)


qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
         Value of relative entropy.
 
+    References
+    ----------
+
+    See Nielsen & Chuang, "Quantum Computation and Quantum Information",
+    Section 11.3.1, pg. 511 for a detailed explanation of quantum relative
+    entropy.
     """
     if rho.type != 'oper' or sigma.type != 'oper':
```
Ericgig, line 253: Not your code, but I would prefer the property `isoper` to the test here.

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
+    # intersection with the support of rho (i.e. rvecs[rvals != 0]).
+    rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
+    svals, svecs = sp_eigs(sigma.data, sigma.isherm, vecs=True, sparse=sparse)
+    nzrvals = rvals[rvals != 0]
+    # Calculate S
+    S = sum(nzrvals * log_base(nzrvals))
+    for i in range(len(rvals)):
+        for j in range(len(svals)):
+            P_ij = (
+                dot(rvecs[i], svecs[j].conjugate()) *
```
Ericgig, line 273: Matching `shape` and `dims` are never tested.
Until this `dot`, operators of different shapes are accepted, it should at least fail before the `sp_eigs`.
Operators with the same `shape` and but different `dims` will not raise any error.
This is not normal qutip behaviours, so if this is on purpose, please leave a comment explaining why.

qutip/entropy.py
hodgestar: Thanks for the suggestion. Done!

qutip/entropy.py
hodgestar: I added a check for shape and dims at the start. Let me know if that addresses the issue.

qutip/entropy.py
```diff
@@ -223,10 +224,8 @@ def entropy_mutual(rho, selA, selB, base=e, sparse=False):
     return out
 
 
-def _entropy_relative(rho, sigma, base=e, sparse=False):
+def entropy_relative(rho, sigma, base=e, sparse=False):
```
jakelishman, line 227: Sparse kwarg isn't documented - just needs mentioning that it's about the eigensolver method.  (If you're at it, you could also make the `qobj` types probably crossrefs too.)

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
-    rnzvals = rvals[rvals != 0]
-    # calculate tr(rho*log sigma)
-    rel_trace = float(real(sum(rnzvals * slogvals)))
-    return -entropy_vn(rho, base, sparse) - rel_trace
+    # S(rho || sigma) = sum_i(p_i log p_i) - sum_ij(p_i P_ij log q_i)
+    #
+    # S is +inf if the kernel of sigma (i.e. svecs[svals == 0]) has non-trivial
+    # intersection with the support of rho (i.e. rvecs[rvals != 0]).
+    rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
+    svals, svecs = sp_eigs(sigma.data, sigma.isherm, vecs=True, sparse=sparse)
```
jakelishman, line 266: Is the relative entropy sensibly defined if the "density matrices" are non-Hermitian?

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
+            P_ij = (
+                dot(rvecs[i], svecs[j].conjugate()) *
+                dot(svecs[j], rvecs[i].conjugate())
+            )
+            if svals[j] == 0 and (rvals[i] * P_ij != 0):
+                # kernel of sigma intersects support of rho
+                return inf
+            if svals[j] != 0:
+                S -= rvals[i] * P_ij * log_base(svals[j])
+    return real(S)
```
jakelishman, line 281: The call to `real(S)` looks unnecessary (assuming that it's only ever Hermitian matrices that can be input) - the eigenvalues should be floats not complex already, and `P_ij` should be real itself, though I suppose depending on how you find it, it might not be typed as such.

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
-    # calculate tr(rho*log sigma)
-    rel_trace = float(real(sum(rnzvals * slogvals)))
-    return -entropy_vn(rho, base, sparse) - rel_trace
+    # S(rho || sigma) = sum_i(p_i log p_i) - sum_ij(p_i P_ij log q_i)
+    #
+    # S is +inf if the kernel of sigma (i.e. svecs[svals == 0]) has non-trivial
+    # intersection with the support of rho (i.e. rvecs[rvals != 0]).
+    rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
+    svals, svecs = sp_eigs(sigma.data, sigma.isherm, vecs=True, sparse=sparse)
+    nzrvals = rvals[rvals != 0]
```
jakelishman, line 267: This is a tricky problem numerically - `eigh` isn't guaranteed to find 0 eigenvalues to be _exactly_ 0, so you may need to take a tolerance argument, and compare `rvals[np.abs(rvals) < tol]` to get the nonzeros.

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
+        for j in range(len(svals)):
+            P_ij = (
+                dot(rvecs[i], svecs[j].conjugate()) *
+                dot(svecs[j], rvecs[i].conjugate())
+            )
+            if svals[j] == 0 and (rvals[i] * P_ij != 0):
+                # kernel of sigma intersects support of rho
+                return inf
+            if svals[j] != 0:
+                S -= rvals[i] * P_ij * log_base(svals[j])
```
jakelishman, line 280: Similar comment about floating-point tolerances.  It's harder here too, because `rvals[i] * P_ij` presumably needs to compare to `tol**2`?

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
+    if not rho.isoper or not sigma.isoper:
+        raise TypeError("Inputs must be density matrices.")
     if base == 2:
-        slogvals = log2(snzvals)
+        log_base = log2
     elif base == e:
-        slogvals = log(snzvals)
+        log_base = log
     else:
         raise ValueError("Base must be 2 or e.")
```
jakelishman, line 260: Perhaps it would be easier to allow arbitrary log bases?  Just use `log` throughout the function, then at the end return `S / np.log(base)` - pretty sure the maths works out the same.  Not, of course, that anything other than base 2 or base e are useful, but the internal logic is easier that way as well.

qutip/entropy.py
jakelishman: Also, given that we're constructing this by subtracting from a start value, should we take special care to ensure that we never return a value less than 0?  I can imagine it could screw up plenty of people's data processing in loops if we let small negative values through, and with floating-point subtraction that's entirely possible.

qutip/tests/test_entropy.py
```diff
@@ -89,6 +89,56 @@ def test_component_selection(self):
         assert abs(qutip.entropy_mutual(dm, [0, 2], [1]) - expect) < 1e-13
 
 
+class TestRelativeEntropy:
+    def test_rho_or_sigma_not_oper(self):
+        rho = qutip.ket("00")
+        sigma = qutip.ket("01")
+        with pytest.raises(TypeError) as exc:
+            qutip.entropy_relative(qutip.ket2dm(rho), sigma)
+        assert str(exc.value) == "Inputs must be density matrices."
```
jakelishman, line 98: minor nit, and maybe it's just because I learned Python later than other stuff, but `exc` is logically of out-of-scope in the assertion.  I'm pretty sure `pytest.raises` takes a message argument anyway, so you can put the test in there.

qutip/tests/test_entropy.py
```diff
@@ -89,6 +89,56 @@ def test_component_selection(self):
...
+        sigma = rho + qutip.ket2dm(qutip.ket("01"))
+        sigma = sigma.unit()
+        assert (
+            qutip.entropy_relative(rho, sigma) == pytest.approx(0.69314718)
+        )
+        assert (
+            qutip.entropy_relative(rho, sigma, base=np.e)
+            == pytest.approx(0.69314718)
+        )
+        assert qutip.entropy_relative(rho, sigma, base=2) == pytest.approx(1)
```
jakelishman, line 126: Are there any other known cases we can test as well?

I'm not sure on best practices here - is testing the default argument of `base` something that we want to enforce via tests like this?

Also, minor nit, but perhaps the base-e comparisons could be to `np.log(2)` for clarity.

qutip/tests/test_entropy.py
```diff
@@ -89,6 +89,56 @@ def test_component_selection(self):
...
+        rho = qutip.rand_dm(8, pure=False)
+        rel = qutip.entropy_relative(rho, rho)
+        assert abs(rel) < 1e-13
+
+    @pytest.mark.repeat(20)
+    def test_random_rho_sigma(self):
+        rho = qutip.rand_dm(8, pure=False)
+        sigma = qutip.rand_dm(8, pure=False)
+        rel = qutip.entropy_relative(rho, sigma)
+        assert rel >= 0
```
jakelishman, line 139: (see above comments about enforcing >= 0 output too.)  In my understanding of floating-point, I don't think the function as written _absolutely_ passes this test for all inputs, though failures would be exceedingly rare.

jakelishman (commented):
Thanks for both finding this and fixing it!  I have a suspicion there was an old issue or mailing list message that was asking for some quantity derived from relative entropy as well, so this is definitely good to get working.

qutip/entropy.py
Albantakis: I just checked the relative entropy function and ran into this issue. The relative entropy between |0> and |+> should be +inf. However, I got a value of ~53.25 because the svals of |+> returned [-8.71576399e-33  1.00000000e+00] instead of [0 1]. 
Formally the function seems correct to me (implementing eq. 11.53 in Nielsen & Chuang). Thank you very much for implementing it!

qutip/tests/test_entropy.py
Albantakis: Possible test cases: 
for pure states, either rho = sigma and S = 0, or rho ~= sigma and S = inf.
comparing pure states against the maximally mixed state should yield the classical rel entropy, e.g. |00> against 0.25(|00> + |01> + |10> + |11>) should return S = 2.
(I checked a few combinations and apart from the tolerance issue above for |0> and |+> I did get the correct results)

qutip/entropy.py
```diff
@@ -244,24 +243,42 @@ def _entropy_relative(rho, sigma, base=e, sparse=False):
...
+    Section 11.3.1, pg. 511 for a detailed explanation of quantum relative
+    entropy.
     """
-    if rho.type != 'oper' or sigma.type != 'oper':
-        raise TypeError("Inputs must be density matrices..")
-    # sigma terms
-    svals = sp_eigs(sigma.data, sigma.isherm, vecs=False, sparse=sparse)
-    snzvals = svals[svals != 0]
+    if not rho.isoper or not sigma.isoper:
+        raise TypeError("Inputs must be density matrices.")
```
jakelishman, line 254: Oh yeah, one more point - I think elsewhere in QuTiP we promote kets to dms in functions like this.  The relative entropy is defined between states, and a ket has an unambiguous representation as a dm without losing information, so we can probably implement that.

qutip/entropy.py
hodgestar: Ah, the push failed because my DNS fell over. Fixed. Shape check removed.

Re `P_ij` -- yep. I'll look at improving that a little once I've addressed the other comments.

qutip/entropy.py
hodgestar: Documentation for sparse and links to Qobj added.

qutip/entropy.py
hodgestar: Good idea. Implemented.

qutip/entropy.py
hodgestar: I'll add a tolerance argument. This was on my mind to add too.

qutip/entropy.py
hodgestar: `S` is a complex number (with zero imaginary part), `real(S)` is a float. Since the relative entropy is always a positive real number, I think `real(S)` makes sense, and agree that clamping to zero makes sense. Will add.

jakelishman: I didn't say it very well originally, but my point was that there's no reason `S` _should_ be a complex number at any point in the calculation, right?  The call to `real` made it look like something was wrong to me while I flicked through it.

If the inputs are allowed to be non-Hermitian, then S would in general have a non-zero imaginary component, so the call to `real` doesn't make sense, and if they _are_, then `rvals` and `svals` are both of type `double`, and `P_ij` should be of type `double` in all cases anyway, so S should be too, right?

qutip/tests/test_entropy.py
hodgestar: Enforcing greater than or equal to zero added, so this should now always work (but isn't a super strong test).

jakelishman: Probably still worth keeping in, I guess?  The test would still catch regressions if in the future we change the implementation of the function.

qutip/entropy.py
hodgestar: Good question. I *think* it will still make sense for PT-symmetric systems where the density matrix eigenvalues are real, positive or zero and sum to one. I don't know how or whether to check that.

qutip/entropy.py
hodgestar: Yes, all the eigenvalues and `P_ij` should be real already. I could convert / raise an error earlier?

qutip/entropy.py
hodgestar: Added.

qutip/tests/test_entropy.py
hodgestar: `pytest.raises` only takes a regular expression pattern via the `match` argument, which results in the equivalent code being something like:
```with pytest.raises(TypeError, match=re.escape("Inputs must be density matrices."))
```
which is not particularly pretty.

I tend to use this form since it easily allows checking other attributes of the exception (not that there are any here).

qutip/entropy.py
jakelishman: I think the intention is probably clearer if we do the real-ness tests when the eigenvalues are generated?  That's also approximately how you'd test for a valid PT matrix, I think, too - if the eigenvalues are all within tolerance of a real value then it's valid to try and continue, and we can cast them down to real then.

I don't know how much effort you want to put into non-Hermitian dynamics right now, though - I don't think very much of the library supports it.  If that's the case, then the function could also just fail out if `not rho.isherm or not sigma.isherm`.

qutip/tests/test_entropy.py
hodgestar: @Albantakis Thanks! I added a test for the pure vs maximally mixed density matrix.

qutip/tests/test_entropy.py
jakelishman: Fair enough, then.  The scoping thing is probably me just being a bit weird - looking at the pytest documentation, they actually have this kind of thing as a direct example.

qutip/tests/test_entropy.py
hodgestar: @jakelishman It's nice to test that the defaults work as expected, and its easy to update the test if it changes (and if it changes accidentally it's probably a bug).

qutip/entropy.py
hodgestar: I addressed this by checking `P_ij` and `rvals[i]` separately.

qutip/entropy.py
hodgestar: It's not a bad idea, but all of the other entropy functions follow this `e` or `2` pattern so I'd prefer to keep things consistent for now and consider changing that later.

jakelishman: Sounds sensible to me

jakelishman:
Oh, I just noticed the comment at the top about vectorisation of the calculation of `P`.  I could be wrong, but I think `P = np.abs(np.inner(rvecs, np.conj(svecs)))**2` gets you there.

hodgestar:
The intermittently failing test seems unrelated and is `test_usage_in_solvers[mesolve-complex-spline,string]`.

qutip/entropy.py
hodgestar: I've added the checks for non-real eigenvalues and converted the eigenvalues to floats immediately after the checks.

qutip/entropy.py
hodgestar: I added a check that the eigenvalues are real.

hodgestar:
@jakelishman @Ericgig @Albantakis Ready for a second round of review, I think!

qutip/entropy.py
```diff
@@ -223,45 +224,79 @@ def entropy_mutual(rho, selA, selB, base=e, sparse=False):
...
-def _entropy_relative(rho, sigma, base=e, sparse=False):
+def entropy_relative(rho, sigma, base=e, sparse=False, tol=1e-12):
     """
-    ****NEEDS TO BE WORKED ON****
-
     Calculates the relative entropy S(rho||sigma) between two density
     matrices.
-
     Parameters
     ----------
```
jakelishman: There's a missing blank line for the section heading here.

qutip/entropy.py
```diff
@@ -223,45 +224,79 @@ def entropy_mutual(rho, selA, selB, base=e, sparse=False):
...
-    # calculate tr(rho*log sigma)
-    rel_trace = float(real(sum(rnzvals * slogvals)))
-    return -entropy_vn(rho, base, sparse) - rel_trace
+    # S(rho || sigma) = sum_i(p_i log p_i) - sum_ij(p_i P_ij log q_i)
+    #
+    # S is +inf if the kernel of sigma (i.e. svecs[svals == 0]) has non-trivial
+    # intersection with the support of rho (i.e. rvecs[rvals != 0]).
+    rvals, rvecs = sp_eigs(rho.data, rho.isherm, vecs=True, sparse=sparse)
+    if any(imag(rvals) >= tol):
+        raise ValueError("Input rho has non-real eigenvalues.")
```
jakelishman, line 281: What if there are negative imaginary components?

(Also, incredibly minor nit, but one of my pet peeves - the pre-existing style of this file [import everything into a global namespace] masks the fact that the `any` being called is the Python iterator built-in, not the vectorised `np.any`.)

qutip/entropy.py
```diff
@@ -223,45 +224,79 @@ def entropy_mutual(rho, selA, selB, base=e, sparse=False):
...
+    svals = real(svals)
+    # Calculate inner products of eigenvectors and return +inf if kernel
+    # of sigma overlaps with support of rho.
+    P = abs(inner(rvecs, conj(svecs))) ** 2
+    if (rvals >= tol) @ (P >= tol) @ (svals < tol):
+        return inf
+    # Avoid -inf from log(0) -- these terms will be multiplied by zero later
+    # anyway
+    svals[abs(svals) < tol] = 1
+    nzrvals = rvals[abs(rvals) >= tol]
```
jakelishman, line 295: We can't use it without swapping how we handle the logarithms, but if it's useful in the future, scipy has `scipy.special.xlogy` that specifically handles `0 ln(0) := 0`.

qutip/entropy.py
```diff
@@ -223,45 +224,79 @@ def entropy_mutual(rho, selA, selB, base=e, sparse=False):
...
+    # Calculate S
+    S = nzrvals @ conj(log_base(nzrvals)) - rvals @ P @ log_base(svals)
```
jakelishman, line 297: Why conjugate `log(nzrvals)`?  It's guaranteed real, right?

qutip/tests/test_entropy.py
```diff
@@ -89,6 +90,113 @@ def test_component_selection(self):
...
+        with pytest.raises(TypeError) as exc:
+            qutip.entropy_relative(rho, sigma)
+        assert str(exc.value) == "Inputs must be density matrices."
+
+    def test_rho_and_sigma_have_different_shape_and_dims(self):
+        rho = qutip.ket("00")
+        sigma = qutip.ket("0")
+        with pytest.raises(ValueError) as exc:
+            qutip.entropy_relative(rho, sigma)
+        assert str(exc.value) == "Inputs must have the same shape and dims."
```
jakelishman, line 139: A good different dims test would be Qobjs that have compatible shapes, but incompatible dims (e.g. one in a [2, 3] space and one in a [3, 2] space) - Numpy/Scipy functions will generally just fail if shapes mismatch, but this test ensures that we're enforcing the Hilbert space constraints as well.

hodgestar:
I'm didn't find an obvious notebook in qutip-notebooks to add an example to. Not sure how we usually proceed in such cases -- skip having an example in a notebook? Make a new notebook?

jakelishman (commented):
Looking good.  For notebooks, we've historically not been good at updating them, I think.  For a single, clearly defined function, probably just adding an `Examples` section in the docstring might be more appropriate?  I'm not entirely sure.

qutip/entropy.py
jakelishman: (Although `np.any` doesn't short-circuit.)

qutip/entropy.py
hodgestar: The robots caught that too -- woot.

qutip/entropy.py
hodgestar: Good catch. Fixed.

Re `any` vs `np.any` -- yeah, the style of this file has been annoying me too. I have been making myself avoid fixing that in this PR. :)

qutip/entropy.py
hodgestar: Bleh. Just me being silly when translating maths from my head to code. Fixed.

qutip/entropy.py
jakelishman: Oh, that's great to see that they're working like that.  Let me see if I can force coloured output of Sphinx, though - might make it a bit easier to pick out the failures.

qutip/entropy.py
jakelishman: Haha, not fixing stuff that's "not my style" is definitely something I have to force myself to do as well!

qutip/tests/test_entropy.py
hodgestar: Given that we only check the dims, I'm not too worried.

jakelishman: Oh yeah, I'm not worried about the function as it is right now either.  I just know that matching shapes but mismatched dims is a bug we that crops up from time to time all over the place in QuTiP.  You can still make the test case in shorthand -
```python
rho = qutip.basis([2, 3], [0, 0])
sigma = qutip.basis([3, 2], [0, 0])
```

hodgestar: Added!

qutip/entropy.py
hodgestar: I'm okay with not involving `scipy.special.xlogy` for now -- it's another thing for readers of the code to think about and I'm not sure where the bottlenecks in the current code are or how much anyone cares about performance (my guess is that finding the eigenvalues is the slowest step now).

jakelishman: Oh yeah, no need to put it in now. Now that everything's vectorised, I'd be surprised if the dominant time factor wasn't the eigenstate calculation.

jakelishman (approved):

hodgestar:
@jakelishman I added an example to the docstring.

jakelishman:
Nice, and it's all rendered correctly in the HTML documentation

jakelishman (approved):

PR #1555, opened 2021-05-13
Convert http to https links where valid

jakelishman:
Not quite all links we refer to have valid SSL certificates, so we leave those ones be.

**Changelog**
Convert http links to https


nathanshammah (approved):
Thanks @jakelishman for this. A major step for the project visibility. One small thing: I guess indexing of previous stuff on Google should be redirected fine.

jakelishman:
Yeah, Google will just update itself over time, I think - the http links aren't dead, they just redirect through to the https ones.  There's a similar PR on the website (qutip/qutip.github.io#145) to fix links there as well.

Issue #1502, opened 2021-04-19
Bloch sphere requires matplotlib >= 3.3


jakelishman:
#1496 updated the Bloch sphere to prevent distortion in matplotlib 3.3+.  This appears to have inadvertantly pinned the effective matplotlib dependency to `matplotlib>=3.3`, because it uses `Axes3D.set_box_aspect` which didn't exist before then.  We might need to add a small version-dependent switch beacuse matplotlib 3.3 is under a year old, so it's a very recent version to be depending on.

I would like to keep the optional matplotlib dependency to at least `matplotlib>=3` if at all possible, since I don't think we're using much cutting-edge stuff.

Note we now have `packaging` as a run-time dependency, which will make any version comparisons much easier (e.g. 
```python
from packaging.version import parse as parse_version

if parse_version(matplotlib.__version__) > parse_version('3.3'):
    ...
else:
    ...
```

@BoxiLi: do you have thoughts?

BoxiLi:
Sounds good! @jakelishman  Thanks for pointing this out.

Minor comment: It should be `parse_version(matplotlib.__version__) >= parse_version('3.3'):` to also include 3.3.0?

jakelishman:
Ok cool, I'll leave it for other contributors for the time being. You're probably right about the version comparison - I hadn't really intended it to be accurate, just illustrative of what's possible.

MrRobot2211:
Hi, just to clarify: the intention here is to have the two behaviours mediated by  a `settings` flag  or just an in-place version check?

BoxiLi:
I think an in-place version check should be sufficient? But I remember you guys had a deeper discussion over the Scipy version check so you might have a different optinion.

PR #1556, opened 2021-05-13
Version checks for `matplotlib` [unitaryHACK]

### LINKED ISSUE #1503: matplotlib 3.4 issues new deprecation warnings
matplotlib 3.4 has begun issuing some new deprecation warnings against some of the stuff we do in `bloch.py`.

Simple reproducer from within `ipython --matplotlib`:
```python
In [1]: import qutip
   ...: b = qutip.Bloch()
   ...: b.add_vectors([0, 1, 0])
   ...: b.render()
```
produces
```
/Users/jake/.anaconda3/envs/qutip/lib/python3.9/site-packages/qutip/bloch.py:451: MatplotlibDeprecationWarning: Axes3D(fig) adding itself to the figure is deprecated since 3.4. Pass the keyword argument auto_add_to_figure=False and use fig.add_axes(ax) to suppress this warning. The default value of auto_add_to_figure will change to False in mpl3.5 and True values will no longer work in 3.6.  This is consistent with other Axes classes.
  self.axes = Axes3D(self.fig, azim=self.view[0], elev=self.view[1])

/Users/jake/.anaconda3/envs/qutip/lib/python3.9/site-packages/qutip/bloch.py:59: MatplotlibDeprecationWarning:
The M attribute was deprecated in Matplotlib 3.4 and will be removed two minor releases later. Use self.axes.M instead.
  xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, renderer.M)
```

This particular environment had `matplotlib==3.4.1`.

The warnings themselves suggest fixes, however some care needs to be taken; I suspect that the additional keyword `auto_add_to_figure` might be new in matplotlib 3.4 (not sure about the `.M` attribute).  We want to keep compatibility with at least `matplotlib>=3` (we don't want to picky so it's easy to use QuTiP with other packages), so it might be necessary to include a version test using `packaging.version.parse`.  If someone could test whether that's necessary, that'd be great.

Yash-10:
**Description**
This PR attempts to add version checks for resolving deprecation warnings for `matplotlib` in the `bloch` and `visualization` modules. Specifically, it attempts to resolve warnings related to `matplotlib>=3.4` and `matplotlib>=3.3`.

Places where the change has been made:
- `proj3d.proj_transform`
- `Axes3D`

Thanks!

**Related issues or PRs**
Fix #1503 and #1502.

**Changelog**
Added a version checking condition to handle specific functionalities depending on the `matplotlib` version.

### DIFF
--- a/doc/guide/guide-bloch.rst
+++ b/doc/guide/guide-bloch.rst
@@ -422,13 +422,13 @@ Directly Generating an Animation
 The code to directly generate an mp4 movie of the Qubit decay is as follows ::
 
    import matplotlib
-   from packaging.version import parse as parse_version  # For checking matplotlib version
+   from qutip.bloch import matplotlib_version_gte  # For checking matplotlib version
    from matplotlib import pyplot, animation
    from mpl_toolkits.mplot3d import Axes3D
 
    fig = pyplot.figure()
 
-   if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+   if matplotlib_version_gte():
       ax = Axes3D(fig,azim=-40,elev=30, auto_add_to_figure=False)
       fig.add_axes(ax)
    else:
--- a/doc/guide/guide-bloch.rst
+++ b/doc/guide/guide-bloch.rst
@@ -421,18 +421,11 @@ Directly Generating an Animation
 
 The code to directly generate an mp4 movie of the Qubit decay is as follows ::
 
-   import matplotlib
-   from qutip.bloch import matplotlib_version_gte  # For checking matplotlib version
    from matplotlib import pyplot, animation
    from mpl_toolkits.mplot3d import Axes3D
 
    fig = pyplot.figure()
-
-   if matplotlib_version_gte():
-      ax = Axes3D(fig,azim=-40,elev=30, auto_add_to_figure=False)
-      fig.add_axes(ax)
-   else:
-      ax = Axes3D(fig,azim=-40,elev=30)
+   ax = Axes3D(fig, azim=-40, elev=30)
    sphere = qutip.Bloch(axes=ax)
 
    def animate(i):
--- a/doc/guide/guide-bloch.rst
+++ b/doc/guide/guide-bloch.rst
@@ -421,11 +421,18 @@ Directly Generating an Animation
 
 The code to directly generate an mp4 movie of the Qubit decay is as follows ::
 
+   import matplotlib
+   from packaging.version import parse as parse_version  # For checking matplotlib version
    from matplotlib import pyplot, animation
    from mpl_toolkits.mplot3d import Axes3D
 
    fig = pyplot.figure()
-   ax = Axes3D(fig,azim=-40,elev=30)
+
+   if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+      ax = Axes3D(fig,azim=-40,elev=30, auto_add_to_figure=False)
+      fig.add_axes(ax)
+   else:
+      ax = Axes3D(fig,azim=-40,elev=30)
    sphere = qutip.Bloch(axes=ax)
 
    def animate(i):
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -35,8 +35,6 @@ __all__ = ['Bloch']
 
 import os
 
-import matplotlib
-
 from numpy import (ndarray, array, linspace, pi, outer, cos, sin, ones, size,
                    sqrt, real, mod, append, ceil, arange)
 
@@ -47,10 +45,20 @@ from qutip.expect import expect
 from qutip.operators import sigmax, sigmay, sigmaz
 
 try:
+    import matplotlib
     import matplotlib.pyplot as plt
     from mpl_toolkits.mplot3d import Axes3D
     from matplotlib.patches import FancyArrowPatch
     from mpl_toolkits.mplot3d import proj3d
+
+    def matplotlib_version_gte(version='3.4'):
+        """
+        Checks if matplotlib version is greater than a specific version
+        """
+        if parse_version(matplotlib.__version__) >= parse_version(version):
+            return True
+        else:
+            return False
 
     class Arrow3D(FancyArrowPatch):
         def __init__(self, xs, ys, zs, *args, **kwargs):
@@ -60,7 +68,7 @@ try:
 
         def draw(self, renderer):
             xs3d, ys3d, zs3d = self._verts3d
-            if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+            if matplotlib_version_gte():
                 xs, ys, zs = proj3d.proj_transform(xs3d, ys3d,
                                                    zs3d, self.axes.M)
             else:
@@ -457,7 +465,7 @@ class Bloch:
             self.fig = plt.figure(figsize=self.figsize)
 
         if not axes:
-            if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+            if matplotlib_version_gte():
                 self.axes = Axes3D(self.fig, azim=self.view[0],
                                    elev=self.view[1], auto_add_to_figure=False)
                 self.fig.add_axes(self.axes)
@@ -478,10 +486,8 @@ class Bloch:
             self.axes.set_zlim3d(-0.7, 0.7)
         # Manually set aspect ratio to fit a square bounding box.
         # Matplotlib did this stretching for < 3.3.0, but not above.
-        if parse_version(matplotlib.__version__) >= parse_version('3.3'):
+        if matplotlib_version_gte(version='3.3'):
             self.axes.set_box_aspect((1, 1, 1))
-        else:
-            pass
 
         self.axes.grid(False)
         self.plot_back()
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -35,8 +35,12 @@ __all__ = ['Bloch']
 
 import os
 
+import matplotlib
+
 from numpy import (ndarray, array, linspace, pi, outer, cos, sin, ones, size,
                    sqrt, real, mod, append, ceil, arange)
+
+from packaging.version import parse as parse_version
 
 from qutip.qobj import Qobj
 from qutip.expect import expect
@@ -56,7 +60,10 @@ try:
 
         def draw(self, renderer):
             xs3d, ys3d, zs3d = self._verts3d
-            xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, renderer.M)
+            if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+                xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, self.axes.M)
+            else:
+                xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, renderer.M)
 
             self.set_positions((xs[0], ys[0]), (xs[1], ys[1]))
             FancyArrowPatch.draw(self, renderer)
@@ -448,7 +455,11 @@ class Bloch:
             self.fig = plt.figure(figsize=self.figsize)
 
         if not axes:
-            self.axes = Axes3D(self.fig, azim=self.view[0], elev=self.view[1])
+            if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+               self.axes = Axes3D(self.fig, azim=self.view[0], elev=self.view[1], auto_add_to_figure=False)
+               self.fig.add_axes(self.axes)
+            else:
+                self.axes = Axes3D(self.fig, azim=self.view[0], elev=self.view[1])
 
         if self.background:
             self.axes.clear()
@@ -463,7 +474,10 @@ class Bloch:
             self.axes.set_zlim3d(-0.7, 0.7)
         # Manually set aspect ratio to fit a square bounding box.
         # Matplotlib did this stretching for < 3.3.0, but not above.
-        self.axes.set_box_aspect((1, 1, 1))
+        if parse_version(matplotlib.__version__) >= parse_version('3.3'):
+            self.axes.set_box_aspect((1, 1, 1))
+        else:
+            pass
 
         self.axes.grid(False)
         self.plot_back()
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -51,10 +51,11 @@ try:
     from matplotlib.patches import FancyArrowPatch
     from mpl_toolkits.mplot3d import proj3d
 
+    # Define a custom _axes3D function based on the matplotlib version.
+    # The auto_add_to_figure keyword is new for matplotlib>=3.4.
     if parse_version(matplotlib.__version__) >= parse_version('3.4'):
-        def _axes3D(*args, **kwargs):
-            fig = args[0]
-            ax = Axes3D(*args, auto_add_to_figure=False, **kwargs)
+        def _axes3D(fig, *args, **kwargs):
+            ax = Axes3D(fig, *args, auto_add_to_figure=False, **kwargs)
             return fig.add_axes(ax)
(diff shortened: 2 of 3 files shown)


qutip/bloch.py
```diff
@@ -35,9 +35,13 @@
 
 import os
 
+import matplotlib
```
hodgestar, line 38: This import should be inside the `try: ... except: ...` block a little futher down. This file should be importable even if matplotlib is not installed (or at least that seems to be the intention).

hodgestar (changes requested):
Looking good! I left some suggestions for tidying it up a little (not quite sure how tidy it can be).

qutip/bloch.py
```diff
@@ -463,7 +478,10 @@ def render(self, fig=None, axes=None):
             self.axes.set_zlim3d(-0.7, 0.7)
         # Manually set aspect ratio to fit a square bounding box.
         # Matplotlib did this stretching for < 3.3.0, but not above.
-        self.axes.set_box_aspect((1, 1, 1))
+        if parse_version(matplotlib.__version__) >= parse_version('3.3'):
+            self.axes.set_box_aspect((1, 1, 1))
+        else:
+            pass
```
jakelishman, line 484: `else: pass` is unnecessary - it's fine to just have the `if ...:` part of the statement

qutip/bloch.py
```diff
@@ -56,7 +60,12 @@ def __init__(self, xs, ys, zs, *args, **kwargs):
 
         def draw(self, renderer):
             xs3d, ys3d, zs3d = self._verts3d
-            xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, renderer.M)
+            if parse_version(matplotlib.__version__) >= parse_version('3.4'):
```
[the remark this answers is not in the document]
jakelishman: `guide-bloch.rst` doesn't need a defended import of `matplotlib` because that's a dependency of the documentation build - it'll always be there.

The alternative is to keep the test like this here, but in all the places where it's about adding the `auto_add_to_figure=False` kwarg to `Axes3D`, you could do the test once at the top of the file, and rebind
```python
if [ ... matplotlib >= 3.4 ... ]:
    def Axes3D(*args, **kwargs):
        return mpl_toolkits.mplot3d.Axes3D(*args, auto_add_to_figure=False, **kwargs)
else:
    def Axes3D(*args, **kwargs):
        return mpl_toolkits.mplot3d.Axes3D(*args, **kwargs)
```
or something like that.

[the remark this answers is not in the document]
jakelishman: Also, I didn't test, but what's the earliest version of `matplotlib` that `axes.M` works in?  If it works in mpl 3.0, then we can ellide the version test in this particular instance entirely, and just use `axes.M` in all cases.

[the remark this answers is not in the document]
Yash-10: Thank you so much for the review @jakelishman ! In the recent commit, I have incorporated the reviously requested changes. I will surely comprehend your suggestions.

> The alternative is to keep the test like this here, but in all the places where it's about adding the auto_add_to_figure=False kwarg to Axes3D, you could do the test once at the top of the file, and rebind

Thank you for this suggestion! I was wondering whether this could take more time for it's instantiation since then we would check versions for every call to `Axes3D`, given that the places where such conditions are required are sparsely spread across the module? I might be missing on some places, and hence would love to get your feedback!

qutip/bloch.py
Yash-10: Thanks!

qutip/bloch.py
Yash-10: Thanks, I transferred it inside the `try` block..

qutip/bloch.py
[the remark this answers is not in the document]
Yash-10: > Also, I didn't test, but what's the earliest version of matplotlib that axes.M works in? If it works in mpl 3.0, then we can ellide the version test in this particular instance entirely, and just use axes.M in all cases.

@jakelishman I checked with `matplotlib==3.0.1` and `self.axes.M` does work. Using `matplotlib==3.0.0` with `python>3.6.6` gives some import errors but was fixed in `matplotlib==3.0.1` (as mentioned [here](https://stackoverflow.com/questions/52979322/matplotlib-3-0-0-cannot-import-name-get-backend-from-matplotlib)).

In any case, you had said that if it works for mpl 3.0, then we could omit version tests for `proj3d.proj_transform` which uses `axes.M`. Could you let me know the reason behind this?

Thanks!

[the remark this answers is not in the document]
Yash-10: > `guide-bloch.rst` doesn't need a defended import of `matplotlib` because that's a dependency of the documentation build - it'll always be there. (_edit_: just realised this wasn't about the defended import at all, but the versioning. I'm actually happy for the documentation to rely on matplotlib 3.4 if it makes things clearer; building the documentation isn't the same as running in a production environment, and I'd rather the documentation be legible without additional version checks. That said, we'll have to update the `requirements.txt` file, and that's a shade non-trivial, because there's actually some pinned versions in there that are very important.)
> 
> The alternative is to keep the test like this here, but in all the places where it's about adding the `auto_add_to_figure=False` kwarg to `Axes3D`, you could do the test once at the top of the file, and rebind
> 
> ```python
> if [ ... matplotlib >= 3.4 ... ]:
>     def Axes3D(*args, **kwargs):
>         return mpl_toolkits.mplot3d.Axes3D(*args, auto_add_to_figure=False, **kwargs)
> else:
>     def Axes3D(*args, **kwargs):
>         return mpl_toolkits.mplot3d.Axes3D(*args, **kwargs)
> ```
> 
> or something like that.


The above suggested change by @jakelishman seems to provide better readability than the current implementation in this PR, especially because the internal implementation of the function do not change within the module. I would make those changes and analyse its outcomes. Thanks!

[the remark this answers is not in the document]
jakelishman: > Thank you for this suggestion! I was wondering whether this could take more time for it's instantiation since then we would check versions for every call to `Axes3D`

The defended `def Axes3D` statement I did only tests the version once, at module initialisation, and then saves it for use later.  You don't need to worry about performance concerns between a single `if` statement and wrapper Python function in general, so prefer the one that's more readable (as you say).

> I checked with `matplotlib==3.0.1` and `self.axes.M` does work. Using `matplotlib==3.0.0` with `python>3.6.6` gives some import errors but was fixed in `matplotlib==3.0.1`

In that case, definitely better to prefer just using `self.axes.M` in all cases.  Don't worry if matplotlib messed up their 3.0.0 release - we still support it, it's just that 3.0.0 happens to be broken with new versions of Python.

> In any case, you had said that if it works for mpl 3.0, then we could omit version tests for `proj3d.proj_transform` which uses `axes.M`. Could you let me know the reason behind this?

In an ideal world, there would only be one way to do things in a library, and we would never have to defensively programme against different versions.  So, if there's a recommended way that works with all our supported versions, we prefer to do that and have as few version checks as we can - every time we have a version check, it means we've got duplicated code, which is harder to read and harder to maintain.

qutip/bloch.py
```diff
@@ -38,16 +38,28 @@
...
 from qutip.operators import sigmax, sigmay, sigmaz
 
 try:
+    import matplotlib
     import matplotlib.pyplot as plt
     from mpl_toolkits.mplot3d import Axes3D
     from matplotlib.patches import FancyArrowPatch
     from mpl_toolkits.mplot3d import proj3d
 
+    def matplotlib_version_gte(version='3.4'):
```
jakelishman, line 54: I wouldn't have a default value here at all - it's not obvious why it's 3.4 (somebody two months in the future will have no idea), so it's not clear what `matplotlib_version_gte()` will do.  Better to make all the calls `matplotlib_version_gte('3.4')` or so.

(a minor style note that isn't important at all: "private" functions that aren't meant to be exported usually start with an underscore `_` in Python convention.  It matters more if a module _doesn't_ specify an `__all__` metadata object, but in QuTiP we always do.)

qutip/bloch.py
```diff
@@ -38,16 +38,28 @@
...
+        """
+        Checks if matplotlib version is greater than a specific version
+        """
+        if parse_version(matplotlib.__version__) >= parse_version(version):
+            return True
+        else:
+            return False
```
jakelishman, line 61: `if x: return True; else: return False` is unnecessary - you can just `return x`, e.g.
```python
return parse_version(matplotlib.__version__) >= parse_version(version)
```

qutip/bloch.py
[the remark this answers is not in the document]
jakelishman: Oh, I should also note that my original redefinition of `Axes3D` should probably have included a catch for the figure and called `figure.add_axes(axes)` call inside it.

qutip/bloch.py
Yash-10: Oh! I should have done that before! Thank you for the suggestion!

qutip/bloch.py
Yash-10: > (a minor style note that isn't important at all: "private" functions that aren't meant to be exported usually start with an underscore _ in Python convention. It matters more if a module doesn't specify an __all__ metadata object, but in QuTiP we always do.)

Thanks, I will keep that in mind!

Yash-10:
Thank you @jakelishman and @hodgestar for your suggestions throughout the PR!

At this stage, following things are added:

-  Getting rid of the previously implemented `matplotlib_version_gte` method. Instead following @jakelishman 's advice to factor version checking once at the top and using it throughout the module. This allowed to prevent any version checking in the user guide since the `axes3D_` method can be directly imported.
- Since `axes.M` seems to work even for `matplotlib==3.0`, converted all instances of `renderer.M` to `axes.M` to keep version checking at the minimum.

Thanks!!

doc/guide/guide-bloch.rst
```diff
@@ -421,11 +421,13 @@ Directly Generating an Animation
 
 The code to directly generate an mp4 movie of the Qubit decay is as follows ::
 
+   import matplotlib
```
hodgestar, line 424: I think this `import matplotlib` is no longer needed now?

hodgestar (changes requested):
Left one more comment, otherwise looks good to me.

doc/guide/guide-bloch.rst
```diff
@@ -421,11 +421,13 @@ Directly Generating an Animation
...
+   from qutip.bloch import axes3D_
    from matplotlib import pyplot, animation
-   from mpl_toolkits.mplot3d import Axes3D
 
    fig = pyplot.figure()
-   ax = Axes3D(fig,azim=-40,elev=30)
+
+   ax = axes3D_(fig, azim=-40, elev=30)
```
jakelishman, line 430: I'd say there's no need to import the "fixed" `Axes3D` in the documentation.  Right now, just use whichever form works with matplotlib 3.3 - the documentation is currently built with matplotlib pinned to that.  Documentation doesn't need to build with _every_ version, it just needs to be repeatable, and I'd rather use the "core" version of `Axes3D` rather than a modified one in code that users will read.

(but @hodgestar if you want to weigh in differently, that's also fine - just mark this as resolved)

doc/guide/guide-bloch.rst
Yash-10: Thank you!

doc/guide/guide-bloch.rst
Yash-10: Due to the way `axes3D_` works, there should never be any deprecation warnings (irrespective of the matplotlib version) that otherwise creep in while using matplotlib's `Axes3D` directly (given `matplotlib<3.4`). 

If it's okay to have deprecation warnings (for users having older versions of matplotlib) due to the code in the user guide, using `Axes3D` from `matplotlib` directly would be a good choice since it provides better readability. As a newcomer, I still would have been hesitant to see `axes3D_` imported from `qutip.bloch` in the user guide :)

Thanks!!

Any thoughts on this? @jakelishman @hodgestar

hodgestar: Having read @jakelishman comment, I think we should remove the use of `axes3D_` from the documentation and maybe even rename it to `_axes3D`. We really don't want our users to be encouraged to use this internal helper function because then it will be harder to remove later.

@Yash-10 Thank you for putting up with the slightly back and forth reviewing! I think that are all the suggestions we're converging on a really nice solution.

Yash-10: Thank you @hodgestar and @jakelishman ! Following your suggestions, I have renamed `axes3D_` to `_axes3D` and used matplotlib's `Axes3D` in the user guide to prevent any confusion there.

> @Yash-10 Thank you for putting up with the slightly back and forth reviewing! I think that after all the suggestions we're converging on a really nice solution.


Yes! Thanks for your guidance throughout.

hodgestar (approved):
Looks good to me! Once @jakelishman is happy too, we can merge.

qutip/bloch.py
```diff
@@ -38,16 +38,28 @@
...
 try:
+    import matplotlib
     import matplotlib.pyplot as plt
     from mpl_toolkits.mplot3d import Axes3D
     from matplotlib.patches import FancyArrowPatch
     from mpl_toolkits.mplot3d import proj3d
 
+    if parse_version(matplotlib.__version__) >= parse_version('3.4'):
+        def _axes3D(*args, **kwargs):
+            fig = args[0]
```
jakelishman, line 56: It's better to spell this as
```python
def _axes3D(fig, *args, **kwargs):
    ...
```
because then if there's an error calling the function (i.e. by passing _no_ arguments), you get the "normal" Python error
```python
TypeError: _axes3D() missing 1 required positional argument: 'fig'
```
What I wrote is a pretty common Python idiom when you want to pick out the first positional argument (or more).  You can similarly pick out special keyword arguments with `def xxx(*args, out=None, linewidth=None, **kwargs)`.

It also would be good if you could write a little comment above the `if parse_version()` line (just a one- or two-line comment) explaining why this is necessary; we understand right now why it was put in place, but the next people to look at the code won't have the context of this PR.

jakelishman: (and a similar comment at the visualisation.py change as well)

jakelishman:
This is pretty much ready to go for me, too.  I added one more thing, mostly asking for a code comment to explain why we're doing this slightly unusual thing - we've always got to be thinking about the people who'll come after us to maintain the code, too.

Thank you very much for going through so many rounds of this.

qutip/bloch.py
Yash-10: Thanks @jakelishman for all these suggestions! I have made the necessary changes.

jakelishman (changes requested):
If you click through to the failing test and scroll through the logs, it's reporting a bug in visualization.py, which needs fixing before this can be merged.

Yash-10:
Ah, just needed to change the position of the import statements in `visualization.py`. It is fixed now!

Thanks !

qutip/visualization.py
```diff
@@ -372,7 +384,7 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
     """
     if fig is None or ax is None:
         fig = plt.figure()
-        ax = Axes3D(fig)
+        ax = _axes3D(fig, azim=-35, elev=35)
```
jakelishman, line 387: Is there a reason to change the arguments in this one?

Yash-10: Sorry for missing all these intricacies! There was absolutely no reason for me to change that. Thanks again.

jakelishman: That's no problem, it happens to everyone when we go back and forth between changes.  I think this is ready now, but we'll wait til the tests have officially passed.

jakelishman (approved):

qutip/visualization.py
Yash-10: Yes, absolutely!

hodgestar (approved):

PR #1561, opened 2021-05-27
Fix tlist QobJEvo with constant c_ops in mesolve

### LINKED ISSUE #1560: `tlist are not compatible` when using collapse operators in `mesolve`
Hi,
I'm trying to use `mesolve` using a sampled hamiltonian to be interpolated (_list array format_ according to [the doc](https://qutip.org/docs/latest/apidoc/functions.html?highlight=mesolve#module-qutip.mesolve)) with a `tlist` that isn't the time list `times` of the Hamiltonian coefficients, but a sublist `eval_times` of `times`. This method works fine and returns the results at `eval_times` when `c_ops=[]`, but if I specify collapse operators, I get a `tlist are not compatible` error.

Here is an example of what I'm saying.
```python
times = np.linspace(0, 1, 100)
x = np.cos(2*np.pi * times)
eval_times=times[0:-1:3]
rho = qutip.basis(2, 0).proj()
ham = qutip.QobjEvo([[qutip.sigmax(), x]], tlist=times)
qutip.mesolve(ham, rho0=rho, tlist=eval_times, c_ops=[]) # works fine
collapse = [qutip.qeye(2)] 
qutip.mesolve(ham, rho0=rho, tlist=times, c_ops=collapse) # works fine as well
qutip.mesolve(ham, rho0=rho, tlist=eval_times, c_ops=collapse) # raises `tlist are not compatible` error
```

Do you have any idea what might cause this behaviour ? I believe this might be an issue with the `mesolve` solver, since when you don't specify `c_ops`, you use `sesolve`.

jakelishman:
Previously, passing a Hamiltonian to `mesolve` which used "array" time dependence with a tlist not equal to `mesolve`'s would fail if any constant collapse operators were present.  This was because `mesolve` was a little over-zealous in promoting `c_ops` to time-dependent `QobjEvo` using the `mesolve` time list.  This would cause an unnecessary failure if the Hamiltonian was a `QobjEvo` specified previously with a different tlist.

Fix #1560

#### Changelog
- Specialise exception types in ``QobjEvo`` failure paths
- Fix ``tlist`` ``QobjEvo`` with constant collapse operators in ``mesolve``

### DIFF
--- a/qutip/qobjevo.py
+++ b/qutip/qobjevo.py
@@ -467,7 +467,7 @@ class QobjEvo:
                 self.const = True
                 self.type = "cte"
             elif op_type == 1:
-                raise Exception("The Qobj must not already be a function")
+                raise TypeError("The Qobj must not already be a function")
             elif op_type == -1:
                 pass
         else:
@@ -566,7 +566,7 @@ class QobjEvo:
                         elif isinstance(op_k[1], np.ndarray):
                             if not isinstance(tlist, np.ndarray) or not \
                                         len(op_k[1]) == len(tlist):
-                                raise TypeError("Time list does not match")
+                                raise TypeError("Time lists do not match")
                             op_type.append(3)
                         else:
                             raise TypeError("Incorrect Q_object specification")
@@ -633,7 +633,7 @@ class QobjEvo:
         try:
             t = float(t)
         except Exception as e:
-            raise TypeError("t should be a real scalar.") from e
+            raise TypeError("Time must be a real scalar.") from e
 
         if state is not None:
             self._dynamics_args_update(t, state)
@@ -871,7 +871,7 @@ class QobjEvo:
                     pass
                 elif len(other.tlist) != len(self.tlist) or \
                         other.tlist[-1] != self.tlist[-1]:
-                    raise Exception("tlist are not compatible")
+                    raise ValueError("Time lists are not compatible")
         else:
             self.cte += other
             self.dummy_cte = False
@@ -915,7 +915,8 @@ class QobjEvo:
             self.cte *= other
             for op in self.ops:
                 op.qobj *= other
-        elif isinstance(other, QobjEvo):
+            return self
+        if isinstance(other, QobjEvo):
             if other.const:
                 self.cte *= other.cte
                 for op in self.ops:
@@ -949,11 +950,8 @@ class QobjEvo:
                 self.num_obj = (len(self.ops) if
                               self.dummy_cte else len(self.ops) + 1)
             self._reset_type()
-
-        else:
-            raise TypeError("QobjEvo can only be multiplied"
-                            " with QobjEvo, Qobj or numbers")
-        return self
+            return self
+        return NotImplemented
 
     def __div__(self, other):
         if isinstance(other, (int, float, complex,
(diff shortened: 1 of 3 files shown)


jakelishman:
I've just pushed an extra commit purely to keep CodeClimate happy - I had to touch a line in a logic block that CodeClimate already didn't like, so it moaned that I hadn't fixed it.  This just separates out the logic to make it a bit easier to read.  As a side-effect, it allows iterable type to be used inside the time-dependent object list of `QobjEvo` (only - doesn't affect anywhere that doesn't use `QobjEvo`), so
```python
qutip.QobjEvo([qutip.sigmax(), (qutip.sigmay(), '1')])
```
is now a valid specifier (note the tuple in the list).

qutip/qobjevo.py
```diff
@@ -539,42 +543,39 @@ def __init__(self, Q_object=[], args={}, copy=True,
...
-                    else:
-                        raise TypeError("Incorrect Q_object specification")
+            return 0
+        if isinstance(Q_object, (FunctionType, BuiltinFunctionType, partial)):
+            return 1
+        if isinstance(Q_object, list):
+            return [self._td_op_type(element) for element in Q_object] or -1
+        raise TypeError("Incorrect Q_object specification")
+
+    def _td_op_type(self, element):
```
Ericgig, line 555: A lot more readable, thank you.

qutip/qobjevo.py
```diff
@@ -539,42 +543,39 @@ def __init__(self, Q_object=[], args={}, copy=True,
...
+            op, td = element
+        except (TypeError, ValueError) as exc:
+            raise TypeError("Incorrect Q_object specification") from exc
+        if (not isinstance(op, Qobj)) or isinstance(td, Qobj):
+            # Qobj is itself callable, so we need an extra check to make sure
+            # that we don't have a two-element list where both are Qobj.
+            raise TypeError("Incorrect Q_object specification")
+        if isinstance(td, Cubic_Spline):
+            out = 4
+        elif callable(td):
```
Ericgig, line 568: We don't test the validity of the function in `__init__` because of the `dynamics_args`: passed `args` could not be usable before parsing them with `_args_checks`. Do you think that testing them if `_args_checks` did no work would be worth it?

Ericgig (approved):

qutip/qobjevo.py
jakelishman: I'm not sure, to be honest - I didn't think about it at all here.  I only changed these parts of `QobjEvo` because I needed CodeClimate to be quiet while I made an unrelated change.  At least in the 4.x branch, I think what we've got is fine.  I think it'll be much easier to be rigorous in the 5.x branch with each term individually gaining its own `Coefficient`.

I didn't even want to make this change, since it's just one more thing to merge up into `dev.major`, but oh well...

I think that right now we shouldn't bother testing functions for validity - with the way dynamic arguments are handled in `QobjEvo` right now, it would probably involve some rather difficult logic to organise that, and I really wouldn't be convinced we'd get all the edge cases.  We can't catch every possible error a function could give anyway, so perhaps we should only attempt tests that are convenient for us to do?  I suppose we have to have some level of trust that the user will pass something usable.

PR #1562, opened 2021-05-27
merge test_rand and test_random

Ericgig:
**Description**
Tests for functions in `random_object.py` are split between `test_rand` and `test_random`. I merged those 2 files and rewrote some using `parametrize`.
The test themselves are mostly the same with some additions:
- `rank_ket` output's norm is checked .
- Given eigenvalue tests actually check the eigenvalues.
- `rand_herm` test for sparse and dense cases since the called function differ.

Fix a bug in `rand_herm` with `pos_def=True` and `density>0.5`.

**Changelog**
Improve random object tests.


qutip/random_objects.py
```diff
@@ -187,18 +187,20 @@ def _rand_herm_sparse(N, density, pos_def):
...
                                                  replace=False)])
-    M = sp.coo_matrix((data, (row_idx,col_idx)),
-                      dtype=complex, shape=(N,N)).tocsr()
+    M = sp.coo_matrix((data, (row_idx, col_idx)),
+                      dtype=complex, shape=(N, N))
     M = 0.5 * (M + M.conj().transpose())
     if pos_def:
-        M.setdiag(np.abs(M.diagonal())+np.sqrt(2)*N)
+        M = M.tolil()
+        M.setdiag(np.abs(M.diagonal()) + np.sqrt(2)*N)
```
jakelishman, line 195: What's the reason for converting to lil?  I'd have expected that the cost of converting to lil, and the increased cost of the lil->csr conversion (coo->csr is optimised in scipy) dominated any other effects.

qutip/random_objects.py
```diff
@@ -212,7 +214,7 @@ def _rand_herm_dense(N, density, pos_def):
         M[row, col] = 0
     if pos_def:
         as_vec = M.ravel()
-        M[::N+1] = (np.abs(M.diagonal())+np.sqrt(2)*N)
+        as_vec[::N+1] = (np.abs(M.diagonal()) + np.sqrt(2)*N)
```
jakelishman, line 217: This looks unsafe without a reassignment back to `M`:
```python
>>> x = np.array(np.eye(2), order='F')
>>> np.shares_memory(x.ravel(), x)
False
```

`np.ravel` doesn't guarantee that it returns a view.  We'd have the same C/Fortran problem with `np.ndarray.flat` and `np.reshape` as well, so we probably just have to reassign back to `M` to be safe.  I imagine we'll never see problems in reality because Numpy always defaults to C-order, but it's probably best not to rely on it just in case.  At any rate, reassignment to `M` is a free operation.

(*edit*: actually, on second thoughts, we can probably just use `np.fill_diagonal`)

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+    assert abs(random_qobj.tr() - 1.0) < 1e-15
+    # verify all eigvals are >=0
+    assert all(random_qobj.eigenenergies() >= -1e-15)
+    # verify hermitian
+    assert random_qobj.isherm
+
+
+@pytest.mark.repeat(5)
+def test_rand_dm_Eigs():
+    """
```
jakelishman, line 97: This shadows an earlier test, so the earlier won't be run.

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+@pytest.mark.repeat(5)
+@pytest.mark.parametrize('density', [0.2, 0.8], ids=["sparse", "dense"])
+@pytest.mark.parametrize('pos_def', [True, False])
+def test_rand_herm(density, pos_def):
+    """
+    Random Qobjs: Hermitian matrix
+    """
+    random_qobj = rand_herm(5, density=density, pos_def=pos_def)
+    if pos_def:
+        assert all(random_qobj.eigenenergies() > 0)
```
jakelishman, line 63: Possibly needs a slightly negative tolerance?  Like `assert all(... > -1e-8)` or whatever the tolerance of `eigvalsh` is.

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+@pytest.mark.repeat(5)
+def test_rand_dm_Eigs():
+    """
+    Random Qobjs: Hermitian matrix - Eigs given
+    """
+    eigs = np.random.random(5)
+    eigs /= np.sum(eigs)
+    eigs.sort()
+    random_qobj = rand_herm(eigs)
+    assert np.allclose(random_qobj.eigenenergies(), eigs)
```
jakelishman, line 76:
```suggestion
    np.testing.assert_allclose(random_qobj.eigenenergies(), eigs)
```

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+
+
+@pytest.mark.repeat(5)
+@pytest.mark.parametrize('func', [rand_dm, rand_dm_hs])
+def test_rand_dm(func):
+    """
+    Random Qobjs: Density matrix
+    """
+    random_qobj = func(5)
+    assert abs(random_qobj.tr() - 1.0) < 1e-15
```
jakelishman, line 88: can also be spelled `assert random_qobj.tr() == pytest.approx(1, abs=1e-15)` if you like, but I don't really mind.  My C upbringing makes me equally happy with the `abs(x - y) < tol` idiom for fp comparisons - don't know what other people think.

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+@pytest.mark.repeat(5)
+def test_rand_dm_Eigs():
+    """
+    Random Qobjs: Density matrix - Eigs given
+    """
+    eigs = np.random.random(5)
+    eigs /= np.sum(eigs)
+    eigs.sort()
+    random_qobj = rand_dm(eigs)
+    assert random_qobj.tr() - 1.0 < 1e-15
```
jakelishman, line 104: needs an abs

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+    assert np.allclose(random_qobj.eigenenergies(), eigs)
```
jakelishman, line 105: np.testing.assert_allclose

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
 
+@pytest.mark.repeat(5)
+@pytest.mark.parametrize('kind', ["left", "right"])
+def test_rand_stochastic(kind):
+    """
+    Random Qobjs: Test random stochastic
+    """
+    random_qobj = rand_stochastic(5, kind=kind)
+    axis = {"left":0, "right":1}[kind]
+    assert np.allclose(np.sum(random_qobj.full(), axis=axis), 1, atol=1e-15)
```
jakelishman, line 128: np.testing.assert_allclose

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+    """
+    Random Qobjs: Test random ket type and norm.
+    """
+    random_qobj = func(5)
+    assert random_qobj.type == 'ket'
+    assert abs(random_qobj.norm() - 1) < 1e-14
+
+
+@pytest.mark.repeat(5)
+def test_rand_super():
```
jakelishman, line 143: Do we want to test both `rand_super` and `rand_super_bcsz` here?

qutip/tests/test_random.py
```diff
@@ -35,34 +35,145 @@
...
+@pytest.mark.repeat(5)
 def test_rand_super_bcsz_cptp():
     """
     Random Qobjs: Tests that BCSZ-random superoperators are CPTP.
     """
-    S = rand_super_bcsz(5)
-    assert S.iscptp
+    random_qobj = rand_super_bcsz(5)
+    assert random_qobj.iscptp
+    assert random_qobj.issuper
```
jakelishman, line 158: Re previous comment: if we're going to test `rand_super_bcsz` only in this function, the `assert x.issuper` should come first, because I'm fairly sure `Qobj.iscptp` autofails if the object isn't super, but the failure mechanism will be masked if we allow the failure to occur there.

jakelishman (approved):
Looks good to me.  Let's merge - I'm interested in getting all our random-number generation swapped over to the "new" Numpy 1.17 methods as well, which should make seeding/reseeding rather easier to deal with, especially with threads.

PR #1573, opened 2021-06-15
matrix_histogram updated - imporved the styling options

mahdiaslanimk:
**Description**
`matrix_histogram` is much more eye-catching and has more options for syling:
1- zticks: ticks for z-axis
2- cmap: used colormap
3- cmap_min and cmap_max: truncating a portion of a colormap is now possible
4- bars_spacing: spacing between bars can be changed
5- bars_alpha: alpha channel of bars can be changed
6- bars_lw: bars' linewidth can be changed
7- bars_edgecolor: bars' color can be changed
8- shade: it's possible to turn on and off the shading
9- azimuthal and elevation angles can be changed
10- stick: bars can stick to xz and yz planes (z values can be read more easily)
11- cbar_pad: padding between the figure and the colorbar can be changed
12- proj_type: projection type can be changed to either 'perspective' or 'orthogonal'
13- xticks and yticks are now perfectly aligned to bars' center

**Changelog**

Improve the default colors and styling of matrix_histogram and provide additional styling options.

### DIFF
--- a/qutip/visualization.py
+++ b/qutip/visualization.py
@@ -443,19 +443,20 @@ def _limit_finder(z):
     return limit
 
 
-def _remove_margins():
+def _remove_margins(axis):
     """
     Removes margins about z=0 and improves the style
+    by monkey patching
     """
-    if not hasattr(Axis, "_get_coord_info_old"):
-        def _get_coord_info_new(self, renderer):
-            mins, maxs, centers, deltas, tc, highs = \
-                self._get_coord_info_old(renderer)
-            mins += deltas/4
-            maxs -= deltas/4
-            return mins, maxs, centers, deltas, tc, highs
-        Axis._get_coord_info_old = Axis._get_coord_info
-        Axis._get_coord_info = _get_coord_info_new
+    def _get_coord_info_new(renderer):
+        mins, maxs, centers, deltas, tc, highs = \
+            _get_coord_info_old(renderer)
+        mins += deltas / 4
+        maxs -= deltas / 4
+        return mins, maxs, centers, deltas, tc, highs
+
+    _get_coord_info_old = axis._get_coord_info
+    axis._get_coord_info = _get_coord_info_new
 
 
 def _truncate_colormap(cmap, minval=0.0, maxval=1.0, n=100):
@@ -642,9 +643,6 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
 
     """
 
-    # patch
-    _remove_margins()
-
     # default options
     default_opts = {'figsize': None, 'cmap': 'jet', 'cmap_min': 0.,
                     'cmap_max': 1., 'zticks': None, 'bars_spacing': 0.1,
@@ -657,7 +655,7 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
     if options:
         # check if keys in options dict are valid
         if set(options) - set(default_opts):
-            raise ValueError("invalid key(s) found in options: "+
+            raise ValueError("invalid key(s) found in options: " + \
                              f"{', '.join(set(options) - set(default_opts))}")
         else:
             # updating default options
@@ -731,6 +729,11 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
         cax, kw = mpl.colorbar.make_axes(ax, shrink=.75,
                                          pad=default_opts['cbar_pad'])
         mpl.colorbar.ColorbarBase(cax, cmap=cmap, norm=norm)
+
+    # removing margins
+    _remove_margins(ax.xaxis)
+    _remove_margins(ax.yaxis)
+    _remove_margins(ax.zaxis)
 
     return fig, ax
 
--- a/qutip/visualization.py
+++ b/qutip/visualization.py
@@ -577,7 +577,7 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
 
     options : dict
         A dictionary containing extra options for the plot.
-        The names (keys) and values of the options are 
+        The names (keys) and values of the options are
         described below:
 
         'zticks' : list of numbers
--- a/qutip/visualization.py
+++ b/qutip/visualization.py
@@ -504,16 +504,9 @@ def _update_zaxis(ax, z_min, z_max, zticks):
     updates the z-axis
     """
     ax.axes.w_zaxis.set_major_locator(plt.IndexLocator(1, 0.5))
-    # ax.set_zlim3d([min(z_min, 0), z_max])
-    if z_min > 0 and z_max > 0:
-        ax.set_zlim3d([0, z_max])
-    elif z_min < 0 and z_max < 0:
-        ax.set_zlim3d([0, z_min])
-    else:
-        ax.set_zlim3d([z_min, z_max])
-
-    if zticks:
+    if isinstance(zticks, list):
         ax.set_zticks(zticks)
+    ax.set_zlim3d([min(z_min, 0), z_max])
 
 
 def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
--- a/qutip/visualization.py
+++ b/qutip/visualization.py
@@ -479,24 +479,24 @@ def _stick_to_planes(stick, azim, ax, M, spacing):
     if stick is True:
         azim = azim % 360
         if 0 <= azim <= 90:
-            ax.set_ylim(1-0.55,)
-            ax.set_xlim(1-0.55,)
+            ax.set_ylim(1 - .5,)
+            ax.set_xlim(1 - .5,)
         elif 90 < azim <= 180:
-            ax.set_ylim(1-0.55,)
-            ax.set_xlim(0, M.shape[0]+(.5-spacing))
+            ax.set_ylim(1 - .5,)
+            ax.set_xlim(0, M.shape[0]+(.5 - spacing))
         elif 180 < azim <= 270:
-            ax.set_ylim(0, M.shape[1]+(.5-spacing))
-            ax.set_xlim(0, M.shape[0]+(.5-spacing))
+            ax.set_ylim(0, M.shape[1]+(.5 - spacing))
+            ax.set_xlim(0, M.shape[0]+(.5 - spacing))
         elif 270 < azim < 360:
-            ax.set_ylim(0, M.shape[1]+(.5-spacing))
-            ax.set_xlim(1-0.55,)
+            ax.set_ylim(0, M.shape[1]+(.5 - spacing))
+            ax.set_xlim(1 - .5,)
 
 
 def _update_yaxis(spacing, M, ax, ylabels):
     """
     updates the y-axis
     """
-    ytics = [x+(1-(spacing/2)) for x in range(M.shape[1])]
+    ytics = [x+(1 - (spacing / 2)) for x in range(M.shape[1])]
     ax.axes.w_yaxis.set_major_locator(plt.FixedLocator(ytics))
     if ylabels:
         nylabels = len(ylabels)
@@ -504,17 +504,17 @@ def _update_yaxis(spacing, M, ax, ylabels):
             raise ValueError(f"got {nylabels} ylabels but needed {len(ytics)}")
         ax.set_yticklabels(ylabels)
     else:
-        ax.set_yticklabels([str(y+1) for y in range(M.shape[1])])
+        ax.set_yticklabels([str(y + 1) for y in range(M.shape[1])])
         ax.set_yticklabels([str(i) for i in range(M.shape[1])])
     ax.tick_params(axis='y', labelsize=14)
-    ax.set_yticks([y+(1-(spacing/2)) for y in range(M.shape[1])])
+    ax.set_yticks([y+(1 - (spacing / 2)) for y in range(M.shape[1])])
 
 
 def _update_xaxis(spacing, M, ax, xlabels):
     """
     updates the x-axis
     """
-    xtics = [x+(1-(spacing/2)) for x in range(M.shape[1])]
+    xtics = [x+(1 - (spacing / 2)) for x in range(M.shape[1])]
     ax.axes.w_xaxis.set_major_locator(plt.FixedLocator(xtics))
     if xlabels:
         nxlabels = len(xlabels)
@@ -522,10 +522,10 @@ def _update_xaxis(spacing, M, ax, xlabels):
             raise ValueError(f"got {nxlabels} xlabels but needed {len(xtics)}")
         ax.set_xticklabels(xlabels)
     else:
-        ax.set_xticklabels([str(x+1) for x in range(M.shape[0])])
+        ax.set_xticklabels([str(x + 1) for x in range(M.shape[0])])
         ax.set_xticklabels([str(i) for i in range(M.shape[0])])
     ax.tick_params(axis='x', labelsize=14)
-    ax.set_xticks([x+(1-(spacing/2)) for x in range(M.shape[0])])
+    ax.set_xticks([x+(1 - (spacing / 2)) for x in range(M.shape[0])])
 
 
 def _update_zaxis(ax, z_min, z_max, zticks):
@@ -544,7 +544,8 @@ def _update_zaxis(ax, z_min, z_max, zticks):
     if zticks:
         ax.set_zticks(zticks)
     else:
-        ax.set_zticks([z_min+0.5*i for i in range(int((z_max-z_min)/0.5)+1)])
+        ax.set_zticks([z_min + 0.5 * i for i in 
+                       range(int((z_max - z_min) / 0.5) + 1)])
 
 
 def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
@@ -647,7 +648,7 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
     default_opts = {'figsize': None, 'cmap': 'jet', 'cmap_min': 0.,
                     'cmap_max': 1., 'zticks': None, 'bars_spacing': 0.1,
                     'bars_alpha': 1., 'bars_lw': 0.5, 'bars_edgecolor': 'k',
-                    'shade': False, 'azim': 65, 'elev': 30,
+                    'shade': False, 'azim': -35, 'elev': 35,
                     'proj_type': 'ortho', 'stick': False,
                     'cbar_pad': 0.04, 'cbarmax_to_zmax': False}
 
@@ -696,7 +697,7 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
         ax = _axes3D(fig,
                      azim=default_opts['azim'] % 360,
                      elev=default_opts['elev'] % 360)
-    ax.set_proj_type(default_opts['proj_type'])
+        ax.set_proj_type(default_opts['proj_type'])
 
     ax.bar3d(xpos, ypos, zpos, dx, dy, dz, color=colors,
              edgecolors=default_opts['bars_edgecolor'],
--- a/qutip/visualization.py
+++ b/qutip/visualization.py
@@ -418,11 +418,13 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
 def _limit_finder(z):
     """Finds nearest proper 0.5 value
     This funtion is used when limits is not passed to matrix_histogtam
+    If z > 0 it returns the higher half value
+    If z < 0 it returns the lower half value
     examples:
-        if z=0.1 returns 0.0
+        if z=+0.1  returns 0.5
         if z=-0.1 returns -0.5
         if z=-2.4 returns -2.5
-        if z=3.8  returns 4
+        if z=+3.8  returns 4.0
         if z=-5.0 returns -5.0
 
     Parameters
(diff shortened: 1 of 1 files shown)


mahdiaslanimk:
Comparing previous code and the new one:
matrix_histogram-1

mahdiaslanimk:
Hi folks
Anyone can help me on how to reduce the **Cognitive Complexity** of function `matrix_histogram`?

qutip/visualization.py
```diff
@@ -414,8 +415,135 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
...
+        cmap(np.linspace(minval, maxval, n)))
+    return new_cmap
+
+
+def _stick_to_planes(stick, azim, ax, M, bars_spacing):
+    """Adjusts xlim and ylim in way that bars will
+    Stick to xz and yz planes
+    """
+    if stick is True:
+        if 0 < azim <= 90 or -360 <= azim < -270 or azim == 0:
```
Ericgig, line 485: Doing `azim_= azim % 360` first would simplify the conditions.
Is it normal that `azim=0` and `azim=360` result in different behaviour?

Ericgig (commented):
Thank you for your PR, the new graphics are really nice. Allowing the options instead of forcing a new style is nice.
Thank you also for respecting the codeclimat.

qutip/visualization.py
```diff
@@ -414,8 +415,135 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
     return fig, ax
 
 
+def _limit_finder(z):
+    """Finds nearest proper 0.5 value
+    This funtion is used when limits is not passed to matrix_histogtam
+    examples:
+        if z=0.1 returns 0.0
```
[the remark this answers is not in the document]
mahdiaslanimk: Hi @Ericgig , thanks for reviewing the code, your corrections and your great suggestions🙏. I'll apply them as soon as possible

qutip/visualization.py
```diff
@@ -414,8 +415,135 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
...
+            limit = int(z)+0.5
+        else:
+            limit = int(z+0.5)
+    elif z < 0:
+        if int(z-0.5) > z > int(z)-0.5 or z % 1 == 0.5:
+            limit = int(z)-0.5
+        else:
+            limit = int(z-0.5)
+    else:
+        limit = 0
```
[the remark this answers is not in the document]
mahdiaslanimk: Yes, much shorter and wiser 👌

qutip/visualization.py
```diff
@@ -414,8 +415,135 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
...
+    """
+    if not hasattr(Axis, "_get_coord_info_old"):
+        def _get_coord_info_new(self, renderer):
+            mins, maxs, centers, deltas, tc, highs = \
+                self._get_coord_info_old(renderer)
+            mins += deltas/4
+            maxs -= deltas/4
+            return mins, maxs, centers, deltas, tc, highs
+        Axis._get_coord_info_old = Axis._get_coord_info
+        Axis._get_coord_info = _get_coord_info_new
```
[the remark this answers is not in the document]
mahdiaslanimk: @Ericgig I don't know how to do it, can you help me?

qutip/visualization.py
mahdiaslanimk: The auto ticking function used in making a 3D matplotlib figure distinguishes between `azim=0` and `azim=360` and that's why I did the same:
```python
from mpl_toolkits.mplot3d import Axes3D

fig = plt.figure()
ax  = Axes3D(fig, azim=0, elev=0)
```
**output:**
azim0

```python
fig = plt.figure()
ax  = Axes3D(fig, azim=360, elev=0)
```
**output:**
azim360

Your suggestion greatly simplified the code, so ended up in something like this:

```python
def _stick_to_planes(azim, ax, M, spacing):
    """Adjusts xlim and ylim in way that bars will
    Stick to xz and yz planes
    """
    azim = azim % 360
    if 0 <= azim <= 90:
        ax.set_ylim(.5,)
        ax.set_xlim(.5,)
    elif 90 < azim <= 180:
        ax.set_ylim(.5,)
        ax.set_xlim(0, M.shape[0]+(.5-spacing))
    elif 180 < azim < 270:
        ax.set_ylim(0, M.shape[1]+(.5-spacing))
        ax.set_xlim(0, M.shape[0]+(.5-spacing))
    elif 270 <= azim < 360:
        ax.set_ylim(0, M.shape[1]+(.5-spacing))
        ax.set_xlim(.5,)
        
def matrix_histogram(M, ...):
    ...
    fig = plt.figure(figsize)
    ax = _axes3D(fig, azim=azim % 360, elev=azim % 360)
    ...
    # stick to xz and yz plane
    if stick is True:
        _stick_to_planes(azim_,ax, M, bars_spacing)
    ...
    return fig, ax
```

qutip/visualization.py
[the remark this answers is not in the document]
mahdiaslanimk: @Ericgig The problem comes from monkey patching the `Axis` class. Instead, we can monkey patch `axis` instances like this:
```python
# monkey patching axis function
def _remove_margins(axis):
    """
    Removes margins about z=0 and improves the style
    by monkey patching
    """
    def _get_coord_info_new(renderer):
        mins, maxs, centers, deltas, tc, highs = \
            _get_coord_info_old(renderer)
        mins += deltas / 4
        maxs -= deltas / 4
        return mins, maxs, centers, deltas, tc, highs

    _get_coord_info_old = axis._get_coord_info
    axis._get_coord_info = _get_coord_info_new
.
.
.

def matrix_histogram(M,...):
    fig = plt.figure(figsize=default_opts['figsize'])
    ax = _axes3D(fig, azim, elev)
    .
    .
    .
    
    _remove_margins(ax.xaxis)
    _remove_margins(ax.yaxis)
    _remove_margins(ax.zaxis)

    return fig, ax
```

qutip/visualization.py
```diff
@@ -452,70 +643,98 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
     if ax is None:
-        fig = plt.figure()
-        ax = _axes3D(fig, azim=-35, elev=35)
-
-    ax.bar3d(xpos, ypos, zpos, dx, dy, dz, color=colors)
+        fig = plt.figure(figsize=default_opts['figsize'])
+        ax = _axes3D(fig,
+                     azim=default_opts['azim'] % 360,
+                     elev=default_opts['elev'] % 360)
+    ax.set_proj_type(default_opts['proj_type'])
```
Ericgig, line 699: I would expect `set_proj_type` to be in the `if ax is None:` block.

qutip/visualization.py
```diff
@@ -414,8 +415,140 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
...
+        ax.set_zlim3d([0, z_max])
+    elif z_min < 0 and z_max < 0:
+        ax.set_zlim3d([0, z_min])
+    else:
+        ax.set_zlim3d([z_min, z_max])
+
+    if zticks:
+        ax.set_zticks(zticks)
+    else:
+        ax.set_zticks([z_min+0.5*i for i in range(int((z_max-z_min)/0.5)+1)])
```
Ericgig, line 547: Please add more spacing around operator.

Ericgig (commented):
Hi mahdiaslanimk, looking good, there are a few place where the mathematics operations are quite dense, we prefer having operators surrounded by space.

I tried the new options most work well.
- `azim`'s default was better at -35 since for diagonal object, 65 create an overlaps.
- `stick`: I could not visually observe a difference (`axim=200`), and I can't understand what it should do from the description.

mahdiaslanimk:
Dear @Ericgig, thanks for reviewing the code.
- I changed value of `azim` to the previous one (-35).
- If you're changing azimuthal view angle by writing `ax.view_init(azim=200, elev=35)`:
```python
...
options = {...}  # No 'azim' in options
fig, ax = matrix_histogram(H, xlabels, xlabels, limits=[-4,4],options=options)
ax.view_init(azim=200, elev=35)
plt.show()
```
the `stick` won't work and that's because `stick` works based on the `azim` value passed via `options`. So if you write:
```python
...
options = {..., 'azim':200, ...}  # 'azim' in options
fig, ax = matrix_histogram(H, xlabels, xlabels, limits=[-4,4],options=options)
plt.show()
```
this will work.

Ericgig (commented):
Looking good.
I noted 2 patterns where the spacing around operators that still look strange.

Other than that, the doc-string about `options` is in a format that match with other functions in this file, but not the recent qutip style... (no article or capitalization) @hodgestar could you look at it?

qutip/visualization.py
```diff
@@ -414,8 +415,141 @@ def sphereplot(theta, phi, values, fig=None, ax=None, save=False):
...
+        elif 270 < azim < 360:
+            ax.set_ylim(0, M.shape[1]+(.5 - spacing))
+            ax.set_xlim(1 - .5,)
+
+
+def _update_yaxis(spacing, M, ax, ylabels):
+    """
+    updates the y-axis
+    """
+    ytics = [x+(1 - (spacing / 2)) for x in range(M.shape[1])]
```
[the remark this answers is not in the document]
mahdiaslanimk: @Ericgig Oh, sorry, I missed those

jakelishman:
As a comparison, here's the output of `matrix_histogram` from [the example in the _current_ build of the docs](https://qutip.org/docs/latest/guide/guide-visualization.html#visualizing-operators) (QuTiP 4.6.2):

QuTiP 4.6.2 guide-visualization-8

compared to the version produced by this PR:

this PR guide-visualization-8

The original certainly isn't great, but I think its defaults have produced a somewhat better result in this case.  It might be worth having a look at some of the defaults here, particularly with regards to the z-axis tic labelling. While we're at it, the rainbow colour scheme is not a good choice - it'd be much better to use a diverging colourmap, and preferably something a bit more perceptually uniform.

---

I'm not completely sold on the idea that our functions should take so many different optional arguments.  We return the figure and axes instances, which means anything the user can do anything that's basic matplotlib commands - we don't need to reimplement their interface.  Things like the figure size, azimuth and elevation are particular examples of this - it's very simple to set them using matplotlib's interface, and we don't really provide any added benefit by having the options here.

jakelishman:
That all said, I don't have strong opinions about the visualisation functions, so I won't block merging if other people are happy - Eric if you're content, go ahead and merge without me.

Ericgig:
I like having the same colour for all 0s and having x and y ticks centred on the columns. But I prefer the old, less flashy, colours. I would also prefer all our visualization functions to use the same options. But overall, I think it's an improvement on what we have.

mahdiaslanimk:
@Ericgig ,@jakelishman thanks for reviews and the comments. I agree with Jake, the default `'jet'` colormap isn't very nice and I believe choosing the right colormap depends on the problem and values of the input matrix `M`. Personaly I love to have 0s in white and the others colored and that's because non-zeros elements maybe important and should catch the attention.
I thinks the code (in docs) producing the input matrix should be changed and that's because:
1- it has lots of off-diagonal elements which raises the renown problem of overlaping bars. 
2- maximum value and minum value of input matrix are very different and nearly non of the colormaps makes the results nice.

Showing a thermal state's density matrix maybe a good option:
```python
from qutip import thermal_dm, matrix_histogram

rho_th = thermal_dm(6,2)
options = {'cmap':'hot_r',
                  'stick': True,
                  'cbar_pad':.0,
                  'zticks':[0, 0.365, 0.5],
                  'azim':65,
                  'elev':35,
                  'shade': False}
matrix_histogram(rho_th, limits=[0,0.5],options=options)
```
anim1

**the old one:**

anim2

I agree that because the function returns `fig` and `ax` the user can modify nearly everything, but most of them may not be easy to implement for everyone.

If we want the colors to be less flashy we have to set `shade` to `True` and there are 2 problems with turning on the shading:
1- bars with nearly equal values have different colors (like 0s)
2- in some azimuthal viewing angle the `lightsource` produces shadows

mat_hist_new_azim=30
**turing around by 180°:**
mat_hist_new_azim=210


the second one can be solved for example by manually flipping the input matrix two times and reversing x and y labels (or maybe by manipuating the `lightsource` which I haven't experiment so far):
```python
...
matrix_histogram(np.flip(np.flip(rho_th,axis=1),axis=0),
                               xlabels=[str(5-i) for i in range(6)],
                               ylabels=[str(5-i) for i in range(6)],
                               limits=[0,0.5], options=options)
```
mat_hist_new_azim=220_trick

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
         The axes context in which the plot will be drawn.
 
-    Returns
+    colorbar : bool (default: True)
+        show colorbar
+
+    options : dict
+        dictionary containing extra options
+        all keys are of type `str` and values have different types
+        key:value  pairs should be as follows:
```
hodgestar, line 582:
```suggestion
        A dictionary containing extra options for the plot.
        The names (keys) and values of the options are described below:
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'zticks' : list of numbers
+            list of z-axis ticks location
```
hodgestar, line 585:
```suggestion
            A list of z-axis tick locations.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'cmap' : string (default: 'jet')
+            colormap name
```
hodgestar, line 588:
```suggestion
            The name of the color map to use.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'cmap_min' : float (default: 0.0)
+            colormap truncation minimum, a value in range 0-1
```
hodgestar, line 591:
```suggestion
            The lower bound to truncate the color map at. A value in range 0 - 1. The default, 0, leaves the lower
            bound of the map unchanged.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'cmap_max' : float (default: 1.0)
+            colormap truncation maximum, a value in range 0-1
```
hodgestar, line 594:
```suggestion
            The upper bound to truncate the color map at. A value in range 0 - 1. The default, 1, leaves the upper
            bound of the map unchanged.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'bars_spacing' : float (default: 0.1)
+            spacing between bars
```
hodgestar, line 597:
```suggestion
            The size of the spacing between bars.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'bars_alpha' : float (default: 1.)
+            transparency of bars, should be in range 0-1
```
hodgestar, line 600:
```suggestion
            The transparency of the bars. An alpha value in range 0 - 1. The default, 1, renders opaque bars and 0
            renders completely transparent bars.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'bars_lw' : float (default: 0.5)
+            linewidth of bars' edges
```
hodgestar, line 603:
```suggestion
            The linewidth of bars' edges.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'bars_edgecolor' : color (default: 'k')
+            color of bars' edges, examples: 'k', (0.1, 0.2, 0.5), '#0f0f0f80'
```
hodgestar, line 606:
```suggestion
            The colors of the bars' edges. For example: 'k', (0.1, 0.2, 0.5) or '#0f0f0f80'.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'shade' : bool (default: True)
+            when True, this shades dark sides of the bars (relative
+            to plot's source of light).
```
hodgestar, line 610:
```suggestion
            Whether to shade the dark sides of the bars (True) or not (False). The shading is relative
            to plot's source of light.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'azim' : float
+            azimuthal viewing angle.
```
hodgestar, line 613:
```suggestion
            The azimuthal viewing angle.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'elev' : float
+            elevation viewing angle.
```
hodgestar, line 616:
```suggestion
            The elevation viewing angle.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'proj_type' : string (default: 'ortho' if ax is not passed)
+            type of projection ('ortho' or 'persp')
```
hodgestar, line 619:
```suggestion
            The type of projection ('ortho' or 'persp')
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'stick' : bool (default: False)
+            changes xlim and ylim in a way that bars next to
+            xz and yz planes will stick to those planes
+            works if ax is not passed to the function
```
hodgestar, line 624:
```suggestion
            Changes xlim and ylim in such a way that bars next to
            XZ and YZ planes will stick to those planes.
            This option has no effect if ``ax`` is passed as a parameter.
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'cbar_pad' : float (default: 0.04)
+            fraction of original axes between colorbar and new image axes
+            (padding between 3D figure and colorbar).
```
hodgestar, line 628:
```suggestion
            The fraction of the original axes between the colorbar and the new image axes
            (i.e. the padding between the 3D figure and the colorbar).
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'cbarmax_to_zmax' : bool (default: False)
+            set color of maximum z-value to maximum color of colorbar
```
hodgestar, line 631:
```suggestion
            Whether to set the color of maximum z-value to the maximum color in the colorbar (True) or not (False).
```

qutip/visualization.py
```diff
@@ -439,7 +573,67 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
...
+
+        'figsize' : tuple of two numbers
+            size of the figure
```
hodgestar, line 634:
```suggestion
            The size of the figure.
```

hodgestar (commented):
Some suggestions for docstring improvements.

mahdiaslanimk:
@hodgestar Thanks a lot for the suggestions Simon 🙏

Ericgig:
There are a few suggestion from Hodgestar you missed. You can commit then in batch in the 'Files Changed' tab.

Ericgig (commented):
Trying this some more, the tick for the z axis is fixed to `0.5`. So for input with values not in the unit range, it look quite strange.
So `qutip.matrix_histogram(qutip.rand_dm(20) / 10)`, looked ok before, but flat now.

Instead of fixing the tick to 0.5, we could compute it with:
```
delta_z = max(dz) - min(dz)
scale = 10**np.floor(np.log10(delta_z))
if delta_z / scale < 5:
    scale /= 2
```
then using this scale in `_limit_finder` and `_update_zaxis`, make the function support any range of values. 
matplotlib automatically scale with input, is `_limit_finder` really needed?

mahdiaslanimk:
@Ericgig Thanks for reviewing the code. As I'm thinking about it now, as you said there are cases with matrix elements in ranges very different than 1.0 (or 0.5). Since passing `zticks` (and `limits`) to the function is also possible, there is no need to auto-tick the Z axis (and hence no need to use `_limit_finder`) and we can rely on matplotlib's auto-ticking functionality (although it might not work properly in some ranges, e.g. `qutip.matrix_histogram(qutip.rand_dm(20, seed=42) / 10)`).

Ericgig (approved):
Look good, thank you for your contribution.

hodgestar (approved):

hodgestar:
I've added this to the [4.7](https://github.com/qutip/qutip/milestone/12) milestone and added a changelog entry to the description.

PR #1541, opened 2021-05-02
Fix handling of "release" flag in sdist

jakelishman:
This removes the `CI_QUTIP_RELEASE` environment variable and the `--release` flag to `setup.py` in favour of deriving this information from the `VERSION` file.  This file should now contain a `.dev` suffix on the `master` and `dev.major` branches to indicate that they are not releases, while release branches will omit it.  This means that sdists will no longer attempt to build in local development mode.

This also fixes the related bug in handling local versioning when `git` is not installed on the system by making sure any relevant errors are caught.

Fix #1514 (after a new sdist release)

PR #1574, opened 2021-06-15
Cherry-pick build and CI changes from `master`

jakelishman:
The main purpose of this patch is to cherry-pick up the commits which allow us to move our testing infrastructure to GitHub Actions, without absolutely messing up the merge of #1481.  After that is merged, I'll merge everything up properly.  This stop-gap measure is necessary because Travis CI is shutting down its free tier _today_, and we haven't managed to effect the merge yet.

Unfortunately, I had to pick out rather more than I had hoped to get this without seriously rewriting `__init__.py`, `setup.py` and the testing workflow, so apologies for the larger PR.  I have, however, cherry-picked out PRs in full, rather than splitting them, to make it logically a bit simpler.  I have also cherry-picked the PRs commit-by-commit, rather than squashing them, to make it easier to understand the history when I have to do a larger merge later (even though it'll still be more confusing in the `git log` than desirable).

The cherry-picked PRs are:
- #1465
- #1471
- #1484
- #1510
- #1541
- #1551 (the actual target).

I had to fix several merge conflicts in `__init__.py` over time, but most stuff was trivial to solve.  The largest merge conflicts that needed to be handled were issues where the old and new methods of handling settings conflicted, but again, it's obvious what is the correct version in this.

I have also disabled warnings-as-errors in the testing infrastructure, until we've fixed them properly on this branch.


Ericgig (approved):

jakelishman:
@hodgestar: I'll leave this for a bit if you want to review it, but it's _almost_ all just code that we've already included in `master`.  There's several points where things can clearly be improved (`__init__.py`), but that's also the case on `master`, and so far it's not been a super high priority.

hodgestar (approved):
Thanks @jakelishman.

We should remember to put back `-Werror` later.

jakelishman:
Absolutely - it already helped a load on `master` to catch pending deprecations and old code. As soon as we've got the warnings fixed on `dev.major` we'll turn it right back on.

PR #1575, opened 2021-06-15
Tidy up pep8 and minor logic bugs

jakelishman:
Some minor pep8 fixes in files I might touch soon with regards to moving to the Numpy 1.17 random-number generation.  There's a bunch of seemingly dead code in `mcsolve` about a "diagonal" system?  Seems to have a few logic bugs in it at the very least, and no way to access it.  I fixed one about `e_ops`, but looking at the log, seems like the whole system might not be needed.

**Changelog**: pep8 changes.

### DIFF
--- a/qutip/control/pulsegen.py
+++ b/qutip/control/pulsegen.py
@@ -59,29 +59,29 @@ def create_pulse_gen(pulse_type='RND', dyn=None, pulse_params=None):
     The pulse generators each produce a different type of pulse,
     see the gen_pulse function description for details.
     These are the random pulse options:
-        
+
         RND - Independent random value in each timeslot
         RNDFOURIER - Fourier series with random coefficients
         RNDWAVES - Summation of random waves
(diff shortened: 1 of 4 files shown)


hodgestar (approved):

Ericgig (approved):
139 lines of only white space change in pulsegen.py!

hodgestar:
It would be good to decide on the fate of the diagonal system. @Ericgig Any thoughts on whether we remove this or should keep it?

Ericgig:
We can remove it here. It is working but does not have a user facing interface or documentation, so I don't think any user use it.

jakelishman:
Ok, cool.  I'll merge this and we can remove unused bits of `mcsolve` in a different PR.

> 139 lines of only white space change in pulsegen.py!

I was just searching for `np.random` in the codebase and fixing files with that and pep8 errors, but I think there might be quite a few more hiding in the control module!

PR #1576, opened 2021-06-16
Fix pep8 and cyclomatic complexity in _mkl

jakelishman:
This simplifies some of the long if/else blocks in the MKL module, making it simpler to parse.  No logic should actually have changed.

Everything is at most 10 cyclomatic complexity now, but I don't think that corresponds exactly to CodeClimate's metrics, so let's see.

**Changelog**: simplify logic in `_mkl` module.


qutip/_mkl/spsolve.py
```diff
@@ -127,95 +136,127 @@ def __init__(self, np_pt=None, dim=None, is_complex=None,
...
+    11: 'Real non-symmetric',
+    13: 'Complex non-symmetric',
+}
+
+
+def _mkl_matrix_type(dtype, solver_args):
+    mtype = (
+        (0b0100 if dtype == np.complex128 else 0b0010)
+        | (0b0000 if solver_args['hermitian'] else 0b1001)
+    )
```
Ericgig, line 250: Why use bytes and not int?
```
    mtype = (
        (4 if dtype == np.complex128 else 2)
        + (0 if solver_args['hermitian'] else 9)
    )
```

qutip/tests/test_mkl.py
```diff
@@ -31,205 +31,120 @@
...
+            [0, 0, 0],
+        ], dtype=dtype)
+        sX = mkl_spsolve(sM, N, verbose=True)
+        X = scipy.linalg.solve(M, N)
+        np.testing.assert_allclose(X, sX)
+
+    def test_rhs_shape_is_maintained(self):
+        """
+        MKL spsolve : Solution shape same as input RHS vec
+        """
```
Ericgig, line 88: Why keep the doc-string only here?

qutip/_mkl/spsolve.py
jakelishman: I had (erroneously, it turns out) assumed that this was interpreted as a bit field.  It's actually just specific numbers.  I'll change it to be more legible.

qutip/tests/test_mkl.py
jakelishman: A mistake - in my editor I have docstrings folded unless I open them, and I just overlooked it.  The docstrings don't add any information that we don't get from the names of the tests, which is why I removed them.

Ericgig (approved):

PR #1578, opened 2021-06-17
Improve coverage of states and operators tests

Ericgig:
Add tests for previously untested function in states and operators and parametrize some of the existing tests.
Should improve the coverall by about 100 lines.

**Changelog**
Improve test coverage of states and operators functions.

### DIFF
--- a/qutip/tests/test_operators.py
+++ b/qutip/tests/test_operators.py
@@ -43,29 +43,34 @@ N = 5
 def test_jmat_12():
     spinhalf = qutip.jmat(1 / 2.)
 
-    paulix = np.array([[0.0 + 0.j, 0.5 + 0.j], [0.5 + 0.j, 0.0 + 0.j]])
-    pauliy = np.array([[0. + 0.j, 0. - 0.5j], [0. + 0.5j, 0. + 0.j]])
-    pauliz = np.array([[0.5 + 0.j, 0.0 + 0.j], [0.0 + 0.j, -0.5 + 0.j]])
+    paulix = np.array([[0. + 0.j, 1. + 0.j], [1. + 0.j, 0. + 0.j]])
+    pauliy = np.array([[0. + 0.j, 0. - 1.j], [0. + 1.j, 0. + 0.j]])
+    pauliz = np.array([[1. + 0.j, 0. + 0.j], [0. + 0.j, -1. + 0.j]])
     sigmap = np.array([[0. + 0.j, 1. + 0.j], [0. + 0.j, 0. + 0.j]])
     sigmam = np.array([[0. + 0.j, 0. + 0.j], [1. + 0.j, 0. + 0.j]])
 
-    np.testing.assert_allclose(spinhalf[0].full(), paulix)
-    np.testing.assert_allclose(spinhalf[1].full(), pauliy)
-    np.testing.assert_allclose(spinhalf[2].full(), pauliz)
+    np.testing.assert_allclose(spinhalf[0].full() * 2, paulix)
+    np.testing.assert_allclose(spinhalf[1].full() * 2, pauliy)
+    np.testing.assert_allclose(spinhalf[2].full() * 2, pauliz)
     np.testing.assert_allclose(qutip.jmat(1 / 2., '+').full(), sigmap)
     np.testing.assert_allclose(qutip.jmat(1 / 2., '-').full(), sigmam)
 
-    np.testing.assert_allclose(qutip.spin_Jx(1 / 2.).full(), paulix)
-    np.testing.assert_allclose(qutip.spin_Jy(1 / 2.).full(), pauliy)
-    np.testing.assert_allclose(qutip.spin_Jz(1 / 2.).full(), pauliz)
+    np.testing.assert_allclose(qutip.spin_Jx(1 / 2.).full() * 2, paulix)
+    np.testing.assert_allclose(qutip.spin_Jy(1 / 2.).full() * 2, pauliy)
+    np.testing.assert_allclose(qutip.spin_Jz(1 / 2.).full() * 2, pauliz)
     np.testing.assert_allclose(qutip.spin_Jp(1 / 2.).full(), sigmap)
     np.testing.assert_allclose(qutip.spin_Jm(1 / 2.).full(), sigmam)
 
-    np.testing.assert_allclose(qutip.sigmax().full(), paulix * 2)
-    np.testing.assert_allclose(qutip.sigmay().full(), pauliy * 2)
-    np.testing.assert_allclose(qutip.sigmaz().full(), pauliz * 2)
+    np.testing.assert_allclose(qutip.sigmax().full(), paulix)
+    np.testing.assert_allclose(qutip.sigmay().full(), pauliy)
+    np.testing.assert_allclose(qutip.sigmaz().full(), pauliz)
     np.testing.assert_allclose(qutip.sigmap().full(), sigmap)
     np.testing.assert_allclose(qutip.sigmam().full(), sigmam)
+
+    spin_set = qutip.spin_J_set(0.5)
+    for i in range(3):
+        assert spinhalf[i] == spin_set[i]
+
 
 
 def test_jmat_32():
@@ -104,7 +109,19 @@ def test_jmat_dims(spin, N):
     assert spin_mat.shape == (N, N)
 
 
+def test_jmat_raise():
+    with pytest.raises(TypeError) as e:
+        qutip.jmat(0.25)
+    assert str(e.value) == 'j must be a non-negative integer or half-integer'
+
+    with pytest.raises(TypeError) as e:
+        qutip.jmat(0.5, 'zx+')
+    assert str(e.value) == 'Invalid type'
+
+
 @pytest.mark.parametrize(['oper_func', 'diag', 'offset', 'args'], [
+    pytest.param(qutip.qeye, np.ones(N), 0, (), id="qeye"),
+    pytest.param(qutip.qzero, np.zeros(N), 0, (), id="zeros"),
     pytest.param(qutip.destroy, np.arange(1, N)**0.5, 1, (), id="destroy"),
     pytest.param(qutip.destroy, np.arange(6, N+5)**0.5, 1, (5,),
                  id="destroy_offset"),
@@ -117,7 +134,7 @@ def test_jmat_dims(spin, N):
     pytest.param(qutip.charge, np.arange(2, N+1)/3, 0, (2, 1/3),
                  id="charge_args"),
 ])
-def test_diagonal_oper(oper_func, diag, offset, args):
+def test_diagonal_operators(oper_func, diag, offset, args):
     oper = oper_func(N, *args)
     assert oper == qutip.Qobj(np.diag(diag, offset))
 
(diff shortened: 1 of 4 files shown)


qutip/tests/test_operators.py
```diff
@@ -33,34 +33,43 @@
...
 
 
 def test_jmat_12():
-    "Spin 1/2 operators"
-    spinhalf = jmat(1 / 2.)
+    spinhalf = qutip.jmat(1 / 2.)
 
     paulix = np.array([[0.0 + 0.j, 0.5 + 0.j], [0.5 + 0.j, 0.0 + 0.j]])
     pauliy = np.array([[0. + 0.j, 0. - 0.5j], [0. + 0.5j, 0. + 0.j]])
     pauliz = np.array([[0.5 + 0.j, 0.0 + 0.j], [0.0 + 0.j, -0.5 + 0.j]])
```
jakelishman, line 48: These aren't actually the definitions of the Pauli matrices??  (I know you didn't add this code - perhaps we could change them to be the actual definitions, and adjust the tests accordingly?)

qutip/tests/test_operators.py
```diff
@@ -79,51 +88,38 @@ def test_jmat_32():
...
+    pytest.param(qutip.create, np.arange(1, N)**0.5, -1, (), id="create"),
+    pytest.param(qutip.create, np.arange(6, N+5)**0.5, -1, (5,),
+                 id="create_offset"),
+    pytest.param(qutip.num, np.arange(N), 0, (), id="num"),
+    pytest.param(qutip.num, np.arange(5, N+5), 0, (5,), id="num_offset"),
+    pytest.param(qutip.charge, np.arange(-N, N+1), 0, (), id="charge"),
+    pytest.param(qutip.charge, np.arange(2, N+1)/3, 0, (2, 1/3),
+                 id="charge_args"),
+])
+def test_diagonal_oper(oper_func, diag, offset, args):
```
jakelishman, line 120: Can we call it `test_diagonal_operators` (and similar for all tests) - the names appear in the log, so they're basically the top-level test documentation.  We never have to type them, so it's probably better to spell everything out in full.

Sidenote: some languages have syntax to let you define a "literal" identifier, which can be an arbitrary string including spaces, or a single keyword!  I wish we had them in Python - it's really nice to have (e.g.)
```fsharp
let ``test diagonal operators are created correctly`` func diag offset args =
  let operator = func diag offset args
  assert oper == (np.diag diag offset |> qutip.Qobj)
```
(sidesidenote: apparently GitHub's F# highlighting is a little naff)

qutip/tests/test_operators.py
```diff
@@ -133,7 +129,7 @@ def test_create():
 @pytest.mark.parametrize("dimension", [1, 5, 100])
 def test_simple_operator_creation(to_test, expected, dimension):
     qobj = to_test(dimension)
-    assert np.allclose(qobj.full(), expected(dimension))
+    np.testing.assert_allclose(qobj.full(), expected(dimension))
```
jakelishman, line 132: Could we not test `qzero` and `qeye` as part of the diagonal checks?

qutip/tests/test_operators.py
```diff
@@ -160,18 +156,20 @@ def test_super_operator_creation(to_test):
...
+@pytest.mark.parametrize(["to_test", "factor", "phase"], [
+    (qutip.position, 1, 1),
+    (qutip.momentum, 1j, -1),
+])
+def test_bidiagonal(to_test, factor, phase):
+    N = 5
+    operator = to_test(N)
+    expected = (np.diag((np.arange(1, N) / 2)**0.5, k=-1) +
+                np.diag((np.arange(1, N) / 2)**0.5, k=1) * phase) * factor
+    np.testing.assert_allclose(operator.full(), expected)
```
jakelishman, line 168: It may be clearer just to test `position` and `momentum` separately and explicitly - it took me a second to work out what was going on, and then it also raises the question why `squeeze` isn't also considered "bidiagonal".

qutip/tests/test_operators.py
```diff
@@ -196,30 +199,23 @@ def test_displace():
...
-    assert_equal(np.allclose(tn.full(), tn_matrix), True) 
-    
-    tn = tunneling(2*N+1,2)
-    tn_matrix = np.diag(np.ones(2*N-1),k=-2) + np.diag(np.ones(2*N-1),k=2)
-    assert_equal(np.allclose(tn.full(), tn_matrix), True)
-
+def test_commutator():
+    A = qutip.qeye(N)
+    B = qutip.destroy(N)
+    assert qutip.commutator(A, B) == qutip.qzero(N)
```
jakelishman, line 217: we should test `commutator(type='anti')` as well.  I'm not sure anyone is actually using the function given how long it is to type compared to how simple it is conceptually, but still!

qutip/tests/test_states.py
```diff
@@ -33,10 +33,13 @@
...
 import numpy as np
-from numpy.testing import assert_, run_module_suite
 import qutip
-from qutip import (expect, destroy, coherent, coherent_dm, thermal_dm,
-                   fock_dm, triplet_states)
+from functools import partial
+from itertools import combinations
+
+
+_2PI = 2 * np.pi
```
jakelishman, line 41: Probably we don't need to define a constant for this

qutip/tests/test_states.py
```diff
@@ -66,63 +69,124 @@ def test_implicit_tensor_projection(size, n, m):
...
-            assert_(abs(triplet.norm() - 1.) < 1e-12)
-
-
-if __name__ == "__main__":
-    run_module_suite()
+@pytest.mark.parametrize("base, operator, args, opargs, eigenval", [
+    pytest.param(qutip.basis, qutip.num, (10, 3), (10,), 3,
+                 id="basis"),
+    pytest.param(qutip.basis, qutip.num, (10, 3, 1), (10, 1), 3,
+                 id="basis_offset"),
```
jakelishman, line 76: it's clearer if you have the id as `basis,offset`, so it doesn't look like we're testing a function `qutip.basis_offset` - by default pytest uses a comma to separate out ids in multi-parametrisation

qutip/tests/test_states.py
```diff
@@ -66,63 +69,124 @@ def test_implicit_tensor_projection(size, n, m):
...
+    alpha = 0.5
+    c1 = qutip.coherent(N, alpha)  # displacement method
+    c2 = qutip.coherent(7, alpha, offset=3)  # analytic method
+    assert abs(qutip.expect(qutip.destroy(N), c1) - alpha) < 1e-10
+    assert (qutip.Qobj(c1[3:]) - c2).norm() < 1e-7
+
+
+def test_TripletStateNorm():
+    for triplet in qutip.triplet_states():
+        assert abs(triplet.norm() - 1.) < 1e-12
```
jakelishman, line 136: There's presumably more we'd like to test about these states, right?  Should we test they're orthonormal, and check their eigenvalues?  Also, the documentation for the function seems to be wrong (although the code isn't) - one of the documentated states is the singlet, not a triplet!!

qutip/tests/test_states.py
```diff
@@ -66,63 +69,124 @@ def test_implicit_tensor_projection(size, n, m):
...
+
+
+def test_ket2dm():
+    N = 5
+    state = qutip.coherent(N, 2)
+    oper = qutip.ket2dm(state)
+    assert np.abs(qutip.expect(oper, state) - 1) < 1e-12
```
jakelishman, line 143: We should probably test more about `ket2dm` as well, e.g. `assert oper.isoper` and `assert oper == state * state.dag()`.  At present, if `ket2dm` just returned `state`, this test would still pass.

qutip/tests/test_states.py
```diff
@@ -66,63 +69,124 @@ def test_implicit_tensor_projection(size, n, m):
...
+        qutip.qstate("dudd") +
+        qutip.qstate("ddud") +
+        qutip.qstate("dddu")
+    ) / 2
+    assert state == qutip.w_state(4)
+
+
+def test_ghz_states():
+    state = (qutip.qstate("uuu") + qutip.qstate("ddd")) * 0.5**0.5
+    assert state == qutip.ghz_state(3)
```
jakelishman, line 174: Probably clearer just to do `state = ( ... ).norm()`

jakelishman (commented):
Nice, always good to improve our coverage!

Looking at the coverage report, seems like the only lines we're missing in states and operators are failure paths now.  Could you add a couple of extra tests to hit those paths as well, to get the files up to 100% coverage?

qutip/tests/test_operators.py
hodgestar: We could get something similar for tests using https://pypi.org/project/pytest-testdox/ if we wanted.

qutip/tests/test_states.py
```diff
@@ -33,10 +33,10 @@
 
 import pytest
 import numpy as np
-from numpy.testing import assert_, run_module_suite
 import qutip
-from qutip import (expect, destroy, coherent, coherent_dm, thermal_dm,
-                   fock_dm, triplet_states)
+from functools import partial
+from itertools import combinations
+N = 5
```
hodgestar, line 39: Is the global `N` actually used anywhere in this file? Even if the answer is yes, I'm not wild about having this global size that is shadowed by numerous local variables and pytest parameters. Could we remove it or at least give it a more specific name and perhaps move it closer to where it's used and make it specific to a specific test? Also happy with various other solutions to this.

qutip/tests/test_states.py
```diff
@@ -66,63 +91,182 @@ def test_implicit_tensor_projection(size, n, m):
...
+    N = 10
+    alpha = 0.5
+    c1 = qutip.coherent(N, alpha)  # displacement method
+    c2 = qutip.coherent(7, alpha, offset=3)  # analytic method
+    c3 = qutip.coherent(N, alpha, offset=0, method="analytic")
+    np.testing.assert_allclose(c1.full()[3:], c2.full(), atol=1e-7)
+    np.testing.assert_allclose(c1.full(), c3.full(), atol=1e-7)
+    with pytest.raises(TypeError) as e:
+        qutip.coherent(7, method="other")
+        assert e.message == ("The method option can only take "
```
hodgestar, line 158: Another `assert` that needs to be moved outside the `with pytest.raises`.

qutip/tests/test_operators.py
```diff
@@ -79,52 +93,60 @@ def test_jmat_32():
...
+def test_jmat_dims(spin, N):
+    spin_mat = qutip.jmat(spin, '+')
+    assert spin_mat.dims == [[N], [N]]
+    assert spin_mat.shape == (N, N)
+
+
+def test_jmat_raise():
+    with pytest.raises(TypeError) as e:
+        qutip.jmat(0.25)
+        assert e.message == 'j must be a non-negative integer or half-integer'
```
hodgestar, line 115: And here.

qutip/tests/test_operators.py
```diff
@@ -79,52 +93,60 @@ def test_jmat_32():
...
+    with pytest.raises(TypeError) as e:
+        qutip.jmat(0.5, 'zx+')
+        assert e.message == 'Invalid type'
```
hodgestar, line 118: And here.

qutip/tests/test_operators.py
```diff
@@ -196,30 +227,43 @@ def test_displace():
...
+    B = qutip.destroy(N)
+    assert qutip.commutator(A, B, 'anti') == qutip.destroy(N) * 2
+
+    sx = qutip.sigmax()
+    sy = qutip.sigmay()
+    assert qutip.commutator(sx, sy, 'anti') == qutip.qzero(2)
+
+    with pytest.raises(TypeError) as e:
+        qutip.commutator(sx, sy, 'something')
+        assert e.message.startswith("Unknown commutator kind")
```
hodgestar, line 261: And here.

hodgestar (changes requested):
Thank you for adding more tests!

Apologies for leaving lots of comments about moving the `asserts` outside of the `with pytest.raises` block -- I stopped leaving comments after awhile, but we should fix all of them.

hodgestar (approved):
Approving.

I'm still not wild about the following things, but I will leave these battles for another day :)

* Heavily parameterized tests where there parameters are `func_to_test`, `args`, `expected` still look like trouble to me. Specifically unit tests should be clearly divided into "these are the tests for this unit of functionality". Writing tests for ten units in three lines of code plus 15 lines of decorators looks fancy, but makes it much harder to answer "what functionality of each unit is directly tested" and makes it much harder to spot errors in tests. The ideal test is dead simple so that it's obvious that its correct and tests everything it should.

* I'm still not wild about the global `N` that is overridden somewhat randomly in tests. It's a small thing, but inconsistent use of and shadowing of global variables feels messy.

Anyway, not letting either of these stand in the way of merging.

PR #1579, opened 2021-06-18
Update licence information in QuTiP

jakelishman:
This updates the licence holders to be the "QuTiP Admin Team" (this is similar to how Numpy handle their licensing).  This also updates the text of the licence file to a GitHub-readable version of the 3-clause BSD licence, so it should present better information to users as well.  At present, erroneous whitespace in the licence file means that GitHub's detection can't tell what exactly the licence is.

In addition, having the text of the licence in each file is redundant, and just made it more annoying to edit the files.  It also made it much harder to update the information in the licence in order to put the whole admin team as copyright holders, so I have removed it.  Again, this is in keeping with how other large-scale projects manage their licensing.

We always distribute `LICENSE.txt` with all copies of the QuTiP source (either by clone of the repository, or by sdist).

**Changelog**: update licence holders with regards to new governance model, and remove extraneous licensing information from source files.


jakelishman:
Note: the reported changes (other than the deletion of the licence from each source file) are all whitespace-only, or simply a reflow of the hard-wrapped text in `LICENSE.txt`.  I didn't change the licence at all, other than the holder.  I've also checked this with the same tool that GitHub uses, and it will definitely recognise the licence correctly after this.

The whitespace-only changes in other files weren't actually 100% intended, but it was a function of how I did the licence detection - I normalised the whitespace at the ends of lines first.  Looking through, I guess I only did that when changing the `doc` folder, not the main source code (I had to handle those two cases separately).

Ericgig:
Can this be done legally?
Is it acceptable to remove the copyright in the files?
We would be changing the owner of the code.

In #1405, it has to have the Nasa copyright by agency policy:
```
Copyright 2020 United States Government as represented by the Administrator
of the National Aeronautics and Space Administration. All Rights Reserved
```
I think we could have legal issues if we removed this without their permission.

While I agree with removing the ~30 lines header with the license in each file, we should keep the copyright lines unless we can confirm it safe to do so. Luckily, it is mostly Paul, Robert and Alex that have specified it and if they agree then I guess it fine. But if any other contributor added their copyright, we should keep it.

jakelishman:
This can - see e.g. Numpy that uses the same licence.  #1405 would add its additional copyright to the top of the file.  This is actually _more_ useful in that case - it's far clearer that that code is under different licence conditions.

That said, I'm really not certain about accepting #1405 in its current form.  It's not using the data-layer correctly at all, and if we were to include it, we'd really want to do something like that in a more structured way in core QuTiP.  Really it should be a properly dispatched implementation of `einsum`.

I'm also very wary of introducing code to QuTiP which is licensed by a different holder.  I can only imagine that that will really give us problems down the line - how long would NASA own the copyright on that code, for example?  Could we modify it?  After how many modifications is it no longer NASA's?  We wouldn't have control over anything like that, and I think it's probably more hassle than it's worth on our side.

jakelishman:
Also, while we should check with Paul, Rob and Alex just for completeness, my understanding was that they had already accepted the governance document, which essentially lays this out as well.  This PR is more about bringing the actual text of the licence in QuTiP into equality with how it's actually owned.

It's worth checking, but I'm fairly sure that there are no other licence holders listed in any of the files other than the current QuTiP admin team, and Rob and Paul.

jakelishman:
In general about the licensing concerns of #1405: really, I think we should be requiring that anybody contributing code to QuTiP has to sign over the rights to the QuTiP development team.  In fact, my understanding is that this is [part of the GitHub terms of service](https://docs.github.com/en/github/site-policy/github-terms-of-service#6-contributions-under-repository-license), and that we _cannot_ accept #1405 under a different licence holder.

nonhermitian:
The NASA copyright needs to be there.  As it stands now, it is not clear what the license actually means though, i.e. it is embedded in the QuTiP license which does not make any sense.  So that needs to be cleaned up before approval. However, if it is truly BSD licensed then the copyright does not matter for derivative works as it gives permission to do whatever one wants provided the original copyright remains.

I am personally fine with the copyright changes going forward.  I am not part of the admin team (by choice), but there is a decade of git commits that show my share the the work, and making the cp more broadly applied to QuTiP admin / devs make s a lot of sense going forward.

Ericgig:
If there is no other license holders, then it should be fine, but I would ask a review from Robert.

jakelishman:
@nonhermitian: we haven't accept #1405 (and also I don't think we will soon, because it'll be part of larger changes), but yeah, if it is accepted, it would need to carry its entire licence with it in the file.  We already do this somewhere with a vendored version of `Qcircuit.tex` though that's a whole other can of worms, because that's GPL'd, and I'm really not sure it's legally correct that we've done that.  (also sidenote: thanks to you, I'm starting at IBM in Lev Bishop's team on the 5th of July - I've just put in the requisition for my computer!)

On second thoughts, surely the more correct solution is to have the main licence read "Copyright 2011-2021 Rob and Paul, 2021 onwards admin team"?  That's more historically accurate, for sure.  Another question: does anybody know about the legality of asserting copyright "2011 and onwards" in general?  Numpy put in specific years, and bump the end year by commit.

I'm mostly in favour of moving the licencing to the single `LICENSE.txt` file in order to make all these instances of separately licenced code easier to see and maintain.  It also seems to be the way other large-scale open-source projects handle their licensing, it makes it much easier for us to manage the licensing of the entire project.

I really don't want to rush this.  I don't want to merge it without a proper discussion in a full admin meeting.  I'm just trying to put it up ahead of time.

nonhermitian:
> though that's a whole other can of worms, because that's GPL'd

Yes, if it is GPL then it should be removed.  That license can be construed to mean that all other pieces of code that use it must be under the same license; One can read the license as all derivative works must be GPL.  So I would revert that PR ASAP.

> legality of asserting copyright "2011 and onwards" in general

Yeah putting years I think is the correct thing.  Didn't know that at the time.

>  I'm starting at IBM in Lev Bishop's team on the 5th of July

Look forward to talking to you in July!

jakelishman:
The `Qcircuit.tex` PR was unfortunately many many years ago, long before I made my first commit to QuTiP (according to `git log` it was by Anubhav Vardhan in 2014), so the damage is already done.  It was on my mind because there was a PR about updating the version of the code in it (#1187).  Nonetheless, we definitely should remove it. @BoxiLi Boxi: `Qcircuit.tex` is a concern for `qutip-qip` more than the main library from 5.0 onwards - it's currently in `qip/circuit_latex.py`, which we'll have removed, but you've gained.

It might be most appropriate to handle this the same way Qiskit does - generate code which does `\usepackage{qcircuit}`, and leave it up to the user to have a properly-provisioned LaTeX environment.  I note that `qcircuit` is included in MikTeX and TexLive since at least 2018 (but not sure when exactly it was added).


> Yeah putting years I think is the correct thing. Didn't know that at the time.

Ah, that's good to know, thanks.  I'm trying to learn more about licensing - I'm trying to get it sorted now into the right form, but it's a really complicated field.  I think if we change `LICENCE.txt` to

```text
Copyright (c) 2011--2021 Paul Nation and Robert Johansson.
Copyright (c) 2021--2021 QuTiP Admin Team.
All rights reserved.

<BSD 3-clause licence>
```

and update the final `2021` as appropriate, we'll be in the best standing.  That's also why we might want to remove the general licence text from each file - it's already redundant, but it'll be much harder to update if we have it scattered everywhere (automation would run the risk of us messing up and changing any external code's licence as well, I think, which would be wrong).

> Look forward to talking to you in July!

Me too!

nathanshammah (changes requested):
The copyright is owned by all of the contributors, notwithstanding the boilerplate. 
I'd split this PR in two PRs: 
1. "Format license.txt to the standard BSD v3.0 template". This way it will also be picked up by Github's metadata. 
2. "Uniform copyright boilerplate in files": with this regard, it would best read `Copyright (c) 2011 and later, Paul D. Nation, Robert J. Johansson and the QuTiP contributors`. It would be best to keep this in all files of the repository. 

I've discussed this in the past with @vanl, but I am happy to arrange an open meeting if more Q&As are needed, this could be done on Unitary Fund Discord server, if it helps. It may be nice, alongside this asynchronous discussion (thanks Jake) to discuss it at a QuTiP admins meeting.

jakelishman:
@nathanshammah: already on the agenda.  The splitting seems like a good plan.

I would have thought that part of making it uniform would be to remove the lines from each file, so it can be centrally managed.  That way, only files that have different licence terms need to note it.  This is the way that Numpy, Scipy, CuPy, pandas, scikit-learn, etc etc all handle licensing - none of those specify copyright information separately, and they all use the exact same licence as QuTiP (except CuPy, which is MIT).  If we were GPL'd (like mitiq), we'd have to include the licence, as I understand it, but 3-clause BSD only requires you include the licence file itself.

jakelishman:
Also, from further looking around, everyone else uses the string "the <project> development team" or "<project> developers" as the copyright owner from 2021 onwards, and leave it as Rob and Paul only until then, because I don't think we can re-licence the past.  I'm generally assuming that Numpy and SciPy in particular have got their licensing done properly, and since they use the exact same licence as us, they seem to be the right model to follow.

BoxiLi:
@jakelishman

> Qcircuit.tex is a concern for qutip-qip more than the main library from 5.0 onwards - it's currently in qip/circuit_latex.py, which we'll have removed, but you've gained.

Yes, this is the reason why that PR in `qutip` wasn't merged. I saw you already closed it. Make sense. I only found this out after checking the new `qcircuit` package. Indeed, it was added way earlier, in 2014 as you found out.

Thanks for the fix!

nathanshammah:
@jakelishman the reason why I suggested having Paul, Rob and the qutip contributors/developers is that the copyright is regulated by the license, and how it is implies already that every contributor who got a PR merged owns copy rights. So the reason to keep Rob and Paul for me was just aesthetic, the alternative being 2011 onwards qutip contributors/developers.

jakelishman:
@nathanshammah, @Ericgig, @ajgpitch, @nwlambert, @quantshah, @BoxiLi, @hodgestar

I've updated this PR to match what I understand we agreed on.

I've also added complete copyright notices to the documentation, including linking the CC-BY-3.0 licence the documentation text is licensed under, and making it clear that it's the documentation _text_ licensed under CC, and that the code is under 3-clause BSD.  I've updated the frontmatter, and added a copyright notice to the index, so it's linked from every page in the documentation releases.  I have updated the short copyright notice at the bottom of the documentation pages to say "Copyright 2011 to 2021 inclusive, QuTiP developers and contributors" to make it consistent with the source code (for the same reasons).  As I've added a `_copyright` Sphinx target, the word "Copyright" at the bottom of the documentation is now hyperlinked as well.

qutip/control/cy_grape.pyx
```diff
@@ -1,36 +1,3 @@
-#cython: language_level=3
```
hodgestar, line 1: We should probably keep these `cython: language_level` lines? Or at least they are not part of the license boilerplate.

qutip/control/dump.py
```diff
@@ -1,37 +1,3 @@
-# -*- coding: utf-8 -*-
```
hodgestar, line 1: We should keep these encoding lines.

qutip/cy/brtools.pxd
```diff
@@ -1,37 +1,3 @@
-#!python
```
hodgestar, line 1: This `#!` line in a .pxd file seems pretty strange, so I guess removing it is a good idea!

qutip/cy/cqobjevo.pyx
```diff
@@ -1,37 +1,3 @@
-#cython: language_level=3
-# distutils: language = c++
```
hodgestar, line 2: This `distutils: language = c++` is also pretty odd and can hopefully go too.

qutip/cy/mcsolve.pyx
```diff
@@ -1,39 +1,3 @@
-#cython: language_level=3
-## cython: profile=True
-## cython: linetrace=True
```
hodgestar, line 3: I'm guessing both of these profile and linetrace lines should be removed and will perhaps make mcsolve slightly faster?

hodgestar (changes requested):
Thanks @jakelishman! This matches my understanding of what we all agreed on in the admin meeting. Thank you for also sorting out the docs licensing.

I left comments on a bunch of changes that I thought were slightly out of scope for this PR, although some of the changes I think can happen here anyway -- just wanted to flag this for everyone to consider. I didn't flag all of them -- just one example of each kind of oddity.

qutip/cy/mcsolve.pyx
jakelishman: If they'd been enabled, you'd _know_ about it - linetrace makes inner-loop nogil code sloooooooow!

qutip/cy/cqobjevo.pyx
jakelishman: It's forced by our `setup.py` anyway, so no big deal.

jakelishman:
@hodgestar: those comments lines you've highlighted were a careless mistake - they shouldn't have been removed, but it's a function of how I implemented the automated removal.  I didn't notice them when I checked it through.  I've restored them.  The `cython: language_level=3` lines in particular are especially important to maintain, so that was very careless of me.

That said, utf-8 is the default encoding for Python 3, so those lines aren't actually necessary at all any more.  That wasn't the case with Python 2 (which assumed ASCII), but it's no longer an issue.  Regardless, I've restored them because this PR wasn't meant to remove that sort of thing.  Similarly, since I had to go through everything anyway, I've excluded the whitespace-only changes within files (though I left the normalisation of whitespace specifically at the ends of files).

Apologies, I've force-pushed it back rather than applying it as a new commit, so it probably won't render these changes as nicely in the GitHub interface - it was rather easier for me to modify the commit than to make a "partial reversion" commit (and it's better for the logs).

qutip/control/cy_grape.pyx
jakelishman: I've updated the PR to skip these removals.  Not sure why these lines haven't gained the "outdated" tag after the push (at least for me), because they should be entirely gone.

qutip/control/dump.py
jakelishman: Noted in the main response, but these encoding lines are an artifact of Python 2, and not actually required any more.  Regardless, I've left them in because they aren't in scope of this PR.

qutip/cy/brtools.pxd
jakelishman: Technically outside scope, but I left this change in anyway because this is definitely a mistake - the file's not even marked as executable, so it could never have had an effect anyway.

hodgestar (approved):

qutip/cy/openmp/cqobjevo_omp.pyx
```diff
@@ -1,37 +1,5 @@
 #cython: language_level=3
-# distutils: language = c++
```
BoxiLi, line 2: Do we need to keep this or not?

hodgestar:
@jakelishman Thanks for cleaning things up. Glad one of the odd changes I spotted was useful to sort out!

qutip/cy/openmp/cqobjevo_omp.pyx
jakelishman: It has no effect, because we override it at the `setup.py` level, so I just left it as removed.

jakelishman:
@nathanshammah: are you happy that I've not made a mistake in doing what we agreed on in the group meeting?  If you give this your tick, I'll merge it.

nathanshammah (approved):
Thanks @jakelishman, LGTM. I guess we'll have to remember yearly to update "to 2021 inclusive" to avoid making things look old from next year.

jakelishman:
@nathanshammah: by my understanding, technically we're only meant to include years in which we make a new release of the code/documention with modifications from a previous release.  I went and checked, and we did it for all of 2011, 2012, ..., 2021, and I assume we'll be doing it for some time beyond that as well, so we should be good.  In this new format, it's also much easier to update - there's only something like 3 places in the entire codebase with the years now.

PR #1580, opened 2021-06-18
Remove vendored Qcircuit.tex

jakelishman:
This file included GPL'd code, which is incompatible with QuTiP's 3-clause BSD licence.  Instead, we rely on the user to have a functional install of the file (it's been included in TeXLive since at least 2014).

See qutip/qutip#1579

**Changelog**: removed vendored version of LaTeX package `qcircuit`

### DIFF
--- a/qutip/qip/circuit_latex.py
+++ b/qutip/qip/circuit_latex.py
@@ -41,7 +41,8 @@ import warnings
 
 _latex_template = r"""
 \documentclass{standalone}
-%s
+\usepackage[braket]{qcircuit}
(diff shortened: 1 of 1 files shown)


jakelishman:
Look at me go, fixing CodeClimate issues... I'm so altruistic.

BoxiLi (approved):
Tested using the Jupyter Notebook on https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/quantum-gates.ipynb.

LGTM. I shall merge if @purva-thakre find out the same on Ubuntu :)

PR #1583, opened 2021-06-20
Faster Husimi Q with saved state

jakelishman:
This is a tidied up and completed version of #934.  The first commit is credited to the original author (although I fixed up some little concerns in the PR and rebased it), because it's logically built off that PR, although in the end it's a complete reimplementation.

This is a complete rewrite of the Husimi Q calculations, to make a more formal split between the iterative (low-memory) and precomputed (faster) algorithms, giving full access to both of them.  It's a bit easier to use from a user's perspective, and it's faster than #934 as well.

The concept of the precomputed matrix is encapsulated into a class with a cached tensor, so that the same precomputed system can be used for many states of different sizes, and autoexpands (up to the memory limit) to accommodate larger states.  This is called `QFunc` as opposed to the standard function interface `qfunc`.

It also adds rather more error checking on its inputs, and adds a complete testing suite for the new implementation.

Closes #934

## Examples

Let's say we have some states, and the phase-space coordinates we want to calculate the Husimi Q function at.
```python
>>> import qutip
>>> states = [qutip.rand_dm(32, density=0.2) for _ in [None]*100]
>>> xs = np.linspace(-2, 2, 401)
```

Now the normal way of calculating the Q distribution for a single state is
```python
>>> qutip.qfunc(states[0], xs, xs)
array( ... )
```

This already has some speed advantages over the pre-PR version of `qfunc`, because it caches a lot of its intermediary results, to avoid recomputing them.  This results in calculating more matrix-vector products than before, but far fewr FLOPS overall.  By default, `qfunc` issues a warning if it exceeds a certain amount of memory, and falls back to the low-memory version.  You can control this limit with the `precompute_memory` option:
```python
>>> qutip.qfunc(states[0], xs, xs, precompute_memory=0)
qutip/qutip/wigner.py:822: UserWarning: Falling back to iterative algorithm due to lack of memory.
Needed 78.52 MB, but only allowed to use 0.00 MB.  Increase `precompute_memory` to raise limit, or
set to `None` to suppress warning.
array( ... )
```
To suppress the warning, and always use the iterative approach, you can set `precompute_memory=None`.

Now, even if you're using the `precompute_memory` option, `qfunc` still can't remember its state between calls.  This means it's inefficient.  Instead, we can use the `QFunc` class, which adds an extra step (set-up and call as opposed to just call), but it encapsulates the state.
```python
>>> qfunc_calculator = qutip.QFunc(xs, xs)
>>> husimiqs = [qfunc_calculator(state) for state in states]
```
This is much much faster.  The class version `QFunc` also takes a memory option, but since this is designed to be the precompute version, it errors out if the amount of memory needed is too great, rather than trying to fall back.  This gives users a way to put a limit on the amount of memory used:
```python
>>> qfunc_calculator_low_memory = qutip.QFunc(xs, xs, memory=0)
>>> qfunc_calculator_low_memory(states[0])
MemoryError: Refusing to precompute up to 32 basis states.  This would require 78.52 MB, but only 0 MB is allowed.
```

## Timings

These benchmarks were taken on a Macbook Pro 2.9GHz i5 (2015) with 8GB of RAM.  The different tests had different numbers of phase-space coordinates (the first number in the first column) in each dimension, and the number after 'ket' or 'dm' is the Hilbert space dimension.  The columns are:

- `original`: The state of the code before this and the previous commit
- `prev def`: Direct calls of `qutip.qfunc` with the given state and phase-space coordinates, but everything else at the default for #934
- `prev pre`: Calls of `qutip.qfunc(..., precompute=pre)`, where the calculation of `pre` was done outside the timing.  This times how long it would take to call in a loop many times.
- `cur def`: Direct calls of `qutip.qfunc` with the given state and phase-space coordinates, but everything else at the default for this commit.
- `cur pre`: Calls to a pre-constructed `qutip.QFunc` instance (which was excluded from the timings) for this commit.  This times how long it would take to call in a loop many times.

```text
            │ original  │ prev def  │ prev pre  │  cur def  │  cur pre
────────────┼───────────┼───────────┼───────────┼───────────┼──────────
11,  ket 5  │ 126.07 µs │ 133.02 µs │  89.63 µs │ 108.95 µs │  25.12 µs
11,  ket 32 │ 192.49 µs │ 197.88 µs │  90.91 µs │ 172.13 µs │  29.33 µs
11,  dm 5   │ 481.68 µs │ 440.45 µs │ 316.93 µs │ 304.86 µs │ 161.00 µs
11,  dm 32  │   4.69 ms │   1.87 ms │   1.66 ms │   1.07 ms │ 823.49 µs
            │           │           │           │           │
101, ket 5  │ 570.59 µs │ 566.65 µs │ 501.86 µs │ 463.96 µs │ 329.25 µs
101, ket 32 │   1.61 ms │   1.43 ms │ 751.97 µs │   1.34 ms │ 507.20 µs
101, dm 5   │   2.51 ms │   3.51 ms │   2.20 ms │   2.62 ms │   1.72 ms
101, dm 32  │  42.91 ms │  26.84 ms │  19.73 ms │  23.30 ms │  16.30 ms
            │           │           │           │           │
401, ket 5  │  10.56 ms │  13.50 ms │   7.91 ms │  11.95 ms │   5.28 ms
401, ket 32 │  34.44 ms │  43.73 ms │  10.37 ms │  41.17 ms │   7.70 ms
401, dm 5   │  46.85 ms │  57.51 ms │  33.19 ms │  52.30 ms │  26.71 ms
401, dm 32  │   1.05  s │ 411.32 ms │ 273.93 ms │ 428.32 ms │ 246.59 ms
```

You can see that there are huge benefits to the precomputation, even compared to the previous precomputation implementation.  This makes calling in a loop much much more efficient.

#### Changelog

Large feature, authored by @dweigand (Daniel Weigand) and me (Jake).

Add new Husimi Q algorithms, improving the speed for density matrices, and giving a near order-of-magnitude improvement when calculating the Q function for many different states (using the new `qutip.QFunc` class, instead of the `qutip.qfunc` function).

### DIFF
--- a/qutip/wigner.py
+++ b/qutip/wigner.py
@@ -1,27 +1,21 @@
-__all__ = ['wigner', 'qfunc', 'spin_q_function',
+__all__ = ['wigner', 'qfunc', 'QFunc', 'spin_q_function',
            'spin_wigner', 'wigner_transform']
 
 import numpy as np
 import warnings
 from numpy import (
-    zeros, array, arange, exp, real, conj, pi, copy, sqrt, meshgrid, size,
-    conjugate, cos, sin, polyval, fliplr,
+    zeros, array, arange, exp, real, conj, pi, copy, sqrt, meshgrid, cos, sin,
 )
 import scipy.sparse as sp
 import scipy.fftpack as ft
 import scipy.linalg as la
-from scipy.special import genlaguerre
-from scipy.special import binom
-from scipy.special import sph_harm
+import scipy.special
+from scipy.special import genlaguerre, binom, sph_harm, factorial
 
-from qutip.qobj import Qobj, isket, isoper
-from qutip.states import ket2dm
+import qutip
+from qutip import Qobj, ket2dm, jmat
 from qutip.parallel import parfor
-from qutip.utilities import clebsch
-from qutip.operators import jmat
-from scipy.special import factorial
 from qutip.cy.sparse_utils import _csr_get_diag
-import qutip as qt
 from qutip.sparse import eigh
 
 
@@ -114,7 +108,7 @@ def _parity(N, j):
         foo = np.ones(mult)
         for n in np.arange(-j, j + 1, 1):
             for l in np.arange(0, mult, 1):
-                foo[l] = (2 * l + 1) * qt.clebsch(j, l, j, n, 0, n)
+                foo[l] = (2 * l + 1) * qutip.clebsch(j, l, j, n, 0, n)
             matrix[np.int32(n + j), np.int32(n + j)] = np.sum(foo)
         return matrix / mult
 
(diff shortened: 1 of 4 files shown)


jakelishman:
(Also apparently I tidied a couple of imports and docstrings in `wigner.py` as well, which I forgot...)

Ericgig (approved):
Look good.
I confused `_qfunc_alpha_matrix.conj` with the `conj()` method at first.

jakelishman:
@Ericgig: yeah, it's not a perfect naming choice.  Can you think of a better one?  It ended up like that because I started off just calling in `_qfunc_alpha_matrix`, then realised about halfway through that I was actually constructing the conjugate of the matrix formed by stacking coherent states (since that's how it's used), but the name isn't really ideal.

@quantshah: I'm just thinking now you might want to have a look at this - you had some thoughts about user interfaces for class-based creation schemes last year, with regards to the displacement operator, which I'll build up into a proper PR at some point as well.  Don't worry if you've no time, though.

quantshah:
Yes I will have a look this weekend.

Ericgig:
I am not great with names myself, but I would go with `grid`, or `mesh`.

quantshah:
@jakelishman this is great, also thanks @dweigand for this PR. I tried out the function and it is very fast and could be great for specific use cases. My idea was a bit more simple - precomputing the `coherent_dm` operators (using qutip.coherent_dm) and using the `expect` function. But that performs worse that the original `qfunc` as well as this implementation of course. So this is great. I would be happy to merge this but have one thought:

This new implementation is sacrificing the ease-of-understanding of how Q-functions are computed with a very efficient and fast, but complicated class-based implementation. This means that users who want to look under the hood will not fully understand how this is working. I believe the crux of this implementation is a fast computation of the tensors of `coherent_dm` operators for all the alpha values at (x + iy).

Personally, on many occasions, I have looked at qutip implementations of quantum operations just to understand what is happening under the hood and I believe several users have often done this to hack around. If someone were to try and understand Q-function computation from this code, it seems simple enough with the call:

```
np.abs(np.dot(alphas, (self._g*0.5)*vector))**2
```

for state vectors, or decomposing a density matrix into state vectors and then running the above call. Or 

```
out = np.polyval(
        (0.5*g * vector / np.sqrt(scipy.special.factorial(ns)))[::-1],
        alpha.conj,
    )
```

But the object `alpha` now constructed with `_qfunc_alpha_matrix` class is a bit difficult to understand. The variable `self.conj = np.empty(x.shape, dtype=np.complex128)` creates a tensor and sets it to a mesh grid (x, -y), but then what does the `__call__()` return? The tensor of `coherent_dm(x+iy)` for each x and y coordinate? But how is it computing it without knowing the Hilbert space dim for the state, which I think is computed on the fly by `ns = np.arange(first, last).reshape(1, 1, -1)`. It seems it caches the `coherent_dm` computed upto `current_size` and then extends it to `size` for new Hilbert space dimensions. 

I am sorry if I am being a bit thick here because I understand the overall idea of the code but somehow I am not confident it will be easy for others to understand on first read.

Do you think we can make it a bit more transparent how the `_qfunc_alpha_matrix` works and that the crux of the method is computing the operators `coherent_dm` in an efficient way and saving them? If I have understood the code correct, I could try to document it a bit more and simplify it. 

Let me know your thoughts and correct me if I misunderstood something @jakelishman @Ericgig

jakelishman:
Well, the default implementation for states is actually nearly completely unchanged (the only difference is that I replaced `np.fliplr(array)` with `array[::-1]`, which I think is a more readable idiom for "backwards").  You're right that in the new form, basically the method is that you cache various parts of the `np.polyval` call into a 3D tensor, rather than recomputing this large polynomial each time.  The `np.polyval` call is how you actually go from an array of different values of `alpha` (which comes from the meshgrid) to the coherent state, but doing it with `polyval` means you construct the coherent state and do the dot product simultaneously to save memory.  If you need to repeat the calculations a bunch, and you've got enough memory to spare, you can build up all the `alpha**n / sqrt(factorial(n))` tensor once, then just dot it against every state.  Since we're dealing with _complex_ Hilbert spaces, the inner product isn't a straight dot product, it needs the conjugate of one of the sides, so we only store the conjugate of the alpha tensor.

You can see in the timings that this isn't _always_ the biggest win you'd expect; in some cases, the additional memory usage causes you to be more heavily bound by memory bandwidth issues rather than CPU pipelined stuff, which is another reason we allow forcing either method to be used.

If you're interested in the most readable implementation, you might want to look at the test:
https://github.com/qutip/qutip/blob/a20c26307e2f6e738f87efd75abafc3dff4daad5/qutip/tests/test_wigner.py#L251-L260

jakelishman:
Oh, also to answer the question more explicitly about `_qfunc_alpha_matrix.__call__`: _that_ function always re-computes the matrix of alphas, returning the relevant bits of the Fock space.  The caching behaviour is in the `QFunc` class itself, which manages how many Fock states it needs (see `QFunc._alphas`, which tries to pull the matrix from a cache, or expands the cache if necessary).  `QFunc.__call__` calculates the Q function of any quantum state over the given phase-space coordinates, no matter how large the size of Hilbert space of the state is - that means it's got feature parity with `qfunc` in the sense that both
```python
iterative = functools.partial(qutip.qfunc, xvec=xs, yvec=ys)
classbased = qutip.QFunc(xs, ys)
```
can be called with the same objects, and will return the same result, even if the objects change size in between.  I see this as sensible functionality in general - if you're plotting a few graphs, it's quite possible that you haven't calculated all your states with the same amount of Fock-space truncation (I frequently scaled the calculation sizes in my PhD for speed), so it's not very useful if the class-based version can only handle fixed Hilbert spaces.

quantshah:
Thanks Jake, I understand better how the cache is working. Okay let me suggest a few comments within the code so that we (or at least I) can keep track of how this implementation works. If you think they are sensible then feel free to accept and then we can merge this.

qutip/wigner.py
```diff
@@ -548,83 +548,297 @@ def _wig_laguerre_val(L, x, c):
...
+    if xs.ndim != 1 or ys.ndim != 1:
+        raise ValueError(
+            f"xs and ys must be 1D, but have shapes {xs.shape} and {ys.shape}."
+        )
+    return xs, ys
+
+
+class _qfunc_alpha_matrix:
+    """
+    Internal class used to centrally manage the creation of tensors related to
```
quantshah, line 593: How about adding this simple example to show what this function is doing:

```
"""
Internal class to compute coherent states. A call to an object of this class gives the same 
results as `qutip.coherent` as seen from this example.

Example
--------
>> N = 64 # Hilbert size
>> xs, ys, g = [1, 0, -1], [1, 0, -1], 1 # Grid

>> X, Y = np.meshgrid(xs, ys)
>> betas = 0.5*g*(X - 1j*Y).ravel() # Complex numbers denoting phase space points

>> # QuTiP's implementation of computing coherent states
>> coherent_states_qutip = np.array([coherent(N, beta).full() 
                                  for beta in betas]).reshape(3, 3, -1)

>> # Function to get coherent state vector coefficients for
>> # different Fock numbers within the #grid

>> get_coherent = _qfunc_alpha_matrix(xs, ys, g=g) 

>> # Coefficients for the Fock numbers between 0, N
coherent_states_with_alpha_matrix = get_coherent(0, N)

>> # Test that the implementations match
>> np.testing.assert_array_almost_equal(coherent_states_with_alpha_matrix,
                                    coherent_states_qutip)

>> np.testing.assert_array_almost_equal(get_coherent(1, 6), coherent_states_qutip[:, :, 1:6])
```

qutip/wigner.py
```diff
@@ -548,83 +548,297 @@ def _wig_laguerre_val(L, x, c):
...
+            return self._cache[:, :, :size]
+        if size > self._max_size:
+            requirement = self._size_mb * size
+            raise MemoryError(
+                f"Refusing to precompute up to {size} basis states."
+                f" This would require {requirement:.2f} MB,"
+                f" but only {self._memory_mb} MB is allowed."
+            )
+        if self._cache is None:
+            self._cache = self._alpha_matrix(self._current_size, size)
```
quantshah, line 710: Could we add a comment here saying self._cache is a [n, m, k] sized numpy array of complex numbers where (n, m) represents the phase space grid and k represents the Hilbert space dimensions for the coherent state.

qutip/wigner.py
```diff
@@ -548,83 +548,297 @@ def _wig_laguerre_val(L, x, c):
...
+
+    def __call__(self, state: Qobj):
+        """
+        Get the Husimi-Q function for the given state vector or density matrix,
+        over the coordinates used to initialise the class.  If called multiple
+        times, the states do not need to have the same dimensions, but none of
+        them can have tensor-product structure.
+        """
+        state = _qfunc_check_state(state)
+        alphas = self._alphas(state.shape[0])
```
quantshah, line 734: Something similar for the variable alphas? . 

# alphas represents a [n, m, k] sized numpy array of complex numbers with (n, m)
# representing the phase space grid and k representing the Hilbert space
# dimensions for the coherent state.

qutip/wigner.py
```diff
@@ -548,83 +548,297 @@ def _wig_laguerre_val(L, x, c):
...
+):
+    r"""
+    Get the Q function (without the :math:`\pi` scaling factor) of a single
+    state vector, using the iterative algorithm which recomputes the powers of
+    the coherent-state matrix.
+    """
+    ns = np.arange(vector.shape[0])
+    out = np.polyval(
+        (0.5*g * vector / np.sqrt(scipy.special.factorial(ns)))[::-1],
+        alpha.conj,
```
quantshah, line 760: A comment here saying:

"""
alpha is an instance of the `_qfunc_alpha_matrix` with alpha.conj specifying the grid of 
complex numbers representing points in the phase space.
"""

It was a bit confusing to use `alpha` since I am expecting something like a complex number here

qutip/wigner.py
```diff
@@ -548,83 +548,297 @@ def _wig_laguerre_val(L, x, c):
...
+    if state.isoper and enough_memory:
+        return QFunc(xs, ys, g)(state)
+    if precompute_memory is not None and state.isoper:
+        warnings.warn(
+            "Falling back to iterative algorithm due to lack of memory."
+            f" Needed {required_memory:.2f} MB, but only allowed to use"
+            f" {precompute_memory:.2f} MB.  Increase `precompute_memory` to"
+            " raise limit, or set to `None` to suppress warning."
+        )
+    alpha = _qfunc_alpha_matrix(xs, ys, g)
```
quantshah, line 828: A comment here:

"""
Calls to `alpha(a: int, b: int) computes a [n, m, a:b] dimensional array
of complex coefficients for coherent state vectors at points on the
phase-space grid (n, m) for Fock numbers between a:b.
"""

quantshah:
@jakelishman Thanks for this really great PR. I have just a few comments to add to keep track of the private class `_alpha_matrix` and how it behaves. No changes to the code so if you think these comments are reasonable and can help to clarify how it all works for future maintenance, then go ahead and add them and we can merge this.

quantshah:
I can actually add these comments myself so no worries @jakelishman. I will do it and merge this.

jakelishman:
Ok, no problem!  When you do, it might be best to do the regular "merge" (i.e. _not_ squash), because there's 2 commits which refer to each other in this PR, and it won't make sense if they're squashed.

qutip/wigner.py
```diff
@@ -574,31 +574,57 @@ def _qfunc_check_state(state: Qobj):
...
+    --------
+    >> N = 64 # Hilbert size
+    >> xvec, yvec, g = [1, 0, -1], [1, 0, -1], 1 # Grid and coefficient g
+
+    >> X, Y = np.meshgrid(xvec, yvec)
+    >> betas = 0.5*g*(X - 1j*Y).ravel() # Complex numbers denoting phase space points
+
+    QuTiP's implementation of computing coherent states
+    >> coherent_states_qutip = np.array([coherent(N, beta).full() 
+                                    for beta in betas]).reshape(3, 3, -1) # reshaped to a tensor
```
jakelishman: Looks like you're using `black` with too high a line limit for QuTiP.  QuTiP's got a hard limit at 79 characters, rather than black's lax limit up to 88.

jakelishman:
You want to be very careful using black - we haven't applied it to most of the codebase, so if you're letting it run, then it's going to produce large unrelated changes everywhere and make the PR harder to understand.  Also, black doesn't abide by the exact set of rules set up in codeclimate, so it'll also add a bunch of errors.

This would be part of a larger change we'd have to make across the whole codebase separately.

jakelishman:
Also, once you're done, leave this PR without merging and I'll tidy up the commit history.  We shouldn't have things like 374debd (a wrong-way merge) because they really mess with the mainline of the history, and make things harder to traceback, and I can collapse all your changes into one "logical" change once it's working.

quantshah:
> Whoops!

If you're using a regex-capable find-and-replace, try using `\<xs\>` or `\bxs\b` instead of just `xs` so that it only matches if the `xs` has word boundaries on either side of it. 

Oops, I will check this and un-black the code.

quantshah:
> Also, once you're done, leave this PR without merging and I'll tidy up the commit history.  We shouldn't have things like 374debd (a wrong-way merge) because they really mess with the mainline of the history, and make things harder to traceback, and I can collapse all your changes into one "logical" change once it's working. 

Ok. But what would be the better way to merge master here then?

jakelishman:
Without the black changes, the patch would have applied cleanly, I think.  Without them, it's better to wait til the end, and then rebase the PR on top of master (in git, not github), then merge the PR into master as normal.  The rebase is a destructive operation (rewrites history of the PR), so we have to take more care when working with more than one person, but it's better afterwards.  I've heard this called a few things - Atlassian calls it a "foxtrot merge", for example.  Here's a blog post they made about them, and some of the problems: https://blog.developer.atlassian.com/stop-foxtrots-now/.

qutip/wigner.py
```diff
@@ -574,31 +574,57 @@ def _qfunc_check_state(state: Qobj):
...
         raise ValueError(
-            f"xs and ys must be 1D, but have shapes {xs.shape} and {ys.shape}."
+            f"xvec and yvec must be 1D, but have shapes {xvec.shape} and {yvec.shape}."
         )
-    return xs, ys
+    return xvec, yvec
 
 
-class _qfunc_alpha_matrix:
+class _CoherentGrid:
```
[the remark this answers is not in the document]
quantshah: I see. I was just a bit confused because I never saw Class names written like that.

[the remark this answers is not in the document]
jakelishman: The distinction between Python "functions" and "classes" is a pretty nebulous one, to be fair.  I don't mind how you want to case it - if you prefer camel case, then by all means go for it.  I think having some reference to "QFunc" in the name will make it easier to skim the code, but that's about the only opinion I have for naming it.

quantshah:
@jakelishman This is okay for me now, please have a look and merge/rebase with the proper history. Thanks.

PR #1584, opened 2021-06-21
Data.__init__ set shape

Ericgig:
Being unable the set the shape is the main constrain in making a new python only data layer.
This allow `Data.__init__` to set it.


hodgestar:
I know this is a very simple change, but could we add a test that it works? E.g. maybe a test in which we subclass `Data` and then create an instance of it and access shape?

hodgestar (approved):
Woot.

jakelishman (approved):

PR #1588, opened 2021-06-28
Fix dimensions in calls to mesolve when calculating propagators for superoperators or with collapse operators present.

### LINKED ISSUE #1585: propagator fails to dimension check for composite system
Hi, I report a strange behavior of `propagator` method.
`propagator` with collapse operator fails to dimension validation on composite system.

At first, single system works correctly.
```python
import qutip as qt
```
```python
H = qt.sigmaz()
c_ops = [np.sqrt(1) * qt.sigmam()]
tlist = np.linspace(0, 10, 201)
rho0 = qt.ket2dm(qt.basis(2, 0)).unit()

# mesolve
rho_f = qt.mesolve(H, rho0, tlist, c_ops=c_ops).states[-1]

# using propagator
F = qt.propagator(H, tlist, c_op_list=c_ops)[-1]
rho_f_prop = qt.vector_to_operator(F * qt.operator_to_vector(rho0))

rho_f.overlap(rho_f_prop)
```

The output is
```
0.9999092149599982
```
(I'm not sure why this is not 1. I guess it's limited by the numerical precision.)

However, same code on composite system fails.
The differences are only `H`, `c_ops`, and `rho0`, using `qt.tensor(*, qt.qeye(2))`
```python
H = qt.tensor(qt.sigmaz(), qt.qeye(2))
c_ops = [np.sqrt(1) * qt.tensor(qt.sigmam(), qt.qeye(2))]
tlist = np.linspace(0, 10, 201)
rho0 = qt.ket2dm(qt.tensor(qt.basis(2, 0), qt.basis(2, 0))).unit()

# mesolve
rho_f = qt.mesolve(H, rho0, tlist, c_ops=c_ops).states[-1]

# using propagator
F = qt.propagator(H, tlist, c_op_list=c_ops)[-1]
rho_f_prop = qt.vector_to_operator(F * qt.operator_to_vector(rho0))

rho_f.overlap(rho_f_prop)
```

```
<ipython-input-1-551bcc27bcfa> in <module>
      8 
      9 # using propagator
---> 10 F = qt.propagator(H, tlist, c_op_list=c_ops)[-1]
     11 rho_f_prop = qt.vector_to_operator(F * qt.operator_to_vector(rho0))
     12 

c:\users\mizuno\research\py38\lib\site-packages\qutip\propagator.py in propagator(H, t, c_op_list, args, options, unitary_mode, parallel, progress_bar, _safe_mode, **kwargs)
    239                 rho0 = Qobj(sp.csr_matrix(([1], ([row_idx], [col_idx])),
    240                                           shape=(N, N), dtype=complex))
--> 241                 output = mesolve(H, rho0, tlist, c_op_list, [], args, options,
    242                                  _safe_mode=False)
    243                 for k, t in enumerate(tlist):

c:\users\mizuno\research\py38\lib\site-packages\qutip\mesolve.py in mesolve(H, rho0, tlist, c_ops, e_ops, args, options, progress_bar, _safe_mode)
    263         raise Exception("Invalid H type")
    264 
--> 265     func, ode_args = ss.makefunc(ss, rho0, args, e_ops, options)
    266 
    267     if _safe_mode:

c:\users\mizuno\research\py38\lib\site-packages\qutip\mesolve.py in _qobjevo_set(HS, rho0, args, e_ops, opt)
    349         # Should be caught earlier in mesolve.
    350         raise ValueError("rho0 must be a ket, density matrix or superoperator")
--> 351     _test_liouvillian_dimensions(H_td.cte.dims, rho0.dims)
    352     return func, ()
    353 

c:\users\mizuno\research\py38\lib\site-packages\qutip\mesolve.py in _test_liouvillian_dimensions(L_dims, rho_dims)
    330         raise ValueError("Liouvillian had nonsquare dims: " + str(L_dims))
    331     if not ((L_dims[1] == rho_dims) or (L_dims[1] == rho_dims[0])):
--> 332         raise ValueError("".join([
    333             "incompatible Liouvillian and state dimensions: ",
    334             str(L_dims), " and ", str(rho_dims),

ValueError: incompatible Liouvillian and state dimensions: [[[2, 2], [2, 2]], [[2, 2], [2, 2]]] and [[4], [4]]
```

My environment is following.
```
Current admin team: Alexander Pitchford, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, Boxi Li, Jake Lishman and Simon Cross.

QuTiP Version:      4.6.2
Numpy Version:      1.19.5
Scipy Version:      1.6.2
Cython Version:     0.29.22
Matplotlib Version: 3.4.1
Python Version:     3.8.9
Number of CPUs:     16
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
Installation path:  c:\users\mizuno\research\py38\lib\site-packages\qutip
```

hodgestar:
**Description**
A lot of the code in `propagators.py` is quite old and manipulates `Qobj` shapes directly and unnecessarily converts backwards and forward between numpy arrays and `Qobj`s. This PR attempts to address two important bugs in dimension checking -- the one reported in #1585 (when calculating propagators for collapse operators) and another I found while reading the nearby code. A more thorough rework of the code should probably happen on `dev.major` for QuTiP version 5.

The `propagator` method could do with more input sanity checking for QuTiP version 5, e.g.
* rename `c_op_list` to `c_ops` for consistency with the solver
* complain if `c_ops` are supplied along with a superoperator (?)
* perhaps remove the `single` or `batch` unitary_mode option, or at least complain if the parameters supplied don't make sense.

**Related issues or PRs**
* #1585

**Changelog**
- Fixed support for calculating the ``propagator`` of a density matrix with collapse operators. QuTiP 4.6.2 introduced extra sanity checks on the dimensions of inputs to mesolve (#1459), but the propagator function's calls to ``mesolve`` violated these checks by supplying initial states with the dimensions incorrectly set. ``propagator`` now calls ``mesolve`` with the correct dimensions set on the initial state. Fixes #1585.
- Fixed support for calculating the ``propagator`` for a superoperator without collapse operators. This functionality was not tested by the test suite and appears to have broken sometime during 2019. Tests have now been added and the code breakages fixed.

### DIFF
--- a/qutip/propagator.py
+++ b/qutip/propagator.py
@@ -236,8 +236,12 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
             for n in range(N * N):
                 progress_bar.update(n)
                 col_idx, row_idx = np.unravel_index(n, (N, N))
-                rho0 = Qobj(sp.csr_matrix(([1], ([row_idx], [col_idx])),
-                                          shape=(N, N), dtype=complex))
+                rho0 = Qobj(
+                    sp.csr_matrix(
+                        ([1], ([row_idx], [col_idx])),
+                        shape=(N, N), dtype=complex),
+                    dims=H0.dims,
+                )
                 output = mesolve(H, rho0, tlist, c_op_list, [], args, options,
                                  _safe_mode=False)
                 for k, t in enumerate(tlist):
--- a/qutip/propagator.py
+++ b/qutip/propagator.py
@@ -211,7 +211,7 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
         else:
             rho0 = qeye(H0.dims[0])
             output = mesolve(H, rho0, tlist, None, [], args, options,
-                             _safe_mode=True)
+                             _safe_mode=False)
             return output.states[-1] if len(tlist) == 2 else output.states
 
     else:
--- a/qutip/propagator.py
+++ b/qutip/propagator.py
@@ -208,10 +208,9 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
                 for k, t in enumerate(tlist):
                     u[:, n, k] = mat2vec(output[n].states[k].full()).T
         else:
-            rho0 = qeye(N, N)
-            rho0.dims = [[sqrt_N, sqrt_N], [sqrt_N, sqrt_N]]
-            output = mesolve(H, psi0, tlist, [], args, options,
-                             _safe_mode=False)
+            rho0 = qeye(H0.dims[0])
+            output = mesolve(H, rho0, tlist, None, [], args, options,
+                             _safe_mode=True)
             return output.states[-1] if len(tlist) == 2 else output.states
 
     else:
--- a/qutip/propagator.py
+++ b/qutip/propagator.py
@@ -48,7 +48,7 @@ from qutip.sparse import sp_reshape
 from qutip.cy.sparse_utils import unit_row_norm
 from qutip.mesolve import mesolve
 from qutip.sesolve import sesolve
-from qutip.states import basis
+from qutip.states import basis, projection
 from qutip.solver import Options, _solver_safety_check, config
 from qutip.parallel import parallel_map, _default_kwargs
 from qutip.ui.progressbar import BaseProgressBar, TextProgressBar
@@ -238,12 +238,8 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
             for n in range(N * N):
                 progress_bar.update(n)
                 col_idx, row_idx = np.unravel_index(n, (N, N))
-                rho0 = Qobj(
-                    sp.csr_matrix(
-                        ([1], ([row_idx], [col_idx])),
-                        shape=(N, N), dtype=complex),
-                    dims=H0.dims,
-                )
+                rho0 = projection(N, row_idx, col_idx)
+                rho0.dims = H0.dims
                 output = mesolve(
                     H, rho0, tlist, c_ops=c_op_list, args=args,
                     options=options, _safe_mode=False)
@@ -311,8 +307,8 @@ def _parallel_sesolve(n, N, H, tlist, args, options):
 
 def _parallel_mesolve(n, N, H, tlist, c_op_list, args, options, dims=None):
     col_idx, row_idx = np.unravel_index(n, (N, N))
-    rho0 = Qobj(sp.csr_matrix(([1], ([row_idx], [col_idx])),
-                              shape=(N, N), dtype=complex), dims=dims)
+    rho0 = projection(N, row_idx, col_idx)
+    rho0.dims = dims
     output = mesolve(
         H, rho0, tlist, c_ops=c_op_list, args=args, options=options,
         _safe_mode=False)
--- a/qutip/propagator.py
+++ b/qutip/propagator.py
@@ -196,15 +196,16 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
         sqrt_N = int(np.sqrt(N))
         dims = H0.dims
 
-        u = np.zeros([N, N, len(tlist)], dtype=complex)
-
         if parallel:
-            output = parallel_map(_parallel_mesolve, range(N * N),
+            output = parallel_map(_parallel_mesolve, range(N),
                                   task_args=(
                                       sqrt_N, H, tlist, c_op_list, args,
                                       options),
+                                  task_kwargs={"dims": H0.dims[0]},
                                   progress_bar=progress_bar, num_cpus=num_cpus)
-            for n in range(N * N):
+
+            u = np.zeros([N, N, len(tlist)], dtype=complex)
+            for n in range(N):
                 for k, t in enumerate(tlist):
                     u[:, n, k] = mat2vec(output[n].states[k].full()).T
         else:
@@ -226,6 +227,7 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
             output = parallel_map(_parallel_mesolve, range(N * N),
                                   task_args=(
                                       N, H, tlist, c_op_list, args, options),
+                                  task_kwargs={"dims": H0.dims},
                                   progress_bar=progress_bar, num_cpus=num_cpus)
             for n in range(N * N):
                 for k, t in enumerate(tlist):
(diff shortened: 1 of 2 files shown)


qutip/propagator.py
```diff
@@ -208,10 +208,9 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
                 for k, t in enumerate(tlist):
                     u[:, n, k] = mat2vec(output[n].states[k].full()).T
         else:
-            rho0 = qeye(N, N)
-            rho0.dims = [[sqrt_N, sqrt_N], [sqrt_N, sqrt_N]]
-            output = mesolve(H, psi0, tlist, [], args, options,
-                             _safe_mode=False)
+            rho0 = qeye(H0.dims[0])
+            output = mesolve(H, rho0, tlist, None, [], args, options,
```
Ericgig, line 212: Was this never tested? `psi0` is not defined and `e_ops` was missing... Thank you for catching it.

qutip/tests/test_propagator.py
```diff
@@ -127,5 +127,29 @@ def testPropHDims():
     U = propagator(H,1, unitary_mode='single')
     assert_equal(U.dims,H.dims)
 
+
+def testPropHSuperWithoutCops():
+    "Propagator: super operator without collapse operators"
+    H = tensor(sigmaz(), qeye(2))
+    H = to_super(H)
```
Ericgig, line 134: Why `to_super` and not `liouvillian`?

qutip/tests/test_propagator.py
```diff
@@ -127,5 +127,29 @@ def testPropHDims():
     U = propagator(H,1, unitary_mode='single')
     assert_equal(U.dims,H.dims)
 
+
+def testPropHSuperWithoutCops():
```
Ericgig, line 131: Could you add another test with super `H` and using `parallel=True`.

qutip/propagator.py
hodgestar: Yep, it was never tested. :/

qutip/tests/test_propagator.py
hodgestar: Because I did not think to use `liouvillian`. Fixed. :)

qutip/tests/test_propagator.py
hodgestar: Added! And also added a parallel test for the case with c_ops.

hodgestar:
@Ericgig Thank you for reviewing. Ready for review again, I think.

Ericgig (approved):
As a fix of the dimension bug, I see no problem, but as you said, propagator need a more thorough rework.

qutip/propagator.py
```diff
@@ -196,22 +196,22 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
...
+                             _safe_mode=True)
```
jakelishman, line 214: For reading clarity, it might be better to remove the `c_ops` and `e_ops` arguments here, and pass `args` and `options` by keyword.  (I personally think `c_ops` and `e_ops` should be keyword-only arguments to `mesolve`, but I'm not going to force that on anyone.)

In theory you could keep `_safe_mode=False` here (`propagator` has checked the dimensions of `H` and `rho`, so `mesolve` doesn't need to), which is consistent with the rest of the file.  That said, the cost of the checks is very low, but we'll hopefully get rid of it in `dev.major`; a stateful object encapsulating `mesolve` should allow users to skip these checks, in addition to us.

qutip/propagator.py
```diff
@@ -236,8 +237,12 @@ def propagator(H, t, c_op_list=[], args={}, options=None,
...
-                                          shape=(N, N), dtype=complex))
+                rho0 = Qobj(
+                    sp.csr_matrix(
+                        ([1], ([row_idx], [col_idx])),
+                        shape=(N, N), dtype=complex),
+                    dims=H0.dims,
+                )
                 output = mesolve(H, rho0, tlist, c_op_list, [], args, options,
                                  _safe_mode=False)
                 for k, t in enumerate(tlist):
```
jakelishman, line 248: This is a confusing way of writing `rho0 = projection(N, col_idx, row_idx); rho0.dims = H0.dims` (possibly with `col_idx` and `row_idx` the other way around).  Also, the `np.unravel_index` is quite confusing - it may be better just as `for row in range(N): for col in range(N)`.

jakelishman: It would also be more efficient if we create a single `QobjEvo` out of `H` and `c_op_list` before this loop, and pass that as the first argument to `mesolve`.  (Similarly to before, I'd also not pass `e_ops` at all because it's empty, and pass the others by keyword.)

jakelishman (approved):
Looks good to me as a fix.  I put in a couple of comments about additional stuff we might want to change in this function in a proper overhaul, but it doesn't need to be done right now.

A lot of the existing test file feels like it's crying out for parametrisation - we basically want to do the exact same thing about 8 times, with different "method" configurations.  Again, though, that's basically an entire re-write, and not necessarily in the scope of this PR.

nwlambert:
Thanks simon!
I checked the logic of the changes and played with the tests, and looks OK to me.

nwlambert (approved):

qutip/propagator.py
hodgestar: Thanks for picking up the `_safe_mode` change -- I initially changed it only for debugging purposes. Restored to `False`.

Agreed about passing `args` and `options` by keyword -- the long lists of unnamed arguments was a bit terrifying, so I've implemented that in this PR too.

qutip/propagator.py
hodgestar: I've made the `projection` and removal of `e_ops` changes because those make the code much more understandable, but I've left out the `QobjEvo` idea for now.

hodgestar:
@nwlambert @Ericgig @jakelishman I made some of the small clean-ups Jake suggested. I'll leave this open in case anyone wants to have a last look, but otherwise I'll merge a bit later today.

qutip/propagator.py
jakelishman: Sure - definitely best to draw the line somewhere, or we'll end up doing an entire re-write bit-by-bit.

jakelishman (approved):
Looks good.  Don't forget the changelog.

hodgestar:
Changelog updated and issue added to 4.6.3 milestone.

PR #1590, opened 2021-06-29
Set `superrep` for operator-ket and operator-bra.

### LINKED ISSUE #1586: dev.major's Qobj's __matmul__ set wrong superrep for operator-ket
**Describe the bug**
`operator-ket`'s `superrep` is `None` at creation.
But `Qobj.__matmul__` set the `superrep` of the output to the one of the left `Qobj`.
Thus `super-operator @ operator-ket` result  in an `operator-ket` with `superrep != None`.
Some operations, such as addition, cannot be done between object with different `superrep`.

**To Reproduce**
```
operket = qutip.operator_to_vector(qutip.qeye(2))
super_oper = qutip.spre(qutip.qeye(2))
super_oper @ operket - operket
```
the last operation result in a `TypeError`.

(#1476)

Ericgig:
**Description**
Set `operator-ket`'s `superrep` to `'super'`.
`operator-ket` are the states of super operators, ie. `super @ operator-ket` make sense but `oper @ operator-ket` does not. So it is natural that they have an attached `superrep`. This will make `choi_super @ operator-ket` error unless care is taken in making the `superrep` match.

TODO? Have `to_choi`, etc. work for operator-kets?
closes #1586

### DIFF
--- a/qutip/core/superoperator.py
+++ b/qutip/core/superoperator.py
@@ -269,6 +269,7 @@ def operator_to_vector(op):
     return Qobj(stack_columns(op.data),
                 dims=[op.dims, [1]],
                 type='operator-ket',
+                superrep="super",
                 copy=False)
 
 
@@ -280,6 +281,8 @@ def vector_to_operator(op):
     """
     if not op.isoperket:
         raise TypeError("only defined for operator-kets")
+    if op.superrep != "super":
+        raise TypeError("only defined for operator-kets in super format")
     dims = op.dims[0]
     return Qobj(unstack_columns(op.data, (np.prod(dims[0]), np.prod(dims[1]))),
                 dims=dims,
--- a/qutip/core/superoperator.py
+++ b/qutip/core/superoperator.py
@@ -266,6 +266,9 @@ def operator_to_vector(op):
     representation.  Note that QuTiP uses the _column-stacking_ convention,
     which may be different to what you expect.
     """
+    if op.type in ['super', 'operator-ket', 'operator-bra']:
+        raise TypeError("Cannot convert object already "
+                        "in super representation")
     return Qobj(stack_columns(op.data),
                 dims=[op.dims, [1]],
                 type='operator-ket',
--- a/qutip/tests/core/test_superoper.py
+++ b/qutip/tests/core/test_superoper.py
@@ -73,11 +73,18 @@ class TestMatVec:
         rho1 = qutip.rand_dm(N)
         as_vec = qutip.operator_to_vector(rho1)
         assert as_vec.superrep == 'super'
+
         with pytest.raises(TypeError) as err:
             as_vec.superrep = ""
             qutip.vector_to_operator(as_vec)
         assert err.value.args[0] == ("only defined for operator-kets "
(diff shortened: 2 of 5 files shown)


qutip/core/superoperator.py
```diff
@@ -269,6 +269,7 @@ def operator_to_vector(op):
     return Qobj(stack_columns(op.data),
                 dims=[op.dims, [1]],
                 type='operator-ket',
+                superrep="super",
```
hodgestar, line 272: Could we add direct unit tests (or asserts to existing tests) that check that:

1. `operator_to_vector` sets `superrep` correctly now.
2. `vector_to_operator` raises the expected error if an `operator-ket` with the wrong `superrep` is given.

hodgestar: Separately, I'm wondering if `operator_to_vector` is missing some input checks? Can one really call it on any Qobj?

hodgestar (commented):
I left some small suggestions, but otherwise looks good to me.

qutip/core/superoperator.py
Ericgig: I added input check. Since we support non-square super-operator, the check is no very strict. We may want to review this once we have better dimension handling.

qutip/tests/core/test_superoper.py
```diff
@@ -67,6 +68,25 @@ def testOperatorVector(self):
...
+        assert as_vec.superrep == 'super'
+
+        with pytest.raises(TypeError) as err:
+            as_vec.superrep = ""
+            qutip.vector_to_operator(as_vec)
+        assert err.value.args[0] == ("only defined for operator-kets "
+                                     "in super format")
+
+        with pytest.raises(TypeError) as err:
+            as_vec.superrep = ""
```
hodgestar, line 84: It's a bit odd setting `superrep = ""` here because 1) it has already been set to `""` above and 2) its not directly relevant to the test. I would perhaps move `as_vec.superrep = ""` above both `with pytest.raises(...) ...` blocks and add a short comment on why it is being set to `""`.

qutip/tests/core/test_superoper.py
```diff
@@ -67,6 +68,24 @@ def testOperatorVector(self):
...
 
+    def testsuperrep(self):
+        N = 3
+        rho1 = qutip.rand_dm(N)
+        as_vec = qutip.operator_to_vector(rho1)
+        assert as_vec.superrep == 'super'
+
+        with pytest.raises(TypeError) as err:
+            as_vec.superrep = ""
+            qutip.vector_to_operator(as_vec)
```
hodgestar, line 79:
```suggestion
            bad_vec = as_vec.copy()
            bad_vec.superrep = ""
            qutip.vector_to_operator(bad_vec)
```

hodgestar (approved):
Looks good!

qutip/tests/core/test_superoper.py
hodgestar: Fixed!

hodgestar:
@Ericgig Tests passed and we both approved, so I've clicked the merge button.

PR #1592, opened 2021-06-29
Change eigs eigenvector output to Data instead of ndarray

Ericgig:
**Description**
Change `data.eigs` evecs output to a Data object instead of a `ndarray`. This will be useful for `tensorflow` for which `eigh` output is a tensor. `Qobj.eigenstate` have been reworked to use the dispatcher or `eigs_csr`. `eigs_csr` do not have the same signature s `eigs.data`, so it is still called directly.

This had side effect in `Qobj.sqrtm` which was rewritten to work with `Data` instead of `ndarray`.

2 new dispatched functions are added:
`split_columns`: Used in `Qobj.eigenstates`, split each cols into a list of ket shaped Data.
`inv`: matrix inverse.

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -936,24 +936,23 @@ class Qobj:
         """
         if self.dims[0] != self.dims[1]:
             raise TypeError('sqrt only valid on square matrices')
-        # TODO: consider another way of handling the dispatch here.
-        if sparse:
-            evals, evecs = _data.eigs_csr(_data.to(_data.CSR, self.data),
+        if isinstance(self.data, _data.CSR) and sparse and self.shape[0] > 3:
+            evals, evecs = _data.eigs_csr(self.data,
                                           isherm=self._isherm,
                                           tol=tol, maxiter=maxiter)
+        elif isinstance(self.data, _data.CSR):
+            evals, evecs = _data.eigs(_data.to(_data.Dense, self.data),
+                                      isherm=self._isherm)
         else:
-            evals, evecs = _data.eigs_dense(_data.to(_data.Dense, self.data),
-                                            isherm=self._isherm)
-        evecs = np.hstack(evecs)
+            evals, evecs = _data.eigs(self.data, isherm=self._isherm)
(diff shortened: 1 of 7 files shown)


qutip/core/qobj.py
```diff
@@ -936,24 +936,23 @@ def sqrtm(self, sparse=False, tol=0, maxiter=100000):
         """
         if self.dims[0] != self.dims[1]:
             raise TypeError('sqrt only valid on square matrices')
-        # TODO: consider another way of handling the dispatch here.
-        if sparse:
-            evals, evecs = _data.eigs_csr(_data.to(_data.CSR, self.data),
+        if isinstance(self.data, _data.CSR) and sparse and self.shape[0] > 3:
```
hodgestar, line 939: `eigs_csr` should handle the `shape[0] < 3` case itself (using `eigs_dense` if needed).

hodgestar (approved):
I've approved because I'm happy that this PR moves the code forward.

In general, I think we should work to avoid code like `if isinstance(self.data, _data.CSR)` inside Qobj because it prevents the data layer abstraction from doing its job (i.e. specializing operations on data types). Specifically, code like this breaks a hypothetical `CuPySparse` data implementation (or at least it would not have access to the `sparse` parameter so it wouldn't be able to special case on it).

Solutions we could use in future:
* Remove the `sparse` parameter and have `eigs_csr` do something sensible for sparse matrices (e.g. return a dense representation of eigenvectors and possibly use a dense eigenvector algorithm if it believes that will be faster).
* Add `sparse` to `.eigs` in the data layer as a hint to the data layer on how to handle this case.

Both of these require a bit more thought, but I fairly sure it's the data layers job to decide what algorithm should be used, not Qobj's.

Ericgig:
I agree the dispatch is not ideal, `eigs_crs` take extra parameters that `eigs` does not and are managed in `Qobj` and should not. Pushing everything to the dispatch force it to have parameters used by only one specialization, which also feel wrong, but should be a lot more practical.

PR #1593, opened 2021-06-29
faster algorithm for enr_destroy

jfeist:
**Description**
The algorithm used in `enr_destroy` was extremely inefficient, with a triple loop. This PR replaces it with a much simpler algorithm with only a single loop. For `dims = [3,8,2,4,6,1,3]` and `Nexc = 6`, on my computer `enr_destroy(dims,Nexc)` took almost 4 s, with the new algorithm it takes 20 ms. I've checked that the output is identical.

**Changelog**
Made enr_destroy significantly faster.

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -892,12 +892,13 @@ def enr_destroy(dims, excitations):
              for _ in range(len(dims))]
 
     for n1, state1 in idx2state.items():
-        for n2, state2 in idx2state.items():
-            for idx, a in enumerate(a_ops):
-                s1 = [s for idx2, s in enumerate(state1) if idx != idx2]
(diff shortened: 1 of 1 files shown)


Ericgig (approved):
Looks good, thank you for the optimization.

hodgestar (approved):
Thank you for this drastic performance improvement.

hodgestar:
@jfeist @Ericgig I've added this to the 4.6.3 release milestone. The result of `enr_destroy` was not incorrect, but I'm sure users will appreciate the drastic performance improvement. Even 4s is a long time to wait at a Jupyter notebook prompt even before one starts to call `enr_destroy` often or allow more excitations.

PR #1594, opened 2021-06-30
faster state_number_enumerate

jfeist:
**Description**
When the number of excitations in `state_number_enumerate` is limited, instead of iterating over all states and discarding the ones with too many excitations, directly choose the limits to only iterate over allowed states. As a small additional optimization, do not redo the same sum every time, but keep track of the sum within the algorithm. For the tests I've made, this is a factor of ~4-10 faster than the current version. Together with #1593, this reduces the runtime of `enr_destroy` (which uses `state_number_enumerate`) from almost 4s to 15 ms for the case I just treated, and another much bigger case takes 2.5 s now, while it hadn't finished after more than an hour with the previous version.

Note that since the two PRs (this one and #1593) are somewhat related, it might make sense to combine them into a single one. I'd be happy to do that.

**Changelog**
Made state_number_enumerate somewhat faster.

### DIFF
--- a/qutip/states.py
+++ b/qutip/states.py
@@ -757,7 +757,7 @@ def bra(seq, dim=2):
 #
 # quantum state number helper functions
 #
-def state_number_enumerate(dims, excitations=None, state=None, idx=0):
+def state_number_enumerate(dims, excitations=None, state=None, idx=0, nexc=0):
     """
     An iterator that enumerate all the state number arrays (quantum numbers on
     the form [n1, n2, n3, ...]) for a system with dimensions given by dims.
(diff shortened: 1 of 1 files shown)


nwlambert:
This and #1593 both look like solid improvements to me, I did  a little testing and couldn't break anything. Also your approach makes the code much more readable than it was.

Probably @hodgestar  can comment on whether it is best to combine them into one?

Ericgig (approved):

hodgestar (approved):
Thanks @jfeist!

I think we could make an even faster non-recursive version. Are you interested in giving that a try? If not, maybe I will find a bit of time to do it.

jfeist:
I've given it a shot in #1604, together with some other (possible) improvements.

PR #1600, opened 2021-07-06
Seed is no longer ignored when pure=True for rand_dm

### LINKED ISSUE #1598: Documentation for rand_dm
**Describe the bug**
Providing seed as an integer to `rand_dm` does not fixate the seed but randomizes it for each run. In the documentation of `rand_dm` a description of seed is missing.

**To Reproduce**
```python
from qutip import rand_dm
rho = rand_dm(2, density=0.75, pure = True, seed = 1)
print(rho)
```
The terminal out put is different for each run

**Expected behavior**
This gives the correct expected behavior
```python
from qutip import rand_dm
rho = rand_dm(2, density=0.75, pure = True, seed = np.random.seed(1))
print(rho)
```

**Additional context**
Please add a description of seed in `rand_dm` that specifies that the seed parameter should be given as `np.random.seed`

Wikstahl:
**Description**
`seed` is no longer ignored when `pure=True` for `rand_dm`

**Related issues or PRs**
fix #1598

### DIFF
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -309,6 +309,8 @@ def rand_ket(N=None, density=1, dims=None, seed=None):
     dims : list
         Dimensions of quantum object.  Used for specifying
         tensor structure. Default is dims=[[N],[1]].
+    seed : int
+        Seed for the random number generator.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


Ericgig (approved):
Thank you for the fix.

hodgestar:
@quantshah @Ericgig I added the milestone https://github.com/qutip/qutip/milestone/19 to this issue so that we remember to include this PR in the 4.6.3 release.

PR #1601, opened 2021-07-06
Import the BoFiN HEOM implementation into QuTiP for 4.7.

hodgestar:
**Description**
This is imports [BoFiN](https://github.com/tehruhn/bofin/) into QuTiP with the intention of replacing `qutip.nonmarkov.heom` with `qutip.nonmarkov.bofin` in 4.7 (where `qutip.nonmarkov.heom` will be a thin wrapper around the new, better HEOM solvers) and completely replacing `qutip.nonmarkov.heom` in 5.0

The strategy used was to import the BoFiN history (52 commits) without the examples and their data files into `qutip/nonmarkov/bofin` and then to proceed to tidy things up from there. Specifically I imported the branch from https://github.com/tehruhn/bofin/pull/7 which has @nwlambert's work on the `HSolverDL` replacement.

**Todo**
- [x] Add more exhaustive tests.
- [x] Address or silence CodeClimate issues.
- [x] Remove the many assigned but unused variables.
- [x] Add better tests that new `HSolverDL` is compatible with the existing `HSolverDL`.
- [x] Replace the existing `HSolverDL` with the new one.
- [x] Decide what to do with `heom.HEOMSolver`. Should it stay as it is? Be incorporated into BoFiN? Be Removed? *-- Neil and I decided we should remove it*
    - [x] Decide what to do about the `renorm` argument to the old `HSolverDL`. Do we add it to the new BoFiN one?
    - [x] Decide what to do about the `stats` argument. I'm planning to just add a note telling people to use `pyinstrument` if they need to profile their code.
- [x] Make HSolverDL use a bath instead of having the Matsubara calculation burried inside the solver.
- [x] Add Pade coefficient calculator somewhere? Maybe also as a bath?
- [x] Clean up the interfaces of `BosonicHEOMSolver` and `FermionicHEOMSolver`.
- [x] Whatever else we discover along the way.
- [x] Keep just the gather_op mode of RHS construction.
- [x] Split bath classes out into their own module.
- [x] Add underdamped spectral density bath.

**Follow up PRs (i.e. before 4.7)**
- [ ] Add Lorentzian fermionic bath.
- [ ] Convert the extension BoFiN documentation and into an extensive QuTiP guide (this PR already adds the docs but under qutip/nonmarkov/bofin/docs -- this allows the preservation of the origina BoFiN commits)
- [ ] Convert the BoFiN examples into QuTiP example notebooks (https://github.com/qutip/qutip-notebooks/pull/132/).
- [ ] Add MKL support to the dynamics solver.

**Possible future work (i.e. look like great ideas but might never happen)**
- [ ] Use hermiticity relation from Schinabeck's PhD to half the number of ADOs stored in the fermionic bath.
- [ ] Possibly add support for time-dependent bath properties (e.g. altering the chemical potential of a Fermionic bath).
  - [ ] Option to include time-dependent exponent coupling operators (Q)
  - [ ] Option to include time-dependent exponent coefficients (c_k's and v_k's)
- [ ] Possibly add a means for automatically generating bath exponents for a given spectral density function.

**Related issues or PRs**
None currently.

**Changelog**
Imported BoFiN (https://github.com/tehruhn/bofin/) into QuTiP and replaced the HEOM solver with a compatibility wrapper around BoFiN bosonic solver.


nwlambert:
Thanks simon!  On the tests we have one that checks the dynamics for the bosonic solver, and one that checks the steady state for the fermionic stuff, but not vice versa. I will try and construct something.

I think for the HSolverDL backwards compatibility test, we can reuse the old test (which is essentially the same test I put in for the bosonic solver).

In bofin.py itself, I think we can replace _pad_csr() with the cy_pad_csr cythonised version in qutip already (that HSolverDL was already using), its a bit faster (I didn't do it so far because of bug checking, but it should be ok to change now).

There are probably lots of small places where things can be optimized a little as well.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1553 @@
...
+            # Use superLU solver
+
+            LU = splu(L)
+            solution = LU.solve(b_mat)
+
+        dims = self.H_sys.dims
+        data = dense2D_to_fastcsr_fmode(vec2mat(solution[:sup_dim]), n, n)
+        data = 0.5 * (data + data.H)
+
+        solution = solution.reshape((nstates, self.H_sys.shape[0] ** 2))
```
nwlambert, line 728: if H is a superoperator, this fails.....  safer to use 

`solution = solution.reshape((nstates, sup_dim))`

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1553 @@
...
```
nwlambert, line 1446: As above, safer to use

`solution = solution.reshape((nstates, sup_dim))`

(credit to Hai-Feng Yang  for spotting this)

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1553 @@
...
+
+            self.progress_bar.finished()
+            return output
+
+        else:
+            self.progress_bar.start(n_tsteps)
+            N_he = self.nhe
+            N = shape[0]
+            hshape = (N_he, N**2)
+            full_hierarchy = [rho0.reshape(hshape)]
```
nwlambert, line 813: minor comment:  this reshape is useful for the utility functions in the example notebooks, for extracting bath properties like current and things. But it is a little inconsistent with how full__init=True is used, where the assumed shape is different. Maybe set this reshape as an option? (if so, the same should be done to the fermion class)

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1553 @@
...
+                        "collated automatically."
+                    )
+                    common_ck.append(ckAR[i])
+                    common_ck.append(ckAI[j])
+                    common_vk.append(vkAR[i])
+                    common_vk.append(vkAI[j])
+                    real_indices.append(i)
+                    img_indices.append(j)
+                    common_coup_op.append(coup_op[i])
+
```
nwlambert, line 334: the above checks if two exponents are similar, and collates them into a single exponent. This is pretty important for efficiency and convergence in some cases, particularly the Drude-Loretnz spectral density, but its a little clunky.  I think at least we need a better warning message.   E.g., since these 'new' collated exponents are added to the end of the list, if the user is trying to do something with the ADOs, they may get mixed up about how various ADOs correspond to various exponents, or various baths, since the ADOs ordering will be inconsistent with the input lists of parameters.  Perhaps returning the ADOs from run() or steady_state() with a dictionary showing which ADOs correspond to which input parameters/baths/etc is an option?   This is a pretty niche issue, and non of the examples we put together ran into it, it came up in an email from a user (Hai-Feng Yang)

qutip/nonmarkov/bofin.py
hodgestar: Thanks. Fixed.

qutip/nonmarkov/bofin.py
hodgestar: Fixed.

qutip/nonmarkov/bofin.py
```diff
@@ -1048,17 +1044,16 @@ def _configure_solver(self):
         RHSmat = RHSmat.tocsr()
 
         if self.isTimeDep:
-            solver_params = []
             h_identity_mat = sp.identity(nstates, format="csr")
             H_list = self.H_sys.to_list()
 
-            # Store each time dependent component
+            # store each time dependent component
+            solver_params = [[RHSmat]]
```
hodgestar, line 1051: @nwlambert I added the RHS matrix to the start of the solver_params to the time-dependent fermionic case because it wasn't there before. None of the tests broke and I added a dummy time-dependent test in an earlier commit. Is this the correct change to make?

nwlambert: Yikes, yeah good spot.  Next time we chat we can sort out any missing tests to catch this kind of thing.

qutip/nonmarkov/bofin.py
```diff
@@ -530,38 +484,77 @@ def boson_grad_prev(self, he_n, k, prev_he):
...
         return self._pad_op(op, he_n, prev_he)
 
-    def boson_grad_next(self, he_n, k, next_he):
-        """
-        Get the next gradient
-        """
+    def _grad_prev_fermionic(self, he_n, k, prev_he):
+        ck = self.bath.ck
+
+        n_excite = sum(he_n)
```
nwlambert, line 490: i guess if we want to try allowing bosonic+fermionic baths at the same time, we need a check at here and elsewhere to make sure we only sum over fermionic indices.  maybe this is all? the use of generic dims in the bathmodes stuff helps us do the cut-offs correctly?

qutip/nonmarkov/bofin.py
```diff
@@ -321,10 +352,24 @@ def __init__(
...
 
+    def _pad_op(self, op, row_he, col_he):
+        """
+        Pad op into its correct position within the larger HEOM liouvillian
+        for the given row and column bath states.
+        """
+        nhe = self.bath.n_states
+        rowidx = self.bath.idx(row_he)
+        colidx = self.bath.idx(col_he)
+        return cy_pad_csr(op, nhe, nhe, rowidx, colidx)
```
nwlambert, line 440: just a note:  I did a lot of testing on difficult cases, compared to the C++ version in bofin_fast, and this cy_pad_csr stuff is still a big bottleneck.  I saw a significant benefit by going back to the ''insertion'' approach @tehruhn  took, but also making sure that the RHS is in lil format first. In the old notation the insertion was done with 

   # Fill into larger L
    nidx = self.he2idx[he_n]
    block = self.N ** 2
    pos = int(nidx * block)
    self.L_helems[pos : pos + block, pos : pos + block] += L

I guess you could have the grad functions just return the small operator L, and do a similar insertion in the RHS construction function (+starting with RHS in lil format, and then switching back to CSR after everything is constructed).

nwlambert (commented):
This is a really great reorganization, thanks@hodgestar.  Makes us a bit future proof, and makes improvements in the future much easier.

qutip/nonmarkov/bofin.py
```diff
@@ -26,6 +28,79 @@
 from qutip.ui.progressbar import BaseProgressBar
 
 
+class BathMode:
+    """
+    Represents a single excitation mode within a bath.
```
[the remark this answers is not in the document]
hodgestar: Thanks! I updated these names and fleshed out the documentation a bit.

I kept the name `modes` for `Bath.modes` because `Bath.exps` looked odd to me. I'm happy to change it to something better -- I'm just not sure what the something better is at the moment.

qutip/nonmarkov/bofin.py
hodgestar: I implement bothed and got the following results on the test suite:

* `pad-csr`: ~11s
* `inplace add, CSR`: ~27s
* `inplace add, LIL`: ~27s
* `inplace add, DOK`: ~22s

The tests use very small systems though, so perhaps we need a bigger system for benchmarking.

The `_rhs` method now has an `_use_pad` option that defaults to `True`. You can set it to `False` if you want to use the `L[..., ...] += op` code though. You can also change `lil_matrix` to `dok_matrix` or `csr_matrix` if you want to experiment with different sparse matrix types.

qutip/nonmarkov/bofin.py
hodgestar: Yep! The idea is to add `self.bath.is_fermionic` (which is an array for 0 or 1) and then do `sum(he_n[k] * fermionic[k] for k in range(len(he_n))` or something like that.

Probably a good idea to get a test case first before fiddling too much though.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1445 @@
...
+            for ck, vk in zip(ck_imag, vk_imag)
+        )
+
+        if combine:
+            exponents = self.combine(exponents)
+
+        super().__init__(exponents)
+
+    @classmethod
+    def combine(cls, exponents, rtol=1e-5, atol=1e-7):
```
nwlambert, line 217: this is much better than the old approach, nice to get real or imaginary parts grouped up as well (rather than just a warning).

Tor the tolerances, I think I chose fiddled with the original values so that they worked well for some specific examples.  I think allowing them to be passed by the user is a good choice.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1445 @@
...
+                b_mat,
+                perm=None,
+                verbose=True,
+                max_iter_refine=mkl_max_iter_refine,
+                scaling_vectors=True,
+                weighted_matching=mkl_weighted_matching,
+            )
+        else:
+            L = L.tocsc()
+            LU = splu(L)
```
nwlambert, line 1096: Very minor comment: Might be worth adding support for lgmres and other methods at some point.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1445 @@
...
+            self.progress_bar.update(t_idx)
+            if t_idx != 0:
+                solver.integrate(t)
+            rho = Qobj(
+                solver.y[:n ** 2].reshape(rho_shape, order='F'),
+                dims=rho_dims,
+            )
+            output.states.append(rho)
+            if ado_return:
+                output.ado_states.append(solver.y.reshape(hierarchy_shape))
```
nwlambert, line 1174: very minor comment:  this reshape is useful for extracting info from the ados, but i think is incompatible with then using this output as input for ado_inti in .run()?

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1445 @@
...
+        self, H_sys, Q, ck_plus, vk_plus, ck_minus, vk_minus, max_depth,
+        options=None,
+    ):
+        bath = FermionicBath(Q, ck_plus, vk_plus, ck_minus, vk_minus)
+        super().__init__(
+            H_sys, bath=bath, max_depth=max_depth, options=options
+        )
+
+
+class _GatherHEOMRHS:
```
nwlambert, line 1358: Just a not for posterity, we did some checking of @hodgestar's new gather method against the direct inserts in CSR matrices, the old cython pad_csr method, and against tarun's c++ code.   This gather method definitely beats both the old python/cython methods, and seems to scale much better as we increase the problem size. C++ still wins overall, but this makes the python construction of the RHS much more useable in big problems.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1773 @@
...
+                ado_state = HierarchyADOsState(
+                    rho, self.ados, solver.y.reshape(hierarchy_shape)
+                )
+            if ado_return:
+                output.ado_states.append(ado_state)
+            for e_key, e_op in e_ops.items():
+                if isinstance(e_op, Qobj):
+                    e_result = (rho * e_op).tr()
+                else:
+                    e_result = e_op(t, ado_state)
```
Ericgig, line 1500: This does not match the description.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1773 @@
...
+        self._spreQ = [spre(op).data for op in Qs]
+        self._spostQ = [spost(op).data for op in Qs]
+        self._s_pre_minus_post_Q = [
+            self._spreQ[k] - self._spostQ[k] for k in range(self._n_exponents)
+        ]
+        self._s_pre_plus_post_Q = [
+            self._spreQ[k] + self._spostQ[k] for k in range(self._n_exponents)
+        ]
+
+        self.progress_bar = BaseProgressBar()
```
Ericgig, line 1065: The solving is probably slow and so we should not hide the progress bar here.
It should at least be mentioned in the doc-string.

Ericgig (commented):
@hodgestar
I gave it a first read.
I would suggest the split the `bofin.py` file and move out the `Bath` to make it less heavy.

The interface for `Bath` seems quite hard to use. An expansion of the correlation function is useful for the solving, but I feel we should have more helper function for the user.

Are `BathExponent` needed? I don't see why there is one object for each term and not one object for the whole decomposition.

hodgestar:
> @hodgestar I gave it a first read

Thank you!

> I would suggest the split the `bofin.py` file and move out the `Bath` to make it less heavy.

I've been thinking of splitting the file up a bit too but have been waiting until I was ready to replace heom.py. The Bath and Bath exponent classes could possibly be separated from the solver and solver support classes.

> The interface for `Bath` seems quite hard to use. An expansion of the correlation function is useful for the solving, but I feel we should have more helper function for the user.

There are three tiers of bath classes:

* Bath: Low-level. One gets to specify exactly the exponents one wants by hand.
* BosonicBath and FermionicBath: Mid-level. Flexible but one has to specify the expansion coefficients.
* DrudeLorentzBath and DrudeLorentzPadeBath: High-level. Support for specific bath models.

I discussed having a class that would automatically fit a correlation function with Niell, and he has experiment with that but it's bit tricky to make it work well in a generic way. E.g. For the DrudeLorentz bath there is a divergence in the exponent series at t = 0 which makes it hard to fit numerically.

We'd like to add more generic support in future, but the current approach provides all the functionality of the existing HSolverDL and is sufficiently flexible to implement many more higher-level Bath classes easily in the future.
 
> Are `BathExponent` needed? I don't see why there is one object for each term and not one object for the whole decomposition.

BathExponent its just a holder for all of the information associated with an expansion term. Three issues with treating whole expansions as one object are:

* some exponents contribute to two expansions (e.g. the RI exponent type)
* some exponents are paired (e.g. the + and - types)
* having exponents within expansions makes dealing with the hierarchy a mess because one has to have lists of lists of exponents with relationships between them and ADO labels would have to become nested too (at least conceptually)

nwlambert:
> I discussed having a class that would automatically fit a correlation function with Niell, and he has experiment with that but it's bit tricky to make it work well in a generic way. E.g. For the DrudeLorentz bath there is a divergence in the exponent series at t = 0 which makes it hard to fit numerically.


Yeah, for the fitting stuff, we kept it for the example notebooks, since as simon said it's very problem dependent. hence having a very generic bath interface helps for inputting these low-level cases.  

For high-level interface stuff, we can also add the underdamped (brownian-motion) bath example (matsubara + pade as well). We can also add the equivalent for the fermionic case.  Simon's setup makes it easy to do.

qutip/nonmarkov/bofin.py
Ericgig: In v4, it's a parameter in the `..solve` function and in v5 it will be an option.

Ericgig:
Having a tool to fit a correlation function would be great for the user, I see that it's certainly not easy nor  a priority.
I believe having high level Bath available will be important to get user to interested.

The issue I see with `BathExponent` is that it contain a `Q` which seems to be usually shared for one Bath. While the exponent themselves can share an instance, `_spreQ`, `_spostQ`, `_s_pre_minus_post_Q` and `_s_pre_plus_post_Q` will be independent for each exponent. I don't know how many exponent per Bath you expect, if it's 1~2, it could be fine, but if it get large, you are computing and storing too many copies of these.

> some exponents are paired (e.g. the + and - types)
If they are to be paired, then why make each exponents independent? I see this as a reason to store them as a whole.

hodgestar:
> The issue I see with `BathExponent` is that it contain a `Q` which seems to be usually shared for one Bath. While the exponent themselves can share an instance, `_spreQ`, `_spostQ`, `_s_pre_minus_post_Q` and `_s_pre_plus_post_Q` will be independent for each exponent. I don't know how many exponent per Bath you expect, if it's 1~2, it could be fine, but if it get large, you are computing and storing too many copies of these.

Usually the number of exponents is less than 10.

I have thought about optimizing the `_spreQ` lists so that they share the instances of the superoperators if the underlying Q are the same, but it seemed a complication for not a huge benefit:

The shape of the RHS is `(block * nhe, block * nhe)` where `(block, block)` is the shape of `spre(Q)` and `nhe` is the number of labels in the hierarchy. `nhe ~ k**cutoff` where `k` is the number of exponents and `cutoff` is the maximum hierarchy depth. So the overall size of the RHS is `~ k**(2*cutoff - 1)` times the size of  the `_spreQ` list.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1773 @@
...
+            kappa.append(term)
+
+        epsilon = [0] + eps
+
+        return kappa, epsilon
+
+    def _calc_eps(self, Nk):
+        alpha = np.zeros((2 * Nk, 2 * Nk))
+        for j in range(2 * Nk):
+            for k in range(2 * Nk):
```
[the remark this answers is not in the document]
hodgestar: Nk will not be large, but I rewrote this with `np.diag` because it makes it a bit more readable.

qutip/nonmarkov/bofin.py
```diff
@@ -0,0 +1,1773 @@
...
+        kappa = [0]
+        prefactor = 0.5 * Nk * (2 * (Nk + 1) + 1)
+        for j in range(Nk):
+            term = prefactor
+            for k in range(Nk - 1):
+                term *= (
+                    (chi[k]**2 - eps[j]**2) /
+                    (eps[k]**2 - eps[j]**2 + self._delta(j, k))
+                )
+            for k in range(Nk - 1, Nk):
```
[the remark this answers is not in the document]
hodgestar: This makes it clear that we're completely the last item of the previous loop, but I replaced `range(...)` with just `[Nk - 1]` to make it obvious that there is only one element.

qutip/nonmarkov/bofin.py
hodgestar: Thanks! Documentation updated.

qutip/nonmarkov/bofin.py
hodgestar: I've added it as a parameter in the v4 style for now. Thanks!

hodgestar:
@nwlambert @Ericgig I've removed the "WIP" tag, so I'd like to start the "let's try merge this" round of reviews!

I've intentionally flagged a few items for follow up PRs (i.e. before 4.7) and some as future work (i.e. might never happen, but are ideas I don't want to lose track of).

qutip/nonmarkov/bofin_baths.py
```diff
@@ -0,0 +1,643 @@
...
+        The Matsubara terminator -- i.e. a liouvillian term representing the
+        contribution to the system-bath dynamics of all exponential expansion
+        terms beyond Nk. It should be used by adding it to the system
+        liouvillian (i.e. ``liouvillian(H_sys)``).
+
+    delta : float
+        The approximation discrepancy. That is, the difference between the true
+        correlation function of the Drude bath and the correlation function
+        represented by this object is approximately ``2 delta * dirac(t)``,
+        where ``dirac(t)`` denotes the Dirac delta function.
```
nwlambert, line 352: by ''this object'' i guess you mean the N_k correlation terms included in the heom?  maybe we can also clarify when this approximation is ok

nwlambert:
great!  For the todo list for 4.7, we can add the ''Underdamped spectral density'' bath option too.

qutip/nonmarkov/bofin_solvers.py
```diff
@@ -0,0 +1,1218 @@
...
+        sign1 = (-1) ** (n_excite + 1)
+
+        n_excite_before_m = sum(he_n[:k])
+        sign2 = (-1) ** (n_excite_before_m)
+
+        if sign1 == -1:
+            op = (-1j * sign2) * self._s_pre_minus_post_Q[k]
+        else:
+            op = (-1j * sign2) * self._s_pre_plus_post_Q[k]
+
```
nwlambert, line 604: I think in 601 and 604 the Q operators should be the adjoint (dagger) of the Q associated with that exponent

qutip/tests/test_bofin_solvers.py
```diff
@@ -0,0 +1,900 @@
...
+        ]
+
+        def exp_sign(exp):
+            return 1 if exp.type == exp.types["+"] else -1
+
+        # right hand modes are the first k modes in ck/vk_plus and ck/vk_minus
+        # and thus the first 2 * k exponents
+        k = self.lmax + 1
+        return 1.0j * sum(
+            exp_sign(exp) * (exp.Q * aux).tr()
```
nwlambert, line 708: I think this should be exp.Q.dag()

qutip/tests/test_bofin_solvers.py
```diff
@@ -0,0 +1,900 @@
...
+        hsolver = FermionicHEOMSolver(
+            H_sys, dlm.Q, ck_plus, vk_plus, ck_minus, vk_minus, Ncc,
+            options=options,
+        )
+
+        tlist = [0, 10]
+        result = hsolver.run(dlm.rho(), tlist, ado_return=True)
+        current = dlm.state_current(result.ado_states[-1])
+        analytic_current = dlm.analytic_current()
+        np.testing.assert_allclose(analytic_current, current, atol=atol)
```
nwlambert, line 868: these tests are passing despite some bugs because atol is too high, and these currents are very small (order 1e-5), so maybe setting an rtol insted of atol is better.  Also, due to me messing up the directionality of the current, we will need a minus sign on one of them!

qutip/nonmarkov/bofin_solvers.py
hodgestar: I agree that these should be based on `Q.dag()` or equivalently taken from `sigma_bar_k` instead of `k`. Do you know which of those two ways of expressing this is best?

After I fixed this, the currents in the test ended up being 0, which doesn't match the analytic result. Any thoughts on what to try next?

qutip/tests/test_bofin_solvers.py
hodgestar: Setting rtol would be better here. I'll do that as soon as we've figured out the correct current results so that I know what to set rtol to.

qutip/nonmarkov/bofin_solvers.py
hodgestar: I think there is also a separate question about what `exp.Q` should be for the `+` and `-` exponent types.

Currently, if the operator `Q` is supplied to the bath, then the + modes have `exp.Q == Q.dag()` and the - modes have `exp.Q == Q`. Perhaps the `exp.Q` should just be `Q` for both + and - exponents though? Then solver should sort out the difference between the two in the various `_grad_next/_prev` methods?

On the other hand, in many calculations one might need to take `exp.Q.dag() if exp.type == exp.types["+"] else exp.Q` which would be annoying if one is working with the exponents outside of the solver.

qutip/nonmarkov/dlheom_solver.py
```diff
@@ -0,0 +1,595 @@
...
+        if stats:
+            stats.add_timing('initialize',
+                             timeit.default_timer() - start_init, ss_run)
+            start_integ = timeit.default_timer()
+
+        dt = np.diff(tlist)
+        n_tsteps = len(tlist)
+        for t_idx, t in enumerate(tlist):
+            if t_idx < n_tsteps - 1:
+                r.integrate(r.t + dt[t_idx])
```
Ericgig, line 480: After integrating, you should check `r.successful()`. If the integrator fails, it does not raise an error but an `IntegrationWarning` and the state `r.y` is wrong, but often it still seems plausible.

The main error that you can expect is that too much work is done in one step, it this case it will stop the integration where it is when it reach the maximum work. `r.t` will be less than what you expect and the state will be at that time. Since you don't check, you may be saving the `state`/`expect` with a small time shift than what you expect.

qutip/nonmarkov/bofin_baths.py
```diff
@@ -0,0 +1,729 @@
...
+    """
+    def __init__(self, exponents):
+        self.exponents = exponents
+
+
+class BosonicBath(Bath):
+    """
+    A helper class for constructing a bosonic bath from the expansion
+    coefficients and frequencies for the real and imaginary parts of
+    the bath correlation function.
```
Ericgig, line 156: expansion is vague, it could be a taylor expansion, fourrier expansion, since nothing is said, it's up to the reader to guess... If you can add a reference to the paper, it would be great.
Otherwise, the equation should be found somewhere in the documentation, just something like: 
`correlation(t) = sum( ck_real * np.exp(-vk_real*t) +1j  ck_imag * np.exp(-vk_imag*t) )`
would help.

Ericgig (changes requested):
I still have doubt about `BathExponent` or more precisely, where the expansion data and operators are stored. But it could be optimized later if ram issues appear. 
I would also suggest to add a reference to the paper in every docstring that reference the expansion.
Otherwise, the pre-made Bath are nice and seems easy to use and the process of creating `rhs` is quite easy to follow.

I would not merge before adding a check for scipy's ode failure and fixing the tests (lowering tolerance in options) instead of ignoring warning.

qutip/nonmarkov/bofin_solvers.py
hodgestar: We ended up going with making `exp.Q` always the bath coupling operator and having the HEOMSolver calculate the daggered operators for `+` modes where needed.

qutip/tests/test_bofin_solvers.py
hodgestar: Results corrected & tolerances fixed.

qutip/tests/test_bofin_solvers.py
hodgestar: Thanks for catching this. Fix added.

qutip/nonmarkov/bofin_solvers.py
```diff
@@ -0,0 +1,1248 @@
...
+                    e_result = (rho * e_op).tr()
+                else:
+                    e_result = e_op(t, ado_state)
+                output.expect[e_key].append(e_result)
+
+        self.progress_bar.finished()
+        return output
+
+
+class BosonicHEOMSolver(HEOMSolver):
```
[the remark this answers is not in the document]
hodgestar: Good suggestion -- `BosonicHEOMSolver` and `FermionicHEOMSolver` made sense before I added the `BosonicBath` and `FermionicBath` and made the `HEOMSolver` support both, but now they are unnecessary.

I have removed them.

qutip/nonmarkov/bofin_baths.py
hodgestar: The paper isn't published yet, but I've added equations to the docstrings.

qutip/nonmarkov/dlheom_solver.py
hodgestar: Thank you! These are important checks to have. Checks and tests for them added.

hodgestar:
@nwlambert & @Ericgig Thank you for the many reviews. I think I have addressed all the comments. Have a final look and them if you and the CI robots are happy, I can merge! :D

Ericgig (approved):
With the check for the integration success, one the tests passes, I am happy with it.
The `@pytest.mark.filterwarnings("ignore::scipy.integrate.IntegrationWarning")` are no longer useful in tests since they happen when the integrator fails which you catch now.

hodgestar:
Tests passed! Merging! Yay, at last!

PR #1603, opened 2021-07-07
Add the missing drift Hamiltonian to the method `run_analytically` of `Processor`

### LINKED ISSUE #1602: qutip.qip.device.processor.Processor.run_analytically does not include drift
When calculating the propagators analytically it does not include the drift hamiltonian, which can result in a significantly different operator. This is particularly relevant to me as the drift hamiltonian changes depending on which qubit is being controlled because the basis is changed to a different rotating frame. This means I replace the drift when operating a pulse. If the drift is not included in the propagator calculation it results in a substantially incorrect matrix.

I think there should be an include_drift=True/False option in the function definition which if true adds the drift hamiltonian before it exponentiates.

BoxiLi:
**Description**
In the analytical calculation of the circuit unitaries in `Processor`, the drift Hamiltonian was missing. Add it and add a test accordingly. Also fix a wrong docstring.

**Related issues or PRs**
fix #1602 

**Changelog**
Add the missing drift Hamiltonian to the method `run_analytically` of `Processor`

### DIFF
--- a/qutip/qip/device/processor.py
+++ b/qutip/qip/device/processor.py
@@ -637,19 +637,25 @@ class Processor(object):
 
         Returns
         -------
-        evo_result: :class:`qutip.Result`
-            An instance of the class
-            :class:`qutip.Result` will be returned.
+        U_list: list
+            A list of propagators obtained for the physical implementation.
         """
         if init_state is not None:
             U_list = [init_state]
         else:
             U_list = []
         tlist = self.get_full_tlist()
-        # TODO replace this by get_complete_coeff
         coeffs = self.get_full_coeffs()
+
+        # Compute drift Hamiltonians
+        H = 0
+        for drift_ham in self.drift.drift_hamiltonians:
+            H += drift_ham.get_qobj(self.dims)
+
+        # Compute control Hamiltonians
         for n in range(len(tlist)-1):
-            H = sum([coeffs[m, n] * self.ctrls[m]
+            H += sum(
+                [coeffs[m, n] * self.ctrls[m]
                     for m in range(len(self.ctrls))])
             dt = tlist[n + 1] - tlist[n]
             U = (-1j * H * dt).expm()
@@ -658,7 +664,9 @@ class Processor(object):
 
         try:  # correct_global_phase are defined for ModelProcessor
             if self.correct_global_phase and self.global_phase != 0:
-                U_list.append(globalphase(self.global_phase, N=self.N))
+                U_list.append(globalphase(
+                    self.global_phase, N=self.num_qubits)
+                )
         except AttributeError:
             pass
 
--- a/qutip/qip/device/processor.py
+++ b/qutip/qip/device/processor.py
@@ -648,13 +648,13 @@ class Processor(object):
         coeffs = self.get_full_coeffs()
 
         # Compute drift Hamiltonians
-        H = 0
+        H_drift = 0
         for drift_ham in self.drift.drift_hamiltonians:
-            H += drift_ham.get_qobj(self.dims)
+            H_drift += drift_ham.get_qobj(self.dims)
 
         # Compute control Hamiltonians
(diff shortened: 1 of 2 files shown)


qutip/qip/device/processor.py
```diff
@@ -637,19 +637,25 @@ def run_analytically(self, init_state=None, qc=None):
...
+        # Compute drift Hamiltonians
+        H = 0
+        for drift_ham in self.drift.drift_hamiltonians:
+            H += drift_ham.get_qobj(self.dims)
+
+        # Compute control Hamiltonians
         for n in range(len(tlist)-1):
-            H = sum([coeffs[m, n] * self.ctrls[m]
+            H += sum(
+                [coeffs[m, n] * self.ctrls[m]
```
hodgestar, line 658: This code seems odd to me. Previously we added a separate `U` based on a separate `H` for each time `t`, but here we add to `H` each time. Do we perhaps need to create an `H_drift` above and then do `H = H_drift + sum([...])` in each loop iteration here?

qutip/qip/device/processor.py
```diff
@@ -658,7 +664,9 @@ def run_analytically(self, init_state=None, qc=None):
 
         try:  # correct_global_phase are defined for ModelProcessor
             if self.correct_global_phase and self.global_phase != 0:
-                U_list.append(globalphase(self.global_phase, N=self.N))
+                U_list.append(globalphase(
+                    self.global_phase, N=self.num_qubits)
+                )
         except AttributeError:
```
hodgestar, line 670: Not directly related to this PR, but what is the `AttributeError` that can be raised here?

qutip/tests/test_processor.py
```diff
@@ -324,11 +324,19 @@ def testDrift(self):
         Test for the drift Hamiltonian
         """
         processor = Processor(N=1)
-        processor.add_drift(sigmaz(), 0)
-        tlist = np.array([0., 1., 2.])
-        processor.add_pulse(Pulse(identity(2), 0, tlist, False))
+        processor.add_drift(sigmax() / 2, 0)
+        tlist = np.array([0., np.pi])
```
hodgestar, line 328: Could we also add a test with more than two items in tlist? Such a test would catch the bug above, I think.

hodgestar (changes requested):
Reviewed!

qutip/qip/device/processor.py
BoxiLi: Oh, yes you are right.

qutip/qip/device/processor.py
BoxiLi: `global_phase` is only defined in some subclass but not in the parent class, like this one. It is only calculated if the pulses are compiled from gates manually. There is no global phase returned if one uses the optimal control algorithms for example. So it not always defined. But I think we can just define it to be 0 so that there is no need for error handling.

hodgestar: Thanks! Let's leave that for another PR.

qutip/tests/test_processor.py
hodgestar: Thanks!

hodgestar (approved):
Happy for this to be merged once the robots are done running their tests. :)

hodgestar:
@BoxiLi Just letting you know that I added this to https://github.com/qutip/qutip/milestone/19 so that we remember to include the fix in the 4.6.3 release.

BoxiLi:
An unrelated test fails after this is merged. It doesn't make sense to me...
https://github.com/qutip/qutip/runs/3008054620

hodgestar:
> An unrelated test fails after this is merged. It doesn't make sense to me...
> https://github.com/qutip/qutip/runs/3008054620

I am very confused too. Could we try re-run it and see what happens?

BoxiLi:
Yeah, let's try.

PR #1604, opened 2021-07-13
improvements in state_number handling

jfeist:
**Description**
Speed up and simplify `state_number_enumerate`, `state_number_index`, `state_index_number`, `state_number_qobj`, and `enr_state_dictionaries`. I've changed quite a few functions in states.py, but all the changes are small. It seemed easier to combine this into a single pull request instead of splitting it up into many smaller ones (but this could be done).

The changes in more detail:
 - `state_number_enumerate` and `state_index_number` now always return tuples. Before, `state_number_enumerate` returned arrays for `excitations is None` and tuples otherwise, and `state_index_number` returned a list. For `excitations is None`, I realized that a simple call to `itertools.product` is enough, and is about 20x faster for a (relatively big) test case. It would be trivial to keep returning arrays instead of tuples here as before, but this slows the code down (quite significantly, in fact) and, in any case, returning arrays or tuples depending on whether `excitations` is set seems like an inconsistent interface, so I decided to make it consistent (and faster). I checked that the uses in the code base should not be affected. The algorithm used when the number of excitations is restricted is a slightly more elegant version of the one from my previous pull request (#1594), and is faster by another factor of 5 or more for the "big" test cases I looked at. ~~It is still recursive, but quite fast, and I did not find an elegant non-recursive algorithm.~~ I came up with an even faster non-recursive algorithm after all. It's arguably a bit less elegant, but I think still readable, and another 50% or so faster than the previous one (see the second commit below).
 - `enr_state_dictionaries` now returns a dictionary and a list instead of two dictionaries. Before, `idx2state` was a dictionary with integer keys from 0 to nstates-1, which behaves essentially like a list. However, before Python 3.7, iteration order was not guaranteed to be insertion order (i.e., iterating over the dictionary could give the indices and states in any order, and it seems to me that `enr_thermal_dm`, which did exactly that, was not guaranteed to work).
- `state_number_index` and `state_index_number` now just call `np.ravel_multi_index` and `np.unravel_index`, which are exactly the same operations (mapping between linear and multi-index in a multidimensional array). These are faster and also include error checking. The previous code had no error checks (such as `state` being allowed for the given dims). 

Since the behavior of the code has changed, but the differences should not be noticeable in most cases, I'm not sure if I should do anything else. 

**Related issues or PRs**
This is in some sense a follow-up of #1594. 

**Changelog**
Speed up state_number_enumerate, state_number_index, state_index_number, and add some error checking
Made enr_state_dictionaries return a list for idx2state.

### DIFF
--- a/qutip/operators.py
+++ b/qutip/operators.py
@@ -859,7 +859,7 @@ def enr_destroy(dims, excitations):
     a_ops = [sp.lil_matrix((nstates, nstates), dtype=np.complex128)
              for _ in range(len(dims))]
 
-    for n1, state1 in idx2state.items():
+    for n1, state1 in enumerate(idx2state):
         for idx, s in enumerate(state1):
             # if s > 0, the annihilation operator of mode idx has a non-zero
             # entry with one less excitation in mode idx in the final state
--- a/qutip/states.py
+++ b/qutip/states.py
@@ -11,6 +11,7 @@ import numbers
 import numpy as np
 from numpy import arange, conj, prod
 import scipy.sparse as sp
+import itertools
 
 from qutip.qobj import Qobj
 from qutip.operators import destroy, jmat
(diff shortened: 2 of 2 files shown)


Ericgig (approved):
Thank you!
The speed up of `state_number_enumerate` from the last released version is a good 10x and the code is a lot cleaner. The use of `itertools.product` and `numpy` function is also great.

jfeist:
Thanks! I just noticed that there was an unnecessary term in the outer loop (since idx is always the last dimension there, state[idx+1:] is the empty tuple). Removing this gives another small speedup, on the order of a few percent, but mostly it's cleaner.

hodgestar (approved):
Thanks for taking this on. It looks great.

All that's left to decide I think is whether this should go into 4.7 or 4.6.3.

I think the removal of the `state`, `idx` and `nexc` parameters from `state_number_enumerate` is fine to include in 4.6.3, since they were internal anyway. I think returning tuples is okay for 4.6.3 too because tuples were already returned in the `excitations=None` case.

Making idx2state a list makes a lot of sense and `idx2state[idx]` still works the same, but changing a list to a dictionary is a bit change (e.g. maybe users did `for idx, state in idx2state.items(): ...`), so perhaps that should go into 4.7 only?

@Ericgig Any thoughts on these questions from your side?

Ericgig:
I would go for 4.7.

hodgestar:
Merged and marked for inclusion in 4.7.

Thanks again @jfeist!

PR #1609, opened 2021-07-14
Allow f(t, **args) as coefficient function alternative

Ericgig:
**Description**
With Coefficient, supporting func(t, **kwargs) in parallel to f(t, args) for function coefficient is mostly free. 
This signature is more 'pythonic' and would probably by appreciated if available. 
With this new scheme, `t` is positional and the `**kwargs` is needed even if unused.
For backward compatibility, we should not remove the old format.

Example:
```
def f(t, args):
    return t*args['w']

def g(t, w, **kwargs):
    return t*w

qevo_old = QobjEvo([qeye(3), f], args={'w':1})

qevo_new = QobjEvo([qeye(3), g], args={'w':1})

for t in any_ts:
    qevo_old(t) == qevo_new(t)
```

**Changelog**
For time-depedent function, f(t, **args) is available with

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -24,6 +24,7 @@ from .data import Data
 from .interpolate import Cubic_Spline
 from .cy.coefficient import (InterpolateCoefficient, InterCoefficient,
                              StepCoefficient, FunctionCoefficient,
+                             KwFunctionCoefficient,
                              ConjCoefficient, NormCoefficient,
                              ShiftCoefficient, StrFunctionCoefficient,
                              Coefficient)
@@ -103,7 +104,13 @@ def coefficient(base, *, tlist=None, args={}, args_ctypes={},
         return coeff_from_str(base, args, args_ctypes, compile_opt)
 
     elif callable(base):
-        op = FunctionCoefficient(base, args.copy())
+        try:
+            # Try f(t, **kwargs)
+            base(0, **args, _checkkwargs=True)
+            op = KwFunctionCoefficient(base, args.copy())
+        except TypeError:
+            # Fallback on f(t, args)
+            op = FunctionCoefficient(base, args.copy())
         if not isinstance(op(0), numbers.Number):
             raise TypeError("The coefficient function must return a number")
         return op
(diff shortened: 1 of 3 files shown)


hodgestar:
This is a really great idea! Since we're putting this into version 5, I have a suggestion to take things a small step further and try the following:

Let's have only one function element type that accepts functions of the form `f(t, ...)` where `...` is any signature that does not include `POSITIONAL_ONLY` arguments (i.e. no `/`) or a `VAR_POSITIONAL` argument (i.e. no `*args`) and is implemented as follows:

When the function element is constructed, we use `inspect.signature(f)` to:

* raise an error if any `POSITIONAL_ONLY` or `VAR_POSITIONAL` arguments are present
* determine the list of `POSITIONAL_OR_KEYWORD` argument
* determine whether a `VAR_KEYWORDS` argument is present

When the function arguments are updated:

* if `VAR_KEYWORDS` was present, we just keep all arguments
* if not, set the arguments to just `{k: new_args[k] for k in argument_list}`

When the function element is called:

* we can just just call `f(t, **args)` as before, because `args` already contains either just the named arguments we want or all the arguments if a `VAR_KEYWORDS` argument was present in f.

Thoughts?

Ericgig:
I other words, make 
```
def f(t, w):
    return t*w

coefficient(f, args={'w':0})
```
just work using inspect...
Not free like this but doable and the function accepted as coefficient would be quite clean.
Lets to it.

Ericgig:
Function based coefficient now support all of these:
`f(t, args)`, `f(t)`, `f(t, w, ..., k=2, ...)`, `f(t, w, ..., **kwargs)`.
The `t` is positional, then we would support any number of keyword arguments. The **kwargs is not used, but can be included. Arguments with default are valid and do not need to be set at creation: `f(t, w=1); coefficient(f, args={})` is accepted and replacing `args` to add a value for `w` will work as expected.

hodgestar:
The tests fail because if one defines `f(t, args, k)` and defines `g = partial(f, k=0.5)` then the signature of g is `<Signature (t, args, *, k=5)>`. There will likely be other cases like this (e.g. many decorators will change the signature).

I vote that we give up on all the complicated detection of the old syntax and simply ask users to change `f(t, args)` to `f(t, **args)` when they upgrade to QuTiP 5.

Ericgig:
That would be punishing long time qutip user and making it hard for package depending on qutip to support both v4 and v5. 
We should do our best to support both fully knowing that it will never be perfect. The new method require inspecting the signature anyway,

hodgestar:
@Ericgig I decided it was easiest to try out some of my ideas for how to simplify this by coding. The result is #1633. Let me know what you think.

PR #1610, opened 2021-07-14
auto_tidyup options can be set per data type

Ericgig:
**Description**
`Qobj` tidyup after each mathematical operation it the option is set to `True`. This is expected when using sparse matrices, but not as useful for Dense data. This also made it a priority to implement it for new data layer while not doing any of the visible work.

This allows to set `auto_tidyup` setting per data type.

**Changelog**
Allow `auto_tidyup` settings per data type.

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -126,7 +126,15 @@ def _tidyup(method):
     @functools.wraps(method)
     def out(*args, **kwargs):
         out = method(*args, **kwargs)
-        if isinstance(out, Qobj) and settings.core['auto_tidyup']:
+        tidy_type = []
+        if isinstance(settings.core['auto_tidyup'], list):
+            tidy_type = [_data.to.parse(_type)
+                         for _type in settings.core['auto_tidyup']]
+        if (
+            isinstance(out, Qobj) and
+            (settings.core['auto_tidyup'] is True
+            or type(out.data) in tidy_type)
+        ):
             out.tidyup()
         return out
     return out
@@ -1744,6 +1752,7 @@ class Qobj:
     def isherm(self):
         if self._isherm is not None:
             return self._isherm
+        self.tidyup()
         self._isherm = _data.isherm(self._data)
         return self._isherm
 
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1752,7 +1752,6 @@ class Qobj:
     def isherm(self):
         if self._isherm is not None:
             return self._isherm
-        self.tidyup()
         self._isherm = _data.isherm(self._data)
         return self._isherm
 
--- a/qutip/core/data/properties.pyx
+++ b/qutip/core/data/properties.pyx
@@ -51,13 +51,13 @@ cdef bint _isherm_csr_full(CSR matrix, double tol) except 2:
     cdef idxint row, ptr_a, ptr_b, col_a, col_b
     for row in range(matrix.shape[0]):
         ptr_a, ptr_a_end = matrix.row_index[row], matrix.row_index[row + 1]
-        ptr_b, ptr_b_end = transpose.indptr[row], transpose.indptr[row + 1]
+        ptr_b, ptr_b_end = transpose.row_index[row], transpose.row_index[row + 1]
         while ptr_a < ptr_a_end and ptr_b < ptr_b_end:
             # Doing this on every loop actually involves a few more
             # de-references than are strictly necessary, but just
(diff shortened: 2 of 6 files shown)


hodgestar:
I'm wondering whether setting `auto_tidyup` per data type is a good idea? One expects the mathematical operations performed by different data backends to give the same results but setting `auto_tidyup` to `["CSR"]` by default will break that assumption quite badly. For example, matrices might become singular when sparse and not when dense.

There is also a danger of making some benchmarks artificially fast. If we're going to compare backends, we should probably look at both backends with tidyup on or off.

Is there a reason that tidyup cannot just be specialized for the new backends that need it?

qutip/core/qobj.py
```diff
@@ -1744,6 +1752,7 @@ def iscptp(self):
     def isherm(self):
         if self._isherm is not None:
             return self._isherm
+        self.tidyup()
```
hodgestar, line 1755: I feel `isherm` should not call `.tidyup()` -- I would be very surprised to discover that a method called `isX` changed the underlying data of a `Qobj`.

Ericgig: `_isherm_csr_full` does not work without it, but it's a bug, not a choice for faster computation. It should totally be fixed there.

Ericgig:
`auto-tidyup` exist to remove 0s stored in sparse matrices.  `auto-tidyup` is quite messy, rounding all values under 1e-12 to 0 per default after each matrix operation can only cause problems. 

As for the benchmark, why ask for dense operations to add an extra step to increase their outputs error to the same level as sparse operations so their benchmark do not look to good in comparisons. I think it is fair that the sparse operation's timing include the time needed to manage it's structure. 

For auto-differentiation, if operations included are done with `Qobj`, it would have to differentiate tidy-up steps after each operation. But since the tidy-up operation does not have a clean derivative (`out[out < 1e-12] = 0`) it will make the auto-differentiation very inefficient.

I see `auto-tidyup` more as a necessary evil to keep sparse matrix efficient. I would argue that we should not have `auto-tidyup` in `Qobj` and have sparse matrix operation clean themselves instead.

AGaliciaMartinez:
I would like to add that at this point CSR and Dense already _dno't_ return the same result in every operations. For example, for matmul:
```python
size = 100
random = np.random.random((size, size))
a = qt.data.Dense(random)
b = qt.data.to['CSR'](a)
np.all((a).to_array() == (b).to_array()) # True
np.all((a+a).to_array() == (b+b).to_array()) # True
#  This one is different for Dense and CSR even though the operation is the same. 
# This seems to be due to the order at which the  operations are performed with Dense
#  and CSR matmul and the fact that floating point addition is not associative. 
# (something I learned by reading the test_mathematics in qutip)
np.all((a@a).to_array() == (b@b).to_array()) # False.
```
Although the change in `auto_tidyup` will probably make bigger the difference, given that with tidy-up "all" we can not ensure the same output for every operation, I would say that it should be ok to set auto_tidyup to `[CSR]`.

jakelishman:
Yeah, auto-tidyup should be a function of CSR matrices only.  It doesn't make sense for dense matrices to have it, that's just burning cycles.  As Asier mentioned, you already can't rely on floating-point operations being exactly identical between different matrix types, so there's no particular problem there.  The point about auto-differentiation is an excellent reason that it must be allowed to vary between types as well.

To be honest, I'm not sure how many cases the auto-tidyup operation actually helps in keeping things efficient.  It may well be worth us having a proper investigation, and seeing how often it actually helps.  I can tweak the addition and matrix-multiplication sparse algorithms to automatically tidy up when generating their output (if they don't already...), in which case we could probably entirely remove all "automatic" tidyup, and just have it be a manual operation?

`isherm_csr_full` is actually explicitly meant to handle all cases, _especially_ those involving matrices that have not been tidied up.  The logic is actually sound (it's been deployed in `master` for a while now), but looks like there's a typo made when I handled the merge up, and the data-layer is missing a few of its mathematical tests so it didn't get spotted.  I'll fix that now.

jakelishman:
Fixed the bad merge in #1614.

hodgestar:
I'm +1 on having auto tidy up being a feature of the individual data layers. Could the auto tidy up setting perhaps be passed down to the data layers and then each layer can handle it in a way that seems sensible to the data layer?

Ericgig:
The consensus seems to be removing the `auto-tidyup` from `Qobj` and have it included in `CSR` operations.
I am closing this and will make a new one in that direction.

hodgestar:
Thanks @Ericgig!

Issue #1607, opened 2021-07-14
Change current behaviour of __mul__ and __rmul__. 


AGaliciaMartinez:
In qutip-tensoflow, we aim to include TensorFlow's autodifferentiation with Qobj. However, this will require the following code to work:
```python
import numpy as np
import qutip

random_matrix = np.random.random((2, 2)) + 1j * np.random.random((2, 2))
complex_number = np.random.random((1, 1)) + 1j * np.random.random((1, 1))

qobj = qutip.Qobj(random_matrix)
qobj*complex_number
``` 
I used numpy `ndarray` instead of a `tf.Tensor` or `tf.Variable` for simplicity and so that the example is reproducible. 

Currently this example raises: 
"TypeError: incompatible dimensions [[2], [2]] and [[1], [1]]"
 as `__mul__(self, other)` defaults to `__matmul__(self, other)` when `other` is not an instance of `numbers.Number`. 

If instead we do:
```python
complex_number*qobj
```
The error raised is:
"TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'Qobj'"
This error is different because `__rmul__` checks first for `other` being an instance of `numbers.Number` and if not, it raises `NotImplementedError`. It does not check whether `Qobj(other)` works. I would say this is undesired and that `__rmul__` should try first converting `other` into `Qobj` and not assume that other is already `Qobj`. 

**Describe the solution you'd like**
I would like to suggest either:
1) changing the behaviour of `matmul` dispatcher so that (1,1) shaped data is multiplied element wise by default. However this, although not terrible, it breaks conceptually what matmul is, as a (1,1) x (2,2) matrix multiplication can not happen.

2) Allow `mul` dispatcher to have as input a `Data` object. In this way we would raise an error when the `Data` for `mul` dispatcher is not of shape (1,1) and perform an elementwise multiplication when it is. I am not sure how difficult it is to code this.

3) Create `mul_elementwise` dispatcher to have as input two `Data` objects and perform elementwise multiplication.

Another change that I would like to suggest is to add a check in `__rmul__` to see if `other` can be represented as a Qobj. 

These two changes should probably go in a different PR.

_Edit: added third option_

jakelishman:
I'm not sure about any of these options:
[...]

PR #1611, opened 2021-07-15
Fixed __rmul__ to be consistent with __mul__ behaviour.

AGaliciaMartinez:
**Description**
Before this fix the following two situations were incosinsten:
```python
matrix = np.random.random((2, 2))
qobj = qutip.Qobj(array)

qobj * matrix # This would perform a matrix multiplication as `__mul__` tries to convert matrix into `Qobj`.
matrix * qobj # This would raise an error as `__rmul__` assumed that matrix was a `Qobj`.
```

I assumed that `qobj * matrix` behaviour is the correct one as it is explicitly programmed to work like that. With this fix `matrix * qobj` will first try to convert matrix to `Qobj` and perform a matmul between matrix and `Qobj`.

**Related issues or PRs**
Issue #1607  is related to this although this PR does not close that issue.

**Changelog**
__rmul__ now accepts as `other` an array_like that is understood by `Qobj`.

### DIFF
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -292,6 +292,43 @@ def test_QobjMultiplication():
 
     assert q3 == q4
 
+def test_QobjNumpyMultiplication():
+    """qutip.Qobj multiplication times another object(ndarray). We test both
+    left multiplication and right multiplication."""
+    data1 = np.array([[1, 2], [3, 4]])
+    data2 = np.array([[5, 6], [7, 8]])
+
+    result1 = np.dot(data1, data2)
+    result1 = qutip.Qobj(result1)
+
+    result2 = np.dot(data2, data1)
+    result2 = qutip.Qobj(result2)
+
+    q1 = qutip.Qobj(data1)
+    expect1 = q1 * data2
+    expect2 = data2 * q1
+
+    assert result1 == expect1
+    assert result2 == expect2
+
+def test_QobjMulNotImplmented():
+    """We test that if Qobj does not know how to handle an object it defaults to
+    the objects it raises NotImplmented, which in turn should raise TypeError as
+    DummyClass does not have __mul__ or __rmul__."""
+    class DummyClass():
+        pass
+
+    dummy_object = DummyClass()
+    qobj = qutip.Qobj(1)
+
+    with pytest.raises(TypeError):
+        dummy_object * qobj
+
+    with pytest.raises(TypeError):
+        qobj * dummy_object
+
+
+
 
 def test_QobjDivision():
     "qutip.Qobj division"
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -292,6 +292,7 @@ def test_QobjMultiplication():
 
     assert q3 == q4
 
+
 def test_QobjNumpyMultiplication():
     """qutip.Qobj multiplication times another object(ndarray). We test both
     left multiplication and right multiplication."""
@@ -311,10 +312,11 @@ def test_QobjNumpyMultiplication():
     assert result1 == expect1
     assert result2 == expect2
 
+
 def test_QobjMulNotImplmented():
-    """We test that if Qobj does not know how to handle an object it defaults to
-    the objects it raises NotImplmented, which in turn should raise TypeError as
-    DummyClass does not have __mul__ or __rmul__."""
+    """We test that if Qobj does not know how to handle an object it defaults
+    to the objects it raises NotImplmented, which in turn should raise
+    TypeError as DummyClass does not have __mul__ or __rmul__."""
     class DummyClass():
         pass
 
@@ -326,8 +328,6 @@ def test_QobjMulNotImplmented():
 
     with pytest.raises(TypeError):
         qobj * dummy_object
-
-
 
 
 def test_QobjDivision():
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -499,7 +499,6 @@ class Qobj:
 
         return other.__matmul__(self)
 
-
     @_tidyup
     def __matmul__(self, other):
         if not isinstance(other, Qobj):
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -487,11 +487,18 @@ class Qobj:
                     copy=False)
 
     def __rmul__(self, other):
-        # Shouldn't be here unless `other.__mul__` has already been tried, so
-        # we _shouldn't_ check that `other` is `Qobj`.
-        if not isinstance(other, numbers.Number):
+        if isinstance(other, numbers.Number):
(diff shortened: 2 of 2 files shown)


qutip/tests/core/test_qobj.py
```diff
@@ -293,6 +293,43 @@ def test_QobjMultiplication():
...
+    dummy_object = DummyClass()
+    qobj = qutip.Qobj(1)
+
+    with pytest.raises(TypeError):
+        dummy_object * qobj
+
+    with pytest.raises(TypeError):
+        qobj * dummy_object
+
+
```
AGaliciaMartinez, line 332: I am not sure if this is the correct way of testing this. I guess `TypeError` could be raised by other reasons. However, I have not found an explicit way of testing for a `NotImplemented` return in pytest.

jakelishman:
I don't think this is the right way to solve your problem, but we probably should be consistent.  `Qobj` is not a container for arbitrary data, and a 1x1 `Qobj` basically should never exist.  If I remember correctly, the auto-promotion to `Qobj` is to avoid breaking backwards compatibility, because 4.x doesn't use the `@` syntax at all.  We almost certainly should be moving to remove that "feature", not add to it - it should never work with compound Hilbert spaces (we should be safe-by-default), and trying to detect special behaviour on 1x1 inputs just leads to weird edge cases and breakages all over the place.

You probably want some sort of different mechanism for deciding what types a `Qobj.data` instance can accept in `mul`.  At the moment, the test is `isinstance(other, numbers.Number)`, but you want to accept `tensorflow.Variable` if the backing data is TF-like.  I think the right solution would be based on having some way for downstream classes to expose what types they can accept, perhaps by a well-defined try/except pattern?

Ericgig:
I agree we need to be consistent, but I would also go with `matrix * Qobj` to be right and change `Qobj * array` to raise an error.
`Qobj * Qobj` can be understood as `Qobj @ Qobj` but everything else could use `mul`.  If `mul` raise a `TypeError`, we can catch it and return `NotImplemented`. With this, we probably won't need to touch the `Dense` and `CSR` `mul` function and `Variable` will be passed to the `TfTensor` specialization.
The only issue I see is that sometime `Variable * Qobj` will work, other time it will not exist.

jakelishman:
Having `Qobj * matrix` work at the moment is to make it easier to transition to the 5.x branch when it releases, since it's the current behaviour of 4.x.  Once we've had a bit of time for people to make the transition, I think we should be completely removing this "functionality".  To me, matrices aren't quantum objects; if you're working with a Numpy array, you should expect Numpy semantics to occur, i.e. Numpy should see `Qobj` as a scalar type, and broadcast against it like that.  It's an implementation detail that the backing data is a matrix; the linear algebra theory certainly doesn't demand it.

Raising-and-catching is slower than just having the `mul` functions return `NotImplemented`.  The only implementations that would need a wrapper are our Cython-based ones, and they can both be wrapped in something like
```python
class only_allow_numeric:
    def __init__(self, inner):
        self.inner = inner
    def __call__(self, other):
        if not isinstance(other, numbers.Number):
            return NotImplemented
        return self.inner(complex(other))
        
mul.add_specialisations([
    (CSR, CSR, only_allow_numeric(mul_csr)),
    (Dense, Dense, only_allow_numeric(mul_dense)),
])
```

I'm generally fine with `Variable * Qobj` only working sometimes; I think the general intent is that you'd configure the dispatchers to only allow TensorFlow data output, so it would always work for you.

jakelishman:
Also, I've just noticed - do Americans spell it "specializations" with a z?  If so, we should go through the data layer and make sure I'm not committing us to any British spellings of anything, just for consistency in the code.

Ericgig:
I would just cut remove it with the v5 release, it will be harder to make breaking changes later and with the removal of `__array__`, we already change the way `Qobj` interact with `numpy`. But if we keep it for a while, we should give a warning when used.

My spellcheck correct to "specializations".

hodgestar:
Thoughts:
* Using `qobj * qobj` is really nice because it matches the mathematical convention that the multiplication used is the one that makes sense for the mathematical objects involved.
* One option is to have `qobj * qobj` always do the same as `qobj @ qobj` if both are Qobjs, and scalar multiplication if not (are there any other cases that make sense?)
* Another option is to have `qobj * qobj` issue a deprecation warning recommending `qobj @ qobj` and have `x * qobj` always do scalar multiplication later.

I think we should take this decision to the qutip-admin meeting. We don't need to wait for that to merge PRs, but we should be okay with the full admin team asking for code to be reverted potentially if as a group we pick a different decision that whatever decision with come to in GitHub PRs.

hodgestar:
During today's GSoC progress call @Ericgig, @AGaliciaMartinez and I thought that:

* `Qobj.__mul__(self, obj)` should:
   * dispatch to `__matmul__` if `obj` is a `Qobj`.
   * dispatch to `__mul__` otherwise.
* The data layer implementation of `__mul__` should treat `obj` as a scalar, but `Qobj` should not change the type of `obj` so that, e.g., the tensorflow data layer, can accept non-complex types (e.g. TensorFlow uses 1x1 matrixes for scalars).

jakelishman:
I agree with this, with some additional points:
1. `QobjEvo` also needs to be handled by `__matmul__`, but this should be handled by returning `NotImplemented`, and letting `QobjEvo.__rmul__` dispatch it to `QobjEvo.__rmatmul__`.
1. however we choose to do this, `__mul__` should return `NotImplemented` if the other element can't be handled correctly; it shouldn't raise an exception by itself.  The "Pythonic" way is to try/catch an exception in handling, but this just means we need to be sure that all `__mul__` methods do proper error checking, and _do_ reject invalid input rather than just silently doing weird stuff.
1. what do we want to do in response to being multiplied with Numpy arrays?  We have to be careful, because otherwise we might do very different things depending on whether we're being multiplied from the left or right by the Numpy array, which isn't really desirable.  My first instinct is to say we should make ourselves appear as a scalar type to Numpy, so it broadcasts the multiplication over the scalar values in its array.  To do that, we may need to play a little with `__array__`, `__array_ufunc__` and `__array_function__`, and return `NotImplemented` from `Qobj.__mul__(array)`.  I think this maybe gives the most useful behaviour?  I can't imagine a situation where it would make sense for a data-layer type to accept a Numpy array in a scalar-multiplication routine, and doing things this way allows people to use the broadcasting support of Numpy for convient looping.

Just for posterity, I'm really strongly against any automatic promotion to `Qobj`, with the exception of handling scalar numbers in `__add__` and `__sub__`, where it's mathematical convention that they mean the relevant identity operator scaled by the scalar.  Having that is also necessary to have `sum()` work with `Qobj` ergonomically.  I know that's not the suggestion here, just thought I'd mention it again!

hodgestar:
@AGaliciaMartinez Is this PR deprecated in favour of #1620? If so, would you close it? I keep looking at both :see_no_evil:.

AGaliciaMartinez:
Yes, this one should be closed.

PR #1612, opened 2021-07-15
Fixed bug where copy argument was not being passed to create dispatcher.

AGaliciaMartinez:
**Description**
Qobj's copy argument was not being passed to create dispatcher. This resulted in a copy of `data` regardless of the `copy` argument when `data` was not an instance of `data.Data` or `Qobj`.

Example (I am using here qutip-tensorflow as I do not know how this translates into numpys ndarray, see comment below):
```python
tfconstant = tf.constant(np.random.random((2,2))) # random tensor
qobj = qutip.Qobj(tfconstant, copy=False) 
print(qobj.data._tf is not tfconstant) # True but should be False
```

**Changelog**
Copy argument is now handled correctly.

### DIFF
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -76,7 +76,7 @@ def test_QobjData():
     assert isinstance(q2.data, qutip.core.data.Data)
 
 
-@pytest.mark.parametrize("data",
+@pytest.mark.parametrize("original_data",
                          [
                             qutip.data.Dense(_random_not_singular(2)),
                             qutip.data.csr.identity(2),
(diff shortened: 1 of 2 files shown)


qutip/tests/core/test_qobj.py
```diff
@@ -76,6 +76,38 @@ def test_QobjData():
...
+    """Tests that Qobj copy argument works properly when instantiating Qobj."""
+    qobj_data = qutip.Qobj(data, copy=copy).data
+
+    if isinstance(data, qutip.Qobj):
+        data = data.data  # Qobj copies the data of another Qobj
+
+    # For numpy object we compare with data's data. This should be dense so we
+    # get its data as ndarray.
+    if isinstance(data, np.ndarray):
+        qobj_data = qobj_data.as_ndarray()
```
AGaliciaMartinez, line 103: This does not seem to work. I am not sure when it should work or how to make it work. What I was expecting is that if the `ndarray` is passed with `copy=False` then I should be able to retrieve it somehow from `qobj.data`. I guess I could check that operating in `qobj` does not change input `data`? Any ideas of how this should be done @jakelishman?

jakelishman: `copy=False` refers to the data, not the entire object.  This allows us to avoid copies even when the shape is being changed (say you started with a 1D array for a ket, or whatever). You can check that the data wasn't copied here with `np.shares_memory`:
```python
>>> x = np.arange(5, dtype=complex)
>>> # This is explicitly a no-op in terms of data, but creates a new Python object.
>>> y = x.reshape(x.shape, order='A')
>>> x is y
False
>>> # But they still point to the same place in memory.
>>> np.shares_memory(x, y)
True
```

Similar things happen inside `data.Dense` in a few places, to allow the initialiser to do its job.

AGaliciaMartinez: Thanks @jakelishman for the clear explanation! It now works!

AGaliciaMartinez:
I am very surprised to see a decrease in the coverage. In any case, this is ready for review.

qutip/tests/core/test_qobj.py
```diff
@@ -1023,3 +1060,4 @@ def test_contract(expanded, contracted, inplace):
     assert out.dims == contracted
     assert out.shape == qobj.shape
     assert np.all(out.full() == qobj.full())
+
```
hodgestar, line 1063:
```suggestion
```
Removing blank line that was accidentally added at the end of the file.

hodgestar (approved):
I requested a small clean up (removing a newline that was added at the end of the test file), but the change looks good to me. Happy for it to be merged once the newline is removed.

Ericgig (approved):

PR #1614, opened 2021-07-18
Fix bad attribute accesses in isherm_csr

jakelishman:
Fix a couple of incorrect attribute accesses in `_isherm_csr_full`, which were incorrectly merged up from `master` - they refer to the old SciPy names for the fields, rather than the new `CSR` type names.  This wasn't caught at the time because the tests hadn't been merged up yet, so I've written those as well.

**Changelog**
- Fix typos in `isherm_csr`
- Add tests of `isherm` for all data types.


hodgestar (approved):

hodgestar:
Woot. And thanks for adding the tests!

jakelishman:
It's interesting to see how my style of writing tests changes and matures.  It's being particularly influenced at the moment, because on Qiskit we use a fairly thin wrapper around raw `unittest`, rather than `pytest`, so I'm being exposed to writing in other styles as well.

PR #1615, opened 2021-07-19
Move auto-tidyup from Qobj to CSR function

Ericgig:
**Description**
The `auto-tidyup` step is moved from `Qobj` to `CSR` operations as discussed in #1610.
Most methods which supported `auto-tidyup` still do, with the exceptions of does based on scalar product.
`__matmul__` and `__add__` already had a `if value != 0` check which was changed to support a tolerance.
This new conditions affect both the real and imaginary part together:
Previously : `[1+1e-15j, 1e-15+1e-15j]` -> `[1+0j, 0]`
Now :  `[1+1e-15j, 1e-15+1e-15j]` -> `[1+1e-15j, 0]`
This can be changed if desired.

`expm_csr` was the only function that is not covered by these changes. 
I added a `tidyup` keyword to `CSR.__init__` for it instead of using a decorator.
It could be dropped if we judge `tidyup` is not needed for `expm`.

**Related issues or PRs**
#1610 

**Changelog**
Move auto-tidyup from Qobj to CSR function


qutip/core/data/add.pyx
```diff
@@ -51,6 +52,10 @@ cdef idxint _add_csr(Accumulator *acc, CSR a, CSR b, CSR c) nogil:
     """
     cdef idxint row, ptr_a, ptr_b, ptr_a_max, ptr_b_max, nnz=0, col_a, col_b
     cdef idxint ncols = a.shape[1]
+    cdef double tol
+    with gil:
+        if settings.core['auto_tidyup']:
+            tol = settings.core['auto_tidyup_atol']
```
jakelishman, line 58: It would be better to pass in the tolerance value in the call from `add_csr`, which holds the GIL, rather than playing yoyo with it here (and below).

qutip/core/qobj.py
```diff
@@ -1055,7 +1032,7 @@ def unit(self, inplace=False, norm=None, kwargs=None):
         """
         norm = self.norm(norm=norm, kwargs=kwargs)
         if inplace:
-            self.data /= norm
+            _data.imul(self.data, 1/norm)
```
jakelishman, line 1035: This is unrelated to the PR - probably best not include it.  It's also not necessarily correct; for example, it'll make data backends with autodifferentiation slower/less accurate, because you're masking the true operation.  It should probably be up to the data backends to decide whether this is an acceptable modification.

Also, this is a good illustration of my reservations about inplace operations on the data layer; we said that they're not guaranteed to be in-place, so you have to assign the result.

qutip/core/data/csr.pxd
```diff
@@ -103,7 +104,10 @@ cdef inline base.idxint acc_gather(Accumulator *acc, double complex *values, bas
     for i in range(acc.nnz):
         position = acc.nonzero[i]
         value = acc.values[position]
-        if value != 0:
+        if (
+            value != 0 and
+            (tol == 0 or fabs(value.real) >= tol or fabs(value.imag) >= tol)
+        ):
```
jakelishman, line 110: why not
```cython
value = acc.values[position]
if fabs(value.real) < tol:
    value.real = 0
if fabs(value.imag) < tol:
    value.imag = 0
if value != 0:
    ...
```
(or alternatively the same way it's currently in `tidyup_csr`).  The way it's written in the PR would allow a value to pass through unaltered if _either_ of its real and imaginary parts are above the tolerance, but that's inconsistent with the behaviour of `tidyup_csr`.

jakelishman (changes requested):
If this is totally removing calls to `tidyup`, do `tidyup.pyx` and `tidyup.pxd` still need to exist, along with the relevant dispatcher?

hodgestar:
Re tidyup dispatcher: I think this should still exist:

* It's useful for implementing efficient explicity tidyup calls on the different data layers.
* It's useful for specializations on data layer implementations if they want to do, e.g., `__matmul__(self, other): other = other.tidyup() ...` at the start of an operation (not 100% sure this is a good example -- probably other.tidyup() and conversion to sparse needs to happen in one step to happen efficiently, but I feel like there are cases like this out there).

qutip/core/data/csr.pxd
Ericgig: Yes, but do we want the current `tidyup` behaviour?
If the aim is to remove small value to keep the matrix sparse, there is no point in cleaning the `real` and `imag` part individually. We only introduce extra numerical error in the operation.
The present `tidyup` help to have nice looking output: `1` looks better than `1+1e-15j`, but other than the look, I don't understand why we round the real and imaginary part independently...

And now that `tidyup` in a method of `Qobj` explicitly called by the user, do `CSR` operations need to be consistent with it? I am tempted to have the `auto_tidyup_atol` and default `tidyup` tolerance being different values and rename the options to make those 2 distinct.

jakelishman: I could see the full `tidyup` going either way, but at least for the purposes of this PR, which isn't changing `tidyup_csr`, the methods should be the same - it would be surprising if not, I think.

An argument in favour of keeping it separate for `real` and `imag` is that it makes the `atol` argument have a proper meaning; if you allow a value if _either_ is above the threshold because you're considering the complex number as a whole, then the true maximum absolute value that can get tidied is `sqrt(2) * atol`, and there are numbers with an absolute value less than that which wouldn't get tidied up (`1.1*atol`, for example).  If you want the absolute value to work properly, you end up with 3 more flops per nonzero (though you save one comparison).  The current method is faster; you can also do it with one CPU-vectorised pass over the data reinterpreted as double-precision floats, where each operation is a conditional move (`cmp [x] y; mov 0 [x]`, but one CPU instruction), which (if I remember correctly) are typically immune to branch mispredication.

hodgestar:
@Ericgig Just checking the status of this PR. Should we update it to the latest dev.major and then I can review it again?

Ericgig:
@hodgestar I resolved the conflicts.

qutip/core/options.py
```diff
@@ -40,5 +40,5 @@ class CoreOptions:
         # general relative tolerance
         "rtol": 1e-12,
         # use auto tidyup absolute tolerance
-        "auto_tidyup_atol": 1e-12
+        "auto_tidyup_atol": 1e-14
```
hodgestar, line 43: Why did the default tolerance change?

qutip/core/options.py
```diff
@@ -25,8 +25,8 @@ class CoreOptions:
         general relative tolerance
         Used to choose QobjEvo.expect output type
 
-    auto_tidyup_atol : float {1e-12}
-        use auto tidyup absolute tolerance
+    auto_tidyup_atol : float {1e-14}
+        The absolute tolerance used in automatic tidyup.
```
hodgestar, line 29: This is also the default value used in explicit calls to `Qobj.tidyup` so we should mention that in the docstring too.

qutip/tests/test_gates.py
```diff
@@ -53,7 +53,7 @@ def _remove_global_phase(qobj):
     Return a new Qobj with the gauge fixed for the global phase.  Explicitly,
     we set the first non-zero element to be purely real-positive.
     """
-    flat = qobj.full().flat.copy()
+    flat = qobj.tidyup(1e-14).full().flat.copy()
```
hodgestar, line 56: Wondering why this got set explicitly here? Not a bad idea to set it explicitly, just wondering if the test failed without it.

hodgestar (changes requested):
I left a few small questions / comments, but generally looks good to me.

hodgestar:
I have some slight reservations about the copying around of the tidyup logic in the CSR methods, but it's not in too many places so far, so I'm happy with the duplication for now. If we get more copies, I would recommend we define an inline method that does the `if real < tol: ... if imag < tol: ...` dance for us.

qutip/tests/test_gates.py
Ericgig: The matrices here can be `Dense`, and it set the phase to the first non-zero.
So without the `auto-tidyip` in `Qobj`, the first element could be 1e-30 and the global phase would not be properly set.

qutip/core/options.py
Ericgig: Now that the cleaning step is in the operation itself, it will probably be called more often in loops so I expect the numerical error will accumulate faster. This is to balance that.

hodgestar: Sounds good.

qutip/core/options.py
```diff
@@ -25,8 +25,9 @@ class CoreOptions:
...
+        Also used as the default value do :method:`Qobj.tidyup`.
```
hodgestar, line 30:
```suggestion
        The absolute tolerance used in automatic tidyup (see the ``auto_tidyup``
        parameter above) and the default value of `atol` used in
        :method:`Qobj.tidyup`.
```

hodgestar (approved):
I left one suggestion for improved wording in the settings documentation, but happy for it to be merged as soon as that's applied.

hodgestar (approved):
@Ericgig Happy for this to be merged if the test failure is unrelated.

PR #1619, opened 2021-07-22
More carefully track the state of the Bloch sphere figure and axes

### LINKED ISSUE #1616: Bloch sphere not showing on b.show() command, only flashes window briefly
When attempting to create (and then draw on) the Bloch sphere, the command b.show() as per the documentation (after doing the correct setup) does not display the Bloch sphere. Rather a matplotlib window appears for a fraction of a second before closing itself.
Using plt.pause() after the b.show() line allows this window to remain open however it is simply black with nothing rendered on it.
I am able to get a graph of the Bloch sphere by using the plt.show() command in place of b.show(). However this also brings up another window at the same time showing only a set of 3d axes.

Here is my MWE

```python
import qutip

b = qutip.Bloch()

b.make_sphere()
b.render()

```
This is the line I can replace with plt.show()
```
b.show()
```

As per the documentation I would expect b.show() to be the only command required to display the Bloch sphere. I have attempted to install correctly but I assume I am missing something still.

My environment is:
QuTiP Version:      4.6.2
Numpy Version:      1.21.1
Scipy Version:      1.6.2
Cython Version:     0.29.24
Matplotlib Version: 3.0.3
Python Version:     3.7.1
Number of CPUs:     4
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
Installation path:  C:\ProgramData\Anaconda3\lib\site-packages\qutip

Thanks

hodgestar:
Previously .render(...) took fig and axes arguments and overwrote self.fig and self.axes if the arguments were None. The actual values of the arguments were completely discarded (!!). In addition, not supplying a value for fig and axes would overwrite the values passed to the constructor as Bloch(..., fig=fig, axes=axes).

The new behaviour is as follows:

* If an external figure is passed to the constructor it is retained.

* If no figure is supplied, .render() will create a figure if one has not been created yet, or if the figure has been closed, will recreate it. As a special case, it will *not* recreate *closed* figures inside Jupyter notebooks using the inline backend, because that closes figures at the end of every cell and displays closed figures quite happily. 

* .render() will always call self.fig.canvas.update() to ensure that any existing figure that is already shown will be updated.

* .show() will now always render the figure and call `IPython.display.display` (if running in an IPython terminal) or `fig.plot()` otherwise.

Previously it was possible to cause .render() to raise obscure exceptions by, e.g., calling .render(), closing the figure, calling
.render() again. See, for example, https://github.com/qutip/qutip/issues/1616.

Previously when running in an IPython shell `.show()` would have to be called twice to have any effect if no figure was passed in. Now it only needs to be called once.

The external figure tracking was backported from Qiskit's copy of Bloch and was inspired by PR https://github.com/Qiskit/qiskit-terra/pull/1091/.

**Related issues or PRs**

* fixes #1616
* inspired by https://github.com/Qiskit/qiskit-terra/pull/1091/
* issue with `.show()` under `IPython` mentioned in #1690

**Changelog**

More carefully track the state of the Bloch sphere figure and axes to prevent exceptions during rendering.


hodgestar:
@Ericgig @nwlambert @jakelishman I'm hoping one of you has a chance to review. This makes the Bloch sphere plotting significantly more robust for me, so it would be nice to merge it.

qutip/bloch.py
```diff
@@ -414,20 +418,21 @@ def run_from_ipython(self):
...
+                self.fig = None
+                self.axes = None
+
+        if not self._ext_fig and self.fig is None:
             self.fig = plt.figure(figsize=self.figsize)
 
-        if not axes:
+        if not self._ext_axes and self.axes is None:
             self.axes = _axes3D(self.fig, azim=self.view[0],
                                 elev=self.view[1])
```
Ericgig, line 437: What happen if `axes` are given but not `fig`? The `if not self._ext_fig:` would overwrite `self.axes` with `None` but they would not be recreated since `if not self._ext_axes and self.axes is None` will be `False`.

hodgestar: Good point. Let me figure out what to do.

qutip/bloch.py
```diff
@@ -112,7 +112,9 @@ def __init__(self, fig=None, axes=None, view=None, figsize=None,
                  background=False):
         # Figure and axes
         self.fig = fig
+        self._ext_fig = fig is not None
         self.axes = axes
+        self._ext_axes = axes is not None
```
Ericgig, line 117: I don't see `self._ext_axes` being used anywhere.

hodgestar: Thanks for catching that -- it was left over from earlier versions of the fix. Removed.

hodgestar:
@Ericgig Ready for re-review!

Ericgig:
I tried it locally and `show` print the figure twice, whether you provide a figure or not. `render` itself print the picture so having it in `show` make `display` redundant. Otherwise, it work fine.

hodgestar:
> I tried it locally and `show` print the figure twice, whether you provide a figure or not. `render` itself print the picture so having it in `show` make `display` redundant. Otherwise, it work fine.

Where were you running your example? Python console, IPython console, Jupyter notebook?

Ericgig:
Jupyter, I ran https://github.com/qutip/qutip-notebooks/blob/master/examples/bloch_sphere_with_colorbar.ipynb

hodgestar:
@Ericgig Thanks for finding that -- I obviously missed a last round of testing in Jupyter notebooks. I've added a special case for the inline backed in Jupyter notebook and even more comments.

The downside is that the whole situation with matplotlib is crazy and the current Bloch implementation is not making things any more sane.

The upside is that things now seem to work fairly nicely for me in practice. Holding thumbs that they work nicely for you too.

Ericgig:
It still print 2 the sphere twice when a figure is given, the `plt.close` in `render` is only in effect when fig is None.
I think a check `show` is needed:
```
if self.run_from_ipython() and not self._ext_fig:
            display(self.fig)
```

hodgestar:
> It still print 2 the sphere twice when a figure is given, the `plt.close` in `render` is only in effect when fig is None. I think a check `show` is needed:
> 
> ```
> if self.run_from_ipython() and not self._ext_fig:
>             display(self.fig)
> ```

I think it's okay if when the user supplies the figure themselves they can close it themselves if they want to?

Adding the `not self._ext_fig` check here would be wrong. The external figure would only be displayed after the cell in which is was created, and `.show()` would never cause it to be displayed.

The figure is only displayed twice because the user explicitly created an open figure in the same cell they called `.show()`. I don't want to close the users figure for them (that might not be what they want) and they can always not call `.show()` in that cell or close the figure themselves.

Happy to think about a better solution or perhaps just to document this more in the docstring of `.show()` or `.render()`.

Thoughts?

Ericgig (approved):
Since it only happen if the figure is opened in the same cell as `show`, I don't see it causing any issues. A note in the documentation would be nice.

hodgestar:
Thanks! Note added to the docstring of `.show()`.

PR #1622, opened 2021-07-23
Pow tests for data layer.

AGaliciaMartinez:
**Description**
Added tests for pow specialisation. I also changed the raised error to `ValueError` to be consistent with the rest of the specialisations (and in particular with trace that also raises `ValueError` if input matrix is not square). 

The tests do not contemplate the cases where:
- n is not an integer (float): Float is cast to integer so: `pow_csr(matrix, 0.1) == pow_csr(matrix, 0)`. We may want to raise `TypeError` instead.
- n is not positive: `OverflowError` is raised instead of `TypeError`. This is due to how Cython works but we may want to raise TypeError for consistency with the rest of the code.

For instance, raising `TypeError` in these cases is what NumPy does. What should we do?

**Related issues or PRs**
None

**Changelog**
Give a short description of the PR in a few words. This will be shown in the QuTiP change log after the PR gets merged.
For example: 
Tests for pow specisalisation.
Pow now raises ValueError instead of TypeError if matrix is not square.

### DIFF
--- a/qutip/tests/core/data/test_mathematics.py
+++ b/qutip/tests/core/data/test_mathematics.py
@@ -93,13 +93,17 @@ def shapes_binary_bad_matmul(dim=100):
 
 
 def shapes_square(dim=100):
+    """Allowed shapes for operations that require square matrices. Examples of
+    these operations are trace, pow, expm and the trace norm."""
     return [
         (pytest.param((1, 1), id="1"),),
         (pytest.param((dim, dim), id=str(dim)),),
     ]
 
 
-def shapes_bad_square(dim=100):
+def shapes_not_square(dim=100):
+    """Disallowed shapes for operations that require square matrices. Examples of
+    these operations are trace, pow, expm and the trace norm."""
     return [
         (x,) for x in shapes_unary(dim) if x.values[0][0] != x.values[0][1]
     ]
@@ -766,7 +770,7 @@ class TestTrace(UnaryOpMixin):
         return np.sum(np.diag(matrix))
 
     shapes = shapes_square()
-    bad_shapes = shapes_bad_square()
+    bad_shapes = shapes_not_square()
     specialisations = [
         pytest.param(data.trace_csr, CSR, complex),
         pytest.param(data.trace_dense, Dense, complex),
@@ -787,7 +791,7 @@ class TestPow(UnaryOpMixin):
         return np.linalg.matrix_power(matrix, n)
 
     shapes = shapes_square()
-    bad_shapes = shapes_bad_square()
+    bad_shapes = shapes_not_square()
     specialisations = [
         pytest.param(data.pow_csr, CSR, CSR),
     ]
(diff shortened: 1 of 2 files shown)


qutip/tests/core/data/test_mathematics.py
```diff
@@ -92,6 +92,18 @@ def shapes_binary_bad_matmul(dim=100):
...
+        (pytest.param((1, 1), id="1"),),
+        (pytest.param((dim, dim), id="100"),),
+    ]
+
+
+def shapes_bad_square(dim=100):
+    return [
+        (x,) for x in shapes_unary() if x.values[0][0] != x.values[0][1]
+    ]
+
```
Ericgig, line 106: Why the `dim` input? In `shapes_square`, if it is use, the `id` will not reflect it. In `shapes_bad_square`, it is not used.

qutip/tests/core/data/test_mathematics.py
```diff
@@ -92,6 +92,18 @@ def shapes_binary_bad_matmul(dim=100):
     ]
 
 
+def shapes_square(dim=100):
```
hodgestar, line 95: Could we add a docstring for `shapes_square`?

qutip/tests/core/data/test_mathematics.py
```diff
@@ -92,6 +92,18 @@ def shapes_binary_bad_matmul(dim=100):
...
+    return [
+        (pytest.param((1, 1), id="1"),),
+        (pytest.param((dim, dim), id=str(dim)),),
+    ]
+
+
+def shapes_bad_square(dim=100):
```
hodgestar, line 102: Could we add a docstring for `shapes_bad_square`?

qutip/tests/core/data/test_mathematics.py
```diff
@@ -92,6 +92,18 @@ def shapes_binary_bad_matmul(dim=100):
...
+    return [
+        (x,) for x in shapes_unary(dim) if x.values[0][0] != x.values[0][1]
+    ]
+
```
AGaliciaMartinez, line 106: - add small docstring square.
- change bad -> not

hodgestar (commented):
Left a request for some short docstrings for the new shape list functions, but looks good otherwise.

AGaliciaMartinez:
I implemented the suggested changes. Given that the cases where n is negative or a floating point number are already handled in `Qobj`, it may not be necessary to handle them here. I think this is ready to be merged.

qutip/tests/core/data/test_mathematics.py
```diff
@@ -775,6 +786,44 @@ def test_incorrect_shape_raises(self, op, data_m):
...
+            assert test.shape == expected.shape
+            np.testing.assert_allclose(test.to_array(), expected,
+                                       atol=self.tol)
+        elif out_type is list:
+            for test_, expected_ in zip(test, expected):
+                assert test_.shape == expected_.shape
+                np.testing.assert_allclose(test_.to_array(),
+                                           expected_, atol=self.tol)
+        else:
+            assert abs(test - expected) < self.tol
```
Ericgig, line 815: We know that the output is a `Data` object. 
```suggestion
        assert test.shape == expected.shape
        np.testing.assert_allclose(test.to_array(), expected,
                                   atol=self.tol)
```

Ericgig (approved):

hodgestar (approved):

PR #1625, opened 2021-07-29
Added clarification in eigen docstring.

AGaliciaMartinez:
Added clarification for what `low` and `high` mean when sorting eigenvalues.


hodgestar (approved):
Thanks @AGaliciaMartinez!

PR #1627, opened 2021-08-02
Fixed deprecation warning from matplotlib.

AGaliciaMartinez:
**Description**
Fixed a deprecation warning from matplotlib. The fix was suggested by the warning.

**Changelog**
Fixed deprecation warning from matplotlib.

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -54,7 +54,7 @@ try:
 
         def draw(self, renderer):
             xs3d, ys3d, zs3d = self._verts3d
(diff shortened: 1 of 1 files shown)


jakelishman:
This is already fixed on `master` and awaiting merge, see #1556.  This fix isn't entirely suitable, because it isn't compatible with older vesrions of matplotlib, which we still support.

jakelishman:
Looking again, this particular line of the fix passed review last time, and it was other deprecation warnings that involved more threading of the needle in deprecated versions.  Regardless, we've still got this fixed on master, and we'll pull it up whenever I finally manage to get time to completely cut out `qutip.qip` and merge.

PR #1630, opened 2021-08-04
Expm and  project tests.

AGaliciaMartinez:
**Description**
This adds a few more tests to the specialisations. I also found a Bug in expm and changed `TypeError` to `ValueError`.

- Expm: 
  - Added tests for its specialisations.
  - There was a _bug_ in expm where if the matrix was csr and diagonal, zero values where not being exponentiated as they did not appear in scipy's data attribute. This case is now handled properly with the same scaling in resources (O(N) where N is the non-cero diagonal elements in the matrix).
  - Incorrect shape now raises ValueError.
 
- UnaryOpMixin: 
  - It did not check for bad shapes. It now does have a function that checks for them (this avoids code duplication in `trace`, `expm` and some others). 
  - As a result of this change a hidden bug emerged: even though `TestProject` had a non-empty `bad_shapes` attribute, it was not doing anything as `UnaryOpMixin` did not generate tests for bad shapes. It now does.

- Project:
  -  Changed `TypeError` to `ValueError`.
  - Removed redundant `TestProject`

- Pow:
  - Added the case where `n=0` as it is a special case.

**Changelog**
`project` specialisations now return `ValueError` when matrix has not valid shapes.
Removed redundant `TestProject`.
Added special test case for Pow (n=0).
`UnaryOpMixin` now has test_incorrect_shapes (by default it does not have any bad shapes)
Added tests for `expm` specialisations.
Fixed bug where diagonal `CSR` matrix was not exponentiating zero values in the diagonal.
`expm` specialisations now raises `ValueError` when matrix is not square.

_Edit: Added "Removed redundant `TestProject`"_
_Edit2: Added changelog_

### DIFF
--- a/qutip/core/data/expm.py
+++ b/qutip/core/data/expm.py
@@ -4,6 +4,7 @@ import scipy.sparse.linalg
 from .dense import Dense
 from .csr import CSR
 from .properties import isdiag_csr
+from .base import idxint_dtype
 
 __all__ = [
     'expm', 'expm_csr', 'expm_csr_dense',
@@ -14,14 +15,18 @@ def expm_csr(matrix: CSR) -> CSR:
     if matrix.shape[0] != matrix.shape[1]:
         raise ValueError("can only exponentiate square matrix")
     if isdiag_csr(matrix):
-        sci = matrix.as_scipy()
-        # Note that expm1 is an elemetwise operation that maps 0->0 which
-        # involves no operation at all. Since we know that the input is
-        # diagonal, elementwise expm1 in the whole matrix has the same
-        # computational cost as doing only a expm1 in the diagonal
-        # (elementwise).
-        sci = sci.expm1() + scipy.sparse.diags(np.ones(matrix.shape[0]))
-        return CSR(sci)
+        matrix_sci = matrix.as_scipy()
+        data = np.ones(matrix.shape[0], dtype=np.complex128)
+        data[matrix_sci.indices] += np.expm1(matrix_sci.data)
+        return CSR(
+            (
+                data,
+                np.arange(matrix.shape[0], dtype=idxint_dtype),
+                np.arange(matrix.shape[0] + 1, dtype=idxint_dtype),
+            ),
+            shape=matrix.shape,
+            copy=False,
+        )
     # The scipy solvers for the Pade approximant are more efficient with the
     # CSC format than the CSR one.
     csc = matrix.as_scipy().tocsc()
--- a/qutip/core/data/expm.py
+++ b/qutip/core/data/expm.py
@@ -12,12 +12,16 @@ __all__ = [
 
 def expm_csr(matrix: CSR) -> CSR:
     if matrix.shape[0] != matrix.shape[1]:
-        raise TypeError("can only exponentiate square matrix")
+        raise ValueError("can only exponentiate square matrix")
     if isdiag_csr(matrix):
-        out = matrix.copy()
-        sci = out.as_scipy()
-        sci.data[:] = np.exp(sci.data)
-        return out
+        sci = matrix.as_scipy()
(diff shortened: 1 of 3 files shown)


AGaliciaMartinez:
I

AGaliciaMartinez:
Sorry for closing and reopening. It was not my intention to close this PR.

qutip/core/data/expm.py
```diff
@@ -12,12 +12,16 @@
...
-        sci.data[:] = np.exp(sci.data)
-        return out
+        sci = matrix.as_scipy()
+        # Note that expm1 is an elemetwise operation that maps 0->0 which
+        # involves no operation at all. Since we know that the input is
+        # diagonal, elementwise expm1 in the whole matrix has the same
+        # computational cost as doing only a expm1 in the diagonal
+        # (elementwise).
+        sci = sci.expm1() + scipy.sparse.diags(np.ones(matrix.shape[0]))
+        return CSR(sci)
```
jakelishman, line 24: This is a good catch, thanks.  We could consider optimising this algorithm a little - it involves several allocations, three instantiations of the slow `scipy.sparse.csr_matrix`, and one CSR addition, which aren't super ideal.  I would imagine something like:
```python
        matrix_sci = matrix.as_scipy()
        data = np.ones(matrix.shape[0], dtype=np.complex128)
        data[matrix_sci.indices] += np.expm1(matrix_sci.data)
        return _data.CSR(
            (
                data,
                np.arange(matrix.shape[0], dtype=_data.base.idxint_dtype),
                np.arange(matrix.shape[0] + 1, dtype=_data.base.idxint_dtype),
            ),
            shape=matrix.shape,
            copy=False,
        )
```
would be more efficient.  We could avoid a couple more allocations if we rewrote this in Cython as well (we wouldn't need to allocate in the `np.expm1` call and we could avoid a couple of internal `CSR.__init__` checks), but that's probably not really necessary.  I would expect the overhead of this method to be reasonably low anyway.

AGaliciaMartinez: I can go  with this implementation. Thanks for the input!

jakelishman: Please do make sure it's right, and it's probably worth doing a proper benchmark! The one I wrote _looks_ right to me  and it makes rather fewer allocations (plus numpy instantiations are cheaper), so I expect it it's noticeably faster, but I'm not certain.

qutip/tests/core/data/test_mathematics.py
```diff
@@ -406,6 +407,16 @@ def test_mathematically_correct(self, op, data_m, out_type):
         else:
             assert abs(test - expected) < self.tol
 
+    def test_incorrect_shape_raises(self, op, data_m):
+        """
+        Test that the operation produces a suitable error if the shapes of the
```
hodgestar, line 412:
```suggestion
        Test that the operation produces a suitable error if the shape of the
```

hodgestar (changes requested):
@AGaliciaMartinez Thanks! I left one small docstring suggestion.

Could you update the changelog entry in the description to include the rest of the changes made and then I can merge?

AGaliciaMartinez:
@hodgestar changelog updated.

hodgestar (approved):

qutip/tests/core/data/test_mathematics.py
```diff
@@ -816,6 +802,18 @@ def test_incorrect_shape_raises(self, op, data_m):
             op(data_m(), 10)
 
 
+class TestExpm(UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return sc.sparse.linalg.expm(matrix)
```
jakelishman, line 807: I didn't catch this in time, but there's no need to use `scipy.sparse.linalg` here; the matrix is guaranteed to be a Numpy array by the test sutie.

PR #1631, opened 2021-08-04
Tests for norm

AGaliciaMartinez:
**Changelog**
Added tests for Norm.
`norm.trace` raises `ValueError` instead of `TypeError` when the input is not a square matrix.


AGaliciaMartinez:
It seems I messed up with the new `ValueError`. Im trying to figure out what is wrong.

AGaliciaMartinez:
Trace norm is well defined for square matrices.

jakelishman:
Really, the trace norm should be defined for non-square matrices as well.  It's just the sum of the singular values, and that's still well defined.

PR #1632, opened 2021-08-05
Fix bugs in spin_q_function and spin_wigner, add test cases

maij:
**Description**

- Fixed some errors in the transcription of the husimi Q and Wigner W functions for spins into code.
- Ensure Q and W functions are normalized over (theta, phi).
- Add tests for correctness and normalization of spin Q function.
- Add tests for normalization and real-ness of spin W function. Correctness is established by comparing the overlap of two states and the overlap of their Wigner functions.
- Spin Q function tests take ~11s to complete, spin Wigner tests take ~10s to complete on my laptop.

**To do**
- [x] Decide if the `spin_wigner` function should return only real values be default (as is currently done for the `spin_q_function`).
- [x] Find a simple test for correctness of the spin W function.
- [x] Decide whether to integrate the W function normalization in `_rho_kq` rather than at the top level. We divide all terms by `(2*j + 1)` only to end up multiplying it back out at the end which is a waste of time (`_rho_kq` is called `d**2` times, `d=2*j+1`) . The benefit of leaving it as-is is that it reads like most definitions in literature.
- [x] Update docstring for spin Q function.
- [x] Decide if normalization factor for Wigner and spin Q function should be included (Davis et al. suggest the normalization factor is only introduced during integration: Eq 14 https://arxiv.org/abs/2008.10167)

**Related issues or PRs**
This issue was raised in another PR #1195 which didn't resolve this issue as intended. The original issue this fixes was #1128.

**Changelog**
Fixed error for calculation of Husimi `spin_q_function` and `spin_wigner`.
Added tests to `tests/test_wigner.py` for these functions.

### DIFF
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -1,8 +1,10 @@
 import pytest
 import numpy as np
+import itertools
 from scipy.special import laguerre
 from numpy.random import rand
-from numpy.testing import assert_, run_module_suite, assert_equal
+from numpy.testing import assert_, run_module_suite, assert_equal, \
+    assert_almost_equal
 
 import qutip
 from qutip.states import coherent, fock, ket, bell_state
@@ -593,6 +595,21 @@ def test_wigner_clenshaw_sp_iter_dm():
         Wdiff = abs(W - Wclen)
         assert_equal(np.sum(abs(Wdiff)) < 1e-7, True)
 
+@pytest.mark.slow
+def test_spin_q_function():
+    for spin in [1/2, 3, 11/2, 21]:
+        d = int(2*spin + 1)
+        rho = rand_dm(d)
 
+        # Points at which to evaluate the spin Q function
+        theta_prime = np.linspace(0, np.pi, 32, endpoint=True)
+        phi_prime = np.linspace(-np.pi, np.pi, 64, endpoint=True)
+        Q, _, _ = qutip.spin_q_function(rho, theta_prime, phi_prime)
+
+        Q = Q.transpose()
+        for k, (t, p) in enumerate(itertools.product(theta_prime, phi_prime)):
+            state = qutip.spin_coherent(spin, t, p)
+            direct_Q = (state.dag() * rho * state).norm() / np.pi
+            assert_almost_equal(direct_Q, Q.flat[k], decimal=9)
 if __name__ == "__main__":
     run_module_suite()
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -606,7 +606,7 @@ def test_wigner_clenshaw_sp_iter_dm():
     pytest.param(False, id="mixed")
 ])
 def test_spin_q_function(spin, pure):
-    d = int(2 * spin + 1)
+    d = int(2*spin + 1)
     rho = rand_dm(d, pure=pure)
 
     # Points at which to evaluate the spin Q function
@@ -653,7 +653,7 @@ def test_spin_q_function_normalized(spin, pure):
     pytest.param(False, id="mixed")
 ])
 def test_spin_wigner_normalized_and_real(spin, pure):
-    d = int(2 * spin + 1)
+    d = int(2*spin + 1)
     rho = rand_dm(d, pure=pure)
 
     # Points at which to evaluate the spin Wigner function
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -616,7 +616,7 @@ def test_spin_q_function(spin, pure):
 
     for k, (phi_prime, theta_prime) in enumerate(itertools.product(phi, theta)):
         state = qutip.spin_coherent(spin, theta_prime, phi_prime)
-        direct_Q = (2 * spin + 1) / (4 * np.pi) * (state.dag() * rho * state).norm()
+        direct_Q = (state.dag() * rho * state).norm()
         assert_almost_equal(Q.flat[k], direct_Q, decimal=9)
 
 @pytest.mark.parametrize(['spin'], [
@@ -638,7 +638,7 @@ def test_spin_q_function_normalized(spin, pure):
     phi = np.linspace(-np.pi, np.pi, 1024, endpoint=True)
     Q, THETA, _ = qutip.spin_q_function(rho, theta, phi)
 
-    norm = np.trapz(np.trapz(Q * np.sin(THETA), theta), phi)
+    norm = d / (4 * np.pi) * np.trapz(np.trapz(Q * np.sin(THETA), theta), phi)
     assert_almost_equal(norm, 1, decimal=5)
 
 
@@ -652,7 +652,7 @@ def test_spin_q_function_normalized(spin, pure):
     pytest.param(True, id="pure"),
     pytest.param(False, id="mixed")
 ])
-def test_spin_wigner_normalized_and_real(spin, pure):
+def test_spin_wigner_normalized(spin, pure):
     d = int(2*spin + 1)
     rho = rand_dm(d, pure=pure)
 
@@ -661,9 +661,7 @@ def test_spin_wigner_normalized_and_real(spin, pure):
     phi = np.linspace(-np.pi, np.pi, 512, endpoint=True)
     W, THETA, PHI = qutip.spin_wigner(rho, theta, phi)
 
-    assert_allclose(W.imag, 0, atol=1e-9)
-
-    norm = np.trapz(np.trapz(W * np.sin(THETA), theta), phi)
+    norm = np.trapz(np.trapz(W * np.sin(THETA) * np.sqrt(d / (4*np.pi)), theta), phi)
     assert_almost_equal(norm, 1, decimal=4)
 
 @pytest.mark.parametrize(['spin'], [
@@ -690,7 +688,7 @@ def test_spin_wigner_overlap(spin, pure, n=10):
         state_overlap = (test_state*rho).tr().real
 
         W_state, _, _ = qutip.spin_wigner(test_state, theta, phi)
-        W_overlap = (4 * np.pi / (2 * spin + 1)) * np.trapz(
+        W_overlap = np.trapz(
             np.trapz(W_state * W * np.sin(THETA), theta), phi).real
         assert_almost_equal(W_overlap, state_overlap, decimal=4)
 
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -616,7 +616,7 @@ def test_spin_q_function(spin, pure):
 
     for k, (p, t) in enumerate(itertools.product(phi_prime, theta_prime)):
         state = qutip.spin_coherent(spin, t, p)
-        direct_Q = (state.dag() * rho * state).norm() * (2 * j + 1) / (4 * pi)
+        direct_Q = (state.dag() * rho * state).norm() * (2 * spin + 1) / (4 * np.pi)
         assert_almost_equal(Q.flat[k], direct_Q, decimal=9)
 
 @pytest.mark.parametrize(["spin"], [
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -606,7 +606,7 @@ def test_wigner_clenshaw_sp_iter_dm():
     pytest.param(False, id="mixed")
 ])
 def test_spin_q_function(spin, pure):
-    d = int(2*spin + 1)
+    d = int(2 * spin + 1)
     rho = rand_dm(d, pure=pure)
 
     # Points at which to evaluate the spin Q function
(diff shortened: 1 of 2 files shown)


hodgestar:
@maij Thanks! The test code looks like a good start. Could you add it to `qutip/tests/test_wigner.py`?

Other tests we could add:

* We could write a more general test that compares the output of spin_q_function to a simpler implementation based on the definition -- something like:
```python
    state = qutip.spin_coherent(spin, theta_prime, phi_prime)
    direct_q = (state.dag() * qutip.ket2dm(psi) * state).norm() / np.pi
```
Note: I named the angles `theta_prime` and `phi_prime` to emphasize that we can apply this test to cases where the angles for the state psi and the ones passed to spin_q_function are different.

* We should also add a test for passing in arrays of theta and phi (i.e. were we really get out a mesh grid for Q values) since that was the primary use case for which `spin_q_function` was defined.

Ideas for future PRs:

* It would be nice to also add tests for `spin_wigner` (I am suspicious of it now because of the problems with spin_q_function)

Thank you for looking at these and taking charge of this bit of QuTiP.

maij:
@hodgestar I've added some tests for the spin_q_function. At the moment it tests a random density matrix for three different spin values. I'm not very familiar with qutip's style for writing tests, I would normally write a test function that would also take arguments but that doesn't seem to be the norm here. We could also do something similar to the `TestHusimiQ` class of tests.

I've had a look at the `spin_wigner` function as well, and I believe that had the same issue of density matrix indexing.
I've re-implemented it (and the `_rho_kq` function) and I've added details to the docstrings including relevant references.
Should I create a separate PR for this or should I expand the scope of this PR?

NB: At the moment I can't think of an easy way to test the spin Wigner function because there is no simpler/alternative definition that I am aware of to verify the implementation against.
There are simple tests such as if the imaginary part is less than some threshold (currently true), does it integrate to 1 (currently not true, not sure why).

qutip/tests/test_wigner.py
```diff
@@ -593,6 +595,21 @@ def test_wigner_clenshaw_sp_iter_dm():
         Wdiff = abs(W - Wclen)
         assert_equal(np.sum(abs(Wdiff)) < 1e-7, True)
 
-
+@pytest.mark.slow
```
hodgestar, line 598: You can parameterise the spin with:
```python
@pytest.mark.parametrize(['spin'], [
    pytest.param(1/2, id="spin-one-half"),
    pytest.param(3, id="spin-three"),
    ...
])
```

hodgestar: Question: How slow is slow?

hodgestar: If it is very slow, perhaps we can just reduce the number of steps down from 32 to say, 7?

hodgestar (commented):
@maij Thanks for the test! I added a small suggestion on how to parameterize it.

I'm happy for the `spin_wigner` fixes to happen in this PR or a separate PR -- either is fine.

Let us think a bit more on how to test `spin_wigner` properly. Maybe there is a way to use the `wigner` function to generate a suitable result to compare with? If not, we can just make up some small tests with specific numbers.

qutip/tests/test_wigner.py
maij: I've tested with pytest and the `spin_q_function` and `spin_wigner` tests together take less than 1 minute to complete.

Thanks for the tip on parametrisation, I've also added an arg to test non-pure state.

maij:
@hodgestar I decided to add the `spin_wigner` changes in which I've resolved the normalization issue and indexing problems that were also in the `spin_q_function`. I've added a simple test to check for normalization and real-ness of the spin Wigner.

qutip/tests/test_wigner.py
```diff
@@ -593,6 +595,104 @@ def test_wigner_clenshaw_sp_iter_dm():
...
+    pytest.param(True, id="pure"),
+    pytest.param(False, id="mixed")
+])
+def test_spin_q_function(spin, pure):
+    d = int(2 * spin + 1)
+    rho = rand_dm(d, pure=pure)
+
+    # Points at which to evaluate the spin Q function
+    theta = np.linspace(0, np.pi, 32, endpoint=True)
+    phi = np.linspace(-np.pi, np.pi, 64, endpoint=True)
```
hodgestar, line 614:
```suggestion
    theta = np.linspace(0, np.pi, 16, endpoint=True)
    phi = np.linspace(-np.pi, np.pi, 32, endpoint=True)
```
Dropping these to 16 and 32 speeds up the tests a bit and I think is a reasonable compromise between more testing and test speed.

qutip/tests/test_wigner.py
```diff
@@ -593,6 +595,104 @@ def test_wigner_clenshaw_sp_iter_dm():
...
+    d = int(2 * spin + 1)
+    rho = rand_dm(d, pure=pure)
+
+    # Points at which to evaluate the spin Q function
+    theta = np.linspace(0, np.pi, 512, endpoint=True)
+    phi = np.linspace(-np.pi, np.pi, 1024, endpoint=True)
+    Q, THETA, _ = qutip.spin_q_function(rho, theta, phi)
+
+    norm = np.trapz(np.trapz(Q * np.sin(THETA), theta), phi)
+    assert_almost_equal(norm, 1, decimal=5)
```
hodgestar, line 642:
```suggestion
    theta = np.linspace(0, np.pi, 128, endpoint=True)
    phi = np.linspace(-np.pi, np.pi, 256, endpoint=True)
    Q, THETA, _ = qutip.spin_q_function(rho, theta, phi)

    norm = np.trapz(np.trapz(Q * np.sin(THETA), theta), phi)
    # decimal=4 is a bit rough-and-ready, but increasing the accuracy of the norm requires
    # more theta and phi steps which makes the test slower
    assert_almost_equal(norm, 1, decimal=4)
```

qutip/tests/test_wigner.py
```diff
@@ -593,6 +595,104 @@ def test_wigner_clenshaw_sp_iter_dm():
...
+
+    # Points at which to evaluate the spin Wigner function
+    theta = np.linspace(0, np.pi, 256, endpoint=True)
+    phi = np.linspace(-np.pi, np.pi, 512, endpoint=True)
+    W, THETA, PHI = qutip.spin_wigner(rho, theta, phi)
+
+    assert_allclose(W.imag, 0, atol=1e-9)
+
+    norm = np.trapz(np.trapz(W * np.sin(THETA), theta), phi)
+    assert_almost_equal(norm, 1, decimal=4)
```
hodgestar, line 667:
```suggestion
    theta = np.linspace(0, np.pi, 128, endpoint=True)
    phi = np.linspace(-np.pi, np.pi, 256, endpoint=True)
    W, THETA, PHI = qutip.spin_wigner(rho, theta, phi)

    assert_allclose(W.imag, 0, atol=1e-9)

    norm = np.trapz(np.trapz(W * np.sin(THETA), theta), phi)
    # decimal=3 is a compromise between accuracy and test speed (i.e. more theta and phi steps)
    assert_almost_equal(norm, 1, decimal=3)
```

qutip/tests/test_wigner.py
```diff
@@ -593,6 +595,104 @@ def test_wigner_clenshaw_sp_iter_dm():
...
+    pytest.param(True, id="pure"),
+    pytest.param(False, id="mixed")
+])
+def test_spin_wigner_overlap(spin, pure, n=10):
+    d = int(2*spin + 1)
+    rho = rand_dm(d, pure=pure)
+
+    # Points at which to evaluate the spin Wigner function
+    theta = np.linspace(0, np.pi, 256, endpoint=True)
+    phi = np.linspace(-np.pi, np.pi, 512, endpoint=True)
```
hodgestar, line 685:
```suggestion
    theta = np.linspace(0, np.pi, 128, endpoint=True)
    phi = np.linspace(-np.pi, np.pi, 256, endpoint=True)
```
This was faster for me, but still passed the accuracy threshold during repeated test runs.

qutip/tests/test_wigner.py
```diff
@@ -593,6 +595,104 @@ def test_wigner_clenshaw_sp_iter_dm():
...
+    assert_allclose(W.imag, 0, atol=1e-9)
+
+    norm = np.trapz(np.trapz(W * np.sin(THETA), theta), phi)
+    assert_almost_equal(norm, 1, decimal=4)
+
+@pytest.mark.parametrize(['spin'], [
+    pytest.param(1/2, id="spin-one-half"),
+    pytest.param(3, id="spin-three"),
+    pytest.param(13/2, id="spin-thirteen-half"),
+    pytest.param(7, id="spin-seven")
```
hodgestar, line 673:
```suggestion
    pytest.param(1/2, id="spin-one-half"),
    pytest.param(1, id="spin-one"),
    pytest.param(3/2, id="spin-three-half"),
    pytest.param(2, id="spin-two")
```
Decreasing the value of the spins also speeds up the tests significantly because it reduces the size of the system substantially. This might also be a good idea for the other tests (and perhaps would mean we wouldn't need to reduce the number of theta and phi steps so much).

hodgestar (changes requested):
@maij Thanks for the great tests. I've made some suggestions for speeding them up a bit.

I will review the code itself shortly.

Apologies for the delay in reviewing -- back from holiday now!

qutip/wigner.py
```diff
@@ -72,7 +72,7 @@ def wigner_transform(psi, j, fullparity, steps, slicearray):
     else:
         rho = psi
 
-    sun = 2   # The order of the SU group
+    sun = 2  # The order of the SU group
```
hodgestar, line 75: I know it's a bit of a pain, but could you undo the style clean ups from this PR? I would like to add something like black  to QuTiP to make these issues go away, but making all the clean-ups in this PR is asking for trouble (e.g. if something breaks or someone wants to read the history it will be hard for them to see what are the important changes -- indeed it is already a bit tricky for me as a reviewer).

qutip/wigner.py
```diff
@@ -882,42 +896,64 @@ def spin_q_function(rho, theta, phi):
 
     Q = np.zeros_like(THETA, dtype=complex)
 
-    for m1 in arange(-j, j+1):
-
-        Q += binom(2*j, j+m1) * cos(THETA/2) ** (2*(j-m1)) * sin(THETA/2) ** (2*(j+m1)) * \
-             rho.data[int(j-m1), int(j-m1)]
+    for m1 in arange(-j, j + 1):
```
hodgestar, line 899: P.S. I'm fine if we keep the style fixes in the  `spin_*` functions because we're rewriting those to some extent anyway.

qutip/wigner.py
```diff
@@ -850,14 +856,17 @@ def qfunc(
 # PSEUDO DISTRIBUTION FUNCTIONS FOR SPINS
 #
 def spin_q_function(rho, theta, phi):
-    """Husimi Q-function for spins.
+    r"""The Husimi Q function for spins is defined as
+    ``Q(theta, phi) = SCS.dag()* rho * SCS*(2*j+1)/(4*pi)`` for the spin
```
hodgestar, line 860:
```suggestion
    ``Q(theta, phi) = SCS.dag()* rho * SCS * (2*j+1) / (4*pi)`` for the spin
```

hodgestar:
@maij The code is looking good to me. I've requested some changes and after that I'd like to do a final review, but I think we're in good shape.

My thoughts on the remaining questions:

> Decide if the spin_wigner function should return only real values be default (as is currently done for the spin_q_function).

I'm +1 on returning real values since the Wigner function is real valued and this is more consistent (unless you're aware of any strange cases where it might somehow not be?)

> Decide whether to integrate the W function normalization in _rho_kq rather than at the top level. We divide all terms by (2*j + 1) only to end up multiplying it back out at the end which is a waste of time (_rho_kq is called d**2 times, d=2*j+1) . The benefit of leaving it as-is is that it reads like most definitions in literature.

I'm +1 on sticking to the definitions in literature unless there is a measured speed penalty. `2*j + 1` likely can't be that big and the small number of arithmetic operations required is probably dwarfed by the matrix multiplications happening. 

> Decide if normalization factor for Wigner and spin Q function should be included (Davis et al. suggest the normalization factor is only introduced during integration: Eq 14 https://arxiv.org/abs/2008.10167)

Here I would try stick as closely as possible to the previous behaviour of these functions (i.e. cases which worked before should hopefully return the same answer now) and then document clearly what the normalization factor is and how it should be added or removed.

hodgestar:
@maij Just checking in on the status of this PR.

maij:
@hodgestar Sorry, I am a bit busy at the moment and haven't found time to complete this PR. I will try and complete this now, there are only small changes left.

maij:
@hodgestar I have implemented most of your suggested changes. I reduced the largest spin used  in the spin Wigner tests, but left the spin Q function as-is because it was fast enough.

I've removed the normalization from the functions themselves and described the normalization criterion in the docstrings.
The spin Wigner is now real.

qutip/tests/test_wigner.py
hodgestar: Okay to leave these as is for now.

hodgestar (approved):
Thanks @maij, this looks great.

PR #1635, opened 2021-08-12
Norm tests for specialisations.

AGaliciaMartinez:
**Changelog**
Added specialisations for norm.
`TraceNorm` now uses `scipy.linalg.norm` with `ord='nuc` by default (sparse=False).
Fixed bug for `eigs_sparse` where if the shape was less than 4 it called with wrong arguments to `eigs_dense`.

### DIFF
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -68,7 +68,7 @@ cpdef double _trace_dense(Dense matrix) except -1:
 cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
     # For column and row vectors we simply use the l2 norm as it is equivalent
     # to the trace norm.
-    if matrix.shape[1]==1 or matrix.shape[0]==1:
+    if matrix.shape[0]==1 or matrix.shape[1]==1:
         return l2_csr(matrix)
 
     # For the sparse=False we default to scipy's nuclear norm but for
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -45,9 +45,8 @@ cpdef double one_csr(CSR matrix) except -1:
         mem.PyMem_Free(col)
 
 cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
-    """Compute the trace norm relaying only on sparse operations. This consists
-    on a digitalization of `matrix@matrix.adjoint()` to obtain the result from
-    the diagonals."""
+    """Compute the trace norm using only sparse operations. These consist
+    of determining the eigenvalues of `matrix @ matrix.adjoint()` and summing their square roots."""
 
     cdef CSR op = matmul_csr(matrix, adjoint_csr(matrix))
     cdef size_t i
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -44,9 +44,20 @@ cpdef double one_csr(CSR matrix) except -1:
     finally:
         mem.PyMem_Free(col)
 
-cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
+
+cpdef double trace_dense(Dense matrix) except -1:
+    """Compute the trace norm relaying scipy for dense operations."""
+    return scipy.linalg.norm(matrix.as_ndarray(), 'nuc')
+
+
+cpdef double trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
     """Compute the trace norm using only sparse operations. These consist
-    of determining the eigenvalues of `matrix @ matrix.adjoint()` and summing their square roots."""
+    of determining the eigenvalues of `matrix @ matrix.adjoint()` and summing
+    their square roots."""
+    # For column and row vectors we simply use the l2 norm as it is equivalent
+    # to the trace norm.
+    if matrix.shape[0]==1 or matrix.shape[1]==1:
+        return l2_csr(matrix)
 
     cdef CSR op = matmul_csr(matrix, adjoint_csr(matrix))
     cdef size_t i
@@ -62,20 +73,6 @@ cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
         total += math.sqrt(abs(eigs[i]))
     return total
 
-cpdef double trace_dense(Dense matrix) except -1:
-    """Compute the trace norm relaying scipy for dense operations."""
-    return scipy.linalg.norm(matrix.as_ndarray(), 'nuc')
-
-cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
-    # For column and row vectors we simply use the l2 norm as it is equivalent
-    # to the trace norm.
-    if matrix.shape[0]==1 or matrix.shape[1]==1:
-        return l2_csr(matrix)
-
-    if sparse:
-        return _trace_csr(matrix, tol=tol, maxiter=maxiter)
-    else:
-        return trace_dense(dense.from_csr(matrix))
 
 cpdef double max_csr(CSR matrix) nogil:
     cdef size_t ptr
@@ -201,9 +198,6 @@ one.add_specialisations([
     (CSR, one_csr),
 ], _defer=True)
 
-
-# TODO: sort out how the trace norm is handled with regards to the `sparse`
-# keyword argument and add a Dense method.
 
 trace = _Dispatcher(
     _inspect.Signature([
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -6,7 +6,7 @@ from libc cimport math
 from cpython cimport mem
 
 from scipy.linalg cimport cython_blas as blas
-from scipy.linalg import norm
+import scipy
 
 from qutip.core.data cimport CSR, Dense, csr, dense, Data
 
@@ -65,7 +65,7 @@ cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
 
 cpdef double trace_dense(Dense matrix) except -1:
     """Compute the trace norm relaying scipy for dense operations."""
-    return norm(matrix.as_ndarray(), 'nuc')
+    return scipy.linalg.norm(matrix.as_ndarray(), 'nuc')
 
 cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
     # For column and row vectors we simply use the l2 norm as it is equivalent
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -209,12 +209,6 @@ one.add_specialisations([
 trace = _Dispatcher(
     _inspect.Signature([
         _inspect.Parameter('matrix', _inspect.Parameter.POSITIONAL_OR_KEYWORD),
-        _inspect.Parameter('sparse', _inspect.Parameter.POSITIONAL_OR_KEYWORD,
-                           default=False),
-        _inspect.Parameter('tol', _inspect.Parameter.POSITIONAL_OR_KEYWORD,
-                           default=0),
-        _inspect.Parameter('maxiter', _inspect.Parameter.POSITIONAL_OR_KEYWORD,
-                           default=None),
     ]),
     inputs=('matrix',),
     name='trace',
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -45,7 +45,9 @@ cpdef double one_csr(CSR matrix) except -1:
         mem.PyMem_Free(col)
 
 cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
-    """Compute the trace norm reliying only on sparse operations."""
+    """Compute the trace norm relaying only on sparse operations. This consists
+    on a digitalization of `matrix@matrix.adjoint()` to obtain the result from
+    the diagonals."""
 
     cdef CSR op = matmul_csr(matrix, adjoint_csr(matrix))
     cdef size_t i
@@ -61,8 +63,8 @@ cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
         total += math.sqrt(abs(eigs[i]))
     return total
 
-cpdef double _trace_dense(Dense matrix) except -1:
-    """Compute the trace norm reliying scipy for dense operations."""
+cpdef double trace_dense(Dense matrix) except -1:
+    """Compute the trace norm relaying scipy for dense operations."""
     return norm(matrix.as_ndarray(), 'nuc')
 
 cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
@@ -71,12 +73,10 @@ cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
     if matrix.shape[0]==1 or matrix.shape[1]==1:
         return l2_csr(matrix)
 
-    # For the sparse=False we default to scipy's nuclear norm but for
-    # sparse=True we compute eigenvalues and get the norm from them.
     if sparse:
         return _trace_csr(matrix, tol=tol, maxiter=maxiter)
     else:
-        return _trace_dense(dense.from_csr(matrix))
+        return trace_dense(dense.from_csr(matrix))
 
 cpdef double max_csr(CSR matrix) nogil:
     cdef size_t ptr
@@ -229,6 +229,7 @@ trace.__doc__ =\
     """
 trace.add_specialisations([
     (CSR, trace_csr),
+    (Dense, trace_dense),
 ], _defer=True)
 
 
--- a/qutip/core/data/norm.pyx
+++ b/qutip/core/data/norm.pyx
@@ -6,6 +6,7 @@ from libc cimport math
 from cpython cimport mem
 
 from scipy.linalg cimport cython_blas as blas
+from scipy.linalg import norm
 
 from qutip.core.data cimport CSR, Dense, csr, dense, Data
 
@@ -43,21 +44,15 @@ cpdef double one_csr(CSR matrix) except -1:
     finally:
         mem.PyMem_Free(col)
 
-cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
-    # We use the general eigenvalue solver which involves a Python call, so
-    # there's no point attempting to release the GIL.
+cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
+    """Compute the trace norm reliying only on sparse operations."""
+
     cdef CSR op = matmul_csr(matrix, adjoint_csr(matrix))
     cdef size_t i
     cdef double [::1] eigs
 
-    # For column and row vectors we simply use the l2 norm.
-    if matrix.shape[1]==1 or matrix.shape[0]==1:
-        return l2_csr(matrix)
+    eigs = eigs_csr(op, isherm=True, vecs=False, tol=tol, maxiter=maxiter)
 
-    if sparse:
-        eigs = eigs_csr(op, isherm=True, vecs=False, tol=tol, maxiter=maxiter)
-    else:
-        eigs = eigs_dense(dense.from_csr(op), isherm=True, vecs=False)
     cdef double total = 0
     for i in range(matrix.shape[0]):
         # The abs is technically not part of the definition, but since all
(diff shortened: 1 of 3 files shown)


hodgestar (approved):
Looks good to me.

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,73 @@
+from . import test_mathematics as testing
+import numpy as np
+import scipy as sc
```
jakelishman, line 3: Can we avoid shortening `scipy`?  I don't really see any benefit to it, and it just makes things harder to read.  Shortening is ok for `numpy` because _everyone_ does that, but imo pretty much anything else should be spelled out in full in library code for readability.

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,73 @@
...
+    specialisations = [
+        pytest.param(data.norm.l2_csr, CSR, numbers.Number),
+        pytest.param(data.norm.l2_dense, Dense, numbers.Number),
+    ]
+
+
+class TestTraceNorm(testing.UnaryOpMixin):
+    def op_numpy(self, matrix):
+        matrix = matrix@np.conj(matrix.T)
+        return np.trace(sc.linalg.sqrtm(matrix))
```
jakelishman, line 65: I think this is just `scipy.linalg.norm(matrix, 'nuc')`, but not 100% certain.

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # TraceNorm probaly needs lower tol due to how it is being computed with
+    # NumPy
+    tol = 1e-5
```
jakelishman, line 69: Hmmm, not keen on this.  1e-5 is really _not_ a good tolerance - we should really be passing tests better than this, I think.

AGaliciaMartinez: Neither do I to be honest. It seems to be pointing that something is not good enough? The change in tol is only necessary for a 2 of the tests. These are:
```
FAILED qutip/tests/core/data/test_norm.py::TestTraceNorm::test_mathematically_correct[CSR[ket,filled,sorted]->Number]
FAILED qutip/tests/core/data/test_norm.py::TestTraceNorm::test_mathematically_correct[CSR[ket,filled,unsorted]->Number]
```

Any idea what could be wrong?

qutip/tests/core/data/test_norm.py
AGaliciaMartinez: It is. thanks for the tip!

qutip/tests/core/data/test_norm.py
AGaliciaMartinez: Actually, if a ket is given as input, there should be a fast path to compute the traceNorm. Will try to implement that and see if it improves.

jakelishman: I'm not sure that this should even be defined for ket?  Technically the matrix norm is defined, but it doesn't make sense for a quantum object.  It's like how `l2` and `fro` are actually the same norm, but one's defined only for kets and the other's only for operators.

jakelishman: At any rate, there's almost certainly a shortcut if we desperately wanted to handle it - `vector @ vector.conj().T` is a projector, whose singular values will all be calculable from the L2 norm of `vector` (probably the square of the l2 norm?), so we should be able to shortcut this calculation if we did want to handle it.

AGaliciaMartinez: I think the norm for kets can be computed faster by just computing the l2 norm of the ket? I will try to implement this.

AGaliciaMartinez: I just read the comments and I see we agree in that there is a faster shortcut. I actually think that in these two cases (fro and trace) it would be better for the data layer to handle any mathematically correct case and let qobj decide whether it makes sense to compute the norm or not. That being said, maybe we want for the moment acod the ket and bra cases with `TraceNorm` just for consistency with how `fro` handles things? We could also change `fro` to handle kets though.

AGaliciaMartinez: Also, 
```
FAILED qutip/tests/core/data/test_norm.py::TestTraceNorm::test_mathematically_correct[CSR[square,sparse]->Number]
```
also fails every now and then. Not sure what to do with this one apart from changing tol or repeating the test multiple times.

jakelishman: The Frobenius norm already allows any shape, it's the L2 norm that rejects anything that isn't a vector.  Mathematically, the L2 norm _is_ only defined for vectors, so we should keep it like that.

For the trace norm, we need to keep the `trace_csr` function so that there's an option to pass `sparse=True` for huge systems, but for cases where `sparse=False` (the majority of cases), we can probably just change the algorithm to resolve to `scipy.linalg.norm(x, 'nuc')` - it'll be faster and more accurate.

jakelishman (commented):
Personally I prefer just using `scipy.linalg` fully qualified rather than `from scipy import linalg` (`linalg` is a very common module name, and I don't see any benefit to shortening things in tests), but I don't care enough to insist on changing it again.

My main concern is just the tolerances - we absolutely cannot merge a test which fails 20% of the time, and 1e-5 is far too lax a tolerance for something that should be pretty well conditioned.  We should swap any dense usage to a `scipy.linalg.norm(x, 'nuc')` form - we'll likely have to keep the sparse form as it is, but we're not as worried about tolerances with that.

AGaliciaMartinez:
So now tests do run properly. They do not seem to fail. I did add an extra test for the case for the parameter `sparse=True`. This led me to a bug in `eigs_csr` that is now fixed.

@jakelishman let me know if you are ok with this implementation. I am not very familiar with Cython so I am not sure how to handle the GIL in this case.

I also find quire surprising to need the `sparse` argument in `trace_csr`. Shouldn't this function just perform a sparse computation of the trace norm (sparse always `True`?). Also,  should the specialisation `traces_dense`, that currently does not exist, be mapped to the function `_trace_dense` I just wrote?

AGaliciaMartinez:
Not sure why the test failed.

qutip/core/data/norm.pyx
```diff
@@ -6,6 +6,7 @@ from libc cimport math
 from cpython cimport mem
 
 from scipy.linalg cimport cython_blas as blas
+from scipy.linalg import norm
```
AGaliciaMartinez, line 9: change this

qutip/core/data/norm.pyx
```diff
@@ -61,6 +61,23 @@ cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
         total += math.sqrt(abs(eigs[i]))
     return total
 
+cpdef double _trace_dense(Dense matrix) except -1:
+    """Compute the trace norm reliying scipy for dense operations."""
+    return norm(matrix.as_ndarray(), 'nuc')
```
AGaliciaMartinez, line 66: make this as specialisation

qutip/core/data/norm.pyx
```diff
@@ -212,6 +229,7 @@ trace.__doc__ =\
     """
 trace.add_specialisations([
     (CSR, trace_csr),
+    (Dense, trace_dense),
```
Ericgig, line 232: `trace_dense` signature does not have the `sparse, tol, maxiter` parameters of the trace dispatcher so:
`qt.data.norm.trace(qt.qeye(5).to('dense').data)` fails: 
```TypeError: qutip.core.data.norm.trace_dense() takes exactly one argument (4 given)```

The dispatcher signature must only contain parameters that all specializations accept.

Comment (AGaliciaMartinez) [reply qutip/core/data/norm.pyx:232]: I did not know that `trace_norm` had to take those arguments necessarily. Sorry for that. What do you thing about adding a check in the `add_specialisaiton` method to check whether the added specialisations accept the expected arguments (?). 

I any case, these arguments dont seem to make sense for dense_norm. 
I am not sure what to do here. I can revert the changes and not include the `trace_dense` in this PR. However, it does seem that a trace_dense is useful as you would like to avoid the conversion to CSR when doing the `norm.trace`. Mainly because, by default, you convert back to dense...

Comment (Ericgig) [reply qutip/core/data/norm.pyx:232]: I suggest we remove the `sparse` parameter and `tol` and `maxiter` from the dispatcher.
`_trace_csr` would become the specialization for sparse. For a user to change the `tol`, he would need to call the specialization directly, but since it does not make sense for other data type, it should be fine.

Comment (AGaliciaMartinez) [reply qutip/core/data/norm.pyx:232]: I was worried that this would break backwards compatibility but I also agree that removing those parameters is probably necessary. Should I proceed then?

Comment (Ericgig) [reply qutip/core/data/norm.pyx:232]: Backwards compatibility could be fixed in `Qobj.norm`. But I don't believe that these options are that commonly used.

**Comment by AGaliciaMartinez** [comment]
I removed the sparse specific kwargs in the trace dispatcher so that trace_dense could be included.  @hodgestar, since there have been significant changes to the code since you last reviewed, could you take another look?

**Review by Ericgig — APPROVED** [review]
The trace dispatcher is fine like this, but we should remember to review `Qobj.norm` to match it.

File: qutip/core/data/norm.pyx [hunk 7]
```diff
@@ -43,16 +44,17 @@ cpdef double one_csr(CSR matrix) except -1:
     finally:
         mem.PyMem_Free(col)
 
-cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
-    # We use the general eigenvalue solver which involves a Python call, so
-    # there's no point attempting to release the GIL.
+cpdef double _trace_csr(CSR matrix, tol=0, maxiter=None) except -1:
+    """Compute the trace norm relaying only on sparse operations. This consists
+    on a digitalization of `matrix@matrix.adjoint()` to obtain the result from
+    the diagonals."""
```
hodgestar, line 50:
```suggestion
    """Compute the trace norm using only sparse operations. These consist
    of determining the eigenvalues of `matrix @ matrix.adjoint()` and summing their square roots."""
```

qutip/core/data/norm.pyx
```diff
@@ -61,6 +63,21 @@ cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
...
+    """Compute the trace norm relaying scipy for dense operations."""
+    return scipy.linalg.norm(matrix.as_ndarray(), 'nuc')
+
+cpdef double trace_csr(CSR matrix, sparse=False, tol=0, maxiter=None) except -1:
+    # For column and row vectors we simply use the l2 norm as it is equivalent
+    # to the trace norm.
+    if matrix.shape[0]==1 or matrix.shape[1]==1:
+        return l2_csr(matrix)
+
+    if sparse:
```
hodgestar, line 76: There is no way to specify `sparse=True` via the data layer, so I propose we just remove it entirely and combine `_trace_csr` with this method.

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,86 @@
+from . import test_mathematics as testing
+import numpy as np
+from scipy import linalg
```
hodgestar, line 3:
```suggestion
import scipy
```
Suggesting making this change now so we don't have to tidy it up later.

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,86 @@
...
+import pytest
+from qutip import data
+from qutip.core.data import CSR, Dense
+import numbers
+
+
+class TestOneNorm(testing.UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return linalg.norm(matrix, 1)
```
hodgestar, line 12:
```suggestion
        return scipy.linalg.norm(matrix, 1)
```

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,86 @@
...
+
+    specialisations = [
+        pytest.param(data.norm.one_csr, CSR, numbers.Number),
+        pytest.param(data.norm.one_dense, Dense, numbers.Number),
+    ]
+
+
+class TestFrobeniusNorm(testing.UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return linalg.norm(matrix, 'fro')
```
hodgestar, line 22:
```suggestion
        return scipy.linalg.norm(matrix, 'fro')
```

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,86 @@
...
+
+    specialisations = [
+        pytest.param(data.norm.max_csr, CSR, numbers.Number),
+        pytest.param(data.norm.max_dense, Dense, numbers.Number),
+    ]
+
+
+class TestL2Norm(testing.UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return linalg.norm(matrix, 'fro')
```
hodgestar, line 44:
```suggestion
        return scipy.linalg.norm(matrix, 'fro')
```

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,86 @@
...
+    ]
+    specialisations = [
+        pytest.param(data.norm.l2_csr, CSR, numbers.Number),
+        pytest.param(data.norm.l2_dense, Dense, numbers.Number),
+    ]
+
+
+class TestTraceNorm(testing.UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return linalg.norm(matrix, 'nuc')
```
hodgestar, line 64:
```suggestion
        return scipy.linalg.norm(matrix, 'nuc')
```

qutip/tests/core/data/test_norm.py
```diff
@@ -0,0 +1,86 @@
...
+    specialisations = [
+        pytest.param(data.norm.trace_csr, CSR, numbers.Number),
+        pytest.param(data.norm.trace_dense, Dense, numbers.Number),
+    ]
+
+# This test tests the special case where the argument sparse in traces_csr is
+# set to True.
+class TestTraceNormSparseTrue(testing.UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return linalg.norm(matrix, 'nuc')
```
hodgestar, line 75:
```suggestion
        return scipy.linalg.norm(matrix, 'nuc')
```

hodgestar (changes requested):
I left some clean-up suggestions (mostly to follow Jake's suggestion to refer to scipy.linalg by its full name), a docstring suggestion and then a bigger picture suggestion to remove the mostly unused `sparse=...` argument from the `trace_csr` function.

We probably need to get used to making these kinds of explicit "I want a better result via dense matrices" via the data layer itself -- i.e. `qobj.to("dense").norm()` -- and get out of the old qutip habits of `qobj.norm(sparse=False)`.

qutip/tests/core/data/test_norm.py
jakelishman: scipy's components are separate packages which aren't auto-imported, so it needs to be `import scipy.linalg`

hodgestar: This works accidentally because `import qutip` does `import scipy.linalg` somewhere.

qutip/core/data/norm.pyx
AGaliciaMartinez:
```suggestion
    """Compute the trace norm using only sparse operations. These consist
    of determining the eigenvalues of `matrix @ matrix.adjoint()` and summing their square roots.
    """
```

hodgestar:
The intermittent test failure in c6d68c0 is unrelated and should be resolved by merging in dev.major.

AGaliciaMartinez:
I added all the suggested changes. However, now the tets for trace norm with csr matrices started failing (in particular `TestTraceNorm.test_mathematically_correct[CSR[square,sparse]->Number]`). Decreasing the tolerance to `1e-8` solves the problem but given that this was not a problem before (at least I do not remember this happening), I wonder if it is actually signaling that there is something wrong.

Ericgig:
I don't think there is something wrong.
We can expect the matmul output and eigen values to have errors of about 1e-16, (double's precision), then it sums the square-roots, resulting in error up to ~1e-8 when the eigen values should be zeros. So with a tol at 1e-8, the test can fail from numerical error with nothing wrong in the test.

If you look are the sparse eigen solver, for a 100x100 matrix with only 10 non-zero values, all along the diagonal, it does not find any` 0` eigen values: there are about 89x `~e-36`, 1x `~e-17` and the 10 non-zeros values. So the `sqrt(~e-17)` is pure numerical error and appear in the test at around 3e-9. 
I ran the test a few thousand times and it fails about 23% of the time, at `tol=3e-8`, it fails about 0.2% of the time and the max is 3.7e-8.

So I expect everything is fine and you just need to loosen the tolerance.

AGaliciaMartinez:
Makes sense, thank your for the explanation Eric! While trying to set a sensible value for `tol` I got the problem that 3% of the times the tests fail independently of the `tol` value. This is because the numpy function returns `nan` whereas `trace_csr` does not. To replicate the problem I used the following code:
```python
import qutip
import scipy
import numpy as np
from qutip.tests.core.data import conftest

# Catch one of those matrices that return nan in the nuclear norm
for i in range(100):
    matrix_csr = conftest.random_csr((100,100), 0.001, True)
    matrix = matrix_csr.to_array() 
    result = np.linalg.norm(matrix, 'nuc')

    # If it is nan  store the matrix for analysis.
    if np.isnan(result):
        print('got one of those tricky matrices')
        nan_matrix = matrix        
        nan_matrix_csr = matrix_csr
    
# Scipy's and numpys methods return nan
sc_norm = scipy.linalg.norm(nan_matrix, 'nuc')
print(f'SciPy linalgn nuc norm: {sc_norm}')

np_norm = np.linalg.norm(nan_matrix, 'nuc')
print(f'NumPy linalgn nuc norm: {np_norm}')

# Using svd wich is similar to what numpy/scipy should be doing
u, s, vh = np.linalg.svd(nan_matrix)
print(f'SVD method: {s.sum()}')

# Qutip sparse method
print(f'QuTiP trace_csr:{qutip.data.norm.trace_csr(nan_matrix_csr)}')

# This prints
# SciPy linalgn nuc norm: nan
# NumPy linalgn nuc norm: nan
# SVD method: 6.456744360029763
# QuTiP trace_csr:6.456744366739419
```

As you can see in the above code, it fails to compute a value for NumPy/SciPy but not for a direct method (SVD) or the QuTiP method (which is different to the SVD method but still works).

Eric, do you get the same results? They look quite weird and I wonder if I am doing something wrong or whether this is a NumPy/SciPy bug.
------------
```
np.__version__='1.21.2'
qutip.__version__='5.0.0.dev0+7a90dbd'
scipy.__version__='1.7.0'
```

Ericgig:
I see the `Nan`'s...
Scipy's `norm` call numpy's `norm` which call `svd` with `compute_uv=False` and it return junk...:

```
np.linalg.svd(nan_matrix, compute_uv=False)
array([            inf,             inf, 1.07311390e+000, 9.83440402e-001,
       9.42395122e-001, 9.27989857e-001, 8.23177182e-001, 2.38050872e-022,
       6.40149943e-032, 5.64518572e-033, 9.14729223e-045, 2.72704923e-049,
       1.56401558e-063, 4.83650926e-068, 6.39334574e-081, 2.83836536e-092,
       8.27175428e-096, 1.73661126e-097, 2.01091260e-114, 1.30856687e-123,
       2.06455678e-127, 4.92744107e-155, 5.08952541e-159, 1.74118770e-186,
       9.44136765e-227, 6.23313780e-246, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000, 0.00000000e+000,
       0.00000000e+000, 0.00000000e+000, 0.00000000e+000,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan,             nan,
                   nan,             nan,             nan, 0.00000000e+000])
```

Put with `compute_uv=True` (default) it work fine...

I guess the best we can do is report to numpy... Do you want to do it?

Not computing the `uv` is 10x faster in case everything goes well and these error seem rare, so for now I would use a fall back:
```
norm = np.linalg.norm(matrix, 'nuc')
if np.isnan(norm):
    # Bug in `numpy svd(..., compute_uv=False)` used by `norm(..., 'nuc')`
    norm = np.linalg.svd(nan_matrix)[1].sum()
return norm
```

AGaliciaMartinez:
It seems that they already know about this bug numpy/numpy#19473. 

The tests runs smoothly now so it should be ready to merge. @hodgestar, is there any change missing?

hodgestar (approved):

PR #1636, opened 2021-08-12
Expect return real value only when possible.

AGaliciaMartinez:
**Description**
Changes expect to cast to real only when attribute real exist. This was a problem for qutip-tensorflow as it returned a `tf.Tensor` that does not have the real attribute. 

I did not add any test yet. I thought of adding a test that creates a dummy_specialisation that always returns a dummy class. However, this is was quite involved test. I can include it although I wondered if it was the correct approach to test the code.

Also, note that I am not using an `isinstance(out, number.Number)` deliberately. There are examples of class that do implement the `real` and `imag` attribute, such us numpy arrays. Furthermore, I opened an [issue](https://github.com/tensorflow/tensorflow/issues/51463) in TensorFlow to see if they would like to support the `real` and `imag` attributes. 

**Changelog**
`expect` can now return arbitrary python objects.

### DIFF
--- a/qutip/core/expect.py
+++ b/qutip/core/expect.py
@@ -33,6 +33,8 @@
 
 __all__ = ['expect', 'variance']
 
+import numbers
+
 import numpy as np
 
 from .qobj import Qobj
@@ -77,6 +79,7 @@ def expect(oper, state):
             return np.array([_single_qobj_expect(op, state) for op in oper],
                             dtype=dtype)
         return [expect(op, state) for op in oper]
+
     elif isinstance(state, (list, np.ndarray)):
         dtype = np.complex128
         if oper.isherm and all(op.isherm or op.isket for op in state):
@@ -99,7 +102,12 @@ def _single_qobj_expect(oper, state):
         )
         raise ValueError(msg)
     out = _data.expect(oper.data, state.data)
-    return out.real if oper.isherm and (state.isket or state.isherm) else out
+
+    # This ensures that expect can return something that is not a number.
+    try:
+        return out.real if oper.isherm and (state.isket or state.isherm) else out
+    except AttributeError:
+        return out
 
 
 def variance(oper, state):
--- a/qutip/core/expect.py
+++ b/qutip/core/expect.py
@@ -33,8 +33,6 @@
 
 __all__ = ['expect', 'variance']
 
-import numbers
-
 import numpy as np
 
 from .qobj import Qobj


Ericgig (approved):

qutip/core/expect.py
```diff
@@ -99,7 +100,12 @@ def _single_qobj_expect(oper, state):
         )
         raise ValueError(msg)
     out = _data.expect(oper.data, state.data)
-    return out.real if oper.isherm and (state.isket or state.isherm) else out
+
+    # This ensures that expect can return something that is not a number.
+    try:
+        return out.real if oper.isherm and (state.isket or state.isherm) else out
```
hodgestar, line 106:
```suggestion
        return out.real if oper.isherm and (state.isket or state.isherm) and hasattr(out, "real") else out
```
I would not use a `try: ... except AttributeError:` around this expression because it accesses many attributes and already contains an `if ... else` that explicitly tests for whether `out.real` is the correct thing to return.

hodgestar (changes requested):
I left a suggestion to use `hasattr` rather than catch `AttributeError`. We should also explain in the comment *why* we need this check (i.e. other data layers such as `qutip-tensorflow` might `qutip-cupy` might need to return a value that is tf.Tensor or a CuPy array that does not have a `.real` attribute).

I would also like this to have a test, because the behaviour is seldom triggered anywhere in QuTiP itself so it could easily gather bugs that lurk silently.

One straightforward way is to have a test that uses pytest's monkeypatch fixture to overwrite `_data.expect` during that one test. The monkeypatch fixture is never a good first option, but it exists for exactly this sort of situation where `.expect` *can* return a value without `.real` but no code in QuTiP itself actually does this.

Another straightforward way is to temporarily add just a dummy specialisation to the expect Dispatcher with `.add_specialisations` (e.g. for Dense, Dense) and then override it again afterwards by calling `.add_specialisations` again with the original specialisation.

What I don't like about the second option is that if we get restoring the specialisation wrong, it could be very confusing, but on the upside it is closer to what will happen in real code.

hodgestar (approved):
I would still ideally have liked a test for this, but the change looks okay to me now.

AGaliciaMartinez:
I will add tests. I changed the status to draft to avoid merging.

AGaliciaMartinez:
@hodgestar The test is now included. It turned out to be quite simple with monkeypatch. Thanks for the suggestion!

PR #1637, opened 2021-08-13
Test reshape specialisations.

AGaliciaMartinez:
**Changelog**
Added tests for reshape specialisations.
Added tests for columns_stack specialisations.
Added tests for columns_unstack specialisations.
Moved tests for split_columns from test_mathematics to test_reshape.
PEP 8 clean test_mathematics.

### DIFF
--- a/qutip/tests/core/data/test_reshape.py
+++ b/qutip/tests/core/data/test_reshape.py
@@ -3,7 +3,6 @@ import pytest
 import numpy as np
 from qutip import data
 from qutip.core.data import CSR, Dense
-from itertools import product
 
 
 class TestSplitColumns(UnaryOpMixin):
--- a/qutip/tests/core/data/test_reshape.py
+++ b/qutip/tests/core/data/test_reshape.py
@@ -1,0 +1,136 @@
+from .test_mathematics import UnaryOpMixin
+import pytest
+import numpy as np
+from qutip import data
+from qutip.core.data import CSR, Dense
+from itertools import product
+
+
+
+class TestSplitColumns(UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return [matrix[:, i].reshape((-1, 1)) for i in range(matrix.shape[1])]
+
+    specialisations = [
+        pytest.param(data.split_columns_csr, CSR, list),
+        pytest.param(data.split_columns_dense, Dense, list),
+    ]
+
+
+class TestColumnStack(UnaryOpMixin):
+    # It seem column_stack assumes F ordered array?
+    def op_numpy(self, matrix):
+        out_shape = (matrix.shape[0]*matrix.shape[1], 1)
+        return np.reshape(matrix, newshape=out_shape, order='F')
+
+    specialisations = [
+        pytest.param(data.column_stack_csr, CSR, CSR),
+        pytest.param(data.column_stack_dense, Dense, Dense),
+    ]
+
+
+class TestColumnUnstack(UnaryOpMixin):
+    # It seem column_unstack assumes F ordered array?
+    def op_numpy(self, matrix, rows):
+        out_shape = (rows, matrix.shape[0]*matrix.shape[1]//rows)
+        return np.reshape(matrix, newshape=out_shape, order='F')
+
+    shapes = [
+        (pytest.param((10, 1), id="ket"), ),
+    ]
+
+    bad_shapes = [
+        (pytest.param((1, 10), id="bra"), ),
+        (pytest.param((10, 10), id="square"), ),
+        (pytest.param((2, 10), id="non_square"), ),
+    ]
+
+    specialisations = [
+        pytest.param(data.column_unstack_csr, CSR, CSR),
+        pytest.param(data.column_unstack_dense, Dense, Dense),
+    ]
+
+    # `ColumnUnstack` has an additional scalar parameters. This is why tests
+    # are a little bit more involved.
+    @pytest.mark.parametrize('rows', [2, 5])
+    def test_mathematically_correct(self, op, data_m, rows, out_type):
+        """
+        Test that the binary operation is correct for all the known type
+        specialisations.
+        """
+        matrix = data_m()
+        expected = self.op_numpy(matrix.to_array(), rows)
+        test = op(matrix, rows)
+        assert isinstance(test, out_type)
+        assert test.shape == expected.shape
+        np.testing.assert_allclose(test.to_array(), expected, atol=self.tol)
+
+    # These tests will check that ValueError is raised when input matrices have
+    # invalid shape.
+    def test_incorrect_shape_raises(self, op, data_m):
+        with pytest.raises(ValueError):
+            op(data_m(), 1)  # We set rows to one so that it is always valid
+
+
+    # We also test that wrong value for `rows` raises ValueError We will
+    # generate tests with valid data_m since we want to make sure that the
+    # error is raised due to wrong `rows` value. For this we will make use of
+    # the generate_mathematically_correct function. This is why out_type is
+    # included although not used at all.
+    def test_incorrect_rows_raises(self, op, data_m, out_type):
+        with pytest.raises(ValueError):
+            op(data_m(), 3)
+
+    # To generate tests we simply call to getenerate_mathematically correct.
+    def generate_incorrect_rows_raises(self, metafunc):
+        self.generate_mathematically_correct(metafunc)
+
+
+
+
+class TestReshape(UnaryOpMixin):
+    def op_numpy(self, matrix, rows, columns):
+        # It seem to_array returns F ordered numpy arrays.
+        out_shape = (rows, columns)
+        return np.reshape(matrix, newshape=out_shape, order='C')
+
+    # I use custom shapes so that they all have 100 elements.
+    shapes = [
+        (pytest.param((1, 100), id="bra"), ),
+        (pytest.param((100, 1), id="ket"), ),
+        (pytest.param((2, 50), id="non_square"), ),
+        (pytest.param((10, 10), id="square"), ),
+    ]
+
+    specialisations = [
+        pytest.param(data.reshape_dense, Dense, Dense),
+        pytest.param(data.reshape_csr, CSR, CSR),
+    ]
+
+    # `reshape` has an additional scalar parameters. This is why tests
+    # are a little bit more involved.
+    @pytest.mark.parametrize('rows, columns', [(5, 20), (10, 10)])
+    def test_mathematically_correct(self, op, data_m, rows, columns, out_type):
+        """
+        Test that the binary operation is correct for all the known type
+        specialisations.
+        """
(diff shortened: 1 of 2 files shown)


AGaliciaMartinez:
I had some problems figuring out the way the reshapes are done in each of the specialisations. It seems I need to use numpys 'F' ordering when doing `column_stack` and `columns_unstack`. However, I need to use the numpys 'C' ordering to match `reshape` specialisations output. Could you confirm me that this is the correct behaviour @jakelishman? I found it a little bit weird.

AGaliciaMartinez:
Test fails do not seem to be related to changes. Could they be rerun?

jakelishman:
The `order` argument of `numpy.reshape` says which order it should read and write the elements in, with regards to the shape of the input.  So `order='f'` for 2D arrays means it reads all of the first column, then all of the second column, and so on, while `order='c'` means it reads all of the first row, then all the second row, and so on.

In other words, when you want to reshape keeping _columns_ intact, you'd use `order='f'`, and when you want to reshape keeping _rows_ intact (and, usually, memory), you'd use `order='c'`.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
+from .test_mathematics import UnaryOpMixin
+import pytest
+import numpy as np
+from qutip import data
+from qutip.core.data import CSR, Dense
+
+
+class TestSplitColumns(UnaryOpMixin):
+    def op_numpy(self, matrix):
+        return [matrix[:, i].reshape((-1, 1)) for i in range(matrix.shape[1])]
```
jakelishman, line 10: Can also just be `return list(matrix.T)`.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+        (pytest.param((2, 10), id="non_square"), ),
+    ]
+
+    specialisations = [
+        pytest.param(data.column_unstack_csr, CSR, CSR),
+        pytest.param(data.column_unstack_dense, Dense, Dense),
+    ]
+
+    # `ColumnUnstack` has an additional scalar parameters. This is why tests
+    # are a little bit more involved.
```
jakelishman, line 52: This comment made sense when it was just one comment on one test (`add`, maybe?) in the original file.  Now that it's pretty common, this is just noise and should be removed.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+        expected = self.op_numpy(matrix.to_array(), rows)
+        test = op(matrix, rows)
+        assert isinstance(test, out_type)
+        assert test.shape == expected.shape
+        np.testing.assert_allclose(test.to_array(), expected, atol=self.tol)
+
+    # These tests will check that ValueError is raised when input matrices have
+    # invalid shape.
+    def test_incorrect_shape_raises(self, op, data_m):
+        with pytest.raises(ValueError):
```
jakelishman, line 69: Unnecessary comment - both the code and its intent are very clear already, so this just adds noise.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+    # included although not used at all.
+    @pytest.mark.parametrize('rows', [-1, 0, 3], ids=['negative', 'zero',
+                                                      'invalid'])
+    def test_incorrect_rows_raises(self, op, data_m, out_type, rows):
+        with pytest.raises(ValueError):
+            op(data_m(), rows)
+
+    # To generate tests we simply call to getenerate_mathematically correct.
+    def generate_incorrect_rows_raises(self, metafunc):
+        self.generate_mathematically_correct(metafunc)
```
jakelishman, line 85: Don't explain what the code does if it's obvious.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+    # invalid shape.
+    def test_incorrect_shape_raises(self, op, data_m):
+        with pytest.raises(ValueError):
+            op(data_m(), 1)  # We set rows to one so that it is always valid
+
+    # We also test that wrong value for `rows` raises ValueError We will
+    # generate tests with valid data_m since we want to make sure that the
+    # error is raised due to wrong `rows` value. For this we will make use of
+    # the generate_mathematically_correct function. This is why out_type is
+    # included although not used at all.
```
jakelishman, line 76: Most of this is unnecessary noise - the test name describes most of this, and the only necessary bit is just a comment saying

> `out_type` is included but not used so that `generate_mathematically_correct` can be re-used.

Comments are much more useful if they're short and say only the minimum that's not already clear from the code.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+            op(data_m(), rows)
+
+    # To generate tests we simply call to getenerate_mathematically correct.
+    def generate_incorrect_rows_raises(self, metafunc):
+        self.generate_mathematically_correct(metafunc)
+
+
+class TestReshape(UnaryOpMixin):
+    def op_numpy(self, matrix, rows, columns):
+        # It seem to_array returns F ordered numpy arrays.
```
jakelishman, line 90: Comments like this don't explain anything, so shouldn't be here.  In these cases, if you're unsure, it would really be better to check - as it stands, `to_array` is _not_ guaranteed to return an F-ordered array in general, so this comment is wrong.

https://github.com/qutip/qutip/blob/cefacfc90a41c8627f1c7d4ae9fe1aad73a7f306/qutip/core/data/dense.pyx#L159-L173

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+        out_shape = (rows, columns)
+        return np.reshape(matrix, newshape=out_shape, order='C')
+
+    # I use custom shapes so that they all have 100 elements.
```
jakelishman, line 94: Explain _why_ not _what_ in comments.  In this case, presumably (?), it's

> All matrices should have the same number of elements in total, so we can use the same `(rows, columns)` parametrisation for each input.

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+        (pytest.param((10, 10), id="square"), ),
+    ]
+
+    specialisations = [
+        pytest.param(data.reshape_dense, Dense, Dense),
+        pytest.param(data.reshape_csr, CSR, CSR),
+    ]
+
+    # `reshape` has an additional scalar parameters. This is why tests
+    # are a little bit more involved.
```
jakelishman, line 108: delete

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+        test = op(matrix, rows, columns)
+        assert isinstance(test, out_type)
+        assert test.shape == expected.shape
+        np.testing.assert_allclose(test.to_array(), expected, atol=self.tol)
+
+    # We also test that wrong value for `rows` or `columns` raises ValueError
+    # We will generate tests with valid data_m since we want to make sure that
+    # the error is raised due to wrong `rows` or columns value. For this we
+    # will make use of the generate_mathematically_correct function. This is
+    # why out_type is included although not used at all.
```
jakelishman, line 126: same comment as before

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+    @pytest.mark.parametrize('rows, columns', [(-2, -50), (-50, -2), (3, 10)],
+                            ids=["negative1","negative2","invalid"])
+    def test_incorrect_rows_raises(self, op, data_m, out_type, rows, columns):
+        with pytest.raises(ValueError):
+            op(data_m(), rows, columns)
+
+    # To generate tests we simply call to getenerate_mathematically correct.
```
jakelishman, line 133: delete

qutip/tests/core/data/test_reshape.py
```diff
@@ -0,0 +1,135 @@
...
+
+    specialisations = [
+        pytest.param(data.reshape_dense, Dense, Dense),
+        pytest.param(data.reshape_csr, CSR, CSR),
+    ]
+
+    # `reshape` has an additional scalar parameters. This is why tests
+    # are a little bit more involved.
+    @pytest.mark.parametrize('rows, columns', [(5, 20), (10, 10)])
+    def test_mathematically_correct(self, op, data_m, rows, columns, out_type):
```
jakelishman, line 110: Why parametrise over two parameters when you could just parametrise over `new_shape`?

jakelishman (commented):
Most of the actual code looks good to me, but I have some comments on tidying up unnecessary comments, so the code's easier to read - unnecessary comments add noise, and make things seem more complex than they are.

AGaliciaMartinez:
> The `order` argument of `numpy.reshape` says which order it should read and write the elements in, with regards to the shape of the input. So `order='f'` for 2D arrays means it reads all of the first column, then all of the second column, and so on, while `order='c'` means it reads all of the first row, then all the second row, and so on.
> 
> In other words, when you want to reshape keeping _columns_ intact, you'd use `order='f'`, and when you want to reshape keeping _rows_ intact (and, usually, memory), you'd use `order='c'`.

Ok, I see. Thanks for the explanation!

qutip/tests/core/data/test_reshape.py
AGaliciaMartinez: This seems to change the shape of each individual element in the list to be (2,) instead of (2,1).

jakelishman: oh yeah, of course.  I guess it would have to be `[column[:, np.newaxis] for column in matrix.T]`, but then I'm not sure it's much better than what you've got already.

qutip/tests/core/data/test_reshape.py
AGaliciaMartinez: So that the parameters match the required parameters for the reshape specialisation.

qutip/tests/core/data/test_reshape.py
AGaliciaMartinez: I actually think is more readable.

qutip/tests/core/data/test_reshape.py
jakelishman: Yeah, fair enough. It looks a little odd to me to parametrise over them rather than a single linked variable "shape", but that's ok.

qutip/tests/core/data/test_reshape.py
jakelishman: Up to you.  Yours is certainly fine to read as well - my original suggestion was only because I thought it was basically a trivial iterator, forgetting the extra shape dimension.

Ericgig (approved):

hodgestar (approved):
Thanks @AGaliciaMartinez!

PR #1643, opened 2021-08-26
Swappable ODE integrators.

Ericgig:
Introduce a common ODE integrator interface for qutip's solver.

Presently qutip's solver use `scipy.integratre.ode`'s zvode ODE solver, which support `adams` and `bdf` methods.
These are great method in most case, but not always optimal.
This PR introduce, a common interface to allow other ODE method to be used by solvers.
It also add 2 new scipy ODE solvers: `lsoda` and `dop853`.
`lsoda` detect if the system is stiff and switch between `adams` and `bdf`.
`dop853` is a 8th order Runge-Kutta method.

The `Integrator` use `Data` as state and return the state as a pair `(t, state)`.
Often solver's run only compute the expectation values, so there is no need to create a `Qobj`.
Returning the time with the state is for `mcsolve` which have the integration advancing with one internal step and the run loop is cleaner in solver. 

I did some benchmark and `adams` method is often the best scipy method.
Sparse system: loglog plot of time in function of system size.
image

Dense time-dependent system.
image

*`vern7`, `vern9`, and `diag` method will be introduced in another PR.

### DIFF
--- a/qutip/solver/integrator.py
+++ b/qutip/solver/integrator.py
@@ -17,7 +17,7 @@ class IntegratorException(Exception):
 class Integrator:
     """
     A wrapper around ODE solvers.
-    It ensure a common interface for Solver usage.
+    It ensures a common interface for Solver usage.
     It takes and return states as :class:`qutip.core.data.Data`, it may return
     a different data-type than the input type.
 
--- a/qutip/solver/integrator.py
+++ b/qutip/solver/integrator.py
@@ -40,6 +40,7 @@ class Integrator:
         self.system = system
         self._stats = {}
         self.options = options
+        self._is_set = False
         self._prepare()
 
     def _prepare(self):
@@ -124,7 +125,8 @@ class Integrator:
 
     def reset(self):
         """Reset internal state of the ODE solver."""
-        self.set_state(*self.get_state)
+        if self._is_set:
+            self.set_state(*self.get_state())
 
     def update_args(self, args):
         """
--- a/qutip/solver/integrator.py
+++ b/qutip/solver/integrator.py
@@ -33,7 +33,7 @@ class Integrator:
         Options for the solver.
 
     Class Attributes
-    ---------------------
+    ----------------
     supports_blackbox : bool
         If True, then the integrator calls only ``system.matmul``,
         ``system.matmul_data``, ``system.expect``, ``system.expect_data`` and
@@ -47,13 +47,19 @@ class Integrator:
     supports_time_dependent : bool
         If True, then the integrator supports time dependent systems. If False,
         ``supports_blackbox`` should usually be ``False`` too.
+
+    integrator_options : dict
+        A dictionary of options used by the integrator and their default
+        values. Once initiated, ``self.options`` will be a dict with the same
+        keys, not the full options object passed to the solver. Options' keys
+        included here will be supported by the :cls:SolverOdeOptions.
     """
     # Used options in qutip.SolverOdeOptions
     integrator_options = {}
     # Can evolve time dependent system
     support_time_dependant = None
     # Use the QobjEvo's matmul_data method as the driving function
-    use_QobjEvo_matmul = None
+    supports_blackbox = None
     name = ""
 
     def __init__(self, system, options):
--- a/qutip/solver/integrator.py
+++ b/qutip/solver/integrator.py
@@ -73,13 +73,38 @@ class Integrator:
         """
         raise NotImplementedError
 
-    def integrate(self, t, step=False, copy=True):
+    def integrate(self, t, copy=True):
         """
         Evolve to t.
 
-        If `step` advance up to t by one internal solver step.
-        Should be able to retreive the state at any time between the present
-        time and the resulting time with subsequent calls.
+        Before calling `integrate` for the first time, the initial state should
+        be set with `set_state`.
+
+        Parameters
+        ----------
+        t : float
+            Time to integrate to, should be larger than the previous time.
+
+        copy : bool [True]
+            Whether to return a copy of the state or the state itself.
+
+        Return
+        ------
+        (t, state) : (float, qutip.Data)
+            The state of the solver at ``t``.
+        """
+        raise NotImplementedError
+
+    def mcstep(self, t, copy=True):
+        """
+        Evolve toward the time ``t``.
+
+        If ``t`` is larger than the present state's ``t``, advance the internal
+        state toward ``t``. If  ``t`` is smaller than the present ``t``, but
+        larger than the precious one, it does an interpolation step and return
+        the state at that time. When advancing the state, it may return it at a
+        time between present time and the asked ``t`` if more efficent for
+        subsequent interpolation step.
 
         Before calling `integrate` for the first time, the initial step should
         be set with `set_state`.
@@ -91,17 +116,18 @@ class Integrator:
             the last integrate call was use with ``step=True``, the time can be
             between the time at the start of the last call and now.
 
-        step : bool [False]
-            When True, integrate for a one internal step of the ODE solver.
-
         copy : bool [True]
             Whether to return a copy of the state or the state itself.
 
         Return
         ------
         (t, state) : (float, qutip.Data)
-            The state of the solver at ``t``. The returned time ``t`` can differ from
-            the input time only when ``step=True``.
+            The state of the solver at ``t``. The returned time ``t`` can
+            differ from the input time only when ``step=True``.
+
+        .. note:
+            Does not need to be implemented it the Integrator does not support
+            :func:`mcsovle`
         """
         raise NotImplementedError
 
--- a/qutip/solver/integrator.py
+++ b/qutip/solver/integrator.py
@@ -9,13 +9,12 @@ class IntegratorException(Exception):
     """Error from the ODE solver being unable to integrate with the given
     parameters.
 
-    Exemple
+    Example
     -------
     - The solver cannot reach the desired tolerance within the maximum number
     of steps.
     - The step needed to be within desired tolerance is too small.
     """
-    pass
 
 
 class Integrator:
@@ -27,11 +26,25 @@ class Integrator:
 
     Parameters
     ----------
-    sytem: qutip.QobjEvo
+    system: qutip.QobjEvo
         Quantum system in which states evolve.
 
     options: qutip.SolverOdeOptions
         Options for the solver.
+
+    Class Attributes
+    ---------------------
+    supports_blackbox : bool
+          If True, then the integrator calls only ``system.matmul``, ``system.matmul_data``, ``system.expect``, 
+          ``system.expect_data`` and ``isconstant``, ``isoper`` or ``issuper``. This allows the solver using the integrator
+          to modify the system in creative ways. In particular, the solver may modify the system depending on *both*
+          the time ``t`` *and* the current ``state`` the system is being applied to.
+        
+          If the integrator calls any other methods, set to False.
+          
+    supports_time_dependent : bool
+        If True, then the integrator supports time dependent systems. If False, ``supports_blackbox`` should usually be
+        ``False`` too.
     """
     # Used options in qutip.SolverOdeOptions
     used_options = []
@@ -101,7 +114,7 @@ class Integrator:
 
         If ``t`` is larger than the present state's ``t``, advance the internal
         state toward ``t``. If  ``t`` is smaller than the present ``t``, but
-        larger than the precious one, it does an interpolation step and return
+        larger than the previous one, it does an interpolation step and returns
         the state at that time. When advancing the state, it may return it at a
         time between present time and the asked ``t`` if more efficent for
         subsequent interpolation step.
@@ -127,7 +140,7 @@ class Integrator:
 
         .. note:
             Does not need to be implemented it the Integrator does not support
-            :func:`mcsovle`
+            :func:`mcsolve`
         """
         raise NotImplementedError
 
--- a/qutip/solver/integrator.py
+++ b/qutip/solver/integrator.py
@@ -1,5 +1,5 @@
 """ `Integrator`: ODE solver wrapper to use in qutip's Solver """
-
+import numpy as np
 
 __all__ = ['Integrator', 'IntegratorException', 'add_integrator',
            'sesolve_integrators', 'mesolve_integrators', 'mcsolve_integrators']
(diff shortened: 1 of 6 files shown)


qutip/solver/integrator.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class IntegratorException(Exception):
+    pass
+
+
+class Integrator:
+    """
+    A wrapper around ODE solvers.
+    It ensure a common interface for Solver usage.
```
AGaliciaMartinez, line 20:
```suggestion
    It ensures a common interface for Solver usage.
```

hodgestar:
@Ericgig Do you have examples of where integrators are used?

Some of the interface choices seem inconsistent -- e.g. `system` is a `QobjEvo` instance (so it has dims) but the states are `data` (which don't have dims, type, etc). Perhaps this makes sense given how things are used, but it would be good to be consistent (especially given that accessing `.data` is very fast and would allow the integrator access to the dims and various high-level operations).

Perhaps one solution is to have two levels of interface -- i.e. a low-level one (`get_state`, `set_state`, `integrate`) and a high-level one (`run`) where the signature of `run` changes to `run(initial_qobj, tlist) -> final_qobj` and `run` calls `get_state`, `set_state` and `integrate`?

I will also do a more line by line review shortly.

Ericgig:
Here is [sesolve](https://github.com/Ericgig/qutip/blob/solver.integrator/qutip/solver/sesolve.py) using it. And here is the [base solver](https://github.com/Ericgig/qutip/blob/solver.integrator/qutip/solver/solver_base.py).
I planned to do the dimension check in the solver. So `Sesolve.step(t)` and will call will `Integrator.integrate` and build the `Qobj`.
Likewise `SeSolver.run` call `Integrator.run`, there the `Result` object create the `Qobj`. I want the solver to copy the state only if it stores it, not if it only compute the expectation values. 
All the physics, which I consider the dims check to be part of, is done at the `Solver` level.

We also don't have a no dims `QobjEvo`. 

I made the `Integrator` to focus more on the ODE mathematics, but the creation of `Qobj` is quite fast so I don't mind either ways for the input types, but the dims check will still be done at the `Solver` level, not `Integrator`.

qutip/solver/integrator.py
```diff
@@ -0,0 +1,185 @@
...
+        Parameters
+        ----------
+        args : dict
+            New arguments
+        """
+        self.system.arguments(args)
+        self.reset()
+
+
+# TODO: These integrator set will be part of the solver classes
```
hodgestar, line 145: Agreed that the add_integrators and these sets belongs closer to the solvers, but happy with this as an intermediate step.

qutip/solver/integrator.py
```diff
@@ -0,0 +1,185 @@
...
+    if not issubclass(integrator, Integrator):
+        raise TypeError(f"The integrator {integrator} must be a subclass"
+                        " of `qutip.solver.Integrator`")
+    if not isinstance(keys, list):
+        keys = [keys]
+    for key in keys:
+        integrator_set[key] = integrator
+    if integrator.used_options and options_class:
+        for opt in integrator.used_options:
+            options_class.extra_options.add(opt)
```
hodgestar, line 185: It's all very implicit that these end up being provided in `options.ode`. It would be good if that was clearer. Perhaps passing just the SolverOdeOptions to the integrator would make this better? Then at least it would be clear that the `options` received are the same class that is passed when registering the options.

Registering all the ODE options on the same `SolverOdeOptions` class also creates a namespace clash. E.g. All the zvode integrators register their options on `SolverOdeOptions` for each type of solver, but if, say, a particular integrator does not support mcsolve, then having only one `SolverOdeOptions` would mean that the option was available for mcsolve even though the integrator is not available on mcsolve.

Should we perhaps have one `SolverOdeOptions` per solver set? Doesn't necessarily have to happen in this PR.

hodgestar (commented):
@Ericgig I've reviewed the integrator base class and solver set mechanism and left some suggestions. I'll review the zvode integrators and tests tomorrow.

Are their tests for the zvode integrators yet?

qutip/solver/integrator.py
```diff
@@ -0,0 +1,185 @@
...
+    # Used options in qutip.SolverOdeOptions
+    used_options = []
+    # Can evolve time dependent system
+    support_time_dependant = None
+    # Use the QobjEvo's matmul_data method as the driving function
+    use_QobjEvo_matmul = None
+
+    def __init__(self, system, options):
+        self.system = system
+        self._stats = {}
```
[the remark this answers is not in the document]
Ericgig: I wanted to have stats, like total number of steps, average step length, number of failed steps, etc. It would help debug and optimize integration.
But it's not that simple to extract them from scipy, so I removed it at the last minute, but I still want it...
I can have them for solver we make ourselves, so it could stay empty for scipy integrator and but be in for verner methods... Or we drop it altogether...
I am removing it for now and think about it later, it's really just a nice to have but not worth delaying v5 for.

qutip/solver/integrator.py
Ericgig: Changed every `SolverOptions` to `SolverOdeOptions`.

Yes, the options handling is not great and one per Solver would be better. But this can wait for now, as long as we can access the desired option.

qutip/solver/integrator.py
```diff
@@ -0,0 +1,185 @@
...
+
+    options: qutip.SolverOptions
+        Options for the solver.
+    """
+    # Used options in qutip.SolverOdeOptions
+    used_options = []
+    # Can evolve time dependent system
+    support_time_dependant = None
+    # Use the QobjEvo's matmul_data method as the driving function
+    use_QobjEvo_matmul = None
```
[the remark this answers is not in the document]
Ericgig: - Since Solver also used option, I thought it would be clearer if self.option was always the same thing, but no they don't need all of options. If we define `self.options` to be a `dict` of only the needed options, we remove a lot of duplication, good idea.

- `support_time_dependent` is used in test, only the [`IntegratorDiag`](https://github.com/qutip/qutip/pull/1644/files#diff-96d5a6258324bcf45eb2b4cdb89e8b15dafa8899f26152046d39c4b60b82f8c2R70), does not support it for now and it check that the system is constant.

- When `use_QobjEvo_matmul` is False, feedback, adaptative hilbert space and other modification to `matmul` will fail. I am thinking that it will also be enough to tell if we can expect auto differentiation to work with the `Integrator`. If we have an Integrator that does not use the `QobjEvo.matmul_data` it will probably be unable to detect it and produce wrong result. It could be part of `add_integrator` and we keep a set of operator that support feedback and one that don't, but adding a flag here is simpler. I expect a strong correlation between `support_time_dependent` and `use_QobjEvo_matmul` but not always. Looking at `fsesolve`, it could be an `Integrator` for `sesolve` that support time-dependant (periodic) system but not feedback.

Ericgig:
Thank you for doing this quickly, I guess you did until quite late.

Yes test_integrator is parameterized over all integrators registered in `sesolve_integrators`, etc.

qutip/solver/integrator.py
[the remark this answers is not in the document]
hodgestar: Having stats seems like a really good idea. Maybe let's leave it for another PR though because it requires a bit of thinking. It would also help with debugging (e.g. we could ask people file GitHub issues to post their solver stats).

qutip/solver/integrator.py
hodgestar: Sounds good for now.

hodgestar (commented):
I've gone over the solvers and tests now.

I think the `.integrate(t, step=....)` interface is currently way too complex, as evidenced by the difficulty of implementing it and the hard to resolve issues.

SciPy has a much simpler ODE interface -- see https://github.com/scipy/scipy/blob/v1.7.1/scipy/integrate/_ivp/base.py#L25-L214. Essentially a single `.step()` function.

Is there a reason we could not implement something similar here, and just have one `.step()` method and then implement `.integrate()` on top of that?

qutip/tests/solver/test_integrator.py
```diff
@@ -0,0 +1,101 @@
...
+    def me_method(self, request):
+        return request.param
+
+    @pytest.fixture(params=list(mcsolve_integrators.keys()))
+    def mc_method(self, request):
+        return request.param
+
+    def test_se_integration(self, se_method):
+        opt = SolverOdeOptions(method=se_method)
+        evol = sesolve_integrators[se_method](self.se_system, opt)
```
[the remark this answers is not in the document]
Ericgig: It's the solver that read `method` and retrieve the `Integrator`, it will rarely be done manually.

qutip/solver/ode/scipy_integrator.py
```diff
@@ -0,0 +1,315 @@
...
+    def _one_step(self, t):
+        """
+        Advance up to t by one internal solver step.
+        Should be able to retreive the state at any time between the present
+        time and the resulting time using `backstep`.
+        """
+        # integrate(t, step=True) ignore the time and advance one step.
+        # Here we want to advance up to t doing maximum one step.
+        # So we check if a new step is really needed.
+        self.back = self.get_state(copy=True)
```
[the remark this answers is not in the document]
Ericgig: Will add sanity checks.

qutip/tests/solver/test_integrator.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class TestIntegratorCte():
+    _analytical_se = lambda _, t: np.cos(t * np.pi)
+    se_system = qutip.QobjEvo(-1j * qutip.sigmax() * np.pi)
+    _analytical_me = lambda _, t: 1 - np.exp(-t)
+    me_system = qutip.liouvillian(qutip.QobjEvo(qutip.qeye(2)),
+                                  c_ops=[qutip.destroy(2)])
+
+    @pytest.fixture(params=list(sesolve_integrators.keys()))
```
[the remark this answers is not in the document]
Ericgig: `sesolve_integrators` is temporary so let's remember this for the next PR.

Ericgig:
> I think the `.integrate(t, step=....)` interface is currently way too complex, as evidenced by the difficulty of implementing it and the hard to resolve issues.
> 
> SciPy has a much simpler ODE interface -- see https://github.com/scipy/scipy/blob/v1.7.1/scipy/integrate/_ivp/base.py#L25-L214. Essentially a single `.step()` function.
> 
> Is there a reason we could not implement something similar here, and just have one `.step()` method and then implement `.integrate()` on top of that?

The `.step()` is essentially `.integrate(self.t)`, sot there would be no need for both.
If you mean having `step()` always advance one internal integrator step, we don't always have that possibility, `dop853` cannot advance only one step. And it would require to safe `self.back` each time even when not needed.
If you want a method to just advance to the time `t`, what ever it is as `scipy` seems to do, it will crush `mcsolve` speed. `mcsolve` has a root finding step that require a some back and forth and ode schemes all have their quirks when doing this.

For actual use `integrate`, `step` and `backstep` can be 3 methods. I had this in the evolver PR but was told to merge them...

The complexity could be split: have `integrate` for normal evolution (`sesovle`, `mesolve`), and `step` or `mcstep` for `mcsolve`.
This would also make it easier to create new Integrator for special use, it is easier to choose not to support `mcsolve` if the logic is split.

hodgestar (changes requested):
Let's also:
* add the default `mcstep` method.
* update the options list on the class to include the default option values.

Ericgig:
@hodgestar 
I added default values to options and a default `mcstep`. I added a doc-string entry for `integrator_options`, but please check it.

About `mcstep` optimization for scipy's `ode`, I tested both the methods and there can be a factor 2 in timing between both, but it can be in both ways depending on the conditions. So I used `mcsolve`'s method for `zvode` and the new method for `lsoda`. This way, the default will run like `v4`, but you have an alternative when it's not ideal.

qutip/solver/integrator.py
```diff
@@ -0,0 +1,262 @@
...
+        keys, not the full options object passed to the solver. Options' keys
+        included here will be supported by the :cls:SolverOdeOptions.
+    """
+    # Used options in qutip.SolverOdeOptions
+    integrator_options = {}
+    # Can evolve time dependent system
+    support_time_dependant = None
+    # Use the QobjEvo's matmul_data method as the driving function
+    supports_blackbox = None
+    name = ""
```
hodgestar, line 63:
```suggestion
    # The name of the integrator
    name = None
```
I suggest we set this to `None` so that it is an error to not specify a name for the solver.

qutip/solver/integrator.py
```diff
@@ -0,0 +1,262 @@
...
+    Parameters
+    ----------
+    system: qutip.QobjEvo
+        Quantum system in which states evolve.
+
+    options: qutip.SolverOdeOptions
+        Options for the solver.
+
+    Class Attributes
+    ----------------
```
hodgestar, line 36:
```suggestion
    ----------------
    name : str
        The name of the integrator.

```

qutip/solver/integrator.py
```diff
@@ -0,0 +1,262 @@
...
+        A dictionary of options used by the integrator and their default
+        values. Once initiated, ``self.options`` will be a dict with the same
+        keys, not the full options object passed to the solver. Options' keys
+        included here will be supported by the :cls:SolverOdeOptions.
+    """
+    # Used options in qutip.SolverOdeOptions
+    integrator_options = {}
+    # Can evolve time dependent system
+    support_time_dependant = None
+    # Use the QobjEvo's matmul_data method as the driving function
```
hodgestar, line 61:
```suggestion
    # Whether the integrator used the system QobjEvo as a blackbox
```

qutip/solver/integrator.py
```diff
@@ -0,0 +1,262 @@
...
+        Parameters
+        ----------
+        t : float
+            Initial time
+
+        state0 : qutip.Data
+            Initial state.
+
+        .. note:
+            It should set the flags `_is_set` to True.
```
hodgestar, line 97: Does `set_state` also need to set `self._back` to something now? Perhaps only if the default `mcstep` implementation is used? If yes, we should add a sentence here mentioning what `set_state` should do.

hodgestar (changes requested):
Left some small suggestions, but otherwise looks good now.

qutip/solver/integrator.py
Ericgig: No, it's not needed, it's used when the desired time between the 2 previous steps. So having them both at the same values will not help. The value set in `__init__` (`t_back = inf`) is enough to catch bad use (starting the evolution in the wrong direction, which we don't support.)

hodgestar: Woot. Let's leave things as they are there.

hodgestar (approved):
Looks good!

PR #1644, opened 2021-08-26
New Integrator methods

Ericgig:
Add new ODE integration methods:
`vern7`, `vern9`  and `diag`.

Verner 's methods are high order Runge-Kutta method with variable step size and dense output. (http://people.math.sfu.ca/~jverner/). Having our own qutip's implementation allow the state to be a qutip data-layer, instead of numpy arrays, which could be useful for gpu or auto differentiation. 

`diag` method diagonalize a constant system and solve it analytically, similar to what `essolve` did.
The states are numpy arrays since we don't have vectorized functions for our data layer.

Benchmark of these method are included with those of #1643.
Verner method tend to be faster than `adams` method for smaller systems and `diag` is also very fast.

For a diff with #1643, see https://github.com/Ericgig/qutip/pull/5.


nathanshammah:
I wonder if the Krylov-based method would fit in here too https://github.com/emilianomfortes/krylovsolver.

Ericgig:
@nathanshammah 
I expect it to be easy to use as an integrator:
` _evolve_krylov_tlist` is essentially the `run`, `time_evolution` is `integrate`.
The interface just need to be adapted.
It would work well with `sesolve` and `mcsolve`.

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+        if dt_needed <= self._dt_safe:
+            return dt_needed
+        return dt_needed / (int(dt_needed / self._dt_safe) + 1)
+
+    cdef double _recompute_safe_step(self, double err, double dt):
+        """ Get maximum safe step in function of the error."""
+        cdef double factor
+        if not self.adaptative_step:
+            return 0.
+        elif err == 0:
```
hodgestar, line 429: Do we need the explicit `err == 0` here? It looks like `factor` is clamped between `0.2` and `10` in the code below and an err of zero will give `0.2`?

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+        cdef double dt_needed = t - self._t_front
+        if not self.adaptative_step:
+            return dt_needed
+        if self.interpolate:
+            return self._dt_safe
+        if dt_needed <= self._dt_safe:
+            return dt_needed
+        return dt_needed / (int(dt_needed / self._dt_safe) + 1)
+
+    cdef double _recompute_safe_step(self, double err, double dt):
```
hodgestar, line 424: `_recompute_safe_step` doesn't seem to actually return anything? Perhaps it's return type should be void or int?

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+    cdef double _get_timestep(self, double t):
+        """ Get the dt for the step. """
+        cdef double dt_needed = t - self._t_front
+        if not self.adaptative_step:
+            return dt_needed
+        if self.interpolate:
+            return self._dt_safe
+        if dt_needed <= self._dt_safe:
+            return dt_needed
+        return dt_needed / (int(dt_needed / self._dt_safe) + 1)
```
hodgestar, line 422: I don't understand why we don't just return `self._dt_safe` at this point? This formula is almost the same as `self._dt_safe` but I'm guessing it is doing something I am not seeing?

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+    def t_prev(self):
+        return self._t_prev
+
+    @property
+    def t(self):
+        return self._t
+
+    def _debug_state(self):
+        if self._y is not None:
+            print("y", self._t, self._y.to_array())
```
hodgestar, line 475: For these could we always print `y`, etc but print `None` if the value is None? E.g. `print("y @ t =", self._t, ": ", None if self._y is None else self._y.to_array())`.

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+        if self._y_prev is not None:
+            print("prev", self._t_prev, self._y_prev.to_array(), self._norm_prev)
+        if self._y_front is not None:
+            print("front", self._t_front, self._y_front.to_array(),
+                  self._norm_front)
+        if self._y_temp is not None:
+            print("temp", self._y_temp.to_array())
+        print(self._dt_safe, self._dt_int)
+        for i in range(self.rk_extra_step):
+            print(self.k[i].to_array())
```
hodgestar, line 485: Here we should also print a label of some sort, maybe `"k[i=", i, "]: "`?

qutip/solver/ode/qutip_integrator.py
```diff
@@ -0,0 +1,119 @@
...
+__all__ = ['IntegratorVern', 'IntegratorDiag']
+
+
+class IntegratorVern(Integrator):
+    """
+    Qutip's implementation of Verner 'most efficient' Runge-Kutta method
+    of order 7 and 9. It's a Runge-Kutta method with variable steps and dense
+    output.
+    Use qutip's Data object for the state, allowing sparse or gpu states.
+    [http://people.math.sfu.ca/~jverner/]
```
hodgestar, line 19:
```suggestion
    QuTiP's implementation of Verner's "most efficient" Runge-Kutta method
    of order 7 and 9. These are Runge-Kutta methods with variable steps and dense
    output.

    The implementation uses QuTiP's Data objects for the state, allowing sparse, GPU or other
    data layer objects to be used efficiently by the solver in their native formats.
    
    See http://people.math.sfu.ca/~jverner/ for a detailed description of the methods.
```

qutip/solver/ode/qutip_integrator.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def mcstep(self, t, copy=True):
+        self._ode_solver.integrate(t, step=True)
+        self._check_failed_integration()
+        return self.get_state(copy)
+
+    def _check_failed_integration(self):
+        if self._ode_solver.successful():
+            return
+        messages = {
```
hodgestar, line 58: Perhaps the messages can made accessible directly on the ODE solver? I know scipy doesn't to this, but perhaps we can make ours nicer?

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+        # If True, will do the steps longer that the target and use
+        # interpolation to get the state at the wanted time.
+        self.interpolate = interpolate
+
+        self.k = []
+        if isinstance(method, tuple):
+            self._init_coeff(*method)
+        elif "vern7" == method:
+            self._init_coeff(*vern7_coeff)
+        elif "vern9" == method:
```
hodgestar, line 123: I wonder if we should name these `vern7-eff` and `vern9-eff` in case we want to implement `vern7-robust` later? Or at least allow `vern7-eff` as an alternative name?

hodgestar (changes requested):
Before doing the full review I didn't understand that these new methods would allow use to have solvers which use the Data objects directly -- that's great!

I think we need to have some explicit tests for the RK implementation because these things are fiddly and we'll be maintaining this for some time & I expect it will get used a lot.

I haven't look in detail, but it might be nice to also provide the corresponding Verner robust methods if those can avoid some solver pitfalls (which seem to catch people all the time).

The rest of my comments are mostly questions and small suggestions.

Happy new year! :D

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+        self.first_step = first_step
+        self.min_step = min_step
+        self.max_step = max_step
+
+        # If True, will do the steps longer that the target and use
+        # interpolation to get the state at the wanted time.
+        self.interpolate = interpolate
+
+        self.k = []
+        if isinstance(method, tuple):
```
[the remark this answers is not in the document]
hodgestar: Thanks!

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,485 @@
...
+        tmp_norm = frobenius_data(<Data> self.k[0])
+        for i in range(1, self.order+1):
+            factorial *= i
+        if tmp_norm != 0:
+            dt1 = ((tol * factorial * norm**self.order)**(1 /(self.order+1))
+                   / tmp_norm)
+        else:
+            dt1 = (tol * factorial * norm**self.order)**(1 /(self.order+1))
+        self._y_temp = copy_to(y0, self._y_temp)
+        self._y_temp = iadd_data(self._y_temp, <Data> self.k[0], dt1 / 100)
```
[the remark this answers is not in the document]
hodgestar: Could we add a comment that the `/ 100` is an arbitrary small fraction of dt1?

qutip/solver/ode/explicit_rk.pyx
```diff
@@ -0,0 +1,509 @@
...
+            factorial *= i
+        if tmp_norm >= (self.atol * 1e-6):
+            dt1 = ((tol * factorial * norm**self.order)**(1 / (self.order+1))
+                   / tmp_norm)
+        else:
+            dt1 = (tol * factorial)**(1 / (self.order+1)) * norm * 0.5
+
+        t1 = t + dt1 / 100
+        self._y_temp = copy_to(y0, self._y_temp)
+        self._y_temp = iadd_data(self._y_temp, <Data> self.k[0], dt1 / 100)
```
hodgestar, line 246:
```suggestion
        # below dt1 / 100 is an arbitrary small fraction of dt1:
        self._y_temp = iadd_data(self._y_temp, <Data> self.k[0], dt1 / 100)
```

hodgestar (approved):
I left one small suggestion that adds a comment, but otherwise looks great.

Thank you for making all the tiny improvements from my previous review!

hodgestar (approved):

PR #1645, opened 2021-08-27
Fix test_interpolate mcsolve use.

Ericgig:
In `test_interpolate`, `qzero` was used as `mcsolve`'s collapse operator. Due to numerical error in the interpolation method, `mcsolve` can detect a collapse, but a collapse for a null operator cause a a division by zero. This sometime made our tests fails.

By changing the collapse operator to `qeye`, the evolution, thus the test is unaffected, but it no longer randomly fail.


jakelishman (approved):

hodgestar:
@Ericgig Thanks!

PR #1647, opened 2021-08-27
Add operator support between Data types

Ericgig:
Doing  #1646, I saw that we use operator between data object in code where types are not strictly checked. We may have other test working only because we do test mostly in `CSR`... 

This PR add operators support between data-layer. This will help clean the code:
`_data.matmul(_data.matmul(A, B), C)` vs `A @ B @ C`.
and it will reduce the risk of them being used in functions where we expect, but don't check, that all data object are of the same type. 

It is added in `base.Data` so any new data-layer will have operators working without having to create them. It use the dispatched function of `add`, `sub`, `matmul`, `mul`, `imul`, and `neg`. But it has the side effect that operators can change the type if the specialization does not exist.

@jakelishman, is this something you though of but choose not to do?

### DIFF
--- a/qutip/core/data/base.pyx
+++ b/qutip/core/data/base.pyx
@@ -2,6 +2,7 @@
 
 import numpy as np
 cimport numpy as cnp
+import qutip.core.data as _data
 
 __all__ = [
     'idxint_dtype', 'Data', 'EfficiencyWarning',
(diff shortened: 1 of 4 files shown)


Ericgig:
Yes, we have binary operator defined for `Data` with object of the same type, but since for a while we only used `CSR`, there are probably some in the code where the `Data` types can be mixed. It will cause problem when we start using multiple types together.

Also `/=` is used in `Qobj`, so some `TfTensor` no having them will cause problem as it is right now.

We could remove operator support or fully support them, but the present something used, partially supported will cause issues.

I prefer full support. By adding operator in `base.Date`, new type don't need to do anything so it's not causing difficulties when creating new types and it will work as expected everywhere, so we don't have to hunt them down. It also make the code cleaner.

hodgestar:
I will put in a vote for removing support entirely because I would like data layer operations to require more typing for the following reasons:

1. Specializations methods and the implementation of an individual data layer should not use full support. By design they're intended to call specific implementation functions on known data types. Going via a dispatch mechanism like `__add__` rather than directly calling `add_csr_csr` adds code, obscures what is being called and hides bugs.

2. Qobj should not use full support because `qobj1 @ qobj2` is very different to `qobj1.data @ qobj2.data` and typing `_data.matmul(qobj1.data, qobj2.data)` is a good visual indicator that these are not at all the same things.

For other use cases I would much prefer if we kept objects as `Qobj` for as long as possible (possibly a Qobj with a known data type) and only used `qobj.data` sparingly.

Happy to hear counter arguments -- this is just my view coming in to the discussion.

jakelishman:
I'm also in favour of removing them entirely.  I think we added them in the first place to ease the transition as we converted the entire library over from v4 to v5, but not entirely sure - if I thought differently back then, I can't think what my arguments would have been. I agree with everything Simon said, and the point Eric brought up at the start is something we explicitly _didn't_ want to occur in practice.  Having a "type system" (in as many scare quotes as is possible) handle it is a much better way of doing this.

Ericgig:
The fear I have with removing them is I expect we will forget some in untested code...
But let's go with removing them.
Do we keep the ones with scalar: `Data / scalar`, `-Data`?
They probably won't cause problems, but to be consistent I would also remove them.

hodgestar:
> The fear I have with removing them is I expect we will forget some in untested code...
> But let's go with removing them.

Woot. Code that is entirely untested is probably already broken. ;) 

> Do we keep the ones with scalar: `Data / scalar`, `-Data`?
> They probably won't cause problems, but to be consistent I would also remove them.

I'm happy with those being removed or being simple calls to other methods on the Data class. The fewer methods Data layer classes have to implement themselves, the better.

hodgestar:
Closing in favour of #1649.

PR #1651, opened 2021-09-03
Add 'bloch_redfield_tensor' as a superoperator

Ericgig:
Change `bloch_redfield_tensor` to be an equivalent to the `liouvillian` function.
As the former, it is now placed in core and support the both constant and time-dependant systems. (Removing the need for `br_codegen.py`).

Previously, `bloch_redfield_tensor` returned the tensor in the frame where the Hamiltonian was diagonalized, now it offer the option between returning the tensor in that frame of the the lab frame.

The function `brterm` as  been added as the equivalent to `lindblad_dissipator`, it compute the contribution of one `a_op` of a bloch Redfielf tensor.

The last user facing addition is `SpectraCoefficient`, it transform a coefficient of `t` to one of `w`. The coupling with the environment include a frequency response function which, in `brmesolve`, could be a function, string or `Cubic_Spline`, thus a `Coefficient`, but with a `w` input. `SpectraCoefficient` is to have interpolation based coefficient take the `w` input.  

`QobjEvo` created by `brterm` use a custom element: `_BlochRedfieldElement` to allow faster matmul with a state. Since the operator in computed in a basis and the state is in another, having `matmul` change the state basis is usually faster than changing the operator basis.

Each `_BlochRedfieldElement` of a `QobjEvo` share a `_EigenBasisTransform`, the object that does the Hamiltonian diagonalization at each call.

It would be nice to support to computation of  the tensor using `Data`. but it would require `ndarray.__mul__` equivalent to the dispatched function.


hodgestar:
@Ericgig Is the plan to update qutip.solve.bloch_redfield and remove br_codegen, etc in a follow up PR?

Ericgig:
Yes, that was for another PR.
I could add it if there a rush.

qutip/solver/brmesolve.py
```diff
@@ -0,0 +1,212 @@
...
+    Solves for the dynamics of a system using the Bloch-Redfield master
+    equation, given an input Hamiltonian, Hermitian bath-coupling terms and
+    their associated spectrum functions, as well as possible Lindblad collapse
+    operators.
+
+    Parameters
+    ----------
+    H : Qobj / list
+        System Hamiltonian given as a Qobj or
+        nested list in string-based format.
```
hodgestar, line 31: We should probably make the description for `H` more similar to the description of `H` in other solvers (i.e. now it can accept anything `QobjEvo` accepts, not just a Qobj or list.

qutip/solver/brmesolve.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    H = QobjEvo(H, args=args, tlist=tlist)
+
+    c_ops = c_ops if c_ops is not None else []
+    if not isinstance(c_ops, (list, tuple)):
+        c_ops = [c_ops]
+    c_ops = [QobjEvo(c_op, args=args, tlist=tlist) for c_op in c_ops]
+
+    new_a_ops = []
+    for (a_op, spectra) in a_ops:
```
hodgestar, line 109: We should check that `a_op` is either `Qobj` or `QobjEvo` here, or try convert it to a QobjEvo perhaps.

qutip/solver/brmesolve.py
```diff
@@ -0,0 +1,212 @@
...
+        else:
+            raise TypeError("a_ops's spectra not known")
+
+    solver = BRSolver(
+        H, new_a_ops, c_ops, options=options,
+        use_secular=(sec_cutoff>=0.), sec_cutoff=sec_cutoff,
+    )
+
+    return solver.run(psi0, tlist, e_ops=e_ops)
+
```
hodgestar, line 133:
```suggestion


```
Just adding a missing newline between classes.

qutip/solver/brmesolve.py
```diff
@@ -0,0 +1,212 @@
...
+            a_ops = [
+                (a+a.dag(), coefficient('w>0', args={'w':0})),
+                (QobjEvo([b+b.dag(), f(t)]), coefficient(g(t, w))),
+                (c+c.dag(), SpectraCoefficient(coefficient(array, tlist=ws))),
+            ]
+
+    c_ops : list of :class:`Qobj`, :class:`QobjEvo`
+        Single collapse operator, or list of collapse operators, or a list
+        of Lindblad dissipator. None is equivalent to an empty list.
+
```
hodgestar, line 169: `use_secular` and `sec_cutoff` should be documented. Perhaps only `sec_cutoff` should be used to match `brsolve` above.

qutip/tests/solver/test_brmesolve.py
```diff
@@ -0,0 +1,294 @@
...
+    delta = 0.0
+    epsilon = 0.5 * 2 * np.pi
+    e_ops = pauli_spin_operators()
+    H = delta * 0.5 * qutip.sigmax() + epsilon * 0.5 * qutip.sigmaz()
+    psi0 = (2 * qutip.basis(2, 0) + qutip.basis(2, 1)).unit()
+    times = np.linspace(0, 10, 100)
+    me = qutip.mesolve(H, psi0, times, c_ops=me_c_ops, e_ops=e_ops).expect
+    brme = brmesolve(H, psi0, times, brme_a_ops,
+                     e_ops=e_ops, c_ops=brme_c_ops).expect
+    for me_expectation, brme_expectation in zip(me, brme):
```
hodgestar, line 39: We should do `zip(me.expect, brme.expect)` here for consistency with the other tests.

qutip/tests/solver/test_brmesolve.py
```diff
@@ -0,0 +1,294 @@
...
+    psi0 = (2 * qutip.basis(2, 0) + qutip.basis(2, 1)).unit()
+    times = np.linspace(0, 10, 100)
+    me = qutip.mesolve(H, psi0, times, c_ops=c_ops, e_ops=e_ops).expect
+    brme = brmesolve(H, psi0, times, [], e_ops=e_ops, c_ops=c_ops).expect
+    for me_expectation, brme_expectation in zip(me, brme):
+        np.testing.assert_allclose(me_expectation, brme_expectation, atol=1e-8)
+
+
+# test_time_dependence
+
```
hodgestar, line 171:
```suggestion
```
This comment seems a bit lost -- if it should stay, let's move it to the test it refers (to even better, just extend the name of the test to make it clear what the test is for).

hodgestar (changes requested):
This is a partial review that covers the `solver` package parts. I'll look at the `core` parts in a follow-up review.

qutip/core/_brtools.pyx
```diff
@@ -0,0 +1,321 @@
...
+        elif fock.shape[0] == self.size**2 and fock.shape[1] == 1:
+            if type(fock) is Dense and (<Dense> fock).fortran:
+                fock = _data.column_unstack_dense(fock, self.size, True)
+                temp = _data.matmul(matmul_var_data(self.evecs(t), fock, 3, 0),
+                                    self.evecs(t))
+                fock = _data.column_stack_dense(fock, True)
+            else:
+                fock = _data.column_unstack(fock, self.size)
+                temp = _data.matmul(matmul_var_data(self.evecs(t), fock, 3, 0),
+                                    self.evecs(t))
```
hodgestar, line 275: Should we restack `fock` here?

qutip/core/_brtools.pyx
```diff
@@ -0,0 +1,321 @@
...
+            if type(temp) is Dense:
+                return _data.column_stack_dense(temp, True)
+            return _data.column_stack(temp)
+
+        if fock.shape[0] == self.size and fock.shape[0] == fock.shape[1]:
```
hodgestar, line 280:
```suggestion
        elif fock.shape[0] == self.size and fock.shape[0] == fock.shape[1]:
```

qutip/core/_brtools.pyx
```diff
@@ -0,0 +1,321 @@
...
+            return _data.matmul(matmul_var_data(self.evecs(t), fock, 3, 0),
+                                self.evecs(t))
+
+        elif fock.shape[0] == self.size**2 and fock.shape[0] == fock.shape[1]:
+            temp = self._S_converter_inverse(t)
+            return _data.matmul(matmul_var_data(temp, fock, 3, 0), temp)
+
+        raise ValueError
```
hodgestar, line 288: Missing error message.

qutip/core/_brtools.pyx
```diff
@@ -0,0 +1,321 @@
...
+
+        elif eig.shape[0] == self.size and eig.shape[0] == eig.shape[1]:
+            temp = self.evecs(t)
+            return matmul_var_data(_data.matmul(temp, eig), temp, 0, 3)
+
+        elif eig.shape[0] == self.size**2 and eig.shape[0] == eig.shape[1]:
+            temp = self._S_converter_inverse(t)
+            return _data.matmul(temp, matmul_var_data(eig, temp, 0, 3))
+
+        raise ValueError
```
hodgestar, line 321: Missing error message.

hodgestar (changes requested):
We seem to also be missing tests for the `core/blochredfield.py` functions.

Was some of the cython copied from the old bloch redfield code?

qutip/core/_brtools.pyx
Ericgig: No need, this was not inplace.

qutip/core/_brtools.pyx
```diff
@@ -0,0 +1,321 @@
...
+class _eigen_qevo:
+    """
+    Callable function to represent the eigenvectors of a QobjEvo at a time
+    ``t``.
+    """
+    def __init__(self, qevo):
+        self.qevo = qevo
+        self.args = None
+
+    def __call__(self, t, args):
```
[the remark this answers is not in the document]
Ericgig: No, this is not directly seen by the user and splitting and rebuilding the dict is less efficient.

qutip/core/_brtools.pyx
```diff
@@ -0,0 +1,321 @@
...
+    """helper function for matmul_var_Dense, fetch the blas flag byte"""
+    if trans == 0:
+        return b'N'
+    elif trans == 1:
+        return b'T'
+    elif trans == 3:
+        return b'C'
+
+
+cpdef Data matmul_var_data(Data left, Data right,
```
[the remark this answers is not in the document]
Ericgig: Should this function be part of `core`?  `A.dag @ B` is quite common. `blas` and `tensorflow` support it,

Ericgig:
@hodgestar , I think I responded to all your comments.
The tests for core/blochredfield.py are present, I think it has a 100% coverage.
I did little cut and paste, but stayed close to the original cython code when possible.

The question I still have is about the `a_ops` format: I made breaking changes from v4, so I want to be sure the new way is clear.

I think `matmul_var_data` should be available in `core/data` not implemented here, but I don't want to touch that part in this PR.

qutip/core/_brtools.pyx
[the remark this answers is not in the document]
hodgestar: Not sure.  I don't want to overcomplicate the data layer, but it's not a giant extra complexity and if blas and tensorflow support it, then perhaps it makes sense.

Let's leave it for a follow up PR though.

qutip/core/blochredfield.py
```diff
@@ -0,0 +1,162 @@
...
+            Can be a `Coefficient` using an 'w' args or a function of the
+            frequency. The `SpectraCoefficient` can be used to use array based
+            coefficient. They can also depend on ``t`` if the corresponding
+            ``a_op`` is a :cls:`QobjEvo`.
+
+        Example:
+            a_ops = [
+                (a+a.dag(), coefficient('w>0', args={"w": 0})),
+                (QobjEvo([b+b.dag(), f(t)]), g(w)),
+                (c+c.dag(), SpectraCoefficient(coefficient(Cubic_Spline))),
```
hodgestar, line 45: I think the new structure of `a_ops` is nice, but I think the documentation here is currently a little imprecise.

- the `a_op` always seems to be converted to a QobjEvo and the spectra is always called with `spectra(t, w=...)` so I'm not sure that functions that depend only on `w` will work?
- the example passes `f(t)` and `g(w)` but presumably one would type `[..., f]` and `(QobjEvo(...), g)` and the `t` and `w` are meant to indicate the signatures of the two functions. Perhaps the signatures could be indicated another way?
- the long `SpectraCoefficient(coefficient(Cubic_Spline))` is a bit unwieldy and I'm not sure what `Cubic_Spline` is supposed to be (maybe this was written before the new order parameter for array  coefficients)

It would be good to have a nicer way to pass in an ndarray of coefficients and perhaps specify the order.

Would it be worthwhile trying to support a 2-d ndarray for the time-dependent case (so that would could specify a mesh of values)? Perhaps that is just too messy -- mesh interpolation is much more involved that 1-d interpolation.

The example implies that the code could be written roughly as is, but that is not the case for the last two elements of a_ops.

hodgestar:
@Ericgig I left some suggestions for improving a_ops, but everything else looks good to me now.

qutip/core/blochredfield.py
Ericgig: Functions of `w` only are accepted in `brmesolve` and only there. Here it need to be already a `Coefficient`, good catch.

The way I see to have the signature clear is to use lambda, I changed it, tell me if you have a better idea.

`Cubic_Spline` is sometime we are removing with #1725. I merged `dev,major` and remove all it's use.

For `ndarray`, the user need to have created the `Coefficient` before calling this, so he can set the order then. As for 2-d ndarrays, I don't think it's worth it. As you say, it a lot more messy that 1-d and it's use case is limited, so for those cases using a function or string is enough. If there is demand for it we can add it later.

hodgestar (approved):
Approved!

PR #1652, opened 2021-09-07
Minor doc fix in QubitCircuit

Erotemic:
**Description**

The `num_cbits` wasn't documented in the QubitCircuit docstring. This adds it.

Took me a second to figure out what it was (as a novice in quantum computing), so I figured I'd save some other novice that second.

FWIW: `reverse_states` isn't documented either, but I'm not confident I can write an accurate description of what that is, otherwise I would have put it here too.

This library is really neat! Thanks to all who are developing and maintaining it! 

**Changelog**

Add docs for `num_cbits` in QubitCircuit docstring

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -364,6 +364,8 @@ class QubitCircuit:
         A list of integer for the dimension of each composite system.
         e.g [2,2,2,2,2] for 5 qubits system. If None, qubits system
         will be the default option.
+    num_cbits : int
+        Number of classical bits in the system.
 
     Examples
     --------


hodgestar (approved):

hodgestar:
@Erotemic Thank you for documenting this. I will merge it shortly & have marked if for inclusion in QuTiP 4.6.3.

Note that qutip-qip has also been split out into it's own package (https://github.com/qutip/qutip-qip) in preparation for QuTiP 5 so we should update it there too.

`reverse_states` seems to reverse the order of the array of quantum and classical bits. It is a bit odd, so I'm happy to leave it undocumented. The original commit just read:

```
commit dcf8d363718b6ee5a8a17ec9af00df4bc8aea166
Author: Robert Johansson
Date:   Mon May 26 19:11:44 2014 +0900

    allow reversing the circuit represenation. added support for multigate symbol
```

@BoxiLi Just pinging you so you know I merged this and can comment if you like.

PR #1656, opened 2021-09-10
Add a tolerance for measurement.

Ericgig:
Measurements would return a `None` state when the probability is 0.
Due to recent changes which affected operation order, a test for `CircuitSimulator` would result in a probability of `~1e-33`. Not being 0, this resulted in an unexpected possible result of the circuit simulation, breaking tests for `dev.major`

I changed the test from `p != 0` to `p >= tol` with `tol` being set in core's settings.

### DIFF
--- a/qutip/core/options.py
+++ b/qutip/core/options.py
@@ -29,6 +29,10 @@ class CoreOptions:
         The absolute tolerance used in automatic tidyup (see the ``auto_tidyup``
         parameter above) and the default value of ``atol`` used in
         :method:`Qobj.tidyup`.
+
+    min_probability : float {1e-30}
+        The minimun probability for measurement considered to be possible in
+        :func:`measurement_statistics_povm`.
     """
     options = {
         # use auto tidyup
@@ -42,5 +46,7 @@ class CoreOptions:
         # general relative tolerance
         "rtol": 1e-12,
         # use auto tidyup absolute tolerance
-        "auto_tidyup_atol": 1e-14
+        "auto_tidyup_atol": 1e-14,
+        # cutoff for impossible collapse in measurment
+        "min_probability": 1e-30
     }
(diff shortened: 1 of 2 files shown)


hodgestar:
@Ericgig Thanks for finding this. I don't think we should add a setting for this though -- it's too obscure. The point of returning `None` when `p == 0` seems to be to avoid division by zero, not to round down small probabilities.

I think we should either 1) update the test or 2) if we lots some precision in a recent change, try to get that precision back. Do you know which change caused the loss of precision?

Ericgig:
No, I don't know what exactly caused this. All operation are through `Qobj`. Maybe `tidyup` have some effect.
But testing for a pure `0` with floating number is never a good idea.

Which test do you think we should update. `test_quantumcircuit` is testing what it should.
`test_measurement` tests case which have `0` probabilities with `np.testing.assert_almost_equal`.

BoxiLi:
The failing test in `CircuitSimulator` is likely from `run_statistics`, which collects all the possible outcomes wrt. the measurement result. If the measurement result is `None`, it will skip that state. I can also set tolerance there.

hodgestar:
I think it's quite important to understand why this happened. Previously we reliably had an exactly correct result, and now we don't. This particular check for `p != 0` of a float is I think okay because it's there specially to avoid division by exactly zero.

I can think of at least two plausible scenarios:

1. The issue was caused by the removal of auto_tidyup, in which case I think either a setting like this with an explicit check in either the measurement functions or the simulator are good solutions (although it should be possible to override the setting on a per-call basis, just like with atol).

2. The issue was caused by the removal of the data layer mathematical operations, and in particular, by the replacement of `data / value` by `data.mul(data, 1 / value)`, in which case I think the solution is to add a `div` operation to the data classes so that we don't lose this accuracy.

PR #1657, opened 2021-09-10
fix seeds in stochastic tests 

Ericgig:
Some tests in stochastic evolution passes only most of the time.
The problematic tests are comparison of different step length, normally error should be smaller with smaller step length. Since the evolution use random number, this is not true for all seeds.

By fixing the test, we should not have the tests randomly failing.
I am adding _improving stochastic tests and having them use pytest_ to my to-do.

Ps. tests fail due to the error fixed in #1656

### DIFF
--- a/qutip/tests/solve/test_stochastic_me.py
+++ b/qutip/tests/solve/test_stochastic_me.py
@@ -92,6 +92,8 @@ def test_smesolve_homodyne_methods():
                         ['explicit1.5', 1e-4],
                         ['taylor2.0', 1e-4]]
     for n_method in list_methods_tol:
+        # Comparisons of error between sol and sol3 depend on the stochastic
+        # noise, thus the seed, fixing the seed remove random fails.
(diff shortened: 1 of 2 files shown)


qutip/tests/solve/test_stochastic_me.py
```diff
@@ -92,6 +92,7 @@ def arccoth(x):
                         ['explicit1.5', 1e-4],
                         ['taylor2.0', 1e-4]]
     for n_method in list_methods_tol:
+        np.random.seed(1)
```
hodgestar, line 95: Maybe the random seed should be outside the for loop and have a comment explaining why it is needed?

hodgestar:
@Ericgig I re-triggered the builds. If that doesn't fix the tests, could you merge dev.major into this PR?

hodgestar (approved):

hodgestar:
Thanks! Merging.

PR #1658, opened 2021-09-10
fix failing circuit test

Ericgig:
Fix the test breaking dev.major by adding the tolerance in the test itself.
This is an alternative to #1656, merge only one.


hodgestar (approved):
I'm approving this so that builds will pass again, but we should continue to try understand what happened in #1656.

PR #1669, opened 2021-09-21
Make mcsolve more memory efficient

### LINKED ISSUE #1667: mcsolve always store the final state.
**Describe the bug**
`mcsolve` always store the last state of a trajectory and throws it away after the evolution if the options is `False`, making `mcsolve` quite memory inefficient.

See 
**Expected behavior**
`mcsolve` should read the options before the evolution and only store the states if needed.

**Additional context**
https://groups.google.com/g/qutip/c/tzU_6BcbUrs

Ericgig:
Presently `mcsolve` always save the final state of all trajectories, needed or not.
With this PR, it will only save them when needed.

**Related issues or PRs**
closes #1667

**Changelog**
Improve mscolve memory efficiency when states are not needed.

### DIFF
--- a/qutip/cy/mcsolve.pyx
+++ b/qutip/cy/mcsolve.pyx
@@ -32,15 +32,15 @@ cdef np.ndarray[complex, ndim=1] normalize(complex[::1] psi):
 
 cdef class CyMcOde:
     cdef:
-        int steady_state, store_states, col_args
+        int steady_state, store_states, col_args, store_final_state
         int norm_steps, l_vec, num_ops
         double norm_t_tol, norm_tol
         list collapses
         list collapses_args
         list c_ops
         list n_ops
-        complex[:,::1] states_out
-        complex[:,::1] ss_out
+        object states_out
+        object ss_out
         double[::1] n_dp
 
     def __init__(self, ss, opt):
@@ -51,6 +51,7 @@ cdef class CyMcOde:
         self.norm_tol = opt.norm_tol
         self.steady_state = opt.steady_state_average
         self.store_states = opt.store_states or opt.average_states
+        self.store_final_state = opt.store_final_state
         self.collapses = []
         self.l_vec = self.c_ops[0].cte.shape[0]
         self.num_ops = len(ss.td_n_ops)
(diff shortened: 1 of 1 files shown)


nathanshammah (approved):
LGTM, thank you @Ericgig.

qutip/cy/mcsolve.pyx
```diff
@@ -32,15 +32,15 @@ cdef np.ndarray[complex, ndim=1] normalize(complex[::1] psi):
...
         int norm_steps, l_vec, num_ops
         double norm_t_tol, norm_tol
         list collapses
         list collapses_args
         list c_ops
         list n_ops
-        complex[:,::1] states_out
-        complex[:,::1] ss_out
+        object states_out
+        object ss_out
```
jakelishman, line 43: This feels like it might possibly have some knock-on performance issues due to less perfect indexing.  I doubt it _will_, because these aren't inner loop variables, but could we check?  I also couldn't immediately see why they needed to be changed at all, but that may just be me missing things.

Ericgig: Inside loop I used `cdef complex[:,::1] states_out = self.states_out` to ensure no impact on performance.

I changed it to keep the numpy array alive. We create it with `self.states_out = np.zeros(...)` in a function and do `np.array(self.states_out)` in another and I am not sure it is a safe usage for reference counting. There haven't been any issues, but I preferred to use only one `ndarray` just in case. 

Also `np.array(self.states_out)` don't work on not initialized memoryview, so we would have to branch the `return`, but as an object, it is `None` when not used. (Which I should have mode explicit...)

jakelishman: I'm surprised that assigning to the memoryview doesn't keep the numpy array alive - I thought the base object got stored within the view, to avoid problems with that (similar to how numpy views keep references to their underlying objects to prevent the memory going stale).

Either way, I'm not actually bothered about it, since as you say, there shouldn't be any performance issues.

hodgestar (approved):

hodgestar:
@Ericgig Looks good. Should this go into 4.6.3 or only 4.7?

hodgestar:
@Ericgig Could you add a changelog entry to the PR description before merging?

PR #1670, opened 2021-09-21
Add `element_wise_multiply`

Ericgig:
Add `element_wise_multiply`, the equivalent to `numpy * numpy` to the dispatched function.
This function is needed to create the Bloch-Redfield tensor working in `Data` object.
The name is willingly long to discourage it's use as it is rarely a valid operation on quantum objects.

### DIFF
--- a/qutip/core/data/matmul.pyx
+++ b/qutip/core/data/matmul.pyx
@@ -15,11 +15,12 @@ import numpy as np
 cimport numpy as cnp
 from scipy.linalg cimport cython_blas as blas
 
+from qutip.core.data.base import idxint_dtype
 from qutip.core.data.base cimport idxint, Data
 from qutip.core.data.dense cimport Dense
 from qutip.core.data.csr cimport CSR
 from qutip.core.data cimport csr, dense
-from qutip.core.data.add cimport iadd_dense
+from qutip.core.data.add cimport iadd_dense, add_csr
 
 cnp.import_array()
 
@@ -38,7 +39,7 @@ cdef extern from "src/matmul_csr_vector.hpp" nogil:
 
 __all__ = [
     'matmul', 'matmul_csr', 'matmul_dense', 'matmul_csr_dense_dense',
-    'element_wise_multiply', 'element_wise_multiply_csr', 'element_wise_multiply_dense',
+    'multiply', 'multiply_csr', 'multiply_dense',
 ]
 
 
@@ -290,7 +291,7 @@ cpdef Dense matmul_dense(Dense left, Dense right, double complex scale=1, Dense 
     return out
 
 
-cpdef CSR element_wise_multiply_csr(CSR left, CSR right):
+cpdef CSR multiply_csr(CSR left, CSR right):
     """Element-wise multiplication of CSR matrices."""
     if left.shape[0] != right.shape[0] or left.shape[1] != right.shape[1]:
         raise ValueError(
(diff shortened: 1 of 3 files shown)


Ericgig:
I also made a [`einsum` dispatched function](https://github.com/Ericgig/qutip/tree/einsum) as an alternative to this, (with a working `einsum_csr` for a pair of csr matrices).
However, since our object are all 2D, most `einsum` can be represented with `matmul`, `element_wise_multiply` and `transpose`.
And with the dispatcher needing a fixed number of input, it cannot be applied to an arbitrary number of array like numpy can.
So I don't see any advantage with it and unless there is good interest, I don't plan to do anything with it.

qutip/core/data/matmul.pyx
```diff
@@ -289,6 +290,61 @@ cpdef Dense matmul_dense(Dense left, Dense right, double complex scale=1, Dense
...
+    cdef idxint col_left, left_nnz = csr.nnz(left)
+    cdef idxint col_right, right_nnz = csr.nnz(right)
+    cdef idxint ptr_left, ptr_right, ptr_left_max, ptr_right_max
+    cdef idxint row, nnz=0, ncols=left.shape[1]
+    cdef CSR out
+    # Fast paths for zero matrices.
+    if right_nnz == 0 or left_nnz == 0:
+        return csr.zeros(left.shape[0], left.shape[1])
+    # Main path.
+    out = csr.empty(left.shape[0], left.shape[1], max(left_nnz, right_nnz))
```
hodgestar, line 311: Assuming we have to support nans, some of this logic will fail. E.g. `[nan, nan] * [0, 0] -> [nan, nan]` even though `right_nnz == 0` and similarly `[nan, 0] * [0, nan] -> [nan, nan]` which has two non-zero elements even though `max(left_nnz, right_nnz) == 1`.

qutip/core/data/matmul.pyx
```diff
@@ -289,6 +290,61 @@ cpdef Dense matmul_dense(Dense left, Dense right, double complex scale=1, Dense
...
+    ptr_left_max = ptr_right_max = 0
+    for row in range(left.shape[0]):
+        ptr_left = ptr_left_max
+        ptr_left_max = left.row_index[row + 1]
+        ptr_right = ptr_right_max
+        ptr_right_max = right.row_index[row + 1]
+        while ptr_left < ptr_left_max and ptr_right < ptr_right_max:
+            col_left = left.col_index[ptr_left] if ptr_left < ptr_left_max else ncols + 1
+            col_right = right.col_index[ptr_right] if ptr_right < ptr_right_max else ncols + 1
+            if col_left == col_right:
```
hodgestar, line 322: A similar issue to the `nan` comment above applies here. If `col_left != col_right` for any `col_right`, we might still need a `nan` entry corresponding to `col_left`.

hodgestar (commented):
Left a naming suggestion and a question about the behaviour for the wierd float values like `nan`, `inf`, etc.

hodgestar:
> I also made a [`einsum` dispatched function](https://github.com/Ericgig/qutip/tree/einsum) as an alternative to this, (with a working `einsum_csr` for a pair of csr matrices).

Cool! I agree that element-wise multiply is a pretty common and simple operation though, so it's probably worthwhile having regardless of whether we have einsum or not.

hodgestar (approved):

PR #1678, opened 2021-09-28
Fix equation number rendering and bring back site.css.

hodgestar:
**Description**
A bug in the CSS of the readthedocs Sphinx theme causes equation numbers to be rendered *above* the equation, which is both surprising and looks terrible. This PR applies fixes that have been languishing in the RTD theme GitHub pull request queue.

While working on this I discovered that our `site.css` file was not actually included in our generated HTML pages, so I added it in.

**Related issues or PRs**
* RTD theme issue: https://github.com/readthedocs/sphinx_rtd_theme/issues/301
* RTD theme fix: https://github.com/readthedocs/sphinx_rtd_theme/pull/383/

**Changelog**
Add back the custom CSS styling to (site.css) to our generated HTML documentation.
Fix the rendering of equation numbers (they now appear on the right as expected, not above the equation).


hodgestar:
@jakelishman @nathanshammah I'm guessing out site.css file go dropped accidentally at some point, but I don't actually know why the custom styling in it is there. I could just delete the styling that was there before and we could slowly re-add it if we need to?

jakelishman:
I don't know anything about that file here - I think it must have been dropped before I started being involved.

hodgestar:
> I don't know anything about that file here - I think it must have been dropped before I started being involved.

Tx. Hopefully Nathan will know. If not, I think I will just delete all the styling that was there to keep us closer to the RTD theme.

hodgestar:
The styling all got added in 5b4777450f5ba2becb790efb3af85482a1704c73 which has the cryptic commit message "fix github issue". :)

hodgestar:
Old styling deleted. Will merge once the builds pass.

PR #1679, opened 2021-09-30
Update steadystates tests to pytest

Ericgig:
**Description**
Update tests in `test_steadystate` to use pytest.
Improve coverage by adding tests for previously untested method and options and tests for raised errors.

It will not reach full coverage: no tests use `debug=True`, `pinv2` is deprecated on `scipy` side, some checks are redundant, etc. Some work on `steadystate.py` will be needed, but this is for another time, possibly for dev.major.

**Changelog**
Update steadystates tests to pytest for better coverage.

### DIFF
--- a/qutip/steadystate.py
+++ b/qutip/steadystate.py
@@ -242,10 +242,10 @@ def steadystate(A, c_op_list=[], method='direct', solver=None, **kwargs):
                 solver = 'mkl'
     elif solver == 'mkl' and \
             (method not in ['direct', 'power']):
-        raise Exception('MKL solver only for direct or power methods.')
+        raise ValueError('MKL solver only for direct or power methods.')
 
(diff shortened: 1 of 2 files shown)


qutip/steadystate.py
```diff
@@ -1206,10 +1206,9 @@ def _pseudo_inverse_sparse(L, rhoss, w=None, **pseudo_args):
         else:
             pspec = pseudo_args['permc_spec']
             diag_p_thresh = pseudo_args['diag_pivot_thresh']
-            pseudo_args = pseudo_args['ILU_MILU']
```
hodgestar, line 1209: This was a bizarre line of code! Yay for removing it.

hodgestar (approved):
Thanks for cleaning this up.

PR #1680, opened 2021-09-30
Change test_utilities to pytest

Ericgig:
**Description**
Update tests in `test_utilities.py` to `pytest` and add test for previouly untested function.
I removed 2 functions:
`linspace_with`: never used, almost equivalent to `np.linspace`.
`view_methods`: supposed to print methods of a `Qobj`, but broken...

**Changelog**
Update tests in test_utilities.py to pytest.
Remove broken qutip.utilities.view_methods.
Remove qutip.utilities.linspace_with (use numpy.linspace instead).

### DIFF
--- a/doc/apidoc/functions.rst
+++ b/doc/apidoc/functions.rst
@@ -334,7 +334,7 @@ Utility Functions
 -----------------
 
 .. automodule:: qutip.utilities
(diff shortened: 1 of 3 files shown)


hodgestar (approved):
+1 on removing `linspace_with` and `view_methods`.

PR #1688, opened 2021-10-13
Hadamard transform efficient

AGaliciaMartinez:
**Description**
Improved efficiency of the hadamard gate creation time as it was using python for loops which are terribly slow. For N=10 it goes down from 522ms to roughly 7ms.

**Changelog**
Improved hadamard_transform efficiency (70x faster for N=10).

### DIFF
--- a/qutip/qip/operations/gates.py
+++ b/qutip/qip/operations/gates.py
@@ -894,7 +894,7 @@ def hadamard_transform(N=1):
     """
     H = Qobj([[1, 1], [1, -1]]) / np.sqrt(2)
 
-    return tensor([H]*N)
+    return tensor([H] * N)
 
 
 def _flatten(lst):


AGaliciaMartinez:
> What made you trip over this?

I was using this function with N=13 for some testing code I have and realised it took too long to create the matrix.

jakelishman:
There's another interesting technique that's sometimes applicable in these sort of repeating cases.  Rather than passing a list of the same elements to `tensor`, you can also consider building up the tensor product manually, going up in powers of two.  So you go along with something like:
```python
out, tmp = H, H
for i in n.bit_length():
    tmp = tensor(tmp, tmp)
    if n & (1 << i):
        out = tensor(out, tmp)
```
(very very approximately - I've almost certainly got the indexing wrong).  The idea is that you only perform `lg(n)` tensor products.

Now, this might not actually give you much of a speed up in this case, because the calculation is still going to be dominated by the final tensor product, since the output matrix keeps getting larger.  But in things like integer matrix powers, bigint calculations, or binomial expansions, this type of thing can be super useful!  QuTiP 5 uses an algorithm like this for matrix powers.  (I imagine SciPy does too, but for all the normal reasons, we don't use their implementation.)

Issue #1684, opened 2021-10-04
What is rhs_xxx.pyx or cqobjevo_compiled_xxx.pyx file?


KosukeMizuno:
**Is your feature request related to a problem? Please describe.**

Sometimes qutip.mesolver generate `rhs_**.pyx` or `cqobjevo_compiled_coeff_**.pyx` files.
According to the docstring of `solver.Options` and `rhs_generate`, I guess rhs-things relate to string-format time-dependent Hamiltonian, but I couldn't find description about such files. I didn't know even a meaning of 'rhs' (abbreviation?).
So I suggest to add documents about them.

Especially, I have the following question:
- Can I pickle the solver result with a single and portable format?

Saving solver result with pickle module, and loading it sometimes fail, for example:
(I'm so sorry I haven't reproduce this behavior with simplest code, so the following is a dummy-code)
```
out = mesolve(...)
with open(filename, 'wb') as f:
    pickle.dump(out, f)
```
```
Exception ignored in: <function QobjEvo.__del__ at 0x7f9e406f4310>
Traceback (most recent call last):
  File "/home/mizuno.kosuke/venvs/default/lib/python3.8/site-packages/qutip/qobjevo.py", line 624, in __del__
    for file_ in self.coeff_files:
AttributeError: 'QobjEvo' object has no attribute 'coeff_files'
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-22-e4ad332ebfe3> in <module>
      2 p_json = Path(filename)
      3 with p_json.with_suffix('.result.pkl').open('rb') as f:
----> 4     res = pickle.load(f)

ModuleNotFoundError: No module named 'cqobjevo_compiled_coeff_877182617539010'
```
Note: I used pickle module instead of `qsave/qload`. I guess it's no difference because `qsave` uses pickle module internally.

Ericgig:
`rhs_**.pyx` or `cqobjevo_compiled_coeff_**.pyx` are cython compiled function for string-format time-dependent systems.
`rhs_**.pyx` create a function for the right hand side (rhs) of the equation solved by the solver. It's now mostly used in `brmesolve`.

`cqobjevo_compiled_coeff_**.pyx` is used in `QobjEvo` with string coefficient. 

Which version are you using?
`QobjEvo` should not be found in `Results` and `QobjEvo` can be pickle in recent version...

KosukeMizuno:
I'm sorry to forgot to show qutip version.
[...]

PR #1689, opened 2021-10-14
Remove saving the Hamiltonian in sesolve result

Ericgig:
`sesolve` store the Hamiltonian (`-1j*H`) in the result. This Hamiltonian is not used elsewhere, other solve function don't save the system and it's causing some issues (#1684).
This remove it.

**Changelog**
Remove saving the Hamiltonian in sesolve result


hodgestar:
PR #1692 passed the test suite after updating to the latest master, so I'm not sure what is happening. I will try re-run these tests.

PR #1694, opened 2021-10-21
Fix failing steadystate tests

hodgestar:
**Description**

We have steadystate tests that fail almost always in CI on Python 3.9 with OpenMP and MKL, and sometimes with just Python 3.9 and MKL.

Issue is currently hard to reproduce locally.

**Related issues or PRs**
* Test failures seen in #1689 and elsewhere

**Progress so far**
* Fixed a small issue in steadystate tests so that I can use pytest-repeat to run `--count=100` on the steadystate tests in the hopes of reproducing the bug locally.
* Removed mutable default c_ops arguments for steadystate and liouvillian.
* Fix reference to method in _pseudo_inverse_sparse.
* ~~Only set method in pseudo_inverse if one is explicitly defined.~~ (revert)

**Changelog**
TODO: Write the changelog once we understand properly what is going on.


hodgestar:
@Ericgig I've started this branch specifically to tackle the strange steadystate (and other) test failures.

hodgestar:
Minimal script to reproduce the error that only uses numpy:
```
# On a CPU with AVX512 extensions and numpy 1.21.2:
# (only tested on Ubuntu)
# It works again on numpy 1.21.4 (and maybe 1.21.3 -- I did not check because 1.21.3 was not conda installable)

import numpy as np

L = np.diag([1+0j, 1, 1, 1])
b = np.array([1+0j, 0, 0, 0])

# commenting out the line below makes everything work, with it solve returns nans.
np.exp(0)
# breakpoint()

v = np.linalg.solve(L, b)
np.testing.assert_allclose(v, b)
```

hodgestar:
Numpy bug report -- https://github.com/numpy/numpy/issues/20356

hodgestar:
Even smaller script for reproducing the issue:
```python
a = np.diag([1+0j, 1])
np.exp(0)
x = np.linalg.det(a)
```

Ericgig:
@hodgestar do we close this now or do we wait for numpy's fix to be on conda.

hodgestar:
@Ericgig I'm happy to leave this open until a new numpy is released and we can update the version of numpy used in CI tests. Probably also good to have an issue open in case users encounter this in the wild.

hodgestar:
The bug fix is scheduled to be included in numpy 1.22.0 -- https://github.com/numpy/numpy/milestone/93.

PR #1700, opened 2021-10-26
Solver rework

Ericgig:
**Description**
Add a `Solver` class for better task distribution in `sesolve` and `mesolve`.

`sesolve`, `mesolve`, etc. are presently mostly independent: each create their time dependant system, the function to integrator with scipy's ode solver, loop over the `tlist` and store results. With this PR, the `[se/me]solve` functions will reuse most of the code and be more modular.
We already have a `Result` class to manage e_ops and store result, `Integrator` to solve the ode system. `QobjEvo` can support any time-dependant format we support. This PR introduce a new class `Solver` to join all these.

Basis usage looks like:
```
rhs = QobjEvo(...)
solver = Solver(rhs, options=opt)
result = solver.run(psi, tlist, e_ops=e_ops)
result.expect
```

It can also do step evolution:
```
rhs = QobjEvo(...)
solver = Solver(rhs, options=opt)
solver.start(psi, t0)
psi_t1 = solver.step(t1)
psi_t2 = solver.step(t2, args=new_args)  # Can be inefficient depending on integrator.
psi_t3 = solver.step(t3, options=new_opt)  # Inefficient but supported.
```
When stepping, the output is always the state as `Qobj`. The solver class promote ket state to density matrices when the rhs is a super-operator. `rhs` support `Qobj` and `QobjEvo`, but not list or function.

This PR also add `MeSolver` and `SeSolver` classes, but they are very light:
`MeSolver(H, c_ops)` is equivalent to `Solver(liouvillian(H, c_ops))`

`Solver` store a list of available `Integrator` as a class property. `Solver.avail_integrator` has all integrators available to all solvers. `SeSolver.avail_integrator` has all these with those registered directly to `SeSolver`. This is to allow Integrator for specific system. We could think for the floquet sesolve method to be made into an integrator specific to `SeSolver`. 

The interface for `sesolve` and `mesolve` is mostly unchanged, but `progress_bar` keyword is moved to an options and the `_safe_mode` keyword is removed.

**Changelog**
Added `Solver` class.


qutip/solver/solver_base.py
```diff
@@ -0,0 +1,299 @@
...
+        self._t, self._state = self._integrator.integrate(t, copy=False)
+        self.stats["run time"] += time() - _time_start
+        return self._restore_state(self._state, copy=copy)
+
+    def _get_integrator(self):
+        """ Return the initialted integrator. """
+        if self.options.ode["Operator_data_type"]:
+            self.rhs = self.rhs.to(
+                self.options.ode["Operator_data_type"]
+            )
```
hodgestar, line 243: Should be change the `Operator_` to `operator_`.

qutip/tests/solver/test_mesolve.py
```diff
@@ -0,0 +1,657 @@
...
+        nc, na = self.jc_integrate(
+            N, wc, wa, g, kappa, gamma, pump, psi0, use_rwa, tlist)
+
+        nc_ex = 0.5 * (1 - np.cos(2 * g * np.sqrt(n + 1) * tlist)) + n
+        na_ex = 0.5 * (1 + np.cos(2 * g * np.sqrt(n + 1) * tlist))
+
+        np.testing.assert_allclose(nc[-1], nc_ex[-1], atol=0.005)
+        np.testing.assert_allclose(na[-1], na_ex[-1], atol=0.005)
+
+    def testCase2(self):
```
hodgestar, line 473: This test should have a better name. :)

qutip/tests/solver/test_mesolve.py
```diff
@@ -0,0 +1,657 @@
...
+
+        nc_ex = (0.5 * (1 - np.cos(2 * g * np.sqrt(n + 1) * tlist)) + n) * \
+            np.exp(-kappa * tlist)
+        na_ex = 0.5 * (1 + np.cos(2 * g * np.sqrt(n + 1) * tlist)) * \
+            np.exp(-gamma * tlist)
+
+        np.testing.assert_allclose(nc[-1], nc_ex[-1], atol=0.005)
+        np.testing.assert_allclose(na[-1], na_ex[-1], atol=0.005)
+
+    def testCase3(self):
```
hodgestar, line 501: And this one. :)

qutip/tests/solver/test_results.py
```diff
@@ -1,84 +1,88 @@
...
 def test_result_expect():
     N = 10
-    res = Result([qt.num(N), qt.qeye(N)],
-                 qt.solver.SolverResultsOptions(store_final_state=False,
-                                                store_states=False), False)
+    res = Result(
+        [qutip.num(N), qutip.qeye(N)],
+        qutip.solver.SolverResultsOptions(store_final_state=False,
+                                          store_states=False),
+        False, False
```
hodgestar, line 21: What is the new second `False` argument to `Result`?

qutip/tests/solver/test_sesolve.py
```diff
@@ -0,0 +1,251 @@
...
+        sx, sy, sz = solver_obj.run(qutip.basis(2,1), [0, 1, 2, 3],
+                                    e_ops=e_ops).expect
+        csx, csy, csz = solver_obj.run(qutip.basis(2,1), [0, 1, 2, 3],
+                                       e_ops=e_ops).expect
+        np.testing.assert_allclose(sx, csx)
+        np.testing.assert_allclose(sy, csy)
+        np.testing.assert_allclose(sz, csz)
+
+    @pytest.mark.parametrize('method', all_ode_method, ids=all_ode_method)
+    def test_sesolver_steping(self, method):
```
hodgestar, line 197:
```suggestion
    def test_sesolver_stepping(self, method):
```

qutip/tests/solver/test_mesolve.py
```diff
@@ -0,0 +1,657 @@
...
+        copy = pickle.loads(pickle.dumps(solver_obj))
+        e1 = solver_obj.run(qutip.basis(self.N, 9), [0, 1, 2, 3],
+                            e_ops=[self.ada]).expect
+        e2 = solver_obj.run(qutip.basis(self.N, 9), [0, 1, 2, 3],
+                            e_ops=[self.ada]).expect
+        np.testing.assert_allclose(e1, e2)
+
+    @pytest.mark.parametrize('method',
+                             all_ode_method, ids=all_ode_method)
+    def test_mesolver_steping(self, method):
```
hodgestar, line 216:
```suggestion
    def test_mesolver_stepping(self, method):
```

hodgestar (changes requested):
I left a bunch of small suggestions. This is looking great overall though!

qutip/solver/mesolve.py
```diff
@@ -0,0 +1,180 @@
...
+        a *list of array* `result.expect` of expectation values for the times
+        specified by `tlist`, and/or a *list* `result.states` of state vectors
+        or density matrices corresponding to the times in `tlist` [if `e_ops`
+        is an empty list of `store_states=True` in options].
+
+    """
+    H = QobjEvo(H, args=args, tlist=tlist)
+
+    c_ops = c_ops if c_ops is not None else []
+    if not isinstance(c_ops, list):
```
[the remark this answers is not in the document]
Ericgig: I don't think we are accepting tuple in most place we check for a list with `isinstance`...
But I agree there is no reason to refuse tuple.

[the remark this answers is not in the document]
hodgestar: If we're not accepting tuple elsewhere, let me just create a v5 project item for it and we can clean it up later if we feel like it.

qutip/solver/result.py
```diff
@@ -71,41 +73,43 @@ def _read_e_ops(self, _super):
             e_ops = [e for e in self._raw_e_ops.values()]
         elif callable(self._raw_e_ops):
             e_ops = [self._raw_e_ops]
+        elif self._raw_e_ops is None:
```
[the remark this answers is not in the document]
hodgestar: Ah, sorry, you are right. What I did for the BoFiN PR was to always keep the results in the same form as e_ops, rather than converting the results back to a dict at the end.

Since this isn't a user facing change, happy to leave it our for now.

hodgestar (approved):

PR #1702, opened 2021-10-29
qutip.lattice: a helper function for display_unit_cell() and display_lattice()

sbisw002:
1. _unit_site_H() is added that is used by display_unit_cell() and display_lattice()
2. ImportError for matplotlib.pyplot is added.
3. Some logical problems were fixed.

@hodgestar Please review this branch.


hodgestar (changes requested):
Thanks for adding `_unit_cell_H`. Could you add some unit tests for it that cover the different possible cases for `cell_tensor_config`?

hodgestar:
@sbisw002 I'm closing this in favour of #1819 which applies my simpler fix from #1697 and also adds the necessary deprecation warnings for QuTiP 5.

We should continue the development of qutip.lattice in https://github.com/qutip/qutip-lattice. As a first step, we should apply the same simple fix there.

PR #1707, opened 2021-11-06
Hinton plots with working title

VolodyaCO:
**Description**
Hinton plots accept a `title` argument, but the code does not actually set the plot's title. Along with this fix, I also removed unused imports and fixed the other function's titles to avoid verbose such as

```python
if title:
  ax.set_title(title)
```
where the simple `ax.set_title(title)` does nothing when `title` is `None`.

**Related issues or PRs**
No related issues or PRs

**Changelog**
Display a title for hinton and matrix_histogram plots when a title is given.


qutip/visualization.py
```diff
@@ -651,8 +651,7 @@ def matrix_histogram(M, xlabels=None, ylabels=None, title=None, limits=None,
     ax.yaxis._axinfo["grid"]['linewidth'] = 0
     ax.xaxis._axinfo["grid"]['linewidth'] = 0
 
-    if title and fig:
-        ax.set_title(title)
+    ax.set_title(title)
```
hodgestar, line 654: This will unset any existing title, so we shouldn't change the logic here (although the `and fig` check could potentially be removed since that is likely to surprise anyone trying to set a title).

qutip/visualization.py
```diff
@@ -1206,8 +1203,7 @@ def plot_expectation_values(results, ylabels=[], title=None, show_legend=False,
             axes[e_idx, 0].plot(result.times, e,
                                 label="%s [%d]" % (result.solver, e_idx))
 
-    if title:
-        axes[0, 0].set_title(title)
+    fig.suptitle(title)
```
hodgestar, line 1206: What does this change do?

hodgestar (changes requested):
It would be better if this pull request didn't change the logic in other plotting functions. Could we reduce it to just the hinton fix and then we can merge it?

qutip/visualization.py
VolodyaCO: If you have many other axes, the title is not assigned to `axes[0, 0]`, but to the whole figure.

qutip/visualization.py
VolodyaCO: Oh yeah... I will remove the `and fig` check, leaving the `if title`, so that the title is not unset if it previously exists.

qutip/visualization.py
```diff
@@ -268,6 +267,7 @@ def hinton(rho, xlabels=None, ylabels=None, title=None, ax=None, cmap=None,
 
     if not (xlabels or ylabels):
         ax.axis('off')
+    ax.set_title(title)
```
VolodyaCO, line 270:
```suggestion
    if title:
        ax.set_title(title)
```

qutip/visualization.py
VolodyaCO:
```suggestion
    if title:
        ax.set_title(title)
```

qutip/visualization.py
```diff
@@ -768,8 +767,7 @@ def matrix_histogram_complex(M, xlabels=None, ylabels=None,
 
     ax.bar3d(xpos, ypos, zpos, dx, dy, dz, color=colors)
 
-    if title and fig:
-        ax.set_title(title)
+    ax.set_title(title)
```
VolodyaCO, line 770:
```suggestion
    if title:
        ax.set_title(title)
```

qutip/visualization.py
```diff
@@ -967,8 +965,7 @@ def plot_fock_distribution(rho, offset=0, fig=None, ax=None,
     ax.set_xlabel('Fock number', fontsize=12)
     ax.set_ylabel('Occupation probability', fontsize=12)
 
-    if title:
-        ax.set_title(title)
+    ax.set_title(title)
```
VolodyaCO, line 968:
```suggestion
    if title:
        ax.set_title(title)
```

qutip/visualization.py
VolodyaCO:
```suggestion
    if title:
        fig.suptitle(title)
```

hodgestar:
@VolodyaCO I'd like to test these changes out locally, but they look good to me now.

hodgestar (approved):

hodgestar:
@VolodyaCO Fix merged. Thank you for contributing it and apologies that such a simple fix took a long time to be merged.

Issue #1703, opened 2021-11-01
Suggested improvements for average gate fidelity and process fidelity


fhopfmueller:
Hello,

I have some suggestions for improvements in some functions in `metrics.py`. I'd be happy to put together a PR.

The current implementation of `average_gate_fidelity` is based on the Kraus representation of the input quantum channel. If the input is a superoperator, it is first converted to its Kraus representation (involving diagonalization) before computing the average gate fidelity, which sacrifices performance and accuracy. There is a direct formula starting from a superoperator, see, e.g., https://qiskit.org/documentation/stubs/qiskit.quantum_info.average_gate_fidelity.html relating the average gate fidelity to the process fidelity, and https://qiskit.org/documentation/stubs/qiskit.quantum_info.process_fidelity.html#qiskit.quantum_info.process_fidelity to compute that from a superoperator. I think it would be better to use that direct formula if the input is not already in Kraus form!

There is also a function `process_fidelity` in Qutip, but I'm having a hard time understanding what it's intended for. It doesn't seem to be the process fidelity explained in the Qiskit docs above. The arguments of Qutip's `process_fidelity` are called `U1` and `U2` suggesting they are expected to be unitary, but this is not checked. It is computed as `(U1 * U2).tr()`, which is not between 0 and 1, and doesn't give 1 as I'd expect if `U1==U2`. Does someone know the intended use of that function? If not, I'd implement the process fidelity as described in the Qiskit docs here, with several version depending on whether the input is Kraus, a superoperator, and maybe a chi matrix.

Lastly, there is an issue with the tests - the lines
```
class Test_dnorm:
    # Skip dnorm tests if we don't have cvxpy or cvxopt available, since it
    # depends on them.
    cvxpy = pytest.importorskip("cvxpy")
    cvxopt = pytest.importorskip("cvxopt")
```
actually skip all the tests in the file if cvxpy is not available, not just the ones in the `Test_dnorm` class. The test output, if `cvxpy` is not installed, is
```
$ pytest qutip/tests/test_metrics.py 
============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/florian/code/qutip/qutip/qutip/tests, configfile: pytest.ini
plugins: rerunfailures-10.1
collected 0 items / 1 skipped  
```
That issue seems to be addressed on the dev.major branch. Would it make sense to take the test file from there and use it in the master branch also?

Thanks!

[...]

PR #1708, opened 2021-11-08
Implement and test new version of process fidelity

fhopfmueller:
A re-implementation of `qutip.metrics.process_fidelity`. 
The definition implemented here follows Gilchrist, Langford, Nielsen, Phys. Rev. A 71, 062310 (2005), https://journals.aps.org/pra/abstract/10.1103/PhysRevA.71.062310 (https://arxiv.org/abs/quant-ph/0408063).
Depending on how the quantum channels passed to `process_fidelity` are represented (unitary, supermatrix, choi, chi, or Kraus), computational shortcuts are used. They are all tested against each other.

**Related issues or PRs**
Proposed in https://github.com/qutip/qutip/issues/1703

**Changelog**
Fixed implementation of qutip.metrics.process_fidelity and added tests

### DIFF
--- a/qutip/metrics.py
+++ b/qutip/metrics.py
@@ -14,7 +14,9 @@ from scipy import linalg as la
 import scipy.sparse as sp
 from qutip.sparse import sp_eigs
 from qutip.states import ket2dm
-from qutip.superop_reps import to_kraus, to_stinespring, to_choi, _super_to_superpauli, to_super
+from qutip.superop_reps import (to_kraus, to_stinespring, to_choi,
+                                _super_to_superpauli, to_super,
+                                kraus_to_choi)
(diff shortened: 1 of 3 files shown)


Ericgig (commented):
Hi fhopfmueller,
Thank you for the contribution.

Adding support for different representation of super, tests and a proper documentation with a reference is great.

But changing the definition and interface of an existing function could cause issue to those using it. This kind of change would be more appropriate for `dev.major`.

The code is quite hard to follow with a lot of branches and returns. You could use a private function for the case where there are no `target` to split the function into 2 and make it easier to read.

fhopfmueller:
Hi Eric,

Thanks for looking at this and your feedback!

Alright, I'll close this PR and submit a new one for `dev.major`, and will split up the function as you suggest.

PR #1709, opened 2021-11-09
Rtol for tests

AGaliciaMartinez:
**Description**
I added the option to specify `rtol` in the tests. The default value is the same as for `numpy.testing.all_close` which was being used 'silently' (tests for the exponential pretty much ignored the absolute tolerance specified by us). I also made the tests more consistent. Now the assertion for scalar outputs also use `rtol` in the same way `numpy.testing.all_close` does. 

**Related issues or PRs**
This is useful for qutip/qutip-tensorflow#34 where I plan to add support for complex64.

**Changelog**
Aded rtol for specialisation tests.
Changed comparison of scalars to be consistent with `numpy.testing.all_close`

### DIFF
--- a/qutip/tests/core/data/test_mathematics.py
+++ b/qutip/tests/core/data/test_mathematics.py
@@ -297,7 +297,7 @@ class _GenericOpMixin:
         but with all data arguments numpy arrays, and returns the expected
         result.
 
-    tol: float
+    atol: float
         The absolute tolerance to use when comparing the test value with the
         expected value.  If the output is a Data type, the tolerance is
         per-element of the output.
@@ -327,7 +327,7 @@ class _GenericOpMixin:
     # addition not being associative; the maths on full numpy arrays will often
     # produce slightly different results to sparse algebra, since the order of
     # multiplications and additions will be different.
-    tol = 1e-10
+    atol = 1e-10
     rtol = 1e-7  # Same default as numpy
     shapes = []
     bad_shapes = []
@@ -406,16 +406,16 @@ class UnaryOpMixin(_GenericOpMixin):
         if issubclass(out_type, Data):
             assert test.shape == expected.shape
             np.testing.assert_allclose(test.to_array(), expected,
-                                       atol=self.tol,
+                                       atol=self.atol,
                                        rtol=self.rtol)
         elif out_type is list:
             for test_, expected_ in zip(test, expected):
                 assert test_.shape == expected_.shape
                 np.testing.assert_allclose(test_.to_array(),
-                                           expected_, atol=self.tol,
+                                           expected_, atol=self.atol,
                                            rtol=self.rtol)
         else:
-            assert abs(test - expected) < self.tol + self.rtol*abs(expected)
+            assert abs(test - expected) < self.atol + self.rtol*abs(expected)
 
     def test_incorrect_shape_raises(self, op, data_m):
         """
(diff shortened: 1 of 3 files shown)


qutip/tests/core/data/test_mathematics.py
```diff
@@ -400,14 +406,16 @@ def test_mathematically_correct(self, op, data_m, out_type):
...
         elif out_type is list:
             for test_, expected_ in zip(test, expected):
                 assert test_.shape == expected_.shape
                 np.testing.assert_allclose(test_.to_array(),
-                                           expected_, atol=self.tol)
+                                           expected_, atol=self.tol,
+                                           rtol=self.rtol)
         else:
-            assert abs(test - expected) < self.tol
+            assert abs(test - expected) < self.tol + self.rtol*abs(expected)
```
AGaliciaMartinez, line 418: We may want to use `np.testing.assert_allclose` for consistency here.

qutip/tests/core/data/test_mathematics.py
```diff
@@ -323,6 +328,7 @@ def op_numpy(self, *args): raise NotImplementedError
     # produce slightly different results to sparse algebra, since the order of
     # multiplications and additions will be different.
     tol = 1e-10
```
AGaliciaMartinez, line 330: I would like to suggest changing `tol` for `atol` for clarity.

AGaliciaMartinez (commented):
A few other changes that could be included in this PR.

qutip/tests/core/data/test_mathematics.py
Ericgig: For scalar, we could also use `pytest.approx`.

hodgestar: I don't mind too much which we pick, but let's not code this ourselves. I'll put in a small vote for `pytest.approx` (it prints nice error messages) although its behaviour is slightly different (which counts against it a bit in this file).

qutip/tests/core/data/test_norm.py
```diff
@@ -63,9 +63,6 @@ class TestTraceNorm(testing.UnaryOpMixin):
     def op_numpy(self, matrix):
         return np.linalg.svd(matrix)[1].sum()
 
-    # CSR[square,sparse] fails if tol is larger.
-    tol = 3.7e-8
```
hodgestar, line 67: Does this no longer fail if the tolerance is larger?

hodgestar (commented):
I left some comments, but generally it looks like a nice improvement to me.

qutip/tests/core/data/test_norm.py
AGaliciaMartinez: This one was using only the absolute tolerance instead of both the relative and the absolute one (which is use in `numpy.testing.all_close`).  With the addition of relative tolerance it no longer fails.

qutip/tests/core/data/test_mathematics.py
AGaliciaMartinez: I would vote for numpy just for consistency in the behaviour but I am really fine with both.

hodgestar: Sounds good!

qutip/tests/core/data/test_norm.py
hodgestar: Nice!

AGaliciaMartinez:
I implemented the suggestions. I think it should now be ready for merge.

hodgestar (approved):
Thanks @AGaliciaMartinez!

PR #1710, opened 2021-11-12
mcsolve rework

Ericgig:
**Description**
Rewrite `mcsolve` using class as was done for `sesolve` and `mesolve` in #1700.

This add new capacities to `mcsolve`:
- `measurement` and `photocurrent` are added to result making the `photocurrent_..solve` function redundant.
- The `H` input can be a super operator, allowing to do evolution with some collapse operator being treated democratically and some with discrete jumps (as was done in `photocurrent_mesolve`.
- Result are compiled at the end of each trajectories. Previously the result of each trajectories would be stored until the end of the map function, even if only the average are desired, making it quite memory inefficient, see #1667.
- New end condition are added. `mcsolve` can run for a desired number of trajectories, a desired time or the expectation values have converged within a desired tolerance (computed using jackknife re-sampling). 
- Stepping capacities have been added. However, collapse are not available this way, so it's still open to improvement.

Other changes include reworking `MultitrajResults`, it was added in a previous PR but was not used. Also the 2 versions: trajectories stored or averaged are merged.

**Related issues or PRs**
Follow up to #1700 

**Changelog**
Rewrite mcsolve with a class approach.


Babalion:
I like your work very much!
Especially the additional option to stop the simulation when a certain accuracy of the observables is reached!

However, I have the problem that the option is not known with me. 
` mcsolve() got an unexpected keyword argument 'target_tol'`.
Possibly there is just a problem with my merge.

Ericgig:
@Babalion Thank you for tyring.
Please let us know if you meet any bug.

I am keeping both for version of `mcsolve` for now for benchmarks.
The new version has more features but it's slower than the previous.
The new version will replace the main one before we release in v5.

AGaliciaMartinez (commented):
Really nice work Eric!! I really like the new class based approach to solve the evolution of a quantum system. For instance, it makes it much easier to simulate a system step-wise evolution, something I tried with the old method. It also helps extensibility and makes things much cleaner and maintainable. I am not sure how the performance will get affected but I a guessing that as long as the `setp` method is efficient, the rest should not matter that much. 


This review not complete yet but as it was taking me quite some time already I decided to share the suggestions I already had to keep things going. Most of them are related to docs although there are a few more general suggestions I wanted to make:

I  have seen multiple times the `run` method not using the `start` and `step` methods. I found this at first a little bit confusing as from the design of the classes it seems that most of the times it should be enough to rewrite those methods to have a solver working. There is also the fact that a lot of code gets duplicated because of `run` using pretty much the same methods as `start` at the beginning of its logic. This is also true for the `Solver` class and I suggest changing this, if only to avoid the duplication of code and hence help maintainability.

I am also not sure about having two classes for both `_TrayectorySolver` and `MultyTrayectory` solver. I guess the idea  is that new trajectories can be easily created from `_TrayectorySolver`. But the fact that this is a private class and that the methods it needs to be overwrite changes with respect to the `Solver` class confuses me. May I suggest keeping one single class, `McSolver` and `_McTrayectorySolver` for the moment and separate it later once another example of Trajectory is included? 

Finally, I am not sure about the approach to parallelize the McSolver. It only parallelizes the `run` method, yet from the class design and its similarity with the rest of the solvers, I would expect the run method to be roughly equivalent to:
```python
solver.start()
for t in t_list:
    t, state = solver.step(t, args)
    compute_expect(state, e_ops)
```
I suggested a different approach although this may still need some thinking. 

I hope these suggestions are useful!

qutip/solver/multitraj.py
```diff
@@ -0,0 +1,464 @@
...
+            seeds = [
+                seed_  if isinstance(seed_, np.random.SeedSequence)
+                else np.random.SeedSequence(seed_)
+                for seed_ in seed[:ntraj]
+            ]
+        else:
+            raise ValueError("A seed list must be longer than ntraj")
+        return seeds
+
+    def start(self, state0, t0, ntraj=1, seed=None, *, safe_ODE=None):
```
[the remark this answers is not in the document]
Ericgig: `mesolve` uses `rho0`, `sesolve` and `mcsolve` use `psi0`.
But kets are accepted where most density matrix are, and operator input are accepted.
I have not passed to unify the documentation, but it will be needed.
I would go for simply `state` everywhere in this PR, with the plan to take another look at all solver's documentation later.

qutip/solver/multitraj.py
```diff
@@ -0,0 +1,464 @@
...
+
+    options : SolverOptions
+        Options for the solver
+    """
+    _traj_solver_class = None
+    name = "generic multi trajectory"
+    optionsclass = SolverOptions
+    odeoptionsclass = SolverOdeOptions
+
+    def __init__(self, rhs, c_ops=(), *, options=None):
```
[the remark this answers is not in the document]
Ericgig: I kind of want to do this everywhere.
`None` need to be converted to an empty list since we need to loop over them etc., but we can't default to a mutable `list`.
But an empty tuple does not need can be used to do the same and is safe.
I don't think that any user uses explicitly `c_ops=None`.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,318 @@
...
+        return mesolve(H, psi0, tlist, e_ops=e_ops, args=args, options=options)
+
+    if isinstance(ntraj, list):
+        if isinstance(options, dict):
+            options = SolverOptions(**options)
+        options = copy(options) or SolverOptions()
+        options.results['keep_runs_results'] = True
+        max_ntraj = max(ntraj)
+    else:
+        max_ntraj = ntraj
```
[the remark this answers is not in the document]
AGaliciaMartinez: I see. I am ok keeping it as long as it is properly documented. I guess we want `mcsolve` to remain as similar as the previous one as possible. I did not look carefully into MultyTrajResult yet but will do next time before I take another look at the review.

qutip/solver/multitraj.py
```diff
@@ -0,0 +1,464 @@
...
+        if options is not None:
+            self.options = options
+        if args is not None:
+            self._argument(args)
+
+        self.result = MultiTrajResult(ntraj, e_ops or [], self._c_ops,
+                                      options=self.options.results,
+                                      target_tol=target_tol)
+
+        self._run_args = state0, tlist, e_ops
```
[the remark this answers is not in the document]
Ericgig: No with `add_trajectories` should not take this as input or it breaks it's purpose.
Say you computed 100 trajectories with run and there is still to much noise, you can add other trajectories to the same result with `add_trajectories`. But these new trajectories cannot have different starting states or the expectation operator or the average and other results you computed will be wrong.

Another way to do it with storing it here could be:
```
res = mc.run(state, tlist, 100)
mc.add_trajectories(res, 100)
```
and storing them in the result object. It already store `tlist` and `e_ops` and it's probably easier to see where the new trajectories go.

qutip/solver/multitraj.py
```diff
@@ -0,0 +1,464 @@
...
+
+        t0 : double
+            Initial time of the evolution.
+
+        seed : int, SeedSequence
+            Seed for the random number generator.
+
+        safe_ODE : bool {False}
+            Whether to save the state locally. Set to ``True`` if using
+            multiple solvers for step evolution at once.
```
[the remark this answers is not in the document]
Ericgig: `step` cannot be used in parallel (yet?) 
When in parallel: probably not.
Right now, `False` if one trajectory is ran, `True` if multiple. However I would like to discuss what step means for multi trajectory solvers, I am not sure if we should support it.

qutip/solver/multitraj.py
```diff
@@ -0,0 +1,464 @@
...
+            self._argument(args)
+
+        self.result = MultiTrajResult(ntraj, e_ops or [], self._c_ops,
+                                      options=self.options.results,
+                                      target_tol=target_tol)
+
+        self._run_args = state0, tlist, e_ops
+        self.result.stats['run time'] = 0
+        self._traj_solver = self._traj_solver_class(self, options=self.options)
+        self._traj_solver._integrator = self._traj_solver._get_integrator()
```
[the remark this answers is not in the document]
Ericgig: Yes and no... There are Integrator that precompute part of there computation that will win big by doing it here.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,318 @@
...
+    options : SolverOptions, [optional]
+        Options for the evolution.
+
+    seeds : int, SeedSequence, list, [optional]
+        Seed for the random number generator. It can be a single seed used to
+        spawn seeds for each trajectories or a list of seed, one for each
+        trajectories. Seed are saved in the result, they can be reused with::
+            seeds=prev_result.seeds
+
+    target_tol : float, list, [optional]
```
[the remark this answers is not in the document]
Ericgig: jackknife compute an error bar on the average as does bootstrap, but is easier and faster to do on the run and can give an estimation on the number of trajectories needed to reach the desired tolerance.
By 'the error is computer using jackknife resampling' I mean the 1 standard derivation error bar on the average of the expectation values.

qutip/solver/mcsolve.py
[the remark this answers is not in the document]
Ericgig: Removed with an error message pointing to the new way of getting these result.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,318 @@
...
+    def options(self, new):
+        if new is None:
+            new = self.optionsclass()
+        elif isinstance(new, dict):
+            new = self.optionsclass(**new)
+        elif not isinstance(new, self.optionsclass):
+            raise TypeError("options must be an instance of" +
+                            str(self.optionsclass))
+        new.results['normalize_output'] = False
+        self._options = new
```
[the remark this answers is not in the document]
Ericgig: It's not that simple, for super with property setter it's `super(McSolver, self.__class__).options.fset(self, new)`.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,318 @@
...
+        Initial state vector
+
+    tlist : array_like
+        Times at which results are recorded.
+
+    ntraj : int
+        Maximum number of trajectories to run. Can be cut short if a time limit
+        is passed in options (per default, mcsolve will stop after 1e8 sec)::
+            ``options.mcsolve['map_options']['timeout'] = max_sec``
+        Or if the target tolerance is reached, see ``target_tol``.
```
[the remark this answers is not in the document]
AGaliciaMartinez: This should be resolved in the latest update.

qutip/solver/result.py
```diff
@@ -187,101 +182,241 @@ def __repr__(self):
...
-
-    Property
-    --------
-
-    runs_states : list of list of Qobj
-        Every state of the evolution for each trajectories. (ket)
-
-    average_states : list of Qobj
-        Average state for each time. (density matrix)
+    Contain result of simulations with multiple trajectories.
```
AGaliciaMartinez, line 185:
```suggestion
    It contains the results of simulations with multiple trajectories.
```

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,318 @@
...
+        if H.issuper:
+            self._c_ops = [
+                spre(c_op) * spost(c_op.dag()) if c_op.isoper else c_op
+                for c_op in c_ops
+            ]
+            self._n_ops = self._c_ops
+            rhs = QobjEvo(H)
+            for c_op in c_ops:
+                cdc = c_op.dag() @ c_op
+                rhs -= 0.5 * (spre(cdc) + spost(cdc))
```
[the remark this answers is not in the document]
AGaliciaMartinez: I see now how this is obtained. Lets mark this as resolved.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+            self._n_ops = [c_op.dag() * c_op for c_op in c_ops]
+            rhs = -1j * QobjEvo(H)
+            for n_op in self._n_ops:
+                rhs -= 0.5 * n_op
+
+        super().__init__(rhs, options=options)
+
+        self.stats['solver'] = "MonteCarlo Evolution"
+        self.stats['num_collapse'] = len(c_ops)
+        self.stats["preparation time"] = time() - _time_start
```
AGaliciaMartinez, line 285: preparation time in mesolve is used to measure the start method and not the init. I find this a little bit inconsistent. Did you perhaps change it because it no longer made sense to be defined in the start method?

qutip/solver/multitraj.py
```diff
@@ -0,0 +1,494 @@
...
+                            str(self.optionsclass))
+        self._options = new
+
+    def _argument(self, args):
+        """Update the args, for the `rhs` and `c_ops` and other operators."""
+        self.rhs.arguments(args)
+
+    @property
+    def traj_args(self):
+        return (self.rhs,)
```
AGaliciaMartinez, line 311: Should this not be overwritten by McSolver to include c_ops?

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+        self.generator = self.get_generator(seed)
+        self.target_norm = self.generator.random()
+        result = super()._run(self.generator, state, tlist, e_ops)
+        result.collapse = list(self.collapses)
+        result.seed = seed
+        return result
+
+    def start(self, state, t0, seed=None, *, safe_ODE=None):
+        self.collapses = []
+        self.generator = self.get_generator(seed)
```
AGaliciaMartinez, line 126: I think this is also done in super.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+    _avail_integrators = {}
+
+    def __init__(self, rhs, c_ops, n_ops, *, options=None, _prepare=False):
+        self._c_ops = c_ops
+        self._n_ops = n_ops
+        super().__init__(rhs, options=options, _prepare=_prepare)
+
+    def _run(self, seed, state, tlist, e_ops):
+        self.collapses = []
+        self.generator = self.get_generator(seed)
```
AGaliciaMartinez, line 117: I think this is also done in super. I prefer to leave the handling of the seed to super such that the resposibilities of the classes are well separated.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def prob_func(self, state):
+        if self.rhs.issuper:
+            return _data.norm.trace(unstack_columns(state))
+        return _data.norm.l2(state)**2
+
+    def norm_func(self, state):
+        if self.rhs.issuper:
+            return _data.norm.trace(unstack_columns(state))
+        return _data.norm.l2(state)
```
AGaliciaMartinez, line 138: If these functions can potentially be overwritten we should both write a docstring and make state be a Qobj instead of a data instance. If this is not meant to be overwritten lets not expose and make it a private functions instead.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def _step(self, t, copy=True):
+        t_old, y_old = self._integrator.get_state(copy=False)
+        norm_old = self.prob_func(y_old)
```
AGaliciaMartinez, line 142: Not sure how inpactful computing this each time could be but I am going to suggest keepint track of norm_old as an attribute to compute it only once.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+        if self.rhs.issuper:
+            return _data.norm.trace(unstack_columns(state))
+        return _data.norm.l2(state)**2
+
+    def norm_func(self, state):
+        if self.rhs.issuper:
+            return _data.norm.trace(unstack_columns(state))
+        return _data.norm.l2(state)
+
+    def _step(self, t, copy=True):
```
AGaliciaMartinez, line 140: The copy seems to be called for safe_ODE. It seems suspicious though to see no use of it in this function. all copy statements are set to False. I am not sure how this affect multiprocessing though.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+                t_old, y_old = t_step, state
+                norm_old = norm
+
+        return t_old, _data.mul(y_old, 1 / self.norm_func(y_old))
+
+    def _do_collapse(self, norm_old, norm, t_prev):
+        """Find and apply a collapse."""
+        t_final, _ = self._integrator.get_state()
+        tries = 0
+        while tries < self.options.mcsolve['norm_steps']:
```
AGaliciaMartinez, line 160: I think these options need a little bit better documentation in options.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+                # t_guess is still > t_jump
+                t_final = t_guess
+                norm = norm2_guess
+            else:
+                # t_guess < t_jump
+                t_prev = t_guess
+                norm_old = norm2_guess
+
+        if tries >= self.options.mcsolve['norm_steps']:
+            raise Exception("Norm tolerance not reached. " +
```
AGaliciaMartinez, line 191: Mabe a more explicit exception could be good here.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+
+        return t_old, _data.mul(y_old, 1 / self.norm_func(y_old))
+
+    def _do_collapse(self, norm_old, norm, t_prev):
+        """Find and apply a collapse."""
+        t_final, _ = self._integrator.get_state()
+        tries = 0
+        while tries < self.options.mcsolve['norm_steps']:
+            tries += 1
+            if (t_final - t_prev) < self.options.mcsolve['norm_t_tol']:
```
AGaliciaMartinez, line 162: We should be very clear on the documentation that this is an absolute tolerance.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+            raise Exception("Norm tolerance not reached. " +
+                            "Increase accuracy of ODE solver or " +
+                            "SolverOptions.mcsolve['norm_steps'].")
+
+        # t_guess, state is at the collapse
+        probs = np.zeros(len(self._n_ops))
+        for i, n_op in enumerate(self._n_ops):
+            probs[i] = n_op.expect_data(t_guess, state).real
+        probs = np.cumsum(probs)
+        which = np.searchsorted(probs, probs[-1] * self.generator.random())
```
AGaliciaMartinez, line 200: A minor thing to consider: the current documentation for [montecarlo](https://qutip.org/docs/latest/guide/dynamics/dynamics-monte.html) says that the chosen collapse operator is based on "r" which I think here would be target_tol and not a new random number. Not sure if this has really a significant effect on the result (I think it does not) but I would suggest to use tartget_norm as the documentation says. What was using the previous method?

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+
+        state_new = self._c_ops[which].matmul_data(t_guess, state)
+        new_norm = self.norm_func(state_new)
+        if new_norm < self.options.mcsolve['mc_corr_eps']:
+            # This happen when the collapse is caused by numerical error
+            state_new = _data.mul(state, 1 / self.norm_func(state))
```
AGaliciaMartinez, line 206: I am not sure how this occurs. I would like to discuss it with you in discord if possible.

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+        return _data.norm.l2(state)
+
+    def _step(self, t, copy=True):
+        t_old, y_old = self._integrator.get_state(copy=False)
+        norm_old = self.prob_func(y_old)
+        while t_old < t:
+            t_step, state = self._integrator.mcstep(t, copy=False)
+            norm = self.prob_func(state)
+            if norm <= self.target_norm:
+                self._do_collapse(norm_old, norm, t_old)
```
AGaliciaMartinez, line 147: May I suggest splitting this in two? one that finds the collapse (`_find_collapse_time`) time and another that applies the collapse . These operations by themselves are relatively complex so I think it will help in the readability/maintanability. Also, I suggest using as argument names for `_find_collapse_time(norm_t_old, t_old, norm_t, t)` so that it is clear that `norm_t_old` refers to the norm at `t_old` whereas `norm_t` refers to the norm at time t. At this moment t is not a parameter and although it can be obtain from integrator I would say it is better to pass it in order to be explicit on what the function does (I believe this will help writing a clear docstring).

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+        state_new = self._c_ops[which].matmul_data(t_guess, state)
+        new_norm = self.norm_func(state_new)
+        if new_norm < self.options.mcsolve['mc_corr_eps']:
+            # This happen when the collapse is caused by numerical error
+            state_new = _data.mul(state, 1 / self.norm_func(state))
+        else:
+            state_new = _data.mul(state_new, 1 / new_norm)
+            self.collapses.append((t_guess, which))
+            self.target_norm = self.generator.random()
+        self._integrator.set_state(t_guess, state_new)
```
AGaliciaMartinez, line 211: I thought setting the state of an integrator in this way could not be done for some integrators (adams or bdf)? Does it not have an drawback (such as having to reset those integrators)?

qutip/solver/mcsolve.py
```diff
@@ -0,0 +1,311 @@
...
+                   * np.log(norm_old / self.target_norm)
+                   / np.log(norm_old / norm))
+            )
+            if (t_guess - t_prev) < self.options.mcsolve['norm_t_tol']:
+                t_guess = t_prev + self.options.mcsolve['norm_t_tol']
+            _, state = self._integrator.mcstep(t_guess, copy=False)
+            norm2_guess = self.prob_func(state)
+            if (
+                np.abs(self.target_norm - norm2_guess) <
+                self.options.mcsolve['norm_tol'] * self.target_norm
```
AGaliciaMartinez, line 178: I see that `norm_tol` refers to relative, unlike `norm_t_tol`. we need to be clear on this in the documentation. I would suggest a slight renaming to `norm_rtol` and `norm_t_atol` so that it is already clear from the name what kin of tolerances each refer to.

qutip/solver/mcsolve.py
Ericgig: The returned state is always a copy because of the re-normalization at the end. In stochastic, it could be the array stored in the solver if `copy=False`.
Without copy here, either it always makes a copy, where stochastic fill suffer, or the function calling this will need to do one, then copies of copies will be made here.

qutip/solver/mcsolve.py
Ericgig: It has very little impact in my tests, but would be easy to cache.

qutip/solver/mcsolve.py
Ericgig: The atol vs rtol does not matter enough to change the names but documentation not mentioning that it's used when finding collapse is certainly lacking.

qutip/solver/mcsolve.py
Ericgig: The documentation is wrong, it needs to be another random number. It was fixed in  #1677.

Ericgig (commented):
Only had time to read half of them...

qutip/solver/mcsolve.py
AGaliciaMartinez: If it has little impact lets just leave it for the moment.

qutip/solver/mcsolve.py
Ericgig: No, it measure both in `mesolve`. `__init__` and `start` and the preparation part of `run` are all included into "preparation time".

qutip/solver/multitraj.py
Ericgig: It is.

AGaliciaMartinez (commented):
A few more suggestions. The class MultiTrajResult is the only one left to review. I also need to take a look at the tests.

Ericgig:
While you could probably make it work but manually updating the state at each times of `tlist` is certainly not a usage I planned for... If you could use `mesolve` instead of `mcsolve` it would be simpler:
```
solver = MeSolver(...)
solver.start(rho0, t0)
for t in tlist[1:]:
    rho_t = solver.step(t)
    solver.start(rho_t @ projection, t)
```

With `mcsolve`, if you use it like this, it would normalize the state at each step possibly breaking the montecarlo part of the evolution. The way I would do it is by making a child class overwriting the step of `McTrajectorySolver`:

```
class MyMcTrajectorySolver(McTrajectorySolver):
    def _step(self, t):
        super()._step(t)
        _, raw_state = self._integrator.get_state() # state not normalized
        new_state = _data.matmul(projector, state) # projector is a Data object: Qobj.data
        self._integrator.set_state(t, new_state)
        return _data.mul(new_state, 1 / _data.norm.l2(new_state))

class MyMcSolver(McSolver):
    _traj_solver_class = MyMcTrajectorySolver
```
If done this way, you should be able to use the normal solver interface. This projection should be done once per call of `step` or per time in `tlist`, but I would confirm it before running simulation.

`step` does not return the `collapse` part of the results. So if a collapse happen, it will not output it nor the time at which it happen. But the state evolution is done normally, collapse will happen as expected. To get them, you will need to fetch them with `solver.traj_solvers[i].collapse`.

Feel free to comment on the new solver interface. If you think that something is missing or confusing, now is a good time to tell us. We should come back on this PR in April.

Ericgig:
Replaced by #1972

PR #1712, opened 2021-11-14
Implement and test process fidelity

fhopfmueller:
**Description**
This PR implements and tests a new version of `process_fidelity`.
The new version follows the definition in Gilchrist et al., Physical Review A 71, 062310 (2005).

**Related issues or PRs**
Discussion at https://github.com/qutip/qutip/issues/1703.
I previously submitted a similar PR at https://github.com/qutip/qutip/pull/1708, targeting `master`. 
This PR targets `dev.major` rather than `master` because existing functionality is changed, and contains improvements following @Ericgig 's suggestions. I have also adapted the tests to match the style of the tests in `dev.major`.

Addresses part 2 of https://github.com/qutip/qutip/issues/1703.

**Changelog**
Implement and test a new version of `process_fidelity` (https://github.com/qutip/qutip/issues/1703)

### DIFF
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -173,15 +173,18 @@ def process_fidelity(oper, target=None):
 
     Notes
     -----
-    See, for example: A. Gilchrist, N.K. Langford, M.A. Nielsen,
-    Phys. Rev. A 71, 062310 (2005).
+    Since Qutip 5.0, this function computes the process fidelity as defined
+    for example in: A. Gilchrist, N.K. Langford, M.A. Nielsen,
+    Phys. Rev. A 71, 062310 (2005). Previously, it computed a function
+    that is now implemented in
+    :func:`control.fidcomp.FidCompUnitary.get_fidelity`.
     The definition of state fidelity that the process fidelity is based on
     is the one from R. Jozsa, Journal of Modern Optics, 41:12, 2315 (1994).
     It is the square of the one implemented in
     :func:`qutip.metrics.fidelity` which follows Nielsen & Chuang,
     "Quantum Computation and Quantum Information"
+
     """
-    warnings.warn("Caution: Behavior of process_fidelity changed in Qutip 5.0")
     if target is None:
         return _process_fidelity_to_id(oper)
     elif not isinstance(target, list) and target.type == 'oper':
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -40,6 +40,8 @@ __all__ = ['fidelity', 'tracedist', 'bures_dist', 'bures_angle',
            'hellinger_dist', 'hilbert_dist', 'average_gate_fidelity',
            'process_fidelity', 'unitarity', 'dnorm']
 
+import warnings
+
 import numpy as np
 from scipy import linalg as la
 import scipy.sparse as sp
@@ -165,8 +167,7 @@ def process_fidelity(oper, target=None):
     Returns
     -------
     fid : float
-        Process fidelity between oper and target,
-        or between oper and identity.
+        Process fidelity between oper and target, or between oper and identity.
     Notes
     -----
     See, for example: A. Gilchrist, N.K. Langford, M.A. Nielsen,
@@ -177,6 +178,7 @@ def process_fidelity(oper, target=None):
     :func:`qutip.metrics.fidelity` which follows Nielsen & Chuang,
     "Quantum Computation and Quantum Information"
     """
+    warnings.warn("Caution: Behavior of process_fidelity changed in Qutip 5.0")
     if target is None:
         return _process_fidelity_to_id(oper)
     elif not isinstance(target, list) and target.type == 'oper':
@@ -203,7 +205,7 @@ def process_fidelity(oper, target=None):
         if oper_choi.dims != target_choi.dims:
             raise TypeError('Dimensions of oper and target do not match')
         d = np.prod(oper_choi.dims[0][0])
-        return fidelity(oper_choi / d, target_choi / d)**2
+        return (fidelity(oper_choi, target_choi )/d)**2
 
 
 def average_gate_fidelity(oper, target=None):
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -43,7 +43,8 @@ __all__ = ['fidelity', 'tracedist', 'bures_dist', 'bures_angle',
 import numpy as np
 from scipy import linalg as la
 import scipy.sparse as sp
-from .superop_reps import to_kraus, to_choi, _to_superpauli, to_super
+from .superop_reps import (to_kraus, to_choi, _to_superpauli, to_super,
+                           kraus_to_choi)
 from .superoperator import operator_to_vector, vector_to_operator
 from .operators import qeye
 from .semidefinite import dnorm_problem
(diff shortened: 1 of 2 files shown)


Ericgig (changes requested):
Looks good.
It's great to have a proper documentation, tests, and a properly defined function.
`_process_fidelity_to_id` makes it easier to read.

Since the definition of the function changed, could you add a warning. We need to be clear that when going from v4 to v5, the result will change and we can't expect everybody will read the documentation.

Ericgig (approved):
Seems good to me.
Thank you fhopfmueller.

qutip/core/metrics.py
```diff
@@ -105,12 +108,104 @@ def fidelity(A, B):
...
+    -------
+    fid : float
     """
-    Calculate the process fidelity given two process operators.
+    msg = 'The process fidelity to identity is only defined for ' \
+          'dimension preserving channels.'
+    if isinstance(oper, list):  # oper is a list of Kraus operators
+        d = oper[0].shape[0]
+        if oper[0].shape[1] != d:
+            raise TypeError(msg)
```
hodgestar, line 129:
```suggestion
            raise ValueError(msg)
```
I suggest we change these to `ValueError`. The formal type (i.e. list or Qobj) is correct, but the value is not correct (i.e. channel has the wrong shape).

qutip/core/metrics.py
```diff
@@ -105,12 +108,104 @@ def fidelity(A, B):
...
+
+def process_fidelity(oper, target=None):
     """
-    out = (U1 * U2).tr()
-    return out / (U1.tr() * U2.tr()) if normalize else out
+    Returns the process fidelity of a quantum channel to the target
+    channel, or to the identity channel if no target is given.
+    The process fidelity between two channels is defined as the state
+    fidelity between their normalized Choi matrices.
+    Parameters
```
hodgestar, line 159:
```suggestion

    Parameters
```
Just adding some paragraph spacing.

qutip/core/metrics.py
```diff
@@ -105,12 +108,104 @@ def fidelity(A, B):
...
+    ----------
+    oper : :class:`qutip.Qobj`/list
+        A unitary operator, or a superoperator in supermatrix, Choi or
+        chi-matrix form, or a list of Kraus operators
+    target : :class:`qutip.Qobj`/list
+        A unitary operator, or a superoperator in supermatrix, Choi or
+        chi-matrix form, or a list of Kraus operators
+    Returns
```
hodgestar, line 167:
```suggestion

    Returns
```

qutip/core/metrics.py
```diff
@@ -105,12 +108,104 @@ def fidelity(A, B):
...
+    -------
+    fid : float
+        Process fidelity between oper and target, or between oper and identity.
+    Notes
```
hodgestar, line 171:
```suggestion

    Notes
```

qutip/core/metrics.py
```diff
@@ -105,12 +108,104 @@ def fidelity(A, B):
...
+    -----
+    See, for example: A. Gilchrist, N.K. Langford, M.A. Nielsen,
+    Phys. Rev. A 71, 062310 (2005).
+    The definition of state fidelity that the process fidelity is based on
+    is the one from R. Jozsa, Journal of Modern Optics, 41:12, 2315 (1994).
+    It is the square of the one implemented in
+    :func:`qutip.metrics.fidelity` which follows Nielsen & Chuang,
+    "Quantum Computation and Quantum Information"
+    """
+    warnings.warn("Caution: Behavior of process_fidelity changed in Qutip 5.0")
```
hodgestar, line 181: My vote would be to remove this warning and instead write a small paragraph in the docstring describing what changed between 4.7 and 5.0.

qutip/core/metrics.py
hodgestar: The same comment applies to the other `TypeError` exceptions too.

Also happy for @Ericgig to chime in with whether he agrees with my suggestion or not and to leave it as is if he doesn't.

hodgestar (commented):
This looks great. Thank you!

qutip/core/metrics.py
Ericgig: This is a new function that replace an old one with only the name and signature in common. 

I don't expect any users to read the documentation of every function they use before trying their code in v5. Most will just try their old code, get different results and have hard time finding where it come from. Or worst, don't realize that they got a different result until much later...
So we should have a loud message that this changed, something that will make sure that any users using it will know to expect a new, unrelated result for that function call from now on.

Another option would be to make this a new function and deprecate the other one. This will remove any risk of confusion.

qutip/core/metrics.py
fhopfmueller: In the rest of this file, metrics raise a `TypeError` when the shape doesn't match, for example in `fidelity` and `trace_dist`. So I'd vote to keep `TypeError` here for consistency.

qutip/core/metrics.py
fhopfmueller: My guess would be that not many people used the old `process_fidelity`, I don't think it was correct. 
`average_gate_fidelity` is closely related and probably more common.

As a next step, I'm planning to send a PR for `average_gate_fidelity`, reducing it to `process_fidelity`. (The current implementation of `average_gate_fidelity` is correct, but not efficient for some `oper.superrep`s.) That change wouldn't be user facing - so if we leave the warning, users of `average_gate_fidelity` would get a warning from `process_fidelity` even though nothing has changed (unless we introduce another private "warning-free" version of process fidelity but that seems complicated).

For those two reasons, I'd also suggest to remove the warning, but it's up to you.

qutip/core/metrics.py
hodgestar: Happy with that -- we can always fix these TypeErrors in a follow up PR if we decide to.

qutip/core/metrics.py
hodgestar: If it was completely wrong before, then we definitely don't need the warning -- and we've fixed such "this function does completely the wrong thing" in bugfix releases before without a warning.

If we do keep the warning, it should definitely say something more helpful than "something changed" which immediately causes the reader to wonder "what changed?".

@Ericgig Thoughts?

Ericgig: I looked at when it was created , #130, the commit messages in the same PR mention GRAPE, so it seems to be a definition used in quantum control. `fidcomp.py` seems to have a similar function with `U1` being `_onto_evo` and `U2` being `_fwd_evo`. So it never mean to a fidelity between two operators but a fidelity contribution of one time slice of an iteration of GRAPE...

Since it was clearly not a mistake, it's not a sign error but a totally different formulae, I expected this was the right behaviour for someone's research and wanted to be safe.  But now that it seems to be part of an old implementation of quantum control that as been removed since. I don't see any issue with replacing it with only a note in the changelogs.

hodgestar: @Ericgig Thank you for digging into the history!

@fhopfmueller If you will make the last change to remove the warning and add a note we can merge this, I think.

fhopfmueller: Thanks a lot @Ericgig , that makes a lot of sense now!

hodgestar:
@fhopfmueller Thank you for making this PR happen! Merging.

PR #1715, opened 2021-11-18
Added packaging to the installation.rst requirements.

AGaliciaMartinez:
**Changelog**
Added packaging to the requirements in `installation.rst`.
Update requirements.txt with the information in setup.cfg.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 cython>=0.29.20
-numpy>=1.16.6,<1.20
+numpy>=1.16.6
 scipy>=1.0
 packaging
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-cython>=0.21
-numpy>=1.12
+cython>=0.29.20
+numpy>=1.16.6,<1.20
 scipy>=1.0
 packaging
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -171,7 +171,7 @@ As of the 4.6.0 release, the build requirements can be installed with
 
 .. code-block:: bash
 
-   pip install setuptools wheel 'cython>=0.29.20' 'numpy>=1.16.6,<1.20' 'scipy>=1.0'
+   pip install setuptools wheel packaging 'cython>=0.29.20' 'numpy>=1.16.6,<1.20' 'scipy>=1.0'
 
 or similar with ``conda`` if you prefer.
 You will also need to have a functional C++ compiler installed on your system.


AGaliciaMartinez:
I realised it was not necessary for me to include setuptools and wheel in the installation. I wonder if this is because setuptools is being taken from the system installation instead of the virtualenv? 

Also, I noticed that requirements.txt does not include stuptools nor wheel. Should I include them?

jakelishman:
`requirements.txt` has no special meaning at all so it doesn't really matter, but typically it's used for runtime, not build time, so it doesn't need `setuptools` or `wheel`.  `packaging` is required at both, but I had actually thought it was a dependency of `setuptools`.  Still, we use it explicitly, so best add it.

The actual main source-of-truth runtime requirements are in `setup.cfg`, and the build requirements are in `pyproject.toml`.

AGaliciaMartinez:
Thank you Jake for taking a quick look!
I checked once more and the following produces an error (used in an "empty" virtual environment):
```
> pip install setuptools wheel 'cython>=0.29.20' 'numpy>=1.16.6,<1.20' 'scipy>=1.0'
> python setup.py develop

  File "/home/kaladin/Documents/git/qutip/qutip4/setup.py", line 13, in <module>
    import packaging.version
ModuleNotFoundError: No module named 'packaging'
```
So I think it is not included with `setuptools`. 

> `requirements.txt` has no special meaning at all so it doesn't really matter,
Lets not add wheels and setuptools to the requirements then. Updating it with the latest dependency versions in setup.cfg dependencies should be fine though.

AGaliciaMartinez:
The test for python 3.6 seems to fail at coveralls, not sure why though. It may not be related as the rest of the tests do not fail. Could it be rerun?

jakelishman:
Sounds good to me.  Yeah, even if `packaging` _were_ in `setuptools`, it'd still be an explicit dependency of ours, so you're right to put it in, thanks.

Yeah, that coveralls thing just looked like a network timeout on their side.  Not the fault of this PR.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-cython>=0.21
-numpy>=1.12
+cython>=0.29.20
+numpy>=1.16.6,<1.20
```
jakelishman, line 2: There's no upper limit on Numpy at runtime, unless something's changed recently.

AGaliciaMartinez: Indeed the runtime does not have upper limit, only the building requirements do. Ive just updated `requirements.txt`.

hodgestar (approved):

PR #1723, opened 2021-11-23
Added Test classes to the be importable from tests/core/data

AGaliciaMartinez:
**Description**

Previously I was using (in qutip-tensorflow):
```python
import qutip.tests.core.data.test_mathematics
import qutip.tests.core.data.test_reshape

class TestAdd_2(test_mathematic.TestAdd):
	pass

class TestReshape_2(test_reshape.TestReshape):
	pass
```

This PR allows using instead:
```python
import qutip.tests.core.data as testing

class TestAdd_2(testing.TestAdd):
	pass

class TestReshape_2(testing.TestReshape):
	pass
```
which is more convenient, specially because there are multiple  modules that need to be imported in this way. Importing the classes individually is not a valid option as it will run qutip's tests also in qutip-tensorflow. 

**Related issues or PRs**
This is currently planned to be used in qutip/qutip-tensorflow#34.

### DIFF
--- a/qutip/tests/core/data/__init__.py
+++ b/qutip/tests/core/data/__init__.py
@@ -1,0 +1,4 @@
+from .test_expect import *
+from .test_mathematics import *
+from .test_norm import *
+from .test_reshape import *
--- a/qutip/tests/core/data/test_expect.py
+++ b/qutip/tests/core/data/test_expect.py
@@ -10,6 +10,7 @@ from itertools import product
 
 __all__ = ['TestExpect', 'TestExpectSuper']
 
+
 class TestExpect(BinaryOpMixin):
     def op_numpy(self, op, state):
         is_ket = state.shape[1] == 1
--- a/qutip/tests/core/data/test_expect.py
+++ b/qutip/tests/core/data/test_expect.py
@@ -8,6 +8,7 @@ from qutip import data
 from qutip.core.data import CSR, Dense
 from itertools import product
 
+__all__ = ['TestExpect', 'TestExpectSuper']
 
 class TestExpect(BinaryOpMixin):
     def op_numpy(self, op, state):
--- a/qutip/tests/core/data/test_mathematics.py
+++ b/qutip/tests/core/data/test_mathematics.py
@@ -8,10 +8,12 @@ from qutip.core.data import Data, Dense, CSR
 
 from . import conftest
 
-__all__ = ['TestAdd', 'TestAdjoint', 'TestConj', 'TestExpm', 'TestInner',
-          'TestTranspose', 'TestInnerOp', 'TestInv', 'TestKron', 'TestMatmul',
-          'TestMul', 'TestTrace', 'TestSub', 'TestProject', 'TestPow',
-          'TestNeg', 'TestMultiply', 'TestMultiply']
+__all__ = [
+    'TestAdd', 'TestAdjoint', 'TestConj', 'TestExpm', 'TestInner',
+    'TestTranspose', 'TestInnerOp', 'TestInv', 'TestKron', 'TestMatmul',
+    'TestMul', 'TestTrace', 'TestSub', 'TestProject', 'TestPow', 'TestNeg',
+    'TestMultiply', 'TestMultiply',
+]
 
 # The ParameterSet is actually a pretty hidden type, so it's easiest to access
 # it like this.
--- a/qutip/tests/core/data/test_mathematics.py
+++ b/qutip/tests/core/data/test_mathematics.py
@@ -8,6 +8,11 @@ from qutip.core.data import Data, Dense, CSR
 
 from . import conftest
 
+__all__ = ['TestAdd', 'TestAdjoint', 'TestConj', 'TestExpm', 'TestInner',
+          'TestTranspose', 'TestInnerOp', 'TestInv', 'TestKron', 'TestMatmul',
+          'TestMul', 'TestTrace', 'TestSub', 'TestProject', 'TestPow',
+          'TestNeg', 'TestMultiply', 'TestMultiply']
+
 # The ParameterSet is actually a pretty hidden type, so it's easiest to access
 # it like this.
 _ParameterSet = type(pytest.param())
--- a/qutip/tests/core/data/test_norm.py
+++ b/qutip/tests/core/data/test_norm.py
@@ -6,6 +6,8 @@ from qutip import data
 from qutip.core.data import CSR, Dense
 import numbers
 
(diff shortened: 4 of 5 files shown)


jakelishman:
I'm not at all a fan of making tests part of the public interface like this.  It's something that we make absolutely no stability guarantees about, and coupling downstream packages to our own tests just sounds like a recipe for disaster - if we find a bug in main QuTiP and need to change tests or add methods (which we should be able to do completely freely), we're going to inject a ton of extra code into downstream packages, and may completely break them.  To be honest, I'm not even wild about how we distribute our tests as part of the module code at all, but that's what we've got.

It _could_ be ok if you import the test case base, and redefine your tests from there.  We still don't make any stability guarantees about that, but it's conceivable that we could (though of course, the problem of "what if we need to add methods" persists).  That would also solve your problems about saying you can't import individual tests.  You can import individual tests anyway, you just need to rename them to things like `_TestAdd`, or use one of the pytest hooks in your local `conftest.py` to reject tests whose `__module__` isn't your local module, or do one of the other methods of filtering.

jakelishman:
Also, this ch

hodgestar:
I agree with @jakelishman's general principle (and maybe we should even move the tests out of the `qutip` package for v5 still) but it's also really important to have an easy way for data layer implementations to run a comprehensive test suite *that is defined by the data layer*.

I suspect that if anyone is running tests against master (i.e. what is currently dev.major) they will *want* tests to break if we add new methods.

I have a slightly different proposal:

What if we create `qutip.core.data.testing` which is an explicitly exported suite of pytest classes that we intentionally expose for data layers to use? Initially the file can just import test classes from `qutip.tests.core.data`.

I would also not use `__all__` in the individual test packages but instead import the test classes explicitly in `qutip.core.data.testing` so that it's 100% clear by reading just that file what the exported interface is (and adding new tests requires explicitly adding them into the exposed package).

jakelishman:
If it's something we want to support, we'll need a bit of a rewrite of the data-layer testing infrastructure, but nothing too major.  I'd be in favour of a bit of a larger restructure:
- rearrange the repo root to have `/src/qutip` and `/test/qutip` instead of `/qutip/tests` (I originally didn't like it, but I now see more and more benefits of the `/src` (or `/lib`) convention in Python projects).
- the current `qutip/tests` should entirely become `/test/qutip`
- a new `/src/qutip/test` becomes a fully-supported _public_ module, including base classes (such as the data-layer mixins) and things like that.

We'd need to be careful, because reducing your possibilities for flexibility in testing generally sounds like a bad idea to me, but if we take care to be very limited, it could work.  I suspect you won't be able to import the test suite from inside `qutip.core.data` because of cyclical dependencies, which is probably a good thing - it'll force us to put the structure the right way round immediately.

For an idea pretty far out of left field:
- the data layer tests should really _discover_ the specialisations rather than us needing to specify them (this is true regardless of this PR)
- if so, in the new form, we _could_ directly export this part of the test suite from `qutip.test` (our test suite would import this as well), and add a little bit of logic to allow filtering of which tests need to be run (class decorators, probably)
- downstream packages could then test just by importing our classes and disabling any specialisations that are only on internal data-layer objects

I'm not 100% sure how I feel about this, but it's plausible - we'd still be able to add extra methods in our own test suite if we wanted them, because we could just derive from the classes, rather than defining the names into the test suite in those cases.

---

About us breaking downstream CI: having been a maintainer of QuTiP and now Qiskit, I have a deep-rooted hatred for any upstream package that forces CI failures on its dependents, because it _requires_ the lower packages to move "fix CI" way up to the top of their priority list; you can't merge _anything_ until CI is running again, which is very frustrating.  It messes with all your workflows.

AGaliciaMartinez:
Thanks both @jakelishman and @hodgestar for your comments! I agree that I may be depending too much on the classes in `qutip.tests.core.data` to do the testing both in qutip-tensorflow and qutip-tensornetwork. I myself was worried of "who is going to tests the tests" if this trend continued with other packages. Reusing the `qutip.tests.core.data`'s code though helps to speed things up in the development process and also allows QuTiP to strictly define what each specialisation is meant to do. 

As you mentioned above, I think it would be best to create a `qutip.core.data.testing` module where all the classes for testing are implemented. Then these classes could be tested to ensure that the proper tests are created and these classes ca be used by both QuTiP and downstream packages. I do not have time for this now but I will come to this later, so lets keep this PR open to discuss any ideas!

----

> the data layer tests should really discover the specialisations rather than us needing to specify them (this is true regardless of this PR)

I am not sure about this one. Automatic discovery of tests for specialisations seems to lead to very restricted flexibility in how tests are parametrized. For instance, I needed to change the `rtol` to accommodate the tests for the complex64 data layer in qutip-tensorflow and I am not sure how this could be added automatically. 

Also, as @hodgestar mentioned today, you probably want to be explicit on adding tests so that you are confident that those tests are running. Otherwise you need to constantly check the log to ensure the proper tests are running. 

Another thing is that by being explicit on what is tested, you could assert if the tested specialisation has not been registered (which actually happened to both qutip-cupy and qutip-tensorflow before: specialisations where being tested but, surprisingly, they were not registered).

jakelishman:
I think the module should really be `qutip.testing` (or `qutip.test` if the name's already taken), not `qutip.core.data.testing`, if you're going to add one.  It's very likely to be the case in the future (if it isn't already) that you'll need additional dependencies to define the tests, and you don't want to force _all_ runtime users to have them to install QuTiP.  If you put the package in a separate top-level module, it's _far_ easier to separate out all the code, and have `qutip/__init__.py` not import `qutip.testing`, so things won't break.  Also, tests generally need to work with a fully imported version of the module.  If we start putting testing infrastructure in the core, we're going to encode a whole lot of cyclical dependencies right away, or end up with `testing` modules scattered all across the codebase, for no real gain.

For discovery: that would only happen during test generation, which comes after importing the modules and defining the classes, so you'd still be able to define tolerances and influence how the tests ran.

The problem with _not_ having discovery is that it becomes very easy to forget to test a function, and with so much of the machinery automated, it's also easy to forget it when reviewing a PR.  That's particularly true once you have external contributors writing these sorts of things - we have it in Qiskit, where people add things to our "standard equivalence library", and new contributors and new reviewers don't know where all the different tests are, so our discovery tests have helped catch bugs.  That happens even with super experienced team members.

I'd argue that code coverage metrics are possibly a better way of detecting missing tests - that fails safe if you add broken code and forget to test it, whereas having to specify the tests explicitly means you can accidentally add broken code to a dispatcher and not know, if you forgot to add the test.

hodgestar:
I'm closing this for now until we find a solution that we like better and time to work on it. Happy to revisit this or re-open the PR if someone wants to work on it.

PR #1726, opened 2021-11-25
Add Matsubara and Padé expansion Lorentzian baths for the HEOM solver.

hodgestar:
**Description**
Add Matsubara and Padé expansion Lorentzian baths for the HEOM solver.

**Todo**
- [x] Clean-up the Pade expansion code.
- [x] ~~Somehow get the Matsubara and Pade results for the current to match.~~ (These are not supposed to match!)
- [x] Add solver tests for the Lorentzian baths.
- [x] Add bath tests for the Lorentzian baths.
- [x] Add tests for the DrudeLorentzPadeBath terminator.
- [x] Convert the slightly awkward terminator= argument for Drude-Lorentz baths to a terminator method.

**Related issues or PRs**
* #1601
* #1724

**Changelog**
Add Matsubara and Padé expansion Lorentzian baths for the HEOM solver.

### DIFF
--- a/qutip/nonmarkov/bofin_baths.py
+++ b/qutip/nonmarkov/bofin_baths.py
@@ -800,12 +800,12 @@ class LorentzianBath(FermionicBath):
             Q, ck_plus, vk_plus, ck_minus, vk_minus, tag=tag,
         )
 
-    def _corr(self, gamma, w, T, lmax, sigma, mu):
+    def _corr(self, gamma, w, T, Nk, sigma, mu):
         beta = 1. / T
         kappa = [0.]
-        kappa.extend([1. for _ in range(1, lmax + 1)])
+        kappa.extend([1. for _ in range(1, Nk + 1)])
         epsilon = [0]
-        epsilon.extend([(2 * ll - 1) * np.pi for ll in range(1, lmax + 1)])
+        epsilon.extend([(2 * ll - 1) * np.pi for ll in range(1, Nk + 1)])
 
         def f(x):
             # kB = 1.0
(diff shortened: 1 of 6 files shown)


qutip/nonmarkov/bofin_baths.py
```diff
@@ -752,3 +752,167 @@ def __init__(self, Q, ck_plus, vk_plus, ck_minus, vk_minus, tag=None):
                 "-", 2, Q, ckm, vkm, sigma_bar_k_offset=-1, tag=tag,
             ))
         super().__init__(exponents)
+
+
+class LorentzianBath(FermionicBath):
+    def __init__(self, Q, gamma, w, T, theta, lmax, tag=None):
+        mu_l = theta / 2.0
+        mu_r = - theta / 2.0
+
```
nwlambert, line 761: One quick suggestion for both functions, as written here this parameterizes two baths (a L (left) and (R) right bath at chemical potentials mu_l and mu_r). Probably better if we just have these functions parameterize a single bath at chemical potential mu, and use the tags to differentiate them.

hodgestar: Great idea -- implemented!

hodgestar:
@nwlambert I removed the "WIP" from this -- it's now ready for a final review.

I've also converted the slightly awkward "bath = DrudeLorentzBath(..., terminator=True); ... ; bath.terminator" dance into a ".terminator()" method, so that now one just does "bath.terminator()" if one wants a terminator.

nwlambert:
The fermionic stuff looks to me!  For the change in the terminator, I see the logic I guess.  Also makes it easier for the user to include it or not, without remaking the bath.  I have some reservations about it just because I can imagine someone following the examples and using jupyter notebook rerunning the same cell multiple times and continuously modifying their Hsys or something?  But that can just be made clear in the examples so lets go ahead and merge this.

nwlambert (approved):

hodgestar:
Thanks!

Re the terminator -- yes, we'll have to be a bit careful in notebooks. I think that was true before this change though.

PR #1727, opened 2021-11-26
Add a compile time option for 64 bit indices.

### LINKED ISSUE #850: Use int64 for sparse indices by default
Recently, we have seen a number of issues with the limitations imposed by using int32 for the sparse matrix indices in QuTiP (#845, #842, #828, etc).  Supporting **both** int32 and int64 indices is quite a bit of work, and we really do not have the resources to make such a big change.  Instead, might it not be better to switch everything over to int64?  This would result in a ~20% increase in the size of the sparse matrices, but would allow for handling the really large problems that people are looking into these days.  @ajgpitch, @Ericgig, @sahmed95, what do you think?

hodgestar:
**Description**
Adds a --with-idxint-64 compilation option for building with 64-bit integer indices in the data layer.

Example usage:
```
$ python setup.py develop --with-idxint-64
```

**Related issues or PRs**
- fix #850

**Changelog**
Added a --with-idxint-64 compilation option for building with 64-bit integer indices in the data layer.


setup.py
```diff
@@ -72,6 +72,12 @@ def _determine_user_arguments(options):
     )
     if "--with-openmp" in sys.argv:
         sys.argv.remove("--with-openmp")
+    options['intidx_64'] = (
+        '--with-intidx-64' in sys.argv
+        or _get_environment_bool('CI_QUTIP_WITH_INTIDX_64')
+    )
+    if "--with-intidx-64" in sys.argv:
+        sys.argv.remove("--with-intidx-64")
     return options
```
jakelishman, line 81: If we're going to extend this a lot, at some point we should probably do it properly with `argparse`, but I don't know how well that's going to interact with `setuptools` getting the arguments.

jakelishman (commented):
Compilation errors aside, a couple of minor comments below

jakelishman:
But also, yeah, this is exactly the change I meant when I wrote that comment in the other issue.  Unless something's broken in the interrim, it definitely works because I tested the swap fairly heavily when I wrote it, which is also why there's some particularly weird templated code in `core/data/src/matmul_csr_vector.{hpp,cpp}`.

qutip/core/data/base.pxd
```diff
@@ -4,7 +4,10 @@ import numpy as np
 cimport numpy as cnp
 
 # If you change the typedef, change the objects in base.pyx too!
-ctypedef cnp.npy_int32 idxint
+IF QUTIP_INTIDX_64:
+    ctypedef cnp.npy_int64 idxint
```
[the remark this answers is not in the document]
hodgestar: Thanks! Doh. Rushing too much. Fixed.

setup.py
hodgestar: I thought something similar. I did a small tidy up for now.

hodgestar:
The current test failures can probably be fixed by merging #1651, so I should review that.

Ericgig:
#1651 create a time-dependent bloch redfield tensor, but do not create a new `brmesolve` using it and remove the `br_codegen.py` etc. that uses it. So it won't solve it yet.

hodgestar:
> #1651 create a time-dependent bloch redfield tensor, but do not create a new `brmesolve` using it and remove the `br_codegen.py` etc. that uses it. So it won't solve it yet.

I don't think we need to rush -- we can finish #1651 and then do the next one, and then hopefully come back to this one. At least coefficients seem to work now, so hopefully people who are keen can try out this branch in the mean time if they want to.

hodgestar:
#1651 has been merged into dev.major and I've merged that back into this branch. Tests should still fail, but this branch is at least up to date again.

hodgestar:
Oooh. Tests pass. Perhaps we can merge the branch?

@Ericgig Would you mind reviewing?

qutip/core/data/kron.pyx
```diff
@@ -15,7 +15,7 @@ __all__ = [
 
 
 @cython.overflowcheck(True)
-cdef cython.numeric _mul_checked(cython.numeric a, cython.numeric b):
+cdef idxint _mul_checked(idxint a, idxint b):
```
Ericgig, line 18: Why change this? Template in cython work so that that  each use of `cython.numeric` are the same type in all specialization. So it would work and be safe both way.

qutip/core/coefficient.py
```diff
@@ -498,8 +499,15 @@ def compile_code(code, file_name, parsed, c_opt):
...
+            compile_time_env = {
+                "QUTIP_IDXINT_64": idxint_dtype is np.int64,
+            }
             with _ignore_import_warning_for_pyximporter():
-                ext_modules = cythonize(coeff_file, force=c_opt['recompile'])
+                ext_modules = cythonize(
+                    coeff_file,
+                    compile_time_env=compile_time_env,
+                    force=c_opt['recompile'],
+                )
```
Ericgig, line 510: The `int32` vs `int64` is for the data in `CSR`, it does not affect the coefficient. If needed, the coefficient should cimport `base.idxint` instead of using environment variable.

qutip/core/data/kron.pyx
Ericgig: Ok, cython doc says it goes up to long which can be 32 or 64 bit depending on the system so we can't rely on it.

Ericgig:
If we look at https://github.com/cython/cython/issues/4310, they plan to deprecate the compilation time `IF` in a future release.
They suggest using cmacro or code generation to replace this.

I suggest to take more time thinking about this and use those method instead.

As it is, we would have to set a compilation environment variable each time something want to `cimport Data`, even from outside qutip...

hodgestar:
@Ericgig I think it's unavoidable to have some sort of global state that tells the C compiler which type to choose, and I think passing it as a compilation option is a simple and straight forward solution. Using macros won't change the need for a compilation flag and will make fetching the correct numpy type trickier. If cython does eventually provide a nicer option, I am happy to use it, but they are still having long discussions about the right direction to take.

One thing that would make things nicer would be to expose the compilation options used in, for example, `qutip.utilities` so that compilation becomes `compile_time_env=qutip.utilities.compile_time_env()`.

hodgestar:
This has been superseded by #1874 which has been merged. Woot.

PR #1734, opened 2021-12-08
Better handling of scipy ode in integrator

Ericgig:
**Description**
Make our integrator re-entrant. Scipy'ode is can only have one active instance active for some of the methods. This cuased no issue when the whole integration is done at once as when used in `mesolve`. But with solver as class, this mean using them will have side effect on other solver instances.  This PR make the check for concurrent integrator usage before scipy does and force a reset if needed. Switching between system will be inefficient, but will return the expected results.

In lsoda's integrator's `_backstep` we are catching a warning before acting on it, but catching the warnings still printed them at the end of tests, bloating the output. In this PR, I also do the check before scipy does so the warning is never raised.

**Related issues or PRs**
Point raised in discussion with @hodgestar and @AGaliciaMartinez on #1710 

**Changelog**
Front run some safety check in scipy.ode

### DIFF
--- a/qutip/solver/ode/scipy_integrator.py
+++ b/qutip/solver/ode/scipy_integrator.py
@@ -86,28 +86,27 @@ class IntegratorScipyZvode(Integrator):
             state = _data.dense.Dense(self._ode_solver._y, copy=copy)
         return t, state
 
-    def integrate(self, t, copy=True):
-        if (
-            self._ode_solver._integrator.initialized
-            and (self._ode_solver._integrator.handle !=
-                 self._ode_solver._integrator.__class__.active_global_handle)
-        ):
-            self._ode_solver._integrator.reset(
-                len(self._ode_solver._y), self._ode_solver.jac is not None)
+    def _check_handle(self):
+        """
+        Do the check for concurrent use of the integrator and reset if used
+        elsewhere.
+        """
+        if self._ode_solver._integrator.initialized:
+            if (
+                self._ode_solver._integrator.handle !=
+                self._ode_solver._integrator.__class__.active_global_handle
+            ):
+                self._ode_solver._integrator.reset(
+                    len(self._ode_solver._y), False)
 
+    def integrate(self, t, copy=True):
+        self._check_handle()
         if t != self._ode_solver.t:
             self._ode_solver.integrate(t)
         return self.get_state(copy)
 
     def mcstep(self, t, copy=True):
-        if (
-            self._ode_solver._integrator.initialized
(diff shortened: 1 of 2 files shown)


qutip/tests/solver/test_integrator.py
```diff
@@ -101,3 +103,21 @@ def me_method(self, request):
...
     def mc_method(self, request):
         return request.param
+
+
+@pytest.mark.parametrize('integrator',
+    [IntegratorScipyZvode, IntegratorScipylsoda], ids=["zvode", "lsoda"])
+def test_concurent_usage(integrator):
+    sys = qutip.QobjEvo(qutip.qeye(1))
+    opt = SolverOdeOptions()
+    sys = qutip.QobjEvo(0.5*qutip.qeye(1))
```
AGaliciaMartinez, line 113: Is the first assign to `sys` necessary given that there is another one?

AGaliciaMartinez: I see it is now resolved. Thank you Eric!

qutip/solver/ode/scipy_integrator.py
```diff
@@ -277,17 +304,16 @@ def _one_step(self, t):
         # Here we want to advance up to t doing maximum one step.
         # We check if a new step is needed.
         t_front = self._ode_solver._integrator.rwork[12]
+        safe_delta = self._ode_solver._integrator.rwork[11]/10 + 1e-15
```
hodgestar, line 307: Could you explain a little what the safe_delta change does and why it is better? I'm happy for it to stay in this PR but it seems unrelated to the concurrency checks?

qutip/solver/ode/scipy_integrator.py
```diff
@@ -296,14 +322,13 @@ def _backstep(self, t):
...
-        # Scipy raise a warning, not an Error.
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore")
+        t_front = self._ode_solver._integrator.rwork[12]
+        delta = self._ode_solver._integrator.rwork[11]
+        if t == self._ode_solver.t:
+            pass
+        elif t_front - delta <= t:
             self._ode_solver.integrate(t)
-        if not self._ode_solver.successful():
```
hodgestar, line 306: Do we not still need a `.successful()` check somewhere?

hodgestar (commented):
I left some questions, but woot for finding a way to check for the concurrent usage.

hodgestar (approved):

PR #1735, opened 2021-12-09
Update build process to support Python 3.10

hodgestar:
**Description**
- [x] Remove the restriction to `numpy<1.20` in `setup.cfg` (numpy >= 1.20 is required by Python 3.10)
- [x] Update the build process to explicitly build with the lowest supported numpy
- [x] While we're here, update to a more recent version of ciwheelbuild. 
- [x] Build documentation on Python 3.9.

**Related issues or PRs**
* Fixes #1732
* Fixes #1720
* Fixes #1740
* We also need to update our conda-forge feedstock -- https://github.com/conda-forge/qutip-feedstock/pull/72
* Alternative implementation from @henryiii in #1738

**Changelog**
* Updated the build process to support building on Python 3.10 by removing the install requirement for `numpy < 1.20`.
* Update the version of cibuildwheel used to build wheels to 2.3.0.
* Update the version of Python used in documentation builds to Python 3.9.

PR #1738, opened 2021-12-09
feat: Support Python 3.10

henryiii:
This is testing using oldest supported numpy and enabling 3.10.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,6 +91,7 @@ jobs:
           else
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
+          conda install packaging cython
           python -m pip install -e .[$QUTIP_TARGET] --no-build-isolation
           python -m pip install pytest-cov coveralls
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,7 +91,7 @@ jobs:
           else
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
-          python -m pip install -e .[$QUTIP_TARGET]
+          python -m pip install -e .[$QUTIP_TARGET] --no-build-isolation
           python -m pip install pytest-cov coveralls
 
       - name: Package information
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,11 @@ install_requires =
     numpy>=1.16.6
     scipy>=1.0
     packaging
+setup_requires =
+     numpy
+     scipy>=1.0
+     cython>=0.29.20
+     packaging
 
 [options.packages.find]
 include = qutip*
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,11 +30,6 @@ install_requires =
     numpy>=1.16.6
     scipy>=1.0
     packaging
-setup_requires =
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           else
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
-          python -m pip install -e .[$QUTIP_TARGET]
+          python -m pip install -e .[$QUTIP_TARGET] --no-build-isolation
```
henryiii, line 94: (If you don't do this, you'll get PyPI requirements when you are building your package, instead of conda ones)

hodgestar:
@henryiii Thank you for taking the time to make this much nicer looking PR. It does look much better, I just have to try convince myself everything will work now.

It looks like the build is currently failing because `--no-build-isolation` doesn't install the build depedencies so we'll need to install them manually. I wonder what the best way to support that is (since users would likely have to do it themselves regularly too).

henryiii:
Users should not be installing pip dependencies into a conda environment. They should be using conda, not pip. You are only running into this because you are using pip inside a conda environment, and yes, you'll need the dependencies (with their proper conda names) in that environment.

(And, since wheels work inside conda, `pip install cutip` should work inside a conda env, even if they should be using conda - it's only because you are building inside it, which is very much not supported for normal users)

.github/workflows/tests.yml
henryiii:
```suggestion
          conda install packaging cython
          python -m pip install -e .[$QUTIP_TARGET] --no-build-isolation
```
(And any more, if there are any others missing)

setup.cfg
```diff
@@ -30,11 +30,6 @@ install_requires =
     numpy>=1.16.6
     scipy>=1.0
     packaging
-setup_requires =
-    numpy>=1.16.6,<1.20
-    scipy>=1.0
-    cython>=0.29.20
-    packaging
```
hodgestar, line 37: Removing this prevents use of `python setup.py develop` and I don't think there is another way to rebuild only the Cython files that have changed while developing at the moment. :/

henryiii: This has been deprecated for a while now, and recently has started producing warnings. I would recommend requiring users running `setup.py` to pre-install requirements. And I'd worry about using this command: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html 

(But removing isn't important for this PR)

henryiii: Put it back in, but without the problematic numpy constraints.

hodgestar: Thanks!

Re the article -- yes, I'm aware of the various pip options, but `pip install -e .` doesn't cache Cython file builds (maybe there is some new intree-builds version of pip that I haven't seen yet that does?).

henryiii: Have you tried this with pip 21.3? I'd have thought this might have started working now. Haven't tested it yet for this, though.

hodgestar:
@henryiii I've been borrowing some of your changes from here and incorporating them into #1735.

setup.cfg
hodgestar: It looks like it does with pip 21.3 if I run with `pip install -e . --no-build-isolation -v`. Not the friendliest command, but progress?

hodgestar:
@henryiii I'm closing this in favour of #1747, which is very similar and was inspired by this PR. Thank you for all your help! Further suggestions welcomed too.

PR #1741, opened 2021-12-14
Fix silent ignoring of offset in coherent and coherent_dm when offset is non-zero.

hodgestar:
**Description**
Raise an error instead of silently ignoring the offset when attempting to generate coherent states or density matrices with method 'operator' and non-zero offset.

**Related issues or PRs**
* Small fix post #1469

**Changelog**
Fix silent ignoring of offset in coherent and coherent_dm when offset is non-zero. A ValueError is now raised instead.


hodgestar:
I also removed an unused import and a bare except because robot minions complained about them while editing.

PR #1744, opened 2021-12-22
Ptrace tests specalisation

AGaliciaMartinez:
**Checklist**

**Description**
Added ptrace specialisation tests. While adding these tests a bug came out with `ptrace_dense` which is now fixed. The problem was it did not work for `sel=[]`, that is, no selection and hence the full trace was required. The bug comes from how `numpy` works as the output of np.prod([]) is float instead of integer. 

I wonder though what should be the behaviour for `qobj.ptrace([])`. Shouldit not be the same as `qobj.trace()`?  `qobj.ptrace([])` currently returns a `Qobj` of dimensions 1x1 whereas `qobj.trace()` returns a complex number.

**Changelog**
Fixed error with ptrace when the input and output are dense and `sel=[]`.
Added specialisations tests for ptrace.


AGaliciaMartinez:
Closing because i messed up selecting master as target branch.

PR #1746, opened 2021-12-22
Added ptrace specialisation tests and fixed a bug for ptrace_dense.

AGaliciaMartinez:
**Checklist**

**Description**
Added ptrace specialisation tests. While adding these tests a bug came out with `ptrace_dense` which is now fixed. The problem was it did not work for `sel=[]`, that is, no selection and hence the full trace was required. The bug comes from how `numpy` works as the output of np.prod([]) is float instead of integer. 

I wonder though what should be the behaviour for `qobj.ptrace([])`. Should it not be the same as `qobj.trace()`?  `qobj.ptrace([])` currently returns a `Qobj` of shape 1x1 whereas `qobj.trace()` returns a complex number.

**Changelog**
Fixed error with ptrace when the input and output are dense and `sel=[]`.
Added specialisations tests for ptrace.
Improved `ptrace` tests to test for edge cases (trace all and trace none).
Added a few more checks to catch invalid dims and raise ValueError.

### DIFF
--- a/qutip/core/data/ptrace.pyx
+++ b/qutip/core/data/ptrace.pyx
@@ -17,34 +17,28 @@ __all__ = [
     'ptrace', 'ptrace_csr', 'ptrace_dense', 'ptrace_csr_dense',
 ]
 
-cdef void _check_shape(Data matrix, object dims) except *:
-    if matrix.shape[0] != matrix.shape[1]:
+cdef tuple _check_inputs(tuple shape, object dims, object sel):
+    cdef Py_ssize_t i
+
+    if shape[0] != shape[1]:
         raise ValueError("ptrace is only defined for square density matrices")
 
-    if matrix.shape[0] != np.prod(dims, dtype=int):
-        raise ValueError(f"the input matrix shape, {matrix.shape} and the"
+    if shape[0] != np.prod(dims, dtype=int):
+        raise ValueError(f"the input matrix shape, {shape} and the"
                          f" dimension argument, {dims}, are not compatible.")
-
-cdef tuple _prepare_inputs(object dims, object sel):
-    cdef Py_ssize_t i
-    dims = np.atleast_1d(dims).astype(idxint_dtype).ravel()
-    sel = np.atleast_1d(sel).astype(idxint_dtype)
-    sel.sort()
-
     if sel.ndim != 1:
         raise ValueError("Selection must be one-dimensional")
 
     for d in dims:
         if d<1:
             raise ValueError("dimensions must be greated than zero but where"
-                             f" dims={dims}")
+                             f" dims={dims}.")
 
     for i in range(sel.shape[0]):
         if sel[i] < 0 or sel[i] >= dims.size:
             raise IndexError("Invalid selection index in ptrace.")
         if i > 0 and sel[i] == sel[i - 1]:
             raise ValueError("Duplicate selection index in ptrace.")
-    return dims, sel
 
 cdef idxint _populate_tensor_table(dims, sel, idxint[:, ::1] tensor_table) except -1:
     """
@@ -93,8 +87,12 @@ cdef inline void _i2_k_t(idxint N, idxint[:, ::1] tensor_table, idxint out[2]):
 
 
 cpdef CSR ptrace_csr(CSR matrix, object dims, object sel):
-    _check_shape(matrix, dims)
-    dims, sel = _prepare_inputs(dims, sel)
+    dims = np.atleast_1d(dims).astype(idxint_dtype).ravel()
+    sel = np.atleast_1d(sel).astype(idxint_dtype)
+    sel.sort()
+
+    _check_inputs(matrix.shape, dims, sel)
+
     if len(sel) == len(dims):
         return matrix.copy()
     cdef idxint[:, ::1] tensor_table = np.zeros((dims.shape[0], 3), dtype=idxint_dtype)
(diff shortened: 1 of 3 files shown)


qutip/tests/core/test_ptrace.py
```diff
@@ -55,7 +55,7 @@ def expected(qobj, sel):
             continue
         tmp_dims = (before, dim, after) * 2
         out = np.einsum('aibcid->abcd', out.reshape(tmp_dims))
-    return qutip.Qobj(out.reshape(new_shape), dims=dims)
+    return qutip.Qobj(out.reshape(new_shape), dims=dims, type='oper')
```
AGaliciaMartinez, line 58: It seems that Qobj is not able to infer the type from an empty dims (dims=[[][]]).

qutip/tests/core/data/test_ptrace.py
```diff
@@ -0,0 +1,96 @@
...
+    def generate_incorrect_dims_raises(self, metafunc):
+        self.generate_mathematically_correct(metafunc)
+
+    @pytest.mark.parametrize('sel',
+                             [[2, 10], [-1, 2]],
+                             ids=['sel_value_larger_than_dims',
+                                  'sel_value_negative',
+                                 ])
+    def test_incorrect_sel_raises(self, op, data_m, out_type, sel):
+        with pytest.raises(IndexError):
```
AGaliciaMartinez, line 92: Should we raise ValueError instead?

hodgestar: Is `sel` not conceptually used as `dims[sel]` effectively? If so, that looks like an incorrect index?

qutip/tests/core/test_ptrace.py
hodgestar: That seems not completely unexpected.

qutip/tests/core/test_ptrace.py
```diff
@@ -120,29 +120,42 @@ def test_ptrace_fails_on_invalid_input(state, selection, exception):
...
+
+@pytest.mark.parametrize('sel', [[], [0, 1, 2], [0], [1], [1, 0], [0, 2]],
+                        ids=['trace_all',
+                             'trace_none',
+                             'trace_one',
+                             'trace_one_2',
+                             'trace_multiple',
+                             'trace_multiple_not_sorted',
+                            ])
+def test_ptrace_dm(dtype, sel):
```
hodgestar, line 145:
```suggestion
def test_ptrace_rand_dm(dtype, sel):
```
Just to match the name of `_rand_ket` above.

hodgestar (changes requested):
Left some small suggestions but woot for adding tests and finding the differing behaviour.

Regarding `qobj.ptrace([])` vs `qobj.trace()`: I *think* the current behaviour is fine. Yes, they are in some sense the same operation, but it's probably better for `ptrace` to always return a Qobj, rather than to suddenly return a complex number when a particular set of arguments is used.

qutip/tests/core/test_ptrace.py
AGaliciaMartinez: I was not very clear what the problem was from the error message so it took me a little bit of digging to figure it out.

qutip/tests/core/data/test_ptrace.py
AGaliciaMartinez: I guess so. I wont change it then.

qutip/core/data/ptrace.pyx
```diff
@@ -17,17 +17,28 @@ __all__ = [
...
     cdef Py_ssize_t i
     dims = np.atleast_1d(dims).astype(idxint_dtype).ravel()
     sel = np.atleast_1d(sel).astype(idxint_dtype)
+    sel.sort()
+
     if sel.ndim != 1:
         raise ValueError("Selection must be one-dimensional")
-    sel.sort()
+
+    for d in dims:
```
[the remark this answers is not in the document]
AGaliciaMartinez: I ended up adding a  `_check_inputs` function that performs all the checks and included the preparation step explicitly in each specialisation (they were only three lines after all). Hope it is more clear now.

AGaliciaMartinez:
@hodgestar, thank you for the review! I believe all the suggestions were applied. Let me know if any other change is needed.

qutip/core/data/ptrace.pyx
```diff
@@ -17,23 +17,28 @@ __all__ = [
...
+        raise ValueError(f"the input matrix shape, {shape} and the"
+                         f" dimension argument, {dims}, are not compatible.")
     if sel.ndim != 1:
         raise ValueError("Selection must be one-dimensional")
-    sel.sort()
+
+    for d in dims:
+        if d<1:
+            raise ValueError("dimensions must be greated than zero but where"
+                             f" dims={dims}.")
```
hodgestar, line 35:
```suggestion
    if any(d < 1 for d in dims):
        raise ValueError("dimensions must be greated than zero but where"
                         f" dims={dims}.")
```

qutip/core/data/ptrace.pyx
```diff
@@ -82,8 +87,12 @@ cdef inline void _i2_k_t(idxint N, idxint[:, ::1] tensor_table, idxint out[2]):
 
 
 cpdef CSR ptrace_csr(CSR matrix, object dims, object sel):
-    _check_shape(matrix)
-    dims, sel = _prepare_inputs(dims, sel)
+    dims = np.atleast_1d(dims).astype(idxint_dtype).ravel()
+    sel = np.atleast_1d(sel).astype(idxint_dtype)
+    sel.sort()
```
hodgestar, line 92: Could we move these into `_check_inputs` somehow? My concern is that it is an error to call `_check_inputs` without running these lines (e.g. `dims` might be very nested) and it would be easy for us to forget this later.  It would require returning `dims` and `sel` and maybe changing the function name to `_parse_inputs` or something that suggests that inputs are parsed and modified.

hodgestar (changes requested):
Two more small suggestions!

hodgestar (approved):

PR #1748, opened 2022-01-02
refactor process_fidelity using new function _hilbert_space_dims

fhopfmueller:
**Description**
This is a refactor of the changes from https://github.com/qutip/qutip/pull/1712. Previously, the logic which extracts the dimensions of the input and output Hilbert spaces of the involved quantum channels was spread out throughout `process_fidelity` and `_process_fidelity_to_id`. This PR collects it in a new private function `core.metrics._hilbert_space_dims`.

This will also allow reducing `average_gate_fidelity` to `process_fidelity` (in a future PR) as explained in https://qiskit.org/documentation/stubs/qiskit.quantum_info.average_gate_fidelity.html, without having to re-implement extracting the dimensions.

**Related issues or PRs**
Part of the work on https://github.com/qutip/qutip/issues/1703.

**Changelog**
refactor process_fidelity using new function _hilbert_space_dims

### DIFF
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -129,9 +129,12 @@ def _hilbert_space_dims(oper):
     """
     if isinstance(oper, list):
         return oper[0].dims
-    if oper.type == 'oper':
+    elif oper.type == 'oper':  # interpret as unitary quantum channel
         return oper.dims
-    return [oper.dims[0][1], oper.dims[1][0]]  # for Choi, chi, or super
+    elif oper.type=='super' and oper.superrep in ['choi', 'chi', 'super']:
+        return [oper.dims[0][1], oper.dims[1][0]]
+    else:
+        raise TypeError('oper is not a valid quantum channel!')
 
 
 def _process_fidelity_to_id(oper):
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -108,6 +108,32 @@ def fidelity(A, B):
     return float(np.real(np.sqrt(eig_vals[eig_vals > 0]).sum()))
 
 
+def _hilbert_space_dims(oper):
+    """
+    For a quantum channel `oper`, return the dimensions `[dims_out, dims_in]`
+    of the output Hilbert space and the input Hilbert space.
+    - If oper is a unitary, then `oper.dims == [dims_out, dims_in]`.
+    - If oper is a list of Kraus operators, then
+     `oper[0].dims == [dims_out, dims_in]`.
+    - If oper is a superoperator with `oper.superrep == 'super'`:
+     `oper.dims == [[dims_out, dims_out], [dims_in, dims_in]]`
+    - If oper is a superoperator with `oper.superrep == 'choi'`:
+     `oper.dims == [[dims_in, dims_out], [dims_in, dims_out]]`
+    - If oper is a superoperator with `oper.superrep == 'chi', then
+      `dims_out == dims_in` and
+      `oper.dims == [[dims_out, dims_out], [dims_out, dims_out]]`.
+    :param oper: A quantum channel, represented by a unitary, a list of Kraus
+    operators, or a superoperator
+    :return: `[dims_out, dims_in]`, where `dims_out` and `dims_in` are lists
+     of integers
+    """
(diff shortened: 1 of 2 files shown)


qutip/core/metrics.py
```diff
@@ -108,6 +108,32 @@ def fidelity(A, B):
     return float(np.real(np.sqrt(eig_vals[eig_vals > 0]).sum()))
 
 
+def _hilbert_space_dims(oper):
```
fhopfmueller, line 111: Should this function be moved to `core.dimensions`?

qutip/tests/core/test_metrics.py
```diff
@@ -396,6 +397,26 @@ def rand_super():
     ])
 
 
+@pytest.mark.parametrize('superrep_conversion',
+                         [lambda x: x, to_super, to_choi, to_kraus])
+def test__hilbert_space_dims(superrep_conversion):
```
hodgestar, line 402:
```suggestion
def test_hilbert_space_dims(superrep_conversion):
```
The double underscore is perhaps a bit easier to read, but we don't generally use it in QuTiP tests.

qutip/tests/core/test_metrics.py
```diff
@@ -396,6 +397,26 @@ def rand_super():
...
+    """
+    Metrics: check _hilbert_space_dims
+    """
+    dims = [[2, 3], [4, 5]]
+    u = Qobj(np.ones((np.prod(dims[0]), np.prod(dims[1]))), dims=dims)
+    assert_(_hilbert_space_dims(superrep_conversion(u)) == dims)
+
+
+def test__hilbert_space_dims_chi():
```
hodgestar, line 411:
```suggestion
def test_hilbert_space_dims_chi():
```

qutip/core/metrics.py
hodgestar: I think this could be useful in `core.dimensions`, but perhaps let's leave it here for this PR and then move it later. There are some additional considerations in moving it into `core.dimensions` -- e.g. when else will it be used, what is needed to make it return appropriate results for all possible qobjs (e.g. what should happen for bras and kets).

qutip/core/metrics.py
```diff
@@ -108,6 +108,32 @@ def fidelity(A, B):
...
+    :param oper: A quantum channel, represented by a unitary, a list of Kraus
+    operators, or a superoperator
+    :return: `[dims_out, dims_in]`, where `dims_out` and `dims_in` are lists
+     of integers
+    """
+    if isinstance(oper, list):
+        return oper[0].dims
+    if oper.type == 'oper':
+        return oper.dims
+    return [oper.dims[0][1], oper.dims[1][0]]  # for Choi, chi, or super
```
hodgestar, line 134: It would be good to check that `oper` really is an appropriate superop here if we can (maybe a check of `.superrep` is enough?).

hodgestar (changes requested):
Thanks! I did a quick review and left a few suggestions, but generally looks much better than things did before.

qutip/core/metrics.py
fhopfmueller: Done!

qutip/core/metrics.py
```diff
@@ -108,6 +108,35 @@ def fidelity(A, B):
...
+    :param oper: A quantum channel, represented by a unitary, a list of Kraus
+    operators, or a superoperator
+    :return: `[dims_out, dims_in]`, where `dims_out` and `dims_in` are lists
+     of integers
+    """
+    if isinstance(oper, list):
+        return oper[0].dims
+    elif oper.type == 'oper':  # interpret as unitary quantum channel
+        return oper.dims
+    elif oper.type=='super' and oper.superrep in ['choi', 'chi', 'super']:
```
hodgestar, line 134:
```suggestion
    elif oper.type == 'super' and oper.superrep in ['choi', 'chi', 'super']:
```
Tiny style suggestion. We really should add black or something to our CI at some point (but not today).

hodgestar (changes requested):

hodgestar (approved):
Approved! Just waiting for builds to complete after committing the small style fix.

PR #1751, opened 2022-01-05
Fix cibuildwheel builds.

hodgestar:
**Description**
Fix cibuildwheel issues introduced in #1747 and make any other necessary post PR clean ups.

**Related issues or PRs**
- Follow on from #1747

**Changelog**
Fix cibuildwheel builds in CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ manylinux-i686-image = "manylinux2014"
 # Change in future version to "build"
 build-frontend = "pip"
 
-[tool.cibuildwheel.override]
+[[tool.cibuildwheel.overrides]]
 # NumPy support manylinux2010 on CPython 3.6-3.9
 select = "cp3?-*"
 manylinux-x86_64-image = "manylinux2010"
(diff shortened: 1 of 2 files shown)


hodgestar:
@Ericgig @jakelishman I'm just going to merge this because it fixes the builds. Building with Python 3.10 doesn't "just work" so I'm going to leave that out for now and look at it later.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ manylinux-i686-image = "manylinux2014"
 # Change in future version to "build"
 build-frontend = "pip"
 
-[tool.cibuildwheel.override]
+[[tool.cibuildwheel.overrides]]
```
jakelishman, line 20: Are these square brackets meant to be doubled?

hodgestar: Amazingly yes -- https://cibuildwheel.readthedocs.io/en/stable/options/#examples!

I think I got it wrong initially because it was so unexpected.

hodgestar: See https://toml.io/en/v1.0.0#array-of-tables. The "O" is "TOML" is apparently not always that obvious. ;)

jakelishman: I can understand the necessity for something like that, but the syntax is awful...

PR #1754, opened 2022-01-06
Fix reading the CPU frequency on Apple M1.

hodgestar:
**Description**
Fix reading the CPU frequency on Apple M1.

Currently Mac OS does not provide the CPU frequency via sysctl (see discussion in this psutils bug -- https://github.com/giampaolo/psutil/issues/1892).

It seems that the only way to obtain the CPU frequency currently is to measure it -- see https://github.com/BitesPotatoBacks/osx-cpufreq/blob/fb82fd7afcffe9d76d98ad7af76217d7a9d34f38/M1cpufreq/M1cpufreq/main.cpp#L7-L35. I don't know what this measures exactly (e.g. is there clock scaling of some sort so that the measured CPU frequency might not be the maximum CPU frequency?) and it seems a lot of work to get a piece of information that isn't used directly by QuTiP, so I just fixed the result to 3.2 GHz for now (which according to Wikipedia is the clock speed for current M1s).

Hopefully sysctl will support CPU frequency on the M1 before we have to worry about this again.

**Related issues or PRs**
- Fixes #1606

**Changelog**
Fixed reading the CPU frequency on Apple M1.

PR #1758, opened 2022-01-11
Remove hardware info

Ericgig:
**Description**
With new version of os, it sometime break our hardware_info function, as with the new apple M1 with #1754.
Instead of fixing it each time, I propose removing the function:
- Qutip only use the number of cpu from that function, we never uses the cpu frequency or any other info obtained.
- There are cross platform modules for that (`psutil`), if we need that information, we should use them instead.

In this PR, I removed `hardware_info`, created a new function `available_cpu_count` which is placed it in `utilities.py` since `blas_info` is also there.  This function is set to detect the number of cpus _available_, not total number, so that when used on a cluster, it return the number of cpus assigned to the job.

I tested it on slurm.

**Changelog**
Remove hardware info


qutip/tests/test_utilities.py
```diff
@@ -105,3 +105,9 @@ def test_unit_clebsch_delta_m(j1, j2):
                 sum_differ += c1*c2
         assert sum_match == pytest.approx(1)
         assert sum_differ == pytest.approx(int(m1 == m1p and m2 == m2p))
+
+
+def test_cpu_count():
+    ncpus = available_cpu_count()
```
hodgestar, line 111: It would be good to test more of the cases if we can:

* We could test the environment variable cases using pytest's `monkeypatch.setenv`.
* I guess `os` vs `multiprocessing` we can test by letting them run on different enviroments one day (probably the `os` version doesn't work on Windows?)

Maybe just testing the environment variable overrides is enough for now.

hodgestar (changes requested):
I'd like to extent the test a little more, but otherwise I'm +1 on this -- the hardware info module was untested and largely just trouble for us.

I'm unsure whether this should go into 4.6.3 though. Can we keep it only for 4.7?

qutip/tests/test_utilities.py
Ericgig: I added a test with `monkeypatch`. Yes it would be better to test more situations, but I think we will need to test them manually.

I looked deeper and both `os` and `psutil` do not add the affinity function when unavailable on the os. So I rewrote the function with a less `try`.

Ericgig:
It feels more 4.6.3 since it's not a new feature but removing unused code in `__init__`, making qutip's import safer. So it looks more like a bug fix to me. Is there any specific reason to wait for 4.7?

hodgestar (approved):

hodgestar:
> It feels more 4.6.3 since it's not a new feature but removing unused code in `__init__`, making qutip's import safer. So it looks more like a bug fix to me. Is there any specific reason to wait for 4.7?

I was just thinking that we're removing functions that users might be using in their own code. On the other hand it is functionality that is untested and breaks on many modern machines, so I am not completely against adding this to 4.6.3.

hodgestar:
I'm okay with including this in 4.6.3. Merging. Thank you, @Ericgig!

PR #1764, opened 2022-01-13
Dev requirements

anna-naden:
**Description**
Module requirements for dev version of qutip as of 1/13/2022

**Related issues or PRs**
1763
**Changelog*
This is the output of pip freeze for an environment in which qutip has been built from scratch and in which "make html" in the doc folder runs cleanly


requirements.txt
```diff
@@ -1,4 +1,74 @@
...
+pybtex==0.24.0
+pybtex-docutils==1.0.1
+Pygments==2.11.2
+pyparsing==3.0.6
+pytest==6.2.5
+pytest-rerunfailures==10.2
+python-dateutil==2.8.2
+pytz==2021.3
+PyYAML==6.0
+qutip @ file:///home/anna_user2/Dropbox/projects/sphinx-warnings/qutip/dist/qutip-4.7.0.dev0%2Bnogit-cp310-cp310-linux_x86_64.whl
```
AGaliciaMartinez, line 53: Hey @anna-naden great work with the review of the documentation! I have been following along the different PR you opened but did not have time to review them properly. I did catch in this one something off. Specifically this line should not be included as it would fail in any other machine.

This requirements file is seldom use but it still only refers to the runtime requirements (see PR #1715 where it was last modified). There is another requirements file in `doc/` that does have the requirements for building the documentation (which maybe was what you were trying to actually modify?).

 In any case, pinning an exact version of a package is not a good idea as it is too restrictive. For example, different packages may have different requirements for numpys version. Forcing an specific version here means that it will conflict with other packages. Hence, I suggest removing the changes in this file.

Ericgig:
I am closing this: change to the documentation are redundant to previous PRs. The `requirements.txt` file refers to the packages needed to use qutip, which we keep to a minimum. Modules required to build the documentation are in `doc/requirements.txt`.

PR #1775, opened 2022-01-18
Guide dynamics intro2

anna-naden:
**Description**
Describe here the proposed change.

**Related issues or PRs**
1774
**Changelog**
Mention that sometimes we want dynamics rather than eigenstates. Classification of dynamical model, open/closed systems, unitary and non-unitary systems, collapse operators.
Table of solvers and return types.


hodgestar (approved):
There are some inaccuracies in the text, but I think it's better to have a short introduction and overview than not to have one, so I will merge this and it can be improved later.

PR #1777, opened 2022-01-19
Add a numpy 1.22 and Python 3.10 build to the CI test matrix.

hodgestar:
**Description**
Add a numpy 1.22 and Python 3.10 build to the CI test matrix.

**Related issues or PRs**
* If test pass consistently with numpy 1.22, we can close #1694

**Changelog**
Add a numpy 1.22 and Python 3.10 build to the CI test matrix.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -106,8 +106,7 @@ jobs:
           else
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
-          python -m pip install packaging
-          python -m pip install --no-build-isolation -e .[$QUTIP_TARGET]
+          python -m pip install -e .[$QUTIP_TARGET]
           python -m pip install pytest-cov coveralls
 
       - name: Package information
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -106,7 +106,7 @@ jobs:
           else
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
-          python -m pip install -e --no-build-isolation .[$QUTIP_TARGET]
+          python -m pip install --no-build-isolation -e .[$QUTIP_TARGET]
           python -m pip install pytest-cov coveralls
 
       - name: Package information
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -107,6 +107,7 @@ jobs:
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
           python -m pip install -e .[$QUTIP_TARGET]
+          python -m pip install coverage==6.2
           python -m pip install pytest-cov coveralls
 
       - name: Package information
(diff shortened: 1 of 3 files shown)


hodgestar:
@Ericgig This likely isn't the end of making things work on Python 3.10 and numpy 1.22, but it adds one build that works and fixes some issues.

qutip/steadystate.py
```diff
@@ -81,7 +81,7 @@ def _default_steadystate_args():
     def_args = {'sparse': True, 'use_rcm': False,
                 'use_wbm': False, 'use_precond': False,
                 'all_states': False, 'M': None, 'x0': None, 'drop_tol': 1e-4,
-                'fill_factor': 100, 'diag_pivot_thresh': None, 'maxiter': 1000,
+                'fill_factor': 100, 'diag_pivot_thresh': 0.1, 'maxiter': 1000,
```
hodgestar, line 84: I think that the default value of `diag_pivot_thresh` (i.e. the one we get when we pass `None`) is somehow wrong in recent versions of SciPy / SuperLU. The documentation of SuperLU (https://portal.nersc.gov/project/sparse/superlu/ug.pdf) has the default threshold as 1.0 for `splu` but `0.1` for everything else and the `splu` default has a comment that implies it should be `0.1` too.

Setting the default here to `1.0` causes it to fail the splu test, but setting it to anything else causes it to pass.

qutip/steadystate.py
```diff
@@ -1218,7 +1218,7 @@ def _pseudo_inverse_sparse(L, rhoss, w=None, **pseudo_args):
         LIQ = lu.solve(Q.toarray())
 
     else:
-        raise ValueError("unsupported method '%s'" % method)
+        raise ValueError("unsupported method '%s'" % pseudo_args['method'])
```
hodgestar, line 1221: This is just a fix while I'm here.

setup.cfg
```diff
@@ -6,7 +6,7 @@ long_description = file: README.md
 long_description_content_type = text/markdown; variant=GFM
 keywords = quantum, physics, dynamics
 license = BSD 3-Clause License
-license_file = LICENSE.txt
+license_files = LICENSE.txt
```
hodgestar, line 9: The name `license_file` is deprecated. Python 3.10 raises a warning if it is used and that causes runtime compilation of coefficient strings to fail (why pyimportx looks at the setup.cfg file is a bit of a mystery -- probably it just happens to be in the current directory when pyimportx runs setuptools or distutils).

.github/workflows/tests.yml
```diff
@@ -64,6 +64,17 @@ jobs:
...
 
+          # Python 3.10 and numpy 1.22
+          # Use conda-forge to provide numpy 1.22
+          # Ignore ImportWarning because pyximport registered an importer
+          # PyxImporter that does not have a find_spec method and this raises
+          # a warning on Python 3.10
+          - case-name: Python 3.10
+            os: ubuntu-latest
+            python-version: "3.10"
+            condaforge: 1
```
hodgestar, line 75: The conda-forge numpy doesn't seem to find / use mkl even if the mkl blas library is installed (or at least qutip.settings.has_mkl remains False and numpy's config has blas set to Generic). I'm don't know if this is expected or not. Perhaps we should explicitly specify `nomkl: 1` here to make that clear.

Ericgig (approved):
Look good.
The coverall version fix should be merged quickly as it is quite annoying to have all our tests hangs.
Python 3.6 is no longer supported so it could be removed from the test matrix in this PR.

PR #1783, opened 2022-01-25
Extra check for shape in Qobj to fix #1782

### LINKED ISSUE #1782: Segmentation fault in mesolve if density matrix was constructed with flat aray
### Bug Description

Construct a density matrix, via some array creation method (in this case numpy - haven't checked with others). Create a `rho0` for mesolve to use with the flat array while explicitly specifying the matrix dimensions in the `dims` parameter. Upon passing the density matrix to `mesolve` a segmentation fault occurs. A simple proof of concept is included below. `np.random.random` was used to fill the matrix in this case, but the issue was encountered using valid density matricies, and the issue seems to be unrelated to whether the matrix is valid or not as a density matrix.

### Code to Reproduce the Bug

```shell
import qutip as qtip
import numpy as np

dim = 100 # Small dims don't work. I assume python has some buffer after the array
H = qtip.Qobj(np.identity(dim),dims=[[dim],[dim]])
rho0 = np.random.random([dim**2])
rho0 = qtip.Qobj(rho0.flatten(),dims=[[dim],[dim]]) # This fails
# rho0 = qtip.Qobj(rho0.reshape([dim,dim]),dims=[[dim],[dim]]) # This works
qtip.mesolve(rho0=rho0,H=H,tlist=np.linspace(0,10,1),progress_bar=qtip.ui.EnhancedTextProgressBar())
```

### Code Output

```shell
[          0%           ] Elapsed 0.00s / Remaining 00:00:00:00[1]    3281154 segmentation fault (core dumped)
```

### Expected Behaviour

The code should exit without failure or throw a python error elaborating on the issue

### Your Environment

```shell
QuTiP Version:      4.6.2
Numpy Version:      1.21.4
Scipy Version:      1.6.0
Cython Version:     0.29.21
Matplotlib Version: 3.4.3
Python Version:     3.9.7
Number of CPUs:     4
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
Installation path:  /home/lala5th/.local/lib/python3.9/site-packages/qutip
```

### Additional Context

I ran the code through `gdb` to somewhat narrow down the cause of the issue. It seems that `dense2D_to_fastcsr_fmode` causes the segmentation fault, but the parameters passed to it seem fine at first glance, since I expect `ncols = 10000`. While I don't know how useful it is I will put the top of the stack trace (after this the only the call to `mesolve` is relevant in the stack trace).

```
#0  __pyx_pf_5qutip_2cy_9spconvert_6dense2D_to_fastcsr_fmode (__pyx_self=<optimised out>, __pyx_v_ncols=10000, __pyx_v_nrows=<optimised out>, __pyx_v_mat=...) at qutip/cy/spconvert.cpp:9472
#1  __pyx_pw_5qutip_2cy_9spconvert_7dense2D_to_fastcsr_fmode (__pyx_self=<optimised out>, __pyx_args=<optimised out>, __pyx_kwds=<optimised out>) at qutip/cy/spconvert.cpp:9237
#2  0x000000000053a8eb in cfunction_call (func=<built-in function dense2D_to_fastcsr_fmode>, args=<optimised out>, kwargs=<optimised out>) at ../Objects/methodobject.c:543
#3  0x000000000051c24b in _PyObject_MakeTpCall (tstate=0x9561b0, callable=<built-in function dense2D_to_fastcsr_fmode>, args=<optimised out>, nargs=<optimised out>, keywords=<optimised out>) at ../Objects/call.c:191
#4  0x0000000000516021 in _PyObject_VectorcallTstate (kwnames=0x0, nargsf=<optimised out>, args=0x1c16930, callable=<built-in function dense2D_to_fastcsr_fmode>, tstate=<optimised out>) at ../Include/cpython/abstract.h:116
#5  _PyObject_VectorcallTstate (kwnames=0x0, nargsf=<optimised out>, args=0x1c16930, callable=<built-in function dense2D_to_fastcsr_fmode>, tstate=<optimised out>) at ../Include/cpython/abstract.h:103
#6  PyObject_Vectorcall (kwnames=0x0, nargsf=<optimised out>, args=0x1c16930, callable=<built-in function dense2D_to_fastcsr_fmode>) at ../Include/cpython/abstract.h:127
#7  call_function (kwnames=0x0, oparg=<optimised out>, pp_stack=<synthetic pointer>, tstate=0x9561b0) at ../Python/ceval.c:5075
#8  _PyEval_EvalFrameDefault (tstate=<optimised out>, f=<optimised out>, throwflag=<optimised out>) at ../Python/ceval.c:3518
#9  0x000000000050f5e9 in _PyEval_EvalFrame (throwflag=0, 
    f=Frame 0x1c166f0, for file /home/lala5th/.local/lib/python3.9/site-packages/qutip/mesolve.py, line 779, in _generic_ode_solve (func=<built-in method mul_vec of qutip.cy.cqobjevo.CQobjCte object at remote 0x7fffd42248e0>, ode_args=(), rho0=<Qobj(_isherm=None, _type='oper', superrep=None, _isunitary=None, _data=<fast_csr_matrix(data=<numpy.ndarray at remote 0x7ffff3f918d0>, indices=<numpy.ndarray at remote 0x7ffff3f91570>, indptr=<numpy.ndarray at remote 0x7ffff3f91030>, _shape=(10000, 1), maxprint=50, format='csr') at remote 0x7fffe022a700>, dims=[[100], [100]]) at remote 0x7ffff7180df0>, tlist=<numpy.ndarray at remote 0x7ffff3f91930>, e_ops=[], opt=<Options(atol=<float at remote 0x7fffd51cfb70>, rtol=<float at remote 0x7fffd51cfb10>, method='adams', nsteps=1000, first_step=0, min_step=0, max_step=0, order=12, average_states=False, average_expect=True, ntraj=500, seeds=None, tidy=True, rhs_with_state=False, rhs_reuse=False, rhs_filename=None, mc_corr_eps=<float at remote 0x7fffd51cf9f0>, num_cpus=<numpy.int64 a...(truncated), tstate=0x9561b0) at ../Include/internal/pycore_ceval.h:40
```

Lala5th:
**Description**
Extra check in `Qobj` constructor to make sure that the underlying data behaves as expected.

**Related issues or PRs**
The PR is a fix to #1782, where the `Qobj` would be constructed in a way subsequent code (i.e. `mesolve`) did not expect causing a segmentation fault.

**Changelog**
Added a check in `Qobj` constructor that the respective members of `data.shape` cannot be larger than what the corresponding `dims` could contain.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -310,11 +310,6 @@ class Qobj(object):
             )
             self.dims = [[int(inpt.shape[0])], [int(inpt.shape[1])]]
 
-        if self._data.shape[0] > np.prod(self.dims[0]) or \
-           self._data.shape[1] > np.prod(self.dims[1]):
-            raise ValueError(f"Qobj has smaller dims {self.dims} " +
-                             f"than underlying shape {self._data.shape}")
-
         if type == 'super':
             # Type is not super, i.e. dims not explicitly passed, but oper-like
             # shape.
@@ -332,6 +327,13 @@ class Qobj(object):
         else:
             if self.type == 'super' and self.superrep is None:
                 self.superrep = 'super'
+
+        if (self._data.shape[0] > np.prod(np.hstack(self.dims[0])) or
+           self._data.shape[1] > np.prod(np.hstack(self.dims[1]))) and \
+           self.type != 'super':
+
+            raise ValueError(f"Qobj has smaller dims {self.dims} " +
+                             f"than underlying shape {self._data.shape}")
 
         # clear type cache
         self._type = None
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -328,6 +328,10 @@ class Qobj(object):
             if self.type == 'super' and self.superrep is None:
                 self.superrep = 'super'
 
+        # While the obvious check would be != that would fail valid
+        # use cases such as enr_fock and other enr_ functions.
+        # This does leave open the possibility of data still being
+        # misused such as Qobj(complex[n**2][1], dims = [[n],[n]])
         if (self._data.shape[0] > np.prod(np.hstack(self.dims[0])) or
            self._data.shape[1] > np.prod(np.hstack(self.dims[1]))) and \
            self.type != 'super':
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -310,6 +310,11 @@ class Qobj(object):
             )
             self.dims = [[int(inpt.shape[0])], [int(inpt.shape[1])]]
 
+        if self._data.shape[0] > np.prod(self.dims[0]) or \
+           self._data.shape[1] > np.prod(self.dims[1]):
+            raise ValueError(f"Qobj has smaller dims {self.dims} " +
+                             f"than underlying shape {self._data.shape}")
+
         if type == 'super':
             # Type is not super, i.e. dims not explicitly passed, but oper-like
             # shape.
--- a/qutip/tests/test_bofin_solvers.py
+++ b/qutip/tests/test_bofin_solvers.py
@@ -179,13 +179,13 @@ class TestHierarchyADOsState:
 
     def mk_rho_and_soln(self, ados, rho_dims):
         n_ados = len(ados.labels)
-        ado_soln = np.random.rand(n_ados, *[2**d for d in rho_dims])
-        rho = Qobj(ado_soln[0, :], dims=[2, 2])
+        ado_soln = np.random.rand(n_ados, *[np.product(d) for d in rho_dims])
+        rho = Qobj(ado_soln[0, :], dims=rho_dims)
         return rho, ado_soln
 
     def test_create(self):
         ados = self.mk_ados([2, 3], max_depth=2)
-        rho, ado_soln = self.mk_rho_and_soln(ados, [2, 2])
+        rho, ado_soln = self.mk_rho_and_soln(ados, [[2], [2]])
         ado_state = HierarchyADOsState(rho, ados, ado_soln)
         assert ado_state.rho == rho
         assert ado_state.labels == ados.labels
@@ -195,7 +195,7 @@ class TestHierarchyADOsState:
 
     def test_extract(self):
         ados = self.mk_ados([2, 3], max_depth=2)
-        rho, ado_soln = self.mk_rho_and_soln(ados, [2, 2])
+        rho, ado_soln = self.mk_rho_and_soln(ados, [[2], [2]])
         ado_state = HierarchyADOsState(rho, ados, ado_soln)
         ado_state.extract((0, 0)) == rho
         ado_state.extract(0) == rho


qutip/qobj.py
```diff
@@ -328,6 +328,13 @@ def __init__(self, inpt=None, dims=None, shape=None,
             if self.type == 'super' and self.superrep is None:
                 self.superrep = 'super'
 
+        if (self._data.shape[0] > np.prod(np.hstack(self.dims[0])) or
+           self._data.shape[1] > np.prod(np.hstack(self.dims[1]))) and \
```
hodgestar, line 332: I'm a bit surprised that the check is whether the shape `>` to the product of the dims, rather than simply `!=`. Is there a situation where being `<` the dims is okay?

Lala5th: Ericgig pointed out that `enr_` functions return data where `<` is OK. The `>` check allows these cases. This does however leave open the possibility of misshapen data where this condition is technically fulfilled i.e. `Qobj(complex[n,1],dims=[[n],[n]])`. This seems however a very niche and possibly from the start an erroneous case.

hodgestar: We should definitely at least at a comment about this here.

I am wondering if we can do something better (especially in QuTiP 5).

@Ericgig Any thoughts?

Ericgig: This is something we definitively need to take into account when making the dims class in v5 and a reason we need to do it sooner than later. For v4, a comment could be enough. Or we could have the constructor being strict and then 3~4 special functions to work around it.

hodgestar: @Lala5th Could you add the comment here and then I'm happy to merge this?
@Ericgig If you could approve too, that would be good (since the QObj constructor is one of the most used parts of the code)

Lala5th: @hodgestar Is this what you had in mind? Some checks (2) unfortunately should still fail, but I couldn't find proper documentation on those to figure out if they are valid use cases or some ill formed code that has not been noticed until now. (The two checks are the ones relating to `TestHierarchyADOsState` in `test_bofin_solvers.py`)

hodgestar: That looks great -- thanks!

hodgestar: I can look at the HEOM / BoFiN test failures. Quite possibly I just got some dims or shapes wrong in those tests.

hodgestar:
@Lala5th Would you mind merging master into this so that the test suite can run again (I had to pin coverage to version 6.2 while coverage 6.3 is being fixed).

hodgestar (approved):
We just need CI to run and to fix any failing tests.

Ericgig (approved):
I think it's good enough for now.
It's certainly better than the no check we had, but I would like to have something stricter eventually.

hodgestar:
I fixed the BoFiN test, but I'm a little bit worried that it might be common for people to accidentally write `dims=[2, 2]` instead of `dims=[[2], [2]]`, for example, and that with the call to `np.hstack(dims[0])` this will now raise a TypeError, because `np.hstack` must be called with a list.

I.e. we are also forcing `dims[0]` and `dims[1]` to be lists, which is not something we checked before.

Lala5th:
@hodgestar I agree that that is a problem and a check for that was something I considered adding, but ultimately decided against it as I wanted to make the check very small. Something like this should solve the problem though:

```diff
diff --git a/qutip/qobj.py b/qutip/qobj.py
--- qutip/qobj.py
+++ qutip/qobj.py
@@ -327,8 +327,14 @@
         else:
             if self.type == 'super' and self.superrep is None:
                 self.superrep = 'super'
 
+        if isinstance(self.dims[0], (int, np.integer)):
+            self.dims[0] = [self.dims[0]]
+
+        if isinstance(self.dims[1], (int, np.integer)):
+            self.dims[1] = [self.dims[1]]
+
         # While the obvious check would be != that would fail valid
         # use cases such as enr_fock and other enr_ functions.
         # This does leave open the possibility of data still being
         # misused such as Qobj(complex[n**2][1], dims = [[n],[n]])
```

Ericgig:
Every `Qobj` created by qutip have `dims[0]` and `dims[1]` being list. `dims=[2, 2]` is an error that we should catch. The main issue is that there is no clear documentation or tests that clearly says what is a valid dims format. 

There are some strange dimension that are produced by our function that would raise an error with `np.stack` like
`tensor(operator_to_vector(oper), oper)`  but I don't think we should accept these kind of `Qobj` in the first place, so I am fine failing there.

hodgestar:
Given that this likely prevents many mistakes, I'm okay with merging it as is and if things break too badly for many users, we can revert or fix part of it and do a new release.

PR #1784, opened 2022-01-25
Cy.shape.check

Ericgig:
**Description**
Add check for shape at the cython level. While these checks should be done before entering cython in QobjEvo or the solvers, as seen in #1782, some bad shape can slip through. 

**Changelog**
Add check for shape in cython qobjevo.

### DIFF
--- a/qutip/cy/cqobjevo.pyx
+++ b/qutip/cy/cqobjevo.pyx
@@ -204,6 +204,9 @@ cdef class CQobjEvo:
         return 0.
 
     def mul_vec(self, double t, complex[::1] vec):
+        if vec.shape[0] != self.shape1:
+            raise ValueError(f"Shape don't match: ({self.shape0}, "
+                             f"{self.shape1}) @ ({vec.shape[0]}, 1) ")
         cdef np.ndarray[complex, ndim=1] out = np.zeros(self.shape0,
                                                         dtype=complex)
         self._mul_vec(t, &vec[0], &out[0])
@@ -214,6 +217,11 @@ cdef class CQobjEvo:
         cdef unsigned int sp_rows = self.shape0
         cdef unsigned int nrows = mat.shape[0]
         cdef unsigned int ncols = mat.shape[1]
+        if mat.shape[0] != self.shape1:
+            raise ValueError("Shape don't match: "
+                f"({self.shape0}, {self.shape1}) "
+                f"@ ({mat.shape[0]}, {mat.shape[1]})")
+
         if mat.flags["F_CONTIGUOUS"]:
             out = np.zeros((sp_rows,ncols), dtype=complex, order="F")
             self._mul_matf(t, &mat[0,0], &out[0,0], nrows, ncols)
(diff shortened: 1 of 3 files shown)


qutip/cy/cqobjevo.pyx
```diff
@@ -204,6 +204,9 @@ cdef class CQobjEvo:
         return 0.
 
     def mul_vec(self, double t, complex[::1] vec):
+        if vec.shape[0] != self.shape1:
+            raise ValueError(f"Shape don't match: ({self.shape0}, "
```
hodgestar, line 208:
```suggestion
            raise ValueError(f"Shapes don't match: ({self.shape0}, "
```

qutip/cy/cqobjevo.pyx
```diff
@@ -214,6 +217,11 @@ cdef class CQobjEvo:
         cdef unsigned int sp_rows = self.shape0
         cdef unsigned int nrows = mat.shape[0]
         cdef unsigned int ncols = mat.shape[1]
+        if mat.shape[0] != self.shape1:
+            raise ValueError("Shape don't match: "
```
hodgestar, line 221:
```suggestion
            raise ValueError("Shapes don't match: "
```

qutip/cy/cqobjevo.pyx
```diff
@@ -224,20 +232,32 @@ cdef class CQobjEvo:
 
     cpdef complex expect(self, double t, complex[::1] vec):
         if self.super:
+            if self.shape1 != vec.shape[0]:
+                raise ValueError(f"Shape don't match: ({self.shape0}, "
```
hodgestar, line 236:
```suggestion
                raise ValueError(f"Shapes don't match: ({self.shape0}, "
```

qutip/cy/cqobjevo.pyx
```diff
@@ -224,20 +232,32 @@ cdef class CQobjEvo:
...
+                                 f"{self.shape1}) @ ({vec.shape[0]}, 1) ")
```
hodgestar, line 237: The `@` sign in the error messages here suggests that the error occurred during multiplication, but it actually occurred during a call to expect. It would be good if the error message reflected where the error occurred.

We should also remove the space at the end of the string.

qutip/cy/cqobjevo.pyx
```diff
@@ -224,20 +232,32 @@ cdef class CQobjEvo:
...
             return self._expect_super(t, &vec[0])
         else:
+            if self.shape1 != vec.shape[0] and self.shape0 != self.shape1:
+                raise ValueError(f"Shape don't match: ({self.shape0}, "
+                                 f"{self.shape1}) @ ({vec.shape[0]}, 1) ")
             return self._expect(t, &vec[0])
 
     def overlapse(self, double t, complex[::1] oper):
```
hodgestar, line 245: Should this method be called `overlap` or `overlaps`? `overlapse` is not an English word (although it is a very cool made up word). Apologies for not noticing this in earlier reviews. Maybe let's fix it in a separate PR to this one.

hodgestar: Or maybe "overlapse" is a technical QM term I'm not familiar with?

hodgestar (changes requested):
Thanks for adding these. I did wonder a bit whether the checks should happen one level down (e.g. inside `_expect_super`) where the shapes are actually relied on, but maybe I like them more where you put them (i.e. at the top of the public interface).

qutip/cy/cqobjevo.pyx
Ericgig: No it's a spelling error.

hodgestar:
@Ericgig Could you merge master into this so that we can check the Cython changes on all the various Pythons? I don't specifically expect a problem, but for this PR it seems good to run the checks.

hodgestar (approved):
Happy with the changes.

Do you think it is okay to rename `.overlap` in 4.6.3? Or is it public enough that the rename should be left for 4.7? I marked it as for 4.6.3 for now.

Ericgig:
I think the renaming is fine for 4.6.3, it is called by `expect` of the python interface, but the cython interface is not expected to be used by the users.

PR #1785, opened 2022-01-26
Fixed dim_H in test_k

Lala5th:
**Description**
A change in `test_k`. In commit 7173a0d `L` was changed, but this was not reflected in `dim_H`. This commit aims to rectify that.

**Related issues or PRs**

**Changelog**
Changed `dim_H` to include `L` in its definition in lattice tests.

### DIFF
--- a/qutip/tests/test_lattice.py
+++ b/qutip/tests/test_lattice.py
@@ -209,7 +209,7 @@ class TestLattice1d:
                 else:
                     kop[row, col] = 1 / (np.exp(2j * np.pi * (row - col)/L)-1)
         kt = np.kron(kop * 2 * np.pi / L, np.eye(6))
-        dim_H = [[2, 2, 3], [2, 2, 3]]
+        dim_H = [[L, 2, 3], [L, 2, 3]]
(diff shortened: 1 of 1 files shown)


qutip/tests/test_lattice.py
```diff
@@ -209,7 +209,7 @@ def test_k(self):
                 else:
                     kop[row, col] = 1 / (np.exp(2j * np.pi * (row - col)/L)-1)
         kt = np.kron(kop * 2 * np.pi / L, np.eye(6))
-        dim_H = [[2, 2, 3], [2, 2, 3]]
+        dim_H = [[L, 2, 3], [L, 2, 3]]
```
hodgestar, line 212: So this didn't fail before because no operations that checked the dims were used? And it fails now if #1783 is merged?

Lala5th: That is my interpretation, yes.

hodgestar (approved):

hodgestar:
@Lala5th Thank you for the test fix.

PR #1787, opened 2022-01-27
add timeout

Ericgig:
**Description**
Since yesterday, our tests hangs for both `master` and `dev.major` and never finish. It seen only a few tests are affected.
I am adding a timeout so the tests that hangs result in a localized error.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -92,7 +92,7 @@ jobs:
             conda install nomkl "numpy${{ matrix.numpy-requirement }}" "scipy${{ matrix.scipy-requirement }}"
           fi
           python -m pip install -e .[$QUTIP_TARGET]
-          python -m pip install pytest-cov coveralls
+          python -m pip install pytest-cov coveralls pytest-timeout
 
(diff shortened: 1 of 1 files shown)


hodgestar:
The timeouts are because of a GitHub actions outage. It appears to have been resolved earlier today -- https://www.githubstatus.com/incidents/29xnrws671f0.

Ericgig:
This run will still timeout, as it was started before the fix.
Hopefuly it will work on a rerun.

Ericgig:
Still failing... I will retry tomorrow in case it just take time to apply the fix.

hodgestar:
Maybe macos needs a slightly longer timeout? It is a bit slow.

.github/workflows/tests.yml
```diff
@@ -145,6 +145,8 @@ jobs:
           #  --strict-markers
           #     error out if a marker is used but not defined in the
           #     configuration file
+          #  --timeout=150
```
hodgestar, line 148:
```suggestion
          #  --timeout=300
```

PR #1788, opened 2022-01-31
Reduce average gate fidelity to process fidelity

fhopfmueller:
This PR reduces the calculation of average gate fidelity between two quantum channels to the recently merged implementation of process fidelity.

Additional minor change: raise an error when the channel passed to `process_fidelity` does not preserve the Hilbert space dimension of the input state. Previously it gave inconsistent results for that case, for different representations of equivalent channels. It's probably possible to define the `process_fidelity` and `average_gate_fidelity` for non-dimension-preserving channels if there is a use case for it? The previous version of `average_gate_fidelity` also was not defined for non-dimension-preserving channel.

**Related issues or PRs**
Last part of the changes proposed in https://github.com/qutip/qutip/issues/1703. Builds on the implementation of process_fidelity from https://github.com/qutip/qutip/pull/1712, https://github.com/qutip/qutip/pull/1748.

**Changelog**
Re-implement average_gate_fidelity using process_fidelity.


hodgestar:
@fhopfmueller If you can merge dev.major into this branch and resolve the conflicts, I can review (I think the conflicts are just with the fixes that you made that got merged into 4.7).

hodgestar:
@fhopfmueller I merged dev.major into your branch, but I ended up dropping most of the `avg_gate_fidelity_test` since the `avg_gate_fidelity_test` decorator didn't seem to be included in the PR commits. Do you know what happened to them? Should I add them back? And if so, where is the decorator implemented?

fhopfmueller:
Thanks a lot for awakening this PR from its slumber @hodgestar !

All the decorator `avg_gate_fidelity_test` does is to skip the decorated test on Mac. Looks like it has been removed in the following commit which is included in your merge, in favour of refactoring the AGF tests into a class and skipping the whole class on Mac: https://github.com/qutip/qutip/commit/99ce2d5c511e2487505a79455c9b53f7676a5731#diff-a9cd5a7620c34d60ffb61360a78bab270214a212c997e0c2cf5a6617443a90ab
(The decorator was originally introduced in https://github.com/qutip/qutip/pull/1034. The underlying issue https://github.com/qutip/qutip/issues/963 seems to be resolved, so maybe skipping the tests on Mac isn't necessary any more?)

The only test that is included in my original PR, but not after your merge, is `test_average_gate_fidelity_against_legacy_implementation` at https://github.com/fhopfmueller/qutip/blob/74080688cbece61a3d095fa6fc9dd347114b6152/qutip/tests/core/test_metrics.py#L443

Would you like to include that test? On the one hand, it contains copy-pasted code from the previous implementation, so the test isn't exactly elegant. On the other hand, it ensures that the new implementation of this PR gives the same result as the previous correct implementation. I'd favor including it.

hodgestar:
@fhopfmueller Thanks for the help! I've added back the test. I also made a separate commit to remove skipping the tests on Mac OS. Let's see what happens.

hodgestar (approved):

hodgestar:
Tests all pass on Mac OS now & the changes look good. Merging. Thanks @fhopfmueller.

PR #1791, opened 2022-02-02
Update dev.major CI tests setup from master.

hodgestar:
**Description**
Update dev.major CI tests setup from master.

This also merges the build changes needed to support Python 3.10 (which is tested in the latest CI test setup from master).

**Related issues or PRs**
* Brings in: #1777

**Changelog**
Update dev.major CI tests setup from master.


qutip/solve/piqs.py
```diff
@@ -98,6 +98,18 @@
...
+
+def _ensure_int(x):
+    """
+    Ensure that a floating-point value `x` is exactly an integer, and return it
+    as an int.
+    """
+    out = int(x)
+    if out != x:
+        raise ValueError(f"{x} is not an integral value")
+    return out
```
Ericgig, line 110: Not a fan, there should be a tolerance when comparing an int and float. 
And I think the test should be in the function calling this, not here, this will make the error harder to trace.

Ericgig: Ok, let's just leave keep in mind that piqs will need some attention eventually.

hodgestar:
@Ericgig I pushed a fix for the Cython compilation issue that works for me on Python 3.10 locally. Let me know what you think.

Ericgig:
Look good.

PR #1794, opened 2022-02-03
Migrate countstat to solver.

Ericgig:
**Description**
`countstat.py` does not depend on solver, however it has a **F** maintainability score on codeclimate.
This clean the code to make it more readable, running codeclimat locally, there are not issues left.
Also added a call using the pseudo inverse method in the test.

**Changelog**
Make `countstat.py`'s code more readable.


hodgestar:
@Ericgig If you can resolve conflicts, I will review this since.

qutip/solver/countstat.py
```diff
@@ -0,0 +1,225 @@
...
+        wlist = [0.]
+
+    J_ops = [op.data for op in J_ops]
+
+    if not sparse or method != 'direct':
+        current, noise = _noise_pseudoinv(L, wlist, rhoss, J_ops,
+                                          sparse, method)
+    else:
+        rhoss_vec = operator_to_vector(rhoss).data
+        current, noise = _noise_direct(L, wlist, rhoss_vec, J_ops)
```
hodgestar, line 223:
```suggestion
    if sparse and method == 'direct':
        rhoss_vec = operator_to_vector(rhoss).data
        current, noise = _noise_direct(L, wlist, rhoss_vec, J_ops)
    else:
        current, noise = _noise_pseudoinv(L, wlist, rhoss, J_ops,
                                          sparse, method)
```
Removing the double negative makes the code much clearer for me.

hodgestar: I would also pass `rhoss` and `J_ops` as Qobj directly and not data instances, since L is a Qobj. I also wouldn't apply `operator_to_vector` here and just apply it inside `_noise_direct` if needed. It's better if even internally function mostly take all Qobj or all data instances.

qutip/solver/countstat.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def _solve(A, V):
+    try:
+        if settings.has_mkl:
+            out = mkl_spsolve(A.tocsc(), V)
+        else:
+            A.sort_indices()
+            out = sp.linalg.splu(A, permc_spec='COLAMD').solve(V)
+    except Exception:
```
hodgestar, line 82: This exception is raises when the sparse methods are used on very small matrices. I wonder if we should just do an explicit check on the matrix size? Trying the sparse method and failing seems a bit ugly.

@nwlambert Any thoughts on how / whether to clean this up since you wrote the original code?

hodgestar: See 401076efb170 for @nwlambert's orignal commit message.

qutip/tests/solver/test_countstat.py
```diff
@@ -35,16 +36,26 @@ def test_dqd_current():
...
         current2 = qutip.countstat_current(L, c_ops, J_ops=J_ops)
         assert abs(current[n] - current2) < 1e-8
 
+        current2 = qutip.countstat_current(L, c_ops)[0]
+        assert abs(current[n] - current2) < 1e-8
+
+        current3, noise3 = qutip.countstat_current_noise(L, c_ops, wlist=[0,1],
+                                                         sparse=False)
+        assert abs(current_[0] - current3[0]) < 1e-6
+        np.testing.assert_allclose(noise_[0, 0, :], noise3[0, 0, :], atol=1e-6)
```
hodgestar, line 57: Could we sort out the variable names and types a bit? It looks we're comparing numpy arrays, e.g. `current2` and scalars, `e.g. current[n]`, and we have `current_`, but `current3` and `current2`. Maybe just `current_1`, `current_2` and `current_3` and always have each be the same type of object?

hodgestar (changes requested):
Thanks for cleaning up this file. I can see why the code climate score was so low!

I left some suggestions and also pinged @nwlambert since he worked on this code ages ago and fixed numerous mathematical issues and errors.

qutip/solver/countstat.py
nwlambert: I did a quick check, and yeah the old example i had in mind still fails sometimes without those exceptions to catch it.  It fails on a 2-dimensional system, so a Liouvillian with shape (4,4),  but is ok with 3-dimensions (shape (9,9)).  I guess other option is just to default to using lsqr() for the sparse case?

hodgestar: Hmm. Computing the determinant of a large sparse matrix efficiently is hard, although there are some ideas mentioned here -- https://stackoverflow.com/questions/19107617/how-to-compute-scipy-sparse-matrix-determinant-without-turning-it-to-dense. I guess this is why `scipy.sparse.linalg.det` does not exist.

I'm okay with leaving this as is for now, but it would be good perhaps to leave the problem to the user by using a fast method and raising an error suggestion another method if the matrix is singular (and a way to specify that other method).

A similar problem arises in the HEOM steady_state function, so nice solution here could be used elsewhere in QuTiP too.

hodgestar (approved):

qutip/solver/countstat.py
Ericgig: `steadystate.py` is also full of multiple methods to solve this kind of problem. 
A `solve` method for `Qobj` could be nice...

hodgestar: That's a good idea. Let's add it to our v5 todo list. Maybe even at the data layer?

hodgestar: I added adding `solve` to the v5 alpha list.

PR #1797, opened 2022-02-07
Ignore deprecationg warnings from SciPy 1.8.0 scipy.sparse.X imports in CI tests.

hodgestar:
**Description**
Ignore deprecationg warnings from SciPy 1.8.0 scipy.sparse.X imports in CI tests.

**Related issues or PRs**
- None

**Changelog**
Ignore deprecationg warnings from SciPy 1.8.0 scipy.sparse.X imports in CI tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,11 +69,13 @@ jobs:
           # Ignore ImportWarning because pyximport registered an importer
           # PyxImporter that does not have a find_spec method and this raises
           # a warning on Python 3.10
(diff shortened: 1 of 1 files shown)


Ericgig:
Why is scipy 1.8 only used by the py3.10?
I expected python3.9 defaults to use the latest scipy.
Doing it like this, the deprecation warning will be visible to the users, so we may have issues complaining about it but since 4.7 is coming soon, I think it's fine.

Ericgig (approved):

PR #1798, opened 2022-02-07
Fix for scipy 1.8 in dev.major

Ericgig:
**Description**
Fix for scipy 1.8 with dev.major:
One class is now in a private file, but still present.

**Changelog**
Fix for scipy 1.8


hodgestar (approved):
I guess this is not accessing anything more private than it was before, so +1.

hodgestar (approved):

PR #1802, opened 2022-02-10
Improve weighting in steadystate solver.

### LINKED ISSUE #1512: Wrong default weight parameter in steadystate function
**Issue**
In the codebase, the weight for setting the unit trace condition in the steadystate function should be set according to the documentation to the average value of the (nonzero) elements in the Liouvillian L, if not specified explicitly.

However in the code the weight is set as 
```python
 ss_args['weight'] = np.mean(np.abs(A.data.data.max()))
```
which is clearly different as it just set to the maximum value in the Liouvillian matrix A and hence the mean and abs are meaningless.

 I think this could be an overlooked mistake, or else just a mistake in the documentation of the parameter.

hodgestar:
**Description**
Improve weighting in steadystate solver. Previously the default weighting did not match the documented behaviour and the sparse and dense solvers applied the weighting in different ways. Now the default matches the documentation and the sparse and dense solvers apply the weighting in the same manner.

**Related issues or PRs**
- Supersedes #1275
- Fixes #1512

**Changelog**
Improve the weighting in steadystate solver, so that the default weight matches the documented behaviour and the dense solver applies the weights in the same manner as the sparse solver.

### DIFF
--- a/qutip/steadystate.py
+++ b/qutip/steadystate.py
@@ -555,8 +555,7 @@ def _iterative_precondition(A, n, ss_args):
                   fill_factor=ss_args['fill_factor'],
                   options=dict(ILU_MILU=ss_args['ILU_MILU']))
 
-        P_x = lambda x: P.solve(x)
-        M = LinearOperator((n ** 2, n ** 2), matvec=P_x)
+        M = LinearOperator((n ** 2, n ** 2), matvec=P.solve)
         _precond_end = time.time()
         ss_args['info']['permc_spec'] = ss_args['permc_spec']
         ss_args['info']['drop_tol'] = ss_args['drop_tol']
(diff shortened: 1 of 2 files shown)


hodgestar:
@davidschlege @fminga -- if you one or both of you have a moment to review these changes it would be appreciated.

Ericgig (approved):

davidschlegel:
> @davidschlegel @fminga -- if you one or both of you have a moment to review these changes it would be appreciated.

It looks okay to me (at least, it makes sense now), but I haven't used it for a while. The entire point of the weight is to set how much the trace condition is enforced, so the change makes sense to me. However, to my knowledge, there is little literature on how to do this optimally, but it seems to be a good rule of thumb.

A possible addition could be to check the trace condition of the steady-state solution and throw a warning if it is violated more than some epsilon. But for now, everything looks good.

hodgestar:
> A possible addition could be to check the trace condition of the steady-state solution and throw a warning if it is violated more than some epsilon. But for now, everything looks good.

@davidschlegel That's a good idea, so I added some assertions to all the tests where it was simple to do so.

PR #1805, opened 2022-02-16
Linblad fix

AGaliciaMartinez:
**Checklist**
- [x] PEP8

**Description**
The behaviour of the `liouvillian` seemed to be wrong. For a `TfTensor` as input it returned a `Dense` as oupu. This is mainly because the identity matrix created within the function was not a `TfTensor`. 

_Testing The tests that I included where would pass without the fix!_ 

Reproducing this behaviour within QuTiP seems to be challenging. This is because of how things are automatically converted from one data type to another. In particular, `liouvillian` worked as expected for both `Dense` and `CSR` but not for `TfTensor`. However, since `TfTensor` is not part of qutip I am not sure how to make it to reproduce this error. 

The issue seems in fact to be related to an inconsistency in how data is automatically transformed. The following code summarises the problem:
```
H = qt.rand_herm(5)
csr = H.to('csr').data
dense = H.to('dense').data
tftensor = H.to('tftensor').data

qutip.data.kron(csr, dense) # output: dense
qutip.data.kron(csr, csr) # output: csr
qutip.data.kron(csr, tftensor) # output: dense?? this is what causes problems!
qutip.data.kron(tftensor, tftensor) # output: tftensor
```

**Changelog**
Fixed liouvillian to not generate dense output when used with TfTensor.

### DIFF
--- a/qutip/core/superoperator.py
+++ b/qutip/core/superoperator.py
@@ -130,7 +130,7 @@ def liouvillian(H=None, c_ops=None, data_only=False, chi=None):
     op_shape = H.shape
     sop_dims = [[op_dims[0], op_dims[0]], [op_dims[1], op_dims[1]]]
     sop_shape = [np.prod(op_dims), np.prod(op_dims)]
-    spI = _data.identity(op_shape[0])
+    spI = _data.identity(op_shape[0], dtype=type(H.data))
 
     data = _data.mul(_data.kron(spI, H.data), -1j)
     data = _data.add(data, _data.kron(H.data.transpose(), spI), scale=1j)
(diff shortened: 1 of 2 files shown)


Ericgig (approved):
Good find. 
`kron(csr, tftensor) -> dense` is certainly strange...

hodgestar (approved):

AGaliciaMartinez:
I am happy to merge it as it is but note that _the tests do not serve to catch the bug!_ I am not sure how to test it properly though.

hodgestar:
Let's merge this fix for now.

If there is a deeper issue, it's around how dispatch works for specializations with an output dtype. While it's possible we can improve that, the fix here is likely the correct one -- i.e. to create an identity with the same dtype as the input Hamiltonian.

PR #1807, opened 2022-02-16
Remove the check on the initial state dimension

### LINKED ISSUE #1803: Unable to add 2-qudit gates to a quantum circuit of 4-level qudits.
### Describe the Issue!

I'm having issues with running a quantum circuit that consists of 4-level qudits. I want to define my custom 2-qudit gate.

I have written a gate function that returns the following object:

`return Qobj(mat, dims=[[4,4],[4,4]], shape = [16, 16], type='oper')`

Since this is a gate acting on two 4-level qudits, I think that these dims are correct.

I then create an initial ket state for my circuit, let's say with three qudits, and create the circuit object:

```
zero_state=tensor(basis(4, 0),basis(4, 0),basis(4, 0))
qc = QubitCircuit(N=3, dims=zero_state.dims)
```

the dims are: [[4, 4, 4], [1, 1, 1]]

when I try to add a gate between qudits 0 and 1:

`qc.add_gate("Ucomm", targets=[0,1])`

I get the following error:

`ValueError: The operator dims [4, 4] do not match the target dims [[4, 4, 4], [1, 1, 1]].`

If I instead try to add a gate between qudits 1 and 2:
`qc.add_gate("Ucomm", targets=[1,2])`

I get:

```
targ_dims = [dims[t] for t in targets]
IndexError: list index out of range
```

Do I understand correctly how dims work in Qutip? Or is this an error? The above example works perfectly fine for qubits, replacing 4->2 in the above.

BoxiLi:
**Description**
Remove the wrong check on the initial state dimension. It assumes a qubit state. However, we also support multilevel systems such as qudits.

**Related issues or PRs**
fix #1803 

**Changelog**
Remove the wrong check on the initial state dimension in QubitCircuit.

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -2080,8 +2080,6 @@ class CircuitSimulator:
         self.state = None
 
         if state is not None:
-            if state.shape[0] != 2 ** self.qc.N:
-                raise ValueError("dimension of state is incorrect")
             if self.mode == "density_matrix_simulator" and state.isket:
                 self.state = ket2dm(state)
             else:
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Thanks, Boxi!

Issue #1184, opened 2020-02-22
More detailed information for GSoC 2020 project


BoxiLi:
I have some detail information for the GSoC 2020  project "Error mitigation in QuTiP". Unfortunately, I don't have writing access to the QuTiP GitHub Wiki page. I post it here so if anyone finding it nice could copy it to the Wiki page.

I add some details based on the original description:

-----------------------------------
## 1. Error mitigation in QuTiP
From the QuTiP 4.5 release, the qutip.qip module now contains the noisy quantum circuit simulator (which was a GSoC project) providing enhanced features for a pulse-level description of quantum circuits and noise models. A new class `Processor` and several subclasses are added to represent different platforms for quantum computing. They can transfer a quantum circuit into the corresponding control sequence and simulate the dynamics with QuTiP solvers. Different noise models can be added to `qutip.qip.noise` to simulate noise in a quantum device. 

This module is still young and many features can be improved, including new device models, new noise models and integration with the existing general framework for quantum circuits (`qutip.qip.circuit`). There are also possible applications such as error mitigation techniques [1-3].

The tutorial notebooks can be found at http://qutip.org/tutorials.html#nisq. A recent presentation on the FOSDEM conference may help you get an overview (https://fosdem.org/2020/schedule/event/quantum_qutip/). See also the Github Project page for a collection of related issues and ongoing Pull Requests.

### Expected outcomes
 - More devices defined in the `qutip.qip.device` module. At the moment, we have two models: spin chain and cavity QED. We would like to include some other commonly used planform such as Superconducting system, Ion trap system or silicon system. Each model will need a new set of control hamiltonian and a compiler that find the control pulse of a quantum gate. This part is gonna involve some physics and study of commonly used hardware platforms. The related code can be found in `qutip.qip.device` and `qutip.qip.compiler`
- Refactoring the circuit compiler to allow simple pulse scheduling, e.g. Restriction one pulses that can be turned on at the same time.
- Features to perform error mitigation techniques in QuTiP, such as zero-error extrapolation.
- APIs to allow import quantum circuits from other software packages or commonly used language e.g. through quantum assembly language (qasm)
----------------------------------------------------------------

Also, the last outcomes have some overlap with the last project. Although I also find this circuit importation very important, I guess two projects should not have overlap if they are both selected.

nathanshammah:
Hi Boxi,

Can you please modify the projects of the Wiki page as you deem best?
All: we can also think of changing the order of them. It may be that
[...]

PR #1813, opened 2022-02-18
Support Pathlike objects in fileio

labay11:
**Description**
- Adds support for pathlib.Path objects in qsave/qload.
- Adds tests for those functions.
- Suffix is optional, if given it won't be appended again.
- Removes prints in qload

**Related issues or PRs**
fix #1184

**Changelog**
`qsave` and `qload` now support `pathlib.Path` objects.

### DIFF
--- a/qutip/fileio.py
+++ b/qutip/fileio.py
@@ -225,7 +225,7 @@ def qsave(data, name='qutip_data'):
     if file.suffix != '.qu':
         file.with_suffix('.qu')
 
-    with open(name, "rb") as fileObject:
+    with open(name, "wb") as fileObject:
         # this writes the object a to the file named 'filename.qu'
         pickle.dump(data, fileObject)
 
--- a/qutip/fileio.py
+++ b/qutip/fileio.py
@@ -222,8 +222,7 @@ def qsave(data, name='qutip_data'):
     """
     # open the file for writing
     file = Path(name)
-    if file.suffix != '.qu':
-        file.with_suffix('.qu')
+    file = file.with_suffix(file.suffix + ".qu")
 
     with open(name, "wb") as fileObject:
         # this writes the object a to the file named 'filename.qu'
@@ -246,8 +245,7 @@ def qload(name):
 
     """
     file = Path(name)
-    if file.suffix != '.qu':
-        file.with_suffix('.qu')
+    file = file.with_suffix(file.suffix + ".qu")
 
     with open(name, "rb") as fileObject:
         if sys.version_info >= (3, 0):
--- a/qutip/fileio.py
+++ b/qutip/fileio.py
@@ -3,8 +3,7 @@ __all__ = ['file_data_store', 'file_data_read', 'qsave', 'qload']
 import pickle
 import numpy as np
 import sys
-from qutip.qobj import Qobj
-from qutip.solver import Result
+from pathlib import Path
 
 
 # -----------------------------------------------------------------------------
@@ -16,7 +15,7 @@ def file_data_store(filename, data, numtype="complex", numformat="decimal",
 
     Parameters
     ----------
-    filename : str
+    filename : str or pathlib.Path
         Name of data file to be stored, including extension.
     data: array_like
         Data to be written to file.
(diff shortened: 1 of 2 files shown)


qutip/fileio.py
```diff
@@ -217,15 +216,18 @@ def qsave(data, name='qutip_data'):
...
 
     """
     # open the file for writing
-    fileObject = open(name + '.qu', 'wb')
-    # this writes the object a to the file named 'filename.qu'
-    pickle.dump(data, fileObject)
-    fileObject.close()
+    file = Path(name)
+    if file.suffix != '.qu':
+        file.with_suffix('.qu')
```
hodgestar, line 226: Previously the code always added `.qu` regardless -- e.g. `foo.qu` would be saved as `foo.qu.qu` and `foo.txt` would be saved as `foo.txt.qu`. The new code saves `foo.qu` as `foo.qu` and `foo.txt` as `foo.qu`.

I'm not a fan of the old behaviour particularly, but the new behaviour of replacing the suffix isn't much better -- both are unnecessary complications that just make life harder for QuTiP users. Typing `qsave("foo.qu")` is more explicit and is less for the user to remember about the underlying function works.

I would suggest we keep the old behaviour for 4.7 (i.e. in this PR) -- e.g. `file = file.with_suffix(file.suffix + ".qu")`.

If you are interested, you could then open up a follow up PR for QuTiP 5 (the `dev.major` branch) and remove the suffix addition entirely there.

hodgestar, line 226: The `file.with_suffix(...)` doesn't replace the suffix of `file`. It returns a new path object which is then discarded because it isn't assigned to anything.

qutip/tests/test_fileio.py
```diff
@@ -44,11 +45,18 @@ def test_separator_detection(self, separator):
...
-def test_qsave_qload():
+@pytest.mark.parametrize('use_path', [True, False], ids=['Path', 'str'])
+@pytest.mark.parametrize('include_suffix', [True, False],
+                         ids=['with_suffix', 'no_suffix'])
+def test_qsave_qload(use_path, include_suffix):
     ops_in = [qutip.sigmax(),
               qutip.num(_dimension),
               qutip.coherent_dm(_dimension, 1j)]
     filename = _random_file_name()
+    if include_suffix:
```
hodgestar, line 56: If we remove the if statement that checks for `.qu` as I suggested above, then we can also delete the extra test case here. Although perhaps testing a few suffixes is not a bad idea. E.g. change `'include_suffix', [True, False]` to `'suffix', ['', '.qu', '.dat']`.

hodgestar (changes requested):
I made a few suggestions, but looks good.

qutip/fileio.py
labay11: I have brought back the functionality and will open a pull request for v5.

qutip/tests/test_fileio.py
labay11: Testing for different suffixes is a good idea, the code can then be reused later in v5.

hodgestar (approved):

hodgestar:
@labay11 Thank you! I've approved the changes and marked them for inclusion in 4.7. I'll merge once the test runs have succeeded.

qutip/fileio.py
```diff
@@ -217,15 +216,17 @@ def qsave(data, name='qutip_data'):
...
         Name of output data file.
 
     """
     # open the file for writing
-    fileObject = open(name + '.qu', 'wb')
-    # this writes the object a to the file named 'filename.qu'
-    pickle.dump(data, fileObject)
-    fileObject.close()
+    file = Path(name)
+    file = file.with_suffix(file.suffix + ".qu")
```
nleehone, line 225: file isn't used in the pickle or when opening the fileObject.

qutip/fileio.py
```diff
@@ -243,23 +245,14 @@ def qload(name):
         Object retrieved from requested file.
 
     """
-    with open(name + ".qu", "rb") as fileObject:
+    file = Path(name)
+    if file.suffix != '.qu':
+        file.with_suffix('.qu')
+
+    with open(name, "rb") as fileObject:
```
nleehone, line 250: The behaviour here has changed as it no longer automatically includes the .qu. I believe this was meant to be dealt with by the file.with_suffix above, but the file Path object is never used.

PR #1814, opened 2022-02-25
Bloch

anna-naden:
Make the conventions for view angle match between the classes Bloch and Bloch3d

Right now Bloch measures azimuth from y axis and elevation from equator, and Bloch3d measures azimuth from x axis and elevation from North pole.

The proposed change will measure azimuth from x axis and elevation from equator for both classes, in accordance with standard physics conventions.


hodgestar:
@anna-naden This PR only changes where the sphere is viewed from -- it doesn't change where in 3-D space anything is drawn. Rotating the view is probably a great feature to have, but it should be user controllable.

I'm going to close this PR for now. Let's revisit it for version 5 (i.e. a PR against dev.major).

PR #1820, opened 2022-03-03
Fix bug in eigen solver

Ericgig:
**Description**
There are some edge cases with the eigen solver that are bugged in `dev.major` that tests missed:
- `_eigs_dense` : It would not compute the proper eigen states when not all eigen states are required and the matrix is non-Hermitian.
- `_eigs_csr` : When all but one eigen values / vectors are needed: the smallest eigen values would always be removed, even when `sort='high'`. Also one element of the eigen vector matrix was removed, not one column.

I rewrote the tests with pytest to catch more cases.

**Changelog**
Fix some edges cases in `eigs`.

### DIFF
--- a/qutip/core/data/eigen.py
+++ b/qutip/core/data/eigen.py
@@ -79,9 +79,10 @@ def _eigs_dense(data, isherm, vecs, eigvals, num_large, num_small):
     if not isherm and eigvals > 0:
         if vecs:
             if num_small > 0:
-                evals, evecs = evals[:num_small], evecs[:num_small]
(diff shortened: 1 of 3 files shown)


Ericgig:
I rebased against dev.major. The documentation build issue appeared here too, but otherwise should be good.

hodgestar (approved):

hodgestar:
The changes look good but there are still some `test_satisfy_eigenvalue_equation` failures -- perhaps the accuracy just needs to be loosened slightly?

hodgestar (approved):

PR #1824, opened 2022-03-07
Bloch

anna-naden:
This fixes an incompatibility between the current version of QuTiP and the current version of matplotlib. When plotting the Bloch sphere, the call to Bloch.show() produces a run-time error because the Arrow3D class does not implement the method do_3d_projection. This method returns a sort key to matlotlib for sorting artists. This fix adds the method, simply returning a value of 1. The assumption here is that there is only one artist, or that the sort order does not matter.

This PR also modifies the conventions for the viewing angles specified in the view attribute of the Bloch and Bloch3d classes, rotating them by a quarter of a turn. See prior PR.


hodgestar:
@anna-naden There is already a PR for adding `do_3d_projection`  -- see #1818. It has less hacky implementation. If you have time, a review of #1818 would be appreciated.

hodgestar:
Superceded by #1818.

PR #1828, opened 2022-03-09
Add contributors image

### LINKED ISSUE #1713: Create image of code contributors
This code could be used:
https://github.com/dynamicwebpaige/nanowrimo-2021/blob/main/15_VS_Code_contributors.ipynb

Leo-am:
**Description**

Following the suggestion given by @nathanshammah in fix #1713, includes a notebook to create an image with the contributors' names. A figure example is also added.

Currently, the notebook is located in the main folder. Suggestions for better locations for the notebook and image are welcome.

**Changelog**
Added contributors image.


BoxiLi:
Thanks! I don't think we should include ipynb files into the main repo though as they are usually pretty large binary files. 

If we want to have this figure, it would be nice to include the source code stored somewhere in a file under `qutip\doc` and generate the figure automatically when building the doc. Just like other matplotlib figures.

In addition, it would be good to avoid using `pandas` and `mpmath` in the code if possible. Python has its own JASON reader and for mpmath, I'm not sure if such as high float precision is necessary.

Leo-am:
Hi, @BoxiLi!

Thank you for your comments, I agree with all of them. These are the changes I implemented: 

- removed Jupyter notebook and .png file from main repo; 
- added plot directive under qutip/doc/contributors.rst;
- removed pandas and mpmath imports.

The image is currently displayed on the top of the "Developers" section of the documentation. 
The team can discuss a more appropriate position for it, as well as changes in the figure aspects, which I expect to be straightforward to implement with the current version.

I am happy to discuss and implement other changes as well.

BoxiLi:
The code looks great. For some reason, the HTML rendering test fails in the first test run, but it seems to be a connection problem and it passes after I reran the test.

@nathanshammah do you have further comments?

nathanshammah:
Looks good to me, thanks @Leo-am!

PR #1830, opened 2022-03-10
Fix tidyup not reacting to change in settings.

### LINKED ISSUE #1829: QuTiP fails to change cut-off tolerance
### Bug Description

QuTiP Qobj has a `tinyup()` method that sets to 0 all values lower than a certain threshold defined by `settings.auto_tidyup_atol`:

https://github.com/qutip/qutip/blob/f5bdb6dda3f121c253a4862235674fa8980b7681/qutip/qobj.py#L1383

As explained in [here](https://qutip.org/docs/latest/guide/guide-settings.html), the user should be able to change this value. 

Playing with this options I realized that changing the `qutip.settings.auto_tidyup` to False works, but changing the `qutip.settings.auto_tidyup_atol` value doesn't.

### Code to Reproduce the Bug

```shell
import qutip

qutip.settings.auto_tidyup_atol = 1e-30
a = qutip.Qobj(1)
print(a * 1e-15)
qutip.settings.auto_tidyup = False
print(a * 1e-15)
```

### Code Output

```shell
Quantum object: dims = [[1], [1]], shape = (1, 1), type = bra
Qobj data =
[[0.]]
Quantum object: dims = [[1], [1]], shape = (1, 1), type = bra
Qobj data =
[[1.e-15]]
```

### Expected Behaviour

The first print should have been:
```
Quantum object: dims = [[1], [1]], shape = (1, 1), type = bra
Qobj data =
[[1.e-15]]
```

When multiplying `a * 1e-15`, QuTiP calls the `__mul__` magic method, where `out.tidyup()` is called:

https://github.com/qutip/qutip/blob/f5bdb6dda3f121c253a4862235674fa8980b7681/qutip/qobj.py#L581

Tidyup contains one argument set to `settings.auto_tidyup_atol` by default. This value should change to 1e-30 when running `qutip.settings.auto_tidyup_atol = 1e-30`, but this never happens. Consequently, the output value is set to 0.

### Your Environment

```shell
QuTiP Version:      4.6.3
Numpy Version:      1.22.3
Scipy Version:      1.8.0
Cython Version:     None
Matplotlib Version: 3.5.1
Python Version:     3.10.2
Number of CPUs:     8
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
```

### Additional Context

_No response_

Ericgig:
**Description**
As observed by @amitjansc, tidyup did not properly reacted to change to the default tolerance in settings.
Fix this and add tests.

**Related issues or PRs**
Close #1829 

**Changelog**
Ensure `tidyup`'s default tolerance is read from settings at each call.

### DIFF
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -1380,7 +1380,7 @@ class Qobj(object):
         q.data.sort_indices()
         return q
 
-    def tidyup(self, atol=settings.auto_tidyup_atol):
+    def tidyup(self, atol=None):
         """Removes small elements from the quantum object.
 
         Parameters
@@ -1395,6 +1395,8 @@ class Qobj(object):
             Quantum object with small elements removed.
 
         """
+        atol = atol or settings.auto_tidyup_atol
+
         if self.data.nnz:
             # This does the tidyup and returns True if
             # The sparse data needs to be shortened
--- a/qutip/tests/test_qobj.py
+++ b/qutip/tests/test_qobj.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 import scipy.sparse as sp
 import scipy.linalg as la
 import numpy as np
@@ -11,6 +12,7 @@ from qutip.superoperator import (spre, spost, operator_to_vector,
                                  vector_to_operator)
 from qutip.superop_reps import to_super, to_choi, to_chi
 from qutip.tensor import tensor, super_tensor, composite
+import qutip.settings as settings
 
 from operator import add, mul, truediv, sub
 import pytest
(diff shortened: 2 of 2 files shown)


qutip/qobj.py
```diff
@@ -1395,6 +1395,8 @@ def tidyup(self, atol=settings.auto_tidyup_atol):
             Quantum object with small elements removed.
 
         """
+        atol = atol or settings.auto_tidyup_atol
```
hodgestar, line 1398:
```suggestion
        if atol is None:
            atol = settings.auto_tidyup_atol
```
The `atol or settings.auto_tidyup_atol` will prevent using `atol=0.0`. Perhaps not the most important use case, but it worked before and I can see people relying on it rather than writing `if atol: obj = obj.tidyup(atol=atol` in their own code.

qutip/tests/test_qobj.py
```diff
@@ -1014,3 +1016,26 @@ def test_unit():
...
+
+
+def test_tidyup():
+    small = Qobj(1e-15)
+    small.tidyup(1e-14)
+    assert small.norm() == 0
+
+
+@contextmanager
+def tidyup_tol(tol):
```
hodgestar, line 1028:
```suggestion
def auto_tidyup_tol(tol):
```
Thought it might be nice to make it a bit clearer what the context manager changed.

qutip/tests/test_qobj.py
```diff
@@ -1014,3 +1016,26 @@ def test_unit():
     psi.unit(inplace=True)
     assert psi == psi2
     assert np.allclose(np.linalg.norm(psi.full()), 1.0)
+
+
+def test_tidyup():
+    small = Qobj(1e-15)
+    small.tidyup(1e-14)
+    assert small.norm() == 0
```
hodgestar, line 1024: Should we not also check that elements greater than 1e-14 are kept? Perhaps like we do for the `test_tidyup_default` test?

hodgestar (commented):
Thanks for sorting this one out, @Ericgig.

hodgestar (approved):

PR #1837, opened 2022-03-21
Set point transparency on Bloch sphere

### LINKED ISSUE #1287: Set point transparency on Bloch sphere
In our simulations we frequently run Monte-Carlo samples that would be great to visualize on the Bloch sphere. However, an alpha value would be very helpful.

Bloch().add_points(pnt, alpha = 0.3)

**Describe alternatives you've considered**
Absolutely not sure.

**Additional context**
In a similiar fashion I wonder if colormaps etc would become possible.

xspronken:
**Description**
Adding an option to set the transparency of points and vectors on the Bloch sphere.

**Related issues or PRs**
fix #1287 

**Changelog**
Added transparency parameter to the add_point, add_vector and add_states methods in the Bloch and Bloch3d classes.

### DIFF
--- a/doc/apidoc/classes.rst
+++ b/doc/apidoc/classes.rst
@@ -36,6 +36,9 @@ Bloch sphere
 .. autoclass:: qutip.bloch.Bloch
     :members:
 
+.. autoclass:: qutip.bloch3d.Bloch3d
+    :members:
+
 
 Distributions
 -------------
--- a/doc/apidoc/classes.rst
+++ b/doc/apidoc/classes.rst
@@ -36,9 +36,6 @@ Bloch sphere
 .. autoclass:: qutip.bloch.Bloch
     :members:
 
-.. autoclass:: qutip.bloch3d.Bloch3d
-    :members:
-
 
 Distributions
 -------------
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -324,7 +324,7 @@ class Bloch:
         alpha : float, default=1.
             Transparency value for the vectors. Values between 0 and 1.
         """
-        if not isinstance(points[0], (list, ndarray)):
+        if not isinstance(points[0], (list, tuple, ndarray)):
             points = [[points[0]], [points[1]], [points[2]]]
         points = array(points)
         if meth == 's':
@@ -335,13 +335,15 @@ class Bloch:
                 pnts = points
             self.points.append(pnts)
             self.point_style.append('s')
+            self.point_alpha.append(alpha)
         elif meth == 'l':
             self.points.append(points)
             self.point_style.append('l')
+            self.point_alpha.append(alpha)
         else:
             self.points.append(points)
             self.point_style.append('m')
-        self.point_alpha.append(alpha)
+            self.point_alpha.append(alpha)
 
     def add_states(self, state, kind='vector', alpha=1.0):
         """Add a state vector Qobj to Bloch sphere.
@@ -384,9 +386,10 @@ class Bloch:
         if isinstance(vectors[0], (list, tuple, ndarray)):
             for vec in vectors:
                 self.vectors.append(vec)
+                self.vector_alpha.append(alpha)
         else:
             self.vectors.append(vectors)
-        self.vector_alpha.append(alpha)
+            self.vector_alpha.append(alpha)
 
     def add_annotation(self, state_or_vector, text, **kwargs):
         """
@@ -714,7 +717,7 @@ class Bloch:
             zs3d = self.vectors[k][2] * array([0, 1])
 
             color = self.vector_color[mod(k, len(self.vector_color))]
-            alpha = self.vector_alpha[mod(k, len(self.vector_alpha))]
+            alpha = self.vector_alpha[k]
 
             if self.vector_style == '':
                 # simple line style
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -321,7 +321,7 @@ class Bloch:
             Type of points to plot, use 'm' for multicolored, 'l' for points
             connected with a line.
 
-        alpha : float, Default = 1.
+        alpha : float, default=1.
             Transparency value for the vectors. Values between 0 and 1.
         """
         if not isinstance(points[0], (list, ndarray)):
@@ -354,7 +354,7 @@ class Bloch:
         kind : {'vector', 'point'}
             Type of object to plot.
 
-        alpha : float, Default = 1.
+        alpha : float, default=1.
             Transparency value for the vectors. Values between 0 and 1.
         """
         if isinstance(state, Qobj):
@@ -378,7 +378,7 @@ class Bloch:
         vectors : array_like
             Array with vectors of unit length or smaller.
 
-        alpha : float, Default = 1.
+        alpha : float, default=1.
             Transparency value for the vectors. Values between 0 and 1.
         """
         if isinstance(vectors[0], (list, tuple, ndarray)):
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -623,11 +623,11 @@ class Bloch:
         self.plot_back()
         self.plot_points()
         self.plot_vectors()
+        self.plot_lines()
+        self.plot_arcs()
         self.plot_front()
         self.plot_axes_labels()
         self.plot_annotations()
-        self.plot_lines()
-        self.plot_arcs()
         # Trigger an update of the Bloch sphere if it is already shown:
         self.fig.canvas.draw()
 
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -309,7 +309,7 @@ class Bloch:
         self._lines = []
         self._arcs = []
 
-    def add_points(self, points, meth='s',alpha=1.0):
+    def add_points(self, points, meth='s', alpha=1.0):
         """Add a list of data points to bloch sphere.
 
         Parameters
@@ -320,9 +320,9 @@ class Bloch:
         meth : {'s', 'm', 'l'}
             Type of points to plot, use 'm' for multicolored, 'l' for points
             connected with a line.
-        
-        alpha : float
-            Transparency value for the vectors. Values between 0 and 1. Default = 1.
+
+        alpha : float, Default = 1.
+            Transparency value for the vectors. Values between 0 and 1.
         """
         if not isinstance(points[0], (list, ndarray)):
             points = [[points[0]], [points[1]], [points[2]]]
@@ -342,7 +342,7 @@ class Bloch:
             self.points.append(points)
             self.point_style.append('m')
         self.point_alpha.append(alpha)
-    
+
     def add_states(self, state, kind='vector', alpha=1.0):
         """Add a state vector Qobj to Bloch sphere.
 
@@ -353,9 +353,9 @@ class Bloch:
 
         kind : {'vector', 'point'}
             Type of object to plot.
-        
-        alpha : float
-            Transparency value for the vectors. Values between 0 and 1. Default = 1.
+
+        alpha : float, Default = 1.
+            Transparency value for the vectors. Values between 0 and 1.
         """
         if isinstance(state, Qobj):
             state = [state]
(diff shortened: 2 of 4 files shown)


AGaliciaMartinez (changes requested):
Hey @xspronken, thanks for including the feature!
I added a small suggestion but otherwise it looks good! However, we do need a test for this new feature. Testing the code for the `Bloch` class is a bit tricky and was not being done until very recently but with the examples in `qutip/tests/test_bloch.py` it should be doable. We would need a similar function to those found in `qutip/tests/test_bloch.py` that ensures the alpha keyword works as expected. Instead of creating new functions, you could also expand the keyword arguments of some of the already existing functions while adding some more parametrized cases. This is only necessary for the `Bloch` methods as we do not have a testing infrastructure for the `Bloch3D` class yet and writing one from scratch should probably go in another PR. So do not worry about testing Bloch3D.

doc/apidoc/classes.rst
```diff
@@ -36,6 +36,9 @@ Bloch sphere
 .. autoclass:: qutip.bloch.Bloch
     :members:
 
+.. autoclass:: qutip.bloch3d.Bloch3d
+    :members:
+
```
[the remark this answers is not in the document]
xspronken: Alright! I will look into it!

xspronken:
Alright I'll look into it!

doc/apidoc/classes.rst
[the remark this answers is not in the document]
xspronken: Ok I will remove it.
 I did check and was able to build the html docs locally after renaming a file (doc/cqobjevo_compiled_coeff_739974011510.pyx) that couldn't be found while building as the last thee numbers were wrong. I dont know if this might be the issue.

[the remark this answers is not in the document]
BoxiLi: It is probably because the docstring of `Bloch3d` includes syntax errors. It wasn't included in API doc so it was hidden.

It is likely caused by https://github.com/qutip/qutip/blob/3eeb22ca747790f80bd8ae3af72713396415cf42/qutip/bloch3d.py#L53
where `>` is a special character in rst. There are 4 of those, corresponding to 4 warnings in the test.

Wrapping the string with backticks may solve this: ` ``['|x>', '']`` `, I think.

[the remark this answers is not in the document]
AGaliciaMartinez: The reason you may be able to build locally is because in the CI we treat the warnings as errors. You should get the same error we get in the CI if you run `make html SPHINXOPTS="-W --keep-going -T"`.

If we can solve the issue as @BoxiLi suggested we may want to include it in this PR.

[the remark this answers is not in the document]
xspronken: I tried building locally using `make html SPHINXOPTS="-W --keep-going -T"` and got an error `make: *** [Makefile:45: html] Error 1` even after adding the backticks as @BoxiLi suggested. So I removed the class from the rst file and tried again but the error persisted. The command did seem to build most the html documentation anyway though so I'm quite confused.

AGaliciaMartinez (commented):
Thanks for adding these tests! I added a small suggestion (contradicting a previous suggestion I made, sorry for that) on how to sort them to keep things clean. I also included a few minor changes in the docstrings.

doc/apidoc/classes.rst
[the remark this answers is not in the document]
AGaliciaMartinez: It seems to run fine now locally for me. Lets not worry too much about this. I will rerun the CI to see if its fixed.

AGaliciaMartinez (approved):

qutip/tests/test_bloch.py
```diff
@@ -187,17 +187,18 @@ def plot_vector_ref(self, fig, vectors):
...
             a = Arrow3D(
                 xs3d, ys3d, zs3d,
-                mutation_scale=20, lw=3, arrowstyle="-|>", color=color)
+                mutation_scale=20, lw=3, arrowstyle="-|>",
+                color=color)
             b.axes.add_artist(a)
 
     @pytest.mark.parametrize([
-        "vectors",
+        "vectors"
```
AGaliciaMartinez, line 195:
```suggestion
        "vectors",
```

qutip/tests/test_bloch.py
```diff
@@ -187,17 +187,18 @@ def plot_vector_ref(self, fig, vectors):
...
     ], [
         pytest.param(
             (0, 0, 1), id="single-vector-tuple"),
         pytest.param(
             [0, 0, 1], id="single-vector-list"),
-       pytest.param(
+        pytest.param(
```
AGaliciaMartinez, line 201:
```suggestion
       pytest.param(
```

qutip/tests/test_bloch.py
```diff
@@ -210,3 +211,49 @@ def plot_vector_ref(self, fig, vectors):
     def test_vector(self, vectors, fig_test, fig_ref):
         self.plot_vector_test(fig_test, vectors)
         self.plot_vector_ref(fig_ref, vectors)
+
+    def plot_vector_test_alpha(self, fig, vectors, alpha):
+                b = Bloch(fig=fig)
+                b.add_vectors(vectors, alpha=alpha)
+                b.render()
```
AGaliciaMartinez, line 218: Same here with the indentation.

qutip/tests/test_bloch.py
```diff
@@ -167,9 +167,9 @@ def test_line(
         self.plot_line_ref(fig_ref, start_ref, end_ref, **kwargs)
 
     def plot_vector_test(self, fig, vectors):
-        b = Bloch(fig=fig)
-        b.add_vectors(vectors)
-        b.render()
+            b = Bloch(fig=fig)
+            b.add_vectors(vectors)
+            b.render()
```
AGaliciaMartinez, line 172: I've just realized this has unnecessary indentation.

qutip/bloch.py
```diff
@@ -353,23 +366,27 @@ def add_states(self, state, kind='vector'):
...
+
+        alpha : float, default=1.
+            Transparency value for the vectors. Values between 0 and 1.
         """
         if isinstance(vectors[0], (list, tuple, ndarray)):
             for vec in vectors:
                 self.vectors.append(vec)
         else:
             self.vectors.append(vectors)
+        self.vector_alpha.append(alpha)
```
hodgestar, line 389: When vectors is a list of vectors, we should add an alpha value for every vector in the list -- otherwise the wrong alpha values will be applied if one does, e.g.:
```python
b.add_vectors([psi0, psi1], alpha=0.5)
b.add_vectors([psi2, psi3], alpha=1.0)
```
Currently `psi0` and `psi2` will have alpha 0.5 and `psi1` and `psi3` will have alpha 1.0, which is not correct.

qutip/bloch.py
```diff
@@ -697,19 +714,21 @@ def plot_vectors(self):
             zs3d = self.vectors[k][2] * array([0, 1])
 
             color = self.vector_color[mod(k, len(self.vector_color))]
+            alpha = self.vector_alpha[mod(k, len(self.vector_alpha))]
```
hodgestar, line 717: We shouldn't apply a `mod(k, ...)` here. It is the correct thing to do for the color because those are expected to cycle through a list of colors, but for the alpha values we have an explicit alpha value for every vector, so there should be no need to wrap.

qutip/bloch.py
```diff
@@ -751,17 +771,17 @@ def plot_points(self):
...
+                                  color=pnt_colors, marker=marker)
 
             elif self.point_style[k] == 'l':
                 color = self.point_color[mod(k, len(self.point_color))]
                 self.axes.plot(real(self.points[k][1]),
                                -real(self.points[k][0]),
                                real(self.points[k][2]),
-                               alpha=0.75, zdir='z',
-                               color=color)
+                               alpha=self.point_alpha[k],
```
hodgestar, line 783: Here we are changing the default alpha from 0.75 to 1.0. I guess we could either have the default in add points depend on the point style used (seems a bit complicated) or add a note to the docstring for `add_points` stating that the default for points with lines changed from 0.75 to 1.0. I think I favour documenting the change, rather than making the implementation more complex.

qutip/bloch3d.py
```diff
@@ -50,15 +50,15 @@ class Bloch3d():
         Width of displayed vectors.
     view : list {[45,65]}
         Azimuthal and Elevation viewing angles.
-    xlabel : list {['|x>', '']}
+    xlabel : list {``['|x>', '']``}
```
hodgestar, line 53: What are the extra quotes here for? I can imagine strange formatting is needed, but I don' know exactly why in this case. :)

qutip/bloch3d.py
```diff
@@ -262,33 +267,39 @@ def add_states(self, state, kind='vector'):
...
+        alpha : float, default=1.
+            Transparency value for the vectors. Values between 0 and 1.
+
         """
         if isinstance(vectors[0], (list, np.ndarray)):
             for vec in vectors:
                 self.vectors.append(vec)
         else:
             self.vectors.append(vectors)
+        self.vector_alpha.append(alpha)
```
hodgestar, line 302: Similar critique here to `add_vectors` for `Bloch`. We need to add an alpha value for each vector.

hodgestar (changes requested):
Hello!

Back from holiday and catching up. I made some suggestions to fix handling the new alpha values for vectors.

In addition, I think we should add tests for:

- The Bloch3D changes
- The case where multiple vectors are added that will currently fail

Generally the code looks good though!

As @AGaliciaMartinez mentioned, I think we would clean up the API more for QuTiP 5, but it's great to have in 4.7 with this minimal API change.

hodgestar:
I resolved the issue with setting the alpha for multiple sets of vectors and added tests for that (and for setting the alpha of points).

I'm not going to add tests for Bloch3d now since that uses mayavi which looks like it might be a whole other testing rabbit hole. I will simply run some checks in a notebook manually tomorrow.

I still need to update the docstring and look at what the docstring format changes are about.

@xspronken @AGaliciaMartinez Reviews of my most recent commits welcomed.

qutip/bloch3d.py
xspronken: > is a special character in .rst, and since the docs weren't able to build I tried changing adding the backticks as @BoxiLi suggested in order to consider it a regular character in a string. I still wasn't able to entirely build the docs for the Bloch3d class locally, so we decided to remove the Bloch3d class from the docs for now. The html was partially built though including the part with these statements, so I was able to check whether the backticks affected anything, which they did: the statement was changed from a hyperlink (that led nowhere) to a regular string. So I decided to keep them even though the Bloch3d class wasn't added to the rst file.

xspronken:
The new commits look good! Was wondering whether we should add support for array-like alpha values so that something like `add_vectors([[0,0,1],[1,0,0]], alpha=[0.75,0.60])` would work. Right now it's already possible to do using multiple `add_vector` methods, each with a different alpha value but the other option might be a bit cleaner for the user.

qutip/bloch.py
hodgestar: Implemented.

qutip/bloch3d.py
hodgestar: Thanks!

hodgestar:
> The new commits look good! Was wondering whether we should add support for array-like alpha values so that something like `add_vectors([[0,0,1],[1,0,0]], alpha=[0.75,0.60])` would work. Right now it's already possible to do using multiple `add_vector` methods, each with a different alpha value but the other option might be a bit cleaner for the user.

I think a single alpha is fine if we take the view that each call to `add_vectors` is one set of related vectors that share some common properties (although they already rotate colours, which is sometimes helpful and sometimes a bit weird :). We need to rethink the `Bloch` interface for version 5 though, so keep ideas like this in mind for then.

hodgestar:
I tried out Bloch3d locally and it seems to work, so I'll merge once the build passes.

@xspronken @AGaliciaMartinez -- thank you for working on this with me.

hodgestar (approved):

PR #1843, opened 2022-03-29
Fix ttmsolve bug with addition of Qobj and None

christian512:
**Description**
The nonmarkov transfer tensor method (TTM) fails in an example notebook (see below). I tracked down the error to the *ttmsolve* method, where None is used to fill an array. I replaced None with an empty Qobj to fix the bug occuring from adding a Qobj and None.

**Related issues or PRs**
The related issue is in the qutip-notebooks repo: https://github.com/qutip/qutip-notebooks/issues/122

**Changelog**
Replace placeholder in ttmsolve with empty Qobj

### DIFF
--- a/qutip/nonmarkov/transfertensor.py
+++ b/qutip/nonmarkov/transfertensor.py
@@ -14,7 +14,7 @@ import numpy as np
 
 
 from qutip import (Options, spre, vector_to_operator, operator_to_vector,
-                   ket2dm, isket)
+                   ket2dm, isket, Qobj)
 from qutip.solver import Result
 from qutip.expect import expect_rho_vec
 
@@ -148,7 +148,7 @@ def ttmsolve(dynmaps, rho0, times, e_ops=[], learningtimes=None, tensors=None,
     K = len(tensors)
     states = [rho0vec]
     for n in range(1, len(times)):
-        states.append(None)
+        states.append(Qobj())
         for k in range(n):
             if n-k < K:
                 states[-1] += tensors[n-k]*states[k]
--- a/qutip/nonmarkov/transfertensor.py
+++ b/qutip/nonmarkov/transfertensor.py
@@ -146,12 +146,13 @@ def ttmsolve(dynmaps, rho0, times, e_ops=[], learningtimes=None, tensors=None,
     K = len(tensors)
     states = [rho0vec]
     for n in range(1, len(times)):
-        # Append empty state
-        states.append(None)
-        for k in range(n):
-            if n - k < K:
-                tmp = tensors[n - k] * states[k]
-                states[-1] = tmp if states[-1] is None else tmp + states[-1]
+        # Set current state
+        state = None
+        for j in range(1, min(K, n + 1)):
+            tmp = tensors[j] * states[n - j]
+            state = tmp if state is None else tmp + state
+        # Append state to all states
+        states.append(state)
     for i, r in enumerate(states):
         if opt.store_states or expt_callback:
             if r.type == 'operator-ket':
--- a/qutip/nonmarkov/transfertensor.py
+++ b/qutip/nonmarkov/transfertensor.py
@@ -11,10 +11,9 @@ method (TTM), introduced in [1].
 """
 
 import numpy as np
-from attr import validate
 
 from qutip import (Options, spre, vector_to_operator, operator_to_vector,
-                   ket2dm, isket, Qobj)
+                   ket2dm, isket)
 from qutip.solver import Result
 from qutip.expect import expect_rho_vec
 
--- a/qutip/nonmarkov/transfertensor.py
+++ b/qutip/nonmarkov/transfertensor.py
@@ -11,7 +11,7 @@ method (TTM), introduced in [1].
 """
 
 import numpy as np
-
+from attr import validate
 
 from qutip import (Options, spre, vector_to_operator, operator_to_vector,
                    ket2dm, isket, Qobj)
@@ -46,7 +46,6 @@ class TTMSolverOptions:
 
     def __init__(self, dynmaps=None, times=[], learningtimes=[],
                  thres=0.0, options=None):
-
         if options is None:
             options = Options()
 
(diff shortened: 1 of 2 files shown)


AGaliciaMartinez (changes requested):
Thanks @christian512 for taking care of this! However, I am going to ask you to take a different approach to solve the problem. The reason is that empty `Qobj` are not well supported in QuTiP so it is better to avoid using them. In this case it is possible to rewrite the two for loops to not need the empty `Qobj`. Could you try this other approach? 

We also need tests to ensure that this error does not happen again. Unfortunately, no tests were included for this function by previous developers. Would you mind adding _at least_ one that would catch the error in the notebook? It may suffice to use part of the code in the notebook itself and it could be placed in a file named `qutip/tests/test_transfertensor.py`.

christian512:
Yes, I will add the tests and modify the loops soon. Thanks for the comments, @AGaliciaMartinez :+1:

christian512:
I am using `mesolve` in the test now. I am not sure if that is considered a good style, since the test also depends on this function. I could also just hardcode the expected output. What do you think is better, @AGaliciaMartinez ?

qutip/nonmarkov/transfertensor.py
```diff
@@ -148,10 +147,12 @@ def ttmsolve(dynmaps, rho0, times, e_ops=[], learningtimes=None, tensors=None,
...
     states = [rho0vec]
     for n in range(1, len(times)):
+        # Append empty state
         states.append(None)
         for k in range(n):
-            if n-k < K:
-                states[-1] += tensors[n-k]*states[k]
+            if n - k < K:
+                tmp = tensors[n - k] * states[k]
+                states[-1] = tmp if states[-1] is None else tmp + states[-1]
```
hodgestar, line 155: This whole loop seems a bit iffy. `k` runs from `max(0, n - K + 1)` to `n - 1` and `n - k` runs from `K - 1` to `1`. So there are `K` tensors but only `K - 1` of them are used which is strange.

It also seems more natural to express this as `for j in range(0, min(K, n)):` with `k = n - j - 1` which would have made the missing `k` more obvious and remove the need for the tangled indexing and the `if n -  k < K`.

On the other hand the test seems to pass so maybe one of the tensors is meant to be ignored? Do you perhaps know what is supposed to happen here?

qutip/tests/test_transfertensor.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import qutip as qt
+import numpy as np
+from qutip.nonmarkov.transfertensor import *
```
hodgestar, line 4: We should add any more uses of `import *` in QuTiP -- it just makes it hard to see what the code is doing. Just import `ttmsolve`.

qutip/nonmarkov/transfertensor.py
hodgestar: I would also just set `state = None` outside the loop and do `state = tmp if state is None else tmp + state` and then do `states.append(state)` afterwards (this also remove some awkward `-1` terms from the indexing.

hodgestar (changes requested):
Thank you for finding and fixing this bug!

I left some questions about the strange indexing in the loop. If we decide the current logic is correct, we can just add a comment explaining why. If it is not correct, we should fix it in the PR, I think.

Using `mesolve` to find the canonical result is great -- or at least it is much easier to see how the result was obtained and how to update it than if the true result was hardcoded.

AGaliciaMartinez (changes requested):
Hey @alexbrc, great work! I added a few more suggestions but its starting to look good!! This function definitely needed some testing.

Regarding tests, ideally tests should compare to analytical solutions. I believe there is an analytical solution to this problem although I do not think it will be trivial. There are other problems that do have analytical solutions. A good place to look for systems with analytical solutions we know is `test_mesolve`. Ideally we would like one of those here although comparing to mesolve is not terrible either. 

Another thing to note in tests is that we want them to be fast as there are many of them that need to be run for each PR. This one is too slow. I already hinted a place in the comments that could be changed to make it faster but may not be enough. I also think we compute too many values. Try to fiddle with times and learning times (having learning times smaller than times) to find a set of values that runs fast enough yet it represents a non-trivial simulation and for which ttmsolve works well enough. Having an extra parameter for the collapse operator, `gamma` instead of `kappa`, may help.

qutip/nonmarkov/transfertensor.py
christian512: I don't know why the first tensor is left out. In order to drop the `if n-k < K` expression I changed the `k` for loop by defining `j = n - k`. This resolves to a for loop `for j in range(1, min(K, n+1))`.

And I replaced the `states.append(None)` part as you described.

qutip/tests/test_transfertensor.py
```diff
@@ -0,0 +1,49 @@
...
+    Checks the output of ttmsolve using an example from Jaynes-Cumming model,
+    which can also be found in the qutip-notebooks repository.
+    """
+    # Define Hamiltonian and states
+    N, kappa = 3, 1.0
+    a = qt.tensor(qt.qeye(2), qt.destroy(N))
+    sm = qt.tensor(qt.sigmam(), qt.qeye(N))
+    sz = qt.tensor(qt.sigmaz(), qt.qeye(N))
+    H = kappa * (a.dag() * sm + a * sm.dag())
+    c_ops = [np.sqrt(kappa) * a]
```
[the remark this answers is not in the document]
christian512: There was a bug in this. I used `kappa` for both, `H` and `c_ops`. That results in a constant expectation value for `qt.sigmaz()`. I now changed `kappa` in `H` to a new coupling constant `g`.

qutip/tests/test_transfertensor.py
```diff
@@ -0,0 +1,49 @@
...
+        Et = qt.mesolve(H, E0, [0., t], c_ops, []).states[-1]
+        return ptracesuper * (Et * superrho0cav)
+
+    # solve using transfer method
+    learning_times = np.arange(0, 10.0, 0.1)
+    learning_maps = [dynmap(t) for t in learning_times]
+    ttmsol = ttmsolve(learning_maps, rho0a, times)
+    ttm_z = qt.expect(qt.sigmaz(), ttmsol.states)
+    # check that ttm result and exact solution are close
+    assert np.allclose(ttm_z, exact_z, atol=0.01)
```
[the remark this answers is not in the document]
christian512: As far as I understood `ttmsolve` it gives accurate results if the learning time is high enough. Than the results are also accurate for extrapolations. I tried a bit and found a good configuration. But the accuracy of the `ttmsolve` solution seems to be similar in the `learning_times` and for extrapolated times. With the chosen parameters I can use `atol = 1e-5`.

AGaliciaMartinez (approved):

AGaliciaMartinez:
@christian512, this looks great! I am happy with this being merged as it is. We could further improve tests by testing that all arguments work correctly but I believe that to be out of the scope of this PR. Since this PR already fixes a bug I would rather have it merged. @hodgestar, are you happy with this being merged after the changes?

hodgestar (approved):
Thanks again @christian512 for finding and fixing the bug and for making the ugly for loop nicer.

hodgestar:
And thanks @AGaliciaMartinez for reviewing!

PR #1845, opened 2022-04-04
Added support of list input in unit conversion functions

ksohan:
**Description**
I have added support for list input for the unit conversion functions. 

**Related issues or PRs**
I was passing a python list of values to convert from one unit to another but got the following error:
`TypeError: can't multiply sequence by non-int of type 'float'`
As the documentation says it accepts an array, I think a list should be supported.

**Changelog**
If the input is a python list, it will be converted to a numpy array. Because numpy array supports multiplying by a float, Python list doesn't.

### DIFF
--- a/qutip/utilities.py
+++ b/qutip/utilities.py
@@ -33,7 +33,9 @@ def n_thermal(w, w_th):
 
 
     """
-
+    if isinstance(w, list):
+        w = np.asarray(w)
+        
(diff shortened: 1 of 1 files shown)


hodgestar:
Closing this for now. Please feel free to open a PR for QuTiP 5 (currently in the `dev.major` branch) though.

PR #1846, opened 2022-04-04
Unify random qobj

### LINKED ISSUE #1838: Unify random `Qobj` creation functions support of `dims`.
### Describe the Issue!

Most of the functions in `random_objects.py` take an `N` input and an optional `dims`. Those input are redundant as the size (`N`) can be obtained from the `dims`. However the way they handle this inconsistent:
- Many functions, such as `rand_super`, `rand_dm`, check that the dims matches the size and raise an error if it doesn't.
- `rand_dm_ginibre`, `rand_super_bcsz` just pass the `dims` to `Qobj` without any checks.
- `rand_ket` has a default `N=0` and obtain the size from `dims` when `N` is `0`, but it raise a division per zero error when neither `N` nor `dims` is passed.
- `rand_ket_haar` also tries to obtain the size from `dims` when `N` is `0`, but the default value of `N` is `2`, thus one need to pass `N=0` to use it. `N=0` would fail almost everywhere else.

Also some function have a default size, other not: `rand_unitary(N, ...)`, `rand_unitary_haar(N=2, ...)`.

The interface of these functions need to be unified.

Some options:
- `N` is always needed and `dims` are checked to match.
- A default of `None` is used for `N` and either `N` or `dims` is needed. If both are given, we check that they are compatible. If none are passed, we raise an appropriate error.
- (For `dev.major`) Remove the `dims` parameter and replace `N` with `dimensions`, which act as `N` if an integer is given and `dims` is a list is. This is already the behavior of `qeye` and `qzero`.

Leo-am:
**Description**

Functions in `random_objects.py` deal with the parameters `N` and `dims` in different ways. Here I unify their interface implementing the following changes:

1) Both `N` and `dims` parameters are now optional, with default value `None` for all functions.
2) If both are given, we check if they match. An error is raised if they don't.
3) If only one of them is given, the other one receives an appropriate value.
4) If none of them are given, an error is raised requesting one of them.

Functions `rand_super` and `rand_super_bcsz` were left unchanged. I believe it's better to leave their changes for v5.0, when the parameters `N` and `dims` will be substituted by `dimensions`.

**Related issues or PRs**
Fix #1838.

**Changelog**
Unify interface from random_objects.py module.

### DIFF
--- a/qutip/random_objects.py
+++ b/qutip/random_objects.py
@@ -83,8 +83,8 @@ def randnz(shape, norm=1 / np.sqrt(2), seed=None):
 def rand_herm(N=None, density=0.75, dims=None, pos_def=False, seed=None):
     """Creates a random NxN sparse Hermitian quantum object.
 
-    If 'N' is an integer or None, uses :math:`H=0.5*(X+X^{+})` where :math:`X` is
-    a randomly generated quantum operator with a given `density`. Else uses
+    If 'N' is an integer or None, uses :math:`H=0.5*(X+X^{+})` where :math:`X`
+    is a randomly generated quantum operator with a given `density`. Else uses
     complex Jacobi rotations when 'N' is given by an array.
 
     Parameters
@@ -217,16 +217,8 @@ def rand_unitary(N=None, density=0.75, dims=None, seed=None):
         NxN Unitary quantum operator.
 
     """
-    if N is None and dims is None:
-        raise ValueError('Specify either the number of rows of unitary'
-                         ' operator (N) or dimensions of quantum object'
-                         ' (dims).')
-    elif N is None and dims:
-        N = np.prod(dims[0])
-    elif N and dims is None:
-        dims = [[N], [N]]
-    elif N and dims:
-        _check_dims(dims, N, N)
+    N, dims = _check_N_and_dims(dims, N, N, 'unitary operator')
+
 
     U = (-1.0j * rand_herm(N, density, seed=seed)).expm()
     U.data.sort_indices()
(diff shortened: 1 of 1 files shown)


Ericgig (commented):
Thank you for the PR.
It look as it should work, but could you add a test to ensure it does.

Most codeclimat issues should be solved if you make a functions to read `(N, dims)`.

For super operator you can use:
`N**2 = np.prod(qutip.dimensions.flatten(dims[0]))`
Ideally, we should also check `type_from_dims(dims) == 'super'`

AGaliciaMartinez:
Hey @Leo-am, thanks for taking care of this issue! 

What about we just keep one of the arguments instead of both `N` and `dims`? We could keep only `dims` and interpret it in the following way:
- If it is a list then it behaves as `dims`.
- If it is an integer it behaves as `N`.

This would mean we would not need to care about `N` and `dims` contradicting each other and since `dims` would not be optional but rather a positional argument, you do not have to care about it being missing. I also think this would be a less confusing alternative for the user as you want to use only one of them anyway.

In any case, since this change is not backwards compatible we should merge it in dev.major.

AGaliciaMartinez:
I see my suggestion was already discussed in the issue... sorry!

Leo-am:
> Thank you for the PR. It look as it should work, but could you add a test to ensure it does.
> 
> Most codeclimat issues should be solved if you make a functions to read `(N, dims)`.
> 
> For super operator you can use: `N**2 = np.prod(qutip.dimensions.flatten(dims[0]))` Ideally, we should also check `type_from_dims(dims) == 'super'`

Hi, @Ericgig! Thank you for the comments.
I will:

[ ] add a test
[ ] create a function to implement the check on `N`, `dims`.
[ ] add a checks for the superoperator functions: using `N**2 = np.prod(qutip.dimensions.flatten(dims[0]))`
[ ] add a checks for the superoperator functions: check `type_from_dims(dims) == 'super'`

Hi, @AGaliciaMartinez! Thanks for your comments. The plan for `dev.major` is to substitute both `N` and `dims` in favour of `dimensions`, just as you described. I'll start working on that, as soon as I finish the current PR.

hodgestar:
@Leo-am @Ericgig I originally marked this PR for inclusion in 4.7, but I'd like to suggest we skip this PR and clean things up for QuTiP 5. My reasoning is as follows:

- To be included in 4.7.X this PR would have to be included in 4.7.0 now and there is still quite a bit of work to do (writing tests, cleaning up _check_dims, handling some edges cases better) and I'd rather not rush those changes.
- If we make this change now, we will have to breach user code again for 5.0 because the interface being introduced here is not the one proposed for 5.0 in #1838.

Leo-am:
Hi, @hodgestar! Fair enough. 

I'll work on the changes for 5.0 and submit a new PR implementing the proposed changes in the next weeks.

hodgestar:
@Leo-am Thank you! Apologies for the wild goose chase. Hopefully getting to grips with the original code here makes writing the QuTiP 5 version much easier. :/

Issue #1842, opened 2022-03-29
Circuit plotting with measurements fails when `reverse_states=False`


jneer:
The LaTeX code generated for the following simple circuit is invalid, and trying to plot the circuit results in a RuntimeError:

```
qc = QubitCircuit(1, num_cbits=1, reverse_states=False)
qc.add_measurement("M0", targets=0, classical_store=0)
qc.png
```
Setting `reverse_states=True` works fine.

The following line in `QubitCircuit.latex_code()` needs to include a sign dependent on the value of `reverse_states`:
https://github.com/qutip/qutip/blob/00d0fe43055cff44fad803bd1ad1530e17fedeee/qutip/qip/circuit.py#L1741

hodgestar:
Fixed by #1847.

PR #1847, opened 2022-04-05
insert sign for plotting with non reversed states

christian512:
**Description**
There was a bug for plotting a QC with `reversed_states = False`, that caused a problem when plotting a measurement with classical storage. This PR fixes this bug.

**Related issues or PRs**
#1842 

**Changelog**
Fix circuit index used when plotting circuits with non-reversed states.


hodgestar (changes requested):
This looks good to me. Could you also add a test for the generated LaTeX?

This should in addition be submitted as a PR on the qutip-qip repository.

hodgestar (approved):
Thanks!

PR #1849, opened 2022-04-08
Add GitHub actions test run on windows-latest.

hodgestar:
**Description**
Add GitHub actions test run on windows-latest.

**Related issues or PRs**
- None so far.

**Changelog**
Add GitHub actions test run on windows-latest.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest]
         # Test other versions of Python in special cases to avoid exploding the
         # matrix size; make sure to test all supported versions in some form.
         python-version: ["3.9"]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -77,6 +77,11 @@ jobs:
             condaforge: 1
             pytest-extra-options: "-W ignore::ImportWarning -W ignore::DeprecationWarning:cvxpy.interface.scipy_wrapper"
 
+          # Windows
+          - case-name: Windows Latest
+            os: windows-latest
+            pytest-extra-options: "-k 'not (test_correlation or test_interpolate or test_dnorm)'"
```
Ericgig, line 83: I would prefer to skip `dnorm` tests from the test files instead of here.
For `test_correlation` and `test_interpolate`, if it's not that they don't pass, but that they are very slow, skipping them here seems appropriate.

Ericgig (commented):
It's great to have test passing on windows.

If the timeout is removed, let's remove the installation of `pytest-timeout`.

hodgestar:
Replacing this with #1853 now that I understand better what works on Windows and what doesn't.

PR #1850, opened 2022-04-10
Merge master into dev.major

Ericgig:
**Description**
Merge master at after #1847 into `dev.major`.

qip's changes are not kept, this part is still the same state as it was before the merge. Any improvement that were added to master should instead be pushed to qutip/qutip-qip.

HEOM: Files where pulled, but any issue was dealt by commenting the code. All related tests are skipped.

Steadystate: Test fails, there seems to be ~3 bugs that I could not fix in time.

The tests now consider warning as error, this cause other tests to fail, but they run fine without this. 

Documentation is still at master's state.

When needed, change from fastsparse to data layer where made as minimally as possible: steadystate, steadystate_floquet, floquet, krylovsolve, etc. use rough shims.


qutip/solve/steadystate.py
```diff
@@ -449,7 +462,10 @@ def _steadystate_LU_liouvillian(L, ss_args, has_mkl=0):
         perm2 = scipy.sparse.csgraph.reverse_cuthill_mckee(L)
         _rcm_end = time.time()
         rev_perm = np.argsort(perm2)
+        print(np.round(L.toarray(),3))
         L = _permute(L, perm2, perm2)
+        print(np.round(L.toarray(),3))
+        print()
```
AGaliciaMartinez, line 468: Are these print statements supposed to be there¿

qutip/core/cy/openmp/src/zspmv_openmp.cpp
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD:qutip/core/cy/openmp/src/zspmv_openmp.cpp
```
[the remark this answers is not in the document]
Ericgig: Fixed!

qutip/simdiag.py
```diff
@@ -78,75 +89,32 @@ def simdiag(ops, evals=True):
...
-    zipped.sort()
-    ds, perm = zip(*zipped)
-    ds = -np.real(np.array(ds))
-    perm = np.array(perm)
-    eigvecs_array = np.array(
-        [np.zeros((A.shape[0], 1), dtype=complex) for k in range(A.shape[0])])
-
-    for kk in range(len(perm)):  # matrix with sorted eigvecs in columns
-        eigvecs_array[kk][:, 0] = eigvecs[:, perm[kk]]
+    # TODO: rewrite using Data object
```
hodgestar, line 92: Is this TODO for now? If not, do we need to add it to one of the TODO lists in the PR?

Ericgig:
Floquet related examples in `doc/guide/scripts` fail. 
I am looking at it.

qutip/solve/steadystate.py
AGaliciaMartinez: weirdly enough, most of the errors disappear if only this file is merged into dev.major. Except for the two related to iterative methods not converging. It also seems to be way slower once only this file is merged.

qutip/tests/solve/test_brtools.py
```diff
@@ -44,26 +11,26 @@
 import platform
 
 
-@pytest.mark.skipif("platform.system() == 'Darwin'")
-def test_zheevr():
+#@pytest.mark.skipif("platform.system() == 'Darwin'")
```
hodgestar, line 14: Should this have been commented out in the merge?

qutip/tests/test_subsys_apply.py
```diff
@@ -147,11 +114,11 @@ def test_ComplexSuperApply(self):
         analytic_result[1] = Qobj(
             vector_to_operator(
                 superop @ operator_to_vector(analytic_result[1])
-            ).full())
+            ))#.full())
```
hodgestar, line 117: Commented out `.full()` should be removed before merge.

qutip/tests/test_subsys_apply.py
```diff
@@ -147,11 +114,11 @@ def test_ComplexSuperApply(self):
...
         analytic_result[3] = Qobj(
             vector_to_operator(
                 superop @ operator_to_vector(analytic_result[3])
-            ).full())
+            ))#.full())
```
hodgestar, line 121: Commented out `.full()` should be removed before merge.

setup.py
```diff
@@ -29,7 +29,7 @@ def process_options():
             Is this a release build (True) or a local development build (False)
         'openmp': bool
             Should we build our OpenMP extensions and attempt to link in OpenMP
-            libraries?
+            libraries? (Not supported in this version.)
```
hodgestar, line 32: We should make a TODO to either remove openmp support or add it back (maybe there is already an open issue?)

hodgestar (commented):
A few more things that looked slightly odd. Done with the (very) high-level review.

qutip/simdiag.py
Ericgig: Not for now, it's more a TODO until the full release.

setup.py
Ericgig: The hope it to add it back, it's on multiple to-dos (https://github.com/qutip/qutip/projects/6), but no issue has been opened.
Aton said he would take a look at it.

qutip/tests/test_subsys_apply.py
Ericgig: Removed

qutip/tests/test_subsys_apply.py
Ericgig: Removed

qutip/solve/steadystate.py
```diff
@@ -363,7 +375,8 @@ def steadystate(A, c_op_list=[], method='direct', solver=None, **kwargs):
 
     # Set weight parameter to max abs val in L if not set explicitly
     if 'weight' not in kwargs.keys():
-        ss_args['weight'] = np.abs(_data.norm.max(A.data))
+        # set the weight to the mean of the non-zero absoluate values in A:
+        ss_args['weight'] = np.mean(np.abs(A.full().data))
```
AGaliciaMartinez, line 379: This change is the source of two errors related to the iterative methods not converging. @Ericgig why is this change necessary?

qutip/solve/steadystate.py
AGaliciaMartinez: It seems I was not runningthe tests that do include 'use_rcm'. We do have to remove the print statements though.

qutip/solve/steadystate.py
Ericgig: This was changed in #1802.

Ericgig:
I started patching documentation:
- Updated import from moved files (`from qutip.states` --> `from qutip.core.states`).
- Changes `Options` to `SolverOptions`.
- Commented some calls to function and method that do not exist anymore (`QobjEvo.to_list`).

There are a lots of warning when building documentattion, but actual error seems to be limited to qip and HEOM changes that where not properly merged.

qutip/solve/steadystate.py
AGaliciaMartinez: This is definitely where the error occurs. I compared the output of `sp_permute` (version in master) with `_permute` and they do not return the same output for the same matrices and permutation indices. I will include a fix but lets talk about it later in the meeting.

Ericgig:
I fixed steadystate tests:
- `_permute` use numpy array, @AGaliciaMartinez is looking to fix the data layer function.
- Lowered the precision required for `test_exact_solution_for_simple_methods`. The test looked for `1e-16 +/- 1e-16` and most calls returned `0`. Most methods cannot match this precision and can fail from numerical error.

Ericgig:
@hodgestar When you have time, could you take a look at the documentation build? It works fine locally but fail on here...

AGaliciaMartinez:
It turns out that the issue with permute is that what previously was called reverse_permute is now permute. The reverse permute was already being computed so a fix can be to just call the previous permute function with `rev_perm` when using rcm: `_permute(L, rev_perm, rev_perm)`.

qutip/solve/steadystate.py
```diff
@@ -76,17 +78,29 @@ def _permute(matrix, rows, cols):
...
+    # TODO: restore usage of _data.permute when fixed
+    new = as_np[rows, :][:, cols]
+
+    if isinstance(matrix, _data.Data):
+        return _data.to('csr', _data.Dense(new))
+    elif scipy.sparse.issparse(matrix):
+        print(new)
+        return matrix.__class__(new)
+    elif isinstance(perm, np.ndarray):
+        return new
```
AGaliciaMartinez, line 103:
```suggestion
    if not np.any(rows):
        rows = np.arange(matrix.shape[0])
    if not np.any(cols):
        cols = np.arange(matrix.shape[1])

    rows = np.argsort(rows)
    cols = np.argsort(cols)

    shape = matrix.shape
    if isinstance(matrix, scipy.sparse.csc_matrix):
        rows, cols = cols, rows
        shape = (shape[1], shape[0])
    temp = _data.CSR((matrix.data, matrix.indices, matrix.indptr),
                     shape=shape)
    temp = _data.permute.indices_csr(temp, rows, cols).as_scipy()
    if isinstance(matrix, scipy.sparse.csr_matrix):
        return temp
    return scipy.sparse.csc_matrix((temp.data, temp.indices, temp.indptr),
                                   shape=matrix.shape)

```

AGaliciaMartinez: Same code as before, using sparse matrices, but with the reverse permutation. I believe `indices_csr` does what is supposed to do (from the docstring).

Ericgig: Thank you for finding this and making the fix.

hodgestar (approved):

hodgestar (approved):
Still looks good to me!

PR #1854, opened 2022-04-12
Activate remaining tests on Windows.

hodgestar:
**Description**
Activate remaining tests on Windows now that pytest-timeout has been removed.

**Related issues or PRs**
- Continues on from #1853

**Changelog**
Activate remaining tests on Windows.

### DIFF
--- a/qutip/qobjevo.py
+++ b/qutip/qobjevo.py
@@ -396,10 +396,7 @@ class QobjEvo:
     def __init__(self, Q_object=[], args={}, copy=True,
                  tlist=None, state0=None, e_ops=[]):
         if isinstance(Q_object, QobjEvo):
-            if copy:
-                self._inplace_copy(Q_object)
-            else:
-                self.__dict__ = Q_object.__dict__
+            self._inplace_copy(Q_object)
             if args:
                 self.arguments(args)
                 for i, dargs in enumerate(self.dynamics_args):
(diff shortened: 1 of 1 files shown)


hodgestar:
It looks like the mcsolve tests hang if the interpolate tests are run. This is probably related to the use of multiprocessing / mcsolve inside both.

hodgestar:
Ah, we seem to have rediscovered #1202.

Ericgig:
Could you try to have `parallel_map` redirect to `serial_map` in parallel.py to see if it makes it better.

hodgestar:
I've opened #1855 to disable the mcsolve tests on Windows for now. I intend to leave this issue open until after 4.7 is released and to see what happens with the mcsolve in QuTiP 5.

hodgestar:
In [3e530c0](https://github.com/qutip/qutip/pull/1854/commits/3e530c01d9cf686fc195c38c2cc26d0ee5bccfa8) I borrowed @nonhermitian's suggestion and switched to `concurrent.futures.ProcessPoolExecutor` in the hopes that the core Python developers have figured out how to make that works on Windows too.

It also simplifies the management of the pool of workers, which is a win by itself.

hodgestar:
Switching to ProcessPoolExecutor seems to have improved parallel_map quite a bit -- now all of the errors seem to be related to compiling coefficients.

hodgestar:
I've given up on solving this for QuTiP 4 for now -- it leads down a long rabbit hole that ends in fixing string coefficient generation, which was a big re-write and has already been done in QuTiP 5.

I'm continue the work for QuTiP 5 in #1948.

PR #1855, opened 2022-04-13
Disable mcsolve tests on windows.

hodgestar:
**Description**
Disable mcsolve tests on windows.

**Related issues or PRs**
- See #1202

**Changelog**
Disable mcsolve tests on windows. The mcsolve solver uses multprocessing under the hood to simulate many trajectories at once and this can deadlock on Windows.


hodgestar:
I'm going to merge this so that tests pass repeatably on master again and so that I can proceed with the 4.7 release, but we should keep #1854 open and address it for QuTiP 5 (if tests don't just pass there already).

PR #1860, opened 2022-04-13
Remove v4's brmesolve

Ericgig:
**Description**
- Remove the old version of bloch redfield functions in `qutip/solve/bloch_redfield.py`, the new ones being at `qutip/solver/brmesolve.py` added in #1651.
- Remove utilities for previous management of time dependent systems and string compilation. These functionalities are included in `QobjEvo` and `Coefficient`. (`_pyxbuilder.py`, `solve/_utilities.py`).
- Remove `interpolate.py`, `Cubic_Spline` was only used in `brmesolve` since  #1725.


qutip/__init__.py
```diff
@@ -83,6 +62,7 @@
 
 from .core import *
 from .solve import *
+from .solver.brmesolve import *
```
hodgestar, line 65: Should this not be imported via `from .solve import *` as we do for the rest of the solvers? Or is there a reason to make `brmesolve` special?

qutip/solver/brmesolve.py
```diff
@@ -3,7 +3,7 @@
 equation.
 """
 
-__all__ = ['mesolve', 'MeSolver']
+__all__ = ['brmesolve', 'BRSolver']
```
hodgestar, line 6: Doh. :)

hodgestar (approved):
Nice! Glad to see these old things gone.

I left one question about whether brmesolve can be included in `qutip/__init__.py` the same way as the other solvers (i.e. via `from qutip.solve import *`).

qutip/__init__.py
Ericgig: It is in `qutip.solver` and not in `qutip.solve`. The plan is to do `from .solver import *`, but for now that would import 2 version of `sesolve` and `mesolve`.

hodgestar: Okay. Happy with it to stay how it is for now. What is the difference between qutip.solve and qutip.solver? Maybe they need better names.

PR #1861, opened 2022-04-13
Shorter krylov tests

Ericgig:
**Description**
Tests related to krylovsolve take a lot of time, close to 1/3 of the whole test suite.
I kept all the tests, but reduced the size of the systems used.

### DIFF
--- a/qutip/tests/solve/test_krylovsolve.py
+++ b/qutip/tests/solve/test_krylovsolve.py
@@ -16,9 +16,9 @@ np.random.seed(0)
 
 
 @pytest.fixture(params=[
-    pytest.param(100, id="small dim"),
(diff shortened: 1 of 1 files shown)


hodgestar (approved):

hodgestar:
@Ericgig Approved, because faster tests are good, but I am wondering if the krylovsolve tests are much slower on v5 because tests didn't seem to timeout on v4?

Ericgig:
Yes they are, they seems about 3x slower.
I am not sure why, but my guess would be the eigen solving being run in sparse instead of dense for not that sparse Hamiltonian in tests. We should be careful about it when porting the solver to v5.

PR #1863, opened 2022-04-14
Fix specifying of wheel to install during post build verification

hodgestar:
**Description**
Fix specifying of wheel to install during post build verification.

When we switched to a newer cibuildwheel the tags of the built Linux wheels changed and broke the pre-upload to PyPI verification step. This fixes the name of the wheel to match that of the ones built by the newer cibuildwheel. 

**Related issues or PRs**
- None

**Changelog**
Fix specifying of wheel to install during post build verification.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,7 +167,7 @@ jobs:
       - name: Verify this is not a dev version
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


hodgestar:
I wasn't sure whether to make this PR against master or dev-major, but I left it on master for now.

Ericgig (approved):

PR #1866, opened 2022-04-14
Fix `isherm_csr` wrongly detecting structure

### LINKED ISSUE #1865: Bug in `isherm_csr`
### Bug Description

The test `Test_isherm.test_compare_implicit_zero_random` can sometime fail randomly.
I ran in locally 1000000 times and it failed 36 times.
It usually fails at high density (>90%), but can fail with a density of 75%.
https://github.com/qutip/qutip/runs/6012031509

### Code to Reproduce the Bug

_No response_

### Code Output

_No response_

### Expected Behaviour

It should always pass.

### Your Environment

```shell
Seen in "ubuntu-latest, python3.10, Python 3.10" test job on `dev.major`.
Reproduced locally on python 3.9.
```

### Additional Context

_No response_

Ericgig:
**Description**
`isherm_csr` did not fully tested the structure, it only checked that each line had as many elements in it than the corresponding column. So 
```
[[1, 1, 0],
 [0, 1, 1],
 [1, 0, 1]]
```
was seen as hermician.

This added a second layer of structure check for such cases and a test that would previously fail.

**Related issues or PRs**
Fixes #1865

### DIFF
--- a/qutip/core/data/properties.pyx
+++ b/qutip/core/data/properties.pyx
@@ -135,6 +135,8 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
                 # Pointer into the "transposed" matrix.
                 ptr_t = out_row_index[col]
                 out_row_index[col] += 1
+                if nrows != matrix.row_index[ptr_t]:
+                    return _isherm_csr_full(matrix, tol)
                 # We tested the structure already, so we can guarantee that
                 # these two elements correspond.
                 if not _conj_feq(matrix.data[ptr], matrix.data[ptr_t], tol):
--- a/qutip/core/data/properties.pyx
+++ b/qutip/core/data/properties.pyx
@@ -137,8 +137,6 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
                 out_row_index[col] += 1
                 if row != matrix.col_index[ptr_t]:
                     return _isherm_csr_full(matrix, tol)
-                # We tested the structure already, so we can guarantee that
-                # these two elements correspond.
                 if not _conj_feq(matrix.data[ptr], matrix.data[ptr_t], tol):
                     return False
         return True
--- a/qutip/core/data/properties.pyx
+++ b/qutip/core/data/properties.pyx
@@ -135,7 +135,7 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
                 # Pointer into the "transposed" matrix.
                 ptr_t = out_row_index[col]
                 out_row_index[col] += 1
-                if nrows != matrix.row_index[ptr_t]:
+                if row != matrix.col_index[ptr_t]:
                     return _isherm_csr_full(matrix, tol)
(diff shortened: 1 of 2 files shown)


qutip/core/data/properties.pyx
```diff
@@ -135,6 +135,8 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
                 # Pointer into the "transposed" matrix.
                 ptr_t = out_row_index[col]
                 out_row_index[col] += 1
+                if nrows != matrix.row_index[ptr_t]:
+                    return _isherm_csr_full(matrix, tol)
```
hodgestar, line 139:
```suggestion
                if row != matrix.col_index[ptr_t]:
                    return False
```
The code as it was in the PR seems wonky -- `.row_index[...]` should be indexed by a row number, not by a ptr, and wouldn't always be `nrows` in any case, so the code falls back to `_isherm_csr_full` almost always (even when the matrix is hermitian).

The suggested code checks that the column of the `ptr_t` is the correct one and thus avoids having to fall back to `_isherm_csr_full`.

This logic is quite tricky, so please check mine too!

hodgestar (changes requested):
Good job catching this error. It would have been bad to release 5.0 with this bug undetected.

qutip/core/data/properties.pyx
Ericgig: Yes this test was wrong, it would always call `_isherm_csr_full`. Your test work, thank you for catching. But we must still fall back to `_isherm_csr_full` since we can have a number under the `tol` on one side and nothing on the other.

qutip/core/data/properties.pyx
```diff
@@ -135,6 +135,8 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
                 # Pointer into the "transposed" matrix.
                 ptr_t = out_row_index[col]
                 out_row_index[col] += 1
+                if row != matrix.col_index[ptr_t]:
+                    return _isherm_csr_full(matrix, tol)
                 # We tested the structure already, so we can guarantee that
                 # these two elements correspond.
```
hodgestar, line 141: I think we can delete this comment now, because we explicitly check that the row and column match above (i.e. the comment was incorrect before and untrue now).

Ericgig:
```suggestion
```

qutip/core/data/properties.pyx
```diff
@@ -135,6 +135,8 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
                 # Pointer into the "transposed" matrix.
                 ptr_t = out_row_index[col]
                 out_row_index[col] += 1
+                if row != matrix.col_index[ptr_t]:
+                    return _isherm_csr_full(matrix, tol)
```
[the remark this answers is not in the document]
Ericgig: If we are leaving this to later, do you approve this for now?

hodgestar (approved):

PR #1871, opened 2022-04-18
quantum circuits inherit user gates from `add_circuit`

Canoming:
**Description**
The `add_circuit` function in `QuantumCircuit` class used to copy the names of all gates without add the definition of user gates to the target circuit. Hence the user gates cannot be recognized by the new circuit. I made the copy of the definition automatic.

### DIFF
--- a/qutip/qip/circuit.py
+++ b/qutip/qip/circuit.py
@@ -556,8 +556,7 @@ class QubitCircuit:
         arg_label : string
             Label for gate representation.
         """
-        if name not in ["RX", "RY", "RZ", "SNOT", "SQRTNOT", "PHASEGATE",
-                        "X", "Y", "Z", "S", "T", "QASMU"]:
+        if name not in _single_qubit_gates:
             raise ValueError("%s is not a single qubit gate" % name)
 
         if qubits is not None:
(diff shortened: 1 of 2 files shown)


qutip/qip/circuit.py
```diff
@@ -593,6 +593,13 @@ def add_circuit(self, qc, start=0):
         if self.N - start < qc.N:
             raise NotImplementedError("Targets exceed number of qubits.")
 
+        # Inherit the user gates
+        for user_gate in qc.user_gates:
+            if user_gate in self.user_gates and overwrite_user_gates:
+                self.user_gates[user_gate] = qc.user_gates[user_gate]
+            else:
+                self.user_gates[user_gate] = qc.user_gates[user_gate]
```
BoxiLi, line 601: They seem to do the same thing regardless of the if branch. Maybe you meant to do nothing if `overwrite_user_gates==False`?

BoxiLi (changes requested):
Thanks @Canoming! Could you also add a test for this? Probably using the existing `test_user_gate` in `test_circuit.py` or adding a new one.

qutip/qip/circuit.py
Canoming: Ah, sorry, I've fixed the problem.

---

I just added a test into the `test_add_circuit`. I just found that the offset was ignored when adding user defined gates. Indeed, classifying the type of gates is unnecessary here. The arguments of the operators are well defined in the original circuit. I've cleaned up the messy branches.

By the way, is there a particular reason to set the default values of the targets and controls to `None` instead of empty lists? Since the unnecessary arguments are ignored in simulation, it seems loses nothing by setting the default values to empty sets. Since we are iterating over the targets and controls when we move the gates around, it should be simpler if we can iterate over the default value. But I'm not sure how big the change would be. The nullity checks exist in many places, and they have to be modified if we change the default values.

qutip/qip/circuit.py
```diff
@@ -593,47 +592,36 @@ def add_circuit(self, qc, start=0):
...
-                                  [gate.targets[0] + start,
-                                   gate.targets[1] + start],
-                                  gate.controls + start, None, None)
-                elif gate.name in self.user_gates:
-                    self.add_gate(
-                        gate.name, targets=gate.targets,
-                        arg_value=gate.arg_value)
-            else:
-                measurement = circuit_op
+                # gate = circuit_op
```
BoxiLi, line 604: This line can be removed.

qutip/qip/circuit.py
```diff
@@ -593,47 +592,36 @@ def add_circuit(self, qc, start=0):
...
+                    circuit_op.name, targets=tar, controls=ctrl,
+                    arg_value=circuit_op.arg_value)
+            elif isinstance(circuit_op, Measurement):
                 self.add_measurement(
-                                measurement.name,
-                                targets=[measurement.targets[0] + start],
-                                classical_store=measurement.classical_store)
+                                circuit_op.name,
+                                targets=[target + start for target in circuit_op.targets],
+                                classical_store=circuit_op.classical_store)
```
BoxiLi, line 622: We can remove some indents, 16 spaces are unnecessary and it also makes code climate complain.

BoxiLi (commented):
Looks good. Only some styling issue that needs to be fixed.

qutip/qip/circuit.py
BoxiLi: > By the way, is there a particular reason to set the default values of the targets and controls to None instead of empty lists? 

I don't think there is a particular reason. I kind of agree that replacing them with an empty list is a better option. But it is gonna be a "large" scale change as it exists in many places (which also means that `None` was indeed a bad choice). So it is better done in the new qutip-qip rather than here.

qutip/qip/circuit.py
Canoming: Sure. I've shortened the lines

BoxiLi (approved):

BoxiLi:
Thanks @Canoming

PR #1874, opened 2022-04-21
Add option at setup to build with int 64

Ericgig:
**Description**
Adds a --with-idxint-64 compilation option for building with 64-bit integer indices in the data layer. (Alternative to #1727)
Instead of using a compile time environment variable, it contains files for both 32 and 64 int and choose at setup which to use.

This way has the advantage that cython can keep track of the version used and recompile all file depending on it when changed.
With compile time environment variable, you can have file compiled with different type of int and only get error when trying to use them. Also 'cimporting' data is simpler since you don't need to define the variable.

However, changing all the import location made diff a lot bigger than #1727.

`qutip.settings.install['intsize']` will return 32 or 64 depending on the size of int used.

**Changelog**
Added a compilation option for compiling with 64-bit integer.


hodgestar:
@Ericgig Can we change this so `base` includes the index type (by importing the new file)? That way we don't have to import the strange new file in 30 places.

hodgestar:
@Ericgig Another suggestion -- if we use a `.h` file instead of a `.pxd` file to hold the setting we should be able get away with a single file that can be included in "base.pxd" with `cdef extern from "idxtype.h": ...` and not need the `.pyx`.

Ericgig:
@hodgestar the `.h` file imported from `base.pxd` makes is a lot simpler.
Changing the header file force recompilation, but the cythonization step is skipped.
If tests passes, it's ready for review.

qutip/core/data/base.pxd
```diff
@@ -1,10 +1,9 @@
 #cython: language_level=3
 
-import numpy as np
-cimport numpy as cnp
+cdef extern from "src/intdtype.h":
+    # cython is smart enough to understand this int can be 32 or 64 bits.
+    ctypedef int idxint
```
hodgestar, line 5: In `intdtype.h` we can do:
```c
#define QUTIP_INT_SIZE 32
```
and then inside this `cdef extern` block add:
```c
int QUTIP_INT_SIZE
```
Of course `QUTIP_INT_SIZE` is not really an integer, but rather a constant, but Cython only cares about the type (i.e. that the constant can be used where an int can be used).

This will take away the need for the cunning `size_check` logic and will also make setting the value in `qutip.settings` simpler if we expose it.

qutip/installsettings.py
```diff
@@ -115,5 +117,7 @@ class InstallSettings:
         # configured, rather than through _logging.get_logger().
         "_logger": _logger,
         # Running on mac with openblas make eigh unsafe
-        "eigh_unsafe": _eigh_unsafe
+        "eigh_unsafe": _eigh_unsafe,
+        # int size for the shape of data layer object
+        "intsize": None
```
hodgestar, line 122: I suggest calling this `idxint_size` for clarity.

hodgestar (changes requested):
I left a couple of small suggestions, but I think this is starting to look great.

qutip/installsettings.py
```diff
@@ -115,5 +117,7 @@ class InstallSettings:
         # configured, rather than through _logging.get_logger().
         "_logger": _logger,
         # Running on mac with openblas make eigh unsafe
-        "eigh_unsafe": _eigh_unsafe
+        "eigh_unsafe": _eigh_unsafe,
+        # int size for the shape of data layer object
+        "idxint_size": None
```
hodgestar, line 122: Could we set idxint_size here rather than in `__init__`?

Ericgig: Sadly no, it create a circular import.
We may be able to set it in `qutip/core/data/base.pyx` instead of `__init__.py`.

hodgestar: Oh well, too bad.

hodgestar (approved):
Approved!

PR #1879, opened 2022-04-25
Solver contain unique integrator 

Ericgig:
**Description**
`Solver` create the integrator at `__init__` instead of in `run` and `start`.
`step` and `run` no longer take an `options` parameter.  `solver.options = new_options` will now reset the integrator to use the new options. However changing options will always reset the integrator.

**Related issues or PRs**
Similar improvement where first made in #1710, splitting it to made that PR simpler.

**Changelog**
Sovlers manage options as a property instead of method's parameter.

### DIFF
--- a/qutip/solver/solver_base.py
+++ b/qutip/solver/solver_base.py
@@ -27,7 +27,6 @@ class Solver:
     name = "generic"
 
     # State, time and Integrator of the stepper functionnality
-    is_set = False
     _integrator = None
     _avail_integrators = {}
 
@@ -125,7 +124,6 @@ class Solver:
         _time_start = time()
         _data0 = self._prepare_state(state0)
         self._integrator.set_state(tlist[0], _data0)
-        self.is_set = True
         self._argument(args)
         self.stats["preparation time"] += time() - _time_start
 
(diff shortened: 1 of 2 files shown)


hodgestar:
This looks great and I think it will help a lot in cleanly implementing more complex solvers light the HEOMSolver. I have two suggestions for future improvements:

- I'd like to see `.options = ...` become a method because it now modifies the existing integrator and `.options(...)` would both make that clearer and more flexible. E.g. with a method we could allow both `.options(new_options)` to completely override the options and `.options(key=new_value)` to update just one setting.

- The first call to `result.add(t, X)` uses `X = state0` but state0 could be almost anything the user passed in (and not necessarily the canonical form returned by `_restore_state`). It would be better if either there was a separate method to override to convert `state0` into the canonical form, or `result.add(t, X)` was called initially with `X = _restore_state(_data0)`.

Ericgig:
@hodgestar, I consider the second point as a bug.
Do you want me to fix it here or do it elsewhere? (and add swap-able `Result` while at it?)

The options method will wait for what will become of #1812.

hodgestar:
> @hodgestar, I consider the second point as a bug. Do you want me to fix it here or do it elsewhere? (and add swap-able `Result` while at it?)

Happy for you to fix it here if it doesn't complicate things too much. It was a bug before this PR though, so also happy for it to be fixed later.

Ericgig:
The fixes are quite simple, added them here.

hodgestar (approved):
Thanks!

PR #1880, opened 2022-04-25
Fix bug in `IntergratorScipyZvode` when starting integration at time other than `0`

Ericgig:
**Description**
`IntergratorScipylsoda.mcstep` read the integration range from the scipy solver's `rwork`. However `rwork` is initialized to 0 when initializing the state and only take the expected value when the first step in done. Therefore it would fail when starting from time other than `0`. `IntergratorScipyZvode` also used `rwork` in an unsafe way.

Fix that and add a test. 

**Related issues or PRs**
Similar improvement where first made in https://github.com/qutip/qutip/pull/1710, splitting it to made that PR simpler.

**Changelog**
`IntergratorScipylsoda` safe for starting at any time.

### DIFF
--- a/qutip/solver/ode/scipy_integrator.py
+++ b/qutip/solver/ode/scipy_integrator.py
@@ -110,19 +110,26 @@ class IntegratorScipyZvode(Integrator):
         if self._ode_solver.t == t:
             # Exact same `t` as the last call, nothing to do.
             pass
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
The less we fiddle with `rwork` the better!

Issue #1875, opened 2022-04-23
setup.py --with-openmp option not compatible with PEP517 (pyproject.toml) build


drew-parsons:
### Bug Description

setup.py manages a `--with-openmp` build option to provide openmp support.

pyproject.toml was introduced in  QuTiP 4.6.0 to allow PEP517 build without (directly) using setup.py.

But the `--with-openmp` option that previously worked with a setup.py build does not seem to be compatible with a PEP517 build.

The problem shows up in the Debian build of the qutip package. Debian provides its own tool (pybuild from the dh-python package) to manage the build of python packages.  PEP517 is relatively new, so by default pybuild uses setup.py if it is available.  pybuild will use PEP517 instead, if the pybuild-plugin-pyproject package is installed and pyproject.toml is available.

Without pybuild-plugin-pyproject installed, the Debian build was invoking
```
/usr/bin/python3.10 setup.py build --with-openmp
```
using setup.py and specifying `--with-openmp`.  A successful build log for this method can be found at https://buildd.debian.org/status/fetch.php?pkg=qutip&arch=amd64&ver=4.6.2-3&stamp=1641397514&raw=0

With pybuild-plugin-pyproject installed, the Debian build invokes a wheel build (command line given below).  But `--with-openmp` causes the build to fail.

### Code to Reproduce the Bug

```shell
python3.10 -m build --skip-dependency-check --no-isolation --wheel --outdir /build/qutip/.pybuild/cpython3_3.10_qutip --with-openmp
```

### Code Output

```shell
usage: python -m build [-h] [--version] [--sdist] [--wheel] [--outdir OUTDIR] [--skip-dependency-check] [--no-isolation] [--config-setting CONFIG_SETTING] [srcdir]
python -m build: error: unrecognized arguments: --with-openmp
```

### Expected Behaviour

The PEP517 build should be able to be configured to support openmp.

### Your Environment

```shell
QuTiP Version:      4.6.2  (trying to build 4.7.0)
Numpy Version:      1.21.5
Scipy Version:      1.8.0
Cython Version:     0.29.28
Matplotlib Version: 3.5.1
Python Version:     3.10.4
Number of CPUs:     4
BLAS Info:          OPENBLAS
OPENMP Installed:   True  (from setup.py build, not PEP517 build)
INTEL MKL Ext:      False
Platform Info:      Linux (x86_64)
Installation path:  /usr/lib/python3/dist-packages/qutip
```

### Additional Context

[...]

PR #1881, opened 2022-04-25
Single definition for map_kw default

Ericgig:
**Description**
Parallel map options' defaults are set in both options.py and parallel.py. As remarked by Asier in #1710, they should be defined in only one place.
The default are now set in parallel.py. Everywhere else can set the value to `None` or not set the key to use the default.
Also the default number of cpus is obtain with `utilities.available_cpu_count` instead of `multiprocessing.cpu_count`.

**Related issues or PRs**
Originally in #1710

**Changelog**
Single definition for map_kw defaults.

### DIFF
--- a/qutip/solver/options.py
+++ b/qutip/solver/options.py
@@ -3,8 +3,6 @@ __all__ = ['SolverOptions',
            'McOptions']
 
 from ..optionsclass import optionsclass
-import multiprocessing
-import threading
 
 @optionsclass("solver")
 class SolverOptions:
@@ -233,8 +231,10 @@ class McOptions:
     map_options: dict
         keys:
             'num_cpus': number of cpus to use.
-            'timeout': maximum time for all trajectories. (sec)
-            'job_timeout': maximum time per trajectory. (sec)
+            'timeout': maximum time (sec) for all trajectories. ``None`` will
+                use the maximum timeout value.
+            'job_timeout': maximum time (sec) per trajectory. ``None`` will
+                use the maximum timeout value.
         Only finished trajectories will be returned when timeout is reached.
 
     mc_corr_eps : float {1e-10}
@@ -257,8 +257,8 @@ class McOptions:
         "mc_corr_eps": 1e-10,
 
         "map_options": {
-            'num_cpus': multiprocessing.cpu_count(),
-            'timeout': threading.TIMEOUT_MAX,
-            'job_timeout': threading.TIMEOUT_MAX,
+            'num_cpus': None,
+            'timeout': None,
+            'job_timeout': None,
         },
     }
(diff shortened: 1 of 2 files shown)


qutip/solver/options.py
```diff
@@ -257,8 +257,8 @@ class McOptions:
         "mc_corr_eps": 1e-10,
 
         "map_options": {
-            'num_cpus': multiprocessing.cpu_count(),
-            'timeout': threading.TIMEOUT_MAX,
-            'job_timeout': threading.TIMEOUT_MAX,
+            'num_cpus': None,
+            'timeout': None,
+            'job_timeout': None,
         },
```
AGaliciaMartinez, line 263: Would it work to have something on the lines:
```
options = {...}
options["map_options"] = qutip.solver.parallel.default_map_kw
```
?
This way both are the same and you do not need `_read_map_kw`.

AGaliciaMartinez (commented):
Great to see this being changed and thank you for opening a separate PR for it!

qutip/core/data/src/intdtype.h
```diff
@@ -1,2 +1,2 @@
-typedef int64_t idxint;
-int _idxint_size=64;
+typedef int32_t idxint;
+int _idxint_size=32;
```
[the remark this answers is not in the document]
Ericgig: This was a mistake when merging #1874. 
It was merged with the 64bit version of this file but each time we run setup.py it will overwrite it with this 32bit version.
So until we merge this 32bit version, it will appear in diff.

Also, since there are alternative build systems that ignore the setup.py (#1875), adding it to gitignore could cause issues...

qutip/solver/options.py
Ericgig: While it could work, I like the added sturdiness of `_read_map_kw`.
It allow to have the default to `None` instead of a mutable `dict` and serve a check that all desired key are there.
So is someone did `options["map_options"] ={'num_cpus': 2}`, it would still work.
Having `options["map_options"]` itself being a dict is risky since once overwritten, you can forget which keys it's supposed to have...

qutip/solver/parallel.py
```diff
@@ -10,21 +10,32 @@
...
+    'num_cpus': available_cpu_count(),
 }
 
 
+def _read_map_kw(options):
+    options = options or {}
+    map_kw = default_map_kw.copy()
+    for key in map_kw:
+        if options.get(key) is not None:
+            map_kw[key] = options[key]
```
hodgestar, line 33:
```suggestion
    map_kw.update({k: v for k, v in options.items() if v is not None})```
```
I think these days we can write this more succinctly.

```

hodgestar (changes requested):
Looks good. I left one suggestion for a slightly nicer way to update the options.

Could we also document map_kw while we're here? Currently it's either undocumented or documented just as "Other options" which makes using it quite unclear.

AGaliciaMartinez (approved):

hodgestar (approved):
Also approved!

PR #1884, opened 2022-04-26
QobjEvo's call take kwargs

Ericgig:
**Description**
We updated the signature for arguments in coefficient from dict to keyword format, changing call from `coeff(t, args)` to `coeff(t, **args)`. However `QobjEvo`, which have a similar call taking time and argument still took the new args as dictionary instead of keyword. Added keywork support for `QobjEvo.__call__` and `QobjEvo.arguments` . The `__init__` still take argument as a dict since it already takes a lot of parameters.

**Related issues or PRs**
Follow up to #1633 

**Changelog**
QobjEvo call accept keywords for new args.

### DIFF
--- a/qutip/core/cy/qobjevo.pyx
+++ b/qutip/core/cy/qobjevo.pyx
@@ -293,6 +293,26 @@ cdef class QobjEvo:
         return out
 
     def __call__(self, double t, dict _args=None, **kwargs):
+        """
+        Get the :cls:`Qobj` at ``t``.
+
+        parameter
+        ---------
+        t : float
+            Time at which the ``QobjEvo`` is to be evalued.
+
+        args : dict [positional / optional]
+            New arguments as a dict. Update args with ``arguments(new_args)``.
+
+        **kwargs :
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
+
+        .. note:
+            If both the positional ``args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``args``
+            dict value will take priority.
+        """
         if _args is not None or kwargs:
             if _args:
                 kwargs.update(_args)
@@ -326,7 +346,23 @@ cdef class QobjEvo:
         return QobjEvo(self, compress=False)
 
     def arguments(QobjEvo self, dict _args=None, **kwargs):
-        """Update the arguments"""
+        """
+        Update the arguments.
+
+        parameter
+        ---------
+        args : dict [positional / optional]
+            New arguments as a dict. Update args with ``arguments(new_args)``.
+
+        **kwargs :
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
+
+        .. note:
+            If both the positional ``args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``args``
+            dict value will take priority.
+        """
         if _args:
             kwargs.update(_args)
         cache = []
--- a/qutip/core/cy/qobjevo.pyx
+++ b/qutip/core/cy/qobjevo.pyx
@@ -301,7 +301,7 @@ cdef class QobjEvo:
         t : float
             Time at which the ``QobjEvo`` is to be evalued.
 
-        args : dict [positional / optional]
+        _args : dict [optional]
             New arguments as a dict. Update args with ``arguments(new_args)``.
 
         **kwargs :
@@ -309,12 +309,12 @@ cdef class QobjEvo:
             ``arguments(**new_args)``.
 
         .. note:
-            If both the positional ``args`` and keywords are passed new values
-            from both will be used. If a key is present with both, the ``args``
+            If both the positional ``_args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``_args``
             dict value will take priority.
         """
         if _args is not None or kwargs:
-            if _args:
+            if _args is not None:
                 kwargs.update(_args)
             return QobjEvo(self, args=kwargs)(t)
         return Qobj(
@@ -351,7 +351,7 @@ cdef class QobjEvo:
 
         parameter
         ---------
-        args : dict [positional / optional]
+        _args : dict [optional]
             New arguments as a dict. Update args with ``arguments(new_args)``.
 
         **kwargs :
@@ -359,11 +359,11 @@ cdef class QobjEvo:
             ``arguments(**new_args)``.
 
         .. note:
-            If both the positional ``args`` and keywords are passed new values
-            from both will be used. If a key is present with both, the ``args``
+            If both the positional ``_args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``_args``
             dict value will take priority.
         """
-        if _args:
+        if _args is not None:
             kwargs.update(_args)
         cache = []
         self.elements = [
--- a/qutip/core/cy/qobjevo.pyx
+++ b/qutip/core/cy/qobjevo.pyx
@@ -292,9 +292,11 @@ cdef class QobjEvo:
 
         return out
 
-    def __call__(self, double t, dict args=None):
-        if args:
-            return QobjEvo(self, args=args)(t)
+    def __call__(self, double t, dict _args=None, **kwargs):
+        if _args is not None or kwargs:
+            if _args:
+                kwargs.update(_args)
+            return QobjEvo(self, args=kwargs)(t)
         return Qobj(
             self._call(t), dims=self.dims, copy=False,
             type=self.type, superrep=self.superrep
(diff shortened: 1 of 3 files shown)


AGaliciaMartinez (commented):
I like the change in the syntax even if its not backwards compatible. It feels more pythonic in this way. I wonder thought if we should keep `_args`. I guess you left it there to  not break code that uses this argument as positional. To keep the code that uses QobjEvo intact. But code that uses this as keyword argument will break any ways so I suggest to stick just to `kwargs` and try change instances in QuTiP code that use QobjEvo. Do you find that feasible @Ericgig ? Not sure how many things would break.

qutip/core/cy/qobjevo.pyx
```diff
@@ -292,9 +292,11 @@ cdef class QobjEvo:
...
         return out
 
-    def __call__(self, double t, dict args=None):
-        if args:
-            return QobjEvo(self, args=args)(t)
+    def __call__(self, double t, dict _args=None, **kwargs):
+        if _args is not None or kwargs:
+            if _args:
+                kwargs.update(_args)
+            return QobjEvo(self, args=kwargs)(t)
```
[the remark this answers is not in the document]
Ericgig: In coefficient we also kept both. 
This kind ob backward compatibility is free and will help users pass from v4 to v5.

We already add `args` starting with "_" in a few places. Somewhere is the documentation we say that `args` cannot start with this, so there is minimal risk of name collision on `_args`. Sadly cython does not support positional only parameter...

hodgestar (approved):

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
AGaliciaMartinez: I personally still think that we should favour the `kwargs` syntax and  remove the `_args` argument when possible. I would not consider this backward compatibility completely free as it will potentially confuse new users (even with the documentation) and even QuTiP 4 users as `qobj(t, args={})` will not work

In any case, if you already discussed this with @hodgestar and you both agree on this, I will not delay this PR on being merged.

AGaliciaMartinez (approved):

qutip/core/cy/qobjevo.pyx
```diff
@@ -323,11 +345,29 @@ cdef class QobjEvo:
...
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
+
+        .. note:
+            If both the positional ``args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``args``
+            dict value will take priority.
+        """
+        if _args:
+            kwargs.update(_args)
```
AGaliciaMartinez, line 367:
```suggestion
        if _args is not None:
            kwargs.update(_args)
```
If by any chance `_args=0` this will not raise an error and it should. Small thing that will most likely neve occur but best to have it.

qutip/core/cy/qobjevo.pyx
```diff
@@ -292,9 +292,31 @@ cdef class QobjEvo:
...
+        """
+        Get the :cls:`Qobj` at ``t``.
+
+        parameter
+        ---------
+        t : float
+            Time at which the ``QobjEvo`` is to be evalued.
+
+        args : dict [positional / optional]
+            New arguments as a dict. Update args with ``arguments(new_args)``.
```
AGaliciaMartinez, line 305:
```suggestion
        _args : dict [positional / optional]
            New arguments as a dict. Update args with ``arguments(_args)``.
```
Given that the argument is actually `_args` I would include it in that way in the docstring.

qutip/core/cy/qobjevo.pyx
```diff
@@ -292,9 +292,31 @@ cdef class QobjEvo:
...
+
+        **kwargs :
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
+
+        .. note:
+            If both the positional ``args`` and keywords are passed new values
```
AGaliciaMartinez, line 312:
```suggestion
            If both the positional ``_args`` and keywords are passed new values
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -292,9 +292,31 @@ cdef class QobjEvo:
...
+            from both will be used. If a key is present with both, the ``args``
```
AGaliciaMartinez, line 313:
```suggestion
            from both will be used. If a key is present with both, the ``_args``
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -323,11 +345,29 @@ cdef class QobjEvo:
...
+        Update the arguments.
+
+        parameter
+        ---------
+        args : dict [positional / optional]
+            New arguments as a dict. Update args with ``arguments(new_args)``.
+
+        **kwargs :
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
```
AGaliciaMartinez, line 359:
```suggestion
            ``arguments(**kwargs)``.
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -323,11 +345,29 @@ cdef class QobjEvo:
...
+
+        .. note:
+            If both the positional ``args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``args``
```
AGaliciaMartinez, line 363:
```suggestion
            If both the positional ``_args`` and keywords are passed new values
            from both will be used. If a key is present with both, the ``_args``
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -292,9 +292,31 @@ cdef class QobjEvo:
...
+        ---------
+        t : float
+            Time at which the ``QobjEvo`` is to be evalued.
+
+        args : dict [positional / optional]
+            New arguments as a dict. Update args with ``arguments(new_args)``.
+
+        **kwargs :
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
```
AGaliciaMartinez, line 309:
```suggestion
            New arguments as a keywors. Update args with
            ``arguments(**kwargs)``.
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -323,11 +345,29 @@ cdef class QobjEvo:
...
-    def arguments(QobjEvo self, dict new_args):
-        """Update the arguments"""
+    def arguments(QobjEvo self, dict _args=None, **kwargs):
+        """
+        Update the arguments.
+
+        parameter
+        ---------
+        args : dict [positional / optional]
+            New arguments as a dict. Update args with ``arguments(new_args)``.
```
AGaliciaMartinez, line 355:
```suggestion
        _args : dict [positional / optional]
            New arguments as a dict. Update args with ``arguments(_args)``.
```

AGaliciaMartinez (commented):
I added a few suggestions in the docstring involving the use of `_args` instead of `args` to match the actual name of the argument.

qutip/core/cy/qobjevo.pyx
Ericgig:
```suggestion
        _args : dict [optional]
            New arguments as a dict. Update args with ``arguments(new_args)``.
```

qutip/core/cy/qobjevo.pyx
Ericgig:
```suggestion
        _args : dict [optional]
            New arguments as a dict. Update args with ``arguments(new_args)``.
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -292,9 +292,31 @@ cdef class QobjEvo:
...
+            New arguments as a keywors. Update args with
+            ``arguments(**new_args)``.
+
+        .. note:
+            If both the positional ``args`` and keywords are passed new values
+            from both will be used. If a key is present with both, the ``args``
+            dict value will take priority.
+        """
+        if _args is not None or kwargs:
+            if _args:
```
Ericgig, line 317:
```suggestion
            if _args is not None:
```

Ericgig:
Thank you @AGaliciaMartinez for the documentation fixes.

AGaliciaMartinez (approved):

qutip/core/cy/qobjevo.pyx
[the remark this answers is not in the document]
hodgestar: The goal here is to try have at least one way to write code that works in both QuTiP 4 and 5 so that users might potentially support both for awhile without too much hardship.

Perhaps the coefficient function and QobjEvo __call__ situations are a bit different however. The coefficient functions are usually defined directly by the user, but I don't know how often users call QobjEvo objects directly if at all.

So maybe @AGaliciaMartinez is correct and we could simplify the signature of __call__ here?

Lots of little trade-offs to be made around these things to try make the lives of our users as painless as possible.

PR #1885, opened 2022-04-26
Faster correlation test

Ericgig:
**Description**
Lower the dims and length of the `tlist` used by `test_correlation_solver_equivalence[mc]`. Since we added a maximum time for the tests, this one test would fail from time to time. This make the test about 20% faster, which hopefully should be enough to have tests consistently passing.

**Changelog**
Faster `test_correlation_solver_equivalence` test.


hodgestar (approved):
Yay!

PR #1889, opened 2022-05-03
Result for mcsolve

Ericgig:
**Description**
Apply the changes to solver/result.py that were made previously for mcsolve in #1710. 
Also add other improvement to `Result`.

Major changes:
- `Result.__init__` signature changed from `(e_ops, options, _super, oper_state)` to `(e_ops, options, tlist, state0)`.
  To know if we can normalize the state, we use the first state's dims instead of asking for information about it.
  The first `add` is done in the `__init__` call.
- `MultiTrajResult` and `MultiTrajResultAveraged` are merged into `MultiTrajResult`: both classes were be interchangeable but did not share code, so any modification needed to be done twice.
- Add capacity to compute trajectories until a tolerance is reached. A new feature in #1710 was to allow to compute trajectories until the error bars on the expectation values are under a certain tolerance. The `add` method compute the error bars using jackknife re-sampling and return an estimation of the number of trajectories still needed to reach the desired tolerance. The mapping function can then stop computing trajectories when this reach `0`.
- `McResult` was created: `collapse` and `photocurrent` are an output of `mcsolve` only. So these were removed from `MultiTrajResult` and move into `McResult`.

Other changes:
- `MultiTrajResult.steady_state` is a function instead of a property. `steady_state` was the average of the states for all times. With long enough `tlist`, it should converge to the steady state, but the first times are usually far from it. As a function, it offer the possibility to average the last `N` states.
- When `e_ops` is a `Qobj`, `QobjEvo` or `callable`, the expect will not be a list but the array of expect:
  `Result(e_ops=[num], ...).expect[0] == Result(e_ops=num, ...).expect`.
  I am not too sure about that one, it feel logic but could break some codes.  

**Related issues or PRs**
#1710


hodgestar (commented):
I'm not a big fan of the new tlist and state0 arguments. I think it would be better if the result class were more of a dumb container and store_final_state and normalize_output were solver options instead.

All of this new logic around storing states twice and special handling for the initial state could go away.

Since I need to do something for results for the HEOM PR anyway, I will try write a simpler Result class as an alternative PR to this one.

Ericgig:
@hodgestar, 
I adapted the `MultiTrajResult` to the new `Result` format, and I have some changes to the `Result` class that I want to do:
- I changed the `Result.__init__` to move setting the `e_ops` in `_post_init`: `MultiTrajResult` can't use them and needed to remove the related state processors.
- `add` was overwritten, but we could use it if we add a `_early_finish_check`: a processor that return the estimated quantity of work remaining. Parallel map will stop computing trajectories when it get under 0. I think we could have it in `Result`: it could be interesting to have with `mesolve` with the options to stop when the steady state is reached, etc. Then if times when added in a state processor, I could use `add` without overloading.
Are you Ok with me going for it or do you prefer to keep `Result` as is?

hodgestar (approved):
Looks good!

hodgestar:
The test failures seem to be only coveralls result upload errors. I'm happy to merge without re-running. Let's just double check that the merge commits pass everything.

Ericgig:
@hodgestar, tests are passing, can I merge?

PR #1890, opened 2022-05-06
Remove qip submodule

Ericgig:
**Description**
Remove the QIP submodule that has been migrated to `qutip/qutip-qip`.

Most files were simply deleted, but there are a few non trivial parts:
- The gates used in tests were added to operators.py, all others were removed. The signature was changed: the `target` parameter was removed and `dtype` keywords added to match other operators. This means that there will be two version of `cnot` etc. available. 

- `qutip.enntropy.entangling_power` was also removed. While it was not strictly in the qip module, it is meant to be used for 2 qubits gates, so it felt more natural to be included there.

- The target qubits options for measurement was removed.  qip has it's own more specialized `Measurement` class, which I guess is the reference for such case. But we could make it stronger so that it can target various kind of subsystems, such as one site on a lattice or one of harmonic oscillator coupled to others, etc.

Bonus: The `-Werror` for pytest was restored.

**Related issues or PRs**
Checklist item in #1850

**Changelog**
Remove the qip submodule.


BoxiLi:
Thanks very much @Ericgig for preparing this!

I feel like it will be quite confusing if gates like `cnot` have different signatures than the ones in qutip-qip. More importantly, many of those gates/transformations are not only relevant to `qip`, but also used in other applications.

One of the reasons that we want to include those gates into `qutip.operators` is because a few other downstream packages (e.g. `krylovsolver` that is now merged into qutip, and `qutip-tensornetwork` also uses Hadamard transform)  also use them for testing (only) purposes. Having them in the main qutip may save them from depending on qutip-qip. I guess it would be better if we either
- Remove all qip and only define a few `_cnot`, `_swap` as private functions for testing inside the main qutip. And ask whoever else wants to use some gates to install `qutip-qip`. This may result in e.g. `qutip-tensornetwork` depending on `qutip-qip or having to implement a copy of some gates.

or
- Include some more commonly used gates like `rx`, `ry` (they are unitaries generated by the Paulis and directly relevant to quantum control) `iswap` etc and keep them identical to the ones in qutip-qip, so that those who just want some gates for testing does not have to download the full qutip-qip.


If the problem is the `qutip.qip.expand_operator`. I would argue that we can even include it into the main qutip. It is `tensor` + `permutation` and works not just for qubit systems, but also for other composite Hilbert spaces (provided a `dims`). It is quite handy not just for `qip`, but also for control and probably some other applications.

What do you think?

Ericgig:
@BoxiLi, 
After talking with @hodgestar, we saw that it would be nice to have some gates in qutip, so I tried to see what it could like if we kept them.
gates.py is in core with most gate present, but without function that modify gates, (rotation, expand, etc.), The signature remove `targets` but add `dtype`. `target` could become a feature of the `Gate` class in qip and be removed from these function on that side too to have the signature match. `expend_operator` is in tensor.py and require a `dims` input . `qft` is in operator.py and take `dimension` instead of `n_qubits`.

The gates are not in qutip namespace, but gates is: `qutip.gates.cnot` instead of `qutip.cnot`...
Please take a look and tell me what you think.
If it doesn't work, we can go back to removing them altogether.

BoxiLi:
Thanks!

> `target` could become a feature of the `Gate` class in qip and be removed from these function on that side too to have the signature match. 

Sounds like this could work.

If I understand correctly, at the side of qutip-qip, I should remove those gate functions and import them from qutip-v5 to keep things consistent? The only thing I see is that if someone is still using qutip-v4 + qutip-qip, I probably need to import them from qutip-v4 (which still has qutip.qip) and try to keep the `Gate` class working?

I'll have a closer look later. Just a few quick questions here:

- If the gates now follow the signature of other Qobj generating functions, why have them in a separate namespace?

- `hadamard_transform` kept somewhere in qutip or only in qutip-qip?

qutip/entropy.py
```diff
@@ -338,39 +337,3 @@ def participation_ratio(rho):
...
     else:
         return 1.0 / (rho ** 2).tr()
-
-
-def entangling_power(U):
-    """
-    Calculate the entangling power of a two-qubit gate U, which
-    is zero of nonentangling gates and 1 and 2/9 for maximally
-    entangling gates.
-
```
BoxiLi, line 348: What do we do with entangling power now? Do we still delete it or move it to `qutip.gate`

BoxiLi (commented):
Sorry for the delay. LGTM! left only one comment.

I tried briefly to update qutip-qip for this. Should be fine. Let's see how it works after it is merged.

Actually, many of the gates were in the main qutip namespace before 4.6 but it makes sense to give them a separate subspace just in case someone uses star importation.

qutip/entropy.py
BoxiLi: It is a bit restricted but might be useful for those who study quantum information.

hodgestar: We should keep `entangling_power` in `entropy`. It mentions the word "gate" in its docstring but really it calculate the entangling power of a two qubit operator, which is meaningful outside of the realm of QIP.

BoxiLi: I also don't entirely understand what it means, probably it should be from 0 to 2/9 from what I see below. But strangely for `swap` it is also 0...
```python
>>> qutip.entangling_power(qutip.cnot())
<stdin>:1: DeprecationWarning: Importing functions/classes of the qip submodule directly from the namespace qutip is deprecated. Please import them from the submodule instead, e.g.
from qutip.qip.operations import cnot

0.22222222222222227
>>> qutip.entangling_power(qutip.qeye([2,2]))
0.0
>>> qutip.entangling_power(qutip.swap())
0.0
>>> qutip.entangling_power(qutip.iswap())
0.22222222222222227
>>> qutip.entangling_power(qutip.sqrtiswap())
0.16666666666666674
>>> qutip.entangling_power(qutip.sqrtswap())
0.16666666666666674
```

hodgestar: The swap gate does not produce any entanglement since it swaps the two qubit completely. One needs to do a fractional swap to such as sqrtswap to get entanglement.

The iswap gate has maximal entangling power as a result of the multiplication by i of one basis element from each qubit (i.e. the "i" part). The swap part doesn't contribute to the entangling power.

hodgestar: The documentation is wrong but this is not the first QuTiP function for which this is the case. ;)

hodgestar:
Is there anything preventing this merging once the documentation builds are passing?

The documentation build failure is quite odd -- maybe something to do with merging in the upstream/dev.major branch?

hodgestar:
I think @BoxiLi should give final approval, but I am happy for this branch to be merged.

BoxiLi (approved):
Yeah, let's merge it.

PR #1892, opened 2022-05-11
Fix add_gate behavior when inserting a gate at multiple indices

### LINKED ISSUE #1497: Checking the behavior of the parameter `index` in the function `QubitCircuit.add_gate()`
I would like to check the behavior of the parameter `index` in the function `QubitCircuit.add_gate()`

In `qutip.qip.QubitCircuit.add_gate()`, there is a optional `list` type attribute called index.
This seems to be used to insert gates in the middle of the gate list (`qutip.qip.QubitCircuit.gates`).
The corresponding code is shown below.
```
            for position in index:
                num_mes = (sum(isinstance(op, Measurement) for op
                               in self.gates[:position]))
                gate = Gate(name, targets=targets, controls=controls,
                            arg_value=arg_value, arg_label=arg_label,
                            classical_controls=classical_controls,
                            control_value=control_value)
                self.gates.insert(position, gate)
```
In this code, if `index` contains two or more elements, then the insert positions would be shifted one by one.
Is this an expected behavior?

In the current code, the positions in `index` would not indicate the original positions during the insertion loop. (e.g. When we want to insert a gate `A` to `gates=[B0,B1,B2,B3,B4,B5]` in `index=[0,2,3]`, then the output gate list would be `gates=[A,B0,A,A,B1,B2,B3,B4,B5]`, not `gates=[A,B0,B1,A,B2,A,B3,B4,B5]`.)

In addition, the variable `num_mes` is unused throughout the code.

If this is not the expected behavior, and if the expected behavior is like the latter gate list in the former example, then we can fix this part into:

```
            gate = Gate(name, targets=targets, controls=controls,
                            arg_value=arg_value, arg_label=arg_label,
                            classical_controls=classical_controls,
                            control_value=control_value)
            for i, position in enumerate(sorted(index)):
                self.gates.insert(position + i, gate)
```

**Additional context**
This issue is also mentioned in the PR #1489 .

hristog:
**Description**
The fix pertains to `add_gate` behaviour when multiple insertion indices are supplied. Prior to the proposed fix, the insertion loop doesn't take into account the fact that inserting an element in the gates list, modifies the indices in the original list, whereas the caller of `add_gate` supplies a list of indices, with positions with respect to the indices in the original list.

Additionally, the PR removes unused variables and unused function (that has also not been imported).

**Related issues or PRs**
Fixes #1497.

**Changelog**
Fixed add_gate behavior when inserting a gate at multiple indices.

### DIFF
--- a/qutip/tests/test_qubitcircuit.py
+++ b/qutip/tests/test_qubitcircuit.py
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+from operator import attrgetter
 from pathlib import Path
 
 from qutip.qip.operations import gates
@@ -224,6 +225,29 @@ class TestQubitCircuit:
             # Single control
             pytest.raises(ValueError, qc.add_gate, gate, [0], [1])
 
+        dummy_gate = Gate("DUMMY")
+        inds = [1, 3, 4, 6]
+        qc.add_gate(dummy_gate, index=inds)
+
+        # Test adding gates at multiple indices at once.
+        # NOTE: Every insertion shifts the indices in the original list of
+        #       gates by an additional position to the right.
+        expected_gate_names = [
+            'CNOT',     # 0
+            'DUMMY',    # 1
+            'SWAP',     # 2
+            'TOFFOLI',  # 3
+            'DUMMY',    # 4
+            'SWAP',     # 5
+            'DUMMY',    # 6
+            'SNOT',     # 7
+            'RY',       # 8
+            'DUMMY',    # 9
+            'RY',       # 10
+        ]
+        actual_gate_names = list(map(attrgetter('name'), qc.gates))
+        assert actual_gate_names == expected_gate_names
+
     def test_add_circuit(self):
         """
         Addition of a circuit to a `QubitCircuit`
@@ -292,7 +316,7 @@ class TestQubitCircuit:
                     qc.gates[i].controls is not None):
                 assert (qc2.gates[i].controls[0]
                         == qc.gates[i].controls[0]+2)
-        
+
         # Test exception when the operators to be added are not gates or measurements
         qc.gates[-1] = 0
         pytest.raises(TypeError, qc2.add_circuit, qc)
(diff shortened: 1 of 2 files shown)


qutip/tests/test_qubitcircuit.py
```diff
@@ -224,6 +225,52 @@ def test_add_gate(self):
...
+            'TOFFOLI',  # 3
+            'DUMMY1',   # 4
+            'SWAP',     # 5
+            'DUMMY1',   # 6
+            'SNOT',     # 7
+            'RY',       # 8
+            'DUMMY1',   # 9
+            'RY',       # 10
+        ]
+        actual_gate_names = list(map(attrgetter('name'), qc.gates))
```
hodgestar, line 248:
```suggestion
        actual_gate_names = [gate.name for gate in qc.gates]
```
List comprehension looks a lot clearer.

hodgestar (commented):
I made a small suggestion, but looks good to me. Thanks!

@BoxiLi Should this go into 4.7.1 or only 5.0?

qutip/tests/test_qubitcircuit.py
hristog: Thanks for reviewing this! Indeed, a list comprehension is clearer in this case. I've updated the PR accordingly.

hodgestar: Pleasure! Thank you for making the PR.

hodgestar (approved):
This looks good to me. I've started the GitHub actions run.

@BoxiLi I'll leave final review to you & the decision about whether to label this 4.7.1 or 5.0.

Ericgig:
`qip/circuit.py` will be removed in 5.0, so a similar PR should go into qutip-qip.

hristog:
> `qip/circuit.py` will be removed in 5.0, so a similar PR should go into qutip-qip.

I'll raise one there as well, thanks!

BoxiLi (approved):
LGTM @hristog thanks for the PR!

PR #1895, opened 2022-05-18
Add fix for cython 0.29.30 for dev.major

Ericgig:
**Description**
Updating cython to 0.29.30 break some tests... This fix the issue.

**Changelog**
Add support for cython 0.29.30


hodgestar (approved):

hodgestar:
I've marked this for inclusion in QuTiP 4.7.1.

Should we test on cython 0.29.30 or leave our testing as is for now?

Ericgig:
v4 has not issue with the new version of cython, only v5.
So we should stay with the newest cython in tests.

PR #1897, opened 2022-05-24
Build documentation with latest pip version

Ericgig:
**Description**
A bug in [pip 22.1](https://github.com/pypa/pip/issues/11116) break the documentation build.  It's fixed in 22.1.1 but 22.1 is still the version packaged with the python we use for the build.

**Changelog**
Build documentation with latest pip version


hodgestar (approved):
Thanks!

PR #1907, opened 2022-05-27
Add a more flexible base result class for v5.

hodgestar:
**Description**
The more flexible base class adds:

- a list of state processors that store the state or e_ops values (but could also do other things)
- a ``_post_init`` method that allows a result sub-class to override what processors are configured
- helper methods ``_store_state``, ``_store_final_state``, ``_pre_copy`` and ``_e_op_func`` that sub-classes can override if they use states that are not ``Qobj``.
- a ``.e_ops`` attribute that stores a dictionary of ``ExpectOP`` instances where ``.e_ops[k](t, state)`` will calculate the expected value and ``.e_ops[k].op`` returns the original object used to create the e_op.
- a ``.e_data`` attribute that store the ``.expect`` values as a dictionary rather than a list, so that ``.e_data[k]`` is ``.expect[i]``.

Other changes:

- the normalization of states is now performed in the base solver class, rather than in results.
- the solver ``stats`` passed to the results are now *only for that solver .run call* rather than being a strange record of everything the solver has done so far.
These allow:

- The HEOM solver to implement a ``HEOMResult`` class and use the base solver ``.run`` method.
- The normalization of states to be applied when using the ``.step()`` solver method.
- Consistently expose ``.expect`` and a list of results (consistent with QuTiP 4 and useful when turning expectation values into a numpy array)
- Provide a dictionary of expectation values via ``.e_data`` so that when e_ops are supplied as a dictionary, the results can be accessed by the same keys, e.g. ``.e_data["nice-name"]``.
- Possibly allow the MC result classes to be refactored more easily.

**Related issues or PRs**
- #1889 -- I think we could implement this PR more easily on top of the refactored result class.
- #1869 -- the v5 HEOM Solver can use much more of the base solver and results class with these changes 

**Todo**
- [X] I would like to add more tests for the base result class, but it would be good to get a +1 on the general approach before I spend time on that.

**Changelog**
Add a more flexible base result class for v5.


hodgestar:
@Ericgig I re-factor the result class a fair bit to make it possible to re-use more of it for the HEOMSolver. Let me know if you like the general approach.

I would like to remove the additional parameters to `_post_init(...)` and supply everything either directly via the result options (since that would avoid the HEOMResult class receiving parameters it doesn't use like ``rhs_is_super``). I'm happy to implement that here if you have an idea, but I'm also happy to leave it for a future PR (maybe post #1889).

hodgestar:
I updated #1869 to use the new BaseResult class.

Ericgig:
You have my approval for most of it.

The approach of using `preprocessors` and `processors` fells overkill for use that I have, but I don't see any reason it wouldn't work. If you need it for HEOM, let's go for it.

I don't like having expect being a dict. I was actually thinking of going with supporting the 3 cases, at least for the output. If we keep `expect` as a property, the format change can be done only there and we can keep it in another format, dict or list, inside.

Simplifying `_post_init` will be tricky, normalization does not depend only on the option, but also the state... We could have `Solver.run` call `add_preprocessor` to add the normalization function, but them it would make it harder for HEOM to call.

Ericgig:
Then out about removing dict expect? Dictionary support was not tested and supported only for some solver for a long time. It would be strange to have it become the default.

There are a some differences for the user. Code using `np.array(result.expect)` or `for expect in result.expect:` will break. 

For MC result, it has little impact, dict are a little more annoying to work with compared to list, but it need the expect as ndarray
not list or dict anyway. 

As for plotting function, I forgot they existed.... We could have them as methods of `Result`.

hodgestar:
> Then out about removing dict expect? Dictionary support was not tested and supported only for some solver for a long time. It 
> would be strange to have it become the default.

Well, this is v5 -- we are trying to make things nicer.

The issue with lists of e_ops is that often the e_ops have no sensible ordering. E.g. if the e_ops are, say, the number operator and projectors on to two different bases of interest then `[qutip.num(N), proj1, proj2]` are later accessed as `.expect[0]`, `.expect[1]` and `.expect[2]` which are then unreadable and fairly arbitrary and unmemorable numbers. Having `.expect["num"]`, `.expect["proj1"]` and `.expect["proj2"]` is much more readable and will result in fewer mistakes.

I *really* like the dict interface because it gives this readability to results.

I'm happy to accept that others may prefer lists, but providing multiple interfaces creates a problem for any code working with results.

So it is a bit of tricky question to resolve.

> For MC result, it has little impact, dict are a little more annoying to work with compared to list, but it need the expect as ndarray
not list or dict anyway.

The MC result classes already have to work around these things in awkward ways, and probably need to be a lot more careful -- e.g. e_ops functions can return strange things like Qobjs or perhaps even stranger things like strings or dicts which can't be averaged over.

> There are a some differences for the user. Code using `np.array(result.expect)` or `for expect in result.expect:` will break.

Agreed, although unless `result.expect` contains a uniform set of things, this are probably not sensible things to be doing.

I have some other ideas but I think they make more of a mess -- e.g. keep `.expect` a list but provide `.expect_d` which is a dict.

If one needs to write v4 compatible code one can always pass a list of e_ops and then work with `.expect` a bit carefully (i.e. as a dict).

Thoughts?

 I have added tests now, so it's just the `.expect` issue to resolve for this PR, I think.

Ericgig:
Removing list and forcing dict does not feels making things nicer... Being forced to use dict is mostly a nuisance for me. Presently we support both list ans dict output without much issues, there are very few place that would have issue supporting multiple interfaces. (I can only think of `plot_expectation`.)

But how about using `namedtuple`?

It act as a list for loop and numpy, but elements can be called by both the number and name, and I like having result's output somewhat immutable. The biggest issue is that we must provide field names for list inputs...

For MC, if e_ops callable return weird things, they will have to choose the option to store all trajectories instead of averaging the result. If they ask for the average or standard derivation of a dict, they will and should get an error, the best we should do is making this error message useful. Custom reduce function could be doable, but probably be a mess, especially if we need one for both `mean` and `std`... So unless someone present a good use case for it, I don't plan going that way.

Ericgig:
Half related to this, what do you think about having data-layer's data _mostly_ immutable? We could ignore the tracking of copies of the states, which are sometime done by the integrator or solver (normalization in  `Solver._restore_state` ?)

There are some gains to work in-place: stochastic evolution is 30% faster inplace. But I see ways around it.

Ericgig (commented):
It's getting so general that it strange... `typically a :obj:~Qobj`... 

If we remove the normalization step, are `preprocessors` needed?
We could them also remove `_post_init`...
Normalization could be done in `_restore_state` and if data layer are immutable, there would be no extra copies.

It is possible to have `processor` that are neither stored state nor expectation values. I think it could be useful to make them accessible from solvers.

v4's result has `num_expect` and `num_collapse` attribute, while I am not sad of seeing them go, I feel we should also keep `num_expect` in `stats`.

qutip/tests/solver/test_results.py
```diff
@@ -1,89 +1,224 @@
...
+        assert repr(res) == "\n".join([
+            "<BaseResult",
+            "  Solver: test-solver",
+            "  Solver stats:",
+            "    stat-a: 1",
+            "    stat-b: 2",
+            "  Times:",
+            "    from 0",
+            "    to 4",
+            "    in 5 steps",
```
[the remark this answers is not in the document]
hodgestar: I didn't like this format either -- it's just how it was when I found it. I changed it to `Time interval: [0, 4] (5 steps)`.

qutip/solver/result.py
```diff
@@ -1,188 +1,328 @@
...
+    """
+    Pickable e_ops callable that calculates the expectation value for a given
+    operator.
+
+    Parameters
+    ----------
+    op : :obj:`~Qobj`
+        The expectation value operator.
+    """
+    def __init__(self, op):
```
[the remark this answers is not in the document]
hodgestar: At least in English, "op" is a much more common abbreviation for "operator" than "oper". E.g. "binop", "opcode", "op" (as in, medical operation).

hodgestar:
`namedtuple` is not a bad idea, although it's a bit odd because `e_ops` is never a namedtuple and it doesn't solve the problem of what type `.e_ops` has. And `namedtuple` forbids attributes which aren't identifiers, which is also problematic.

I guess we could write our own type, `Expect`, that acts enough like a list for numpy, etc but still allows item and attribute access. That feels a bit crazy, but maybe it is worth it?

I still feel that all of these options are overly complicated. `e_ops` is not a list conceptually. Lists are typically ordered, but unstructured. The `e_ops` are clearly not ordered because `e_ops[0]` doesn't "come before" `e_ops[1]` in any meaningful way. The order of the indexes in a list `e_ops` are meaningless -- they exist only because that is how Python lists work.

*Sometimes* there may be a natural ordering to e_ops -- e.g. if they happen to be the projectors onto a complete set of basis elements -- but this a small subset of the more general situation where e_ops might be any set of operators and functions.

I do see that you feel lists are the most important case though, so it would be good to support them well.

hodgestar:
@Ericgig Ready for review again.

I haven't implemented anything for detecting whether an .e_op returns a number that can be averaged, and I'd like to leave that for a later PR because it's really about a better implementation of the MultiTrajResult and MultiTrajResultAveraged classes and that feels quite separate (even if it tweaks the Result class a bit more).

In the `HEOMResult` class in #1869 I override `_store_state`, `_store_final_state`, `_pre_copy` and `_e_op_func` which gives quite a lot of flexibility and might be sufficient to implement `MultiTrajResult` as a sub-class of `Result`.

There is also a question of whether there should be a way to supply processors directly -- either in e_ops or a separate lists. Processors are almost identical to e_ops except their values are not stored in `.expect` or `.e_data` and they don't appear in `.e_ops`. I'd like to leave that for a later (maybe never) PR too -- currently I don't have an immediate use case for it, and I don't want to expose the extra feature directly until we do.

Ericgig (changes requested):
Looking good.
I would like one more test for `store_final_state`.

It would also be good to have one for normalization since it doesn't work for density matrices right now, but did not trigger anything.

hodgestar:
Ah, the test does not fail because I forgot to make psi0 a density matrix before passing it to sesolve. Fixing.

Ericgig:
May I propose another approach to testing normalization: use a non-hermitian Hamiltonian and `qeye` as the e_ops.
Them it is clear if the normalization step does it's work or not. Otherwise the `tol` should be very strick, the error between before and after the normalization is usually quite small.

`sesolve` does not differentiate between an `unitary` and a `dm`.

hodgestar:
@Ericgig Tests changed to use non-hermitian H.

qutip/tests/solver/test_sesolve.py
```diff
@@ -79,6 +80,29 @@ def test_sesolve(self, H, analytical, unitary_op):
...
+        # non-hermitean H causes state to evolve non-unitarily
+        H = qutip.Qobj([[1, -0.1j], [-0.1j, 1]])
+        psi0 = qutip.basis(2, 0)
+        options = SolverOptions(normalize_output=True, progress_bar=None)
+
+        if state_type == "ket":
+            output = sesolve(H, psi0, self.tlist, e_ops=[], options=options)
+            norms = [state.norm() for state in output.states]
+            np.testing.assert_allclose(
+                norms, [1.0 for _ in self.tlist], atol=5e-3,
```
Ericgig, line 97:
```suggestion
                norms, [1.0 for _ in self.tlist], atol=1e-15,
```
This should be very precise.

qutip/tests/solver/test_mesolve.py
```diff
@@ -194,14 +194,41 @@ def testME_TDDecayliouvillian(self, c_ops):
...
 
+    @pytest.mark.parametrize(['state_type'], [
+        pytest.param("ket", id="ket"),
+        pytest.param("dm", id="dm"),
+        pytest.param("unitary", id="unitary"),
+    ])
+    def test_mesolve_normalization(self, state_type):
+        # non-hermitean H causes state to evolve non-unitarily
+        H = qutip.Qobj([[1, -0.1j], [-0.1j, 1]])
+        H = qutip.liouvillian(H)  # ensure use of McSolve
```
Ericgig, line 205:
```suggestion
        H = qutip.Qobj([[1, -0.1j], [-0.1j, 1]])
        H = qutip.sprepost(H, H) # ensure use of MeSolve
```
The liouvillian is always trace preserving even when the Hamiltonian is not hermitian because we use:
`-1j * (spre(H) - spost(H))` without calling `H.dag()`...

qutip/tests/solver/test_mesolve.py
```diff
@@ -194,14 +194,41 @@ def testME_TDDecayliouvillian(self, c_ops):
...
+        psi0 = qutip.basis(2, 0)
+        options = SolverOptions(normalize_output=True, progress_bar=None)
+
+        if state_type in {"ket", "dm"}:
+            if state_type == "dm":
+                psi0 = qutip.ket2dm(psi0)
+            output = mesolve(H, psi0, self.tlist, e_ops=[], options=options)
+            norms = [state.norm() for state in output.states]
+            np.testing.assert_allclose(
+                norms, [1.0 for _ in self.tlist], atol=1e-6,
```
Ericgig, line 215:
```suggestion
                norms, [1.0 for _ in self.tlist], atol=1e-15,
```

Ericgig (approved):
Looks good.
I left some suggestions to make tests stricter.

hodgestar:
@Ericgig Thanks! Suggestions committed. I will merge if tests pass.

PR #1908, opened 2022-05-29
HEOMSolver return states and progress bar

christian512:
**Description**
While working on the [heom-tls notebook] I discovered two wrong behaviors of the `HEOMSolver`. 
1. Running `hsolver.run()` does not return the states if `e_ops=None`, but the documentation says that it should do so. I added the missing if condition to `HEOMSolver`. 
2. If passing a progress bar like `qutip.ui.progressbar.TextProgressBar()` to `HEOMSolver` it results in an error, because internally it only checks for `None` or `True`. The documentation says it should also work if I pass an instance of `BaseProgressBar` (or a subclass). I added the checks and a test in the testfile for this.

**Related issues or PRs**
None

**Changelog**
Fix HEOMSolver state return and progress bar

### DIFF
--- a/qutip/nonmarkov/bofin_solvers.py
+++ b/qutip/nonmarkov/bofin_solvers.py
@@ -938,7 +938,7 @@ class HEOMSolver:
                 solver.y[:n ** 2].reshape(rho_shape, order='F'),
                 dims=rho_dims,
             )
-            if self.options.store_states:
+            if self.options.store_states or not e_ops:
                 output.states.append(rho)
             if ado_return or e_ops_callables:
                 ado_state = HierarchyADOsState(
--- a/qutip/nonmarkov/bofin_solvers.py
+++ b/qutip/nonmarkov/bofin_solvers.py
@@ -455,12 +455,13 @@ class HEOMSolver:
             for k in range(self._n_exponents)
         ]
 
-        self.progress_bar = progress_bar
         if progress_bar is None:
             self.progress_bar = BaseProgressBar()
-        if progress_bar is True:
+        elif progress_bar is True:
             self.progress_bar = TextProgressBar()
-        if not isinstance(self.progress_bar, BaseProgressBar):
+        elif isinstance(progress_bar, BaseProgressBar):
+            self.progress_bar = progress_bar
+        else:
             raise TypeError(
                 "ProgressBar (progress_bar) is not and instance of"
                 "qutip.ui.BaseProgressBar"
--- a/qutip/nonmarkov/bofin_solvers.py
+++ b/qutip/nonmarkov/bofin_solvers.py
@@ -455,10 +455,16 @@ class HEOMSolver:
             for k in range(self._n_exponents)
         ]
 
+        self.progress_bar = progress_bar
         if progress_bar is None:
             self.progress_bar = BaseProgressBar()
         if progress_bar is True:
             self.progress_bar = TextProgressBar()
+        if not isinstance(self.progress_bar, BaseProgressBar):
+            raise TypeError(
+                "ProgressBar (progress_bar) is not and instance of"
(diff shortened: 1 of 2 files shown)


qutip/nonmarkov/bofin_solvers.py
```diff
@@ -455,10 +455,16 @@ def __init__(
             for k in range(self._n_exponents)
         ]
 
+        self.progress_bar = progress_bar
         if progress_bar is None:
             self.progress_bar = BaseProgressBar()
         if progress_bar is True:
             self.progress_bar = TextProgressBar()
+        if not isinstance(self.progress_bar, BaseProgressBar):
```
hodgestar, line 463: Thanks for sorting this out. I would recommend we change it a little more to be:
```
if progress_bar is None:
    ...
elif progress_bar is True:
    ...
elif isinstance(self.progress_bar, BaseProgressBar):
    ...
else:
    raise TypeError(...)
```
just to make it easy to see that the logic is correct.

hodgestar (changes requested):
Thank you for fixing these and for adding the test. I requested one small further improvement to the progress bar setting, but otherwise it looks good.

qutip/nonmarkov/bofin_solvers.py
christian512: Thanks for the feedback :) I changed the if-statement to this cleaner version.

qutip/tests/test_bofin_solvers.py
```diff
@@ -528,6 +528,9 @@ def test_create_progress_bar(self):
         hsolver = HEOMSolver(H, bath, 2, progress_bar=True)
         assert isinstance(hsolver.progress_bar, TextProgressBar)
 
+        hsolver = HEOMSolver(H, bath, 2, progress_bar=TextProgressBar())
+        assert isinstance(hsolver.progress_bar, BaseProgressBar)
+
```
hodgestar, line 533: Could we also add a test for case where TypeError is raised?

qutip/nonmarkov/bofin_solvers.py
```diff
@@ -457,8 +457,15 @@ def __init__(
...
         if progress_bar is None:
             self.progress_bar = BaseProgressBar()
-        if progress_bar is True:
+        elif progress_bar is True:
             self.progress_bar = TextProgressBar()
+        elif isinstance(progress_bar, BaseProgressBar):
+            self.progress_bar = progress_bar
+        else:
+            raise TypeError(
+                "ProgressBar (progress_bar) is not and instance of"
```
AGaliciaMartinez, line 466:
```suggestion
                "ProgressBar (progress_bar) is not and instance of "
```

qutip/nonmarkov/bofin_solvers.py
```diff
@@ -457,8 +457,15 @@ def __init__(
...
         if progress_bar is None:
             self.progress_bar = BaseProgressBar()
-        if progress_bar is True:
+        elif progress_bar is True:
             self.progress_bar = TextProgressBar()
+        elif isinstance(progress_bar, BaseProgressBar):
+            self.progress_bar = progress_bar
+        else:
+            raise TypeError(
+                "ProgressBar (progress_bar) is not and instance of "
```
hodgestar, line 466:
```suggestion
                "ProgressBar (progress_bar) is not an instance of "
```

qutip/nonmarkov/bofin_solvers.py
```diff
@@ -457,8 +457,15 @@ def __init__(
...
+                "qutip.ui.BaseProgressBar"
```
hodgestar, line 467:
```suggestion
                "progress_bar is not an instance of "
                "qutip.ui.BaseProgressBar"
```

hodgestar (approved):

PR #1909, opened 2022-05-30
Bloch cleanup dev

AGaliciaMartinez:
**Description**
Clean-up of the point method in the `Bloch` class. This PR does not add any new feature but it does:
- Improve readability and logic in the point methods for the Bloch class. 
- Extended test coverage of the point methods.
- Now add_points raises error if the arguments `points` or `meth` are wrong.

**Changelog**
Clean-up of the point methods in the `Bloch` class.

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -784,7 +784,6 @@ class Bloch:
                 self.axes.add_artist(a)
 
     def plot_points(self):
-
         # -X and Y data are switched for plotting purposes
         for k, points in enumerate(self.points):
             points = np.asarray(points)
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -358,12 +358,10 @@ class Bloch:
         if meth == 's' and points.shape[1] == 1:
             points = np.append(points[:, :1], points, axis=1)
 
-
         self.point_style.append(meth)
         self.points.append(points)
         self.point_alpha.append(alpha)
         self.point_color.append(colors)
-
 
     def add_states(self, state, kind='vector', colors=None, alpha=1.0):
         """Add a state vector Qobj to Bloch sphere.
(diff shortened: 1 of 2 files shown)


qutip/bloch.py
```diff
@@ -798,53 +784,54 @@ def plot_vectors(self):
...
-                dist, indperm = zip(*zipped)
-                indperm = array(indperm)
+        for k, points in enumerate(self.points):
+            points = np.asarray(points)
+            num_points = points.shape[1]
+
+            dist = np.linalg.norm(points, axis=0)
+            if not np.allclose(dist, dist[0], rtol=1e-12):
+                indperm = np.argsort(dist)
+                points = points[:, indperm]
```
AGaliciaMartinez, line 796: This part of the code was also there before. I seems to sort points based on their distance to the bloch centre, but I am not sure why you would want to do so.... It is neither documented nor tested.

AGaliciaMartinez:
I was surprised to see that the color argument was added long time ago for the point methods. These were not ported to master. These changes break backwards compatibility as code that used `point_colors` will not work anymore. I wonder if this change of variable name is something we want to keep as it should be possible to keep the behavior backwards compatible AND include the new colors argument (which I think is a very necessary feature that motivated me to clean the code in master in the first place, I later realized that this was already added in dev.major).

Ericgig:
If my memory is good, `color` was added in a PR for GSoC application. It fixed the related issue, but still needed some polishing...  If you can fix any issue with, it would be appreciated.

qutip/bloch.py
hodgestar: I suspect this is related to the behaviour of older versions of matplotlib. Before matplotlib 3.5 the rendering order of Axes3D was determined by the zordering of the Artist objects, which is a concept that only made sense for 2D rendering. In 3.5, Axes3D gained a "computed_zorder" option that is now the default and which should order everything correctly. See https://matplotlib.org/stable/api/_as_gen/mpl_toolkits.mplot3d.axes3d.Axes3D.html.

I would vote that we remove this reordering and live with possibly suboptimal rendering in old versions of matplotlib.

qutip/bloch.py
```diff
@@ -799,52 +785,52 @@ def plot_vectors(self):
...
-                                  color=self.point_color[k][0], marker=marker)
+                indperm = np.arange(num_points)
+
+            s = self.point_size[mod(k, len(self.point_size))]
+            marker = self.point_marker[mod(k, len(self.point_marker))]
+            user_color = self.point_color[k]
+            style = self.point_style[k]
+            if user_color is not None:
+                color = user_color
+            elif user_color is None and self.point_style[k] in ['s', 'l']:
```
hodgestar, line 805:
```suggestion
            elif self.point_style[k] in ['s', 'l']:
```
`user_color` is always `None` in the `elif`.

hodgestar (changes requested):
Thanks @AGaliciaMartinez! I left a couple of small suggestions but it mostly looks good to me. I suggest that we leave sorting out the point_color mismatch with master and the removal of the distance ordering for separate PRs (assuming we decide to fix those things).

AGaliciaMartinez:
I agree on leaving this PR as it is. @hodgestar, I committed your suggestions and removed the `user_color` temporary variable (was not really necessary).

hodgestar (approved):
Approved. Happy for it to be merged once tests pass.

PR #1911, opened 2022-05-31
Removed unused imports.

AGaliciaMartinez:
I removed some of the imports from numpy. I favored the writing style of `np.linspace` vs `linspace` in most cases except for `outer`, `ones`, `cos` and `sin`. I kept these for readability in lines 706-708 but could be changed to `np.outer` and so on.

**Changelog**
Clean-up imports Bloch.

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -2,9 +2,8 @@ __all__ = ['Bloch']
 
 import os
 
-from numpy import (ndarray, array, linspace, pi, outer, cos, sin, ones, size,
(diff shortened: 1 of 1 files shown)


AGaliciaMartinez:
@hodgestar @Ericgig, feel free to reject this PR if you think this change is not needed.

hodgestar (approved):

Ericgig (approved):
I would also change `size` to `np.size` since we also use `font_size`, `point_size`, etc.

Issue #1916, opened 2022-06-02
Bloch arrow_style bug


AGaliciaMartinez:
### Bug Description

The option `vector_style=''` does not work. 

Even though the fix is quite simple, I would like to remove it since it is not documented, it would simplify the code and the option '-' achieves the same thing anyways.

@hodgestar and @Ericgig, any thoughts?

### Code to Reproduce the Bug

```shell
import qutip
bloch = qutip.Bloch()
bloch.add_vectors([0,0,1])
bloch.vector_style = ''
bloch.show()
```

### Code Output

```shell
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [4], in <cell line: 4>()
      2 bloch.add_vectors([0,0,1])
      3 bloch.vector_style = ''
----> 4 bloch.show()

File ~/git_repo/qutip/qutip5/qutip/bloch.py:883, in Bloch.show(self)
    868 def show(self):
    869     """
    870     Display Bloch sphere and corresponding data sets.
    871 
   (...)
    881     in the cell you create the figure in.
    882     """
--> 883     self.render()
    884     if self.run_from_ipython():
    885         display(self.fig)

File ~/git_repo/qutip/qutip5/qutip/bloch.py:677, in Bloch.render(self)
    675 self.plot_back()
    676 self.plot_points()
--> 677 self.plot_vectors()
    678 self.plot_lines()
    679 self.plot_arcs()

File ~/git_repo/qutip/qutip5/qutip/bloch.py:779, in Bloch.plot_vectors(self)
    775     color = self.vector_default_color[idx]
    777 if self.vector_style == '':
    778     # simple line style
--> 779     self.axes.plot(xs3d, ys3d, zs3d,
    780                    zs=0, zdir='z', label='Z',
    781                    lw=self.vector_width, color=color,
    782                    alpha=alpha)
    783 else:
    784     print(
    785         xs3d, ys3d, zs3d,
    786                 self.vector_mutation,
   (...)
    789                 color,
    790                 alpha)

File ~/.virtualenvs/qutip5/lib/python3.10/site-packages/mpl_toolkits/mplot3d/axes3d.py:1490, in Axes3D.plot(self, xs, ys, zdir, *args, **kwargs)
   1488     zs, *args = args
   1489     if 'zs' in kwargs:
-> 1490         raise TypeError("plot() for multiple values for argument 'z'")
   1491 else:
   1492     zs = kwargs.pop('zs', 0)

[...]

Issue #1913, opened 2022-05-31
Bloch various updates


AGaliciaMartinez:
### Describe the Issue!

As briefly described in #1910 there are some functions in `Bloch` class that are not fully tested. Besides, there are a few other changes that may benefit `Bloch`. I would like to list them here and provide more information on the change for discussion. 

The main changes I propose to do in the Bloch class are: 
- [x] clean-up of the code in `add_points` and `plot_points` methods (addressed in #1909).
- [x]  Clean import statements.
- [x] clean-up of the code in `add_vectors` and `plot_vectors` methods (#1917).
- [ ] clean-up of the code in `add_state` method. 
- [ ] make the plot functions (`plot_points, plot_vectors, ...`) private as they are meant to be used only in the render method (the do not even have a docstring). This _breaks backwards compatibility_ but I imagine it will be ok as I do not expect anyone using these methods.
- [x] Test `_repr_svg_` and `_repr_png_` (there seems to be a bug in `_repr_svg_` #1912).
- [ ] We have `render` and `make_sphere` methods that do _exactly_ the same (`make_sphere` calls `render`) I would vote to deprecate `make_sphere` for QuTiP 5.0.
- [ ] Make the interface consistent among the methods `add_state`, `add_vector`, `add_points`, `add_arc` and `add_line`. (This would address #1105 )
- [ ] Address #1106.
- [ ] Update documentation AND notebooks with changes.

These task are planned to be addressed each in a separate PR. I plan to finish these tasks myself but feel free to tackle any of these (or add a new one) if no PR is open and you mention it in this issue.

hodgestar:
If you would like another item for the todo list, I'd really like to see the matplotlib aspects of the interface cleaned up. When we added the new lines and arcs features, we allowed passing arbitrary matplotlib plotting arguments as additional parameters. Perhaps we can do the same for the other functions and remove the old means of setting colours, etc?

AGaliciaMartinez:
Yes, definitely! I would like to tackle that when with the item list "Make the interface consistent ..."

Ericgig:
- [ ] Add control for the wireframe's linewidth #2396

PR #1917, opened 2022-06-02
Bloch vector methods

AGaliciaMartinez:
**Description**
Clean-up of the methods `add_vectors` and `plot_vectors`.
- raise ValueError when wrong values on inputs for vectors and colors.
- fixes issue #1916 
- colors argument was bugged as it was not being taken into account when plotting. I kept the colors argument behaving as it was first intended, although we may want to change it later to match the behavior of points.
- Added tests for new ValueErrors and for the colors argument.

**Related issues or PRs**
#1913.

**Changelog**
Clean-up of the methods `add_vectors` and `plot_vectors`.
Now add_vectors raises ValueErrors for wrong vectors and colors argument
Fix issue #1916 
Fix colors argument
Improve test coverage

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -434,7 +434,7 @@ class Bloch:
         else:
             colors = np.asarray(colors)
 
-        if colors.ndim != 1 and colors.size != n_vectors:
+        if colors.ndim != 1 or colors.size != n_vectors:
             raise ValueError("The included colors are not valid. colors must "
                              "be equivalent to a 1D array with the same "
                              "size as the number of vectors. ")
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -45,7 +45,6 @@ try:
             xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, self.axes.M)
             self.set_positions((xs[0], ys[0]), (xs[1], ys[1]))
             return np.min(zs)
-
 except ImportError:
     pass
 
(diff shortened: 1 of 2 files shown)


qutip/bloch.py
```diff
@@ -415,29 +415,33 @@ def add_vectors(self, vectors, colors=None, alpha=1.0):
...
+        if vectors.ndim != 2 or vectors.shape[1] != 3:
+            raise ValueError(
+                "The included vectors are not valid. Vectors must "
+                "be equivalent to a 2D array where the first "
+                "index represents the iteration over the vectors and the "
+                "second index represents the position in 3D of vector head.")
+
+        n_vectors = vectors.shape[0]
+        if colors is None:
+            colors = np.array([None]*n_vectors)
```
hodgestar, line 432:
```suggestion
            colors = np.array([None] * n_vectors)
```

hodgestar: Maybe we should allow colors to be a scalar and then expand that to all vectors?

hodgestar (changes requested):
Thanks -- this looks much nicer!

I've always found the wrapping around of the colours a bit odd, but I guess it's useful for plotting small number of vectors and not having to make up the colours oneself. Certainly using the auto-colouring and explicit colouring together gives rather bizarre results, but I think that's fine. Either the person writing the code cares enough to specify their own colours, or not.

qutip/bloch.py
AGaliciaMartinez: Yes definitely. If you do not mind though, I'd rather do that in another PR once I have a better idea of how I would like to implement the kw arguments for plotting. There are in fact some challenges with points using two different methods for plotting and some arguments being attributes in the bloch sphere. I will try to address all of that in a single PR. For the moment, this fixes a few things and it already adds some small features.

hodgestar: Sound good.

hodgestar (approved):

PR #1918, opened 2022-06-02
Fix#1912

### LINKED ISSUE #1912: Bug with Bloch and Ipython.
### Bug Description

`Bloch` raises an error when used in jupyter notebook. This seems to be due to the output of `print_figure` in `_repr_svg_` not being bytecode (maybe it was in the past?) it then defaults to `_repr_png_` and renders correctly the bloch sphere.

### Code to Reproduce the Bug

```shell
import qutip
qutip.Bloch()
```

### Code Output

```shell
~/.virtualenvs/qutip4/lib/python3.10/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/git_repo/qutip/qutip4/qutip/bloch.py in _repr_svg_(self)
    293         from IPython.core.pylabtools import print_figure
    294         self.render()
--> 295         fig_data = print_figure(self.fig, 'svg').decode('utf-8')
    296         plt.close(self.fig)
    297         return fig_data

AttributeError: 'str' object has no attribute 'decode'
```

### Expected Behaviour

The Bloch sphere should be plotted correctly without any Error.

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+ee51e50
Numpy Version:      1.22.3
Scipy Version:      1.8.1
Cython Version:     None
Matplotlib Version: 3.5.2
Python Version:     3.10.4
Number of CPUs:     8
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
```

### Additional Context

_No response_

AGaliciaMartinez:
**Description**
Fix #1912. The issue was that previously the print_figure function returned byte code, but since ipython/ipython#5452 it returns a Unicode string.

**Related issues or PRs**
Fix #1912

**Changelog**
Fixed an error with svg representation of Bloch.


qutip/tests/test_bloch.py
```diff
@@ -419,3 +419,8 @@ def test_vector(self, vector_kws, fig_test, fig_ref):
             vector_kws = [vector_kws]
         self.plot_vector_test(fig_test, copy.deepcopy(vector_kws))
         self.plot_vector_ref(fig_ref, copy.deepcopy(vector_kws))
+
+
+def test_repr_svg():
+    svg = Bloch()._repr_svg_()
+    assert isinstance(svg, str)
```
hodgestar, line 426: We could probably make this test a little stronger -- e.g. `assert svg.startswith(...)` and maybe `.endswith(...)`.

AGaliciaMartinez:
I added ipython to the testing workflow. Let me know if it should have been included somewhere else.

qutip/tests/test_bloch.py
AGaliciaMartinez: Great idea! I added it into the test.

hodgestar:
> I added ipython to the testing workflow. Let me know if it should have been included somewhere else.

Hmm. I don't like adding random dependencies into the workflow config. On the other hand, this is specific to IPython. Maybe we can add an "ipython" section in `setup.cfg` and add that to `full`?

qutip/tests/test_bloch.py
```diff
@@ -434,7 +434,7 @@ def test_vector(self, vector_kws, fig_test, fig_ref):
             vector_kws = [vector_kws]
         self.plot_vector_test(fig_test, copy.deepcopy(vector_kws))
         self.plot_vector_ref(fig_ref, copy.deepcopy(vector_kws))
-
+
```
AGaliciaMartinez, line 437:
```suggestion

```

AGaliciaMartinez:
I think I still needed to change the test workflow. Lets see if this works but let me know if you prefer another approach to include ipython in testing.

AGaliciaMartinez:
Great! @hodgestar Let me know if any other change is required.

PR #1920, opened 2022-06-06
Setting qutip-qip as an optional plugin for qutip-5 [unitaryhack]

### LINKED ISSUE #1893: Setting `qutip-qip` as an optional package for qutip-5
### Problem Description

From version 5.0, the subpackage `qutip.qip` will be removed and replaced by a separate package `qutip-qip`. To make it easier for others to migrate from 4.7 to 5.0, we can set up `qutip-qip` as an optional plug-in for qutip, so that if `qutip-qip` is installed, the `qutip.qip` name space can be used as before.

### Expected behaviour
- If only `qutip` is installed, importing `qutip.qip` will raise an `ImportError` with instruction for installing `qutip_qip`.
- If `qutip-qip` is installed, using `qutip.qip` will automatically use the `qutip_qip` namespace, with no difference in the user interface.
- Update `setup.cfg` so that `pip install qutip[qip]` will automatically install `qutip_qip`.

E.g. the following should work if `qutip`-5.0 and `qutip-qip` are both installed
```python
# The following should be the same
from qutip_qip.circuit import QubitCircuit
```

### Proposed Solution

See https://github.com/qutip/qutip/discussions/1500 for ideals on the implementation.

**Note: The implementation should be done against the `dev.major` branch of qutip.**

HGSilveri:
**Description**

- Allow imports of `qutip.qip` seamlessly as imports of `qutip_qip`
- Test the import behaviour when `qutip_qip` is installed and when not.

**Related issues or PRs**
Incorporates the ideas in the discussion of #1500.
Fixes #1893.

**Todos and questions**
- [ ] ~~Have `qutip-qip` be an optional dependency~~ (will be left for a future PR)
- [ ] ~~Add installation of `qutip-qip` to the CI workflow~~

I have a draft of these changes locally, but they fail due to a dependency conflict: `qutip_qip` currently requires `qutip<5 and >=4.6 ` while these changes are being implemented in `v5.0.0.dev`, ie `>5`. This will cause the unit test where `qutip_qip` is installed to be skipped. Any ideas on how to get around this?

**Edit**: Uses a stub for the `qutip_qip` package in the units tests instead, which removes the need of having `qutip-qip` installed.

**Changelog**

Allow imports of `qutip.qip` seamlessly as imports of `qutip_qip`.
Have `qutip-qip` be an optional dependency.

### DIFF
--- a/qutip/qip.py
+++ b/qutip/qip.py
@@ -8,6 +8,6 @@ try:
     del qutip_qip
 except ImportError:
     raise ImportError(
-        "'qutip.qip' imports require the `qutip_qip` package. Install it with "
+        "'qutip.qip' imports require the 'qutip_qip' package. Install it with "
         "`pip install qutip-qip` or with `pip install qutip[qip]`."
     )
--- a/qutip/qip.py
+++ b/qutip/qip.py
@@ -9,5 +9,5 @@ except ImportError:
     raise ImportError(
         "Importing 'qutip.qip' requires the 'qutip_qip' package. Install it "
         "with `pip install qutip-qip` (for more details, go to "
-        "https://qutip-qip.rtfd.io/en/stable/)."
+        "https://qutip-qip.readthedocs.io/)."
     )
--- a/qutip/qip.py
+++ b/qutip/qip.py
@@ -1,0 +1,13 @@
+"""Module replicating the qutip_qip package from within qutip."""
+import sys
+
+try:
+    import qutip_qip
+
+    sys.modules["qutip.qip"] = sys.modules["qutip_qip"]
+    del qutip_qip
+except ImportError:
+    raise ImportError(
+        "'qutip.qip' imports require the `qutip_qip` package. Install it with "
+        "`pip install qutip-qip` or with `pip install qutip[qip]`."
+    )
--- a/qutip/qip.py
+++ b/qutip/qip.py
@@ -3,11 +3,11 @@ import sys
 
 try:
     import qutip_qip
-
-    sys.modules["qutip.qip"] = sys.modules["qutip_qip"]
     del qutip_qip
+    sys.modules["qutip.qip"] = sys.modules["qutip_qip"]
 except ImportError:
     raise ImportError(
-        "'qutip.qip' imports require the 'qutip_qip' package. Install it with "
-        "`pip install qutip-qip` or with `pip install qutip[qip]`."
+        "Importing 'qutip.qip' requires the 'qutip_qip' package. Install it "
+        "with `pip install qutip-qip` (for more details, go to "
+        "https://qutip-qip.rtfd.io/en/stable/)."
     )
--- a/qutip/tests/test_qip.py
+++ b/qutip/tests/test_qip.py
@@ -1,0 +1,39 @@
+import sys
+import unittest
+
+import pytest
+
+
+class TestWithoutQip(unittest.TestCase):
+    def setUp(self):
+        self._temp_qip = None
+        if sys.modules.get("qutip_qip"):
+            self._temp_qip = sys.modules["qutip_qip"]
+        sys.modules["qutip_qip"] = None
+
+    def tearDown(self):
+        if self._temp_qip:
+            sys.modules["qutip_qip"] = self._temp_qip
+        else:
+            del sys.modules["qutip_qip"]
+
+    def test_failed_import(self):
+        # Ensure 'qutip.qip' is not imported yet
+        assert "qutip.qip" not in sys.modules
+        with pytest.raises(
+            ImportError,
+            match="'qutip.qip' imports require the 'qutip_qip' package.",
+        ):
+            import qutip.qip
+
+
+def test_with_qip():
+    # Skips test if 'qutip_qip' is not installed
+    qutip_qip = pytest.importorskip("qutip_qip")
+    import qutip.qip
+    import qutip.qip.circuit as circuit
+    from qutip.qip.circuit import QubitCircuit
+
+    assert qutip.qip is qutip_qip
+    assert circuit is qutip_qip.circuit
+    assert QubitCircuit is qutip_qip.circuit.QubitCircuit
--- a/qutip/tests/test_qip.py
+++ b/qutip/tests/test_qip.py
@@ -1,38 +1,44 @@
 import sys
-import unittest
 
 import pytest
 
 
-class TestWithoutQip(unittest.TestCase):
-    def setUp(self):
-        self._temp_qip = None
-        if sys.modules.get("qutip_qip"):
-            self._temp_qip = sys.modules["qutip_qip"]
-        sys.modules["qutip_qip"] = None
+class QutipQipStub:
+    class CircuitModuleStub:
+        QubitCircuit = "FakeQubitCircuit"
 
-    def tearDown(self):
-        if self._temp_qip:
-            sys.modules["qutip_qip"] = self._temp_qip
-        else:
-            del sys.modules["qutip_qip"]
-
-    def test_failed_import(self):
-        # Ensure 'qutip.qip' is not imported yet
-        assert "qutip.qip" not in sys.modules
-        with pytest.raises(
-            ImportError,
-            match="'qutip.qip' imports require the 'qutip_qip' package.",
-        ):
-            import qutip.qip
+    circuit = CircuitModuleStub()
 
 
-def test_with_qip():
-    # Skips test if 'qutip_qip' is not installed
-    qutip_qip = pytest.importorskip("qutip_qip")
+@pytest.fixture
+def without_qutip_qip(monkeypatch):
+    monkeypatch.setitem(sys.modules, "qutip_qip", None)
+    monkeypatch.delitem(sys.modules, "qutip.qip", raising=False)
+
+
+@pytest.fixture
+def with_qutip_qip_stub(monkeypatch):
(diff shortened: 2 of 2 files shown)


BoxiLi:
Ah yes, that constraint was there to prevent accidentally using `qutip-qip` with `qutip@dev.major` because the support was only added a few weeks ago. Only the master branch of `qutip-qip` support this. So this basically prevents installing the `qutip@dev.major` and the released `qutip-qip` at the same time. One has to install `qutip-qip@master` to test them.

This also means that adding `qutip-qip` as an optional package right now is probably does not make much sense. Probably we can drop this.

To test it on GitHub action we need to install `qutip-qip@master` for now and test the importation. @hodgestar Is that ok? Otherwise, I can try to get out a new release of `qutip-qip` in a day or so. Just to update the change log and click the button.

HGSilveri:
> This also means that adding `qutip-qip` as an optional package right now does not make much sense. Probably we can drop this.
> 
> To test it on GitHub action we need to install `qutip-qip@master` for now and test the importation. @hodgestar Is that ok? Otherwise, I can try to get out a new release of `qutip-qip` in a day or so. Just to update the change log and click the button.

Completely up to you, of course! If you choose to drop the optional dependency for now, I can also just put those changes on another PR to be merged when the time is right.
I'll stand by for your decision regarding the CI tests.

qutip/qip.py
```diff
@@ -0,0 +1,13 @@
+"""Module replicating the qutip_qip package from within qutip."""
+import sys
+
+try:
+    import qutip_qip
+
+    sys.modules["qutip.qip"] = sys.modules["qutip_qip"]
+    del qutip_qip
```
hodgestar, line 8:
```suggestion
    import qutip_qip
    del qutip_qip
    sys.modules["qutip.qip"] = sys.modules["qutip_qip"]
```
You can actually `del qutip_qip` staight away if you're going to look it up in sys.modules anyway.

qutip/qip.py
```diff
@@ -0,0 +1,13 @@
...
+except ImportError:
+    raise ImportError(
+        "'qutip.qip' imports require the 'qutip_qip' package. Install it with "
+        "`pip install qutip-qip` or with `pip install qutip[qip]`."
```
hodgestar, line 12:
```suggestion
        "Importing 'qutip.qip' requires the 'qutip_qip' package. Install it with "
        "`pip install qutip-qip` or with `pip install qutip[qip]`."
```
The phrasing just felt a bit clearer to me this way.

qutip/tests/test_qip.py
```diff
@@ -0,0 +1,39 @@
...
+        self._temp_qip = None
+        if sys.modules.get("qutip_qip"):
+            self._temp_qip = sys.modules["qutip_qip"]
+        sys.modules["qutip_qip"] = None
+
+    def tearDown(self):
+        if self._temp_qip:
+            sys.modules["qutip_qip"] = self._temp_qip
+        else:
+            del sys.modules["qutip_qip"]
```
hodgestar, line 18: We should just delete these two methods and use fixtures instead.

qutip/tests/test_qip.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_failed_import(self):
+        # Ensure 'qutip.qip' is not imported yet
+        assert "qutip.qip" not in sys.modules
```
hodgestar, line 22: Instead of this, let's use pytest properly and write two fixtures, something like:
```python
@pytest.fixture
def without_qutip_qip(monkeypatch):
     monkeypatch.delitem(sys.modules, "qutip_qip", raising=False)
     monkeypatch.delitem(sys.modules, "qutip.qip", raising=False)

@pytest.fixture
def with_qutip_qip_stub(monkeypatch):
    monkeypatch.setitem(sys.modules, "qutip_qip", QutipQipStub())
     monkeypatch.delitem(sys.modules, "qutip.qip", raising=False)
```
where `QutipQipStub` is a class we write that is a very fake `qutip_qip` module (maybe with just one or two attributes).

qutip/tests/test_qip.py
```diff
@@ -0,0 +1,39 @@
...
+        with pytest.raises(
+            ImportError,
+            match="'qutip.qip' imports require the 'qutip_qip' package.",
+        ):
+            import qutip.qip
+
+
+def test_with_qip():
+    # Skips test if 'qutip_qip' is not installed
+    qutip_qip = pytest.importorskip("qutip_qip")
```
hodgestar, line 32: Once we have the `with_qutip_qip_stub` we can just always run this test.

hodgestar (changes requested):
I made some suggestions to make it possible to always run the tests.

hodgestar:
Regarding the suggestions in the TODO list:

> Have qutip-qip be an optional dependency

I'm wondering about this. Having `pip install qutip[qip]` is nice, but what should happen with `pip install qutip[full]`? Should that also install `qutip_qip`? If not should we then leave `qutip_qip` out of `full`? And should we have `qutip[full-family]`?

> Add installation of qutip-qip to the CI workflow

This we should not do -- we don't want anything in core qutip to accidentally require qutip_qip (that was sort of the point of the family packages).

qutip/tests/test_qip.py
BoxiLi: I see, in this way you fake a `qutip_qip` package but don't install it. That is clever.

BoxiLi:
> I'm wondering about this. Having pip install qutip[qip] is nice, but what should happen with pip install qutip[full]? Should that also install qutip_qip? If not should we then leave qutip_qip out of full? And should we have qutip[full-family]?

I understand `[full]` as "installing everything that might be needed so that I can run everything without any ImportError". So this means that it includes `[qip]`. However, I do get your concern that this will probably make the option `[full]` larger and larger in the future.

Nonetheless, since there is not yet a released `qutip-qip` that supports qutip-5. Let's leave it to a different PR.

qutip/qip.py
HGSilveri: True, will do

qutip/qip.py
BoxiLi: Yes, let's drop that in this PR. Could you also add a reference to https://qutip-qip.readthedocs.io/en/stable/index.html?

HGSilveri: Sure! It's done

HGSilveri:
> This we should not do -- we don't want anything in core qutip to accidentally require qutip_qip (that was sort of the point of the family packages).

Sure, no need for that anymore, I'll edit the description.

hodgestar:
> I understand `[full]` as "installing everything that might be needed so that I can run everything without any ImportError". So this means that it includes `[qip]`. However, I do get your concern that this will probably make the option `[full]` larger and larger in the future.

Perhaps installing `qutip_qip` as part of `full` is a good first step. It means that QuTiP 5 will just "work" as it did before and we can always remove it from `full` in QuTiP 5.1 or later.

qutip/tests/test_qip.py
HGSilveri: Done. 
By the way, to raise the ImportError one has to set `sys.modules["qutip_qip"]=None" - deleting it causes the module to be reimported instead (which will work fine on the CI test environment because `qutip_qip` is indeed not installed, but failed locally on my environment).

qutip/tests/test_qip.py
```diff
@@ -0,0 +1,45 @@
...
+        import qutip.qip
+
+
+def test_with_qip(monkeypatch, with_qutip_qip_stub):
+    import qutip.qip
+
+    monkeypatch.setitem(sys.modules, "qutip.qip.circuit", qutip.qip.circuit)
+    import qutip.qip.circuit as circuit
+    from qutip.qip.circuit import QubitCircuit
+    import qutip_qip
```
hodgestar, line 41: I agree that it would be nice to test that sub-modules are also importable (i.e. does our approach handle that case) but I don't think monkeypatching sys.modules in the test shows that has has worked.

I guess a much better test is to instead of creating a dummy `qutip_qip` here and sticking it in sys.modules, is to, in the `with_qutip_qip_stub` fixture to use `tempfile` to create a temporary folder, add that temporary folder temporarily to sys.path, and write into the temporary folder a very tiny package with modules in it.

hodgestar (changes requested):
I had a few more ideas about how to test this nicely.

qutip/qip.py
```diff
@@ -0,0 +1,13 @@
...
+
+try:
+    import qutip_qip
+    del qutip_qip
+    sys.modules["qutip.qip"] = sys.modules["qutip_qip"]
+except ImportError:
+    raise ImportError(
+        "Importing 'qutip.qip' requires the 'qutip_qip' package. Install it "
+        "with `pip install qutip-qip` (for more details, go to "
+        "https://qutip-qip.rtfd.io/en/stable/)."
```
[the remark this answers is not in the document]
HGSilveri: I was going to go with that, but the canonical name goes to the "latest" and @BoxiLi asked for the "stable"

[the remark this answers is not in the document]
BoxiLi: @HGSilveri I just checked and found out that I can set the canonical one to be directed to the stable version. Now we can use the canonical domain name. Thanks!

qutip/tests/test_qip.py
HGSilveri: That's an ingenious idea! I gave it a shot, let me know what you think

BoxiLi (approved):
Looks great! Also tested it locally with/without an installed `qutip-qip`.

PR #1924, opened 2022-06-08
Reworking settings and core options

Ericgig:
**Description**
In #1812, I worked on removing the tree structure of option and having options per solver instead of having them together.
This way, we don't need to pass `mcsolve` specific options to `mesolve`, etc. and it makes it simpler to know which options are used by which solver.

That PR is getting quite large and went through multiple revision, so I am replacing it with two smaller ones.

In this PR, I simplify the options in core.

`install` options are put back into `settings`, these are most options that the user cannot change, such as the presence of mkl or running in `ipython`. The exceptions is the location of the .qutip folder which is a setting but can be changed. Most of these settings are properties of the object and only computed when being called instead of at import in `__init__.py`. 

`CoreOptions` and `CompilationOptions` are now both made from a simpler `QutipOptions` class instead of the `optionsclass` which generated many method at import. Some options have been moved around: `debug`, `log_handler`, `colorblind_safe` and `use_cython` has been moved to `CoreOptions`.

Since instance of `CoreOptions` cannot be passed to functions using it. It supports context:
```
with CoreOptions(auto_tidyup=False):
    op = sparse1 + sparse2
```
and global options change using:
```
qutip.settings.core['atol'] = 1e-6
```

`CompilationOptions` instance must be passed to `coefficient` to be used.

`configrc` was removed.  It was used to store settings in a file that would be read when importing qutip. This was mainly used to  store the openmp threshold. Without openmp, we don't have any settings to store between sessions.

Some test related to mkl had to be fixed, it seems that they were always skipped before these changes.

**Related issues or PRs**
Replace #1812

**Changelog**
Simplify settings and core options.


qutip/core/options.py
```diff
@@ -1,9 +1,50 @@
...
+try:
+    import cython
+    _use_cython = True
+except ImportError:
+    _use_cython = False
+
+
+class QutipOptions:
+    """
+    Class for basic functionallity for qutip's options.
```
hodgestar, line 14:
```suggestion
    Class for basic functionality for qutip's options.
```

qutip/core/options.py
```diff
@@ -1,9 +1,50 @@
...
+        out = [f"<{self.__class__.__name__}("]
+        for key, value in self.options.items():
+            out += [f"    '{key}': {repr(value)},"]
+        out += [")>"]
+        return "\n".join(out)
+
+
+class CoreOptions(QutipOptions):
     """
     Settings used by the Qobj.  Values can be changed in qutip.settings.core.
```
hodgestar, line 49: Is `Settings used by the Qobj` correct? It feels like these are options for QuTiP as a whole.

hodgestar (approved):
Two small suggestions, but looks good!

hodgestar (approved):

Issue #1894, opened 2022-05-17
Set up towncrier for managing the changelog.


hodgestar:
### Problem Description

Currently the most time consuming part of making releases is managing the change log (#1055 , #1108, #1533). In addition to the time taken, the release manager is often poorly placed to write change log entries for the PRs involved (the PRs might have been merged some time ago, and the release manager might not have been the author).

### Proposed Solution

[Towncrier](https://pypi.org/project/towncrier/) is a Python package used by Python itself and many other libraries to manage the change log.

I propose that we configure towncrier and ensure that each PR adds a towncrier entry.

This includes:
- [x] Add suitable towncrier configuration to the respository.
- [x] Update the release documentation with instructions for using towncrier to update the change log.
- [x] Update the development documentation to include the requirement to add towncrier entries for each PR.
- [x] Update the GitHub PR template to mention the requirement to add a towncrier entry.
- [x] Add a check to the GitHub action tests workflow that checks that a towncrier entry was added.

### Alternate Solutions

There are a few other tools that do similar things to towncrier, and we could use one of those instead if they are better.

I would recommend that we not roll our own solution. The problem is fairly simple, but writing and documenting our own solution and fixing the inevitable bugs is a lot of work. Plus, I would prefer to have a solution that has already made choices that work, rather than us having to make all of those choices ourselves.

divshacker:
Hey, I can try working on this issue.

trentfridey:
Hi - I think I might have a solution for this issue in #1927.  Setting up the utility was pretty straightforward, but I am not sure on how best to shape the process around using it. There are two issues that I foresee:
- Generating a news fragment when making a PR:
 In order to reference PRs in the changelog, you have to pass the PR number to 
 `towncrier create <PR number>.<change type>` 
 So the process would be:
   1. Make draft PR
   2. Get PR number
   3. run towncrier create with PR number and commit & push resulting news fragment to PR
   4. Mark PR as ready

A way around this extra step is to reference the issue number, but you may not want that.
- When generating changelogs for releases, towncrier needs to be passed the version number in one of three ways (from the documentation):
  - For Python 2/3 compatible projects, a __version__ in the top level package. This can be either a string literal, a tuple, or an [Incremental](https://github.com/hawkowl/incremental) version.
  - Manually passing --version=<myversionhere> when interacting with towncrier.
  - Defining a version option in a configuration file
 Since there's no __version__ string or configuration file, I guess we would use the second option when calling `towncrier build`?

[...]

PR #1927, opened 2022-06-10
[unitaryhack] Add towncrier for managing the changelog

trentfridey:
**Description**
This sets up the towncrier utility to manage the changelog

- Add suitable towncrier configuration to the respository.
- Update the release documentation with instructions for using towncrier to update the change log.
- Update the development documentation to include the requirement to add towncrier entries for each PR.
- Update the GitHub PR template to mention the requirement to add a towncrier entry.
- Add a check to the GitHub action tests workflow that checks that a towncrier entry was added.

**Related issues or PRs**
#1894

**Changelog**
See doc/changes/1927.doc


hodgestar:
@trentfridey Thank you for looking at this. It looks good to me, but I'd like @Ericgig and @AGaliciaMartinez to read and approve too since they're also developing very actively.

I see this is still marked as a work in progress -- are there changes you still want to make from your side?

I guess we also need to add fragments for all the PRs already merged for milestone 4.7.1. If you want to give that a try that would be awesome, but otherwise I'm happy to do it next week when I have a bit more time again.

AGaliciaMartinez (changes requested):
Thank you for the contribution! I added some suggestions but my biggest concern is regarding how things are explained in the documentaiton. I think we should try to not force the user to install towncrier _and_ use it from the command line. This is fine with me but I think it will be a barrier for new contributors. Since it is possible to manually create a file with the changelog, we should favor the manual method as it does not require any additional setup for the user. For instance, this is what pytest does in their repository (https://github.com/pytest-dev/pytest/blob/main/CONTRIBUTING.rst). I think an approach similar to the one followed by pytest would suit us best.

Ericgig:
Is there a way to set towncrier to not require the PR number?
Otherwise we need to open a PR without the changelog, which will fail the towncrier test, to modify it to add the change log...
It would a lot better if we could add it when developing or have a towncrier read it from the Changelog section of PR description.

trentfridey:
> Is there a way to set towncrier to not require the PR number? Otherwise we need to open a PR without the changelog, which will fail the towncrier test, to modify it to add the change log... It would a lot better if we could add it when developing or have a towncrier read it from the Changelog section of PR description.

Yes, I think we could use the issue number that the change is for, if this workflow makes sense for QuTiP (I don't know, I'm a first-time contributor). This is what [pytest does](https://github.com/pytest-dev/pytest/blob/main/CONTRIBUTING.rst#preparing-pull-requests).

doc/development/contributing.rst
```diff
@@ -132,6 +132,20 @@ You can also use the ``-k`` selector to only run tests whose names include a par
...
+When making a pull request, we require that you add a towncrier entry along with the code changes.
+The way to do this is to run 
+
+   towncrier create <PR number>.<change type>
+
+where the PR number should be substituted for ``<PR number>``, and ``<change type>`` is either ``feature``, ``bugfix``, ``doc``, ``removal``, ``misc``, or ``deprecation``,
+depending on the type of change included in the PR.
+
+Running this will create a file in the ``doc/changes`` directory with a filename corresponding to the argument you passed to ``towncrier create``.
+In this file, you should add a short description of the changes that the PR introduces.
```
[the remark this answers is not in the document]
trentfridey: Yes I can change this once we have figured out what we want to do about the PR number being required as @Ericgig pointed out

AGaliciaMartinez:
Unfortunately, not all the PR we open are related to an existing issue.

Could it be possible to include the creation of the changelog as part of a github workflow? If we are able to read the PR number and the Changelog section of the PR description in a workflow, we may be able to create the changelog file during the merge step to later be used with towncrier. This would also mean that new contributors would only require to write the Changelog section of the PR description (which they already have to do), instead of having to create a new file.

hodgestar:
We currently link to the PR in the changelog text, and that that seems a good thing to keep, so we already have to write or update the text after the PR is created.

I'm a bit against pulling information from the PR description. My observation is that the descriptions (even my own) are not very good, because they're written when the PR is just opened, before all of the improvements that happen during review.

@trentfridey Does the PR number have to appear in the filename or confer any special advantages? Or can one just use a meaningful unique name for the file? E.g. `fix-20220614-add-towncrier`?

hodgestar (approved):

hodgestar:
@Ericgig @AGaliciaMartinez I've approved this PR and I'd like to merge it once @trentfridey has fixed the "Verify Towncrier entry added". I actually preferred the situation before @trentfridey removed the requirement for the PR number -- it was simpler and didn't rely on us to remember to write those numbers into the changelog.

For UnitaryHack, I think we have to merge it before the end of tomorrow for it to count, and I'd like to do that since it's a big improvement and now probably we just need to try it and tweak how it is set up if needed.

doc/development/contributing.rst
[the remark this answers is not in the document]
trentfridey: addressed in [79b9c1f](https://github.com/qutip/qutip/pull/1927/commits/79b9c1fcec317c5b6d843adb678d4e96a2a0da2d)

AGaliciaMartinez:
I also prefer to have the PR number for reference in the changelog. Otherwise this looks great. Thanks @trentfridey!

AGaliciaMartinez (approved):

AGaliciaMartinez:
@trentfridey, there seems to be an error in the newly added towncrier github workflow that needs to be fixed before merging. Would you mind taking a look?

Ericgig:
I don't want to approve or block it.
I can't understand the template so I don't know if the generated changelog will be useful for release.
And the tests don't pass...

Personally the issue number as the tag looks good to me. Yes we have many PRs without issues, but it may be a good idea to open them just to tell other what we are working on and it's easier for first time contributor who take a `good first issue`.

hodgestar:
> I don't want to approve or block it.

I don't want to rush it either. I asked the UnitaryHack organizers if we could merge it after the deadline and still have it count for @trentfridey and they said yes!

Let's restore things to the previous "PR name required and no template" state and then we can merge something that we all like without rushing too much.

@trentfridey Thanks for all your work so far and @Ericgig and @AGaliciaMartinez thank you for reviewing with a bit of a deadline (which is not how we usually work).

trentfridey:
Ok, I've reverted to using the PR number for the news fragment file names, and updated the contributing/PR instructions.

I also looked into why the `towncrier-check` test is failing. It looks like towncrier is not finding the `pyproject.toml` config file because it is not running in the root directory. I added a `working-directory: /` specification to the action step to address this, hopefully this should fix it.

pyproject.toml
```diff
@@ -22,3 +22,33 @@ build-frontend = "pip"
...
+package = "qutip"
+package_dir = "qutip"
+directory = "doc/changes"
+filename = "doc/changelog.rst"
+[[tool.towncrier.type]]
+directory = "deprecation"
+name = "Deprecations"
+showcontent = true
+[[tool.towncrier.type]]
+directory = "features"
```
hodgestar, line 36:
```suggestion
directory = "feature"
```

pyproject.toml
```diff
@@ -22,3 +22,33 @@ build-frontend = "pip"
...
+name = "Features"
+showcontent = true
+[[tool.towncrier.type]]
+directory = "bugfixe"
```
hodgestar, line 40:
```suggestion
directory = "bugfix"
```

hodgestar (approved):
I made two small suggestions to fix typos in the names of change types in the .toml file. I will apply these shortly.

hodgestar:
@AGaliciaMartinez @Ericgig Looks good to me. I will merge later this afternoon if there are no objections.

.github/workflows/tests.yml
```diff
@@ -181,6 +181,16 @@ jobs:
           COVERALLS_PARALLEL: true
         run: coveralls --service=github
 
+  towncrier-check: 
+    name: Verify Towncrier entry added
+    runs-on: ubuntu-latest
+    working-directory: /
```
hodgestar, line 187: GitHub rejects `working-directory` here as invalid -- it would need to be inside the step (i.e. same level a `run:`) -- but I'm also moderately sure that the default working directory is already the folder that contains `pyproject.toml`, so I'm not sure what is happening here.

hodgestar (changes requested):

pyproject.toml
```diff
@@ -22,3 +22,33 @@ build-frontend = "pip"
 select = "cp3?-*"
 manylinux-x86_64-image = "manylinux2010"
 manylinux-i686-image = "manylinux2010"
+
+[tool.towncrier]
+package = "qutip"
+package_dir = "qutip"
```
hodgestar, line 28:
```suggestion
```
Remove the package configuration -- we are using the directory directly.

hodgestar (approved):
I fixed the issue, which was firstly that the repo hadn't been checked out at all, and then that GitHub doesn't check out enough of the history of the repository by default in GitHub actions.

nathanshammah:
@trentfridey we cannot find your contact details on unitaryHACK, please sign up https://airtable.com/shrJeycewBFqdot2B to be able to claim your bounty!

trentfridey:
I just signed up. Let me know if you need anything else.

PR #1931, opened 2022-06-13
Reset and clear up figures more correctly in dynamics-time guide.

hodgestar:
**Description**
The guide for time-dependent dynamic includes plots fro the guide before (stochastic dynamics) because it fails to reset its context and properly close figures. This PR fixes that.

The same goes for many other uses of `.. plot::` in the guides and other documentation, but I will leave fix those for when someone has time to go through them more carefully.

See https://qutip.org/docs/latest/guide/dynamics/dynamics-time.html for the issue.

**Related issues or PRs**
None

**Changelog**
Fix spurious plots that appear after many code blocks in the dynamics-time guide.


AGaliciaMartinez (approved):
I find it weird that we need to add the `:nofigs:` options to the plot directive every time but we seem to do that already in other parts of the documentation. Otherwise, this change looks good to me and it does fix the issue (at least locally).

Issue #1841, opened 2022-03-28
summing a list of qobj in qutip 5.0


shohamjac:
### Bug Description

Summing a list of qobj does not work anymore since the change in the qobj class.

### Code to Reproduce the Bug

```shell
from qutip import *
qobj_list = [fock(3,i) for i in range(3)]
sum(qobj_list)
```

### Code Output

```shell
Traceback (most recent call last):
  File "C:\Users\shoha\miniconda3\envs\qutip_fork\lib\site-packages\IPython\core\interactiveshell.py", line 3361, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-27c43f8aaf45>", line 1, in <cell line: 1>
    sum(qobj_list)
  File "C:\Users\shoha\miniconda3\envs\qutip_fork\lib\site-packages\qutip-5.0.0.dev0+8187a26-py3.9-win-amd64.egg\qutip\core\qobj.py", line 443, in __radd__
    return self.__add__(other)
  File "C:\Users\shoha\miniconda3\envs\qutip_fork\lib\site-packages\qutip-5.0.0.dev0+8187a26-py3.9-win-amd64.egg\qutip\core\qobj.py", line 153, in out
    raise ValueError(msg)
ValueError: incompatible dimensions [[3], [1]] and [[1], [1]]
```

### Expected Behaviour

Summing should be equal to (qobj_list[0]+qobj_list[1]+....+qobj_list[N-1])

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+8187a26
Numpy Version:      1.16.6
Scipy Version:      1.7.3
Cython Version:     0.29.25
Matplotlib Version: 3.4.3
Python Version:     3.9.7
Number of CPUs:     4
BLAS Info:          INTEL MKL
INTEL MKL Ext:      False
Platform Info:      Windows (AMD64)
```

### Additional Context

If I understand correctly, this happens because sum starts with 0 and then every object in the list, as written here:
http://www.marinamele.com/2014/04/modifying-add-method-of-python-class.html

Ericgig:
In python `sum` start at `0`. 
So it's doing `(0 + qobj_list[0] + ...)`
And `int + Qobj` is only defined for square `Qobj`.
`sum(qobj_list[1:], start=qobj_list[0])` would work.

hodgestar:
Fixed in #1934.

PR #1934, opened 2022-06-14
Allow `sum(kets)`

Ericgig:
**Description**
Allow `qobj +/- 0` for `Qobj` of any shape and type to allow the use of `sum(kets)` as discussed in #1841.
Addition and subtraction of any other number will still raise an error if the object is not square as before.

**Related issues or PRs**
Issue raised in #1841

**Changelog**
Allow `ket + 0`.

### DIFF
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -1081,3 +1081,19 @@ def test_contract(expanded, contracted, inplace):
     assert out.dims == contracted
     assert out.shape == qobj.shape
     assert np.all(out.full() == qobj.full())
+
+
+@pytest.mark.parametrize(["shape"], [
+    pytest.param((5, 1), id='ket'),
+    pytest.param((5, 2), id='tall'),
+    pytest.param((1, 5), id='bra'),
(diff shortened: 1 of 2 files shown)


qutip/tests/core/test_qobj.py
```diff
@@ -1081,3 +1081,19 @@ def test_contract(expanded, contracted, inplace):
...
+    pytest.param((2, 5), id='wide'),
+    pytest.param((3, 3), id='oper'),
+])
+def test_sum_zero(shape):
+    data = np.random.rand(*shape) + 1j*np.random.rand(*shape)
+    ket = qutip.Qobj(data)
+    assert ket + 0 == ket
+    assert ket - 0 == ket
+    assert 0 + ket == ket
+    assert 0 - ket == -ket
```
hodgestar, line 1099: Could we name these 'obj' or 'qobj' since they are only a `ket` in one of the five test runs?

hodgestar (changes requested):
We should add a test that uses `sum`, but otherwise this looks great.

hodgestar (approved):

Issue #1904, opened 2022-05-26
Restore openmp support on `dev.major`


Ericgig:
### Problem Description

`master` has qutip/qutip/cy/openmp/src/[zspmv_openmp.hpp](https://github.com/qutip/qutip/blob/master/qutip/cy/openmp/src/zspmv_openmp.hpp), for csr @ dense product with openmp, but we did not have the time to add when making the data layer for `dev.major`. In v4, to use openmp, the user needs to compile from source locally with `python setup.py install --with-openmp` and then an option has to be passed to the solver at run time to use it.

For v5, it would be great if openMP was part of the Data layer and used without much thought by the user. (No need to pass `Options(use_openmp=True)` to `mesolve`).  We would probably keep a flag and threshold as an option in core or settings so we can run tests with and without.

We should add support for cython's `prange`. Presently, using it breaks mac's tests. If made available, there are quite a few functions for the data layer that could gain from it.

While openmp is available in v4, it was not used much since it can only be used when compiling locally and the options isn't well known. Ideally, it would be packaged in the wheels (for all but Mac?) instead of needing a local installation. However it does not seems easy, [scipy decided against it](https://github.com/scipy/scipy/blob/38261c1141b8aa80bceeef419911fb4f31df5372/doc/source/dev/toolchain.rst#openmp-support), because it does not mix well with [other parallel environments](https://github.com/scipy/scipy/issues/10239#issuecomment-795030817). It may be possible with conda since it has a openmp package. 

### Proposed Solution

Tasks:
- Restore the compilation option for openmp and have `matmul_csr_dense_dense` use `zspmvpy_openmp`.
- Add options in core to control openmp. In `parallel_map`, turn it of when using multiprocessing.
- Add `prange` in a function (`mul_dense`) and ensure that tests passes and wheels build.
- Add `prange` elsewhere where useful.

Ericgig:
Interesting discussion about the relationship between OpenMP and multiprocessing -- https://scikit-learn.org/stable/faq.html#why-do-i-sometime-get-a-crash-freeze-with-n-jobs-1-under-osx-or-linux (in case this is relevant when bringing OpenMP back).

_Originally posted by @hodgestar in https://github.com/qutip/qutip/issues/1948#issuecomment-1171423571_

ajgpitch:
Assigned to @veronikakurth due to her success with previous C related issues

PR #1941, opened 2022-06-22
Drop tests using older python versions.

Ericgig:
**Description**
Remove tests using python 3.6 and 3.7 and openmp tests.
SInce we need to add a new job for each python version, we also need to remove older jobs so the whole test suite don't get bloated. Numpy has a suggestion for version support in scientific project in [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) which we can follow to decide when we can remove some tests and if it's fine to use newer feature forcing us to improve our requirements.

I added a note in the contributing and release guides to refer to this when adding a change using feature of newer version of package we use.


AGaliciaMartinez:
Should we also update the setup.cfg file with the latest version of numpy and python employed in tests? Allowing the installation of previous numpy versions that are not included in the CI may lead unanticipated bugs. Is there any reason to allow older versions of numpy in the setup.cfg (currently `numpy>=1.16") instead of matching those with the versions used for testing?

doc/development/contributing.rst
```diff
@@ -85,6 +85,15 @@ Cython will detect and compile only the files that have been changed, so this co
...
 
+.. note::
+
+   When developing for a new major or minor version (not a patch release), new feature introduced could require newer versions of some of the required packages.
+   In that case, update the requirements as needed.
+   We follow `NEP29`_ for minimum supported versions of scipy and numpy.
+   For other dependencies, we support the oldest version that is supporting the suggested python and numpy's versions.
+
+.. _NEP29: https://numpy.org/neps/nep-0029-deprecation_policy.html
+
```
AGaliciaMartinez, line 96: I think this information is mostly necessary only for core developers. I would suggest to keep only the documentation on `reslease_distribution.rst`. This seems to also be the approach followed by scipy for instance.

doc/development/release_distribution.rst
```diff
@@ -46,6 +46,17 @@ For every change that is going to be part of your release, make sure that:
...
+.. _update-requirement:
+
+Updating the Requirements
+-------------------------
+
+Ensure that QuTip's tests pass on the oldest version supported in the requirements.
+On major and minor version, requirements can be updated to remove support of older version of some packages.
+We follow `NEP29`_ for minimum supported versions.
+
+.. _NEP29: https://numpy.org/neps/nep-0029-deprecation_policy.html
```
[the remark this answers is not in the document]
AGaliciaMartinez: I would also suggest to be more specific on what we mean by "we follow NEP29". Referencing NEP29 is great but I think it is best to be explicit and explain what we support and how. For example, adding something similar to this may help:

> This project supports:
> 
> - All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
> - All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
> 
>     In setup.py, the python_requires variable should be set to the minimum supported version of Python. All supported minor versions of Python should be in the test matrix and have binary artifacts built for the release.
> 
>     Minimum Python and NumPy version support should be adjusted upward on every major and minor release, but never on a patch release.
> 

This is copied from the NEP29.

doc/development/release_distribution.rst
```diff
@@ -46,6 +46,17 @@ For every change that is going to be part of your release, make sure that:
 
 Please make a normal PR to ``master`` correcting anything missing from these points and have it merged before you begin the release, if necessary.
 
+.. _update-requirement:
+
+Updating the Requirements
+-------------------------
+
+Ensure that QuTip's tests pass on the oldest version supported in the requirements.
```
AGaliciaMartinez, line 54:
```suggestion
Ensure that QuTiP's tests pass on the oldest version supported in the requirements.
```

AGaliciaMartinez (commented):
Thank you Eric for tackling this! I think following NEP29 for which versions we support and which ones we don't is a great idea. However, as the documentation stands it is a little bit confusing for me. I added a few comments on the documentation explaining what I mean and also included a suggestion. 

I would like to also suggest updating the run-time requirements in this PR to make them match with those in the tests. Would that be possible?

.github/workflows/tests.yml
```diff
@@ -33,29 +33,21 @@ jobs:
...
           # No MKL runs.  MKL is now the default for conda installations, but
           # not necessarily for pip.
           - case-name: no MKL
-            os: ubuntu-latest
-            python-version: "3.7"
-            nomkl: 1
-
-          # OpenMP runs.  We only really care about getting OpenMP working
-          # properly on Linux at the moment.
-          - case-name: OpenMP
```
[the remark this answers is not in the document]
Ericgig: v5 does not support openmp. Adding this options does nothing for now, and it may stay this way, see #1904.

doc/development/contributing.rst
Ericgig: I was not sure where to add it, so I added it in both. But I think it could be useful to have something here:
If we say that we support python >=3.8, that mean we can now use the `:=` operator, but not `dict | dict` which was introduced in python 3.9.
This information should be available to contributors.

AGaliciaMartinez: I see your point. However, I think that the average contributor will care only about the specific supported versions, not the procedure employed to select them. The supported versions will be ultimately dictated by tests, so if tests fail we can discuss in the appropriate PR to update the  tests dependencies and run-time requirements. What do you think about using instead the following note:
```suggestion
.. note::

QuTiP follows `NEP29`_ when selecting the supported version of its dependencies.
The currently supported versions can be checked in the `setup.cfg`.
These coincide with the versions employed for testing in continuous integration.

In the event of a feature requiring a version upgrade of python or a dependency, it will 
be considered appropriately in the pull request.
In any case, python and dependency upgrades will only happen in mayor or minor 
versions of QuTiP, not in a patch.

.. _NEP29: https://numpy.org/neps/nep-0029-deprecation_policy.html
```
Feel free to reject this suggestion if you think is not necessary.

Ericgig: `setup.cfg` should list what we actually support, not what we want to support. 

But I agree that having a specific version written somewhere would be useful. I will put it in the road-map, this show that it is more a target than something fixed: if we released a new version last week, we would have to support numpy 1.19.

doc/development/contributing.rst
```diff
@@ -85,6 +85,18 @@ Cython will detect and compile only the files that have been changed, so this co
    When undertaking Cython development, the reason we use ``python setup.py develop`` instead of ``pip install -e .`` is because Cython's changed-file detection does not reliably work in the latter.
    ``pip`` tends to build in temporary virtual environments, which often makes Cython think its core library files have been updated, triggering a complete, slow rebuild of everything.
 
+.. note::
+
+    QuTiP follows `NEP29`_ when selecting the supported version of its dependencies.
+    To see which versions are planned to be supported in future release, please refer to the :ref:`release roadmap`.
```
AGaliciaMartinez, line 91:
```suggestion
    To see which versions are planned to be supported in the next release, please refer to the :ref:`release roadmap`.
```

doc/development/release_distribution.rst
```diff
@@ -46,6 +46,22 @@ For every change that is going to be part of your release, make sure that:
...
+-------------------------
+
+Ensure that QuTiP's tests pass on the oldest version supported in the requirements.
+On major and minor version, requirements can be adjusted upwards, but patch release must not change minimum requirements.
+We follow `NEP29`_ for minimum supported versions ::
+
+    - All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
+    - All minor versions of numpy and scipy released in the 24 months prior to the project, and at minimum the last three minor versions.
+
+Ensure that PyPI wheels and conda builds cover at least these versions.
```
AGaliciaMartinez, line 61:
```suggestion

If dependency versions need to be updated, update them in the master branch. The following files may need to be updated: `.github/workflows/tests.yml`, `setup.cfg` and `roadmap.rst`. Finally, ensure that PyPI wheels and conda builds cover at least these versions.
```

AGaliciaMartinez (approved):
I added two small suggestions in the documentation but otherwise this looks good to me.

hodgestar:
I've been thinking about how this interacts with our build and test infrastructure, and I'm writing some of those thoughts here partly for myself:

- We currently build with oldest-supported-numpy, and we want to keep that because which is the earliest supported numpy varies by Python version and operating system and changes over time, so it's good if someone else worries about that for us.

- We require for installation currently numpy >= 1.16.6 and that is probably what we want to update to match the NEP. There is a question here of whether we should specify our minimum supported numpy in "install_requires" in "setup.cfg" or not.

- The majority of our test matrix should *build* qutip with oldest-supported-numpy to match what we distribute.

- We should include at least one test run for the oldest numpy we require (i.e. oldest NEP) and one for the most recent (i.e. latest numpy).  

- We should include at least one test run for the oldest Python we support and one for the newest.

- We should include at least one test run for the oldest SciPy we support and one for the newest.

- The NEP doesn't say anything about SciPy, but I propose we select the oldest SciPy that supports our oldest NEP numpy.

And as Asier suggested, I think it would be good to write this all down. The test matrix requirements could just be a big comment in the GitHub action file.

hodgestar:
Thanks for documenting everything, @Ericgig. It looks good to me now, but I didn't merge in case you wanted to make any last changes.

PR #1942, opened 2022-06-23
Fix exception type in sesolve

eendebakpt:
**Description**

Return `TypeError` in `sesolve` when the argument `H` is of incorrect type.


doc/changes/1924.misc.py
```diff
@@ -0,0 +1 @@
+Return TypeError instead of Exception for type error in sesolve argument
```
hodgestar, line 1: Thank you for adding the changelog with towncrier -- I think you are officially the first person to use this (after the PR that added it).

I think the filename should be just "1924.misc" though, without the `.py`.

qutip/tests/test_sesolve.py
```diff
@@ -352,3 +352,10 @@ def test_sesolve_bad_e_ops():
     tlist = np.linspace(0, 20, 200)
     with pytest.raises(TypeError) as exc:
         sesolve(H, psi0, tlist=tlist, e_ops=[qeye(3)])
+
+def test_sesolve_bad_H():
+    H = sigmaz(),
+    psi0 = basis(2, 0)
+    tlist = np.linspace(0, 20, 200)
+    with pytest.raises(TypeError) as exc:
+        sesolve(H, psi0, tlist=tlist, e_ops=[qeye(3)])
```
hodgestar, line 361:
```suggestion
        sesolve(H, psi0, tlist=tlist, e_ops=[qeye(3)])
    assert str(exc.value).startswith("Invalid H:")
```
I guess we should check that the right exception message is raised too.

hodgestar (changes requested):
Thank you for making the suggested change and adding the changelog. I was going to offer to make the changelog for you, but you beat me to it.

doc/changes/1924.misc.py
eendebakpt: I was already wondering why the `changes` folder was so empty :-)

hodgestar (approved):
Approved! Will merge once the CI runs complete successfully.

PR #1943, opened 2022-06-24
Fix _repr_svg_ for recent IPython

eendebakpt:
The ipython `print_figure` method was modified to return `str` instead of `bytes` when the `svg` option is used. See https://github.com/ipython/ipython/commit/4a2e85faf8e30c70a5d2c1af6bb45e2cdee5a067

This PR fixes the conversion in `qutip`, but does it using a check so that older versions of ipython still work.

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -293,7 +293,7 @@ class Bloch:
     def _repr_svg_(self):
         from IPython.core.pylabtools import print_figure
         self.render()
-        fig_data = print_figure(self.fig, 'svg').decode('utf-8')
+        fig_data = print_figure(self.fig, 'svg')
         plt.close(self.fig)
         return fig_data
 
--- a/qutip/tests/test_bloch.py
+++ b/qutip/tests/test_bloch.py
@@ -361,3 +361,7 @@ class TestBloch:
             vector_kws = [vector_kws]
         self.plot_vector_test(fig_test, copy.deepcopy(vector_kws))
         self.plot_vector_ref(fig_ref, copy.deepcopy(vector_kws))
+
+def test_repr_svg():
+    svg = Bloch()._repr_svg_()
+    assert isinstance(svg, str)
--- a/qutip/tests/test_bloch.py
+++ b/qutip/tests/test_bloch.py
@@ -363,5 +363,9 @@ class TestBloch:
         self.plot_vector_ref(fig_ref, copy.deepcopy(vector_kws))
 
 def test_repr_svg():
-    svg = Bloch()._repr_svg_()
-    assert isinstance(svg, str)
+    try:
+        # test when ipython is available
+        svg = Bloch()._repr_svg_()
+        assert isinstance(svg, str)
+    except ImportError:
+        pass


hodgestar:
@eendebakpt Could you merge the latest master branch into this PR so that the diff looks right in the PR?

AGaliciaMartinez:
Thanks @eendebakpt for the fix! 
However, a similar fix was already included in dev.major in #1918. The only difference with your approach is that I did not include a try expect for import of iPython because the relevant change for the print_figure function was done 8 years ago. Besides, there is also the issue that currently our tests workflows do not install ipython so `repr` in your tests will not call `_repr_svg_`. Sorry I did not include this fix in master as well, but if you want it to be fixed I recommend cherry-picking the commits of #1918.

eendebakpt:
@AGaliciaMartinez What exactly is this `dev.major` branch? . `dev.major` diverged from the qutip `master` quite a lot.
I could not find anything relevant on https://qutip.org/docs/latest/development/contributing.html

The only thing I could find is https://github.com/qutip/qutip/milestone/18, but it is not clear what the timeline is for that.

hodgestar:
> What exactly is this `dev.major` branch?

Hi @eendebakpt. `dev.major` is the branch in which we are developing QuTiP version 5. We're still doing important fixes for 4.7.1, but we're not planning to release any more feature for the 4.X series (although *maybe* that could change -- we'll see how the move to 5 goes).

I'll update the contributing guidelines to mention QuTiP 5.

QuTiP 5 has been in the works for a long time, but we're now in the "cleaning things up" phase, and hope to release an alpha version in the next month or so.

qutip/bloch.py
```diff
@@ -284,16 +284,24 @@ def __str__(self):
         return s
 
     def _repr_png_(self):
-        from IPython.core.pylabtools import print_figure
+        try:
+            from IPython.core.pylabtools import print_figure
+        except ImportError:
+            return None
```
hodgestar, line 290: I would prefer not to have these `try ... except ...`  and instead just let the import error propagate. I would expect only IPython will call `_repr_png_` and `_repr_svg` so if IPython isn't installed, something strange has happened. Returning `None` just silences the error.

qutip/tests/test_bloch.py
```diff
@@ -361,3 +361,10 @@ def test_vector(self, vector_kws, fig_test, fig_ref):
             vector_kws = [vector_kws]
         self.plot_vector_test(fig_test, copy.deepcopy(vector_kws))
         self.plot_vector_ref(fig_ref, copy.deepcopy(vector_kws))
+
+    def test_representations(self):
+        b = Bloch()
+        assert 'Bloch' in repr(b)
+        assert str(b).startswith('Bloch data:')
+        repr_svg = b._repr_svg_()
+        assert repr_svg is None or isinstance(repr_svg, str)
```
hodgestar, line 370: As @AGaliciaMartinez pointed out, this call to `_repr_svg_` exits early and returns `None` since ipython isn't installed during the tests.

hodgestar (commented):
I would recommend following @AGaliciaMartinez's suggestion and making this look more like the changes in #1918 (see specific comments for my reasoning). Happy to hear reasons to do things differently though.

eendebakpt:
@hodgestar @AGaliciaMartinez Thanks for the info. I decided to cherry-pick one of AGaliciaMartinez commits and push that. I will also look into qutip 5

hodgestar:
I'm closing this in favour of just cherry-picking @AGaliciaMartinez's fix for dev.major (#1918) to master. See #1970.

PR #1947, opened 2022-06-28
Options per solver

Ericgig:
**Description**
This change solver's options to be per solver. Thus mcsolve specific options are not included in options passed to sesolve etc.
It also simplify the options class which is now closer to a `dict` than to a `dataclass`.

`SovlerOptions` take a new optional parameter `solver` to know which options to support.  By passing it, it will fill the instance with the defaults and update the doc string. Passing an options not supported by the solver will raise an error. When not used, any options items will be accepted.

```
>>> SolverOptions(atol=1e-5, store_ados=True)
SolverOptions(solver='', atol=1e-05, store_ados=True)

>>> SolverOptions('sesolve', atol=1e-5, store_ados=True)
KeyError: "'store_ados' is not a supported options."

>>> SolverOptions('heomsolver', atol=1e-5, store_ados=True)
SolverOptions(solver=heomsolver, atol=1e-05, store_ados=True)
```

`SolverOptions` can be used as `dict`, but most methods only use provided keys:
```
>>> {**SolverOptions('sesolve', atol=1, rtol=2)}
{'atol': 1, 'rtol': 2}
```
`store_states` etc. are not in `opt.keys()`.
The exceptions are `in` and `__getitem__` which knows of the default values:
```
>>> opt = SolverOptions('sesolve', atol=1, rtol=2)
>>> "store_states" in opt
True
>>> print(opt["store_states"])
None
```

`print` create a human readable list of options, including default:
```
>>> print(SovlerOptions("sesolve", atol=1e-5, store_states=True))

Options for sesolve:
    progress_bar      : 'text'                                                 (default)
    progress_kwargs   : {'chunk_size': 10}                                     (default)
    store_final_state : False                                                  (default)
    store_states      : True                                                 
    normalize_output  : True                                                   (default)
    method            : 'adams'                                                (default)
Options for adams integrator:
    atol       : 1e-05                                                       
    rtol       : 1e-06                                                         (default)
    nsteps     : 2500                                                          (default)
    order      : 12                                                            (default)
    first_step : 0                                                             (default)
    max_step   : 0                                                             (default)
    min_step   : 0                                                             (default)
```

Some integrators options will change depending on the method:
```
>>> "order" in SovlerOptions("sesolve", method="adams")
True
>>> "order" in SovlerOptions("sesolve", method="dop853")
False
```
passing options not supported by the options will raise an error.

Solver accept options as `dict` or `SolverOptions`. They do not refuse options for other solver.
Solver can support options changes: `my_solver.options['atol'] = 1e-10`, will reset the integrator to use the new value.


hodgestar:
I've had a thorough look. Simplifying the SolverOptions class is an excellent idea, but I think the PR currently doesn't go far enough and ends up with the solver classes, the integrator classes and the solver option classes required to share too much internal state with each other.

As a next step to try untangle things, I suggest the following:

- Let's remove the `known_solvers` and the solver `name` from the SolverOptions and instead have the base solver pass the allowed options with defaults in to the SolverOptions constructor directly. If the allowed options list is `None`, then we can allow all options (for the convenience of users who really want to make their own SolverOptions instances).
- Having the options able to notify a listener of changes seems great -- let's keep that.
- When a solver is created, let's have the base solver create a new SolverOptions instance, either from a dictionary or another SolverOptions instance, but with the allowed options and defaults now specified.
- To deal with the integrator options, let's go back to having a single option named `ode`, but have the value of `ode` have type `SolverOptions`, and be a child of the solver options (if the child has a value modified, it should notify the parent of the change).
- For convenience, when passing options as a dictionary, one should be able to use `{"ode.method": "vern9"}` or `{"ode": {"method": "vern9"}`

I think that should be enough to disentangle things from each other. The base solver would need to update the integrator if integrator options change and deal with any changes to the base options.

It's not completely clear to me whether the result options should also be sub-options or just be included in the solver options. I'm happy for them to be included in the solver options, as long as the solver options are passed to the result class constructor. That keeps things a bit simpler and if solver needs new result options it can always create its own result sub-class and just process its options there.

Thoughts?

Ericgig:
>     Let's remove the `known_solvers` and the solver `name` from the SolverOptions and instead have the base solver pass the allowed options with defaults in to the SolverOptions constructor directly. If the allowed options list is `None`, then we can allow all options (for the convenience of users who really want to make their own SolverOptions instances).

Users who makes their own SolverOptions are the base case. Using the `??solve` function with an `Options` instance and with documentation for the options items in the `Options` class, is the way in v4 and will probably stay that way for a while. 
Having a options not knowing about the solver let it accept bad options and typos. Also it hides the options documentation: solver's options could be in the `??solve` function, but not integrator options. The integrator was merged into the solver in v4 and is not made to be used directly by the average user.  So the user should not need to know they exist to use qutip, nor refer to them directly to see which options are supported.

>     For convenience, when passing options as a dictionary, one should be able to use `{"ode.method": "vern9"}` or `{"ode": {"method": "vern9"}`.

Not sure I understand why this would be convenient, flat seems a easier to use and less confusing. Ideally the user should not need to know that there are 2 levels of options.

hodgestar:
> Users who makes their own SolverOptions are the base case. Using the `??solve` function with an `Options` instance and with documentation for the options items in the `Options` class, is the way in v4 and will probably stay that way for a while. Having a options not knowing about the solver let it accept bad options and typos. Also it hides the options documentation: solver's options could be in the `??solve` function, but not integrator options. The integrator was merged into the solver in v4 and is not made to be used directly by the average user. So the user should not need to know they exist to use qutip, nor refer to them directly to see which options are supported.

I'm not sure I am quite following, but my thought was that options would be checked by the solver class when they are passed in. The mechanism the solver class would use for doing the check, would be to pass the description of the allowed options to the SolverOptions class, along with the options themselves.

> Not sure I understand why this would be convenient, flat seems a easier to use and less confusing. Ideally the user should not need to know that there are 2 levels of options.

The reason for nesting the ode options a little, is to make it clear which options might change if the ODE method changes. If one hides the structure completely, there is nothing to guide the user as to which options are related to what. E.g. If ODE method "A" took option "opt1" and method "B" took "opt2" then it's much clearer what is happening if one has to pass `"ode": {"method": "A", "opt1": 1}` and `"ode": {"method": "B"}` than `"method": "A", "opt1": 1" and `"method": "B", "opt2": 2`. In the flat case, one might accidentally have `"method": "B": "opt1": 1` and "opt1" would be ignored.

I am not too hung up on the exact approach for this bit though -- I just want it to be clear to the user what is going wrong, and for it to be a little be hard to options for different ODE methods clash with each other or other solver options.

I wrote quite a lot in my previous post, but mostly I was suggesting one way to simplify and disentangle things more. I am happy for use to try other ways too.

Ericgig:
The main function of the `Options` class is to hold the documentation of the options supported by solver, mostly integrator related ones are an issue. Since sub-packages can add new integration methods, it should be somewhere dynamic that can fetch them at runtime.

We can't have them in the description of the function interface directly, but it must be clear from that documentation where to find them.
We can't have them in the documentation of the solver class. It could be in a solver instance, but that too late: needing to initialize a `BrSolver` or `HEOMSolver` just to know which options `bsmesolve` and `heomsolve` support is not user friendly.
The Integrator class is useless for the users: if they want to solve ODE, they should use `scipy.integrate`, for quantum systems evolution, `mesolve` etc. So we should not ask them to look there.

`SolverOptions(name)` make it reasonably easy to check the supported options and their default before using the solver.

Another approach I see is to fix the integrator when using the functions interface, this way we could properly document the function with all items from both the solver and integrator. When using the class interface, we keep both options separated (and make the choice of integrator it's own keyword):
```
solver = SeSolver(H, integrator_method='adams', options=solver_options, ode_options=ode_options)
solver.options # Solver options only

solver.set_integrator('vern7', ode_options)
help(solver.ode_options) # See description of integrator options
```

That way the function calls is very close to v4 and easy to use, and the class interface give access to new features. 
This also makes `krylovsolve` fit more.

Since `method` is used by the solver, not the integrator, it should not be under the `options.ode`, but it seems confusing to have it with the normal options, so having it as an independent parameter could make it clearer.

hodgestar:
I think it's okay for the integrator options to be described in the documentation for the integration -- it can't really go anywhere else. The solver documentation can then point to the documentation for the built-in integrators, and mention that it is possible for third-party packages to provide additional integrators.

Issue #1190, opened 2020-03-01
String-based QobjEvo outputs several compiling warnings on Windows


BoxiLi:
String-based QobjEvo generates several warning messages when it is compiled. The terminal is overwhelmed by those messages when using string-based solvers. Is it possible to suppress/correct them?

Example:
```python
import numpy as np
from qutip import *
tlist = np.array([0., 1., 2., 3.])
H = [[sigmaz(), 'exp(-t)']]
qu = QobjEvo(H, tlist=tlist)
qu.compile()
```

Terminal output:
```
cl : Command line warning D9025 : overriding '/W3' with '/w'
cl : Command line warning D9025 : overriding '/Ox' with '/O1'
cqobjevo_compiled_coeff_4851660043300.cpp
c:\users\boxi\miniconda3\envs\qutip-dev-py3\lib\site-packages\numpy\core\include\numpy\npy_1_7_deprecated_api.h(14) : Warning Msg: Using deprecated NumPy API, disable it with #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
cqobjevo_compiled_coeff_4851660043300.obj : warning 
LNK4197: export 'PyInit_cqobjevo_compiled_coeff_4851660043300' specified multiple times; using first specification
   Creating library C:\Users\Boxi\.pyxbld\temp.win-amd64-3.6\Release\Users\Boxi\.pyxbld\temp.win-amd64-3.6\Release\pyrex\cqobjevo_compiled_coeff_4851660043300.cp36-win_amd64.lib and object C:\Users\Boxi\.pyxbld\temp.win-amd64-3.6\Release\Users\Boxi\.pyxbld\temp.win-amd64-3.6\Release\pyrex\cqobjevo_compiled_coeff_4851660043300.cp36-win_amd64.exp
Generating code
Finished generating code
```

Environment:
```
QuTiP Version:      4.5.0.dev0+0f412cd4
Numpy Version:      1.17.5
Scipy Version:      1.2.1
Cython Version:     0.29.14
Matplotlib Version: 2.2.4
Python Version:     3.6.7
Number of CPUs:     12
BLAS Info:          Generic
OPENMP Installed:   False
INTEL MKL Ext:      False
Platform Info:      Windows (AMD64)
```

hodgestar:
Thanks @BoxiLi. I think we should try address this for QuTiP 5, but I'm not 100% sure where to start so I will leave it for others for the moment.

hodgestar:
[...]

PR #1948, opened 2022-06-29
Activate mcsolve tests on Windows in QuTiP 5

hodgestar:
**Description**
Activate the mcsolve tests in QuTiP 5.

Changes:
- [x] Swap parallel_map over to ProcessPoolExecutor.
- [x] Switch ``mcsolve`` to the new parallel_map. 
- [x] Reactivated mcsolve tests on Windows. 

**Related issues or PRs**
- Continues work from #1853 and #1854 in v4
- #1202 (maybe already resolved in QuTiP 5)
- #1190 (hopefully already resolved in QuTiP 5)

### DIFF
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -155,6 +155,8 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
     if sys.version_info >= (3, 7):
         # ProcessPoolExecutor only supports mp_context from 3.7 onwards
         ctx_kw = {"mp_context": mp_context}
+    else:
+        ctx_kw = {}
 
     os.environ['QUTIP_IN_PARALLEL'] = 'TRUE'
     try:
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -9,13 +9,14 @@ import os
 import sys
 import time
 import threading
+from concurrent.futures import ProcessPoolExecutor
 from qutip.ui.progressbar import progess_bars
 from qutip.settings import available_cpu_count
 
 if sys.platform == 'darwin':
-    Pool = multiprocessing.get_context('fork').Pool
+    mp_context = multiprocessing.get_context('fork')
 else:
-    Pool = multiprocessing.Pool
+    mp_context = multiprocessing.get_context()
 
 
 default_map_kw = {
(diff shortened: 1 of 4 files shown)


hodgestar:
A bit surprisingly, all tests now pass on Windows on QuTiP 5.

I'm not 100% sure the parallel map change fixed everything, or if everything was mostly working before. I would like to keep the parallel map change regardless, because it removes some of the management of making multiprocessing work and leaves it with the core Python developers.

Ericgig (commented):
`ProcessPoolExecutor` being enough to fix most windows tests is surprising.
Now we just need to speed up correlation tests.

qutip/solver/parallel.py
```diff
@@ -145,39 +146,41 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
...
-    except multiprocessing.TimeoutError:
-        pass
+    if sys.version_info >= (3, 7):
+        # ProcessPoolExecutor only supports mp_context from 3.7 onwards
+        ctx_kw = {"mp_context": mp_context}
+    else:
+        ctx_kw = {}
 
+    os.environ['QUTIP_IN_PARALLEL'] = 'TRUE'
+    try:
```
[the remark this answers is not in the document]
hodgestar: I hadn't realized the intention of catching the error. Let me see if I can fix things up so that results are retained properly.

Maybe I can also fix `ProcessPoolExecutor` waiting too long.

qutip/solver/parallel.py
```diff
@@ -145,39 +146,41 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
...
+
+            results = []
+            for f in result_futures:
+                remaining_time = min(end_time - time.time(), job_time)
+                result = f.result(timeout=remaining_time)
+                if reduce_func is not None:
+                    reduce_func(result)
+                else:
+                    results.append(result)
+                progress_bar.update()
```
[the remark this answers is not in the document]
hodgestar: Thanks, I will try a variation on these!

hodgestar:
Interesting discussion about the relationship between OpenMP and multiprocessing -- https://scikit-learn.org/stable/faq.html#why-do-i-sometime-get-a-crash-freeze-with-n-jobs-1-under-osx-or-linux (in case this is relevant when bringing OpenMP back).

hodgestar:
@Ericgig I implemented a version of your suggestion. Let's see how it goes on this test run.

Other notes:
- I switched the fork method to `forkserver` for Linux, as that should play better with threads.
- I left the correlation tests in for now, but will remove them again if they still timeout.

Perhaps we should set some timeouts in the correlation function tests as a way to prevent the whole test run from dying.

hodgestar:
All tests passing again -- ready for another review!

qutip/solver/parallel.py
```diff
@@ -9,13 +9,18 @@
 import sys
 import time
 import threading
+import concurrent.futures
 from qutip.ui.progressbar import progess_bars
 from qutip.settings import available_cpu_count
 
 if sys.platform == 'darwin':
-    Pool = multiprocessing.get_context('fork').Pool
+    mp_context = multiprocessing.get_context('fork')
```
Ericgig, line 17: For mac, we used `fork` since `spawn` would fail pickling, but it may work with `concurrent`. We should try it eventually.

qutip/tests/solver/test_parallel.py
```diff
@@ -68,4 +68,4 @@ def test_map_accumulator(map, num_cpus):
     y1 = [_func1(xx) for xx in x]
 
     map(_func2, x, args, kwargs, reduce_func=y2.append, map_kw=map_kw)
-    assert ((np.array(y1) == np.array(y2)).all())
+    assert ((np.array(sorted(y1)) == np.array(sorted(y2))).all())
```
Ericgig, line 71: Is this still needed? Isn't this redundant with the `future._i`?

Ericgig (approved):
Seems to be working fine.
Timeout work as expected and `KeyboardInterrupt` stop the jobs.
It continues when tasks raise exceptions, returning the results with `None`s mixed in.

It's good to know that correlation is only slow but works.

qutip/solver/parallel.py
hodgestar: I *think* spawn would actually work (or at least it works for our Windows tests). Either `spawn` or `forkserver` do require more complex data sharing since things are no longer in the same process. I think moving data around is what kills performance (Linux correlation tests timeout with `forkserver` too).

qutip/tests/solver/test_parallel.py
hodgestar: It's needed when supplying a `reduce_func` to `parallel_map`. We sort the results list, but the `reduce_func` is called in whatever order the tasks complete.

PR #1952, opened 2022-07-13
Add missing transform to Floquet-Markov Solver in v5

christian512:
**Description**
I have added a missing transformation call in the Floquet-Markov Master Equation Solver. The function returned the results in the wrong basis and calculated expectation values in the wrong basis. With the transformation the correct basis should be chosen.

### DIFF
--- a/doc/changes/1952.bugfix
+++ b/doc/changes/1952.bugfix
@@ -1,0 +1,1 @@
+Add missing state transformation to floquet_markov_mesolve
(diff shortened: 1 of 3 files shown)


hodgestar:
The code change look good, but could we also add a test for this so it doesn't fail again in the future? Maybe there is a simple example from a notebook that can be borrowed.

hodgestar (approved):
Thanks, @christian512!

PR #1955, opened 2022-07-15
Deprecate Solveroptions

Ericgig:
**Description**
`Options` and `SolverOptions` are objects that list all available options and pass them a solver. However in v5, with the choice of integrators, the supported options are not fixed. Most integration methods support `"atol"`, but `order` or `beta`, etc. are specific to some methods. SInce new integrators could be added by project such as qutip-cupy, `Options` can no longer be the place to list all supported options. It become little more that a dict. So I propose deprecating it.

The documentation of `sesolve`, `mesolve`, etc. now list the options used by the solver and most common ode options and link the the integrator documentation.

The class interface has the options property which document the options used by the solver and the integrators have the same. In the online documentation, the integrator api entries only contain the description of the method and the supported options. Since users don't need to call them directly, other methods are hidden and will be included in the dev documentation only.

Usage will look like this, for functions:
```
sesolve(H, psi, tlist, options={"store_states": True, "method": "lsoda", "atol": 1e-5})
```
and for the class interface:
```
solver = SeSolver(H, options={"store_states": True, "method": "lsoda", "atol": 1e-5})
solver.run(...)
solver.options["atol"] = 1e-9
solver.run(...)  # run with the new tol.
solver.options = {"atol": 1e-2}  # only change the affected items. 
solver.run(...) # Also run with the new tol.
solver.options["method"] = "adams"  # reset all ode options, including "atol" supported by both.
solver.options = {"method": "bdf", "atol": 1e-2}  # reset all ode options then apply the new atol 
```

Inside solvers, options are stored in a `_SolverOptions` which is a dict with frozen keys and feedback (chaging atol will reset the integrator).
 
Some `Intergrator` are slip so one method correspond to one integrator. Otherwise, `method` was an options used by both the solver to know which integrator to use and by some integrators.

`Options` and `SolverOptions` still both exist for backward compatibility: they raise a warning but return the dict of the same options so the change should not immediately break any code.

**Related issues or PRs**
Alternative to #1947


qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -34,11 +34,18 @@
 )
 from qutip.solver import (
     IntegratorException,
-    SolverOptions,
-    SolverResultsOptions,
 )
 
 
+def fill_options(_=None, **kwargs):
```
hodgestar, line 40: What is the `_=None` for? I'm also not sure why the `fill_options` method is here? Can't we just leave the options empty or supply the dict directly in most of the tests with less code / typing?

hodgestar (changes requested):
I left one question about the `fill_options` in the HEOM tests, but otherwise looks good to me. Is there anything in particular you'd like me to look at more closely? E.g. Anything that might have changed post SciPy discussions?

qutip/tests/solver/heom/test_bofin_solvers.py
Ericgig: When testing `Result`, the option cannot be empty since the result does not check if the options it receive is missing items. Normally the solver would ensure that they are present, but when using `Result` directly for test, we must do it manually.

I don't remember why the `_=None` is there, removing.

hodgestar (approved):

qutip/tests/solver/heom/test_bofin_solvers.py
hodgestar: Ah, I see!

PR #1957, opened 2022-07-16
Add hypothesis tests for data operators.

hodgestar:
**Description**
Add hypothesis strategies for data objects and some simple property-based tests for data operations.

**Related issues or PRs**
- None

### DIFF
--- a/qutip/tests/core/data/test_operators_hypothesis.py
+++ b/qutip/tests/core/data/test_operators_hypothesis.py
@@ -1,0 +1,83 @@
+import numpy
+
+from hypothesis import given, strategies as st
+
+from qutip.core import data as _data
+from qutip.tests import strategies as qst
+
+
+@given(qst.qobj_np(), qst.qobj_dtypes())
+def test_data_create(np_array, dtype):
+    if len(np_array.shape) == 1:
+        np_array = numpy.atleast_2d(np_array).transpose()
+    else:
+        np_array = numpy.atleast_2d(np_array)
+    data = _data.to(dtype, _data.create(np_array))
+    numpy.testing.assert_allclose(
+        data.to_array(), numpy.atleast_2d(np_array), rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas())
+def test_data_neg_operator(data):
+    neg = -data
+    numpy.testing.assert_allclose(
+        neg.to_array(), -data.to_array(), rtol=1e-15
+    )
+
+
+same_shape = st.shared(qst.qobj_shapes())
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_add_operator(a, b):
+    result = a + b
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() + b.to_array(), rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_minus_operator(a, b):
+    result = a - b
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() - b.to_array(), rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_matmul_operator(a, b):
+    result = a @ b
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() @ b.to_array(), rtol=1e-15
+    )
+
+
+@given(st.complex_numbers(), qst.qobj_datas(shape=same_shape))
+def test_data_scalar_multiplication_left_operator(x, a):
+    result = x * a
+    numpy.testing.assert_allclose(
+        result.to_array(), x * a.to_array(), rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas(shape=same_shape), st.complex_numbers())
+def test_data_scalar_multiplication_right_operator(a, x):
+    result = a * x
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() * x, rtol=1e-15
+    )
+
+@given(qst.qobj_datas(shape=same_shape), st.complex_numbers())
+def test_data_scalar_division_operator(a, x):
+    result = a / x
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() / x, rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_equality_operator(a, b):
+    result = (a == b)
+    assert result == numpy.allclose(
+        a.to_array(), b.to_array(), rtol=1e-15
+    )
--- a/qutip/tests/strategies.py
+++ b/qutip/tests/strategies.py
@@ -1,0 +1,51 @@
+""" Hypothesis strategies for QuTiP. """
+
+import numpy
+
+from hypothesis import strategies as st
+from hypothesis.extra import numpy as npst
+
+from qutip.core import data as _data
+
+
+def qobj_dtypes():
+    """
+    A strategy for Qobj data-layer dtypes.
+    """
+    dtype_list = sorted(_data.to.dtypes, key=lambda t: t.__name__)
+    return st.sampled_from(dtype_list)
+
+
+def qobj_shapes():
+    """
+    A strategy for Qobj data-layer shapes.
+    """
+    return npst.array_shapes(max_dims=2)
+
+
+def qobj_np(shape=qobj_shapes()):
+    """
+    A strategy for returning Qobj compatible numpy arrays.
+    """
+    return npst.arrays(shape=shape, dtype=numpy.complex128)
+
+
+@st.composite
+def qobj_datas(draw, shape=qobj_shapes(), dtype=qobj_dtypes()):
+    """
+    A strategy for returning Qobj data-layer instances.
+
+    Parameters
+    ----------
+    shape : strategy
+        A strategy to produce the array shapes. Defaults to `qobj_shapes()`.
+    dtype : strategy
+        A strategy to produce the array QuTiP dtypes. Defaults to
+        `qobj_dtypes()`.
+    """
+    # TODO: In future it might be good to have flags like unitary,
+    #       hermitian, ket, dm, oper, etc to restrict the kinds of
+    #       objects produced.
+    dtype = draw(dtype)
+    data = draw(qobj_np(shape=shape))
+    return _data.to(dtype, _data.create(data))
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,5 @@ benchmark/benchmark_data.js
 *-tasks.txt
 *compiled_coeff*
 result_images/
+
+.hypothesis/
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,7 @@ semidefinite =
 tests =
     pytest>=5.2
     pytest-rerunfailures
+    hypothesis
 ipython =
     ipython
 ; This uses ConfigParser's string interpolation to include all the above


hodgestar:
@Zac-HD Any suggestions for ways we could use hypothesis better? It all looks quite clean right now.

Is there a way we can change what hypothesis prints when an example is found? E.g. we don't want the data layer object repr to include the contents of the array, but it would be nice to see that in the test failure output.

qutip/tests/core/data/test_operators_hypothesis.py
```diff
@@ -0,0 +1,83 @@
...
+        result.to_array(), a.to_array() / x, rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_equality_operator(a, b):
+    result = (a == b)
+    assert result == numpy.allclose(
+        a.to_array(), b.to_array(), rtol=1e-15
+    )
```
Zac-HD, line 83: This looks like it should use exact equality rather than allclose?  I'd also add a test that these objects are always equal to themselves.

qutip/tests/core/data/test_operators_hypothesis.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+same_shape = st.shared(qst.qobj_shapes())
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_add_operator(a, b):
+    result = a + b
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() + b.to_array(), rtol=1e-15
+    )
```
Zac-HD, line 36: You *could* use a `@pytest.mark.parametrize("op", [operator.add, ...])` here, but it's really a matter of taste.

Zac-HD (commented):
All looks good to me!

> Is there a way we can change what hypothesis prints when an example is found?

You can't change the failing example (except by changing the repr or hacking our internal pretty-printer - please don't!), but you *can* add a call to `hypothesis.note("whatever representation you want to see")`  inside the test function, and you'll see that on failure.

qutip/tests/core/data/test_operators_hypothesis.py
Ericgig: The `Data.__eq__` test up to the numerical tolerance in core's settings which is `1e12` per default. So this allclose is stricter than the operation.

qutip/tests/core/data/test_operators_hypothesis.py
```diff
@@ -0,0 +1,83 @@
...
+        result.to_array(), a.to_array() - b.to_array(), rtol=1e-15
+    )
+
+
+@given(qst.qobj_datas(shape=same_shape), qst.qobj_datas(shape=same_shape))
+def test_data_matmul_operator(a, b):
+    result = a @ b
+    numpy.testing.assert_allclose(
+        result.to_array(), a.to_array() @ b.to_array(), rtol=1e-15
+    )
```
AGaliciaMartinez, line 52: For matrix multiplication the following restriction would be required `a.shape[1]==b.shape[0]`. Requiring same shape will lead to errors for non square matrices. I wonder if there is a quick way of generating such strategy similar to how same_shape works.

AGaliciaMartinez (commented):
The hypothesis testing approach seems to generate very clean and readable code! I really like this approach for testing as seems to simplify writing tests. I see that some tests do not fully work yet but the errors seem to be mostly related to inf and nan values not being properly handled. We can skip those tests by using the elements argument in `npst.arrays` but im not sure if we should. Also, absolute tolerances may need to be included in tests if we test with values close to zero.

@hodgestar What is the goal for this PR? Is this meant to ultimately replace the current testing code employed for the data layer? or is it just mean to test Data's operators?

qutip/tests/core/data/test_operators_hypothesis.py
Zac-HD: Check out [the `mutually_broadcastable_shapes()` strategy](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.mutually_broadcastable_shapes); you can see a quick example from `$ hypothesis write numpy.matmul` in your terminal 🙂

Zac-HD:
(unsubscribing because I think there's not much for me to add, but please @ me if that changes)

qutip/tests/core/data/test_operators_hypothesis.py
hodgestar: I quite like the separate tests.

qutip/tests/core/data/test_operators_hypothesis.py
hodgestar: I ended up writing my own simpler `MatrixShapesStrategy` because I couldn't get `MutuallyBroadcastableShapes` to not generate any extra loop dimensions.

I also added a helper `qobj_shared_shapes` that allows `shape_a, shape_b = qobj_shared_shapes([("m", "k"), ("k", "n")])` so that for tests one can write `@given(shape_a, shape_b)` and get the compatible shapes.

hodgestar:
@Ericgig @AGaliciaMartinez This is ready for another round of review.

I fixed small bugs in `mul` and `iszero` -- we should discuss whether the fixes are good or not.

There are still some failing examples I'm encountering locally, that we should also decide what to do with:

**scalar division**:
```
# result = a / x
# numpy returns array([[nan+nanj]])
# our result:
result [Data, Dense]: array([[8.98846567e+307+8.98846567e+307j]])
a [Data, Dense]: array([[9.97920155e+291+1.79769313e+308j]])
x: (1+1j)
```

**matmul**:
The current failing example I'm seeing locally is just an `atol` issue encountered for large exponents, but I have encounter other failing examples still that I need to dig into.

hodgestar:
@Zac-HD Would you mind doing a quick sanity check of my `MatrixShapesStrategy` class (https://github.com/qutip/qutip/blob/2aee70cb3981e60e236de831c56505a7caabd0aa/qutip/tests/strategies.py#L15-L62)? Mostly I want to know whether I should have been able to do this with `mutually_broadcastable_shapes`.

I also wouldn't mind a quick check of `qobj_shared_shapes` (https://github.com/qutip/qutip/blob/2aee70cb3981e60e236de831c56505a7caabd0aa/qutip/tests/strategies.py#L113-L131). Is this a good thing to be doing with shared strategies?

Zac-HD:
> Mostly I want to know whether I should have been able to do this with `mutually_broadcastable_shapes`.

Yes, you should use [`mutually_broadcastable_shapes(signature=)`](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.mutually_broadcastable_shapes) for that.  The signature syntax is a little different, since Numpy does it with strings, but we already support named dimensions, constant-size dimensions, and optional dims.  Should be a bit more efficient too.

> I also wouldn't mind a quick check of `qobj_shared_shapes`. Is this a good thing to be doing with shared strategies?

It does technically work?  But in such a situation I'd probably just use `st.data()` and draw in the body of the test; see `hypothesis write numpy.matmul` for an example.  If you expected to use it in many tests maybe a custom strategy, but for just `test_data_matmul_operator` that doesn't seem worth it.

hodgestar:
> Yes, you should use [`mutually_broadcastable_shapes(signature=)`](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.mutually_broadcastable_shapes) for that. The signature syntax is a little different, since Numpy does it with strings, but we already support named dimensions, constant-size dimensions, and optional dims. Should be a bit more efficient too.

I tried with `signature=` quite a bit but `(m, k), (k, n)` always produced thing like `(1, 1, 1), (1, 1)` despite setting `max_dims` and `min_dims` to various things.
 
> It does technically work? But in such a situation I'd probably just use `st.data()` and draw in the body of the test; see `hypothesis write numpy.matmul` for an example. If you expected to use it in many tests maybe a custom strategy, but for just `test_data_matmul_operator` that doesn't seem worth it.

I expected there to be other use cases.  I started with the output of `write numpy.matmul` but ended up playing around with my own strategy (see difficulties with mutually_broadcastable_shapes` above) and liking the idea of not have to unpack `a` and `b` inside the test, and of being able to supply the shapes for `a` and `b` to other strategies (e.g. maybe we will have a strategy for Qobj and QobjEvo objects and want to try various combinations of multiplying things together)

Zac-HD:
> I tried with `signature=` quite a bit but `(m, k), (k, n)` always produced thing like `(1, 1, 1), (1, 1)` despite setting `max_dims` and `min_dims` to various things.

Huh, that's surprising.  Share some example code and I'll have a look?

> I expected there to be other use cases. I started with the output of `write numpy.matmul` but ended up playing around with my own strategy (see difficulties with mutually_broadcastable_shapes`above) and liking the idea of not have to unpack`a`and`b`inside the test, and of being able to supply the shapes for`a`and`b` to other strategies (e.g. maybe we will have a strategy for Qobj and QobjEvo objects and want to try various combinations of multiplying things together)

👍 sounds good, carry on then!

hodgestar:
@Ericgig and I had a brief discussion about how to handle `inf` and `nan` generally. One option is forbid them entirely from the data layer -- i.e. to check and raise an error whenever a new data layer is created. A potential downside to this is we'd have to check often to be sure, since any operation can potentially create `nan` and `inf`, even simple addition:
```python
>>> 1e308 + 1e308
inf
>>> (1e308 + 1e308) - (1e308 + 1e308)
nan
```

Zac-HD:
(out of my wheelhouse again; @ me again if that changes (again!))

Ericgig (commented):
Hypothesis still feel somewhat magic, but I am starting to understand it.
Right now it seems to create data with mostly full arrays. I don't know how hard it is to create a strategy for different densities and to add the flags for Hermitian, etc. but it will be needed eventually.

The tolerances used are too tight, making some tests fail.


Also, from the old scipy test:
```
>>> a = np.array([[7.59109827e+14+9.96089747e+14j]])
>>> b = np.array([[7.22335902e+307+1.64618697e+308j]])
>>> np.allclose(b, a), np.allclose(a, b)
(False, True)
```
I am not sure we should ignore the overflow warning...

qutip/tests/strategies.py
```diff
@@ -0,0 +1,199 @@
...
+        min_side=1,
+        max_side=64,
+    ):
+        super().__init__()
+        self.side_strat = st.integers(min_side, max_side)
+        self.shapes = tuple(shapes)
+        self.min_side = min_side
+        self.max_side = max_side
+
+    def do_draw(self, data):
```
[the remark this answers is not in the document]
hodgestar: The `data` is the thing that manages drawing samples from strategies (via `data.draw(strategy)`). I don't know what it looks like under the hood, but I imagine it abstractly as something like a tree of counters that help draw the right examples for the sub-strategies within more complex strategies like this one.

hodgestar:
@Ericgig I've expanded the tests and reworked this so that it's less of a WIP. I've also documented the expected behaviour of the data layer in the presence of `nans`, etc in the description of the PR. I'll find somewhere in the docs to add that before merging the PR.

Would you mind taking another look and letting me know what you think?

qutip/core/data/mul.pyx
```diff
@@ -15,15 +17,24 @@ cpdef CSR imul_csr(CSR matrix, double complex value):
...
+            f"Multiplying CSR matrix by non-finite value {value!r}"
+            " would generate a dense matrix of NaNs."
+        )
     zscal(&l, &value, matrix.data, &ONE)
     return matrix
 
 cpdef CSR mul_csr(CSR matrix, double complex value):
     """Multiply this CSR `matrix` by a complex scalar `value`."""
-    if value == 0:
-        return csr.zeros(matrix.shape[0], matrix.shape[1])
```
Ericgig, line 24: I hate the removal of this shortcut.
`mul(data, 0)` can happen often in QobjEvo when the coefficient is a pulse or step function.
Without this shortcut, they need to loop over the values twice, once for this `mul`, once for the `add` or `matmul` following.
Thus this shortcut will actually help users.
It is also one of the only places where there is a `mul(data, 0)` in the code.

In some theoretical case, it can shallow a NaN, but I don't see any realistic case where this could hide an error.

So everyone loses for a case that will probably never happen...

Can you write a code example where this would save us? Something that looks like it should work, but cause infinities or NaN that would be erased by a multiplication by 0 without returning garbage.

If we can't even think of such a case, let's keep the shortcut.
If you can find one, let's loop over the values to find Nans and return the zeros crs if none are found so we at least loop only once over the values.

hodgestar: Maybe we just need to make `_data.zeros[type(x)](*x.shape)` easier to use so that we don't need to write `mul(data, 0)`? I can add `_data.zeros_like(x)` if that would be better?

In the specific case where `mul(data, 0)` is intended to create a matrix of zeros of the right shape, the current fast path can never be wrong since that is exactly what it does.  In general, `mul(data, 0)` is currently just incorrect floating point arithmetic.

I was wondering about adding `has_non_finite` flag to CSRs. It will be a bit of pain to keep track of everywhere but it would allow us to fast path the all finite cases without sacrificing correctness.

If this is a blocker for this PR, I'm happy to undo the current change and just make the tests pass by just explicitly stating that the expected result is `0` in those cases in the test.

I also don't want to sacrifice performance, but it is important to have a clear set of tests that show how we expect the data layer to behave in these edge cases.

Ericgig: I am more thinking about the use in `QobjEvo`. I don't think we ever use it as `_data.zeros_like`, in the code base, but that function would be useful, as would be an `identity_like` as we discussed somewhere.

I would be fine with a `has_non_finite` flag, but I think we should raise an error anytime we would raise that flag.
I think that we could do the check when we tidyup: we already check the scale of all values and it's done in most operation that could create zeros, thus functions at risk.

I see a CSR matrix full of zeros as wrong, if not an error, so this is a blocker to me. 
This and tests should pass reliably.

Ericgig: I am using `mul` as `_data.zeros_like` in stochastic...

hodgestar:
Closing -- I don't have bandwidth for this currently. Let's create a new MR when we have capacity again.

PR #1961, opened 2022-07-27
Floquet rates integration

christian512:
**Checklist**
- [ ] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#Changelog%20Generation) for more information).

**Description**
The numerical integration for the `X` matrix uses the rectangular rule with a fixed number of time steps `nT = 100`. I changed `nT` to vary with the given `kmax` since the integrands frequency increases with `kmax`.

### DIFF
--- a/qutip/floquet.py
+++ b/qutip/floquet.py
@@ -612,7 +612,7 @@ def floquet_master_equation_rates(f_modes_0, f_energies, c_op, H, T,
     Gamma = np.zeros((N, N, M))
     A = np.zeros((N, N))
 
-    nT = 100
+    nT = 2 * kmax * np.ceil(omega)
     dT = T / nT
     tlist = np.arange(dT, T + dT / 2, dT)
 
--- a/qutip/floquet.py
+++ b/qutip/floquet.py
@@ -612,7 +612,8 @@ def floquet_master_equation_rates(f_modes_0, f_energies, c_op, H, T,
     Gamma = np.zeros((N, N, M))
     A = np.zeros((N, N))
 
-    nT = 2 * kmax * np.ceil(omega)
+    # time steps for integration of coupling operator
+    nT = np.max([2 * kmax * np.ceil(omega), 100])
     dT = T / nT
     tlist = np.arange(dT, T + dT / 2, dT)
 


hodgestar:
Woot for finding this buglet. I have some questions:

- What happens in v5?
- Can we also add a test for this case where a different nT is needed?
- Should we make nT a parameter which defaults to your new value? How closely do the signatures of the various Floquet solvers match currently? If they already have different options we can likely just add a new parameter. If not, perhaps we can anyway, but we should think a bit more.
- We should update the docstring to match.

christian512:
Regarding your questions:

- It's the same in QuTiP 5 currently. I am creating a PR also for that version, but I will include support for multiple coupling operators to the PR.
- Yes, I will add a test for this.
- This parameter is only relevant for `fmmesolve` ( master equation solver ). There is already the parameter `kmax` which sets the truncation of sidebands to be considered for the integration of the coupling operator. The integral to solve reads as:

$$ \frac{1}{T} \int_0^T dt e^{-ik\Omega t} f(t) $$

where $k \in \[-k_{max}, k_{max} \]$ and $f(t)$ is a $T$-periodic function. Hence, we know the frequency of the integrand and therefore the required number of grid points for the numerical integration can be set based on `kmax`.
I think, we don't need to create another parameter for `nT`.

qutip/floquet.py
```diff
@@ -612,7 +612,8 @@ def floquet_master_equation_rates(f_modes_0, f_energies, c_op, H, T,
     Gamma = np.zeros((N, N, M))
     A = np.zeros((N, N))
 
-    nT = 100
+    # time steps for integration of coupling operator
+    nT = int(np.max([2 * kmax * np.ceil(omega), 100]))
```
hodgestar, line 616: This formula for `nT` looks odd to me. If I plug it into `dT = T / nT` I get `dT ~ T**2`.

`omega` shouldn't affect `nT` because `omega * t` always traces out one period over the interval of the integral, i.e. `t = [0, T]`. So the full number of periods of `e**(1j * k * omega * t)` is just `k`.

This seems to suggest that we should use `nT = max(20 * kmax, 100)` or similar to give a fixed number of steps per period.

doc/changes/1961.bugfix
```diff
@@ -0,0 +1 @@
+adaptive time steps for integration of floquet rates
```
hodgestar, line 1:
```suggestion
Change floquet_master_equation_rates(...) to use an adaptive number of time steps scaled by the number of sidebands, kmax.
```

hodgestar: Could you also fix the docstring for `kmax` (it's currently named `k_max`) and add a note to it to mention how the number of integration steps is scaled?

hodgestar:
@christian512 Thank you for adding the test and explaining the logic. I did a second mini-review.

qutip/floquet.py
christian512: Thanks for the hint. I did not see that the number of periods of the exponential is fixed by `k` only.  I'll change the definition of `nT`

hodgestar:
I see codeclimate failed because one line of a docstring is too long. Could you fix?

hodgestar (approved):
Thanks for finding and fixing this, @christian512! Looks great now.

hodgestar:
The test failure was just the occasional bloch sphere test failure that happens on Windows sometimes.

hodgestar:
Change floquet_master_equation_rates(...) to use an adaptive number of time steps scaled by the number of sidebands, kmax.

hodgestar:
Change floquet_master_equation_rates(...) to use an adaptive number of time steps scaled by the number of sidebands, kmax.

Issue #925, opened 2018-10-12
`qutip.metrics.fidelity` needs fix for pure states represented by bra or ket


eunjongkim:
I found an instance where putting two kets in `fidelity` function gives nonphysical result while inputting corresponding density matrices gives the right result:
![image](https://user-images.githubusercontent.com/8111538/46898454-a5d08300-ce3e-11e8-920c-1cc98d45838a.png)

Currently, I have no idea what's going wrong, but my observation is that this starts to happen when Hilbert space is getting large (for this simulation, I use ket vector in a Hilbert space of 12 two-level systems). I can send you the data with e-mail.

nathanshammah:
Could you send me the code?

lucainnocenti:
I also experienced problems with `fidelity`. Consider the following MWE:

```
import qutip
import numpy as np

final = np.asarray([0.867314655330313 - 0.4576338188944636j, 0.17267952805615244 - 0.09111291375544905j, 0.013020232894921149 -0.006875138252288391j])
final_state = qutip.Qobj(final)
target = np.asarray([0.9806443568092577, 0.19524328915024022, 0.014720852555238875])
target_state = qutip.Qobj(target)

numpy_fidelity = np.abs(np.vdot(final_state.full(), target_state.full())**2)
qutip_good_fidelity = np.abs((final_state.dag() * target_state)[0, 0])**2
qutip_bad_fidelity = qutip.fidelity(final_state, target_state)**2
print('Correct: {}'.format(numpy_fidelity))
print('Still correct: {}'.format(qutip_good_fidelity))
print('Wrong: {}'.format(qutip_bad_fidelity))
```

The problem is clearly in the fact that you compute the fidelity of ket states by converting them into density matrices and using the general formula ([as seen here](https://github.com/qutip/qutip/blob/564c987bdca385ae179730d0aad7b0d1e624e14d/qutip/metrics.py#L112)). This is fine in general, but can be very inefficient for large states, and is less numerically stable.

An easy solution is to just compute the fidelity as the inner product when both states are kets.

harman97365:
Ya Qutip needs to fix the issue, the formula suggested by @lucainnocenti is better.

Ericgig:
Solved in #1964

Issue #361, opened 2015-08-26
Fidelity greater than one


dick-t:
fidelity() returns >1 by a significant amount for certain pure states. States are normalised, and taking the inner product returns 1. Code below provides two examples of states which give fidelity >1 with themselves.

```
#to test qutip's fidelity routine
import qutip as qu

psi0=qu.Qobj()

i=0
while i<=1:
    j=0
    while j<=1:
        psi0+=qu.state_number_qobj([2,2],[i,j]) #even superposition of qubit states
        j+=1
    i+=1

psi0=psi0.unit() #normalise

print(qu.fidelity(psi0,psi0))
print((psi0.dag()*psi0).norm()**2)
print("\n")

psi0=qu.tensor(psi0,qu.basis(10,1)) #tensor product with Fock state

print(qu.fidelity(psi0,psi0))
print((psi0.dag()*psi0).norm()**2)

```

Output:

```
1.00000002107
1.0


1.00000003485
1.0
```

cgranade:
I suspect this is because of ``A.sqrtm()``, which is not necessary for a pure state. I'll go on and try to see what I can do with it.

nonhermitian:
There is numerical error in the answer due to the need for the sqrtm method. I am not sure if I would call this error significant as the error is of order 1e-7, while the fidelity is of order unity. However, as Chris said, there might be a better way for pure states.

Paul

cgranade:
I went on and gave a stab at it (PR #362), as it turns out we don't need ``sqrtm(sqrtm(A) * B * sqrtm(A))``  even for mixed ``A`` and ``B``--- we only need the trace of the outer ``sqrtm``, which can be done by looking at the eigenenergies directly. After informal checking, this resulted in more precise answers when F = 1. I also added more checks to make sure that the fidelity is bounded by [0, 1] to within 1e-7.

ajgpitch:
@cgranade did you test your PR #362 against dick-t example? If so what were the results?
It would seem that dick-t is comparing a method that works with states in general (mixed or pure) with one that assumes states are pure. Do we have some method in qutip that gives the fidelity when states are assumed pure? It would seem more efficient (and less numerical error risk) to compute the fidelity without the sqrtm (as dick-t did in the comparison) if we knew the states were pure.

cgranade:
[...]

PR #1964, opened 2022-07-29
Fix fidelity for pure states

PierreGuilmin:
**Description**
The previous code converted both states to density matrices, which is both inefficient and numerically inaccurate. We propose to use the reduced fidelity formula for pure states, given by the modulus of their inner product.

**Related issues or PRs**
Related to #361 and #925.

Reusing @lucainnocenti example (https://github.com/qutip/qutip/issues/925#issuecomment-542318121):
```python
import qutip as qt
import numpy as np

A = qt.Qobj(np.array([
    0.867314655330313 - 0.4576338188944636j,
    0.17267952805615244 - 0.09111291375544905j,
    0.013020232894921149 -0.006875138252288391j
]))
B = qt.Qobj(np.array([
    0.9806443568092577, 0.19524328915024022, 0.014720852555238875
]))

fid1 = np.abs(np.vdot(A.full(), B.full()))
fid2 = np.abs((A.dag() * B)[0, 0])
fid3 = np.abs(A.overlap(B))
fid_qt = qt.fidelity(A, B)
print(f'Correct:\n{fid1}\n{fid2}\n{fid3}')
print(f'Incorrect:\n{fid_qt}')
```
```
Correct:
0.9999999999849929
0.9999999999849929
0.9999999999849929
Incorrect:
1.0000000028441378
```

Note that this PR does not fix the proposition from @lucainnocenti https://github.com/qutip/qutip/issues/925#issuecomment-542318121 to perform the general computation using `scipy.linalg.svdvals` at the cost of computing the square root of both matrices.

### DIFF
--- a/doc/changes/1964.bugfix
+++ b/doc/changes/1964.bugfix
@@ -1,0 +1,1 @@
+Fix fidelity for pure states.
--- a/doc/changes/1964.bugfix
+++ b/doc/changes/1964.bugfix
@@ -1,1 +1,1 @@
-Fix fidelity for pure states.
+Change fidelity(A, B) to use the reduced fidelity formula for pure states which is more numerically efficient and accurate.
--- a/qutip/metrics.py
+++ b/qutip/metrics.py
@@ -53,12 +53,13 @@ def fidelity(A, B):
     >>> np.testing.assert_almost_equal(fidelity(x,y), 0.24104350624628332)
     """
     if A.isket or A.isbra:
+        if B.isket or B.isbra:
+            # The fidelity for pure states reduces to the modulus of their
+            # inner product.
+            return np.abs(A.overlap(B))
         # Take advantage of the fact that the density operator for A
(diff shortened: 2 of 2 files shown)


Ericgig (changes requested):
Looks good, thank you.

The changes file name is off by one, could you rename it.

PierreGuilmin:
Sorry, fixed 🙇‍♂️

doc/changes/1964.bugfix
```diff
@@ -0,0 +1 @@
+Fix fidelity for pure states.
```
hodgestar, line 1:
```suggestion
Change fidelity(A, B) to use the reduced fidelity formula for pure states which is more numerically efficient and accurate.
```

hodgestar (approved):

hodgestar:
Thanks @PierreGuilmin!

I made the changelog message a bit more detailed and will merge now.

Issue #1965, opened 2022-08-02
Strange behaviour of brmesolve on Apple M1


jsmarsha11:
### Bug Description

The output is not consistent with what we get on an Intel chip. It either fails, does complete but has invalid output data (times and states data not the same sizes), or gives different simulation results entirely.

### Code to Reproduce the Bug

```shell
from qutip import *
import numpy as np

def psd(w):
    return w * np.exp(-w)

out = brmesolve(sigmax(), psi0=basis(2,0), tlist=np.arange(0, 1, 0.01), a_ops=[[sigmax(), psd]])

assert len(out.states) == len(out.times), f'{len(out.states)} vs {len(out.times)}'
```

### Code Output

```shell
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Input In [11], in <cell line: 6>()
      2     return w * np.exp(-w)
      4 out = brmesolve(sigmax(), psi0=basis(2,0), tlist=np.arange(0, 1, 0.01), a_ops=[[sigmax(), psd]])
----> 6 assert len(out.states) == len(out.times), f'{len(out.states)} vs {len(out.times)}'

AssertionError: 10 vs 100
```

### Expected Behaviour

As we see, the above code does run and produces an output, but the states list is not the same as the times lists (these should be the same length as they have a 1-1 correspondence). It actually appears it has only given data for the first 10 time-steps in this case.
The same code works fine on Intel chip. 
We can increase the `nsteps` parameter to give the correct number of output states, but they do not agree with Intels output (also see below for related issue), and actually are not always even quantum states, so something in the solver went wrong.

Another example, using the cython string formatting:

```brmesolve([[sigmax(), 'cos(t)']], psi0=basis(2,0), tlist=np.arange(0, 1, 0.01), a_ops=[[sigmax(), '1']])``` works fine on Intel, but gives error ```Exception: ODE integration error: Try to increase the allowed number of substeps by increasing the nsteps parameter in the Options class.``` on M1. 

Moreover, if I do increase the `nsteps` sufficiently, this does appear to run correctly, but the output states are not the same as we get on Intel (and so I assume are incorrect).

### Your Environment

```shell
QuTiP Version:      4.7.0
Numpy Version:      1.21.6
Scipy Version:      1.7.3
Cython Version:     0.29.28
Matplotlib Version: 3.5.1
Python Version:     3.9.13
Number of CPUs:     10
BLAS Info:          Generic
OPENMP Installed:   False
[...]

PR #1966, opened 2022-08-04
`Brmesolve` properly raise error when ode fails.

Ericgig:
**Description**
For constant systems, when the ODE solver is not successful, `brmesolve` returned an incomplete result without telling the user something failed.
With this PR. an exception is raised in these situations.

**Related issues or PRs**
#1965

### DIFF
--- a/doc/changes/1965.bugfix
+++ b/doc/changes/1965.bugfix
@@ -1,0 +1,1 @@
+`brmesolve` raise an error when ode is not successful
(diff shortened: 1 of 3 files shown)


hodgestar (approved):
I might have chosen `RuntimeError` instead of `Exception`, but I'm happy with either.

The removal of end of line white space is fine, although I'm a bit worried it will make merging to v5 a bit difficult.

Happy for these to stay as-is though.

hodgestar (approved):

PR #1967, opened 2022-08-05
Update the roadmap and ideas to show completed work as of August 2022.

hodgestar:
**Description**
Update the roadmap and ideas to show completed work as of August 2022.

**Related issues or PRs**
None.

### DIFF
--- a/doc/changes/1967.doc
+++ b/doc/changes/1967.doc
@@ -1,0 +1,1 @@
+Update the roadmap and ideas to show completed work as of August 2022.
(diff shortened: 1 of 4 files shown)


AGaliciaMartinez (approved):
I went through the changes and they look good to me. They only question I have is whether we should include other finished GSoC projects.

It is nice to have updated information as this may be confusing otherwise. Thanks for taking care of this Simon!

hodgestar:
It's probably a good idea to add all the finished projects we remember, but let's do it in a follow up PR.

Issue #1963, opened 2022-07-29
parfor and brmesolve - Cythonizing problem in v5


christian512:
### Bug Description

When combining `parfor` with `brmesolve` and a time-dependent Hamiltonian, the compilation of the Hamiltonian seems to interfer with the correct functioning of `parfor`. The call results in an error, that the coefficient file is too short. 

I guess this could also happen when using `parfor` with other solvers.

### Code to Reproduce the Bug

```shell
from qutip import brmesolve, sigmaz, sigmax, basis, parfor
import numpy as np

psi0 = basis(2, 0)
H_S = np.pi * sigmaz()
H_I = sigmax()
tlist = np.linspace(0, 100, 101)
pulse_shape = "exp(- t**2)"

def brme(A):
    H_tmp = [A * H_S, [H_I, pulse_shape]]

    return brmesolve(
        H_tmp, psi0, tlist, e_ops=[sigmaz()]
    ).expect[0][-1]

results = parfor(brme, np.linspace(0, np.pi, 10))
```

### Code Output

```shell
ImportError: /home/chris/.qutip/qutip_coeffs_1.1/qtcoeff_4c859672674b8b603bd243ba3f4a8d.cpython-38-x86_64-linux-gnu.so: file too short
```

### Expected Behaviour

`parfor` should wait for the compilation to finish and then start the parallel processes.

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+c64f38e
Numpy Version:      1.23.0
Scipy Version:      1.8.1
Cython Version:     0.29.30
Matplotlib Version: 3.5.2
Python Version:     3.8.10
Number of CPUs:     4
```

Ericgig:
This won't be easy to fix, if the compilation is in the task, `parfor` cannot know about it to wait for it.
The simplest would be to have only the main process able to compile. This code will still raise an error, but the solution would be in the error message.
Otherwise communication between process would be needed and this can get tricky.

Also `parfor` will be removed. It says in its [documentation](https://qutip.org/docs/latest/apidoc/functions.html#qutip.parallel.parfor) to use `parallel_map` instead since v3.

hodgestar:
@christian512 Perhaps we should try replace parfor with parallel_map in the notebooks and see what happens?

hodgestar:
@Ericgig Should we just drop parfor in v5?

Ericgig:
Yes, there is no reason to keep both `parallel_map` and `parfor` in v5.

[...]

PR #1968, opened 2022-08-05
Protect compilation of coefficient using filelock

Ericgig:
**Description**
Add a lock for creating string coefficient. This allow to use the coefficient inside a parallel loop without issues.
The lock use `filelock` which works across processes, so it's safe to use with parallel coming from outside python.
The lock is per file, so different coefficient can compile in parallel. If there is a conflict, the first process will create and compile the coefficient and the other will use it (raising a hash collision error if not actually the same string.)

`filelock` is a new requirement for runtime compilation.

**Related issues or PRs**
#1963

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -15,6 +15,7 @@ from collections import defaultdict
 from setuptools import setup, Extension
 try:
     from Cython.Build import cythonize
+    import filelock
 except ImportError:
     pass
 
@@ -217,6 +218,7 @@ class CompilationOptions(QutipOptions):
 
     try:
         import cython
+        import filelock
         _use_cython = True
     except ImportError:
         _use_cython = False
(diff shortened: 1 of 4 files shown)


hodgestar (approved):
Woot!

hodgestar:
The Windows test failure is the strange "Tcl is not properly installed one". Someone else reported this at https://github.com/microsoft/azure-pipelines-tasks/issues/16426 so I mentioned that we see it too. I linked to this build failure there, so please don't re-run it.

hodgestar:
Happy for this to be merged with the build failure and we can just check that the merge commit builds correctly.

Ericgig:
@hodgestar 
Sorry, I reran the test before reading your comment, here is the failing attempt:
https://github.com/qutip/qutip/actions/runs/2805431012/attempts/1

PR #1970, opened 2022-08-06
Backport fix for IPython helper _repr_svg_ from dev.major.

hodgestar:
**Description**
Backport the fix for #1912 to master.

The issue was that previously the print_figure function returned byte code, but since https://github.com/ipython/ipython/pull/5452 (in 2014) it returns a Unicode string.

**Related issues or PRs**
- #1912 (original issue)
- #1918 (fix for dev.major)
- #1943 (previous attempt at a fix for 4.7.X)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -101,7 +101,7 @@ jobs:
         # rather than in the GitHub Actions file directly, because bash gives us
         # a proper programming language to use.
         run: |
-          QUTIP_TARGET="tests,graphics,semidefinite"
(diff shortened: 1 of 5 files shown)


hodgestar:
@Ericgig After reading the following comment in `setup.cfg`
```
; This uses ConfigParser's string interpolation to include all the above
; dependencies into one single target, convenient for testing full builds.
```
I changed my mind a bit about the role of `full`. It seems that the intention of full is to install everything for testing and development, so it makes sense for the ipython dependency to be included there.

After I changed my mind, I just cherry-picked Asier's merge commit for #1918 to master.

Ericgig (approved):

PR #1975, opened 2022-08-16
Have coefficient compilation use relative path

Ericgig:
**Description**
A change in setuptools (pypa/setuptools#3521) raise a warning when compiling coefficient using absolute path, which breaks our tests for `dev.major`:
```
DeprecationWarning: Absolute path '/home/runner/.qutip/qutip_coeffs_1.1/qtcoeff_8208be2129365a87dda9c12028c457' is being replaced with a relative path 'home/runner/.qutip/qutip_coeffs_1.1/qtcoeff_8208be2129365a87dda9c12028c457' for outputs. This behavior is deprecated. If this behavior is desired, please comment in pypa/distutils#169.
```
This changes the path to be relative.

I change the path before creating the lock files, which otherwise were polluting the active directory.

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -482,10 +482,12 @@ cdef class StrCoefficient(Coefficient):
 
 def compile_code(code, file_name, parsed, c_opt):
     pwd = os.getcwd()
-    root = qset.coeffroot
+    os.chdir(qset.coeffroot)
+    # Files with the same name, but differents extension than the pyx file, are
+    # erased during cythonization process, breaking filelock.
+    # Adding a prefix make them safe to use.
+    lock = filelock.FileLock("compile_lock_" + file_name + ".lock")
     try:
-        os.chdir(root)
-        lock = filelock.FileLock(file_name + ".lock")
         lock.acquire(timeout=0)
         for file in glob.glob(file_name + "*"):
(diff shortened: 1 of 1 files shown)


hodgestar (approved):
Thanks for finding and fixing this.

hodgestar:
I assume we need a similar fix for master?

Ericgig:
I don't think we need a fix for master, I haven't seen the test fail yet. 
In master, we create the file in the active directory and use pyx import. So the absolute path should never be used.

hodgestar:
The last build on master failed with:
```
  File "<string>", line 1, in <module>
  File "/home/runner/work/qutip/qutip/qutip/__init__.py", line 64, in <module>
    import qutip.cy.pyxbuilder as _pyxbuilder
  File "/home/runner/work/qutip/qutip/qutip/cy/__init__.py", line 1, in <module>
    from qutip.cy.spmatfuncs import *
ModuleNotFoundError: No module named 'qutip.cy.spmatfuncs'
```
While trying to run `qutip.about()`. See https://github.com/qutip/qutip/actions/runs/2843095517. I triggered rerunning the Python 3.9 default build to see it if was temporary.

hodgestar:
It looks like the Python 3.9 build managed to start running the tests, so hopefully we are good!

PR #1976, opened 2022-08-17
Calls to random object funtions to take `dimensions` instead of `N`.

### LINKED ISSUE #1838: Unify random `Qobj` creation functions support of `dims`.

Ericgig:
**Description**

Update the interface of random functions:
- The first input is `dimensions` instead of `N`. It act the same way as the input of `qeye`. (#1838)
```
qutip.rand_herm(N).dims == [[N], [N]]
qutip.rand_herm(dims).dims == [dims, dims]

qutip.rand_super(N).dims == [[[N], [N]],  [[N], [N]]]
qutip.rand_super([N, M]).dims == [[[N, M], [N, M]],  [[N, M], [N, M]]]
qutip.rand_super([[N], [N]]).dims == [[[N], [N]],  [[N], [N]]]
```
- Random Qobj creation use numpy new random interface. The seed now also accept `numpy.random.SeedSequence` or  `numpy.random.Generator`.
- We had `unitary = (1j*rand_herm(N, density)).expm()`, therefore the resulting density for `rand_unitary` (and `rand_dm`) used to be denser than what was entered. I changed the density keyword to correspond to the density of output object. I do this by creating a smaller dense matrices and shuffling them together.
- Random super operator function now accept a `superrep` keyword and return the object in that representation. This work for `super`, `chi` and `choi` representation, but not `kraus` since it's not a `Qobj` and this cause issue with the `dtype` keyword.
- Lastly I merged some similar functions:  `rand_unitary_haar` is now included as `rand_unitary(N, distribution="haar")`, etc. We had a few functions that returned the same type of object, but with different algorithm. I found this confusing, so I merged them to show they are equivalent. The algorithms having a reference to a paper are used as the defaults. I did not keep the previous functions, but we could keep them with a deprecation warning.

Lastly I merged the `test_rand` and `test_random` files. Tests in those files where overlapping.

**Related issues or PRs**
fix #1838


qutip/random_objects.py
```diff
@@ -1,60 +1,138 @@
...
     'rand_kraus_map',
-    'rand_super_bcsz', 'rand_super'
+    'rand_super',
+    "rand_super_bcsz",
 ]
 
 import numbers
 
 import numpy as np
+from numpy.random import Generator, SeedSequence, default_rng
```
hodgestar, line 19: Which version of numpy was the new random interface added in?

qutip/random_objects.py
```diff
@@ -65,115 +143,154 @@ def rand_jacobi_rotation(A, *, seed=None):
...
 
 
-def _randnz(shape, norm=np.sqrt(0.5), *, seed=None):
+def _get_block_sizes(N, density, generator):
     """
-    Returns an array of standard normal complex random variates.
-    The Ginibre ensemble corresponds to setting ``norm = 1`` [Mis12]_.
+    Obtain a list of matrices block sizes in a way that a NxN matrix composed.
+    Of full matrices of these size allong the diagonal would be of desired
+    density.
```
hodgestar, line 150:
```suggestion
    Obtain a list of matrix block sizes in such a way that an NxN matrix composed
    of full matrices of these sizes along the diagonal would be of desired density.
```

qutip/random_objects.py
```diff
@@ -65,115 +143,154 @@ def rand_jacobi_rotation(A, *, seed=None):
...
+            end = end + block.shape[0]
+            brow, bcol = np.meshgrid(idx[start:end], idx[start:end])
+            matrix[brow, bcol] = block
+    return _data.create(matrix, copy=False)
 
 
-def rand_herm(N, density=0.75, dims=None, pos_def=False,
-              *, seed=None, dtype=_data.CSR):
-    """Creates a random NxN sparse Hermitian quantum object.
+def rand_herm(dimensions, density=0.30, pos_def=False, eigenvalues=None, *,
```
hodgestar, line 212: Why was the default changed from 0.75 to 0.30?

qutip/random_objects.py
```diff
@@ -65,115 +143,154 @@ def rand_jacobi_rotation(A, *, seed=None):
...
-    else:
-        raise TypeError('Input N must be an integer or array_like.')
-    out = Qobj(M, dims=dims or [[N]]*2, type='oper',
-               isherm=True, copy=False).to(dtype)
-    if dtype:
-        out = out.to(dtype)
-    return out
+        out = Qobj(M, type='oper', isherm=True, copy=False)
+
+    out.dims = dims
```
hodgestar, line 280: Why not pass dims when the Qobj is created?

hodgestar (commented):
I left some questions, but looks good.

qutip/random_objects.py
Ericgig: 1.17, not that new.

qutip/random_objects.py
Ericgig: I forgot I changed that. Do you prefer the previous value?
We often use this for a random Hamiltonian, since they are usual usually quite sparse, this seems a more useful value.

qutip/random_objects.py
hodgestar: And we'll drop support for numpy 1.16 for v5, so that looks okay.

qutip/random_objects.py
hodgestar: I'm personally happy with a sparser value, but I'm wondering if it will affect users. Probably not, since it's already random in any case.

hodgestar (approved):

Ericgig:
Doing a last check, I saw that some function have `rand(dimensions, density, distribution, ...)` while other have `rand(dimensions, distribution, density, ...)`. Fixing it to have density first everywhere.

hodgestar (approved):

PR #1982, opened 2022-08-19
Update scipy requirements

eendebakpt:
**Description**

This PR updates the scipy requirements to version 1.4+. The 1.4 version was released December 2019.
Versions before 1.4 would not have worked with dev.major as an import of functools was missing.

With the updated requirements we can remove the version checks in qutip.solve.steadystate which speeds up import times a bit.


requirements.txt
```diff
@@ -1,4 +1,4 @@
 cython>=0.29.20
 numpy>=1.16.6
-scipy>=1.0
```
hodgestar, line 3: I discussed this with @Ericgig and remembered that for QuTiP v5 we are planning to support scipy 1.5+. Could you update the version here to 1.5 and then it's done?

eendebakpt: @hodgestar Done

hodgestar (approved):

hodgestar:
Thanks @eendebakpt!

PR #1984, opened 2022-08-25
Guide - Superoperators

christian512:
**Description**
Adding a guide on *Superoperators, Pauli Basis and Channel Contraction* which is moved from this notebook:
https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/superop-contract.ipynb

**Related issues or PRs**
The corresponding notebook was removed from the new `qutip-tutorials` repository in this PR https://github.com/qutip/qutip-tutorials/pull/34


hodgestar (changes requested):
Could you make the python commands actually run and generate the plots in the guide? I think the `.. plot::` directives you have don't do anything currently because they have no contents.

Using the `plot` directive is a bit of a pain -- one has to set up imports at the top of the guide, and then clean the context at the bottom, and specify if a particular plot directive doesn't contain any plots.

See guide/heom/bosonic.rst for an example usage. In the top plot cell you want as a minimum something like
```
.. plot::
    :context: reset
    :nofigs:

    from qutip import func1, func2, func3
```
and at the end
```
.. plot::
    :context: reset
    :include-source: false
    :nofigs:

    # reset the context at the end
```
```

hodgestar (approved):
Looks good to me. I will just make sure the GitHub Actions pass.

hodgestar (approved):

hodgestar:
@christian512 Sorry this took so long to merge!

PR #1986, opened 2022-08-26
Krylov as an integrator

Ericgig:
**Description**
Add `krylov` method as an `Integrator` for `SeSolver` porting `krylovsolve` to v5 format.

The state evolution code was rewritten to use the qutip's data object instead of numpy arrays, but is otherwise the same.

Since an integrator can be used without knowing the length of the evolution, some changes where as to when the krylov matrices are recomputed. Instead of choosing the step length so the error of the whole evolution is under `atol`, it is chosen so each steps stays under it and an options is available to compute it only once or at each step.

A default `krylov_dim` was needed for the integrator, so I set it the scale as the square root of the size of the system. This is not ideal, but seems to work fine. When using it through `krylovsolve`, `krylov_dim`is a required parameter.


hodgestar:
@Ericgig Could you merge in dev.major and resolve conflicts?

qutip/solver/ode/krylov.py
```diff
@@ -0,0 +1,237 @@
...
+        self._step = 0
+        krylov_dim = self.options["krylov_dim"]
+        if krylov_dim < 0 or krylov_dim > self.system.shape[0]:
+            raise ValueError("The options 'krylov_dim', must be a positive "
+                             "integer smaller that the system size")
+
+        if krylov_dim == 0:
+            # TODO: krylov_dim, max_step and error (atol) are related by
+            # err ~= exp(-krylov_dim / dt**(1~2))
+            # We could ask for 2 and determine the third one.
```
hodgestar, line 44: Where does the approximate expression for err come from? I would expect it to depend on the structure of the eigenvalues of the system, and not just the dimension.

qutip/solver/ode/krylov.py
```diff
@@ -0,0 +1,237 @@
...
+
+        nsteps : int, default=100
+            Max. number of internal steps/call.
+
+        min_step, max_step : float, default=(1e-5, 1e5)
+            Minimum and maximum step size.
+
+        krylov_dim: int, default=0
+            Dimension of Krylov approximation subspaces used for the time
+            evolution approximation.
```
hodgestar, line 218:
```suggestion
            Dimension of Krylov approximation subspaces used for the time
            evolution approximation. If the defaut 0 is given, the dimension is calculated
            from the system size N, using `min(int((N + 100)**0.5), N-1)`.
```

hodgestar (changes requested):
It's great that we can implement this as an integrator now. I left some suggestions, but it looks good overall.

qutip/solver/ode/krylov.py
```diff
@@ -0,0 +1,237 @@
...
+        while j < krylov_dim and T_subdiag[j] > self.options['sub_system_tol']:
+            j += 1
+            v.append(_data.mul(w, 1 / T_subdiag[j-1]))
+            w_prime = _data.matmul(H, v[-1])
+            T_diag[j] = _data.inner(w_prime, v[-1])
+            w = _data.add(w_prime, v[-1], -T_diag[j])
+            w = _data.add(w, v[-2], -T_subdiag[j-1])
+            T_subdiag[j] = _data.norm.l2(w)
+
+        krylov_tridiag = _data.diag["dense"](
```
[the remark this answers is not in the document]
Ericgig: The `N` is usually quite small and we need to compute it's eigen decomposition. The sparse eigen solver is a lot slower and less precise..

qutip/solver/ode/krylov.py
```diff
@@ -0,0 +1,237 @@
...
+        err = krylov_error(dt)
+        if err > 0:
+            ValueError(
+                f"With the krylov dim of {self.options['krylov_dim']}, the "
+                f"error with the minimum step {dt} is {err}, higher than the "
+                f"desired tolerance of {self.options['atol']}."
+            )
+
+        while krylov_error(dt * 10) < 0 and dt < self.options["max_step"]:
+            dt *= 10
```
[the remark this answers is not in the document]
Ericgig: The `root_scalar` call need to look for a `0` and converge better with the log.

qutip/solver/ode/krylov.py
Ericgig: I tried it with various Hamiltonian and states. The eigenvalues and state affect the constant and power of dt.
However, larger dims decrease the error and `dt` increase it. 
`_compute_max_step` already compute `dt` as a function of `err` for fixed `krylov_dims`. We could compute `krylov_dims` with fixed `dt` and `err` or find optimal `krylov_solve` and `dt`with fixed. But I am not sure it's worth it so I just left a comment that it could be done.

qutip/solver/ode/krylov.py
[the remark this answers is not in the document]
hodgestar: Ah, I see. I will add a comment about it.

qutip/solver/ode/krylov.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        if not krylov_state:
+            krylov_state = self._compute_krylov_set(krylov_tridiag, krylov_basis)
+
+        small_tridiag = _data.Dense(krylov_tridiag.as_ndarray()[:-1, :-1])
+        small_basis = _data.Dense(krylov_basis.as_ndarray()[:, :-1])
+        reduced_state = self._compute_krylov_set(small_tridiag, small_basis)
+
+        def krylov_error(t):
+            return np.log(_data.norm.l2(
```
hodgestar, line 134:
```suggestion
            # we divide by atol and take the log so that the error returned is 0
            # at atol, which is convenient for calling root_scalar with.
            return np.log(_data.norm.l2(
```

hodgestar (approved):

hodgestar:
Thanks for addressing all the issues so quickly. I added a commit for the krylov_error comment I wanted. Happy for this to be merged if the latest test run passes.

Ericgig:
I am working on a merge master into dev.major PR and there will be conflict with the krylov documentation. It would be easier to wait for that and fix the conflict here. Otherwise it will be lost in a 60 files PR.

hodgestar:
Happy for you to merge this once you're done with the merge of master.

Ericgig:
I updated the krylov page after the merger. The code was not touched since the review.

hodgestar (approved):

PR #1992, opened 2022-09-14
Add `Qobj.logm`

### LINKED ISSUE #1990: Add the matrix logrithm method logm() to Qobj.
### Problem Description

I don't find matrix logrithm logm() in Qobj methods, so to compute matrix logrithm I call logm() in scipy.linalg, which changes the type from qobj to ndarray. And to continue the subsequent computation, I need to reinstantiate the resulting ndarray to Qobj, which is cumbersome.
It would be better if the matrix logrithm logm() is added to the Qobj

### Proposed Solution

Add the matrix logrithm method logm() to Qobj.

### Alternate Solutions

_No response_

### Additional Context

_No response_

Ericgig:
**Description**
Add `Qobj.logm()` for the matrix logarithm operation.

**Related issues or PRs**
Fix #1990 for master

### DIFF
--- a/doc/changes/1990.feature
+++ b/doc/changes/1990.feature
@@ -1,0 +1,1 @@
+Add `Qobj.logm` 
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -23,6 +23,7 @@ from numpy import (arccos, arccosh, arcsin, arcsinh, arctan, arctan2, arctanh,
 
 import numpy as np
 import scipy.sparse as sp
+import scipy.linalg
 import qutip.settings as settings
 from qutip import __version__
 from qutip.fastsparse import fast_csr_matrix, fast_identity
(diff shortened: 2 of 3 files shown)


hodgestar (approved):
The code looks good, but should we add this to 4.7.X? And if not, should we add it? I'm not completely against adding it to 4.7.X.

hodgestar:
The Mac OS ci failed with a strange pardiso error from the HEOM tests.

Ericgig:
I am not sure. Bug release should not add new features, but this feels too small for a minor release. Waiting for it in v5 could take some time, so I think it's fine.
But if we choose to wait for v5 for this, I am fine with it. It will be one less merge conflict when merging master to dev.major.

hodgestar:
Hmm. I think let's leave it out. We can reply to the issue with "it's implemented in v5 and here is how you do it in v4". That avoids adding a new feature in a point release (and potentially confusing users) while still solving everyone's immediate problems and making the future better.

PR #1993, opened 2022-09-14
Add `Qobj.logm` for dev.major

### LINKED ISSUE #1990: Add the matrix logrithm method logm() to Qobj.

Ericgig:
**Description**
Add `Qobj.logm()` for the matrix logarithm operation.
Only create the `Dense` operation was created.

`expm_dense` was also added.

**Related issues or PRs**
Fix #1990 for dev.major

### DIFF
--- a/qutip/core/data/expm.py
+++ b/qutip/core/data/expm.py
@@ -1,14 +1,19 @@
 import numpy as np
 import scipy.sparse.linalg
+import scipy.linalg
 
 from .dense import Dense
 from .csr import CSR
-from .properties import isdiag_csr
+from . import csr
+from .make import diag
+from .add import add_dense, add_csr
+from .properties import isdiag_csr, iszero_csr, iszero_dense
 from qutip.settings import settings
 from .base import idxint_dtype
 
 __all__ = [
-    'expm', 'expm_csr', 'expm_csr_dense',
+    'expm', 'expm_csr', 'expm_csr_dense', 'expm_dense',
+    'logm', 'logm_csr', 'logm_dense',
 ]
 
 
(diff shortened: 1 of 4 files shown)


qutip/core/data/expm.py
```diff
@@ -58,6 +66,28 @@ def expm_csr_dense(matrix: CSR) -> Dense:
 expm.add_specialisations([
     (CSR, CSR, expm_csr),
     (CSR, Dense, expm_csr_dense),
+    (Dense, Dense, expm_dense),
```
hodgestar, line 69: What did v5 do for Dense .expm before this?

hodgestar (commented):
The code looks good. It would be nice to have a CSR implementation but I didn't find one in scipy.sparse.

hodgestar:
I see everything using pardiso is failing. :/

qutip/core/data/expm.py
Ericgig: It would convert it to `CSR` automatically and use `expm_csr_dense`. The Dispatcher automatically fill missing specializations.

Ericgig:
Since `scipy.sparse.linalg` does not have `logm`, we need to convert it to an array to do the computation.
Without the csr version, the dispatcher will do it for us.

qutip/core/data/expm.py
hodgestar: I suspected as much -- good to fix that.

hodgestar (approved):

PR #1994, opened 2022-09-18
Added division by scalar to be more flexible.

AGaliciaMartinez:
**Description**
Fix issue part of the issue with qutip/qutip-tensorflow#37.

We rely on `__mul__` to decide what to do with other instead of making it complex in `__truediv__`.

**Related issues or PRs**
This fix was already discussed in #1620.

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -512,9 +512,7 @@ class Qobj:
                     copy=False)
 
     def __truediv__(self, other):
-        if not isinstance(other, numbers.Number):
-            return NotImplemented
-        return self.__mul__(1 / complex(other))
+        return self.__mul__(1 / other)
 
     def __neg__(self):
         return Qobj(_data.neg(self._data),
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -351,6 +351,41 @@ def test_QobjMulNotValidScalar(not_scalar):
         q1 * not_scalar
 
 
+# Allowed division operations (scalar)
+@pytest.mark.parametrize("scalar",
+                         [2+2j,  np.array(2+2j), np.array([2+2j])],
+                         ids=[
+                             "python_number",
+                             "scalar_like_array_shape_0",
+                             "scalar_like_array_shape_1",
+                         ])
+def test_QobjDivisionValidScalar(scalar):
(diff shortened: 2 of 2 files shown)


qutip/tests/core/test_qobj.py
```diff
@@ -351,6 +351,41 @@ def test_QobjMulNotValidScalar(not_scalar):
...
+                         [np.array([2j, 1]), np.array([[1, 2], [3, 4]])],
+                         ids=[
+                             "not_scalar_like_vector",
+                             "not_scalar_like_matrix",
+                         ])
+def test_QobjDivisionNotValidScalar(not_scalar):
+    q1 = qutip.Qobj(np.array([[1, 2], [3, 4]]))
+
+    with pytest.raises(TypeError):
+        not_scalar / q1
```
hodgestar, line 386: I expected this to be `q1 / not_scalar`.

qutip/core/qobj.py
```diff
@@ -512,9 +512,7 @@ def __matmul__(self, other):
                     copy=False)
 
     def __truediv__(self, other):
-        if not isinstance(other, numbers.Number):
-            return NotImplemented
-        return self.__mul__(1 / complex(other))
+        return self.__mul__(1 / other)
```
hodgestar, line 515: Do we need a `try: ... except TypeError: return NotImplemented` here like we did in #1620 for `__mul__`?

qutip/tests/core/test_qobj.py
AGaliciaMartinez: Yes indeed.

qutip/core/qobj.py
AGaliciaMartinez: Thats a good point. With the current implementation it will return `NotImplemented` when doing `mul` so I assume it will default to not trying `(1/other).__mul__`. Not terrible but maybe not ideal. Should I implement it to default to `other.__truediv__`?

AGaliciaMartinez: Well, it turns out that the `NotImplemented` returned by `__mul__` should already work. I added some new tests to ensure that is the case.

Ericgig (approved):

AGaliciaMartinez:
Merging this as failing test seems unrelated.

PR #1997, opened 2022-09-22
Correlation using v5 solver

Ericgig:
**Description**
Update correlation functions to use v5 solvers.
The interface remain the mostly the same, I only made a few changes:
 - Support for mcsolve as a back-end is removed. It was very slow, had a low precision and the saved memory is not as useful as it was 10 years ago. Removing it simplify the code a lot, but i can add it back if there is demand for it.
 - Time-dependent Hamiltonian no longer support numpy array format: some correlation have 2 times list: `tlist` and `taulist`. It was not clear which one was to be used when using that format. `QobjEvo` input are supported, so building it before passing it to the correlation function is the way to go.
 - `a_op`, `b_op`, `c_op` can be `QobjEvo`.

Spectrum functions have been moved to their own file, but are mostly unchanged.

I added `correlation_3op` which work as other correlation functions, but take a `Solver` instead of an Hamiltonian and `c_ops`. This allows any solver able to evolve density matrices to be used (`BRSolver`, `HEOMSolver`). It is only one function that support it directly: adding support of `Solver` to existing function would make the interface quite complicated, and changing it would break a lot of code and calling them using `H` is useful. 

Fixed a small bugs in brmesolve and heom. I added data type option for the eigen solver of the diag integrator as it would otherwise sometime fail with cases in correlation (system and state not hermitian.)


qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+        List of collapse operators
+    a_op : Qobj, QobjEvo
+        Operator A.
+    b_op : Qobj, QobjEvo
+        Operator B.
+    c_op : Qobj, QobjEvo
+        Operator C.
+    solver : str
+        Choice of solver (`me` for master-equation, and `es` for exponential
+        series).
```
AGaliciaMartinez, line 249: I think we can give a little bit more information about the solver as string:
```suggestion
        Choice of solver (`me` for master-equation, and `es` for exponential
        series). Both solver use `MeSolver` internally. However, `es`
        method sets the integrator to be 'diag'. If a custom integrator is to be
        employed, use 'me' option.
```

qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+    options : dict
+        Options for the solver.
```
AGaliciaMartinez, line 251:
```suggestion
        Options for the solver. 'method' option is overridden by solver
        option 'es'.
```

qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+        raise ValueError("MC solver for correlation has been removed")
+    return solver_instance
+
+
+def correlation_3op(solver, state0, tlist, taulist, A=None, B=None, C=None):
+    r"""
+    Calculate the three-operator two-time correlation function:
+    :math:`\left<A(t)B(t+\tau)C(t)\right>`
+
+    Note: it is not possibly to calculate a physically meaningful correlation
```
AGaliciaMartinez, line 427:
```suggestion
    Note: it is not possible to calculate a physically meaningful correlation
```

qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+    of this form where :math:`\tau<0`.
```
AGaliciaMartinez, line 428:
```suggestion
   where :math:`\tau<0`.
```

qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+    C : Qobj, QobjEvo, optional, default=None
+        Operator C.
+    """
+
+    taulist = np.asarray(taulist)
+
+    dims = state0.dims[0]
+    A = _fix_op_type(A, dims)
+    B = _fix_op_type(B, dims)
+    C = _fix_op_type(C, dims)
```
AGaliciaMartinez, line 456:
```suggestion
    A = qeye(dims) if A in [None, 1] else A
    B = qeye(dims) if B in [None, 1] else B
    C = qeye(dims) if C in [None, 1] else C
    A, B, C = QobjEvo(A), QobjEvo(B), QobjEvo(C)
```
I know this lead to some repetition in the code but I believe readability is much improved in this way. Besides, this is hopefully just one more line.

qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+
+    if isinstance(solver, (MeSolver, BRSolver, HEOMSolver)):
+        out = _correlation_3op_dm(solver, state0, tlist, taulist, A, B, C)
+    elif isinstance(solver, McSolver):
+        raise TypeError("Monte Carlo support for correlation was removed. "
+                        "Please tell us on github if you need it.")
+    else:
+        raise TypeError("Only solver able to evolve density matrices"
```
AGaliciaMartinez, line 464:
```suggestion
        raise TypeError("Only solvers able to evolve density matrices"
```

qutip/solver/correlation.py
```diff
@@ -0,0 +1,497 @@
...
+    dims = state0.dims[0]
+    A = _fix_op_type(A, dims)
+    B = _fix_op_type(B, dims)
+    C = _fix_op_type(C, dims)
+
+    if isinstance(solver, (MeSolver, BRSolver, HEOMSolver)):
+        out = _correlation_3op_dm(solver, state0, tlist, taulist, A, B, C)
+    elif isinstance(solver, McSolver):
+        raise TypeError("Monte Carlo support for correlation was removed. "
+                        "Please tell us on github if you need it.")
```
AGaliciaMartinez, line 462:
```suggestion
                        "Please, tell us on GitHub issues if you need it!")
```

qutip/solver/ode/qutip_integrator.py
```diff
@@ -174,14 +174,20 @@ def get_state(self, copy=True):
         return self._t, _data.matmul(self.U, _data.dense.Dense(self._y))
 
     def set_state(self, t, state0):
+        self._is_set = True
         self._t = t
         self._y = _data.matmul(self.Uinv, state0).to_array()
         self._is_set = True
```
AGaliciaMartinez, line 180: I am not sure what `self._is_set` is meant to do but I find it weird that it is "set" twice. Can you check that both are needed?

qutip/tests/solver/test_correlation.py
```diff
@@ -127,6 +126,7 @@ def _2ls_g2_0(H, c_ops):
     sp = qutip.sigmap()
     start = qutip.basis(2, 0)
     times = _2ls_times
+    H = qutip.QobjEvo(H, args=_2ls_args, tlist=times)
```
AGaliciaMartinez, line 129: Just a minor question here. Why is it necessary to explicitly make H a QobjEvo? I am worried that if this is necessary there might have been a breaking change.

qutip/tests/solver/test_correlation.py
```diff
@@ -269,3 +267,37 @@ def test_correlation_2op_1t_known_cases(solver,
...
+def test_alternative_solver():
+    from qutip.solver.mesolve import MeSolver
+    from qutip.solver.brmesolve import BRSolver
+
+    H = qutip.num(5)
+    a = qutip.destroy(5)
+    a_ops = [(a+a.dag(), qutip.coefficient(lambda _, w: w>0, args={"w":0}))]
+
+    br = BRSolver(H, a_ops)
+    me = MeSolver(H, [a])
```
AGaliciaMartinez, line 297: If we support `HEMOSolver` I guess we should also add it here.

AGaliciaMartinez (commented):
Hey @Ericgig , I have not finished yet the review entirely but wanted to share the progress so far to get this merger quicker. Overall I feel the code is GREAT! Code is very readable and most of the comments are really minor nits. Really nice that different solvers can be used to compute correlation. Also, the SolverClass really seem to shine in the implementation by geting read of the not so clear shift in time function. Great work!

I see that support for McSolver has not been ported yet. I agree though it is best to leave it for later based on the community feedback.

doc/guide/guide-correlation.rst
```diff
@@ -28,21 +28,24 @@ QuTiP provides a family of functions that assists in the process of calculating
...
+| :func:`qutip.correlation_2op_2t` | :math:`\left<A(t)B(t+\tau)\right>`.              |
++----------------------------------+--------------------------------------------------+
+|                                  | :math:`\left<A(\tau)B(0)\right>` or              |
+| :func:`qutip.correlation_2op_1t` | :math:`\left<A(0)B(\tau)\right>`.                |
++----------------------------------+--------------------------------------------------+
+| :func:`qutip.correlation_3op_1t` | :math:`\left<A(0)B(\tau)C(0)\right>`.            |
++----------------------------------+--------------------------------------------------+
+| :func:`qutip.correlation_3op_2t` | :math:`\left<A(t)B(t+\tau)C(t)\right>`.          |
++----------------------------------+--------------------------------------------------+
+| :func:`qutip.correlation_3op`    | :math:`\left<A(t)B(t+\tau)C(t)\right>`.          |
```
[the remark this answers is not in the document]
AGaliciaMartinez: We may want to be clear in the in the docstrings. Maybe a See Also section in all the old functions that point to correlation_3op could be helpful to explain how to use other solvers.

qutip/tests/solver/test_correlation.py
Ericgig: To test array based coefficient, we need to pass a `tlist`. `correlation` has `tlist` and `taulist` which would case confusion as to which should be used. 
In v4, array as coefficient where accepted, but it uses `tlist` in the first call and `taulist` in the following, not user how it handled the shift in the taus.

qutip/tests/solver/test_correlation.py
Ericgig: I do not know how to build a `Bath` that would represent a similar physic to compare with these results.
@hodgestar do you know any `HEOM` setup that can be compared to result of other solver or that we could compute analytically?
Otherwise I can add a test that `HEOMSolver` can run without error in correlation, but it could return junk and I would not know.

qutip/solver/ode/qutip_integrator.py
Ericgig: This was a mistake, fixing.

doc/guide/guide-correlation.rst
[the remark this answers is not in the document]
Ericgig: Added the `see also`

qutip/solver/correlation.py
Ericgig: Reworded in my words

AGaliciaMartinez (commented):
I have made some last suggestions but after this the PR should be ready to merge.

Note: I have not reviewed thoroughly the spectrum file as it seems to me mostly move/pep8 fix from the previous functions. I think we should eventually clean up spectrum.py as it seems it could benefit from some refactoring, but I do not think that is the scope of this PR.

qutip/solver/ode/qutip_integrator.py
```diff
@@ -181,7 +181,12 @@ def set_state(self, t, state0):
     @property
     def options(self):
         """
-        Diagonalization method do not use any options.
+        Supported options by verner method:
```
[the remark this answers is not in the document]
AGaliciaMartinez: I assumed this was diag from the integrator name, but let me know if im wrong.

qutip/solver/correlation.py
```diff
@@ -0,0 +1,520 @@
...
+    return out
+
+
+def _correlation_3op_dm(solver, state0, tlist, taulist, A, B, C):
+    old_opt = solver.options.copy()
+    try:
+        # We don't want to modify the solver
+        # TODO: Solver could have a ``with`` or ``copy``.
+        solver.options["normalize_output"] = False
+        solver.options["progress_bar"] = False
```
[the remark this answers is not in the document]
AGaliciaMartinez: Maybe for the moment we could just raise a warning if the values are being changed and later figure a better way of handling these situations?

qutip/solver/correlation.py
```diff
@@ -0,0 +1,520 @@
...
+        element `0`.
+    taulist : array_like
+        List of times for :math:`\tau`. taulist must be positive and contain
+        the element `0`.
+    A : Qobj, QobjEvo, optional, default=None
+        Operator A.
+    B : Qobj, QobjEvo, optional, default=None
+        Operator B.
+    C : Qobj, QobjEvo, optional, default=None
+        Operator C.
```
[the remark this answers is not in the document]
Ericgig: I reworked it, merged all three into one entry to remove duplication.
I prefer the `None` to `1` since it shows better that you skip that operator, allowing to compute 2 operators correlation, instead of measuring a 3 operators correlation with one being the identity.

qutip/solver/correlation.py
```diff
@@ -0,0 +1,520 @@
...
+        state0 = steadystate(H, c_ops)
+        n = np.array([expect(state0, a_op.dag() * a_op)])
+    else:
+        n = solver.run(state0, taulist, e_ops=[a_op.dag() * a_op]).expect[0]
+
+    # calculate the correlation function G2 and normalize with n to obtain g2
+    G1 = correlation_3op(solver, state0, [0], taulist,
+                         a_op.dag(), a_op.dag() * a_op, a_op)
+
+    g2 = G2 / (n[0] * n)
```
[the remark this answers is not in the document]
AGaliciaMartinez: Thank you Eric!

qutip/tests/solver/test_correlation.py
Ericgig: Talking to Simon, `HEOMSolver` would not return the proper results since the density matrix does not contain the full state of the solver. It should be doable, but need it's own code. Removing for now.

qutip/solver/correlation.py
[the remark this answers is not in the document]
Ericgig: Simple test added. 
There where other bugs. Nice catch.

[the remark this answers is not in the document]
AGaliciaMartinez: Thanks for the work Eric!

AGaliciaMartinez (approved):
Thanks Eric for your work!

PR #1999, opened 2022-09-23
Allow propagator to accept Solver input

Ericgig:
**Description**
Allow `Propagator` to take a `Solver` instead of an Hamiltonian, this allow to make propagator work with brmesolve or HEOM.

To help with that, I made the system's solver dimensions accessible through `sys_dims` and whether they represent super dims in `evolve_dm`.

### DIFF
--- a/qutip/solver/brmesolve.py
+++ b/qutip/solver/brmesolve.py
@@ -285,7 +285,6 @@ class BRSolver(Solver):
         self._integrator = self._get_integrator()
         self._state_metadata = {}
         self.stats = self._initialize_stats()
-        self.sys_dims = self.rhs.dims[0]
         self.evolve_dm = True
 
 
--- a/qutip/solver/brmesolve.py
+++ b/qutip/solver/brmesolve.py
@@ -285,7 +285,6 @@ class BRSolver(Solver):
         self._integrator = self._get_integrator()
         self._state_metadata = {}
         self.stats = self._initialize_stats()
-        self.evolve_dm = True
 
 
     def _initialize_stats(self):
--- a/qutip/solver/brmesolve.py
+++ b/qutip/solver/brmesolve.py
@@ -66,7 +66,7 @@ def brmesolve(H, psi0, tlist, a_ops=[], e_ops=[], c_ops=[],
             ]
 
         .. note:
-            ``Cubic_Spline`` has been replaced by :class:`Coefficient`\:
+            ``Cubic_Spline`` has been replaced by :class:`Coefficient`:
                 ``spline = qutip.coefficient(array, tlist=times)``
 
             Whether the ``a_ops`` is time dependent is decided by the type of
@@ -285,6 +285,8 @@ class BRSolver(Solver):
         self._integrator = self._get_integrator()
         self._state_metadata = {}
         self.stats = self._initialize_stats()
+        self.sys_dims = self.rhs.dims[0]
+        self.evolve_dm = True
 
 
     def _initialize_stats(self):
--- a/qutip/solver/heom/bofin_solvers.py
+++ b/qutip/solver/heom/bofin_solvers.py
@@ -1135,13 +1135,6 @@ class HEOMSolver(Solver):
         """
         super().start(state0, t0)
 
-    def step(self, t, *, args=None, copy=True):
-        ado = super().step(t, args=args, copy=copy)
-        if self.options["store_ados"]:
-            return ado
-        else:
-            return ado.rho
-
     @property
     def options(self):
         """
--- a/qutip/solver/heom/bofin_solvers.py
+++ b/qutip/solver/heom/bofin_solvers.py
@@ -654,12 +654,11 @@ class HEOMSolver(Solver):
         self._init_rhs_time = time() - _time_start
 
         super().__init__(rhs, options=options)
-        self.evolve_dm = True
 
     @property
     def sys_dims(self):
         """
-        Dimensions of the space that the system use:
+        Dimensions of the space that the system use, excluding any environment:
 
         ``qutip.basis(sovler.dims)`` will create a state with proper dimensions
         for this solver.
--- a/qutip/solver/heom/bofin_solvers.py
+++ b/qutip/solver/heom/bofin_solvers.py
@@ -16,7 +16,7 @@ import numpy as np
 import scipy.sparse as sp
 from scipy.sparse.linalg import spsolve
 
-from qutip import settings
+from qutip.settings import settings
 from qutip import state_number_enumerate
 from qutip.core import data as _data
 from qutip.core.data import csr as _csr
--- a/qutip/solver/heom/bofin_solvers.py
+++ b/qutip/solver/heom/bofin_solvers.py
@@ -1075,25 +1075,13 @@ class HEOMSolver(Solver):
             rho0_he = rho0_he.reshape(n ** 2 * self._n_ados)
             rho0_he = _data.create(rho0_he)
         else:
-            if rho0.dims == rho_dims:
-                n_sub = 1
-            elif rho0.dims[0] == rho_dims:
-                n_sub = rho0.shape[1]
-            else:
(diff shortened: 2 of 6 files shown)


qutip/solver/brmesolve.py
```diff
@@ -285,6 +285,8 @@ def __init__(self, H, a_ops, c_ops=None, sec_cutoff=0.1, *, options=None):
         self._integrator = self._get_integrator()
         self._state_metadata = {}
         self.stats = self._initialize_stats()
+        self.sys_dims = self.rhs.dims[0]
```
AGaliciaMartinez, line 288: I will suggest making this a property if it always depends rhs.

qutip/solver/propagator.py
```diff
@@ -138,22 +145,25 @@ class Propagator:
...
-            self.solver = MeSolver(Hevo, c_ops=c_ops, options=options)
-        else:
-            self.props = [qeye(Hevo.dims[0])]
-            self.solver = SeSolver(Hevo, options=options)
+        self.props = [qeye(self.solver.sys_dims)]
         self.cte = self.solver.rhs.isconstant
-        self.unitary = (not self.solver.rhs.issuper
-                        and isinstance(H, Qobj)
-                        and H.isherm)
+        self.unitary = isinstance(system, SeSolver)
```
AGaliciaMartinez, line 166: Is this correct? If the solver is created with a non-hermitian matrix, should the propagator _not_ be unitary?

AGaliciaMartinez:
The feature is very interesting and makes the propagator class very flexible. I like it! I left some small comments but otherwise looks good.

qutip/solver/solver_base.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, rhs, *, options=None):
             self.rhs = QobjEvo(rhs)
         else:
             TypeError("The rhs must be a QobjEvo")
+        self.sys_dims = rhs.dims[0]
```
hodgestar, line 49: We should document the two new attributes in the class docstring.

qutip/solver/solver_base.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, rhs, *, options=None):
...
+        self.evolve_dm = rhs.issuper
```
hodgestar, line 50: I'm guessing `evolve_dm` means that a density matrix can be passed to `.run()` and similar, but the `evolve_dm` attribute doesn't appear to be used by anything in this PR yet.

qutip/solver/propagator.py
```diff
@@ -138,22 +145,25 @@ class Propagator:
         into a :class:`QobjEvo` with ::
             U = QobjEvo(Propagator(H))
     """
-    def __init__(self, H, c_ops=(), args=None, options=None,
+    def __init__(self, system, *, c_ops=(), args=None, options=None,
                  memoize=10, tol=1e-14):
-        Hevo = QobjEvo(H, args=args)
-        c_ops = [QobjEvo(op, args=args) for op in c_ops]
+        if isinstance(system, McSolver):
```
hodgestar, line 150: I would prefer it if we didn't have to explicitly check for `McSolver` here so that people who write other solvers can specify themselves whether it supports use in the Propagator.

I'm not sure is `evolve_dm` was the attribute intended for use here instead or not.

hodgestar (changes requested):
This looks like a great feature.

In addition to the individual comments, I think it would be good to add a test to each solver class to test it's use with Propagator (including those solver classes which should raise a not-supported error).

qutip/solver/brmesolve.py
Ericgig: `HEOMSolver` is the exception. It has the `_sys_dims` parameter for the physical dimensions, but rhs is expanded to include the bath.
But a property make sense, I don't see any reason to overwrite the value.

qutip/solver/propagator.py
Ericgig: `evolve_dm` is something I wanted for correlation when doing #1997.
I will use it instead of `isinstance(solver, SeSolver)`

Ericgig:
@hodgestar, to make it work with HEOM, I had to change `step` to return `rho` directly, (controlled by the `store_ados` options). Could you take a look and tell me if you are fine with this approach.

Also, do you know  a `Bath` configuration that would result in evolution similar to `mesolve` so we could use it in tests.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -1111,6 +1135,13 @@ def start(self, state0, t0):
         """
         super().start(state0, t0)
 
+    def step(self, t, *, args=None, copy=True):
```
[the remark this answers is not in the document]
hodgestar: Doh -- `.step()` keeps its own state, so resolving this particular comment.

hodgestar:
The current `.evolve_dm` and `.sys_dims` concepts both seem a bit awkward, so I'd like to suggest the following:

**For `.evolve_dm`**:

Replace `.evolve_dm` with a `.unitary` attribute on the solver itself. The attribute should be `True` if that particular instance of the solver can guarantee that the evolution it produces is unitary, and `False` otherwise.

This will be more generally useful and will simplify the code inside the propagator a bit.

**For `.sys_dims`**:

The meaning of what `.sys_dims` returns is already unclear in this PR. The docstring says its the dimension of the system. Practically I assume that to mean "the dimension of the initial state or dm passed to .run()". However, in the PR its used as the size of the propagator. These are the same if there is no environment, but will be different otherwise.

I suggest we keep `.sys_dims` but document that it is the dimensions of the system excluding any environment.

**Specifying which solvers support propagators**:

Currently we have a list of explicit checks at the top of the propagator constructor, but we could instead have a "propagator_class" attribute on solvers the same way we have a "result_class" and then that could be set to some suitable default for ordinary solvers and None or "PropagatorNotSupported" for solvers that don't support them.

Ericgig:
> The current `.evolve_dm` and `.sys_dims` concepts both seem a bit awkward, so I'd like to suggest the following:
> 
> **For `.evolve_dm`**:
> 
> Replace `.evolve_dm` with a `.unitary` attribute on the solver itself. The attribute should be `True` if that particular instance of the solver can guarantee that the evolution it produces is unitary, and `False` otherwise.
> 
> This will be more generally useful and will simplify the code inside the propagator a bit.

I removed `.evolve_dm`.

> **For `.sys_dims`**:
> 
> The meaning of what `.sys_dims` returns is already unclear in this PR. The docstring says its the dimension of the system. Practically I assume that to mean "the dimension of the initial state or dm passed to .run()". However, in the PR its used as the size of the propagator. These are the same if there is no environment, but will be different otherwise.
> 
> I suggest we keep `.sys_dims` but document that it is the dimensions of the system excluding any environment.

Changed in HEOMSolver. Since environment only means something for HEOM, so it took some time to understand what it referred to. Adding the change elsewhere would be confusing.

 
> **Specifying which solvers support propagators**:
> 
> Currently we have a list of explicit checks at the top of the propagator constructor, but we could instead have a "propagator_class" attribute on solvers the same way we have a "result_class" and then that could be set to some suitable default for ordinary solvers and None or "PropagatorNotSupported" for solvers that don't support them.

There is only one class for propagator and I don't see us needing any other, so there is no use for a `propagator_class` . We could have a `support_propagator` flag, but I prefer it this way as we can set a special error message for `HEOMSolver`. 
`HEOMSolver` is the only exceptions where we could support it but choose not to take the time to implement it. 
All deterministic solver we have can be used for propagator, trajectories solver don't have a properly defined propagator for single trajectory and I don't see any use to create a propagator for the averaged states. It would take a lot of trajectories to obtain what is effectively mesolve's propagator. 

We could add the flag to the integrator instead, some krylov integrator cannot evolve 2D states, but it's the only one we have plan for and it's not merged yet...

hodgestar (changes requested):
I'm happy with the rest of the PR.

hodgestar (approved):

PR #2000, opened 2022-09-23
Keep isherm flag in constant QobjEvo

Ericgig:
**Description**
In solvers, we often wrap `Qobj` into `QobjEvo` to remove code duplication while allowing support for time-dependant operators even when constant case are the norm (`e_ops`). Doing so we lose the `isherm` flag, which can sometime be nice to have. 
This PR allow the flag to be kept when calling the `QobjEvo` in such case:
```
>>> QobjEvo(qeye(N))(0)._isherm
True
```

This will be useful for the `diag` integration method which need compute the eigenstates of a constant `QobjEvo`.

### DIFF
--- a/qutip/core/cy/qobjevo.pyx
+++ b/qutip/core/cy/qobjevo.pyx
@@ -312,11 +312,27 @@ cdef class QobjEvo:
             if _args is not None:
                 kwargs.update(_args)
             return QobjEvo(self, args=kwargs)(t)
+
+        t = self._prepare(t, None)
+
         if self.isconstant:
             # When the QobjEvo is constant, it is usually made of only one Qobj
             return sum(element.qobj(t) for element in self.elements)
+
+        cdef _BaseElement part = self.elements[0]
+        cdef double complex coeff = part.coeff(t)
+        obj = part.qobj(t)
+        cdef Data out = _data.mul(obj.data, coeff)
+        cdef bint isherm = <bint> obj._isherm and coeff.imag == 0
+        for element in self.elements[1:]:
+            part = <_BaseElement> element
+            coeff = part.coeff(t)
+            obj = part.qobj(t)
+            isherm &= <bint> obj._isherm and coeff.imag == 0
+            out = _data.add(out, obj.data, coeff)
+
         return Qobj(
-            self._call(t), dims=self.dims, copy=False,
+            out, dims=self.dims, copy=False, isherm=isherm or None,
             type=self.type, superrep=self.superrep
         )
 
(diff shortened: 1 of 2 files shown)


AGaliciaMartinez:
Hey @Ericgig.This is a great feature! I wanted to suggest a slightly different implementation that would work also for non constant cases. We could do checking that the f(t) value is real and that all elements are hermitian. However, the following does not seem to work for me (on a jupyter noteebok with latest dev.major):
```python
qutip.QobjEvo(qt.sigmax()).elements
```
are elements only accessible from cython?

qutip/core/cy/qobjevo.pyx
```diff
@@ -312,6 +312,9 @@ cdef class QobjEvo:
             if _args is not None:
                 kwargs.update(_args)
             return QobjEvo(self, args=kwargs)(t)
+        if self.isconstant:
+            # When the QobjEvo is constant, it is usually made of only one Qobj
```
hodgestar, line 316:
```suggestion
            # For constant QobjEvo's, we sum the contained Qobjs directly in order to
            # retain the cached values of attributes like .isherm when possible, rather than
            # calling _call(t) which may lose this cached information.
```

hodgestar:
@AGaliciaMartinez The `.elements` are intentionally only accessible from Cython (it's not safe to modify the list from Python because the QobjEvo element lists are meant to be immutable). If you have an idea for a cleaner implementation of this though, maybe just write the idea here and then we can all look at it?

AGaliciaMartinez:
I was thinking something on the same lines as in `__mul__`for `Qobj`:
``
coeff_real = all(element.coeff(t).imag==0 for element in self.elements) 
data_herm = all(element.part(t).isherm for element in self.elements)
return Qobj(
            self._call(t), dims=self.dims, copy=False,
            type=self.type, superrep=self.superrep,
            isherm=(coeff_real and data_herm) or None
        )
```
Although I imagine it can be made more efficient by not looping twice in self.elements and probably using the loop in `_call`method. This way you always try to guess the hermeticity of QobjEvo even if it is not constant.

**Comment by Ericgig** [comment]
I added the code to guess Hermiticity in non-constant cases, but it mean some code duplication (or a 2x slowdown.)

**Review by hodgestar — APPROVED** [review]
I'm not wild about the code duplication -- I feel like we're missing some sort of sum operation for Qobjs so that we can write:
```
return Qobj.sum(element.qobj(t) for element in self.elements)
```
and have that do the right thing in all cases.

I'm okay with leaving that for another day though.

AGaliciaMartinez:
Just one last small thing, do we need `_call` anymore?

Ericgig:
`_call` is used for bloch redfield tensor. It allow to skip the overhead of using `Qobj` and can be call from cython space.

AGaliciaMartinez (approved):

AGaliciaMartinez:
Thanks for the answer! I will also approve this.

PR #2002, opened 2022-09-29
Add `solve` and `svn` to dispatched function

Ericgig:
**Description**
Add data layer functions needed to update `steadystate` to use qutip's data layer.

Most of `steadystate`'s methods solve the system `L(rho_ss) = 0` using different solver from numpy, scipy, mkl.
This add the `solve` dispatched function to wraps this operation at the data layer level.
For the `CSR` version, it gives access to all methods available in `scipy.sparse.linalg` and our mkl version.
For the `Dense` version, `solve` and `lstsq` from `numpy.linalg` are made available.
Options supported by numpy / scipy for each function are pass through:
`qutip.data.solve(A, B, method, options)` call `scipy.sparse.linalg.method(A, B, **options)`.

Other methods of `steadystate` decompose the Liouvillian with either `eigen` or `svd`.
`svd` was added as a dispatched function, but only with specialization for `Dense`. 
I made a sparse version, but  `svds` can't readily compute all singular values and does not respond well to splitting the problem into 2 like it was done for sparse eigensolver. 

This PR only include the new dispatched functions, not updated `steadystate` using them or `Qobj` interface.

### DIFF
--- a/qutip/core/data/eigen.py
+++ b/qutip/core/data/eigen.py
@@ -1,6 +1,7 @@
 import numpy as np
 import scipy.linalg
 import scipy.sparse as sp
+import scipy.sparse.linalg
 from itertools import combinations
 
 from .dense import Dense, from_csr
@@ -10,6 +11,7 @@ from qutip.settings import settings
 
 __all__ = [
     'eigs', 'eigs_csr', 'eigs_dense',
+    'svd', 'svd_csr', 'svd_dense',
 ]
 
 
@@ -276,6 +278,8 @@ def eigs_dense(data, isherm=None, vecs=True, sort='low', eigvals=0):
 
 
 from .dispatch import Dispatcher as _Dispatcher
+import inspect as _inspect
+
 
 # We use eigs_dense as the signature source, since in this case it has the
 # complete signature that we allow, so we don't need to manually set it.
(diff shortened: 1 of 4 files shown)


qutip/core/data/eigen.py
```diff
@@ -323,4 +327,127 @@ def eigs_dense(data, isherm=None, vecs=True, sort='low', eigvals=0):
...
+    -------
+    U : Dense
+        Left singular vectors as columns. Only returned if ``vecs == True``.
+    S : np.ndarray
+        Singular values.
+    Vh : Dense
+        Right singular vectors as rows. Only returned if ``vecs == True``.
+    """
+# Dense implementation return all states, but sparse implementation compute
+# only a few states. So only the dense version is registed.
```
hodgestar, line 446:
```suggestion
# only a few states. So only the dense version is registered.
```

hodgestar (commented):
I think allowing specializing solving and SVD is extremely useful. I have a few questions though:

- Do you have a specific use case that you're adding it for right now?
- It will be hard to support methods such as "mkl_spsolve" on a strange data backend, so essentially inside qutip itself it is not okay to pass a method except when it was supplied to us by the user. Do you have any idea how to ensure we don't forget that rule? Or any ideas on how to provide an interface that allows specifying a method but which also doesn't break if someone hardcodes a method name in a place that is difficult for users to override?

Ericgig:
`svd` is used by `steadystate`.

There are no method in common between `Dense` and `CSR`, the default is `solve` for `Dense`, but `spsolve` for `CSR`. All iterative methods are only available in `sparse.linalg`... So we cannot pass it anywhere thus I think it won't be hard to remember.

If we want to have the default change depending on whether we have mkl or not, we could read the default method from core's settings.

hodgestar (approved):

PR #2008, opened 2022-10-14
Add missing Dense specialisations

Ericgig:
**Description**
`Dense` was missing some dispatch specialisations:
`isherm`, `isdiag`, `inner`, `inner_op`, `pow`, `indices` and `dimensions`.
Some of them are quite commonly used:
- `isherm` is called in a few method of `Qobj` like `eigenstates`.
- `inner` is used in `Qobj.__matmul__` between `bra` and `ket`.

Tests for new functions are added.
For `isherm`, it was already tested for Dense, but relied on the automatic conversion of the dispatcher.

Also fix some docstings which used `Arguments` instead of `Parameters`.

### DIFF
--- a/qutip/core/data/properties.pyx
+++ b/qutip/core/data/properties.pyx
@@ -15,8 +15,8 @@ cdef extern from *:
     void *PyMem_Calloc(size_t nelem, size_t elsize)
 
 __all__ = [
-    'isherm', 'isherm_csr',
-    'isdiag', 'isdiag_csr',
+    'isherm', 'isherm_csr', 'isherm_dense',
+    'isdiag', 'isdiag_csr', 'isdiag_dense',
     'iszero', 'iszero_csr', 'iszero_dense',
 ]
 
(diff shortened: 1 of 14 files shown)


Ericgig:
Also fixed dense matmul to use `zgemv` for `oper@ket`.
In the `oper @ ket` [matmul scaling benchmarks](https://qutip.org/qutip-benchmark/s_matmul.html), `Dense` is slower than `numpy`. Our dense data layer used `zgemm` for both matrix @ matrix and matrix @ vector operations. By switching to `zgemv` we get closer to numpy's times: faster in `C` format but slower in `Fortran`.
`bra@oper` still use `zgemm`.

qutip/core/data/properties.pyx
```diff
@@ -144,6 +144,39 @@ cpdef bint isherm_csr(CSR matrix, double tol=-1):
...
+    -------
+    bint
+        Boolean True if it is Hermitian, False if not.
+    """
+    if matrix.shape[0] != matrix.shape[1]:
+        return False
+    tol = tol if tol >= 0 else settings.core["atol"]
+    cdef size_t row, col, size=matrix.shape[0]
+    for row in range(size):
+        for col in range(size):
```
hodgestar, line 170:
```suggestion
        for col in range(row + 1):
```
We could just check the lower triangle plus diagonal.

hodgestar (changes requested):
I think we can improve is_herm_dense, but otherwise looks great.

hodgestar (approved):

PR #2015, opened 2022-11-01
Remove limitation for autodiff

Ericgig:
**Description**
Make changes so enable autodiff with `jax` or `tensorflow`:
- In `QobjEvo`, `Elements`, `Coefficient`: `t` is a python object instead of a double as is coefficient output.
  This allows to pass jax traced variable or tensorflow's scalar tensor.
- QobjEvo's elements attribute are accessible and new methods `QobjEvo._getstate`, `QobjEvo._restore`.
  Used by jax layer to register `QobjEvo` as a pytree.
- Allow subproject to add their own coefficient.
  This allow to make a coefficient that can jit well or can be differentiated. `FuncCoefficient` without type restriction could mostly work, but there are some issues with registering them as a pytree or pickling.
- In CoreOptions, an options for the default data type is added:
  When not specified, each `Qobj` creation function keeps its default.
```
with CoreOptions(default_dtype="jax"):
    obj = qeye(N)
    assert isinstance(obj.data, JaxArray)
```
- Add `QobjEvo.to_list` to restore the list format of the `QobjEvo`.

**Related issues or PRs**
Replace #1816


qutip/core/coefficient.py
```diff
@@ -47,8 +47,24 @@ class StringParsingWarning(Warning):
...
+# `Coefficient` specialization here.
+coefficient_builders = {
+    Coefficient: _return,
+    np.ndarray: InterCoefficient,
+    scipy.interpolate.PPoly: InterCoefficient.from_PPoly,
+    scipy.interpolate.BSpline: InterCoefficient.from_Bspline,
+}
+
+
 def coefficient(base, *, tlist=None, args={}, args_ctypes={},
```
AGaliciaMartinez, line 66: The documentation for this function is great! although I still miss the Parameters section. For instance, it took me some digging to understand what kwargs are meant to be used for. Is it mainly used to supprt the builders from jax?

qutip/core/coefficient.py
```diff
@@ -353,6 +367,9 @@ def coeff_from_str(base, args, args_ctypes, compile_opt=None):
     return coeff(base, keys, const, args)
 
 
+coefficient_builders[str] = coeff_from_str
```
AGaliciaMartinez, line 370: Could we move this one with the rest for clarity?

qutip/core/cy/coefficient.pyx
```diff
@@ -499,11 +501,11 @@ cdef class InterCoefficient(Coefficient):
         return out
 
     @classmethod
-    def from_PPoly(cls, ppoly):
+    def from_PPoly(cls, ppoly, **_):
```
AGaliciaMartinez, line 504: I am wondering if we should keep the **_. This can lead to confusing behavior were some arguments get ignored.
```
def f(a):
    pass

f(1, **{}) # this does not raise error
f(1, **{"m" : 1})  # I think we should make this raise an error.
```

AGaliciaMartinez (changes requested):
Hi Eric, nice work! It is actually surprising that not so many changes are needed to make this work. I left some suggestions but seems this is pretty much ready to be merged.

My biggest concern is with the addition of `**_` to some of the functions, I would argue that it is best to raise an error if the function does not have arguments. this may lead to code that works on jax being incompatible with code that works on just qutip though, so it is something to think about. I will let you decide what would work best here (perhaps @hodgestar has a suggestion for this?).

Otherwise, I like the addition of `default_dtype` although I think we should be more clear on what happens when `dtype=None` and `defaul_dtype=None`. This could go in a follow up pr. I could be an interesting idea for the future to consider adding a  similar context manager that removes other dtypes (`CSR` and `Dense`) so that it is clear what the code is doing?

qutip/core/_brtensor.pyx
```diff
@@ -265,14 +265,14 @@ cdef class _BlochRedfieldElement(_BaseElement):
             return _br_term_data(A_eig, self.spectrum, self.skew, cutoff)
         raise ValueError('Invalid tensortype')
 
-    cpdef object qobj(self, double t):
+    cpdef object qobj(self, t):
         return Qobj(self.data(t), dims=self.dims, type="super",
                     copy=False, superrep="super")
 
-    cpdef double complex coeff(self, double t) except *:
+    cpdef object coeff(self, t):
```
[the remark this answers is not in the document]
Ericgig: Errors do not propagate when returning a c variable. Here we were returning a `double complex` thus erasing all raised error without the `except *`. Without the output type, a python object is returned and it pass the error along.

qutip/core/coefficient.py
Ericgig: Easier said than done, we would need to split the file in 2...
We must add it after defining `coeff_from_str` thus we can't add it with the other, but building `coefficient_builders` too far from the function would be confusing.

qutip/core/cy/coefficient.pyx
Ericgig: The `**_` is not for the users defined functions, but for the function that build the `Coefficient` object.
The `coefficient` function has the `args`, `order`, `tlist`, etc. that it need to pass down, function based coefficients need the `args`, the string coefficients use the `order` and `tlist`. 
Since `coefficient` no longer have special cases for each type, we don't filter the input anymore.

qutip/core/cy/qobjevo.pyx
```diff
@@ -285,6 +285,33 @@ cdef class QobjEvo:
 
         return out
 
+    @classmethod
+    def _restore(cls, elements, dims, shape, type, superrep, flags):
+        """ Recreate a QobjEvo without using __init__ """
```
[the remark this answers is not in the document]
Ericgig: Yes

qutip/core/_brtensor.pyx
```diff
@@ -265,14 +265,14 @@ cdef class _BlochRedfieldElement(_BaseElement):
             return _br_term_data(A_eig, self.spectrum, self.skew, cutoff)
         raise ValueError('Invalid tensortype')
 
-    cpdef object qobj(self, double t):
```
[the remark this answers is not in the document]
Ericgig: I did a few tests and it's actually faster in average:
For string and array based coefficient there is a small (~10 ns) cost.
But function coefficient are faster since they already return a python object and we did a `python -> C -> python` transformations.

qutip/core/_brtensor.pyx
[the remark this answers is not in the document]
AGaliciaMartinez: Thanks, very clear explanation!

Ericgig:
> My biggest concern is with the addition of `**_` to some of the functions, I would argue that it is best to raise an error if the function does not have arguments. this may lead to code that works on jax being incompatible with code that works on just qutip though, so it is something to think about.

We pass unused arguments when we mix different types of coefficients in solver and `QobjEvo`. So we cannot raise an error.

>  I could be an interesting idea for the future to consider adding a similar context manager that removes other dtypes (`CSR` and `Dense`) so that it is clear what the code is doing?

I don't understand what you mean by this... A context manager that remove dtype from those known to the dispatcher?

qutip/core/coefficient.py
AGaliciaMartinez: I see, thanks for clarifying it!

AGaliciaMartinez (approved):
Thanks for the clarifications and the improvements Eric! Looks good to me, I would say it is ready to be merged :D.

AGaliciaMartinez:
> I don't understand what you mean by this... A context manager that remove `dtype` from those known to the dispatcher?

Yes, I know this goes a little bit against the original idea of the dispatcher, where different dtypes could be used in the same operation, but it seems to me that jax and tensorflow dtypes lean more towards being independent. You may want to ensure that no transformation goes under the hood because there is a missing specialization for instance. Technically it is not necessary if all the specializations are implemented but could be something to consider (?).

PR #2016, opened 2022-11-02
Remove shift in coefficient and qobjevo after 

Ericgig:
**Description**
With the update of correlation to v5 in #1997, the need for  `shift` in `QobjEvo` and `Coefficient` disappeared.
`ShiftCoefficient`, `QobjEvo._insert_time_shift` and related tests are removed.

I also restored the correlation tests on windows hoping they are now fast enough to not timeout.

**Related issues or PRs**
#1997

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,7 +77,6 @@ jobs:
           # error prone. See, e.g., https://github.com/qutip/qutip/issues/1202
           - case-name: Windows Latest
             os: windows-latest
-            pytest-extra-options: "-k 'not (test_correlation)'"
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


AGaliciaMartinez:
I am happy to see this removed! Could explain a little bit more in detail why we need to keep the ´_prepare´ method?

Ericgig:
It will be useful for solver feedback.
In v4 we can define Hamiltonian to depend on the state: `H(t, args, psi/rho)`.
We don't have it working in v5, but we will need to restore it before release.
This `_prepare` catch the time and state at the start of `matmul` and `expect` and update the arguments according to feedback.

hodgestar (commented):
This looks like a good simplification.

Could you explain briefly how this interferes with Jax? A simple shift in `t` doesn't seem like it would create much trouble in the derivatives.

Ericgig:
This pr does not have anything to do with JAX...

hodgestar:
> This pr does not have anything to do with JAX...

Ah, that would explain my confusion. :)

hodgestar (approved):

AGaliciaMartinez (approved):

PR #2019, opened 2022-11-16
Run mac tests without mkl

Ericgig:
**Description**
Mac tests have been failing for a while because of MKL. 
Set the mac test to run with openblas, skipping failing tests.
Added `rcond=None` to `linalg.lstsq` to remove a warning.

I also updated versions of github action as some where raising warnings.


hodgestar (approved):
This seems like a good compromise. +1

PR #2032, opened 2022-12-06
Momentum and position default isherm values.

AGaliciaMartinez:
**Description**
The momentum and position did not have the _isherm attribute set to True. Setting it by default to True means hamiltonians created with those operators only will also have the attribute set to True and, for example, eigenvalues computed will, by default, return a real value (and I think be computed faster but I am not completely sure if we take this into account in our algorithms).


AGaliciaMartinez:
Good point with the setter!

PR #2034, opened 2022-12-07
Floquet using v5's Solver interface

Ericgig:
**Description**

Write `fmesolve` as a `Solver` for v5.

I changed most of the function in the module.

The 9 functions to generate the floquet modes, states, etc. have been encapsulated into one class `FloquetBasis`.
This make them easier to use, faster and more precise.

The previous method had you first compute the initial floquet modes, computing the propagator over one period, then using these to compute at other time, recomputing the propagator up to that time. When the modes at multiple times were needed, we could compute a list of them over one period using `floquet_modes_table`, but at least 2 computations over a period were needed. With the and the `Propagator` memoize capacities added in v5, only one computation of the propagator is required. Also with the table computation of v4, if a mode was not included in the table, the closest one was used. `fmesolve ` and `fsesolve` computed 100 slices over the period. It did not used the values in `tlist` and caused artifacts in the result. With `FloquetBasis`, the mode are always computed at the exact time. 

I left the v4's functions with warnings.

The construction of the floquet tensor was done using 2 functions: `floquet_master_equation_rates` which returned 4 tensors and `floquet_master_equation_tensor`. I added `floquet_tensor` which computed the tensor in one call and split `floquet_master_equation_rates` in 4 functions, one for each of the outputs.

Floquet, like `brmesolve`, need a function for the noise power spectrum.  Unlike `brmesolve`, string are not supported, only functions. We should have them match, but supporting string is quite heavy since coefficient are not a great fit.

`fsesolve` have been rewritten without the class interface. As an integrator, the floquet method has no advantage since the propagator much be computed using another method.

`fmesolve` work similarly to other solver. The `e_ops` are always in the laboratory basis. The result states are also in the laboratory basis. The option `"store_floquet_states"` is used to store the states in the floquet basis in `result.floquet_states`.

In v4, there it an option for the environment temperature, but no tests for it. I kept the feature, but it's still untested.
Also all tests for `fmesolve` are either for constant systems or without collapse operator.


hodgestar:
@Ericgig Could you merge dev.major into this branch? Or would you like me to review first and then merge it in?

doc/guide/dynamics/dynamics-floquet.rst
```diff
@@ -104,15 +104,17 @@ In QuTiP we can define this Hamiltonian as follows:
    >>> args = {'w': omega}
    >>> H = [H0, [H1, 'sin(w * t)']]
 
-The :math:`t=0` Floquet modes corresponding to the Hamiltonian :eq:`eq_driven_qubit` can then be calculated using the :func:`qutip.floquet.floquet_modes` function, which returns lists containing the Floquet modes and the quasienergies
+The :math:`t=0` Floquet modes corresponding to the Hamiltonian :eq:`eq_driven_qubit` can then be calculated using the :class:`qutip.FloquetBasis` class, which encapsulated the Floquet modes and the quasienergies
```
hodgestar, line 107:
```suggestion
The :math:`t=0` Floquet modes corresponding to the Hamiltonian :eq:`eq_driven_qubit` can then be calculated using the :class:`qutip.FloquetBasis` class, which encapsulates the Floquet modes and the quasienergies:
```

doc/guide/dynamics/dynamics-floquet.rst
```diff
@@ -166,24 +171,25 @@ Given the Floquet modes at :math:`t=0`, we obtain the Floquet mode at some later
...
 
    >>> psi0 = rand_ket(2)
-   >>> f_coeff = floquet_state_decomposition(f_modes_0, f_energies, psi0)
+   >>> f_coeff = floquet_basis.to_floquet_basis(psi0)
    >>> f_coeff # doctest: +SKIP
    [(-0.645265993068382+0.7304552549315746j),
    (0.15517002114250228-0.1612116102238258j)]
 
-and given this decomposition of the initial state in the Floquet states we can easily evaluate the wavefunction that is the solution to :eq:`eq_driven_qubit` at an arbitrary time :math:`t` using the function :func:`qutip.floquet.floquet_wavefunction_t`
+and given this decomposition of the initial state in the Floquet states we can easily evaluate the wavefunction that is the solution to :eq:`eq_driven_qubit` at an arbitrary time :math:`t` using the function :meth:`FloquetBasis.from_floquet_basis`
```
hodgestar, line 186:
```suggestion
and given this decomposition of the initial state in the Floquet states we can easily evaluate the wavefunction that is the solution to :eq:`eq_driven_qubit` at an arbitrary time :math:`t` using the function :meth:`FloquetBasis.from_floquet_basis`:
```

doc/guide/dynamics/dynamics-floquet.rst
```diff
@@ -194,7 +200,9 @@ The following example illustrates how to use the functions introduced above to c
...
+When evaluating the Floquet states or the wavefunction at many points in time it is useful to pre-compute the Floquet modes for the first period of the driving with the required resolution.
+In QuTiP the function :func:`qutip.floquet.floquet_modes_table` calculates a table of Floquet modes which later can be used together with the function :func:`qutip.floquet.floquet_modes_t_lookup` to efficiently lookup the Floquet mode at an arbitrary time.
+The following example illustrates how the example from the previous section can be solved more efficiently using these functions for pre-computing the Floquet modes.
```
hodgestar, line 205:
```suggestion
The following example illustrates how the example from the previous section can be solved more efficiently using these functions for pre-computing the Floquet modes:
```

doc/guide/dynamics/dynamics-floquet.rst
```diff
@@ -194,7 +200,9 @@ The following example illustrates how to use the functions introduced above to c
 Pre-computing the Floquet modes for one period
 ----------------------------------------------
 
-When evaluating the Floquet states or the wavefunction at many points in time it is useful to pre-compute the Floquet modes for the first period of the driving with the required resolution. In QuTiP the function :func:`qutip.floquet.floquet_modes_table` calculates a table of Floquet modes which later can be used together with the function :func:`qutip.floquet.floquet_modes_t_lookup` to efficiently lookup the Floquet mode at an arbitrary time. The following example illustrates how the example from the previous section can be solved more efficiently using these functions for pre-computing the Floquet modes.
+When evaluating the Floquet states or the wavefunction at many points in time it is useful to pre-compute the Floquet modes for the first period of the driving with the required resolution.
+In QuTiP the function :func:`qutip.floquet.floquet_modes_table` calculates a table of Floquet modes which later can be used together with the function :func:`qutip.floquet.floquet_modes_t_lookup` to efficiently lookup the Floquet mode at an arbitrary time.
```
hodgestar, line 204:
```suggestion
When evaluating the Floquet states or the wavefunction at many points in time it is useful to pre-compute the Floquet modes for the first period of the driving with the required times.
The list of times to pre-compute modes for may be passed to :class:`FloquetBasis` using `precompute=tlist`, and then `:meth:`FloquetBasis.from_floquet_basis` and :meth:`FloquetBasis.to_floquet_basis` can be used to efficiently retrieve the wave function at the pre-computed times.
```

doc/guide/dynamics/dynamics-floquet.rst
```diff
@@ -263,15 +271,15 @@ The noise spectral-density function of the environment is implemented as a Pytho
...
+        return (omega>0) * 0.5 * gamma1 * omega/(2*pi)
 
 The other parameters are similar to the :func:`qutip.mesolve` and :func:`qutip.mcsolve`, and the same format for the return value is used :class:`qutip.solve.solver.Result`. The following example extends the example studied above, and uses :func:`qutip.floquet.fmmesolve` to introduce dissipation into the calculation
 
 .. plot:: guide/scripts/floquet_ex3.py
    :width: 4.0in
    :include-source:
 
-Alternatively, we can let the :func:`qutip.floquet.fmmesolve` function transform the density matrix at each time step back to the computational basis, and calculating the expectation values for us, by using:
+Finally, :func:`qutip.solver.floquet.fmmesolve` consider the ``e_ops`` to always be in the laboratory basis, thus we can simply use:
```
hodgestar, line 282:
```suggestion
Finally, :func:`qutip.solver.floquet.fmmesolve`  always expects the ``e_ops`` to be specified in the laboratory basis (as for other solvers) and we can calculate expectation values using:
```

hodgestar (changes requested):
This is just a review of the documentation changes. Will follow up with a review of the changes to the actual solver.

qutip/solver/floquet.py
```diff
@@ -0,0 +1,937 @@
...
+                memoize += 1
+        else:
+            # Default computation
+            tlist = np.linspace(0, T, 101)
+            memoize = 101
+        self.U = Propagator(H, args=args, options=options, memoize=memoize)
+        for t in tlist:
+            # Do the evolution by steps to save the intermediate results.
+            self.U(t)
+        self.U_T = self.U(self.T)
```
hodgestar, line 102: Do we need to store `U_T` as an attribute?

qutip/solver/floquet.py
```diff
@@ -0,0 +1,937 @@
...
+        # w_th is 0, thermal = 0s
+        A = gamma[0]
+        for k in range(1, kmax + 1):
+            A = _data.add(gamma[k], A)
+            A = _data.add(gamma[-k], A)
+
+    return A
+
+
+def floquet_master_equation_tensor(A):
```
hodgestar, line 411: Let's give all of these internal `floquet_` methods a leading underscore. We can always make them public again later if someone wants them.

qutip/solver/floquet.py
```diff
@@ -0,0 +1,929 @@
...
+        Parameters
+        ----------
+        state0 : :class:`Qobj`
+            Initial state of the evolution.
+
+        t0 : double
+            Initial time of the evolution.
+
+        floquet : bool, optional {False}
+            Whether the initial state in the floquet basis or laboratory basis.
```
hodgestar, line 825:
```suggestion
            Whether the initial state is in the floquet basis or laboratory basis.
```

qutip/solver/floquet.py
```diff
@@ -0,0 +1,929 @@
...
+        state = super().step(t)
+        if not floquet:
+            state = self.floquet_basis.from_floquet_basis(state, t)
+        elif copy:
+            state = state.copy()
+        return state
+
+    def run(self, state0, tlist, *, floquet=False, args=None, e_ops=None):
+        """
+        Do the evolution of the Quantum system.
```
hodgestar, line 866:
```suggestion
        Calculate the evolution of the quantum system.
```

qutip/solver/floquet_bwcomp.py
```diff
@@ -0,0 +1,227 @@
+""" Functions for backward compatibility.
+To remove by v5.1
+"""
```
hodgestar, line 3:
```suggestion
""" Floquet solver compatibility functions that behave like the corresponding functions from QuTiP 4.7.

These functions are indented to be used when porting code from QuTiP 4.7 to QuTiP 5. They are deprecated and will be removed in QuTiP 5.1.
"""
```

hodgestar (changes requested):
Done with reviewing for now. Happy for this to be merged once the changes have been applied.

hodgestar (approved):

Issue #2036, opened 2022-12-08
Qutip fails to install on Python 3.11.0


HGSilveri:
### Bug Description

Installing `qutip` with `pip` on Python 3.11 fails at the "Getting the requirements to build the wheel" stage.

Potentially related to [these changes](https://github.com/pypa/packaging/pull/407) in `packaging`, introduced in the latest update to `22.0`.

### Code to Reproduce the Bug

```shell
pip install qutip --no-cache-dir
```

### Code Output

```shell
Collecting qutip
  Downloading qutip-4.7.0.tar.gz (3.6 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3.6/3.6 MB 18.2 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [20 lines of output]
      Traceback (most recent call last):
        File "/Users/hsilver/opt/miniconda3/envs/py311/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 351, in <module>
          main()
        File "/Users/hsilver/opt/miniconda3/envs/py311/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 333, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/hsilver/opt/miniconda3/envs/py311/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/h9/04p7x4ds0hvfmd80jt8gx3kc0000gn/T/pip-build-env-dzdmg7x6/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 338, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2037, opened 2022-12-08
Fix parsing of package versions in packaging 22.

hodgestar:
**Description**
Fix parsing of package versions in packaging 22.

**Related issues or PRs**
- #2036
- https://github.com/pypa/packaging/issues/631

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -134,7 +134,10 @@ def _determine_version(options):
     with open(version_filename, "r") as version_file:
         version_string = version_file.read().strip()
     version = packaging.version.parse(version_string)
-    if isinstance(version, packaging.version.LegacyVersion):
+    # LegacyVersion was removed in packaging 22, but is still returned by
+    # packing <= 21
+    LegacyVersion = getattr(packaging.version, "LegacyVersion", type(None))
+    if isinstance(version, LegacyVersion):
         raise ValueError("invalid version: " + version_string)
     options['short_version'] = str(version.public)
     options['release'] = not version.is_devrelease
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@
 import collections
 import os
 import pathlib
-import re
 import subprocess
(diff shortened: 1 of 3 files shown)


hodgestar:
This fix will need to be duplicated on dev.major and in all our other repositories that used the qutip setup.py as a template.

Ericgig (approved):
The failing test seems unrelated to this. 
Thank you for fixing.

setup.py
```diff
@@ -134,7 +133,10 @@ def _determine_version(options):
     with open(version_filename, "r") as version_file:
         version_string = version_file.read().strip()
     version = packaging.version.parse(version_string)
-    if isinstance(version, packaging.version.LegacyVersion):
+    # LegacyVersion was removed in packaging 22, but is still returned by
+    # packing <= 21
+    LegacyVersion = getattr(packaging.version, "LegacyVersion", type(None))
+    if isinstance(version, LegacyVersion):
         raise ValueError("invalid version: " + version_string)
```
pradyunsg, line 140: This is a drive-by suggestion for something that doesn't rely on `LegacyVersion` in any way:

```
try:
   version = packaging.version.Version(version_string)
except packaging.version.InvalidVersion:
   raise ValueError("invalid version: " + version_string) from None
```

hodgestar: Thanks! After the discussion in the packaging repo issue, I have been thinking about simplifying things this way, but I need to go look at the history of `Version` first and see which packaging releases it has been around for and whether the behaviour has changed over time so that I don't accidentally mess anything up for users still using older releases of packaging.

hodgestar: We could even just do `version = packaging.version.Version(version_string)` because we don't care about what kind of exception is raised, just that we get an error if we try to build a package with a bad version string.

hodgestar: Thanks!

PR #2040, opened 2022-12-09
Add a default max_step to processors using the smallest gate time

### LINKED ISSUE #2003: processor.run_state and qc.run returning different states
### Bug Description

Perhaps I am not understanding how a qutip device works.

I have been following the example from https://qutip.org/docs/latest/guide/qip/qip-processor.html#qip-processor as I want to add noise to my quantum circuits.

However, I saw that using ```processor.run_state(init_state)``` returns a different state compared to ```qc.run(init_state)``` for certain circuits.

### Code to Reproduce the Bug

```shell
import numpy as np

from qutip import basis, tensor
from qutip.qip.device import LinearSpinChain


num_qubits = 3
init_state = tensor([basis(2,1) for x in range(num_qubits)])
qc = QubitCircuit(num_qubits)
qc.add_gate("CNOT", targets=1, controls=[0])

# Uses a LinearSpinChain processor to obtain the final state
processor = LinearSpinChain(num_qubits)
processor.load_circuit(qc)
processor_state = processor.run_state(init_state).states[-1].tidyup(1.0e-6)

# Obtains state by running the circuit directly
circuit_state = qc.run(init_state)

print(processor_state)
print()
print(circuit_state)
print()

# Calculates the fidelity between both states to show that states are not just differing by a phase
fidelity = np.abs((circuit_state.dag() *  processor_state)[0])

print("Fidelity:", np.abs(fidelity[0]))
```

### Code Output

```shell
Quantum object: dims = [[2, 2, 2], [1, 1, 1]], shape = (8, 1), type = ket
Qobj data =
[[0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [1.]]

Quantum object: dims = [[2, 2, 2], [1, 1, 1]], shape = (8, 1), type = ket
Qobj data =
[[0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [1.]
 [0.]
 [0.]]

Fidelity: [0.]
```

### Expected Behaviour

By just adding an RY gate to the circuit, both methods of obtaining the state now returns the same state with unit fidelity. 

```
import numpy as np

from qutip import basis, tensor
from qutip.qip.device import LinearSpinChain


num_qubits = 3
init_state = tensor([basis(2,1) for x in range(num_qubits)])
qc = QubitCircuit(num_qubits)
qc.add_gate("RX", targets=1,arg_value = np.pi/2)
qc.add_gate("CNOT", targets=1, controls=[0])

# Uses a LinearSpinChain processor to obtain the final state
processor = LinearSpinChain(num_qubits)
processor.load_circuit(qc)
processor_state = processor.run_state(init_state).states[-1].tidyup(1.0e-6)

# Obtains state by running the circuit directly
circuit_state = qc.run(init_state)

print(processor_state)
print()
print(circuit_state)
print()

# Calculates the fidelity between both states to show that states are not just differing by a phase
fidelity = np.abs((circuit_state.dag() *  processor_state)[0])

print("Fidelity:", np.abs(fidelity[0]))

```

### Your Environment

```shell
Current admin team: Alexander Pitchford, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, Boxi Li, Jake Lishman and Simon Cross.

QuTiP Version:      4.7.0
Numpy Version:      1.21.5
Scipy Version:      1.7.3
Cython Version:     0.29.30
Matplotlib Version: 3.5.2
Python Version:     3.7.13
Number of CPUs:     8
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      False
```

### Additional Context

_No response_

BoxiLi:
**Checklist**

- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#Changelog%20Generation) for more information).

**Description**
Add a default max_step to processors using the smallest gate time.

**Related issues or PRs**
fix #2003

### DIFF
--- a/qutip/qip/device/processor.py
+++ b/qutip/qip/device/processor.py
@@ -742,7 +742,7 @@ class Processor(object):
         # set the max step size as 1/2 of the smallest gate time.
         options = kwargs.get("options", Options())
         if options.max_step == 0.:
-            options.max_step = np.min(np.diff(self.get_full_tlist())) / 22
+            options.max_step = np.min(np.diff(self.get_full_tlist())) / 2
         kwargs["options"] = options
         # choose solver:
         if solver == "mesolve":
--- a/qutip/qip/device/processor.py
+++ b/qutip/qip/device/processor.py
@@ -12,6 +12,7 @@ from qutip.qip.operations.gates import expand_operator, globalphase
 from qutip.tensor import tensor
 from qutip.mesolve import mesolve
 from qutip.mcsolve import mcsolve
+from qutip import Options
 from qutip.qip.circuit import QubitCircuit
 from qutip.qip.noise import (
     Noise, RelaxationNoise, DecoherenceNoise,
@@ -738,6 +739,11 @@ class Processor(object):
         else:
             kwargs["c_ops"] = sys_c_ops
 
+        # set the max step size as 1/2 of the smallest gate time.
+        options = kwargs.get("options", Options())
+        if options.max_step == 0.:
+            options.max_step = np.min(np.diff(self.get_full_tlist())) / 22
+        kwargs["options"] = options
         # choose solver:
         if solver == "mesolve":
             evo_result = mesolve(
--- a/doc/changes/2040.bugfix
+++ b/doc/changes/2040.bugfix
@@ -1,0 +1,1 @@
+Add a default max_step to processors
--- a/qutip/tests/test_device.py
+++ b/qutip/tests/test_device.py
@@ -91,7 +91,7 @@ def test_numerical_evolution(
         init_state = qutip.tensor(extra, state)
     else:
         init_state = state
-    options = qutip.Options(store_final_state=True, nsteps=50_000)
+    options = qutip.Options(store_final_state=True, nsteps=100_000)
     result = device.run_state(init_state=init_state,
                               analytical=False,
                               options=options)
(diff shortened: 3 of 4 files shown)


BoxiLi:
Ok, although adding a default `max_step` using the smallest gate time solves the issue. It also significantly increases the computation time in some cases. In particular, for DispersiveCavityQED, the two qubit gate time is ~1000 while the single qubit gate takes only ~1 time unit. Using the `max_step=1` drastically increases the number of time steps needed.

qutip/qip/device/processor.py
```diff
@@ -738,6 +739,11 @@ def run_state(self, init_state=None, analytical=False, states=None,
         else:
             kwargs["c_ops"] = sys_c_ops
 
+        # set the max step size as 1/2 of the smallest gate time.
+        options = kwargs.get("options", Options())
+        if options.max_step == 0.:
+            options.max_step = np.min(np.diff(self.get_full_tlist())) / 22
```
hodgestar, line 745: Why 22? :)

hodgestar (changes requested):
The code looks great, I would just like a comment explaining the significance of the number 22. :)

qutip/qip/device/processor.py
BoxiLi: Ah, that was a mistake, it should just be `/2`. But still, this does not fully solve the problem above.

hodgestar:
@Ericgig I am wondering whether in v5.1 we should add a way for users to specify the timescale of a QobjEvo. At the moment if one knows the timescale when calling the solver one can just set max step, but in code like Boxi's where one is handling QobjEvos or pulses that were created somewhere else, it can be a lot of extra work to determine what the max_step should be, and it would be good if the QobjEvo already had that information.

The downside is that one might have to specify the time step when creating the QobjEvo, and that might be too annoying or too easy to forget to do.

hodgestar (approved):
Ah! 2 makes a lot more sense. :)

PR #2041, opened 2022-12-09
Add python 3.11 build using 3.11 from conda-forge.

hodgestar:
**Description**
Add python 3.11 build using 3.11 from conda-forge.

**Related issues or PRs**
- None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     env:
       # Set up wheels matrix.  This is CPython 3.6--3.10 for all OS targets.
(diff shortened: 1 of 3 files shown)


hodgestar:
I'm keen to merge this and make this the last PR for the 4.7.1 release.

Ericgig:
@hodgestar Are you sure this was ready to be merged?
While the tests part look fine, have you checked that the wheels will build properly? Is there no need to tell the tools building the wheels were to find python 3.11? With 3.10 we skipped some build, (win32, etc.), don't we need to do the same for 3.11?

hodgestar:
We do have to check. I will do the checks against #2046 though since I have to check builds there anyway.

PR #2042, opened 2022-12-09
Documentation for v5 mcsolve

Ericgig:
**Description**
- In `SeSolver`,   `MeSolver`,   `McSolver`, only capitalized 2 letters while `BRSolver` and `HEOMSolver` capitalized all abbreviated terms. Changed it use the same capitalization everywhere: `SESolver`,   `MESolver`,   `MCSolver`.

- Update dynamics-master for v5. Only changing a few links is needed. No reference to the class interface is added since it comes before the explanation to time dependent system.

- Updated dynamics-monte: Add description for new results, with photocurrent, average and runs version of expect and state. The class interface was quickly introduced in the `Reusing Hamiltonian Data` section.

### DIFF
--- a/doc/guide/dynamics/dynamics-monte.rst
+++ b/doc/guide/dynamics/dynamics-monte.rst
@@ -4,10 +4,6 @@
 Monte Carlo Solver
 *******************************************
 
-.. plot::
-      :include-source: False
-
-      from qutip.solver.mcsolve import *
 
 .. _monte-intro:
 
@@ -75,6 +71,8 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP, 
 .. plot::
     :context:
 
+    from qutip.solver.mcsolve import MCSolver, mcsolve
+
     times = np.linspace(0.0, 10.0, 200)
     psi0 = tensor(fock(2, 0), fock(10, 5))
     a  = tensor(qeye(2), destroy(10))
@@ -110,7 +108,7 @@ If we want to run 1000 trajectories in the above example, we can simply modify t
 .. plot::
     :context: close-figs
 
-    data = mcsolve(H, psi0, times, [np.sqrt(0.1) * a], [a.dag() * a, sm.dag() * sm], ntraj=1000)
+    data = mcsolve(H, psi0, times, [np.sqrt(0.1) * a], e_ops=[a.dag() * a, sm.dag() * sm], ntraj=1000)
 
 where we have added the keyword argument ``ntraj=1000`` at the end of the inputs.
 Now, the Monte Carlo solver will calculate expectation values for both operators, ``a.dag() * a, sm.dag() * sm`` averaging over 1000 trajectories.
@@ -132,17 +130,22 @@ we can extract the relevant expectation values using:
     expt100 = data.expect_traj_avg(100)
     expt1000 = data.expect_traj_avg(1000)
 
-		plt.figure()
-    plt.plot(times, expt1, label="ntraj=1")
-		plt.plot(times, expt10, label="ntraj=10")
-		plt.plot(times, expt100, label="ntraj=100")
-		plt.plot(times, expt1000, label="ntraj=1000")
+    plt.figure()
+    plt.plot(times, expt1[0], label="ntraj=1")
+    plt.plot(times, expt10[0], label="ntraj=10")
+    plt.plot(times, expt100[0], label="ntraj=100")
+    plt.plot(times, expt1000[0], label="ntraj=1000")
     plt.title('Monte Carlo time evolution')
     plt.xlabel('Time')
     plt.ylabel('Expectation values')
     plt.legend()
     plt.show()
 
+
+Monte Carlo Solver Result
+-------------------------
+
+...
 
 .. _monte-reuse:
 
@@ -196,7 +199,7 @@ In addition to the initial state, one may reuse the Hamiltonian data when adding
     sm = tensor(destroy(2), qeye(10))
 
     H = 2*np.pi*a.dag()*a + 2*np.pi*sm.dag()*sm + 2*np.pi*0.25*(sm*a.dag() + sm.dag()*a)
-    solver = MCSolver(H, c_ops=[np.sqrt(0.1) * a])
+    # solver = MCSolver(H, c_ops=[np.sqrt(0.1) * a])
     data1 = solver.run(psi0, times, e_ops=[a.dag() * a, sm.dag() * sm], ntraj=10)
     data2 = solver.run(psi0, times, e_ops=[a.dag() * a, sm.dag() * sm], ntraj=10)
     data_merged = data1 + data2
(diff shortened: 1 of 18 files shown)


doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -70,12 +71,14 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
 .. plot::
     :context:
 
+    from qutip.solver.mcsolve import MCSolver, mcsolve
```
AGaliciaMartinez, line 74:
```suggestion
    from qutip.solver.mcsolve import mcsolve
```
I would suggest just importing it down when it is necessary.

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -89,44 +92,80 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
...
 
+Monte Carlo Solver Result
+-------------------------
 
-.. _monte-ntraj:
-
-Changing the Number of Trajectories
------------------------------------
+The Monte Carlo solver returns a :class:`qutip.MultitrajResult` object consisting of expectation values and/or states.
+The main difference with :func:`qutip.mesolve`'s :class:`qutip.Result` is that it optionally store the result of each trajectories with the averages.
```
AGaliciaMartinez, line 100:
```suggestion
The main difference with :func:`qutip.mesolve`'s :class:`qutip.Result` is that it optionally stores the result of each trajectory together with their averages.
```

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -89,44 +92,80 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
...
+When trajectories are stored, ``result.runs_expect`` is a list over the expectation operators, trajectories and times in that order.
+The averages are stored in ``result.average_expect`` and the standard derivation of the expectation values in ``result.std_expect``.
+When states are returned, ``result.runs_states`` will be an array of length ``ntraj``, with each element containing an array of ket-type qobjs with the same number of elements as ``times`` and ``result.average_states`` is a list of density matrices for each times.
```
AGaliciaMartinez, line 103:
```suggestion
When the states are returned, ``result.runs_states`` will be an array of length ``ntraj``. Each element contains an array of "Qobj" type ket with the same number of elements as ``times``. ``result.average_states`` is a list of density matrices computed as the average of the states at each time step.
```

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -89,44 +92,80 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
...
+Furthermore, the output will also contain a list of times at which collapse occurred, and which collapse operators did the collapse, in the ``col_times`` and ``col_which`` properties, respectively.
```
AGaliciaMartinez, line 104:
```suggestion
Furthermore, the output will also contain a list of times at which the collapse occurred, and which collapse operators did the collapse. These can be obtained in  ``result.col_times`` and ``result.col_which`` respectively.
```

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -89,44 +92,80 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
...
+    plt.show()
+
+
+.. _monte-ntraj:
+
+Changing the Number of Trajectories
+-----------------------------------
 
-where we have added the keyword argument ``ntraj=1000`` at the end of the inputs.  Now, the Monte Carlo solver will calculate expectation values for both operators, ``a.dag() * a, sm.dag() * sm`` averaging over 1000 trajectories.  Sometimes one is also interested in seeing how the Monte Carlo trajectories converge to the master equation solution by calculating expectation values over a range of trajectory numbers.  If, for example, we want to average over 1, 10, 100, and 1000 trajectories, then we can input this into the solver using:
+As mentioned earlier, by default, the ``mcsolve`` function runs 500 trajectories.
```
AGaliciaMartinez, line 131:
```suggestion
By default, the ``mcsolve`` function runs 500 trajectories.
```
I think the paragraph that mentions mcsolve has by default 500 trajectories was deleted.

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -89,44 +92,80 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
...
+This value was chosen because it gives good accuracy, Monte Carlo errors scale as :math:`1/n` where :math:`n` is the number of trajectories, and simultaneously does not take an excessive amount of time to run.
+However, like many other options in QuTiP you are free to change the number of trajectories to fit your needs.
```
AGaliciaMartinez, line 133:
```suggestion
However, you can change the number of trajectories to fit your needs.
```

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -89,44 +92,80 @@ To illustrate the use of the Monte Carlo evolution of quantum systems in QuTiP,
...
+If we want to run 1000 trajectories in the above example, we can simply modify the call to ``mcsolve`` like:
```
AGaliciaMartinez, line 134:
```suggestion
In order to run 1000 trajectories in the above example, we can simply modify the call to ``mcsolve`` like:
```

AGaliciaMartinez (commented):
Great job with the documentation (and with he Monte Carlo solver!)! I left a few suggestions. Some of these are definitely subjective so feel free to reject them if you do not like them.

The condition for the std is still puzzling me a bit. Is it because numerical error or could there be a hidden bug/wrong assumption somewhere :thinking:.

qutip/solver/result.py
```diff
@@ -916,10 +926,12 @@ def __add__(self, other):
...
             new.e_data = new.average_e_data
 
-            new.std_e_data = {
-                k: np.sqrt(avg_expect2 - abs(avg_expect**2))
-                for k, avg_expect, avg_expect2 in zip(self._raw_ops, avg, avg2)
-            }
+            new.std_e_data = {}
+            for i, key in enumerate(self._raw_ops):
+                std2 = avg2[i] - abs(avg[i]**2)
+                std2[std2 < 0] = 0.
```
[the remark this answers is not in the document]
Ericgig: Floating point calculations are not exact, values that should be 0 can very small negative numbers and raise a warning for the square root. This is not an issue the user can deal with, it will just happen sometime when subtracting large floating point number.
For example, when I compute the std of 1000 measurement of 4/3 I get:
```
>>> N = 1000
>>> v = 4/3
>>> x = [v]*N
>>> x2 = [v**2]*N
>>> (sum(x)/N)**2 - sum(x2) / N
-5.040412531798211e-14
```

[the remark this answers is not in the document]
AGaliciaMartinez: Thanks for the detailed explanation!

AGaliciaMartinez (approved):

PR #2043, opened 2022-12-10
Simplify version parsing.

hodgestar:
**Description**

Calling Version directly has given the strict parsing we require since the Version class was created and the class predates the parse(...) function itself. This simpler code should give the correct behaviour on all versions of packaging since 2014.

**Related issues or PRs**
- See #2037
- See #2036

### DIFF
--- a/doc/changes/2043.misc
+++ b/doc/changes/2043.misc
@@ -1,0 +1,1 @@
+Simplify version parsing by using packaging.version.Version.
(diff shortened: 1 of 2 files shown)


hodgestar:
Merging since this looks straight forward and it would be good to have it for the 4.7.1 release, but would appreciate post-merge review.

PR #2050, opened 2022-12-16
Update `spectrum` to use the data layer

Ericgig:
**Description**
When I updated correlation to use the data layer in #1997. `spectrum` was moved to its own file, but was left mostly unchanged.
This update `spectrum` to use the data layer,


hodgestar (approved):
Looks good!

PR #2056, opened 2023-01-12
Remove duplicate solvers

Ericgig:
**Description**
Remove the previous version of `sesolve`, `mesolve` and `mcsolve` so only one version exist.
Set the import for the class interface of the solver in imported with qutip:
```
import qutip
qutip.SESolver(...)
```

`qutip.settings` now refer to the object, not the file.


qutip/solver/__init__.py
```diff
@@ -1,11 +1,17 @@
 from .result import *
 from .options import *
-from .integrator import *
-from .ode import *
+import qutip.solver.integrator
+from .integrator import IntegratorException
```
hodgestar, line 4: What are the imports of `qutip.solver.integrator` and `qutip.solver.ode` for? I assume they are to make the integrators accessible without directly exposing them in the `qutip.solver` namespace? If that is the case, I think we should make that more explicit and ensure that they're not exposed here at all (e.g. import them as `_integrator` and maybe `del _integrator` immediately after).

hodgestar (commented):
The changes look good to me so far. Are there still a few things left in `solve` or is it entirely gone?

doc/guide/scripts/correlation_ex3.py
```diff
@@ -20,7 +20,7 @@
 
 # calculate the correlation function G1 and normalize with n to obtain g1
 G1 = qutip.correlation_2op_1t(H, rho0, taus, c_ops, a.dag(), a)
-g1 = G1 / np.sqrt(n[0] * n)
+g1 = np.array(G1) / np.sqrt(n[0] * np.array(n))
```
hodgestar, line 23: What made it necessary to convert `G1` and `n` to np.array?

qutip/solver/__init__.py
Ericgig: I want them available in a less direct way, I moved `integrator.py` to the `ode` folder and renamed it `integrator`.
Thus they are available as `qutip.integrator.Integrator...` instead of `qutip.Integrator...`.

doc/guide/scripts/correlation_ex3.py
Ericgig: This is when we made `expect` as an attribute of `Result` instead of a property.

Ericgig:
@hodgestar  `solve` is not entirely gone, there is `stochastic` (working on it but give me a few weeks), `nonmarkov`, `piqs`, `rcsolve`, and `pdpsolve`.
I believe that we could remove `rcsolve` and move it to a notebook.
`pdpsolve` is a variant of `mcsolve` and could also be removed.
`piqs` is not just a solver, but a representation of quantum system and we should go back to it after dimension rework.
`nonmarkov`, I thought this was related to HEOM and could be removed, but it does not seems so... We will need to look more into it.

qutip/solver/__init__.py
hodgestar: Sounds good.

Issue #1112, opened 2019-10-20
confusing scaling in wigner() and plot_wigner()


gsteele13:
The default scaling factor of \sqrt(2) used by wigner() and plot_wigner() is confusing

```
alpha = 3
psi = coherent(N,alpha)
plot_wigner(psi)
```

produced a plot with coherent state that is NOT centered at alpha = 3 (mega confusing)

Does anyone know where this default scaling of $\sqrt{2}$ comes from? 

In particular, I would suggest that the default be changed to g = 1  to make the above code segment not internally self-inconsisitent

I can easily submit a patch if needed

gsteele13:
(also, although wigner() allow you to override the scaling factor, plot_wigner() does not...)

gsteele13:
minor correction: the parameter g in the function call should actually be set to g=2 to get the correct scaling

(also confusing...)

nonhermitian:
It was originally for compatibility with the qotoolbox.

KosukeMizuno:
Hi, since it've been intended for compatibility, just changing the default could cause compatibility issues.
I also frequently use `plot_wigner` but g=2 is proper for me, so if it can be changed, I'm happy.
Thus I added argumens for plotter functions to specify a desired `g` value instead of changing the default.

BeateAsenbeck:
I have come across the same problem. If you set g=2 you will have a different normalization for the Wigner function. W(0,0)=-0.63661977 is the 0,0 point of the single photon, which usually should be -1/pi so here we multiplied it by two. Same goes for g=1 where you divide it by 2 as far as I see. This does not change the normalization of the marginal distributions so all of those Wigner functions with different g are valid phase-space quasi-probabilities. What does change in fact with the factor g is the commutator relation [x,p]=a*i, where if g!=sqrt(2) a!=1. This will in fact change how we connect annihilation, creation, position and momentum operators together. 
In the case g=sqrt(2) we have the definition x=1/sqrt(2)(a^\dagger + a) and p = i/sqrt(2)(a^\dagger - a).
[...]

PR #2057, opened 2023-01-13
add arguments for plot_wigner_fock_distribution

KosukeMizuno:
**Description**

Related to #1112. The scalling of Winger fuction can be changed in `wigner()` (default is `g=sqrt(2)`). But it cannot be specified with `plot_wigner()` or `plot_wigner_fock_distribution()`. For my usage, it should be `g=2`, so it was inconvenient. I modified these two plotter functions to pass `g` (and the other args of `wigner()`).


qutip/visualization.py
```diff
@@ -1162,7 +1175,7 @@ def plot_wigner(rho, xvec=None, yvec=None, method='clenshaw',
 
     fig : a matplotlib Figure instance, optional
         The Figure canvas in which the plot will be drawn.
-
+
```
Ericgig:
```suggestion
```

qutip/visualization.py
```diff
@@ -1180,6 +1193,7 @@ def plot_wigner(rho, xvec=None, yvec=None, method='clenshaw',
 
     if not isinstance(rho, list):
         rhos = [rho]
+
```
Ericgig:
```suggestion
```

Ericgig (approved):
Sorry @KosukeMizuno for letting this fall though the cracks.
The change looks good, thank you for the contribution.

Ericgig:
Since this PR was made, a new function, `anim_wigner`, was added to make animations using `plot_wigner`. 
Changing `plot_wigner` signature created a conflict that was not visible using github conflict resolution tool...
@KosukeMizuno would you be able to fix that conflict?

KosukeMizuno:
oh you were much faster than me. Anyway thanks for approving this change :)

PR #2064, opened 2023-01-29
Description for the qutip-jax related package.

AGaliciaMartinez:
This is a draft for the qutip-jax related project for GSoC 2023.

Based on the ideas discussed internally, I made the focus of the project to be the design of an integrator that can benefit from autodifferentiation. @quantshah, @Ericgig, @hodgestar what are your thoughts? for instance, I am worried that the project is too specific (?).

### DIFF
--- a/doc/development/ideas/jax_integrator.rst
+++ b/doc/development/ideas/jax_integrator.rst
@@ -1,0 +1,63 @@
+JAX integrator
+--------------
+
+QuTiP has recently rewritten its solver module to extend its flexibility and
+allow different ``Integrator`` (ODE solvers). ``Integrator`` in qutip are the
+classes in responsible for the low level computation of the evolution of a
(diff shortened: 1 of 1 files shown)


Ericgig:
@AGaliciaMartinez,
I already made an integrator for jax in qutip/qutip-jax#14 and made and an example for auto-differentiation in `sesolve` using it.
There is still some polishing to do, but half of the deliverable are there.

We should ask the student to use it to makes new tools with it and improve the interface.
Using it for quantum control is impractical since `jax.grad` only support real scalar output and our inputs are also complex.
We need to find an user friendly a way to tell which part of the result are subject to the derivation and how to expand to jacobians, etc.

AGaliciaMartinez:
Hi @Ericgig, thanks for the update! when I was checking the status of the project I did not take a look at the PR. Indeed, with halve the deliverables already finished in the PR this proposal does not make much sense. I am closing this for now.

PR #2066, opened 2023-01-31
Remove example using parfor in the documentation.

Ericgig:
**Description**
We officially removed `parfor` in v5. This remove it's mentions in the using qutip in parallel documentation page.
Also removed it the the namespace of `ipynbtools.py` and `solve.parallel.py` (the later only exist until the stochastics solver are migrated to the class interface).

### DIFF
--- a/doc/guide/guide-parfor.rst
+++ b/doc/guide/guide-parfor.rst
@@ -7,29 +7,13 @@ Parallel computation
 Parallel map and parallel for-loop
 ----------------------------------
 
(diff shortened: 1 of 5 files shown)


doc/guide/guide-parfor.rst
```diff
@@ -142,6 +108,8 @@ The :func:`qutip.parallel.parallel_map` function also supports progressbar, usin
    100.0%. Run time:  25.15s. Est. time left: 00:00:00:00
    Total run time:  28.91s
 
+There is a function called :func:`qutip.solver.parallel.serial_map` that works as a non-parallel drop-in replacement for :func:`qutip.solver.parallel.parallel_map`, which allows easy switching between serial and parallel computation.
+Qutip as also the function :func:`qutip.solver.parallel.loky_map` as another drop-in replacement. It use the `loky` module instead of `multiprocessing` to run in parallel.
```
hodgestar, line 112:
```suggestion
Qutip also has the function :func:`qutip.solver.parallel.loky_map` as another drop-in replacement. It use the `loky` module instead of `multiprocessing` to run in parallel.
```

hodgestar (commented):
One small comment, but looks good!

hodgestar (approved):

PR #2067, opened 2023-01-31
Add rerun for simdiag tests

Ericgig:
**Description**
Tests in `simdiag` sometime fail with one or two values over the tolerance. This cause automated tests to fail for reason unrelated to the PR. This add the flaky mark to rerun these and increase some tolerances.


hodgestar (approved):
I would prefer these not to be flaky and require re-runs, but this is better than tests randomly failing.

PR #2068, opened 2023-02-01
Remove `expect_traj_avg`

Ericgig:
**Description**
Remove the function `MCResult.expect_traj_avg`. It was added to replace list input to mcsolve `ntraj` argument, but as discussed in qutip/qutip-tutorials#58, it was not a useful implementation of the feature.

Also improved the warning message in floquet's backward compatibility functions to include the new way and changed warning to error for any functions that would not work as a drop-in replacement.


qutip/solver/mcsolve.py
```diff
@@ -148,8 +148,7 @@ def mcsolve(H, state, tlist, c_ops=(), e_ops=None, ntraj=500, *,
 
     if isinstance(ntraj, list):
```
hodgestar, line 149: I would change this to:
```
if isinstance(ntraj, (list, tuple)):
    raise TypeError(
        "ntraj must be an integer. A list of numbers is not longer supported."
    )
```
This covers both lists and tuples and explains what ntraj should be.

Apologies for not making this a suggestion -- GitHub seems not to allow suggestions if there are deleted lines, which is a bit odd.

qutip/solver/floquet_bwcomp.py
```diff
@@ -126,13 +134,9 @@ def floquet_wavefunction(f_modes_t, f_energies, f_coeff, t):
...
     """
-    warnings.warn(FutureWarning("`floquet_wavefunction` is deprecated."))
-    return sum(
-        [
-            f_modes_t[i] * np.exp(-1j * f_energies[i] * t) * f_coeff[i]
-            for i in np.arange(1, len(f_energies))
-        ],
-        start=f_modes_t[0] * np.exp(-1j * f_energies[0] * t) * f_coeff[0],
+    raise NotImplementedError(
+        "`floquet_wavefunction` is deprecated. "
```
hodgestar, line 138:
```suggestion
        "`floquet_wavefunction` is not longer provided. "
```
Deprecated implies it is still usable right now, which it is not.

qutip/solver/floquet_bwcomp.py
```diff
@@ -82,8 +85,7 @@ def floquet_modes_t_lookup(f_modes_table_t, t, T):
     f_modes_table_t = fbasis = FloquetBasis(...)
     f_mode_t = f_modes_table_t.mode(t)
     """
-    warnings.warn(FutureWarning("`floquet_modes_t_lookup` is deprecated."))
-    return f_modes_table_t.mode(t)
+    raise NotImplementedError("`floquet_modes_t_lookup` is deprecated.")
```
hodgestar, line 88:
```suggestion
    raise NotImplementedError(
        "`floquet_modes_t_lookup` is no longer provided. Use `FloquetBasis` instead."
    )
```

hodgestar (changes requested):
I suggested some minor changes, but this looks good.

hodgestar (approved):

PR #2069, opened 2023-02-01
Update the release action to push to testpypi and allow dev release.

Ericgig:
**Description**
- Added a job that update to testpypi when the "test" is entered as the branch name.
- Removed the check that stopped release with `dev` in the version number.
- Added a token for testpypi.

While it's a draft PR, the action will only print, not run the upload command.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,14 +12,15 @@ on:
 
 
 jobs:
-  # The deploy_test job is part of the test of whether we should deploy to PyPI.
-  # The job will succeed if either the confirmation reference is empty or if the
-  # confirmation is the selected branch or tag name.  It will fail if it is
-  # nonempty and does not match.  All later jobs depend on this job, so that
-  # they will be immediately cancelled if the confirmation is bad.  The
-  # dependency is currently necessary (2021-03) because GitHub Actions does not
-  # have a simpler method of cancelling an entire workflow---the normal use-case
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -105,7 +107,7 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     env:
       # Set up wheels matrix.  This is CPython 3.6--3.10 for all OS targets.
-      CIBW_BUILD: "cp3{6,7,8,9,10,11}-*"
+      CIBW_BUILD: "cp3{8,9,10,11}-*"
```
hodgestar, line 110: The comment above should be updated too.

.github/workflows/build.yml
```diff
@@ -162,17 +164,55 @@ jobs:
...
       - name: Upload sdist and wheels to PyPI
         run: |
           python -m pip install "twine"
-          python -m twine upload --verbose wheels/*.whl sdist/*.tar.gz
+          echo python -m twine upload --verbose wheels/*.whl sdist/*.tar.gz
+
+
+  deploy_testpypi:
+    name: "Deploy to TestPyPI if desired"
+    # The confirmation is tested explicitly in `deploy_test`, so we know it is
```
hodgestar, line 185: This comment and the similar one in `deploy` should be updated to match the new logic (i.e. the ref can now be valid non-test, test, or empty.

hodgestar (approved):

PR #2070, opened 2023-02-02
Nonmarkovian mcsolve

pmenczel:
**Description**
Adds support for Monte-Carlo simulations of master equations with possibly negative rates. The method implemented here is described in [arXiv:2209.08958](https://arxiv.org/abs/2209.08958) [quant-ph]. A simple example can be found in [this notebook](https://github.com/pmenczel/nm-mcsolve-example/blob/master/nonmarkovian_mcsolver.ipynb).

Based on code provided by B. Donvil.

**Todo-List**
- [X] Create users guide
    - [X] Add example notebook to qutip-tutorial and link it in users guide ([PR in qutip-tutorials repo](https://github.com/qutip/qutip-tutorials/pull/59))
- [X] Store the trajectory-average of the influence martingale in the `Result` object
    - Use this as end condition, or to warn users about bad convergence?  
    (Unclear how effective that would be. Will not be implemented for now)
- [X] Make associated `nm_mcsolve` function (name up for debate?)
- [X] Write docstring comments
- [X] Write tests
- [X] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#Changelog%20Generation) for more information).


qutip/solver/nm_mcsolve.py
```diff
@@ -0,0 +1,103 @@
...
+    # collapses is a list of (t_k, i_k)
+    def _discrete_martingale(self, collapses):
+        o_r = self.ops_and_rates
+        factors = [o_r[ik][1](tk) / (o_r[ik][1](tk) + self._rate_shift(tk))
+                   for tk, ik in collapses]
+        return np.prod(factors)
+
+    # Override "run" to initialize continuous part of martingale evolution
+    def run(self, state, tlist, *args, **kwargs):
+        self._mu_c = self._continuous_martingale(tlist)
```
Ericgig, line 87: We allow `step` to be used with `MCSolver`. While I don't expect it will get much use, it should still work. By pre-computing `_mu_c` here, `step` will fail.

Ericgig (commented):
Thank you for the PR.
You are the first one not on the team that create a new solver using the class interface, hopefully it was not too confusing. 

Could you describe in more details when to use this solver? It seems to be able to produce states which trace is not one, which seems strange.

I am not too sure about adding the time parameter to `_restore_state`. In floquet, the state produced need a basis change depending on time, but it was put in the `Result` so the state before and after the transformation can be stored. Maybe something like this would be useful, it would allow the `mu` to be saved in results.

Do you plan to make an associated function `nm_mcsolve`?

Please add test for this solver.
We merged `dev.major` in to `master` a few weeks ago. So please make the PR to master directly.

pmenczel:
> Could you describe in more details when to use this solver? It seems to be able to produce states which trace is not one, which seems strange.

Thank you for your detailed comments! We are here at RIKEN working together with the authors of the paper linked above to try and add this functionality to qutip. So far, this PR is just a starting point for the collaboration, but proper documentation and examples are high on our priority list.

To add some very quick explanation here: the master equation describing a general completely positive (but not necessarily CP-divisible) map can involve rate functions that become negative at some times. The usual Monte-Carlo algorithm cannot be applied in this case, but Donvil et al. describe [here](https://www.nature.com/articles/s41467-022-31533-8) and [here](https://arxiv.org/abs/2209.08958) a workaround where states are weighted with an "influence martingale". Including this weight, the involved states are indeed not normalized; however, the state that results from an average over sufficiently many trajectories is normalized. (Checking *how well* the average is normalized could maybe be used as an indicator of how good the convergence is, as noted at the top.)

> I am not too sure about adding the time parameter to `_restore_state`. In floquet, the state produced need a basis change depending on time, but it was put in the `Result` so the state before and after the transformation can be stored. Maybe something like this would be useful, it would allow the `mu` to be saved in results.

Putting this calculation into the `Result` instead does sound reasonable to me. I am unsure how to achieve that in practice though, since the result class used for a single trajectory is hard-coded in mcsolve:
```
result = Result(e_ops, {**self.options, "normalize_output": False})
```
Note that in principle, we could also get the time from the integrator in _restore_state without adding it as a function parameter. Tagging @hodgestar who suggested adding the time parameter to `_restore_state`.

qutip/solver/nm_mcsolve.py
```diff
@@ -0,0 +1,103 @@
...
+        a = max(np.real(w))
+        L = scipy.linalg.sqrtm(a * np.eye(len(op)) - op)  # new Lindblad operator
+        return a, L
+
+    # Shifts all rate function by the function rate_shift
+    # Returns c_i = L_i * sqrt(gamma_i) as QobjEvo objects
+    def _compute_paired_c_ops(self):
+        c_ops = []
+        for f in self.ops_and_rates:
+            def sqrt_gamma(t, f=f):
```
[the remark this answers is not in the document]
pmenczel: I was surprised and happy that pickle didn't complain. However, it turns out that I never even tested it in parallel. My assumption that parallel would be the default was so strong that I never even checked...
I removed the inner function and it also works in parallel now.

qutip/solver/nm_mcsolve.py
pmenczel: Thanks for the catch, that should be fixed now.

qutip/solver/nm_mcsolve.py
```diff
@@ -0,0 +1,103 @@
...
+
+        a_candidate = op[0, 0]
+        if np.allclose(a_candidate * np.eye(len(op)), op,
+                       **tolerance_settings):
+            return np.real(a_candidate), None
+
+        w, _ = np.linalg.eig(op)
+        a = max(np.real(w))
+        L = scipy.linalg.sqrtm(a * np.eye(len(op)) - op)  # new Lindblad operator
+        return a, L
```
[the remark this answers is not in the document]
pmenczel: I see your point, but I don't know how this calculation could be done using the `Qobj` API alone

qutip/solver/nm_mcsolve.py
```diff
@@ -0,0 +1,148 @@
...
+
+        a_candidate = op[0, 0]
+        if np.allclose(a_candidate * np.eye(len(op)), op,
+                       **tolerance_settings):
+            return np.real(a_candidate), None
+
+        w, _ = np.linalg.eig(op)
+        a = max(np.real(w))
+        L = scipy.linalg.sqrtm(a * np.eye(len(op)) - op)  # new Lindblad operator
+        return a, Qobj(L)
```
Ericgig, line 75: Here how it can be written using `Qobj`:
```suggestion
        op = sum((L.dag() * L) for L, _ in ops_and_rates)

        a_candidate = op.tr() / op.shape[0]
        with qt.CoreOptions(atol=self.options["completeness_atol"]):
            if op == a_candidate * qt.qeye(op.dims[0]):
                return np.real(a_candidate), None

        a = max(np.real(op.eigenenergies()))
        L = (a * qt.qeye(op.dims[0]) - op).sqrtm()
        return a, qt.Qobj(L)
```
This need to give a default float value to `completeness_atol` and do `self.options = options` in `__init__`.
(merge the latest changes in master if it raise an error.)

Ericgig:
I would propose overwriting `_run_one_traj`, it give you control over what you feed to the results and you could overwrite the call of `_restore_state` for this class only.

pmenczel:
We have made some progress! But I also have some questions for the experts here.
* We have added a Users Guide for the new solver, as a new subsection of the "Monte Carlo Solver" guide. On my machine, I can build the documentation without problems (there are some warnings, but unrelated to this pull request). However, the check here failed and I do not see a useful error message. Does someone have an idea what is going on?
* We have two example notebooks to add to qutip-notebooks. How to proceed in practice: do we open a separate pull request in qutip-notebooks?
* The _run_one_traj function now exists 3 times, with practically identical code:
https://github.com/qutip/qutip/blob/master/qutip/solver/multitraj.py#L197
https://github.com/qutip/qutip/blob/master/qutip/solver/mcsolve.py#L427
https://github.com/pmenczel/qutip/blob/nonmarkovian-mcsolve/qutip/solver/nm_mcsolve.py#L160
I've noticed a small bug that has gone unnoticed because of this: "step" [here](https://github.com/qutip/qutip/blob/master/qutip/solver/multitraj.py#L206) should be "integrate", I think.

    Maybe it would be a good idea to make the type of `Result` that is used for single trajectories generic in multi-trajectory solvers (just like the `resultclass` is already generic in solvers in general)? I have implemented that [here](https://github.com/pmenczel/qutip/commit/4a25709f90ed7c6dc2153011ee5a8110f3957c95) and could merge that commit into this PR if you think that's a good idea.

Side note: due to a WSL bug, some of my commit timestamps are messed up. Because of that, some commits seem to be shown out of order here...

Ericgig:
>     * We have added a Users Guide for the new solver, as a new subsection of the "Monte Carlo Solver" guide. On my machine, I can build the documentation without problems (there are some warnings, but unrelated to this pull request). However, the check here failed and I do not see a useful error message. Does someone have an idea what is going on?

The error in the documentation (HTTP Error 403) is unrelated the the actual build. It happen from time to time randomly... you can ignore it.
However the documentation build will fail on warnings.

>     * We have two example notebooks to add to qutip-notebooks. How to proceed in practice: do we open a separate pull request in qutip-notebooks?
 
qutip-notebook is no longer used, we migrated to qutip-tutorial. Please open a new PR there.

>     * The _run_one_traj function now exists 3 times, with practically identical code:
>       https://github.com/qutip/qutip/blob/master/qutip/solver/multitraj.py#L197
>       https://github.com/qutip/qutip/blob/master/qutip/solver/mcsolve.py#L427
>       https://github.com/pmenczel/qutip/blob/nonmarkovian-mcsolve/qutip/solver/nm_mcsolve.py#L160
>       Maybe it would be a good idea to make the type of `Result` that is used for single trajectories generic in multi-trajectory solvers (just like the `resultclass` is already generic in solvers in general)? I have implemented that [here](https://github.com/pmenczel/qutip/commit/4a25709f90ed7c6dc2153011ee5a8110f3957c95) and could merge that commit into this PR if you think that's a good idea.

There is a fourth `_run_one_traj` coming for the stochastic solver with also almost the same code...
Yes, some cleaning is needed and having a swap-able trajectory result class would be nice.
I looked at your implementation and it look fine.
I would just rename `trajectoryclass` to `trajectory_result_class` or something similar, otherwise it make me think it refer to the removed `TrajectorySolver`.


>       I've noticed a small bug that has gone unnoticed because of this: "step" [here](https://github.com/qutip/qutip/blob/master/qutip/solver/multitraj.py#L206) should be "integrate", I think. 
`step` was renamed to `integrate` a while ago. Good catch!

qutip/solver/result.py
```diff
@@ -970,3 +980,110 @@ def runs_photocurrent(self):
...
+        self._solver = solver
+        self.trace = []
+
+    # This gets called during the Monte-Carlo simulation of the associated
+    # completely positive master equation. To obtain the state of the actual
+    # system, we simply multiply the provided state with the current martingale
+    # before storing it / computing expectation values.
+    def add(self, t, state):
+        if isket(state):
+            state = ket2dm(state)
```
hodgestar, line 1004: Is converting a ket state to a dm the right thing to do here? If MCSolve, wouldn't it be better to rather do something like:
```
if isket(state):
    state = state * np.sqrt(mu)
else:
   state = state * mu
```
so that the kind of state is preserved in the results?

pmenczel: Unfortunately, that would not work. The method only works with density matrices, not with ket states. The reason is that mu may be negative and in that case, `ket2dm(ket * sqrt(mu))` is not the same as `mu * ket2dm(ket)`.

I will add a note in the docs to document this behavior.

qutip/core/coefficient.py
```diff
@@ -191,6 +191,18 @@ def conj(coeff):
     return ConjCoefficient(coeff)
 
 
+def const(value):
+    """ return a Coefficient with a constant value.
+    """
+    return ConstantCoefficient(value)
+
+
+# XXX: HACKITY HACK
```
hodgestar, line 200: @Ericgig Is this a nice way to expose these extra coefficient constructors at the top level of QuTiP, or too hacky? If you like it, I can add tests.

hodgestar:
@pmenczel I've worked on this quite a bit, mostly focusing on putting initial tests in place, supporting the full set of coefficient functions supported by the other solvers for the rates, shifting the martingale calculations into the integrator and moving the rate shift calculations to Cython to avoid the overhead of Python function calls and loops.

@Ericgig I ended up doing some odd things because QobjEvo doesn't seem like natural fit for these rate functions and coefficients are a bit awkward to use directly inside solvers.

@pmenczel & @Ericgig Could you take a look at everything again and let me know what you think?

Some of the bigger todo items I have in mind at the moment:

- How do we tell that the negative rate case is really working correctly? I added a test but it doesn't match mesolve well and I'm unhappy with that.
- Can we strengthen any of the existing tests with more assertions?
- Now that the discrete and continuous martingale calculations are mostly handled when collapses happen, we could expose them on the result along with `col_which` and `col_time`. Should we?
- Are the new martingale and rate shift calculations correct?
- The way the shared coefficients are used inside `RateSet` and exposed to `MCSolve` required a lot of careful thinking, did I get the design of that right?
- How do we test whether this works when using `parallel_map`? Being able to run in parallel is a key strength of the Monte Carlo method, so it would be good to have tests that show that that works.
- Can we drop the number of trajectories in the docs to make the doc build faster?
- Using the SciPy integrator is a bit awkward. Is there anything we can replace it with? If not, should we expose more of its options? I exposed the `limit` option because I kept having the integration fail during testing.

There is also a bunch of general cleaning up still to do, but lets try get the big picture happy first.

qutip/core/coefficient.py
Ericgig: I have more a question about why they should be exposed.
Coefficient are seldom manipulated directly and the name of these functions is too common to make available.
So having them available only when manually imported seems good enough.
The method `Coefficient.conj()` is giving the same effect and is already available.

qutip/solver/nm_mcsolve.py
```diff
@@ -0,0 +1,379 @@
...
+                      seed=seeds, target_tol=target_tol, timeout=timeout)
+    return result
+
+
+def _parse_op_and_rate(op, rate, **kw):
+    """ Sanity check the op and convert rates to coefficients. """
+    if not isinstance(op, Qobj):
+        raise ValueError("NonMarkovianMCSolver ops must be of type Qobj")
+    if isinstance(rate, numbers.Number):
+        rate = coefficient.const(rate)
```
Ericgig, line 175: I would propose using
`rate = ConstantCoefficient(rate)`
or adding `ConstantCoefficient` to the `coefficient_builders` dict and not having a special case here.

qutip/core/coefficient.py
hodgestar: I agree that `conj` is already easily available. `norm` seems fairly useful and at the moment it requires knowing to import it from `qutip.core.coefficient`. `const` isn't that useful for `QobjEvo` but is very useful if one has to work with coefficients more directly.

Since this seems a little bit contentious, let me just revert this part for now and import `const` manually where I need it for `nm_mcsolve`.

qutip/solver/nm_mcsolve.py
hodgestar: I didn't add it as a special case because I was a bit worried about constants accidentally being converted to `Coefficient`, which is not what one wants when working with `QobjEvo`.

Let me remove `const` entirely for now and just use `ConstantCoefficient` directly as you suggest.

hodgestar: Removed.

hodgestar:
@pmenczel I updated the PR again to simplify the new Cython code a lot and shift parts to Python. I'm much happier with this version and hopefully it makes it easier for you to understand too.

hodgestar (changes requested):
Thank you for fixing my bug and for reorganizing the caching of the martingale. I still wish we could find a simpler caching set up, but I'm happy for it to stay as is for now.

I think if we can make the test `rtol` a bit better, we can perhaps merge this and finish the tutorial without major revisions.

qutip/tests/solver/test_nm_mcsolve.py
```diff
@@ -0,0 +1,582 @@
...
+        args=args, e_ops=e_ops,
+    )
+
+    np.testing.assert_allclose(mc_result.trace, [1.] * len(times), rtol=0.25)
+    np.testing.assert_allclose(
+        me_result.expect[0], mc_result.expect[0], rtol=0.25,
+    )
+    np.testing.assert_allclose(
+        me_result.expect[1], mc_result.expect[1], rtol=0.25,
+    )
```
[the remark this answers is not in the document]
pmenczel: As far as I can tell, the tolerance is the same as in the `mcsolve` tests, compare e.g. [here](https://github.com/qutip/qutip/blob/master/qutip/tests/solver/test_mcsolve.py#L111).

qutip/tests/solver/test_nm_mcsolve.py
```diff
@@ -62,6 +62,56 @@ def test_rough_agreement_with_mesolve_for_negative_rates():
...
+        "sin(t)",
+    ]
+    args = [
+        None,
+        {'constant': 1},
+        None,
+    ]
+    for rate, arg in zip(rates, args):
+        solver = NonMarkovianMCSolver(H, [(L, rate)], args=arg)
+        pickle.dumps(solver)
```
hodgestar, line 112: We should also try to load the pickled solver here.

qutip/tests/solver/test_nm_mcsolve.py
```diff
@@ -62,6 +62,56 @@ def test_rough_agreement_with_mesolve_for_negative_rates():
...
+    # test using all combinations of the following operators
+    from itertools import combinations
+    all_ops_and_rates = [
+        (qutip.sigmap(), 1),
+        (qutip.sigmam(), 1),
+        (qutip.sigmaz(), 1),
+        (1j * qutip.qeye(2), 1),
+    ]
+    # empty ops_and_rates not allowed
+    for n in range(1, len(all_ops_and_rates) + 1):
```
hodgestar, line 83: Nice way to test all the combinations!

qutip/tests/solver/test_nm_mcsolve.py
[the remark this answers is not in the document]
hodgestar: Not great, but let's leave it as it is for now if the test is no worse than the one for `mcsolve`.

hodgestar (changes requested):
It largely looks good to me now.

Could you fix the conflicts and check that the pickle can be loaded in the pickle test and then I think we can merge.

hodgestar (approved):
Approved!

Will wait for the CI tests to run and then merge if they pass.

hodgestar (approved):

PR #2071, opened 2023-02-02
Simplify initiating options in solver

Ericgig:
**Description**
Options in solver are usually set in `Solver.__init__`, but solver that needed options at the building of the rhs needed to initiate them themselves and it was more complex than needed.
Allow to use `self.options = options` at anytime without having to set `_options` and filter `None` first.

### DIFF
--- a/qutip/solver/mcsolve.py
+++ b/qutip/solver/mcsolve.py
@@ -444,7 +444,7 @@ class MCSolver(MultiTrajSolver):
 
     def _get_integrator(self):
         _time_start = time()
-        method = self._options["method"]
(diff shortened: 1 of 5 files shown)


qutip/solver/mcsolve.py
```diff
@@ -461,7 +461,7 @@ def _get_integrator(self):
     @property
     def options(self):
         """
-        Options for bloch redfield solver:
+        Options for monte carlo solver:
```
hodgestar, line 464: Whoops. :)

hodgestar (approved):

Issue #1722, opened 2021-11-22
Better operator dimension parameterisation


pierreguilmin:
**Is your feature request related to a problem? Please describe.**
The dimension argument of qutip operators are quite inconsistent in the **naming**, **type** and **description**. Here are a few examples picked from the [functions doc page](https://qutip.org/docs/latest/apidoc/functions.html):
- `basis` > `dimensions: int or list of ints` _Number of Fock states in Hilbert space. If a list, then the resultant object will be a tensor product over spaces with those dimensions._
- `enr_fock` > `dims: list` _A list of the dimensions of each subsystem of a composite quantum system._
- `enr_state_dictionaries` > `dims: list` _A list with the number of states in each sub-system._
- `phase_basis` > `N: int` _Number of basis vectors in Hilbert space._
- `projection` > `N: int` _Number of basis states in Hilbert space._
- `momentum` > `N: int` _Number of Fock states in Hilbert space._
- `num` > `N: int` _The dimension of the Hilbert space._
- `qeye` > `dimensions: (int) or (list of int) or (list of list of int)` _Dimension of Hilbert space. If provided as a list of ints, then the dimension is the product over this list, but the dims property of the new Qobj are set to this list. This can produce either oper or super depending on the passed dimensions._

Is there a reason for this apparent disorder? I'd be happy to propose a PR.

**Describe the solution you'd like**
Harmonise the naming, type and description of the operators dimension argument.

**Describe alternatives you've considered**
Even if not all operators are meant to work on composite Hilbert spaces, harmonising the naming and docstring would already be helpful.

**Additional context**
Cc @francois-marie for indicating the issue.

PierreGuilmin:
We can maybe make a distinction between operators that act on any Hilbert space, including composite Hilbert space, (like `qeye`) vs operators that act on a specific Hilbert space (like `momentum` or `num`). The first would take a `dimensions: int or list of ints` as argument, and the second a `N: int` argument.

Ericgig:
Thank you for pointing, these functions were added and modified over years and it would be a good idea to unify the documentation. 

However we are short handed, so it will probably wait until we rework the dimension representation in a future major release
( #1476). We are open to help until then.

soma2000-lang:
@Ericgig can i work on this issue

Ericgig:
Hi @soma2000-lang,
Yes, it would be a great help.

PriyanshuPansari:
Is someone working on it? if not can I work on this?

[...]

PR #2075, opened 2023-02-04
Better operator dimension parameterisation

PriyanshuPansari:
Removed the inconsistency in the dimension argument of Qutip operators in the naming, type, and description.
Proposed generalization - use N to denote the dimension of Hilbert space if it's int or denote a list with the number of states in each sub-system if it's a list.

fixes https://github.com/qutip/qutip/issues/1722

Used the following conventions-
in States.py-
dimensions : int or list of ints
      Number of Fock states in Hilbert space. If a list, then the resultant
      object will be a tensor product over spaces with those dimensions.

N: int 
      Number of Fock states in Hilbert space.

dims : list or array
     The quantum state dimensions array, as it would appear in a Qobj.

dim: list
      A list with the number of states in each sub-system.

in Opearator.py- 

dimensions : (int) or (list of int) or (list of list of int)
        Dimension of Hilbert space. If provided as a list of ints, then the
        dimension is the product over this list, but the ``dims`` property of
        the new Qobj are set to this list.  This can produce either `oper` or
        `super` depending on the passed `dimensions`.
N: int 
      Number of Fock states in Hilbert space.

### DIFF
--- a/qutip/core/states.py
+++ b/qutip/core/states.py
@@ -48,7 +48,7 @@ def basis(dimensions, n=None, offset=None, *, dtype=_data.Dense):
     Parameters
     ----------
     dimensions : int or list of ints
-        Number of Fock states in Hilbert space.  If a list, then the resultant
+        Number of Fock states in Hilbert space. If a list, then the resultant
         object will be a tensor product over spaces with those dimensions.
 
     n : int or list of ints, optional (default 0 for all dimensions)
@@ -389,7 +389,7 @@ def thermal_dm(N, n, method='operator', *, dtype=_data.CSR):
     Parameters
     ----------
     N : int
-        Number of basis states in Hilbert space.
+        Number of Fock states in Hilbert space.
 
     n : float
         Expectation value for number of particles in thermal state.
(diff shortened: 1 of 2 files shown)


PriyanshuPansari:
@Ericgig

qutip/core/states.py
```diff
@@ -955,9 +958,8 @@ def enr_fock(dims, excitations, state, *, dtype=_data.Dense):
 
     Parameters
     ----------
-    dims : list
-        A list of the dimensions of each subsystem of a composite quantum
-        system.
+    dims : list or array
+        The quantum state dimensions array, as it would appear in a Qobj.
```
Ericgig, line 962: All `enr_...` functions take the same `dims` input. It's doc-string should match between the enr_operators and enr_states.

Ericgig (commented):
Please add a towncrier entry in `qutip/doc/change/2075.misc`.
Also, do not close this pull request and open a new one. You can just push new commits to your branch and this PR will be updated.

Ericgig:
Thank you again @PriyanshuPansari, but this is no longer relevant.

PR #2077, opened 2023-02-06
expect as a array

Ericgig:
**Description**
Change back `Result.expect`to be a `list of np.array` as it was in v4.
`Result.e_data` is still a `dict of list`.

Expectation result for multi trajectories solver use the same output types.

`np.array` being called on each expectation values individually should allow function `e_ops` to return other types than scalar. However this may fails for multi trajectories result when averaging expectation between trajectories or using `target_tol`.

### DIFF
--- a/qutip/solver/result.py
+++ b/qutip/solver/result.py
@@ -126,9 +126,10 @@ class Result(_BaseResult):
         operator given the state at ``t``. If the element is a function, ``f``,
         the value recorded is ``f(t, state)``.
 
-        The values are recorded in the ``.expect`` attribute of this result
-        object. ``.expect`` is a list, where each item contains the values
-        of the corresponding ``e_op``.
+        The values are recorded in the ``e_data`` and ``expect`` attributes of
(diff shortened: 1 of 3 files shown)


qutip/solver/result.py
```diff
@@ -492,17 +492,20 @@ def _add_first_traj(self, trajectory):
             state = trajectory.final_state
             self._sum_final_states = qzero(state.dims[0])
 
-        self._sum_expect = np.array(trajectory.expect) * 0.
-        self._sum2_expect = np.array(trajectory.expect) * 0.
+        self._sum_expect = [expect * 0 for expect in trajectory.expect]
+        self._sum2_expect = [expect * 0 for expect in trajectory.expect]
```
hodgestar, line 496:
```suggestion
        self._sum_expect = [np.zeros_like(expect) for expect in trajectory.expect]
        self._sum2_expect = [np.zeros_like(expect) for expect in trajectory.expect]
```
Using `zeros_like` seems clearer to me.

hodgestar (approved):

PR #2079, opened 2023-02-07
Clean up qutip.settings.Settings

hodgestar:
**Description**

The `Settings` class unnecessarily implements a singleton pattern and has a few unused underscore attributes (e.g. `_integartors`). This PR removes the unused / necessary code and simplifies the class.

**Related issues or PRs**

- None

### DIFF
--- a/qutip/settings.py
+++ b/qutip/settings.py
@@ -128,7 +128,8 @@ class Settings:
             self.tmproot = os.path.join(os.path.expanduser("~"), '.qutip')
         except OSError:
             self._tmproot = "."
-        self.core = None
+        self.core = None  # set in qutip.core.options
+        self.compile = None  # set in qutip.core.coefficient
         self._debug = False
         self._log_handler = "default"
         self._colorblind_safe = False


qutip/settings.py
```diff
@@ -122,21 +122,13 @@ class Settings:
...
-            cls._instance = super(Settings, cls).__new__(cls)
-        return cls._instance
-
     def __init__(self):
         self._mkl_lib = ""
         try:
             self.tmproot = os.path.join(os.path.expanduser("~"), '.qutip')
         except OSError:
             self._tmproot = "."
         self.core = None
```
Ericgig, line 131: The `compile` attribute is sneakily added in coefficient.py. Can we add it here just as a note comment that it's will be added.
```suggestion
        self.core = None
        self.compile = None
```

Ericgig (approved):
Look good.

qutip/settings.py
hodgestar: Good idea. I added `.compile` with a comment saying where the two are set.

PR #2087, opened 2023-02-12
Build documentation on ReadTheDocs

hodgestar:
**Description**
Build documentation on ReadTheDocs.

See https://qutip.readthedocs.io/en/feature-build-docs-on-readthedocs/index.html for the docs built for this branch.

**Related issues or PRs**
- None


Ericgig (approved):
It looks fine on https://qutip.readthedocs.io/en/feature-build-docs-on-readthedocs/index.html .

When a new version is released, we will need to update the `stable` tag on readthedocs. The release documentation should be updated to reflect that (In another PR after the website is updated?).
We may also want to remove the pdf build in the documentation github action.

hodgestar:
I would like CI to check that the documentation builds work, but maybe we can reduce what is built somewhat once we're happy with how ReadTheDocs is set up.

I suggest that on ReadTheDocs we build docs for all the active release branches (so currently qutip-4.7.X and qutip-5.0.X) and then yes, we'll have to update which is stable when that changes.

PR #2089, opened 2023-02-13
Add support for combining bosonic and fermionic HEOM baths.

hodgestar:
**Description**
Add support for combining bosonic and fermionic HEOM baths.

**Related issues or PRs**
- None.


Ericgig (approved):
The changes seems minimal for supporting a new feature.

The only test mixing boson and fermion I found had the boson bath decoupled. I would prefer to have a test with both baths participating to see if nothing strange is happening in their interaction, but I don't expect there is any reference to compare to so I will approve as is.

hodgestar:
I have been wondering whether one can truncate a bath and compare to simulating the full system using `sesolve` or `mesolve`. I will see if I can put something together.

nwlambert:
thanks simon, i think the logic of the changes looks good to me.  I can do some quick checks against some examples from the literature.

For tests, I think if we set up the correlation functions for the bosonic bath to basically mimic an undamped lorentzian we can probably use the existing discrete boson example as a benchmark. 
https://github.com/qutip/qutip-notebooks/blob/master/examples/heom/heom-5b-fermions-discrete-boson-model.ipynb
I will try and set it up and see.

hodgestar:
The failing test was unrelated:
```
FAILED qutip/tests/solver/test_floquet.py::TestFloquet::testFloquetUnitary - assert (0.9999980885...77441045e-05j) == 1.0 ± 5.0e-05
  comparison failed
  Obtained: (0.9999980885612143+5.1789504077441045e-05j)
  Expected: 1.0 ± 5.0e-05
```
Recording it here so that we can fix it (looks like just a tolerance issue in the Floquet tests).

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -876,6 +942,107 @@ def test_discrete_level_model_fermionic_bath_with_decoupled_bosonic_bath(
...
+        )
+
+        result = hsolver.run(dlm.rho(psi0), tlist)
+        current = [
+            dlm.state_current(ado_state, tags=["fermionic"])
+            for ado_state in result.ado_states
+        ]
+        sys_occupation = dlm.sys_occupation(result.states)
+
+        np.testing.assert_allclose(current_ref, current, atol=1e-3)
```
nwlambert, line 1025: just a minor thing, these currents are very small, is it better to use an rtol here?  i guess the steady-state results below are checked more carefully, so this would instead just spot really large deviations in the results?

nwlambert (approved):

qutip/tests/solver/heom/test_bofin_solvers.py
hodgestar: The currents at later time are larger, but will switch back to `rtol` since the currents at early times are small. I switched to `atol` because that `rtol` can cause issues in tests with random data, but this is not such a test.

PR #2095, opened 2023-02-14
Fix test requirement

Ericgig:
**Description**
Test for the bloch sphere require matplotlib and ipython to pass, but these were not installed with `qutip[tests]`.
These tests are now skipped when these packages are missing instead of failing.

Also add `loky` and `tqdm` to extra require as they can now be used in solvers.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ tests =
     pytest-rerunfailures
 ipython =
     ipython
-extra =
+extras =
     loky
     tqdm
 ; This uses ConfigParser's string interpolation to include all the above
@@ -67,4 +67,4 @@ full =
     %(semidefinite)s
     %(tests)s
     %(ipython)s
-    %(extra)s
+    %(extras)s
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -56,6 +56,9 @@ tests =
     pytest-rerunfailures
 ipython =
     ipython
+extra =
```
hodgestar, line 59:
```suggestion
extras =
```
This is grammatically slightly more correct and is more consistent with `graphics` and `tests`.

setup.cfg
```diff
@@ -64,3 +67,4 @@ full =
     %(semidefinite)s
     %(tests)s
     %(ipython)s
+    %(extra)s
```
hodgestar, line 70:
```suggestion
    %(extras)s
```

hodgestar (changes requested):
I'm not wild about a catch all `extras` dependency, but on the other hand don't have a better suggestion that adding separate extra dependencies named `loky` and `tqdm`. So I'm happy for this to be merged once the name is changed to `extras`.

hodgestar (approved):

PR #2110, opened 2023-03-03
Fix version_table to work without Cython installed

### LINKED ISSUE #2001: Clean up qutip.ipynbtools for version 5
### Bug Description

In https://github.com/qutip/qutip-tutorials/pull/42#pullrequestreview-1119409348 @BoxiLi noticed that `qutip.ipynbtools` imports Cython without a `try: ... except: ...` around the import. This results in `qutip.ipynbtools.version_table()` failing even though Cython isn't a required dependency of QuTiP.

We should do a thorough clean up of this module in dev-major and fix this particular issue while we're there.

### Code to Reproduce the Bug

_No response_

### Code Output

_No response_

### Expected Behaviour

`qutip.ipynbtools.version_table()` can be called without Cython installed.

### Your Environment

```shell
QuTiP version: 5.0.0.dev
```

### Additional Context

_No response_

gadhvirushiraj:
fix #2001

**Description**
- added `try: ...except:` in import of `Cython`
- fixed `version_table` to work without Cython installed

**Behaviour**
`qutip.ipynbtools.version_table()` can be called without Cython installed


qutip/ipynbtools.py
```diff
@@ -39,10 +39,14 @@
...
 import numpy
 import scipy
-import Cython
 import matplotlib
 import IPython
 
+try:
+    import Cython
+except:
+    pass
```
AGaliciaMartinez, line 48:
```suggestion
except ImportError:
    pass
```

For try except it is best to be specific with the type of exception to avoid unexpected bugs.

AGaliciaMartinez (changes requested):
Hi @gadhvirushiraj,

Thank you for the fix! We will need some tests to make sure that things work as intended. Our CI tools test several environments, including some with and without Cython. We would need to write a test for this, which is a little bit tricky in this case but Im sure you can handle it :smile:. This would be located in a new file in `qutip/tests/test_ipynbtools.py`. You will need to call the function `version_table` and make sure, for example, that Cython only appears in the output data if `Cython` is in `sys.modules`.  For this you will need to inspect the output string which is contained under the `html.data` attribute of the `HTML` object.

Feel free to ask for help if you get stuck!

gadhvirushiraj:
Hello @AGaliciaMartinez,
Thanks for the review. I will look into it and update.

qutip/ipynbtools.py
```diff
@@ -39,10 +39,14 @@
...
 import scipy
-import Cython
 import matplotlib
 import IPython
 
+try:
+    import Cython
+except ImportError:
+    pass
+
```
hodgestar, line 49:
```suggestion
try:
    import Cython
    _cython_available = True
except ImportError:
    _cython_available = False

```
Let's rather explicitly record whether the import of Cython succeeded. This is much more obviously correct than digging in `sys.modules` where all kinds of strange things may occur.

qutip/ipynbtools.py
```diff
@@ -67,13 +71,15 @@ def version_table(verbose=False):
...
-                ("Cython", Cython.__version__),
                 ("Number of CPUs", available_cpu_count()),
                 ("BLAS Info", _blas_info()),
                 ("IPython", IPython.__version__),
                 ("Python", sys.version),
                 ("OS", "%s [%s]" % (os.name, sys.platform))
                 ]
+    
+    if "Cython" in sys.modules:
+        packages.append(("Cython", Cython.__version__))
```
hodgestar, line 82:
```suggestion
    if _cython_available:
        packages.append(("Cython", Cython.__version__))
```

qutip/ipynbtools.py
```diff
@@ -368,4 +374,4 @@ def update(n):
     video_encoded = b64encode(video).decode("ascii")
     video_tag = '<video controls src="data:video/x-m4v;base64,{0}">'.format(
         video_encoded)
-    return HTML(video_tag)
+    return HTML(video_tag)
```
hodgestar, line 377: Please put back the new line at the end of the file.

qutip/tests/test_ipynbtools.py
```diff
@@ -0,0 +1,9 @@
+from qutip.ipynbtools import version_table
+import sys
+
+
+def test_version_table():
```
hodgestar, line 5: Let's split this into two tests:

- `test_version_table` with is always called and sanity checks the version table.
- `test_version_table_with_cython` which is only called when Cython is installed.

E.g. 
```
@pytest.mark.parametrize('verbose', [False, True])
def test_version_table(verbose):
    html = version_table(verbose=verbose)
    assert  "<th>Software</th>" in html.data
    assert "<th>Version</th>" in html.data
    if verbose:
        assert "<td>Installation path</td>" in html.data


@pytest.importorskip("Cython")
def test_version_table_with_cython():
    html = version_table()
    assert "<td>Cython</td>" in html.data
```

hodgestar (changes requested):
Thank you for your PR. I've left some suggestions for improving the implementation and tests. Once you've implemented them, please ping us for a second review.

qutip/tests/test_ipynbtools.py
gadhvirushiraj: Hello @hodgestar,
Thanks for the review. 
Should I also check for other library dependencies under the `test_version_table` ?

hodgestar: Yes, adding a few more checks to `test_version_table` would be good.

qutip/ipynbtools.py
```diff
@@ -67,13 +72,15 @@ def version_table(verbose=False):
...
                 ("SciPy", scipy.__version__),
                 ("matplotlib", matplotlib.__version__),
-                ("Cython", Cython.__version__),
                 ("Number of CPUs", available_cpu_count()),
                 ("BLAS Info", _blas_info()),
                 ("IPython", IPython.__version__),
                 ("Python", sys.version),
                 ("OS", "%s [%s]" % (os.name, sys.platform))
                 ]
+
```
hodgestar, line 81:
```suggestion

```

hodgestar (approved):

hodgestar:
@gadhvirushiraj Thanks! I've fixed the Code Climate issue and approved. I'll merge once the CI checks finish (and pass, which they should do).

PR #2116, opened 2023-03-13
Remove qutip.control and replace with qutip_qtrl.

hodgestar:
**Description**
Remove `qutip.control` and replace it with the QuTiP family package `qutip-qtrl`.

**Related issues or PRs**
- None


Ericgig (approved):
Happy to see this moved out.
But `qutip-qip` and some tutorial use it.
So we should wait for qutip-ctrl to be added to their requirement before merging.

hodgestar:
> But `qutip-qip` and some tutorial use it. So we should wait for `qutip-qctrl` to be added to their requirement before merging.

Good point. I will talk to @BoxiLi.

Ericgig:
Some of the example notebook on the web page call it directly: https://nbviewer.org/github/qutip/qutip-notebooks/blob/master/examples/control-pulseoptim-QFT.ipynb
It seems they are still in `qutip-notebook`.

hodgestar:
> Some of the example notebook on the web page call it directly: https://nbviewer.org/github/qutip/qutip-notebooks/blob/master/examples/control-pulseoptim-QFT.ipynb It seems they are still in `qutip-notebook`.

The `qutip-notebooks` are only for QuTiP v4, but yes, we do have to migrate those to `qutip-tutorials` and make them v5 compatible.

BoxiLi (approved):

PR #2126, opened 2023-03-17
Add operation to trace oper-ket without reshaping

Ericgig:
**Description**
Add and operation to compute the trace of column stacked operators without reshaping them.
The reshape would otherwise do a temporary copy of the data.
Another addition to improve stochastic solvers timings.

### DIFF
--- a/qutip/core/data/trace.pyx
+++ b/qutip/core/data/trace.pyx
@@ -2,11 +2,14 @@
 #cython: boundscheck=False, wraparound=False, initializedcheck=False
 
 cimport cython
+from libc.math cimport sqrt
 
 from qutip.core.data cimport Data, CSR, Dense
+from qutip.core.data cimport base
 
 __all__ = [
     'trace', 'trace_csr', 'trace_dense',
+    'trace_oper_ket', 'trace_oper_ket_csr', 'trace_oper_ket_dense',
 ]
 
 
@@ -14,6 +17,13 @@ cdef void _check_shape(Data matrix) nogil except *:
     if matrix.shape[0] != matrix.shape[1]:
         raise ValueError("".join([
             "matrix shape ", str(matrix.shape), " is not square.",
+        ]))
+
+
+cdef void _check_shape_oper_ket(int N, Data matrix) nogil except *:
+    if matrix.shape[0] != N * N or matrix.shape[1] != 1:
+        raise ValueError("".join([
+            "matrix ", str(matrix.shape), " is not a stacked square matrix."
         ]))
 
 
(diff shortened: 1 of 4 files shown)


qutip/core/data/trace.pyx
```diff
@@ -39,6 +49,28 @@ cpdef double complex trace_dense(Dense matrix) nogil except *:
     return trace
 
 
+cpdef double complex trace_oper_ket_csr(CSR matrix) nogil except *:
+    cdef int N = <int>sqrt(matrix.shape[0])
+    _check_shape_oper_ket(N, matrix)
+    cdef size_t row, ptr
```
hodgestar, line 55: We're no longer using `ptr` here.

Perhaps we can make both the sparse and dense versions a bit clearer and simpler by doing:
```
for ptr in range(0, N * N, N + 1):
    ... 
```
I'm also happy with using stride for the sparse version too.

hodgestar (changes requested):
Looks good.

I don't see a way to do this without adjusting the data layer somehow, since it doesn't have the information to decide whether a matrix is a stacked operator or not.

What happens in one wants to trace the adjoint of an oper_ket? I.e. an oper_bra? Or does QuTiP not really support that?

qutip/core/data/trace.pyx
Ericgig: Can't do range with step without the gil...
Cython can't detect the step sign at compilation, I tried using uint, but no luck...

qutip/solver/mcsolve.py
```diff
@@ -225,12 +225,12 @@ def reset(self, hard=False):
 
     def _prob_func(self, state):
         if self.issuper:
-            return _data.norm.trace(unstack_columns(state))
+            return _data.trace_oper_ket(state).real
```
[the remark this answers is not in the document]
Ericgig: We have both `data.trace` and `data.norm.trace`.
`data.norm.trace` is defined as `tr(sqrt(A @ A.adjoint()))`...
For dm, they give the same result and the normal trace is faster, so I should have used that one from the start.

Not a fan of having 2 functions with the same name, I sometime get confused and use the wrong one.
But this one should not be in `data.norm`.

qutip/core/data/trace.pyx
Ericgig: If by explicitly set stride you mean defining a variable for it instead of using `N+1` everywhere, yes I did the change.

Ericgig:
@hodgestar, is there anything else you want me to fix in this PR?

ps. Right now, to trace oper-bra, you would need to get it's adjoint first, use this or unstack then trace.

hodgestar (approved):
Looks good!

hodgestar:
@Ericgig Apologies for the delay. I wanted to go look at `data.norm.trace`. The implementation there is completely different and the default one used by Qobj for `qobj.norm()` for operators. I'm wondering if the more complex eigenvalue decomposition definition is really worth it? Not a question for this PR though.

PR #2128, opened 2023-03-20
Speed up the HEOM RHS construct

hodgestar:
**Description**

Adds a `_from_csr_blocks` constructor to speed up the construction of the HEOM RHS by a factor of ~4x.

What allows `from_csr_blocks` to be fast is that we can order the blocks in a way that makes it possible to write the rows of the output one after the other without a lot of work, and we never have to add a new non-zero element into a row or make a copy of the big output matrix.

The implementation in Python was slow just because there is a lot of Python looping and arithmetic when the number of operators being combined becomes large.

`_from_csr_blocks` is a private constructor for now. If use cases emerge we can consider making something similar part of the public interface later.

**Related issues or PRs**
- None

### DIFF
--- a/qutip/core/data/csr.pyx
+++ b/qutip/core/data/csr.pyx
@@ -897,6 +897,9 @@ cpdef CSR _from_csr_blocks(
             " the same length."
         )
 
+    if n_ops == 0:
+        return zeros(shape, shape)
+
     # check op shapes and calculate nnz
     for op in block_ops:
         nnz_ += nnz(op)
@@ -904,10 +907,6 @@ cpdef CSR _from_csr_blocks(
             raise ValueError(
                 "Block operators (block_ops) do not have the correct shape."
             )
-
-    cdef CSR out = empty(shape, shape, nnz_)
-    if n_ops == 0:
-        return out
 
     # check ops are ordered by (row, column)
     row_idx = block_rows[0]
@@ -925,8 +924,9 @@ cpdef CSR _from_csr_blocks(
         col_idx = block_cols[i]
 
     if nnz_ == 0:
-        return out
+        return zeros(shape, shape)
 
+    cdef CSR out = empty(shape, shape, nnz_)
     cdef base.idxint op_idx = 0
     cdef base.idxint prev_op_idx = 0
     cdef base.idxint end = 0
--- a/qutip/core/data/csr.pyx
+++ b/qutip/core/data/csr.pyx
@@ -886,48 +886,75 @@ cpdef CSR from_csr_blocks(
     cdef CSR op
     cdef base.idxint shape = n_blocks * block_size
     cdef base.idxint nnz_ = 0
+    cdef base.idxint n_ops = len(block_ops)
+    cdef base.idxint i, j
+    cdef base.idxint row_idx, col_idx
 
+    # check arrays are the same length
+    if len(block_rows) != n_ops or len(block_cols) != n_ops:
+        raise ValueError(
+            "The arrays block_rows, block_cols and block_ops should have"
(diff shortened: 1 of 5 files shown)


hodgestar:
@nwlambert Would you mind giving this a try with some bigger realistic HEOM examples you have lying around? Eric's recent improvements to CSR `mul` and `imul` also help HEOM RHS construction since there is a lot of `c * op` happening. The 3.5x improvement from this branch is on top of that.

@Ericgig Would you mind giving your thoughts on the Cython interface both from a "is this good from a technical point of view" and from a "do we want this point of view". I guess we could also name it `_from_csr_blocks` if we don't want to expose it.

hodgestar:
> It doesn't look robust enough to be user facing as it is, but as a private function, if you get a 3.5x, I'd say we want it.

I'll rename it to `_from_csr_blocks` and add some more checks on the ordering and shape of the ops if those aren't too expensive and add some tests. We can expose it publicly if that's ever useful.

> Technically, just give a type to `i` and it's all running in c code.

Thanks! Adding the type to `i` did make it a little faster.

qutip/tests/core/data/test_csr.py
```diff
@@ -343,6 +343,105 @@ def test_one_element_error(self, shape, position, value):
                                          " out of bound: ")
 
 
+class TestFromCSRBlocks:
+    class _blocks:
+        def __init__(self, rows, cols, ops, n_blocks=2, block_size=2):
+            self.rows = np.array(rows, dtype=data.base.idxint_dtype)
+            self.cols = np.array(cols, dtype=data.base.idxint_dtype)
+            if isinstance(ops, int):
+                ops = [csr.empty(block_size, block_size, 0)] * ops
```
Ericgig, line 352:
```suggestion
            if isinstance(ops, int):
                ops = [csr.zeros(block_size, block_size, 0)] * ops
```
Here I feels `csr.zeros` is what you want. 
`csr.zeros` is a sparse array with no elements.
`csr.empty`, like `np.empty`, allocate the memory for you to fill manually, but it's not to use as is.

Ericgig (approved):
Seems fine.
This fixes the used data type to `CSR` and this should be documented.
I think you want `csr.zeros` instead of `csr.empty`.

qutip/core/data/csr.pyx
```diff
@@ -844,3 +844,122 @@ def diags(diagonals, offsets=None, shape=None):
...
+        ):
+            raise ValueError(
+                "The arrays block_rows and block_cols must be sorted"
+                " by (row, column)."
+            )
+        row_idx = block_rows[i]
+        col_idx = block_cols[i]
+
+    if nnz_ == 0:
+        return out
```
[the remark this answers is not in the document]
hodgestar: Specifying a bad operator shape or supplying unsorted rows and columns is still a logical error even if the operators are all zeros.

qutip/tests/core/data/test_csr.py
hodgestar: Thanks! I realized that `csr.empty` is really an internal method half way through and then forgot to clean up all the cases in the tests and outputs. I have sorted them out now.

hodgestar:
> This fixes the used data type to `CSR` and this should be documented.

It is documented in the docstring for `block_ops`? Or did you want more documentation somewhere?

hodgestar:
@Ericgig Thank you for the review. Would you mind having a last look? I made quite a few small changes.

Ericgig (approved):
Better with `zeros`.

I would document in `HEOMSolver` say there that the Hamiltonian and bath operator will be coerced to `csr`.

PR #2130, opened 2023-03-21
Add annotate cython files options when installing the module

Ericgig:
**Description**
Add options to annotate the cython files at installation.
I used it when updating the stochastic solvers, but do not what to include it in that PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,13 +44,7 @@ def process_options():
     options = _determine_user_arguments(options)
     options = _determine_version(options)
     options = _determine_compilation_options(options)
-    options["annotate"] = False
-    if "--annotate" in sys.argv:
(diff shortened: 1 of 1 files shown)


hodgestar (changes requested):
This looks very useful. Thanks!

hodgestar (approved):
Looks good to me.

One last thought -- do we want to keep support for `-a`? The `setup` command already parses a bunch of options on its own while `-a` doesn't appear to clash with anything right now, it is easy for a single letter to clash (e.g. `-b` and `-c` are already used by the `build` subcommand).

Ericgig:
I mostly use it as `-a`, I would remove it if/when it clash.

hodgestar (approved):

PR #2131, opened 2023-03-21
Stochastic solver using v5 class interface

Ericgig:
**Checklist**
This rewrite the stochastic solver to use the class interface.

- `smesolve`, `ssesolve` have been rewritten to match other solver with matching class interface `SMESolver`, `SSESolver`. The main change is that the options used to be passed as keywords (`**kwargs`) and are now passed as an options dict.
  - `m_ops` and `dW_factors` are properties of the classes, but not available to overwrite through the function interface. 
  - The multiples end condition (`target_tol`, `timeout`) are added.
  - The `solver` keyword is now `method` options, as it is in other solver.
  - The method that was referring the homodyne or heterodyne detection is now the `heterodyne` keyword. ie before `method="heterodyne"` now `heterodyne=True`.
  - The default integration method is now `rouchon`, instead of `taylor1.5`. ie the method coming from a physic paper and not a mathematics textbook. 

- The stochastic integration methods have been spitted in smaller chunk instead of the huge classes in _stochastic.pyx. I kept most methods, only skipping the taylor order 2.0 method. The python interface is kept very similar to the `Integrator` as `SIntegrator`, the main difference being that `set_state` and `get_state` use an extra random number generator with the time and state. The integration methods are usually 5%~10% slower than they where for medium size problems (shape ~100). Most method are computed the same way as before, but the rouchon method no longer use super operator, but keep the density matrix as a matrix and not an oper-ket.

- The derivative computation for used by the integration method has also been split into multiple functions and made only available for `smesolve`. For there methods to be valid, the each diffusion contribution must commute. For `smesolve`, this means that each `sc_ops` must commute, as documented. For `ssesolve`, each `sc_ops` must not only commute, but also commute with other `sc_ops`'s adjoint, this last part being missing from v4's documentation. It also meant that the sc_ops needed to commute with it's own adjoint when using heterodyne detection...

- `ssepdpsolve`, `smepdpsolve`, `photocurrent_mesolve`, `photocurrent_sesolve` removed. They are all less efficient variants of `mcsolve`. `mcsolve` already include the photocurrent in the output.

- `general_stochastic`, removed. It is a mostly generic stochastic equation solver. The user to define their equations to integrate, but the results are forced into `Results` with expectation values and dims checks, making it hard to use.


Ericgig:
I changed back the default integration method to `taylor1.5` for `smesolve` and `platen` for `ssesolve`.
The property that the state stay physical of `rouchon` is actually problematic since even when the `dt` is too large, the state still look fine but is totally wrong. Whereas other methods will get negative eigenvalues or Nan.

qutip/solver/sode/_sode.pyx
```diff
@@ -0,0 +1,510 @@
...
+    def __init__(self, _StochasticSystem system):
+        self.system = system
+
+    @cython.wraparound(False)
+    def run(
+        self, double t, Data state, double dt,
+        double[:, :, ::1] dW, int num_step
+    ):
+        cdef int i
+        cdef Data out
```
hodgestar, line 24: `out` appears to be unused.

qutip/solver/sode/_sode.pyx
```diff
@@ -0,0 +1,510 @@
...
+    cdef Data inv
+    cdef double prev_dt
+    cdef dict imp_opt
+
+    def __init__(self, _StochasticSystem system, imp_method=None, imp_options={}):
+        self.system = system
+        self.prev_dt = 0
+        if imp_method == "inv":
+            if not self.system.L.isconstant:
+                raise TypeError("")
```
hodgestar, line 407:
```suggestion
                raise TypeError("The 'inv' integration method requires that the system Hamiltonian or Liouvillian be constant.")
```

qutip/solver/sode/_sode.pyx
```diff
@@ -0,0 +1,510 @@
...
+            self.use_inv = True
```
hodgestar, line 408:
```suggestion
            self.use_inv = True
            self.imp_opt = {}
```
Probably makes sense to initialize `imp_opt` here explicitly too even if it isn't used.

qutip/solver/sode/_sode.pyx
```diff
@@ -0,0 +1,510 @@
...
+            self.imp_opt = {"method": imp_method, "options": imp_options}
+
+
+    @cython.wraparound(False)
+    def run(self, double t, Data state, double dt, double[:, :, ::1] dW, int ntraj):
+        cdef int i
+        if type(state) != _data.Dense:
+            state = _data.to(_data.Dense, state)
+        cdef Dense tmp = _data.mul_dense(state, 0)
+        state = state.copy()
```
hodgestar, line 420: Do we need to copy the state here? It's only passed to `step`.

qutip/solver/sode/ssystem.pyx
```diff
@@ -0,0 +1,550 @@
...
+    cpdef Data a(self):
+        if not self._is_set:
+            raise RuntimeError
+        if not self._a_set:
+            self._compute_a()
+        return self._a
+
+    cdef void _compute_a(StochasticOpenSystem self) except *:
+        if not self._is_set:
+            raise RuntimeError
```
hodgestar, line 259: Needs an error message.

qutip/solver/sode/ssystem.pyx
```diff
@@ -0,0 +1,550 @@
...
+        imul_dense(self._a, 0)
+        self.L.matmul_data(self.t, self.state, self._a)
+        self._a_set = True
+
+    cpdef Data bi(self, int i):
+        if not self._is_set:
+            raise RuntimeError
```
hodgestar, line 266: Needs an error message.

qutip/solver/sode/ssystem.pyx
```diff
@@ -0,0 +1,550 @@
...
+        if not self._b_set:
+            self._compute_b()
+        return _dense_wrap(self._b[i, :])
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    cdef void _compute_b(self) except *:
+        if not self._is_set:
+            raise RuntimeError
```
hodgestar, line 275: Same here.

qutip/solver/stochastic.py
```diff
@@ -0,0 +1,754 @@
...
+
+        if self.issuper and not self.H.issuper:
+            self.dims = [self.H.dims, self.H.dims]
+        else:
+            self.dims = self.H.dims
+
+    def __call__(self, options):
+        if self.issuper:
+            return StochasticOpenSystem(
+                self.H, self.sc_ops, self.c_ops, options.get("dt", 1.0 - 6)
```
hodgestar, line 217:
```suggestion
                self.H, self.sc_ops, self.c_ops, options.get("dt", 1e-6)
```
Was this intended to be `1e-6`? `-5` seems an odd default value for `dt`.

qutip/solver/stochastic.py
```diff
@@ -0,0 +1,754 @@
...
+        - map : str {"serial", "parallel", "loky"}, ["serial"]
+          How to run the trajectories. "parallel" uses concurent module to run
+          in parallel while "loky" use the module of the same name to do so.
+        - job_timeout : NoneType, int, [None]
+          Maximum time to compute one trajectory.
+        - num_cpus : NoneType, int, [None]
+          Number of cpus to use when running in parallel. ``None`` detect the
+          number of available cpus.
+        - dt : float [0.001]
+          The finite steps lenght for the Stochastic integration method.
```
hodgestar, line 325:
```suggestion
          The finite steps length for the Stochastic integration method.
```

qutip/solver/stochastic.py
```diff
@@ -0,0 +1,754 @@
...
+          Main ones are {"euler", "rouchon", "platen", "taylor1.5_imp"}
+        - map : str {"serial", "parallel", "loky"}, ["serial"]
+          How to run the trajectories. "parallel" uses concurent module to run
+          in parallel while "loky" use the module of the same name to do so.
+        - job_timeout : NoneType, int, [None]
+          Maximum time to compute one trajectory.
+        - num_cpus : NoneType, int, [None]
+          Number of cpus to use when running in parallel. ``None`` detect the
+          number of available cpus.
+        - dt : float [0.001]
```
hodgestar, line 324: It seems the default value for `dt` is not always `0.001` but it is set differently in a few places. Is the default integrator specific? Perhaps this should be left in the integrator options? Not sure.

qutip/tests/solver/test_stochastic.py
```diff
@@ -0,0 +1,323 @@
...
+
+    return H, sc_ops
+
+
+@pytest.mark.parametrize("system", [
+    "simple", "2 c_ops", "H td", "complex", "c_ops td",
+])
+@pytest.mark.parametrize("heterodyne", [True, False])
+def test_smesolve(heterodyne, system):
+    "Stochastic: smesolve: homodyne, time-dependent H"
```
hodgestar, line 46:
```suggestion
```
Docstring seemed wildly inaccurate.

qutip/tests/solver/test_stochastic.py
```diff
@@ -0,0 +1,323 @@
...
+            for w in res.wiener_process
+        ])
+        assert all([
+            m.shape == (len(sc_ops), len(times)-1)
+            for m in res.measurement
+        ])
+
+
+def test_reuse_seeds():
+    "Stochastic: smesolve: homodyne, time-dependent H"
```
hodgestar, line 246:
```suggestion
```

qutip/tests/solver/test_stochastic.py
```diff
@@ -0,0 +1,323 @@
...
+
+    np.testing.assert_allclose(res.expect, res2.expect, atol=1e-14)
+
+    for out1, out2 in zip(res.final_state, res2.final_state):
+        assert out1 == out2
+
+
+@pytest.mark.parametrize("heterodyne", [True, False])
+def test_m_ops(heterodyne):
+    "Stochastic: smesolve: homodyne, time-dependent H"
```
hodgestar, line 289:
```suggestion
```

qutip/tests/solver/test_stochastic.py
```diff
@@ -0,0 +1,323 @@
...
+    if heterodyne:
+        m_ops = m_ops * 2
+
+    times = np.linspace(0, 1.0, 51)
+
+    options = {"store_measurement": True,}
+
+    solver = SMESolver(H, sc_ops, heterodyne=heterodyne, options=options)
+    solver.m_ops = m_ops
+    solver.dW_factors = [0.] * len(m_ops)
```
hodgestar, line 306: I'm not sure what is happening in these two lines. Hopefully they are unneeded?

hodgestar (changes requested):
The original giant `.pyx` from QuTiP 4 was way too big. I think more reorganizing could be done, but let's leave that for the future. For now let's just finish cleaning this up and merge it.

qutip/tests/solver/test_stochastic.py
hodgestar: Okay, let's keep this for now. It's a bit strange and different to how we do things elsewhere, but let's sort it out later when we have more use cases.

doc/guide/dynamics/dynamics-stochastic.rst
```diff
@@ -126,46 +133,45 @@ where :math:`x` is the operator passed using ``m_ops``. The results are availabl
...
+For other examples on :func:`qutip.solver.stochastic.smesolve`, see the `following notebook <https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/development/development-smesolve-tests.ipynb>`_, as well as these notebooks available at `QuTiP Tutorials page <https://qutip.org/tutorials.html>`_: `heterodyne detection <https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/smesolve-heterodyne.ipynb>`_, `inefficient detection <https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/smesolve-inefficient-detection.ipynb>`_, and `feedback control <https://nbviewer.ipython.org/github/jrjohansson/reproduced-papers/blob/master/Reproduce-SIAM-JCO-46-445-2007-Mirrahimi.ipynb>`_.
```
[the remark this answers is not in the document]
hodgestar: Okay.

qutip/solver/__init__.py
```diff
@@ -16,3 +16,5 @@
 from .brmesolve import *
 from .krylovsolve import *
 from .parallel import *
+import qutip.solver.sode
```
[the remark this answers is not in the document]
hodgestar: Is the idea that one can then access `qutip.solver.sode.XXX` does this do this? It looks like it will make `XXX` available as `qutip.solver.sode.qutip.solver.XXX`? Maybe I am missing something.

hodgestar:
@Ericgig Thank you for adding the mcsolve photocurrent docs. I just left one more question about the `qutip.solver.sode` import.

qutip/solver/__init__.py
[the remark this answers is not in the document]
Ericgig: Updated to `import qutip.solver.sode as sode`, I thought this renaming was automatic...

hodgestar (approved):

PR #2132, opened 2023-03-23
Add tools to list dispatcher specializations

Ericgig:
**Description**
Add a function that list all dispatched functions and their specialization.
It automatically adapt to new data-layer and dispatched functions.
Only show pure specialization: `solve[Dense, Dense]` not `solve[CSR, Dense]`.

I also register 2 specializations that were created, but not registered.

Example:
```
>>> from qutip.core.data.dev_utils import specialisation_table
>>> specialisation_table()

Output only specialisation
|Dispatched function | JaxArray |   CSR    |  Dense   |
|zeros               |   True   |   True   |   True   |
|identity            |   True   |   True   |   True   |
|diag                |   True   |   True   |   True   |
|one_element         |   True   |   True   |   True   |
...
Binary specialisation
|Dispatched function | JaxArray |   CSR    |  Dense   |
|add                 |   True   |   True   |   True   |
|sub                 |   True   |   True   |   True   |
|matmul              |   True   |   True   |   True   |
|multiply            |   True   |   True   |   True   |
|kron                |   True   |   True   |   True   |
|solve               |  False   |  False   |   True   |
...
```

This could also go into a notebook if it fit better there.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,8 @@
 [run]
 source = qutip
 omit =
+    # Skip dev tools
+    *dev_utils.py
     # QuTiP test files
     */qutip/tests/*
 
--- a/qutip/core/data/dev_utils.py
+++ b/qutip/core/data/dev_utils.py
@@ -1,0 +1,79 @@
+"""
+Tools to help with data layer development.
+"""
+
+from collections import defaultdict
+from qutip.core import data as _data
+
+def all_specialisation():
+    """
+    Create a dict of all specialisations of the dispatcher sorted in number of
+    data layer inputs.
+    """
+    all_dispatchers = qt.data.to.dispatchers
+    dispatchers = defaultdict(list)
+    for dispatcher in all_dispatchers:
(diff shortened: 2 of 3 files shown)


.coveragerc
```diff
@@ -1,6 +1,8 @@
 [run]
 source = qutip
 omit =
+    # Skip dev tools
```
hodgestar, line 4: I don't think we should skip the dev tools in coverage. It's always good to know what code is being covered by tests, even if we decide not to cover them.

Aside: I also think that tests *should* be included in coverage. I've often picked up issues where a test or part of a test isn't being run by looking at coverage (e.g. the test is skipped if a dependency isn't installed and no test runner instance has that dependency).

Ericgig:
I just cleaned up a tool I used when making the jax data layer, but I don't want to add something that will ask for our time to maintain, thus no test or coverage.

In this meaning, it doesn't seems belong here.

I am closing this and making a new one with just the missing specializations + will check if they were also skipped in test.

PR #2135, opened 2023-03-27
Optimize dispatcher by using positional only arguments

Ericgig:
**Description**
I optimized the dispatched by only supporting positional arguments for matrix to dispatch on:

The dispatcher was supporting any signature for dispatched functions.
Since each dispatchers is an instance of the `Dispatcher` class and cython does not support patching methods, it had to find the inputs matrix to convert from `__call__(*args, **kwargs)` and re-implemented the `_bind` method to do so.
But all our dispatched functions have the matrix as the first positional argument, by embracing this we get a nice speed up for small matrices.

Also when the specialization existed, it would not call it directly, but call  a `_constructed_specialisation` that called ti. Adding another unneeded layer to each calls. This was changed so it would call the function directly.

For a 2x2 matrix:
|                  | Before | After |
|------------------|--------|-------|
| data.add_dense   | 214ns  | 210ns |
| data.add (dense) | 774ns  | 388ns |
| data.add[Dense, Dense] | 541ns  | 182ns |
| data.add_csr     | 429ns  | 447ns |
| data.add (csr)   | 986ns  | 649ns |
| data.add[CSR. CSR]  | 744ns  | 401ns |
| numpy | 473ns | - |

Sadly it will not improve the benchmarks that much since the `Qobj` operations overhead is quite large:
|                  | Before | After |
|------------------|--------|-------|
| Qboj + Qobj (dense)   | 2430ns  | 2070ns |
| Qboj + Qobj (csr)   | 3070ns  | 2330ns |

### DIFF
--- a/qutip/core/data/dispatch.pyx
+++ b/qutip/core/data/dispatch.pyx
@@ -1,4 +1,5 @@
 #cython: language_level=3
+#cython: boundscheck=False
 
 import functools
 import inspect
@@ -48,7 +49,7 @@ cdef class _constructed_specialisation:
     """
     cdef readonly bint _output
     cdef object _call
-    cdef readonly Py_ssize_t _n_inputs
+    cdef readonly Py_ssize_t _n_inputs, _n_dispatch
     cdef readonly tuple types
     cdef readonly tuple _converters
     cdef readonly str _short_name
@@ -57,10 +58,8 @@ cdef class _constructed_specialisation:
     cdef public str __module__
     cdef public object __signature__
     cdef readonly str __text_signature__
-    cdef readonly bint direct
 
-    def __init__(self, base, Dispatcher dispatcher, types, converters, out,
-                 direct):
+    def __init__(self, base, Dispatcher dispatcher, types, converters, out):
         self.__doc__ = inspect.getdoc(dispatcher)
         self._short_name = dispatcher.__name__
         self.__name__ = (
@@ -74,10 +73,11 @@ cdef class _constructed_specialisation:
         self._output = out
         self._call = base
         self.types = types
-        self.direct = direct
         self._converters = converters
+        self._n_dispatch = len(converters)
         self._n_inputs = len(converters) - out
 
+    @cython.wraparound(False)
     def __call__(self, *args, **kwargs):
         cdef int i
         cdef list _args = list(args)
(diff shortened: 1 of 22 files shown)


qutip/core/data/dispatch.pyx
```diff
@@ -564,20 +369,25 @@ cdef class Dispatcher:
         return "<dispatcher: " + self.__text_signature__ + ">"
 
     def __call__(self, *args, dtype=None, **kwargs):
-        cdef list args_, dispatch
-        cdef dict kwargs_
+        cdef list dispatch = []
```
hodgestar, line 372:
```suggestion
        cdef list dispatch = [None] * self._n_dispatch
```
Might also be a bit faster here than append, but not sure. It doesn't change the number of objects created. If we avoid append further up, it might be consistent to avoid it here too.

In both cases, it would be awesome to avoid creating intermediate Python objects entirely but I'm not sure if `from cpython cimport PyTuple_New, PyTuple_SetItem` is necessarily a route we want to go down unless it gives a useful performance improvement.

hodgestar (changes requested):
I left some small suggestions, but overall this is a great simplification.

qutip/core/data/dispatch.pyx
```diff
@@ -248,13 +46,12 @@ cdef class _constructed_specialisation:
     See `self.__signature__` or `self.__text_signature__` for the call
     signature of this object.
     """
-    cdef _bind _parameters
-    cdef bint _output
+    cdef readonly Py_ssize_t _output
     cdef object _call
-    cdef Py_ssize_t _n_dispatch
+    cdef readonly Py_ssize_t _n_input_converters
```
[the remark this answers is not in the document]
Ericgig: It's not the total number of inputs, but the number data layer of inputs we dispatch on.

qutip/core/data/dispatch.pyx
Ericgig: I tried and timed it and other variant (`boundcheck=False`, ...) and nothing had any observable effect.
I think `PyTuple_New` will create python object so it I don't expect it to give much improvement.
`dispatch`'s length is variable, so it's cleaner with `append`.

hodgestar: PyTuple_New avoids the intermediate list, but I don't like it for other reasons (it does weird things with references and we don't really want to get into the job of keep track of reference counts ourselves). If you didn't see a performance benefit, happy to keep things as they are.

hodgestar (approved):

PR #2141, opened 2023-03-29
Add `Qobj.get` to obtain the data in common type

Ericgig:
**Description**
I think that with the different data type a `Qobj`  can contain, it would be useful to be able get the underlying data (ndarray, csr_matrix, jax.Array, etc.) easily from the `Qobj`. I added the method `Qobj.get(copy)` for that.

I allow direct access to the object in case it's needed for auto-differentiation or other advanced features.

I am open to a better name than `get` if it's too generic.


qutip/core/data/extract.py
```diff
@@ -0,0 +1,102 @@
...
+        Whether to pass a copy of the object or not.
+    """
+    if format not in [None, "ndarray"]:
+        raise ValueError(
+            "Valid format for Dense is 'ndarray'"
+        )
+    if copy:
+        return matrix.to_array()
+    else:
+        return matrix.as_ndarray()
```
hodgestar, line 34:
```suggestion
def extract_dense(matrix, format=None, copy=True):
    """
    Return an array representation of the Dense data object.

    Parameters
    ----------
    matrix : Data
        The matrix to convert to the given format.

    format : str {"ndarray"}, default="ndarray"
        Type of the output.

    copy : bool, default: True
        Whether to return a copy of the data. If False,
        a view of the data is returned when possible.
    """
    if format not in [None, "ndarray"]:
        raise ValueError(
            f"Dense data instances do not support being extracted to format {format!r}"
        )
    if copy:
        return matrix.to_array()
    else:
        return matrix.as_ndarray()
```

qutip/core/data/extract.py
```diff
@@ -0,0 +1,102 @@
...
+        raise ValueError(
+            "Valid format for CSR is 'csr_matrix'"
+        )
+    csr_mat = matrix.as_scipy()
+    if copy:
+        csr_mat = csr_mat.copy()
+    return csr_mat
+
+
+get = _Dispatcher(
```
hodgestar, line 63:
```suggestion
extract = _Dispatcher(
```

qutip/core/data/extract.py
```diff
@@ -0,0 +1,102 @@
...
+    _inspect.Signature([
+        _inspect.Parameter('matrix', _inspect.Parameter.POSITIONAL_OR_KEYWORD),
+        _inspect.Parameter(
+            'format', _inspect.Parameter.POSITIONAL_OR_KEYWORD, default=None
+        ),
+        _inspect.Parameter(
+            'copy', _inspect.Parameter.POSITIONAL_OR_KEYWORD, default=True
+        )
+    ]),
+    name='get',
```
hodgestar, line 73:
```suggestion
    name='extract',
```

hodgestar (changes requested):
I like this new interface and I'm fairly happy with the name `extract` unless someone thinks of a better option.

The doc strings also need some updating.

AGaliciaMartinez:
What about `data_as`, `get_as` or `extract_as`? I think adding `as` to any of the method names could facilitate understanding what this method does without reading documentation. I personally prefer `data_as` given that `data` is already a method for qobj.

Ericgig:
@hodgestar, is the current naming fine for you? I

hodgestar (approved):
Sorry, I think I missed the last commit. Happy with the new name.

PR #2142, opened 2023-03-30
Update `ttmsolve` for v5

Ericgig:
**Description**
Move `ttmsolve` to `qutip/solver` and update it.

No class interface was created, `ttmsolve` work differently to other solver and would be hard to force into the `Solver` mould.

Changed the interface a little, `learningtimes` needed use the first values `times` to be valid, it was replaced by an options for the number of values in ``times`` to use.


Ericgig:
Example in qutip/qutip-tutorials#63

qutip/solver/nonmarkov/transfertensor.py
```diff
@@ -0,0 +1,171 @@
...
+          On `None` the states will be saved if no expectation operators are
+          given.
+        - normalize_output : bool
+          Normalize output state to hide ODE numerical errors.
+        - threshold : float
+          Threshold for halting. Halts if  :math:`||T_{n}-T_{n-1}||` is below
+          treshold.
+        - num_learning : int
+          Number of times used to construct the dynmaps operators when
+          ``dynmaps`` is a callable.
```
hodgestar, line 60: Maybe this shouldn't be an option since it's not optional if `dynmaps` is callable?

qutip/solver/nonmarkov/transfertensor.py
```diff
@@ -0,0 +1,171 @@
...
+    if callable(dynmaps):
+        if not options["num_learning"]:
+            raise ValueError(
+                "When dynmaps is a callable, options['num_learning'] must be "
+                "the number of dynamical maps to compute."
+            )
+        dynmaps = [dynmaps(t) for t in times[: opt["num_learning"]]]
+
+    if not dynmaps[0].issuper:
+        raise ValueError("`dynmaps` entries must be super operators.")
```
hodgestar, line 92: Is it worthwhile just checking that all dynmaps are super operators since it isn't expensive? We should perhaps also check that `dynmaps` isn't an empty list if that isn't allowed.

qutip/solver/nonmarkov/transfertensor.py
```diff
@@ -0,0 +1,171 @@
...
+    start = time.time()
+    results = Result(e_ops, opt, solver="ttmsolve", stats=stats)
+    states = [rho0vec]
+    results.add(times[0], state0)
+
+    for n in range(1, len(times)):
+        # Set current state
+        state = 0
+        for j in range(1, min(K, n + 1)):
+            tmp = tensors[j] * states[n - j]
```
hodgestar, line 126:
```suggestion
            tmp = tensors[j] @ states[n - j]
```
Should we use `@` here?

hodgestar (changes requested):
I left a couple of questions and suggestions, but overall looks good.

hodgestar (approved):

PR #2120, opened 2023-03-14
Replace solve

tamakoshi2001:
**Checklist**
Thank you for contributing to QuTiP! Please make sure you have finished the following tasks before opening the PR.

- [ ] Please read [Contributing to QuTiP Development](http://qutip.org/docs/latest/development/contributing.html)
- [ ] Contributions to qutip should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/).
You can use [pycodestyle](http://pycodestyle.pycqa.org/en/latest/index.html) to check your code automatically
- [ ] Please add tests to cover your changes if applicable.
- [ ] If the behavior of the code has changed or new feature has been added, please also update the documentation in the `doc` folder, and the [notebook](https://github.com/qutip/qutip-tutorials). Feel free to ask if you are not sure.
- [ ] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#changelog-generation) for more information).

Delete this checklist after you have completed all the tasks. If you have not finished them all, you can also open a [Draft Pull Request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to let the others know this on-going work and keep this checklist in the PR description.

**Description**
Changed _

**Related issues or PRs**
Please mention the related issues or PRs here. If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id, e.g. fix #1184

PR #2144, opened 2023-03-31
Make `data.solve` more robust to singular matrices for `countstat`

Ericgig:
**Description**
In #2120, `countstat` was updated to use `data.solve`.
`countstat` would sometime try to solve a system with singular matrix and has a fallback for it using `try: ... except:`
However, while numpy raises an error when the matrix is singular, `scipy.sparse` raises a warnings, thus the tests only pass because we use the warnings as error flags... Without it, the singular matrix would not be caught and the solution from `countstat` would contain a `NaN` and the tests would fail...

- Redirect the warning from scipy sparse and numpy's `LinAlgError` to a `ValueError`to allow easy catching of these case.
- Add the small `1e-15` in `countstat` with `auto_tidyup` turned off.
- Add support for the method `lstsq` and `solve` for `data.solve_csr`. These methods are supported by numpy, jax, tensorflow, scipy.sparse is the exception which use `spsolve` and `lsqr` instead, probably to avoid name collision the the dense version they also have.

### DIFF
--- a/qutip/tests/core/data/test_linalg.py
+++ b/qutip/tests/core/data/test_linalg.py
@@ -74,8 +74,9 @@ class TestSolve():
     def test_singular(self):
         A = qutip.num(2).data
         b = qutip.basis(2, 1).data
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError) as err:
             test1 = _data.solve(A, b)
+        assert "singular" in str(err.value).lower()
 
 
     def test_incorrect_shape_non_square(self):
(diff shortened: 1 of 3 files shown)


qutip/tests/core/data/test_linalg.py
```diff
@@ -65,6 +71,13 @@ def test_mathematically_correct_Dense(self, method, opt):
                                    atol=1e-7, rtol=1e-7)
 
 
+    def test_singular(self):
+        A = qutip.num(2).data
+        b = qutip.basis(2, 1).data
+        with pytest.raises(ValueError):
+            test1 = _data.solve(A, b)
```
hodgestar, line 78: We should check the error message here. `ValueError` is raised by many things and we should check that the correct `ValueError` is raised and that we didn't accidentally trigger some other error path.

hodgestar (changes requested):
Only a small request to also check the error message. Otherwise looks good.

hodgestar (approved):

PR #2145, opened 2023-03-31
tomography uses itertools

Ericgig:
**Description**
Simplify `_index_permutations` using `itertools`.


hodgestar (approved):
I imagine this is also a big performance improvement in addition to being much clearer and simpler. Woot.

PR #2146, opened 2023-03-31
Remove logging utils

Ericgig:
**Description**
Remove `logging_utils`, no longer used in `qutip/qutip`.
It is still used in `qutip-qtrl` and displaced there in qutip/qutip-qtrl#9.

The entry `log_handler` of `settings` is kept since creating `settings` there just for this entry wouldn't be practical.


hodgestar (approved):
We should think about what logging should look like for the qutip family packages and that should dictate what happens to the `log_handler` entry in settings, but I'm happy with this PR being merged now. Logging needs a better plan, but maybe that should happen after 5.0.

hodgestar:
The test failure was just the random superoperator `iscptp` and `isherm` test that is being fixed in #2147.

Issue #2148, opened 2023-04-03
EnhancedTextProgressBar not coherent with docstring of BaseProgressBar


NathalyaaHikari:
### Bug Description

When initializing an `EnhancedTextProgressBar` object the `__init__` method is calling the `start` method of it's super class instead of it's own. When then using `update` (without calling `start` manually again) there will be an `AttributeError` because `width` is not getting initialized by `__init__`

### Code to Reproduce the Bug

```shell
from qutip.ui.progressbar import *

n = 10
a = EnhancedTextProgressBar(n)
# a.start(n)
for i in range(n):
    a.update(i)
a.finished()
```

### Code Output

```shell
Traceback (most recent call last):
  File "C:\Users\Name\Desktop\qutip_bug.py", line 6, in <module>
    a.update(i)
  File "C:\Users\Name\AppData\Local\Programs\Python\Python311\Lib\site-packages\qutip\ui\progressbar.py", line 96, in update
    all_full = self.width - 2
               ^^^^^^^^^^
AttributeError: 'EnhancedTextProgressBar' object has no attribute 'width'
```

### Expected Behaviour

Proper initialization of the `width` attribute by `EnhancedTextProgressBar.__init__`

### Your Environment

```shell
QuTiP Version:      4.7.1
Numpy Version:      1.24.2
Scipy Version:      1.10.1
Cython Version:     0.29.33
Matplotlib Version: 3.7.1
Python Version:     3.11.0
Number of CPUs:     8
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
Platform Info:      Windows (AMD64)
```

### Additional Context

Currently in master the `__init__` for both `TextProgressBar` and `EnhancedTextProgressBar` have been commented out and made to just pass. See https://github.com/qutip/qutip/blob/af49884aa1db47a980626ac00adfe83277d8d261/qutip/ui/progressbar.py#L66 and https://github.com/qutip/qutip/blob/af49884aa1db47a980626ac00adfe83277d8d261/qutip/ui/progressbar.py#L94

I would suggest making them both call their own `start` method to omit an additional call of start by the user right after initialization. This would also make both classes be in line with the usage suggested by the docstring of `BaseProgressBar.` Otherwise I would suggest to include the call to start in said docstring.

As I'm pretty new to github and git in general I would like to try and create a PR myself once you give me approval for the suggested change here.

Ericgig:
Fixed in v5

PR #2127, opened 2023-03-18
Add tests for progress_bar 

HarshKhilawala:
**Checklist**
Thank you for contributing to QuTiP! Please make sure you have finished the following tasks before opening the PR.

- [x] Please read [Contributing to QuTiP Development](http://qutip.org/docs/latest/development/contributing.html)
- [x] Contributions to qutip should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/).
You can use [pycodestyle](http://pycodestyle.pycqa.org/en/latest/index.html) to check your code automatically
- [x] Please add tests to cover your changes if applicable.
- [x] If the behavior of the code has changed or new feature has been added, please also update the documentation in the `doc` folder, and the [notebook](https://github.com/qutip/qutip-tutorials). Feel free to ask if you are not sure.
- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#changelog-generation) for more information).

**Description**
Add tests for ui.progressbar

PR #2150, opened 2023-04-11
Add a test for progress bar

Ericgig:
**Description**
- Add a test for progress bars.
- Make `start` private. `start` was mostly used right after the initialization. The docstring indicated that it was not needed, but it was for most bars. #2148
- Remove the iteration argument in update.
- Have `HTMLProgressBar` import ipython.

**Related issues or PRs**
Fix the issue #2148 for master.
Replace #2127

### DIFF
--- a/qutip/ipynbtools.py
+++ b/qutip/ipynbtools.py
@@ -273,14 +273,14 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
     else:
         if progress_bar is True:
             progress_bar = HTMLProgressBar(len(ar_list))
-
-        n = len(ar_list)
+        prev_finished = 0
         while True:
             n_finished = sum([ar.progress for ar in ar_list])
-            progress_bar.update()
+            for _ in range(prev_finished, n_finished):
+                progress_bar.update()
+            prev_finished = n_finished
 
             if view.wait(ar_list, timeout=0.5):
-                progress_bar.update()
                 break
         progress_bar.finished()
 
--- a/qutip/ipynbtools.py
+++ b/qutip/ipynbtools.py
@@ -272,16 +272,15 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
         view.wait(ar_list)
     else:
         if progress_bar is True:
-            progress_bar = HTMLProgressBar()
+            progress_bar = HTMLProgressBar(len(ar_list))
 
         n = len(ar_list)
-        progress_bar.start(n)
         while True:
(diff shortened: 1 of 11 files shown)


qutip/solve/pdpsolve.py
```diff
@@ -200,6 +197,9 @@ def __init__(self, H=None, state0=None, times=None, c_ops=[], sc_ops=[],
         else:
             self.map_func = serial_map
 
+        if progress_bar is None:
+            self.progress_bar = TextProgressBar(self.ntraj)
```
hodgestar, line 201: Now that the progress bar constructor starts the progress bar, we should construct the progress bar where it is used.

qutip/tests/test_progressbar.py
```diff
@@ -0,0 +1,34 @@
...
+
+@pytest.mark.parametrize("pbar", names)
+def test_progressbar(pbar, capsys):
+    N = 5
+
+    try:
+        bar = progress_bars[pbar](N)
+    except ImportError:
+        # ipython or tqdm could be missing
+        pytest.skip(reason="module not available")
```
hodgestar, line 23: We should probably list the progress bars explicitly and mark which ones to skip. Otherwise this will skip tests if something fails to import and it might take a long time to realize.

hodgestar (changes requested):
Thanks for sorting this out.

All the calls to construct progress bar should move to where `.start()` was originally. I would also prefer if the tests where a bit more explicit about what they were testing. Having tests decide what to test based on the code is a bad idea, because the code can easily be wrong (e.g. we might forget to add something to the list of progress bars) and it's good to have a pattern of "if the code is changed, the tests should change too".

qutip/ipynbtools.py
```diff
@@ -272,16 +272,15 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
         view.wait(ar_list)
     else:
         if progress_bar is True:
-            progress_bar = HTMLProgressBar()
+            progress_bar = HTMLProgressBar(len(ar_list))
 
         n = len(ar_list)
```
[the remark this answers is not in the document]
Ericgig: I am hoping to remove that `parallel_map` function. It cannot be used interchangeably with other functions.

qutip/solve/pdpsolve.py
Ericgig: This file is removed in #2131.
As well as `solve/stochastic.py`.

hodgestar (approved):

PR #2151, opened 2023-04-12
Enable cython 3

### LINKED ISSUE #2117: Prepare qutip for cython 3.0
### Describe the Issue!

Cython is getting close to release a new major release. It bring 2 changes that we must asses to be compatible with it.
- The order of the `nogil` and `except` keyword. In cython 0.29, both `nogil except ...` and `except ... nogil` are supported. From 3.0, a warning is raised when the first is used. We used the first order almost everywhere. We need to reorder them all... 
- Cython 0.29 do not use reverse operations such as  `__rmatmul__` and `__rmul__`, but call `__mul__` with reverse order (`__mul__(other, self)`). With cython 3, `__rmul__` is used and `self` is always the first input. This breaks some of our tests.  The failing tests are related to `Coefficient.__mul__` in qutip/core/cy/coefficient.pyx,  `Data.__mul__` in qutip/core/data/base.pyx and `_ConstantElement.__matmul__` and  `_EvoElement.__matmul__` both in qutip/core/cy/_element.pyx.  While never used, the `__mul__` methods in the last file would break if called in the wrong order.
- One of the automated tests in github action should run using cython 3.

ps. This could be solved in 2 PR, an easy one for the first bullet point and another one more involved for the 2 others.

Ericgig:
**Description**
Ensure qutip works with cython 3.0.
- `nogil` must come after `except`.
- Set the options `c_api_binop_methods=True` to keep previous behaviour of operators.
- Have one of the test case use cython 3.

There is a bug in cython 3.0.0b2 with `overflowcheck` that makes it unable to compile qutip. I works with 3.0.0.b1 and should work with the next release. See cython#5358

**Related issues or PRs**
close #2117


Ericgig:
This require cython>=0.29.32...

.github/workflows/tests.yml
```diff
@@ -115,6 +116,9 @@ jobs:
           if [[ -z "${{ matrix.nocython }}" ]]; then
             QUTIP_TARGET="$QUTIP_TARGET,runtime_compilation"
           fi
+          if [[ -z "${{ matrix.cython3 }}" ]]; then
+            pip install cython --pre
```
Ericgig, line 120:
```suggestion
            pip install cython --pre --upgrade
```

Ericgig:
@hodgestar 
Instead of waiting for the fix to be released in cython, I commented the missing feature: data specializations' `__module__`.

hodgestar (approved):

PR #2152, opened 2023-04-12
Move `Memorycascade` to legacy

Ericgig:
**Description**
- Fix memory cascade to work in v5.
- Move it to a new legacy folder. 
  I would like to remove it, but I don't know if anybody uses it and it can't be replaced by another solver.
  However, it create evolution with jump in the output coming from numerical methods, see https://github.com/qutip/qutip-notebooks/blob/master/examples/nonmarkov-coherent-feedback.ipynb.


nwlambert:
yep just a note that i will take a look at this. I was using a very similar method for the problem in the example notebook, will take a closer look at this and get back with some thoughts on whether to keep it or not and a potential test.

ajgpitch:
Perhaps we could ask Arne himself if he wants to write a test for it? I am a bit surprised that he did not submit it with a notebook at least. @nwlambert are you still in touch with Arne?

ajgpitch:
I think this is a valuable piece of qutip and we should try to keep it in if possible. Comparing with other non-Markov solvers seems like an interesting idea. But I agree with @Ericgig that if we don't have tests and a note book then it should be moved to legacy.

hodgestar (approved):

.coveragerc
```diff
@@ -5,6 +5,8 @@ omit =
     */qutip/tests/*
     # Tool for tests
     */qutip/solver/sode/_noise.py
+    # Feature we know are not fully tested.
+    */qutip/legacy/*
```
Ericgig, line 9:
```suggestion
```

PR #2154, opened 2023-04-14
Use `qeye_like`

Ericgig:
**Description**
Change `qeye` to `qeye_like` and `qzero` to `qzero_like` when appropriate.

`qeye_like` is now able to accept `QobjEvo` to use it in propagator etc.
Fix the bug in propagator with super-operator input found in #2152...

### DIFF
--- a/qutip/solver/countstat.py
+++ b/qutip/solver/countstat.py
@@ -89,7 +89,7 @@ def _noise_direct(L, wlist, rhoss, J_ops):
     current = np.zeros(N_j_ops)
     noise = np.zeros((N_j_ops, N_j_ops, len(wlist)))
 
-    tr_op = tensor(qeye(L.dims[0][0]))
+    tr_op = qeye(L.dims[0][0])
     tr_op_vec = operator_to_vector(tr_op)
 
     Pop = _data.kron(rhoss_vec, tr_op_vec.data.transpose())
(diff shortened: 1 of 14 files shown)


qutip/solver/countstat.py
```diff
@@ -89,7 +89,7 @@ def _noise_direct(L, wlist, rhoss, J_ops):
     current = np.zeros(N_j_ops)
     noise = np.zeros((N_j_ops, N_j_ops, len(wlist)))
 
-    tr_op = tensor([qeye(n) for n in L.dims[0][0]])
+    tr_op = tensor(qeye(L.dims[0][0]))
```
hodgestar, line 92:
```suggestion
    tr_op = qeye(L.dims[0][0])
```
Can we remove the tensor here? This was pretty odd code to start with.

qutip/solver/propagator.py
```diff
@@ -61,12 +61,15 @@ def propagator(H, t, c_ops=(), args=None, options=None, **kwargs):
     if not isinstance(H, (Qobj, QobjEvo)):
         H = QobjEvo(H, args=args, **kwargs)
 
-    if H.issuper or c_ops:
-        out = mesolve(H, qeye(H.dims), tlist, c_ops=c_ops,
```
hodgestar, line 65: Did this code do the wrong thing when `c_ops` were present?

hodgestar (commented):
I left a couple of questions, but looks good overall.

There are a couple of places where we not create intermediate density matrices instead of using the dims of a state, but I'm happy with that. If anything specific case is a performance issue it will be relatively easy to fix.

qutip/solver/propagator.py
Ericgig: The code was wrong when `H` was a super operator. It that case it should have been `qeye(H.dims[0])`.

hodgestar (approved):

PR #2157, opened 2023-04-19
Add dispatcher capacity to support any type specialisation

Ericgig:
**Description**
Allow the dispatcher to accept specialisations on generic `Data`.
This as 2 uses:
- Simplify the creation of complex function using simpler ones: `inner` defined in term of `matmul`. 
- Allow us to create niche dispatched functions for some data layer without the need to be supported by all: blas' zgemm (matmul) can transpose and adjoint of each input matrix in the same operation. 

The `Data` dispatched functions have less priority than direct specialization. They are used directly (no `_constructed_specialisation`) when the `Data` is in the inputs, but when the output data type is set, there will always be a conversion from `Data` to the desired type. 

In this PR:
- Add dispatch on `Data`.
- Add direct tests for the dispatcher.
- Create `kron_transpose` which is `kron(left.T, tright)`. It's commonly used for superoperator and it can be optimized for `Dense` and `Jax` by doing it in one operation.
- Add `Data` specialization for `inner`, `inner_op`, `expect`, `expect_super` and `trace_oper_ket`.
- Make `zeros_like` and `identity_like` as dispatched function so the order is kept during operation for dense.

### DIFF
--- a/qutip/core/data/convert.pyx
+++ b/qutip/core/data/convert.pyx
@@ -24,6 +24,44 @@ from qutip.core.data.base cimport Data
 __all__ = ['to', 'create']
 
 
+class _Epsilon:
+    """
+    Constant for an small weight non-null weight.
+    Use to set `Data` specialisation just over direct specialisation.
+    """
+    def __repr__(self):
+        return "EPSILON"
+
+    def __eq__(self, other):
+        if isinstance(other, _Epsilon):
+            return True
+        return NotImplemented
+
+    def __add__(self, other):
+        if isinstance(other, _Epsilon):
+            return self
+        return other
+
+    def __radd__(self, other):
+        if isinstance(other, _Epsilon):
+            return self
+        return other
+
+    def __lt__(self, other):
+        """ positive number > _Epsilon > 0 """
+        if isinstance(other, _Epsilon):
+            return False
+        return other > 0.
+
+    def __gt__(self, other):
+        if isinstance(other, _Epsilon):
+            return False
+        return other <= 0.
+
+
+EPSILON = _Epsilon()
+
+
 def _raise_if_unconnected(dtype_list, weights):
     unconnected = {}
     for i, type_ in enumerate(dtype_list):
@@ -152,7 +190,6 @@ cdef class _to:
     cdef dict _convert
     cdef readonly dict weight
     cdef readonly dict _str2type
-    cdef readonly float anydataweight
 
     def __init__(self):
         self._direct_convert = {}
@@ -161,7 +198,6 @@ cdef class _to:
         self.weight = {}
         self.dispatchers = []
         self._str2type = {}
-        self.anydataweight = 0.001
 
     def add_conversions(self, converters):
         """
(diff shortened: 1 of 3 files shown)


qutip/core/data/convert.pyx
```diff
@@ -69,19 +70,23 @@ cdef class _converter:
                 + ">")
 
 
+def dummyconverter(arg):
+    return arg
```
hodgestar, line 74:
```suggestion
def identity_converter(arg):
    return arg
```
This is a real converter. :)

qutip/core/data/convert.pyx
```diff
@@ -258,6 +267,11 @@ cdef class _to:
                 self.weight[(to_t, from_t)] = weight
                 self._convert[(to_t, from_t)] =\
                     _converter(convert[::-1], to_t, from_t)
+        for dtype in self.dtypes:
+            self.weight[(dtype, Data)] = 1.
+            self.weight[(Data, dtype)] = self.anydataweight
```
hodgestar, line 272: The constant here is rather awkward.

Could we add an small implementation of an epsilon value? Something like:
```
class _Epsilon:
    def __repr__(self):
        return "EPSILON"

    def __eq__(self, other):
        if isinstance(other, _Epsilon):
            return True
        return NotImplemented

    def __add__(self, other):
        if isinstance(other, _Epsilon):
            return self
        return other

EPSILON = _Epsilon()
```
Then we can check whether a weight is all epsilons by doing `if weight == EPSILON`.

Alternatively, could we just check whether the weight is `0.0`?

Maybe there are some practical issues here I am not seeing yet.

hodgestar (commented):
I left a few small comments, but overall this looks great.

qutip/core/data/convert.pyx
Ericgig: There are issues with using `0.0`, but `_Epsilon` should work fine.

hodgestar: Could you write a sentence about the issues with `0.0` so that I understand a bit the issues?

Ericgig: When there are both a specific specialization and a general one:
(`kron_dense` vs `kron_data`)
We want `kron_dense` to have a lower weight so it does not decide to use `kron_data` instead.
`kron_dense`'s weight is zero, so I used an arbitrary small number.

qutip/core/cy/_element.pyx
```diff
@@ -594,7 +594,7 @@ cdef class _ProdElement(_BaseElement):
         cdef double complex out = self._left.coeff(t) * self._right.coeff(t)
         return conj(out) if self._conj else out
 
-    cdef Data matmul_data_t(_ProdElement self, t, Data state, Data out=None):
+    cpdef Data matmul_data_t(_ProdElement self, t, Data state, Data out=None):
```
hodgestar, line 597: What caused all the changes to `cpdef`?

Ericgig: Sorry, I was multitasking.
This wasn't mean to be here.

qutip/core/data/convert.pyx
hodgestar: Ah, that makes sense.

PR #2017, opened 2022-11-03
edits for working latex renders in vscode

drodper:
**Description**
Changes the string generated by `_repr_latex()` in `qobj.py` so that it plays nicely with VSCode jupyter notebooks. 
Specifically, changes the output to be encased in `$...$`, and changes from `\begin{equation} \begin{array}...` to `\begin{matrix} ...`

This lets matrices be output nicely in VSCode.

**Related issues or PRs**
fixes #1887

PR #2167, opened 2023-05-28
Dev uhack solver docs

### LINKED ISSUE #2163: Solver options
### Problem Description

The options that can be passed to the solvers are not entirely clear by looking at the documentation. For example:
image
Does not show the explicit options even thought it refers to them in the parameter documentation. For comparison, SESolver does show the options.

This I assume is because the options are inherited from SESolver. But this is not clear either by looking at `MESolver`'s rendered documentation.
image

When using `help(MESolver)` the documentation of the class is much clearer but perhaps too verbose for the online documentation.

### Proposed Solution

I would suggest we render the options property explicitly for each solver. I would also consider rendering the documentation related to the `run` method. If not, we should render at least `Solver` documentation which does not seem to be listed in the class API documentation for solvers.

Affected solvers:
- [ ] MESolver
- [ ] SMESolver
- [ ] SSESolver
- [ ] NonMarkovianMCSolver

### Alternate Solutions

_No response_

### Additional Context

The documentation was taken from qutip 5.0 "ReadTheDocs".

EmilianoG-byte:
**Description**
Add inherited members (properties and methods) in the API docs for the solvers mentioned in the issue.
Add inheritance of these solvers showing from which base classes they inherit.

**Related issues or PRs**
fixes #2163

### DIFF
--- a/doc/apidoc/classes.rst
+++ b/doc/apidoc/classes.rst
@@ -46,18 +46,25 @@ Solvers
 
 .. autoclass:: qutip.solver.sesolve.SESolver
     :members:
-
+    
 .. autoclass:: qutip.solver.mesolve.MESolver
     :members:
-
+    :inherited-members:
+    :show-inheritance:
+      
 .. autoclass:: qutip.solver.brmesolve.BRSolver
     :members:
 
+
 .. autoclass:: qutip.solver.stochastic.SMESolver
     :members:
+    :inherited-members:
+    :show-inheritance:
 
 .. autoclass:: qutip.solver.stochastic.SSESolver
     :members:
+    :inherited-members:
+    :show-inheritance:
 
 
 
@@ -71,6 +78,8 @@ Monte Carlo Solvers
 
 .. autoclass:: qutip.solver.nm_mcsolve.NonMarkovianMCSolver
     :members:
+    :inherited-members:
+    :show-inheritance:
 
 
 .. _classes-non_markov_heom:
--- a/doc/apidoc/classes.rst
+++ b/doc/apidoc/classes.rst
@@ -46,6 +46,8 @@ Solvers
 
 .. autoclass:: qutip.solver.sesolve.SESolver
     :members:
+    :inherited-members:
+    :show-inheritance:
     
 .. autoclass:: qutip.solver.mesolve.MESolver
     :members:
@@ -54,6 +56,8 @@ Solvers
       
 .. autoclass:: qutip.solver.brmesolve.BRSolver
     :members:
+    :inherited-members:
+    :show-inheritance:
 
 
 .. autoclass:: qutip.solver.stochastic.SMESolver
@@ -75,6 +79,8 @@ Monte Carlo Solvers
 
 .. autoclass:: qutip.solver.mcsolve.MCSolver
     :members:
+    :inherited-members:
+    :show-inheritance:
 
 .. autoclass:: qutip.solver.nm_mcsolve.NonMarkovianMCSolver
     :members:
--- a/doc/changes/2163.doc
+++ b/doc/changes/2163.doc
@@ -1,1 +1,0 @@
-added inherited members to API doc of solvers: MESolver, SMESolver, SSESolver, NonMarkovianMCSolver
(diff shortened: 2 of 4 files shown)


EmilianoG-byte:
I had also the idea that showing from which base classes these solvers inherit would be enough, but since the issue asks to show `property` and `run` explicitly on each of the solver docs, then I opted do to it this way.

Also, from my implementation, also other methods are now documented (e.g. add_integrator).  Which I believe since these are public methods, they should be documented as well (?)

EmilianoG-byte:
I see there is an error when building the pdf but don't understand why. Running the `make pdflatex` command works for me locally. Also the html page is rendered as expected:

Screenshot 2023-05-29 at 11 56 07

BoxiLi:
In the CI test, we convert warnings to errors. Have you tried `make latexpdf -W`?

EmilianoG-byte:
I just ran locally:
`make latexpdf SPHINXOPTS="-W --keep-going -T"`
And got `build succeeded` message.
The error on the project pipeline shows:
 `Unknown interpreted text role "cls".` for the `me_solver`.

Do you have an idea of why could this be happening?

EmilianoG-byte:
resolved 😄 !

Ericgig (approved):
Looks good, `options` property is properly displayed for all solver, thank you.

EmilianoG-byte:
Awesome :D, do I just close the PR then?

Ericgig:
No, we will merge it.

AGaliciaMartinez (changes requested):
Thanks for the PR! At the begining I was thinking to include only the `run` and `options` method but I think including all the other ones is also fine. However, I do think that if we are including all methods we should do so consistently for all the classes. This would require to do the same changes you have done to the rest of the classes to `SESolver` and `MCSolver`. After that I think the PR is ready to Merge. Thanks again for the changes!!

AGaliciaMartinez (approved):
Thanks for the changes!

EmilianoG-byte:
Thank you for the comments! 

Also, as someone who uses QuTip on a daily basis as part of my work, I was really happy with the changes and restructuring of the solver module. Thank you for the work 😄 

If I may take this opportunity to ask two things:
* I have been looking on Qutip 4.7 for ways to speed up the mc solver simulations, and one of them was to not compute the average density matrix for all times, but only for the last time. Since the behaviour at 4.7 of `mcsolve` was to average the state anyways (despite what I passed on the options object), I found a workaround by having an independent list to create my time dependant Hamiltonian and the `tlist` argument of the solver. Is this something achievable more straightforwardly/differently in Qutip 5.0?
* (shorter): Is there a roadmap for when will Qutip 5.0 be released?

Thank you for your time, and I hope its fine me asking here 😆

EmilianoG-byte:
Btw, there seems to a small bug. When running:

```
from qutip.solver.mcsolve import MCSolver, mcsolve
import numpy as np
from qutip import tensor, fock, qeye, destroy

times = np.linspace(0.0, 10.0, 200)
psi0 = tensor(fock(2, 0), fock(10, 8))
a  = tensor(qeye(2), destroy(10))
sm = tensor(destroy(2), qeye(10))
H = 2*np.pi*a.dag()*a + 2*np.pi*sm.dag()*sm + 2*np.pi*0.25*(sm*a.dag() + sm.dag()*a)
data2 = mcsolve(H, psi0, times, [np.sqrt(0.1) * a])
data2.states[0]
```
The output is not rendered in latex bur rather shows the raw code:
Screenshot 2023-06-02 at 01 37 59

In case you want  to create an `issue` for it, perhaps.

BoxiLi:
It looks like VSCode with Jupyter notebook terminal. I also have this problem. We had this issue solved once in https://github.com/qutip/qutip/pull/2017 but it seems to be back again. You are very welcome to open a separate issue for this.

EmilianoG-byte:
I have added the issue. I have spotted already some suggestions of what can be going wrong and note them down on the issue details!

EmilianoG-byte:
Also, @Ericgig please let me know if you have time to answer my question regarding the MC solver. I would greatly appreciate it😄

EmilianoG-byte:
@Ericgig, I see. Thank you so much! I had the impression that v5 would change something related to the calculation of density matrices being more flexible. But all in order now. Thank you!

PR #2169, opened 2023-05-31
Add signal check in vern integrator

Ericgig:
**Description**
In cython code, signals (Keyboard Interrupt) are not automatically checked.
And when checked, if error are not properly passed (`expect *`) the signal can be last.

This happened in the Runge Kutta integrator, which did not properly passed errors and only when back to python space at time in `tlist` which can be far apart.

The stochastic solver stop on (Keyboard Interrupt), the `except` are not needed since the output is a class. I made it explicit for clarity.

Doing a few tests, I saw that the error was often caught in the `Qobj.data` property called by `QobjEvo.matmul`. This means more back and forth between python and cython was done than needed.

### DIFF
--- a/qutip/solver/integrator/explicit_rk.pyx
+++ b/qutip/solver/integrator/explicit_rk.pyx
@@ -11,6 +11,7 @@ from qutip.core.data.tidyup import tidyup_csr
 from qutip.core.data.norm import frobenius_data
 from .verner7efficient import vern7_coeff
 from .verner9efficient import vern9_coeff
+from cpython.exc cimport PyErr_CheckSignals
 cimport cython
 import numpy as np
 
@@ -198,7 +199,7 @@ cdef class Explicit_RungeKutta:
         self.b_factor_np = np.empty(self.rk_extra_step, dtype=np.float64)
         self.b_factor = self.b_factor_np
 
-    cpdef void set_initial_value(self, Data y0, double t):
+    cpdef void set_initial_value(self, Data y0, double t) except *:
         """
         Set the initial state and time of the integration.
         """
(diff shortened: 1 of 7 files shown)


qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -300,6 +301,7 @@ cdef class Explicit_RungeKutta:
             self._t_prev = self._t_front
             self._norm_prev = self._norm_front
             nsteps_left -= self._step_in_err(t, nsteps_left)
+            PyErr_CheckSignals()
```
hodgestar, line 304: `PyErr_CheckSignals` returns `-1` if a Python signal handler raises an exception. Should we check for that and return?

hodgestar (commented):
I left one question about the error return from `PyErr_CheckSignals` but the rest of the changes look like nice improvements.

qutip/solver/integrator/explicit_rk.pyx
Ericgig: It also raise an error which is what we need.
If we check for that and return ourselves, wouldn't the error trace be lost?

hodgestar: At least in C code, one would do:
```
if (PyErr_CheckSignals == -1) {
    return;
}
```
which would then return `void` and the caller would check to see if an error was set.

So my question is does Cython do this for us automatically?

Ericgig: Yes cython catch the error and return with the error value without executing the rest of the function by itself.
So it should never reach the return.
Also you probably don't want to return void. With `except *`, I believe that it check for error with any return, but with with `except -1`, `-1` should be returned etc. It's better to just let cython handle that.

hodgestar (approved):

PR #2173, opened 2023-06-05
Add `herm_matmul` to speed up `mesolve`.

Ericgig:
**Description**
Add a variant to `matmul` specialised for case where the right matrix is a column stacked hermitian matrix, and the output is the same. Such as for `mesolve` using super operators.

Open solvers `mesolve`, `brmesolve`, `smesolve` and `fmesolve` all have a new options `use_herm_matmul` to use this operation. The default is `False` since we cannot easily check for exception (we can't test that a time dependant Hamiltonian is Hermitian, ...).

With large enough systems, this result in a visible speed up:
```
>>> N = 100
>>> H = qt.rand_herm(N, density=3/N)
>>> c_ops = [qt.destroy(N)]

>>> qt.mesolve(H, qt.basis(N, N-2), [0, 10], c_ops=c_ops, options={"use_herm_matmul": True}).stats["run time"]
0.8431086540222168
>>> qt.mesolve(H, qt.basis(N, N-2), [0, 10], c_ops=c_ops, options={"use_herm_matmul": False}).stats["run time"]
1.0559077262878418

qt.smesolve(H, qt.basis(N, N-2), [0, 2], sc_ops=c_ops, ntraj=1, options={"use_herm_matmul": True}).stats["run time"]
0.9043550491333008
qt.smesolve(H, qt.basis(N, N-2), [0, 2], sc_ops=c_ops, ntraj=1, options={"use_herm_matmul": False}).stats["run time"]
2.111673355102539
```

This is a use case for the capacity to dispatch on `Data` added in #2157. When `herm_matmul` is not available, it can be better to fallback on `matmul` using the same type than to do conversions between data types. For cupy, moving the data to the cpu to halves the work is probably not worth it.

 
**Related issues or PRs**
This is build on top of #2157, it should be merged first.

### DIFF
--- a/qutip/tests/core/data/test_specialized_operations.py
+++ b/qutip/tests/core/data/test_specialized_operations.py
@@ -11,7 +11,7 @@ import qutip
     (Dia, Dense, data.herm_matmul_dia_dense_dense),
     (Dense, CSR, data.herm_matmul_data),
 ])
-@pytest.mark.parametrize("N", [1, 2, 5, 10, 25])
+@pytest.mark.parametrize("N", [2, 5, 10, 25])
 def test_herm_matmul(N, op_type, state_type, function):
     H = qutip.rand_herm(N, dtype=op_type)
     a = qutip.destroy(N, dtype=op_type)


qutip/core/data/herm_matmul.pyx
```diff
@@ -0,0 +1,241 @@
...
+herm_matmul.__doc__ =\
+    """
+    Compute the matrix multiplication of two matrices, with the operation
+        scale * (left @ right)
+    where `scale` is (optionally) a scalar, and `left` and `right` are
+    matrices. The output matrix should be a column stacked Hermitian matrix.
+    This is not tested but only half of the matrix product is computed. The
+    other half being filled with the conj of the first.
+
+    Intented to be used in ``mesolve``.
```
hodgestar, line 214:
```suggestion
    Intended to be used in ``mesolve``.
```

qutip/core/data/herm_matmul.pyx
```diff
@@ -0,0 +1,241 @@
...
+
+    if type(out) is Dense:
+        inplace_out = out
+        out = None
+    else:
+        inplace_out = dense.zeros(left.shape[0], right.shape[1], right.fortran)
+
+    # cdef idxint row, ptr, idx_r, idx_out, dm_row, dm_col
+    # cdef idxint nrows=left.shape[0], ncols=right.shape[1]
+    # cdef double complex val
```
hodgestar, line 99: Should the commented out cdefs be removed?

hodgestar (commented):
I'd like to take a more detailed look at this once the Data specialization PR is merged, but the concept looks good.

Would you mind pinging me for another review here once the other PR is done?

qutip/solver/stochastic.py
```diff
@@ -521,7 +533,7 @@ def m_ops(self, new_m_ops):
 
         Default are:
 
-            m_ops = sc_ops + sc_ops.dag()
+            m_ops = sc_ops + sselfc_ops.dag()
```
hodgestar, line 536:
```suggestion
            m_ops = sc_ops + sc_ops.dag()
```

qutip/core/data/herm_matmul.pyx
hodgestar, line 99:
```suggestion
```

qutip/core/cy/qobjevo.pyx
```diff
@@ -1017,3 +1017,32 @@ cdef class QobjEvo:
...
+        t = self._prepare(t, state)
+        if out is None and type(state) is Dense:
+            out = dense.zeros(self.shape[0], state.shape[1],
+                      (<Dense> state).fortran)
+        elif out is None:
+            out = _data.zeros[type(state)](self.shape[0], state.shape[1])
+
+        for element in self.elements:
+            part = (<_BaseElement> element)
+            out = part.matmul_data_t_herm(t, state, self.subsize, out)
```
hodgestar, line 1047: This requires each matrix element of the `QobjEvo` to be Hermitian, which is neither necessary not sufficient for the `qobjevo(t)` to be Hermitian.

On the up side, it is fairly easy to check whether each matrix element in a QobjEvo is hermitian, so perhaps we could have each element use the hermitian version automatically if it knows its Qobj is hermitian?

Ericgig: This is not meant to be used when `qobjevo(t)` to be Hermitian. 
The output of `vector_to_operator(qobjevo(t) @ operator_to_vector(rho))` should be hermitian.
Since the `out` is only accumulated on, it does not need to be true for every element independently, just the result.

qutip/solver/brmesolve.py
```diff
@@ -351,6 +372,15 @@ def _apply_options(self, keys):
...
             self.rhs = self._prepare_rhs()
 
+        if hasattr(self, "_rhs") and "use_herm_matmul" in keys:
+            self._rhs = self._update_rhs()
+            # Ensure the integrator is reset
+            keys = set(keys)
+            keys.add('method')
+
+        if hasattr(self, "_rhs") and "use_herm_matmul" in keys:
+            self._rhs = self._update_rhs()
```
hodgestar, line 382: Are these two lines meant to be duplicated here? It seems odd to call `self._update_rhs()` twice.

qutip/solver/brmesolve.py
```diff
@@ -274,6 +280,7 @@ def __init__(self, H, a_ops, c_ops=None, sec_cutoff=0.1, *, options=None):
         self._num_collapse = len(c_ops)
         self._num_a_ops = len(a_ops)
         self.rhs = self._prepare_rhs()
+        self._rhs = self._update_rhs()
```
hodgestar, line 283: Duplicating the RHS here feels very messy. Could we just store a single RHS as `self.rhs` and convert to and from `QobjEvoHerm` as needed when options are updated?

hodgestar (commented):
I left a couple more comments. On the whole I'm not wild about the extra QobjEvo class, but if this is a use case you'd like to support, what about the following suggestion:

We allow the user to supply the `QobjEvo` class to use as an option, and allow them to specify either the class itself or a name for the class. E.g. `qobjevo_cls=QobjEvoHerm` or `qobjevo_cls=herm`.

That allows many different classes to be used without the solvers needing to have special support added each time.

Perhaps we should also not allow users to switch the class later by changing options? That might simplify the logic because we wouldn't need to modify the RHS when options are updated.

Ericgig:
> We allow the user to supply the `QobjEvo` class to use as an option, and allow them to specify either the class itself or a name for the class. E.g. `qobjevo_cls=QobjEvoHerm` or `qobjevo_cls=herm`.

I would prefer the user not needing to know how we do it, just that there an option that speed up the simulation by 40 % in normal cases. Knowing that we forced it in an alternative qobjevo class is not useful and I hope we won't be forced to add many kinds of qobjevo.

> Perhaps we should also not allow users to switch the class later by changing options? That might simplify the logic because we wouldn't need to modify the RHS when options are updated.

Since changing options does not change the physic, I would like them to be changeable. But the `rhs`, and `_rhs` is certainly not great. I will rethink the way to do it.

qutip/tests/core/data/test_specialized_operations.py
```diff
@@ -0,0 +1,25 @@
...
+import qutip
+
+
+@pytest.mark.parametrize(["op_type", "state_type", "function"], [
+    (Dense, Dense, data.herm_matmul_dense),
+    (CSR, Dense, data.herm_matmul_csr_dense_dense),
+    (Dia, Dense, data.herm_matmul_dia_dense_dense),
+    (Dense, CSR, data.herm_matmul_data),
+])
+@pytest.mark.parametrize("N", [1, 2, 5, 10, 25])
```
Ericgig, line 14:
```suggestion
@pytest.mark.parametrize("N", [2, 5, 10, 25])
```

Ericgig:
@hodgestar 

I removed the `_rhs` by changing it so the rhs is built when making the integrator instead of in `__init__`.

I also simplified the way options are updated so brmesolve no longer need to overwrite `_apply_options`.

qutip/solver/solver_base.py
```diff
@@ -39,12 +39,13 @@ class Solver:
...
         "method": "adams",
     }
+    _reset_options = {"method"}
     resultclass = Result
 
     def __init__(self, rhs, *, options=None):
         if isinstance(rhs, (QobjEvo, Qobj)):
             self.rhs = QobjEvo(rhs)
-        else:
+        elif rhs is not None:
```
hodgestar, line 48: All subclasses of `Solver` should have one path for building the RHS. Having two paths makes the code much harder to reason about.

I think `_build_rhs()` is a good idea, although I think we should implement it on all sub-classes of `Solver` and have `Solver` do `self.rhs = self._build_rhs()` at the end of `__init__` and when options are changed (if needed). That way `.rhs` is always correct for the given options.

PR #2195, opened 2023-07-13
Remove `solve` folder

Ericgig:
**Description**
Remove the `solve` folder.
All that was remaining was piqs which was moved to it's own directory.


hodgestar (approved):
Yay! Good to have this gone at last.

PR #2196, opened 2023-07-14
`Dia` data layer

Ericgig:
**Description**
New sparse type: `Dia`.
Optimal for diagonal operators such as `qeye`, `destroy`. But it's quite bad for vectors ('bra', 'ket').

For a tri-diagonal matrix, it's about 25% faster than `CSR` for `oper @ dense`.
And while not implemented yet, it can do an very efficient `dense @ dia`, while `CSR` is only efficient `csr @ dense`, not the opposite. 

This include most specializations, change the default in operators, add tests.

The implementation mostly match the one in scipy. 
The matrices are stored in two arrays, the diagonals offsets `shape=num_offsets` and the data `shape=[num_offsets, ncols]`. The diagonals are lined so that for each entry, the column in the data array match the matrix column.
The main differences with scipy's implementation are fixed dtype and the shape of the data array being fixed. (scipy accept longer diagonals and ignore entries outside the matrix.)


Ericgig:
Here are some timing:
Operators are all 100x100

Sparse matrices composed of one diagonal:

|             | C       | Fortran |
|-------------|---------|---------|
| dia @ dense | 14.2 µs | 14.7 µs |
| dense @ dia | 14.2 µs | 12.8 µs |
| csr @ dense | 49.8 µs | 33.3 µs |
| dense @ csr | 165 µs  | 113 µs  |

Here for matrices with 5 diagonals:

|             | C       | Fortran |
|-------------|---------|---------|
| dia @ dense |  67 µs | 63 µs |
| dense @ dia | 60 µs | 55 µs |
| csr @ dense | 236 µs | 85 µs |
| dense @ csr | 270 µs  | 218 µs  |

qutip/core/_brtools.pyx
```diff
@@ -201,7 +201,7 @@ cdef class _EigenBasisTransform:
     def __init__(self, QobjEvo oper, bint sparse=False):
         if oper.dims[0] != oper.dims[1]:
             raise ValueError
-        if type(oper(0).data) is _data.CSR and not sparse:
+        if type(oper(0).data) in [_data.CSR, _data.Dia] and not sparse:
```
hodgestar, line 204:
```suggestion
        if type(oper(0).data) in (_data.CSR, _data.Dia) and not sparse:
```
Maybe slightly better for this to be a tuple?

hodgestar (approved):
I didn't do a thorough review, but we have lots of tests for the data layer and I can see you extended them to the new `dia` layer.

Happy for this to be merged.

Should we add it to the QuTiP benchmarking suite?

Ericgig:
Yes, let's add it to the benchmarks.

PR #2202, opened 2023-07-20
cython 3 patch for 4.7

Ericgig:
**Description**
Fix issues from cython 3.
- Manually set cython.bindings(False) for functions mean to be used with ode.
- Removed use of pxyimport. It only import `c` code and se used to monkeypatch it to compile `c++`. It's internal changed and I rewrote the compilation/importation using inspiration from the `cython.inline_module` PR.
- Added one test using cython 0.29
- Changed the temporary file name construction to have less name collision.

**Related issues or PRs**
cython#3145, cython#555

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,7 +91,6 @@ jobs:
             python-version: "3.10"
             condaforge: 1
             nocvxopt: 1
-            old_cython: 1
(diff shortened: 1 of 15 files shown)


hodgestar (approved):
Thanks for sorting this out @Ericgig.

The compile and import looks a bit similar to what we do in v5, yes?

Ericgig:
@hodgestar , 
Yes but use cython's build stuff instead of setuptools.

I think it's too soon to merge, I still have no idea why I get all these segfault...

PR #2203, opened 2023-07-24
add animation functions

tamakoshi2001:
**Description**
I enabled plot functions in visualization.py to make animations when users put time evolutions. You can see animations with ```%matplotlib notebook%``` on jupyter notebook, but it sometimes does not work. In that case, ```make_html_video``` helps users to save and see animations.


tamakoshi2001:
sample movie
https://github.com/qutip/qutip/assets/72233550/a20da15e-b318-4f6a-bd04-0d07d334b8e0

tamakoshi2001:
sample movie
sample

tamakoshi2001:
Here is an example
```
from qutip import ket, sigmaz, tensor, qeye, mesolve
import qutip as qt
import numpy as np
# Hamiltonian
H = qt.sigmay().unit()

# initial state
psi0 = qt.sigmax().unit()

# list of times for which the solver should store the state vector
tlist = np.linspace(0, 2*np.pi, 50)

results = mesolve(H, psi0, tlist, [], [])

fig, ani, html = qt.anim_hinton(results.states)
HTML
```
<div><video controls src="https://github.com/qutip/qutip/assets/72233550/b5a842ef-fcf8-4137-a8dc-7c3efcefdb62" muted="false"></video></div>

qutip/visualization.py
```diff
@@ -7,7 +7,12 @@
            'matrix_histogram', 'plot_energy_levels', 'plot_fock_distribution',
            'plot_wigner', 'plot_expectation_values',
            'plot_spin_distribution', 'complex_array_to_rgb',
-           'plot_qubism', 'plot_schmidt']
+           'plot_qubism', 'plot_schmidt', '_cyclic_cmap', '_diverging_cmap',
+           '_sequential_cmap', '_is_fig_and_ax', '_cb_labels',
+           '_get_matrix_components', '_update_xaxis', '_update_yaxis',
+           '_update_zaxis', '_stick_to_planes', '_remove_margins',
+           '_set_ticklabels', '_blob', '_sequence_to_latex',
+           '_index_to_sequence', '_to_qubism_index_pair']
```
Ericgig, line 15: Function that start with `_` are not mean to be made public and included here.
You can still import them using:
```
from .visualization import (_cyclic_cmap, _diverging_cmap, ...)
```

qutip/animation.py
```diff
@@ -0,0 +1,1447 @@
+"""
+Functions for creating animations of results of quantum dynamics simulations,
+visualizations of quantum states and processes.
+"""
+
+__all__ = ['anim_wigner_sphere', 'anim_hinton', 'anim_sphereplot',
+           'anim_matrix_histogram', 'anim_fock_distribution',
+           'anim_wigner', 'anim_spin_distribution', 'anim_qubism',
+           'plot_schmidt']
```
Ericgig, line 9:
```suggestion
           'anim_schmidt']
```

qutip/animation.py
```diff
@@ -0,0 +1,1447 @@
...
+
+import os
+from functools import partial
+from IPython.display import HTML
+from base64 import b64encode
+import matplotlib.animation as animation
+from qutip import plot_wigner
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+from .visualization import *
```
Ericgig, line 52: pyplot is imported twice
Qutip's import should ideally be in a block instead if mixed here are there.
Please clean / sort these import.

qutip/animation.py
```diff
@@ -0,0 +1,1447 @@
...
+        save_options['codec'] = None
+
+    ani.save(save_options['name'] + '.mp4', fps=10,
+             writer=save_options['writer'], codec=save_options['codec'])
+    video = open(save_options['name'] + '.mp4', "rb").read()
+    video_encoded = b64encode(video).decode("ascii")
+    video_tag = '<video controls src="data:video/x-m4v;base64,{0}">'.format(
+        video_encoded)
+
+    return ani, HTML(video_tag)
```
Ericgig, line 75: Using `%matplotlib notebook` I can show animation in a notebook without writing it to a file.
No `save`, thus no issue with codex and it's somewhat interactive as I can zoom in.
This function is therefore not needed. We can just return the animation object and let the user see it as is, or save it in a file if desired.

tamakoshi2001: Thanks for your info. I will change them to return fig and ani, but we need ```make_html_video```. In my case, I can use the interactive mode in one jupyter notebook, but not in the other and the google colab.

tamakoshi2001:
@Ericgig Hi. I merged animation functions with functions in visualization.py. Now, users can have animations simply by passing a list.

tamakoshi2001:
Hi. I finished coding. Here is sample  animations.

ani

tamakoshi2001:
@Ericgig Could you review the code?

qutip/visualization.py
```diff
@@ -328,51 +381,58 @@ def hinton(rho, x_basis=None, y_basis=None, color_style="scaled",
...
-            # Create default labels if none are given.
-            if x_basis is None or y_basis is None:
+    if not isinstance(rho, list):
+        rhos = [rho]
+    else:
+        rhos = rho
+
+    Ws = list()
+    w_max = 0
+    for rho in rhos:
```
Ericgig, line 391: Should we ask for the rho's dims to match?
Otherwise it will not make a continuous animation.

qutip/visualization.py
```diff
@@ -828,6 +948,11 @@ def matrix_histogram(M, x_basis=None, y_basis=None, limits=None,
                      options['bars_spacing'])
     ax.view_init(azim=options['azim'], elev=options['elev'])
 
+    # removing margins
+    _remove_margins(ax.xaxis)
+    _remove_margins(ax.yaxis)
+    _remove_margins(ax.zaxis)
```
Ericgig, line 954: Does this need to be inside to loop so it's applied to all artists?

Ericgig (commented):
I am not sure I like it merged in the plot function... It's clean, but feels like an hidden feature.
It will need some example in the documentation and notebooks.

qutip/visualization.py
tamakoshi2001: No. This is already applied to all artists.

qutip/visualization.py
tamakoshi2001: I added an inner function to assert shapes.

qutip/visualization.py
```diff
@@ -982,31 +1120,49 @@ def plot_fock_distribution(rho, fock_numbers=None, color="green",
...
+        artist = ax.bar(np.arange(N), np.real(rho.diag()),
+                        color=color, alpha=0.6, width=0.8).patches
+        artist_list.append(artist)
+
+    if len(rhos) == 1:
+        output = ax
+    else:
+        ani = animation.ArtistAnimation(fig, artist_list, interval=50,
+                                        blit=True, repeat_delay=1000)
+        output = ani
```
Ericgig, line 1153: Why not 
```suggestion
        output = animation.ArtistAnimation(
            fig, artist_list, interval=50,
            blit=True, repeat_delay=1000
        )
```

qutip/visualization.py
```diff
@@ -7,12 +7,14 @@
...
            'plot_spin_distribution', 'complex_array_to_rgb',
-           'plot_qubism', 'plot_schmidt']
+           'plot_qubism', 'plot_schmidt', 'make_html_video']
 
 import warnings
 import itertools as it
 import numpy as np
+import mimetypes
 from numpy import pi, array, sin, cos, angle, log2
+from IPython.display import HTML
```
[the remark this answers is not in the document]
tamakoshi2001: Thank you for you advice in the last meeting. The code below is useful to show animations in a non-interactive mode so I deleted ```make_html_video``` 
```
from IPython.display import HTML
HTML(ani.to_jshtml())
```

qutip/visualization.py
```diff
@@ -101,13 +104,53 @@ def _set_ticklabels(ax, ticklabels, ticks, axis, fontsize=14):
...
+def _equal_shape(matrices):
+    first_shape = matrices[0].shape
+
+    text = "All inputs should have the same shape."
+    if not all(matrix.shape == first_shape for matrix in matrices):
+        raise ValueError(text)
+
+
+def make_html_video(ani, file_path, writer=None, codec=None):
+    """Plots a coloured Bloch sphere.
```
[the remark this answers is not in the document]
tamakoshi2001: I deleted ```make_html_video```

Ericgig (commented):
I will let you deceid if you prefer to have `anim_wigner_sphere` or just `plot_wigner_sphere`... I see arguments for both.

Thank you for fixing issues we discussed yesterday.

tamakoshi2001:
@Ericgig I chose ```anim_wigner_sphere```. I will add test tomorrow.

tamakoshi2001:
@Ericgig Hi. I added tests. Could you review the code?

qutip/animation.py
```diff
@@ -51,3 +55,517 @@ def anim_wigner_sphere(wigners, reflections=False, *, cmap=None,
...
+       and G. Sierra, *Qubism: self-similar visualization of many-body
+       wavefunctions*, `New J. Phys. 14 053028
+       <https://dx.doi.org/10.1088/1367-2630/14/5/053028>`_, arXiv:1112.3560
+       (2012), open access.
+    """
+
+    fig, ani = plot_qubism(ket, theme, how, grid_iteration,
+                           legend_iteration, fig=fig, ax=ax)
+
+    plt.close(fig)
```
Ericgig, line 517: Why do you close the figure?
It stop me from having it shown in the notebook with `%matplotlib notebook`.

Ericgig (commented):
Could the animation functions also accept a `Result` object and read the `result.states` from it?
This could be a nice quality of life.

qutip/animation.py
tamakoshi2001: Thanks. This is a mistake.

tamakoshi2001:
@Ericgig Hi. Animation functions now accept ```Result```. I edit a doc file to see doc in qutip.org.

qutip/animation.py
```diff
@@ -0,0 +1,579 @@
...
+
+try:
+    import matplotlib.pyplot as plt
+except Exception:
+    pass
+
+
+def _result_state(obj):
+    if isinstance(obj, Result):
+        return obj.states
```
Ericgig, line 21: `results` sometime do not save the states and `result.states` is None.
I would check that case and raise an error telling the users to add `options={"store_states": True}` to the simulation.

Ericgig (commented):
The addition of support for Result input is a little rough, did you try using it?
It would not work for animation that takes list of arrays only since result.states is a list of `Qobj`.

qutip/animation.py
tamakoshi2001: I added an assertion. ```result.states``` returns a list so it raises an error when the length of it is zero.

tamakoshi2001:
@Ericgig I saw it works for other functions and thought it is ok. I deleted them and updated the docs.

PR #2204, opened 2023-07-24
Fix cython version to pre 3.0.0

Ericgig:
**Description**
Couldn't make #2202 work...
There are segfault in brmesolve with cython 3.0.0 and the new compilation method does not work properly on windows. 
This replace the PR with a simple limitation to not use cython 3.0.0.

### DIFF
--- a/doc/changes/2204.misc
+++ b/doc/changes/2204.misc
@@ -1,0 +1,1 @@
+Exclude cython 3.0.0 from requirement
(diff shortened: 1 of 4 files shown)


hodgestar (approved):
It was worth a shot making Cython 3.0 work, but perhaps better to leaving fixing it for another time when either we have a better idea or Cython 3 has matured a bit.

PR #2207, opened 2023-07-26
Fix import with cython 3

Ericgig:
**Description**
When cython 3 is installed, qutip will run in no cython mode and a warnings will be raised.

I also removed the `rhs_generate` function. It was only used in an example in the documentation which no longer do what it says.

(See #2202 for my tries in making home made runtime compilation works with cython 3.)


hodgestar (approved):
Changes look good.

Is `rhs_clear` used for anything with `rhs_generate` gone?

Ericgig:
`rhs_clear` is still used to reset solver stored with `rhs_reuse`.

PR #2210, opened 2023-08-02
Restore feedback for solver

Ericgig:
**Description**
Restore the feedback capacities of the solver.
In v4, this was done by using special names for the `args`, but this was confusing. 

For v5, I propose to do it using a `add_feedback` method of the solver:
```
H = QobjEvo([op, lambda t, ket: bra @ ket])
solver = SESolver(H)
solver.add_feedback("ket", "qobj")
solver.run(...)
```
With this, `args["ket"]` will be updated with the state during the evolution.

Most solver support it, the exception are floquet, nm_mcsolve. 
HEOM support it, but the state will be seen as a single ket.
@hodgestar , is that fine?

Available form are as `Qobj`, as `Data`, `expect(oper, data)`.
`mcsolve` also has the collapse and `stochastic` the wiener process.


hodgestar:
Early thought:

The `solver.add_feedback` approach doesn't look too bad, but it does seem a bit odd because on does to specify the state parameters when constructing the operators that the define the system -- i.e. H, c_ops, etc.

Could we not expose the feedback on the QobjEvo itself so that users can then do:

```python
H = QobjEvo([op, lambda t, ket: bra @ ket], feedback={"ket": "qobj"})
...
```

or
```python
H = QobjEvo([op, lambda t, ket: bra @ ket])
H.add_feedback(ket="qobj")
...
```

or even
```python
H = QobjEvo([op, lambda t, ket: bra @ ket])
solver = Solver(...)
solver.rhs.add_feedback(ket="qobj")
```

Thoughts?

It will get a bit trickier for the feedback that is less easily available from the state, but perhaps solvers can just check whether the ops they're using need the special feedback?

Ericgig:
My example was incomplete, they need to set an initial value when creating the `QobjEvo`:
```
H = QobjEvo([op, lambda t, ket: bra @ ket], args={"ket": basis(...)})
```
Otherwise there is an issue with creating the coefficient and if the operator is call without a state first `H(t)`.

For a stand alone QobjEvo, having a feedback does not mean much, I prefer to attach it to the solver where it get's it's meaning. It is also the solver that knows about the collapse and wiener_process.

add_feedback exist on QobjEvo, it just private:
```
H = QobjEvo([op, lambda t, ket: bra @ ket], {"ket":basis(...)})
H._add_feedback("ket", "qobj")
```
Using dict instead of a pair of input does look good.

qutip/tests/solver/test_mesolve.py
```diff
@@ -680,3 +680,19 @@ def test_mesolve_bad_state():
...
+    def f(t, A):
+        return (A-4.)
+
+    N = 10
+    tol = 1e-14
+    psi0 = qutip.basis(N, 7)
+    a = qutip.destroy(N)
+    H = qutip.QobjEvo(qutip.num(N))
+    solver = qutip.MESolver(H)
+    solver.add_feedback("A", qutip.num(N))
```
hodgestar, line 696:
```suggestion
    H = qutip.QobjEvo(qutip.num(N))
    solver = qutip.MESolver(H)
    solver.add_feedback("A", qutip.num(N))
```
I still really don't like the above pattern. `H` is can't sensibly be used by a solver without the specified feedback, so one should immediately be able to introspect `H` to find out what feedback it needs.

If we have that, the solver can look at `H` to see what feedback is needed and either supply that feedback or raise an error if it is unable to supply it.

This way mistakes cannot happen where `H` is silently not given the feedback it need and defaults to some initial value that is wrong, and we don't have `solver.add_feedback` repeatedly updating and overriding information on `H`.

hodgestar:
@Ericgig Will you ping me when you'd like another round of review? Thanks for making the changes so far.

Ericgig:
@hodgestar 
It's ready for another look.
I made it so the feedback can be set at `QobjEvo` creation, with `QobjEvo.add_feedback` method and with a method of the solver.

`QobjEvo` with feedback can be used outside solver with a default value for the argument passed at creation. This default value is needed when creating the coefficient. When used in a solver an error will be raised if the solver cannot provide the feedback.

The `QobjEvo` print will show it has feedback. 

The solver method is mostly there for documentation.

qutip/solver/solver_base.py
```diff
@@ -396,6 +397,29 @@ def add_integrator(cls, integrator, key):
 
         cls._avail_integrators[key] = integrator
 
+    def add_feedback(self, key, type):
```
hodgestar, line 400: Do we still need these `add_feedback` methods now that they are present on the on `QobjEvo` objects? Perhaps we can just document the supported feedback in the documentation for the solver classes? That probably also makes the existence of feedback more discoverable by users,

qutip/solver/solver_base.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, rhs, *, options=None):
         self._integrator = self._get_integrator()
         self._state_metadata = {}
         self.stats = self._initialize_stats()
+        self.rhs._register_feedback({}, solver=self.name)
```
hodgestar, line 54: Maybe `Solver._register_feedback` should be a method that solver instances can override to register their feedback? And it can be called whenever the RHS or equivalent are updated.

hodgestar (commented):
This looks really good to me now. I left some comments about things to clean up, but it's now just polishing left, I think.

qutip/solver/solver_base.py
Ericgig: No instance, but each solver subclass that has something to offer overwrite it.
It's called mostly in `set_state` since it must be called once per trajectories.

qutip/solver/solver_base.py
Ericgig: Having it in `QobjEvo` still feels wrong to me and I prefer using this one.

qutip/core/cy/qobjevo.pyx
```diff
@@ -1160,6 +1174,9 @@ cdef class _Expect:
             f"does not match the state ({state.shape})."
         )
 
+    def __repr__(self):
+        return "Expect"
```
hodgestar, line 1180:
```suggestion
        return "Expect"
```
These `__repr__`s make some sense when included in the `__repr__` of `QobjEvo` but they are very strange on their own. If we're making these names visible as we seem to be doing, we either make them clearer and document what they are.

Perhaps we can rename the classes to `_ExpectFeedback`, `_QobjFeedback` and `_DataFeedback` and then change the reprs to `ExpectFeedbak`, `QobjFeedback` and `DataFeedback`?

I'm not completely happy with my suggested names, so better suggestions welcome.

qutip/core/cy/_element.pyx
```diff
@@ -263,7 +263,8 @@ cdef class _BaseElement:
     def __call__(self, t, args=None):
         if args:
             cache = []
-            self = self.replace_arguments(args, cache)
+            new = self.replace_arguments(args, cache)
+            return new.qobj(t) * new.coeff(t)
```
hodgestar, line 267: Was this change needed? I'm fine with either version, just checking that I am not missing something and they are the same.

qutip/core/cy/qobjevo.pyx
```diff
@@ -187,13 +198,18 @@ cdef class QobjEvo:
         qevo = H0 + H1 * coeff
 
     """
-    def __init__(QobjEvo self, Q_object, args=None, tlist=None,
-                 order=3, copy=True, compress=True,
-                 function_style=None, boundary_conditions=None):
+    def __init__(QobjEvo self, Q_object, args=None, copy=True, compress=True, *,
```
hodgestar, line 201:
```suggestion
    def __init__(QobjEvo self, Q_object, args=None, *, copy=True, compress=True,
```
If we're going to make some arguments keyword-only, perhaps these should include `copy` and `compress`? The `Q_object` and `args` are the two really common ones.

qutip/solver/multitraj.py
```diff
@@ -47,7 +73,11 @@ class MultiTrajSolver(Solver):
     }
 
     def __init__(self, rhs, *, options=None):
-        self.rhs = rhs
+        if isinstance(rhs, QobjEvo):
+            self.system = _MTSystem(rhs)
+        else:
```
hodgestar, line 78: Could we check here that `rhs` really is an instance of `_MTSystem`? I guess someone could pass anything for the RHS accidentally and it would be nice to raise and error straight away.

qutip/solver/sode/rouchon.py
```diff
@@ -132,5 +160,13 @@ def options(self):
     def options(self, new_options):
         Integrator.options.fset(self, new_options)
 
+    def reset(self, hard=False):
+        if self._is_set:
+            state = self.get_state()
+        if hard:
+            raise NotImplementedError
```
hodgestar, line 167:
```suggestion
            raise NotImplementedError("hard reset not implemented")
```

hodgestar: How come this has a `NotImplementedError` but some other `reset` implementations just ignore the `hard` option?

qutip/solver/solver_base.py
hodgestar: I meant that if this call was `self._register_feedback()` rather than `self.rhs._register_feedback(...)` then each Solver class could update all the state it needed inside `_register_feedback` (e.g. add feedback to rhs and c_ops) and the feedback registration could all happen in one method.

qutip/solver/solver_base.py
hodgestar: Let me resolve this thread and I'll write a short note in the main conversation instead.

hodgestar (changes requested):

hodgestar:
> What if we rethink the approach from the beginning:
> 
> ```
> QobjEvo([op, func], args={"state": Solver.StateFeedback(default=psi0)})
> QobjEvo([op, func], args={"e1": SESolver.ExpectFeedback(e_op)})
> QobjEvo([op, func], args={"state": MESolver.StateFeedback(raw_data=True)})
> QobjEvo([op, func], args={"collapse": MCSolver.CollapseFeedback()})
> QobjEvo([op, func], args={"W": SMESolver.WeinerFeedback()})
> ```
> 
> And we remove the `add_feedback` method.

This looks great. I really like that it makes it clearer where the feedback comes from and that it is solver dependent. Removing the need to specify the feedback in two places when constructing a `QobjEvo` is really nice too.

Let's do it in a separate PR though (unless you strongly want to update this one).

I have a slight concern that it might be a bit complicated for users to find the feedback classes (e.g. they have to know to import the right solver) but I'm sure we can make that nice the PR itself.

Ericgig:
@hodgestar
I changed the interface.

Ericgig:
Questions / doubt I still have is about making the `default` mandatory for states.
At first I wanted to pass `qzeros` with the dims fetched from those of the `QobjEvo`, but sometime we need to use the args first to get the QobjEvo's dimensions...

I also added test for `QobjEvo.dag().dims`, previously the test passed even without the fix because we only tested for the returned qobj's dims.

hodgestar:
The question about the mandatory default for states is a good one. So currently we call the `QobjEvo` with `t=0` to get its dims. This is slightly hacky, but very convenient, and usually `t=0` is a valid time.

Feedback breaks this assumption though because now `t=0` is insufficient information. We also need the current state, but we can't create a dummy state because we don't know what its dims should be yet. So there is a cycle we have to break somehow. Currently we break this by having the user supply an initial state, but perhaps we could do other things.

The worst case is when the Qobj part of an EvoElement is supplied by a function. Then we can't get the dims without calling it. When the Qobj part is just a Qobj, getting the dims is fairly straight forward.

So one option would be to ask the user to supply the dims if there are no EvoElements that provide a Qobj to look at, or to guarantee that their function is callable with just `t=0`. I'm not sure this is entirely nicer, but it does avoid having to supply a dummy initial state.

Thoughts?

Ericgig:
The way it is now, the coefficient functions need to be callable with `t=0, state=None, expect=0.` when a default is not passed.

If we require a default, we could suggest the users to provide the the initial state of the solver.
For the same QobjEvo dimensions, the state dimensions could be ket, density matrix or propagator.
Passing a default state would remove the need for the QobjEvo to choose between each format.

Not a fan of having pass dims at QobjEvo creation when no Qobj are passed and StateFeedback is used.
The condition for when it's needed is quite specific and can't be used from solver function call:
`sesolve(H_func, psi0, tlist, args={"psi": StateFeedback()})` could not work since `sesolve` get it's dimensions from the Hamiltonian but now we would need the dims to parse it.

So my preference are asking for a default or accepting `func(t=0, **args)` with the state replaced to `None`.

qutip/solver/_feedback.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class _ExpectFeedback(_Feedback):
+    def __init__(self, oper, default=0.):
+        self.oper = QobjEvo(oper)
+        self.N = oper.shape[1]
+        self.N2 = oper.shape[1]**2
+        self.default = default
+
+    def check_consistancy(self, dims):
```
hodgestar, line 15:
```suggestion
    def check_consistency(self, dims):
```
Just correcting the spelling. Also needs to be changed in a few other places.

hodgestar (approved):
I've looked through the recent changes. I'm still in favour of merging this and continuing the clean-up work in another PR if needed.

hodgestar:
Yay!

PR #2218, opened 2023-08-21
Improved sampling algorithm for mcsolve

dkweiss31:
**Description**
Added functionality associated with an improved monte-carlo sampling algorithm described in https://journals.aps.org/pra/abstract/10.1103/PhysRevA.99.052327 . This algorithm samples the no-jump trajectory only once, then only samples jump trajectories afterwards. Expectation values, final states, etc are then computed by appropriately weighting the no jump and jump trajectories by the no-jump probability.


Ericgig (commented):
Thank you for the contribution, this seems an interesting variant the the Monte-carlo algorithm. 

Should it raise a warning for users that tries to access each trajectories? Or add a note in the docstring? The trajectories would look strange if you don't read the documentation.

I am wondering if `MCSolverImprovedSampling` could be merged in `MCSolver`. I am not a fan of having `mcsolve` use 2 different solvers and it should be accessing by the new class interface. Another options would be that make `MCSolverImprovedSampling` public and add a new function to call it `mcissolve`?

I would prefer that you revert the changes made by black. It add a lot of changes that are unrelated to the new feature and makes reviewing the PR annoying. Also we use the pep8 line width of 80 instead of black's 88.
For now, we mostly use black in new code and adding black to existing file should be their own PR.

dkweiss31:
Thanks for the quick reply! 

> Should it raise a warning for users that tries to access each trajectories? Or add a note in the docstring? The trajectories would look strange if you don't read the documentation.
> 
Maybe, though I'm not sure why they would look strange? It's possible a user could get confused if they don't see any no-jump trajectories other than the first one, but hopefully they wouldn't be confused if they set the flag to use this algorithm.

> I am wondering if `MCSolverImprovedSampling` could be merged in `MCSolver`. I am not a fan of having `mcsolve` use 2 different solvers and it should be accessing by the new class interface. Another options would be that make `MCSolverImprovedSampling` public and add a new function to call it `mcissolve`?
> 
Yeah this might be the way to go. I only had it the way I did to have `MultiTrajSolverImprovedSampling` get inherited before `MCSolver` so that the functionality there gets used and also have `resultclass = McResultImprovedSampling`. We could do a conditional inheritance in `MCSolver`, but I'm worried that might be messy/ugly?

> I would prefer that you revert the changes made by black. It add a lot of changes that are unrelated to the new feature and makes reviewing the PR annoying. Also we use the pep8 line width of 80 instead of black's 88. For now, we mostly use black in new code and adding black to existing file should be their own PR.
>
Sorry about that, will revert

qutip/solver/result.py
```diff
@@ -861,19 +853,166 @@ def __add__(self, other):
...
+            self._sum_final_states_no_jump += dm_final_state
+        else:
+            self._sum_final_states_jump += dm_final_state
+
+    def _average_computer(self):
+        avg = np.array(self._sum_expect_jump) / (self.num_trajectories - 1)
+        avg2 = np.array(self._sum2_expect_jump) / (self.num_trajectories - 1)
+        return avg, avg2
+
+    def _add_first_traj(self, trajectory):
```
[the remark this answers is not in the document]
dkweiss31: Agreed re testing num_trajectories!=1, that's more elegant and would eliminate the need for a custom `add` function. Where did you have in mind when you say "here"?

qutip/tests/solver/test_mcsolve.py
```diff
@@ -56,8 +56,13 @@ def test_states_and_expect(self, hamiltonian, args, c_ops, expected, tol):
         result = mcsolve(hamiltonian, self.state, self.times, args=args,
                          c_ops=c_ops, e_ops=self.e_ops, ntraj=self.ntraj,
                          options=options, target_tol=0.05)
+        result_improved_sampling = mcsolve(hamiltonian, self.state, self.times, args=args,
+                         c_ops=c_ops, e_ops=self.e_ops, ntraj=self.ntraj,
+                         options=options, improved_sampling=True)
```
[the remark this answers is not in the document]
dkweiss31: Yes sorry about that

qutip/solver/mcsolve.py
```diff
@@ -1,17 +1,30 @@
...
+    c_ops=(),
+    e_ops=None,
+    ntraj=500,
+    *,
+    args=None,
+    options=None,
+    seeds=None,
+    target_tol=None,
+    timeout=None,
+    improved_sampling=False,
```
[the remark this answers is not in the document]
dkweiss31: Right so I guess the question is: do we make a new function mcissolve or add this as a flag to options. Maybe I'd be more inclined for a new function, but I don't have a strong preference

qutip/solver/result.py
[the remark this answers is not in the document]
Ericgig: By here, I meant that the logic to add the no jump trajectory could be in `_add_first_traj`.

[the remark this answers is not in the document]
dkweiss31: I'm not positive that would be the best option, since the purpose of `_add_first_traj` is only to initialize a bunch of zero arrays (could be a motivation to change the name of that function to something like `_initialize_data_arrays`). All of the data for each trajectory, including the first trajectory, is added by the `_reduce*` functions. 

I could add the no-jump trajectory at the end of `_add_first_traj` but I would still have to check e.g. `if self.num_trajectories != 1:` in each of the `_reduce*` functions. For now I've committed a version that leaves the behavior of `_add_first_traj` unmodified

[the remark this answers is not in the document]
dkweiss31: I'm not positive that would be the best option, since the purpose of `_add_first_traj` is only to initialize a bunch of zero arrays (could be a motivation to change the name of that function to something like `_initialize_data_arrays`). All of the data for each trajectory, including the first trajectory, is added by the `_reduce*` functions.

I could add the no-jump trajectory at the end of `_add_first_traj` but I would still have to check e.g. `if self.num_trajectories != 1:` in each of the `_reduce*` functions. For now I've committed a version that leaves the behavior of `_add_first_traj` unmodified

qutip/solver/multitraj.py
```diff
@@ -193,13 +205,16 @@ def run(self, state, tlist, ntraj=1, *,
         result.stats['run time'] = time() - start_time
         return result
 
-    def _run_one_traj(self, seed, state, tlist, e_ops):
+    def _run_one_traj(self, seed, state, tlist, e_ops,
+                      no_jump=False, jump_prob_floor=0.0):
```
Ericgig, line 209: `MultiTrajSolver` also apply to stochastic evolution which have continuous noise, no jump...
These new input do not belong here, but in `MCSolverImprovedSampling`.

dkweiss31:
> I believe that `MultiTrajSolverImprovedSampling` and `MultiTrajResultImprovedSampling`. We only need one new class for the solver and result, not 2 each. It would even be nice if it could be inserted into `MCSolver` without making the code to messy...
>
Agreed 
> I believe that a scaling should be applied in to the `McResult.photocurrent` method. Could you check. It's an histogram of the jumps times.
> 
Yes you're right, sorry I missed that. Should be fixed now.
> The last issue is whether to make a new function or not. I feel it belong as an options in the existing mscolve, but it would be easy to overlook by users. As a new function it will get more attention. I also don't have a strong opinion on the matter.
>
I think you are right. I added it as an option to options and updated tests, docs accordingly.

qutip/solver/mcsolve.py
```diff
@@ -385,6 +408,9 @@ def __init__(self, H, c_ops, *, options=None):
                 rhs -= 0.5 * n_op
 
         self._num_collapse = len(self._c_ops)
+        self.options = options
+        if self.options["improved_sampling"]:
+            self.resultclass = McResultImprovedSampling
```
dkweiss31, line 413: This is something I'm not super happy about: this wants to be defined outside of init (where we have ``resultclass = McResult`` now). Maybe still not a good enough reason to define a whole new class ``MCSolverImprovedSampling`` though

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -148,6 +148,60 @@ Now, the Monte Carlo solver will calculate expectation values for both operators
...
+and for all future runs we only sample random numbers :math:`r_1` where :math:`r_1>p`, thus ensuring
+that a jump will occur. When it comes time to compute expectation values, we weight the no-jump
+trajectory by :math:`p` and the jump trajectories by :math:`1-p`. This algorithm is described
+in [Abd19]_ and can be utilized by setting the option ``"improved_sampling"`` in the call to
+``mcsolve``:
+
+.. plot::
+    :context: close-figs
+
+    data = mcsolve(H, psi0, times, [np.sqrt(0.1) * a], e_ops=[a.dag() * a, sm.dag() * sm],options={"improved_sampling": True})
```
Ericgig, line 168:
```suggestion
    data = mcsolve(H, psi0, times, [np.sqrt(0.1) * a], e_ops=[a.dag() * a, sm.dag() * sm], options={"improved_sampling": True})
```

qutip/solver/mcsolve.py
Ericgig: Or maybe set `resultclass` to be a property that check the option and return the right type.

Another options would by using a metaclass. It feels overkill, but would be clean (here at least).

doc/guide/dynamics/dynamics-monte.rst
```diff
@@ -148,6 +148,60 @@ Now, the Monte Carlo solver will calculate expectation values for both operators
...
+    plt.plot(times, np.exp(-gamma * times), label=r"$\exp(-\gamma t)$")
+    plt.title('Monte Carlo: improved sampling algorithm')
+    plt.xlabel("time [ns]")
+    plt.ylabel(r"$p_{1}$")
+    plt.legend()
+    plt.show()
+
+
+The original sampling algorithm samples the no-jump trajectory on average ``97.8\%" of the time, while the improved
+sampling algorithm only does so once.
```
Ericgig: The type of string / equation format don't match.
```suggestion
The original sampling algorithm samples the no-jump trajectory on average ``97.8\%`` of the time, while the improved
sampling algorithm only does so once.
```

Ericgig (commented):
Looking good. 

I left some ideas on how to improve the resultclass setting.

qutip/solver/result.py
```diff
@@ -982,6 +986,167 @@ def runs_photocurrent(self):
...
+        cols = [[] for _ in range(self.num_c_ops)]
+        tlist = self.times
+        for collapses in self.collapse:
+            for t, which in collapses:
+                cols[which].append(t)
+        mesurement = [
+            (1 - self.no_jump_prob) *
+            np.histogram(cols[i], tlist)[0] / np.diff(tlist) / (self.num_trajectories - 1)
+            for i in range(self.num_c_ops)
+        ]
```
Ericgig, line 1146: Line too long otherwise:
```suggestion
        mesurement = [
            (1 - self.no_jump_prob) / (self.num_trajectories - 1) *
            np.histogram(cols[i], tlist)[0] / np.diff(tlist)
            for i in range(self.num_c_ops)
        ]
```

qutip/solver/mcsolve.py
dkweiss31: The metaclass idea sounds fun but maybe the below simple fix works?

Ericgig (approved):
Thank you @dkweiss31.
Everything looks fine to me.

PR #2223, opened 2023-08-31
Deprecate `orbital`

Ericgig:
**Description**
Remove orbital.
It is still used in tutorials.


qutip/__init__.py
```diff
@@ -60,6 +59,13 @@
 from .about import *
 from .cite import *
 
+
+def orbital(*_, **__):
```
hodgestar, line 63: I'm happy for us to just drop `orbital`. We can add this note to the release notes instead.

hodgestar (approved):
I've prefer to drop the not implemented orbital function and just replace it with a note in the release notes, but otherwise looks good.

qutip/__init__.py
```diff
@@ -60,6 +59,13 @@
...
+    raise NotImplementedError(
+        "'orbital was removed. Use scipy.special.sph_harm instead'"
+    )
+
+
```
Ericgig, line 68:
```suggestion
```

hodgestar (approved):

PR #2224, opened 2023-09-01
Add insufficient memory checks

Ericgig:
**Description**
When allocating memory in cython, we did not check if the memory was allocated or not due to lacking RAM.
`kron_dia` was particularly inefficient for `bra & ket` combination., which caused a segfault in the `steadystate` benchmarks.


hodgestar (commented):
The change looks good. Do we have unit tests for both code paths?

PR #2225, opened 2023-09-06
Fix fsesolve fallback

Ericgig:
**Description**
Fallback to `fsesolve` passed unused argument.
Add a test.


hodgestar (approved):
Looks good!

PR #2241, opened 2023-10-06
Update minimum versions

Ericgig:
**Description**
Update minimum version supported 
- python >= 3.10
- numpy >= 1.22
- scipy >= 1.8
- matplotlib >= 3.5

According to scientific python spec 0.

Cython 3.0.3 does not support python under 3.9. The changelog does not seems to indicate if it is intentional or not.
It also simplify some parts where we branch over module versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,7 +107,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     env:
-      # Set up wheels matrix.  This is CPython 3.8--3.10 for all OS targets.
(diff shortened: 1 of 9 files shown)


hodgestar:
Python 3.9 only becomes end-of-line at the end of 2025 (https://devguide.python.org/versions/), so I would hope that Cython still plans to support it. Maybe we can ask them?

hodgestar (approved):
I'm happy with the changes. I don't think we should reject Python 3.8 and 3.9 yet, but I'm also happy for them to be added back into the test matrix again later.

PR #2242, opened 2023-10-06
Conditional cython requirement

Ericgig:
**Description**
Atempt to fix the issue with cython 3.0.3 and python <=3.9 with conditional requirements.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -133,7 +133,8 @@ jobs:
           if [[ -n "${{ matrix.conda-extra-pkgs }}" ]]; then
             conda install "${{ matrix.conda-extra-pkgs }}"
           fi
-          python -m pip install -e .[$QUTIP_TARGET]
+          conda list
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -133,7 +133,7 @@ jobs:
           if [[ -n "${{ matrix.conda-extra-pkgs }}" ]]; then
             conda install "${{ matrix.conda-extra-pkgs }}"
           fi
-          python -m pip install -e .[$QUTIP_TARGET]
+          python -m pip install -e .[$QUTIP_TARGET] -vv
```
hodgestar, line 136: Should we remove the `-vv` now that the tests pass?

hodgestar (approved):

.github/workflows/tests.yml
Ericgig:
```suggestion
          python -m pip install -e .[$QUTIP_TARGET]
```

PR #2247, opened 2023-10-18
Fix cython 3.0.4

Ericgig:
**Description**
Another cython release, another fix needed...

The missing type definition in `dia.diags_` was causing an error during the cythonize step.

The other changes are related to the new "performance hint".
`nogil` and `except *` do not work well together.

### DIFF
--- a/qutip/core/data/expect.pxd
+++ b/qutip/core/data/expect.pxd
@@ -3,13 +3,13 @@
 
 from qutip.core.data cimport CSR, Dense, Data, Dia
 
-cpdef double complex expect_csr(CSR op, CSR state) except * nogil
-cpdef double complex expect_super_csr(CSR op, CSR state) except * nogil
+cpdef double complex expect_csr(CSR op, CSR state) except *
+cpdef double complex expect_super_csr(CSR op, CSR state) except *
 
-cpdef double complex expect_csr_dense(CSR op, Dense state) except * nogil
+cpdef double complex expect_csr_dense(CSR op, Dense state) except *
 cpdef double complex expect_super_csr_dense(CSR op, Dense state) except * nogil
 
-cpdef double complex expect_dense(Dense op, Dense state) except * nogil
+cpdef double complex expect_dense(Dense op, Dense state) except *
 cpdef double complex expect_super_dense(Dense op, Dense state) except * nogil
 
 cpdef double complex expect_dia(Dia op, Dia state) except *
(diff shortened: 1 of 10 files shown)


qutip/core/data/expect.pxd
```diff
@@ -3,13 +3,13 @@
...
 from qutip.core.data cimport CSR, Dense, Data, Dia
 
-cpdef double complex expect_csr(CSR op, CSR state) except * nogil
-cpdef double complex expect_super_csr(CSR op, CSR state) except * nogil
+cpdef double complex expect_csr(CSR op, CSR state) except *
+cpdef double complex expect_super_csr(CSR op, CSR state) except *
 
-cpdef double complex expect_csr_dense(CSR op, Dense state) except * nogil
+cpdef double complex expect_csr_dense(CSR op, Dense state) except *
 cpdef double complex expect_super_csr_dense(CSR op, Dense state) except * nogil
```
hodgestar, line 10: How come we still have `expect * nogil` here?

hodgestar (commented):
Left one question about why some `except * nogil` instances were not removed.

qutip/core/data/expect.pxd
Ericgig: It's not wrong to use `except * nogil`, it's just not optimal as cython need to get the gil to check if an error was raised when the function return. Returning a fix value for errors (`except -1`) is more efficient, as cython can check the error without the gil, but it's not supported for complex return.

For a function that is called outside another nogil block, it has no performance impact so we don't need to remove them.

hodgestar: I would make these function signatures more similar and just remove `nogil` from `expect_super_csr_dense`. Otherwise the slight performance differences will be annoying.

Ericgig: I did not randomly chose to skip `expect_super_csr_dense`, it use the nogil while `expect_csr_dense` call other function that run in `nogil`, but itself do not benefit from it.
Since they don't do the same computation, performance differences are expected.
And performance should be closer to each other this way as `expect_csr_dense` would go back and forth gil/nogil twice as it was instead of once like `expect_super_csr_dense`. Now it's once for both function...

Ericgig:
@hodgestar I will merge this as is since it block the tests from passing everywhere.

The situation with `except * nogil` is not perfect, but it's not a bottleneck in our computations.

If you look at the generated code, the issue is only when calling an `except * nogil` function returning complex or void within another `nogil` function. Therefore I removed the nogil from function that called other nogil functions, not those who did the work.

If you see optimisation I missed, please open another PR.

PR #2248, opened 2023-10-19
Add deprecation warnings to solver

Ericgig:
**Description**
Add warning messages went passing input to solver that moved / were renamed in v5.
Hopefully this should help with the transition.

When the parameter affect the physical result, an error is raised instead of a warning.

### DIFF
--- a/qutip/solver/solver_base.py
+++ b/qutip/solver/solver_base.py
@@ -7,6 +7,7 @@ from .result import Result
 from .integrator import Integrator
 from ..ui.progressbar import progress_bars
 from time import time
+import warnings
 
 
 class Solver:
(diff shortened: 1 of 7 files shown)


qutip/solver/solver_base.py
```diff
@@ -394,3 +395,98 @@ def add_integrator(cls, integrator, key):
...
+        )
+        del kwargs["tol"]
+
+    if "map_func" in kwargs and solver in ["mc", "stoc"]:
+        warnings.warn(
+            '"map_func" is now included in options:\n'
+            'Use `options={"map": "serial" / "parallel" / "loky"}`'
+        )
+        # 4.X pass the function, 5.X use a string
+        del kwargs["map_func"]
```
hodgestar, line 441: We could try the function's `__name__` here and use it if it looks right. Perhaps not worth the complexity though.

hodgestar (approved):
Looks good.

We should make these [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning) so that they are visible to users.

qutip/solver/solver_base.py
Ericgig: I made it so function are also accepted in v5.

qutip/solver/parallel.py
```diff
@@ -385,9 +385,12 @@ def loky_pmap(task, values, task_args=None, task_kwargs=None,
 
 
 _get_map = {
+    parallel_map: parallel_map,
```
hodgestar, line 388: I would prefer not to do this -- it's a mess. The string keys make such arguments simple and, e.g., serializable, and including functions as keys breaks that.

I'd be much happier to just ask people to change their arguments to strings in v5.

hodgestar (approved):

PR #2251, opened 2023-10-24
Fix temp file removal

### LINKED ISSUE #2231: .pyx file is not deleted after compilation
### Bug Description

In `qutip.qobjevo.compile`, on line 1646, the `.pyx` file created by qutip is registered into the global variable `coeff_files` containing a `_file_list` object using the method `coeff_files.add(file_)`. The `add` method appends a `.pyx` extension on line 60 to the string `file_`
```
self.files += [file_ + ".pyx"]
```
However, the `file_` string generated by `_compiled_coeffs` already has the file extension, judging from the return value of `qutip.qobjevo_codegen._import_str`, on line 54
```
    return coeff_obj, try_file + ext
```
I believe this causes `coeff_files.clean()` on line 1697 of `qutip.qobjevo.compile` to not delete the `.pyx` files after the cython class has been loaded.

### Code to Reproduce the Bug

```shell
import qutip as qt
qt.sesolve([qt.identity(2), [qt.sigmax(), 'sin(t)']], qt.basis(2, 0),
    tlist=[0, 2])
```

### Code Output

```shell
creates the file cqobjevo_compiled_coeff_6095589624050.pyx in the working directory.
```

### Expected Behaviour

The `.pyx` file in the working directory should be deleted.

### Your Environment

```shell
QuTiP Version:      4.7.2
Numpy Version:      1.25.2
Scipy Version:      1.11.2
Cython Version:     0.29.36
Matplotlib Version: 3.7.2
Python Version:     3.10.12
Number of CPUs:     4
BLAS Info:          Generic
OPENMP Installed:   False
INTEL MKL Ext:      False
```

### Additional Context

_No response_

Ericgig:
**Description**
Properly remove temp pyx files.

**Related issues or PRs**
fix #2231

### DIFF
--- a/qutip/qobjevo.py
+++ b/qutip/qobjevo.py
@@ -87,7 +87,7 @@ class _file_list:
         self.files = []
 
     def add(self, file_):
-        self.files += [file_ + ".pyx"]
+        self.files += [file_]
 
     def clean(self):
         to_del = []


hodgestar (approved):

Ericgig:
Not sure how to make a test for this. The file name is random and it's not always reliable when the `__del__` is called which is where the file is erased.

hodgestar:
> Not sure how to make a test for this. The file name is random and it's not always reliable when the `__del__` is called which is where the file is erased.

You could create a coefficient and check that the last entry of `coeff_files.file_list` looks correct and exists. The wouldn't check the deletion but would check that the filename is correct.

You can do `del qutip.qobjevo.coeff_files` to trigger the clean-up of the object, but that is a bit GC dependent and implementing it properly would require calling `gc.collect()`. Not sure if that is worth the hassle.

HarshBabla99:
Hi, thanks for fixing this issue. 
However, its release seems to be stalled because of other incomplete features in v4.7.3.
Is there something I can to fix this issue for myself locally, while we wait for v4.7.3 to be released publicly?

Thanks,
Harsh.

HarshBabla99:
Thanks

PR #2252, opened 2023-10-24
Avoid int overflow in Qobj dims check

### LINKED ISSUE #2249: enr_destroy throws a ValueError in the Qobj creation step
### Bug Description

In the "return" statement of enr_destroy, we have `return [Qobj(a, dims=[dims, dims]) for a in a_ops]`. Then, when attempting to create the Qobjs in this return statement, we have the following conditional statement:

```
if (self._data.shape[0] > np.prod(np.hstack(self.dims[0])) or
           self._data.shape[1] > np.prod(np.hstack(self.dims[1]))) and \
           self.type != 'super':

            raise ValueError(f"Qobj has smaller dims {self.dims} " +
                             f"than underlying shape {self._data.shape}")
```
                            
Because np.prod assumes that the input numbers are 32 bit integers, there is a possibility for an overflow, causing the result to be negative and hence this ValueError will always be raised.

### Code to Reproduce the Bug

```shell
from qutip import enr_destroy

enr_destroy([12]*12,2)
```

### Code Output

```shell
ValueError: Qobj has smaller dims [[12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12], [12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12]] than underlying shape (91, 91)
```

### Expected Behaviour

enr_destroy would work

### Your Environment

```shell
QuTiP Version:      4.7.3
Numpy Version:      1.20.1
Scipy Version:      1.9.3
Cython Version:     0.29.21
Matplotlib Version: 3.6.2
Python Version:     3.8.6
Number of CPUs:     16
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
```

### Additional Context

This can be easily fixed by doing np.prod(np.hstack(self.dims[0])**, dtype = np.int64**)  instead of what is currently being done.

Ericgig:
**Description**
Avoid int overflow in Qobj dims check 

**Related issues or PRs**
fix #2249

### DIFF
--- a/doc/changes/2252.bugfix
+++ b/doc/changes/2252.bugfix
@@ -1,0 +1,1 @@
+Avoid integer overflow in Qobj creation.
--- a/qutip/qobj.py
+++ b/qutip/qobj.py
@@ -28,6 +28,7 @@ from qutip import __version__
 from qutip.fastsparse import fast_csr_matrix, fast_identity
 from qutip.cy.ptrace import _ptrace
 from qutip.permute import _permute
+from qutip.dimensions import flatten
 from qutip.sparse import (
     sp_eigs, sp_expm, sp_fro_norm, sp_max_norm, sp_one_norm, sp_L2_norm,
 )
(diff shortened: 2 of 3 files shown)


hodgestar (approved):

hodgestar:
This probably also fixes some cases for the bigger idxint.

PR #2254, opened 2023-11-02
Faster steadystate, mainly for Dia

Ericgig:
**Description**

Speed up `steadystate`, mainly for the `direct` method with the `dia` method, but some of the optimization affect most calls.

- liouvillian(H) + lindblad_dissipator(c_ops) is slower that liouvillian(H, c_ops) by about 2x.
-  Dia are bad with vectors, the `kron(bra, ket)` in `_steadystate_direct` alone was ~90% of the compute time. Using `CSR` instead is much faster.
- Dia to CSR conversion are about made 100x faster.

Overall, the timing for `Dia` is about 20x faster and 10% faster for `CSR`

### DIFF
--- a/qutip/core/data/solve.py
+++ b/qutip/core/data/solve.py
@@ -75,6 +75,9 @@ def solve_csr_dense(matrix: Union[CSR, Dia], target: Dense, method=None,
         raise ValueError("mkl is not available")
     elif method == "mkl_spsolve":
         solver = mkl_spsolve
+        # mkl does not support dia.
+        if isinstance(matrix, Dia):
+            matrix = _data.to("CSR", matrix)
     else:
         raise ValueError(f"Unknown sparse solver {method}.")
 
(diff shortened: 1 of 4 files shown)


hodgestar (approved):
Changes look good. Happy for this to be merged if the tests already run the solver with `Dia` and not-`Dia` data types.

Ericgig:
Tests were only with `Dia`, added tests with other types.

hodgestar (approved):

qutip/core/data/solve.py
```diff
@@ -73,8 +73,11 @@ def solve_csr_dense(matrix: Union[CSR, Dia], target: Dense, method=None,
         solver = getattr(splinalg, method)
     elif method == "mkl_spsolve" and mkl_spsolve is None:
         raise ValueError("mkl is not available")
-    elif method == "mkl_spsolve":
+    elif method in "mkl_spsolve":
```
hodgestar, line 76: Surely this should stay `==`?

hodgestar (approved):

PR #2256, opened 2023-11-03
Add `build_dir` to compilation options.

### LINKED ISSUE #2230: provide a setting for user specified path to store cython compiled files
### Problem Description

String based coefficients with time-dependent Hamiltonians in qutip 4.7 causes compiled files (.so, .o, .cpp files) to be stored into a hidden folder `.pyxbld` in the user's home directory. This is a problem because, when using qutip on an HPC cluster (the cluster uses Red Hat Enterprise Linux 8), I submit multiple qutip simulation jobs to the cluster, each of which generating cached file in my home directory and using up my storage. Qutip is not aware of these files and they are never deleted.

The root cause of the problem is in the top level __init__.py, on line 72, `_pyxbuilder.install()`, which calls the install() function in qutip.cy.pyxbuilder, on line 47:

```
def install():
    """Install the pyximport interface."""
    return pyximport.install(setup_args={'include_dirs': [np.get_include()]})
```
Since the `build_dir` keyword argument is not supplied to `pyximport.install`, the default path is used.

### Proposed Solution

Supply an option in `qutip.settings`, or `qutiprc`, or `qutip.solver.Options` to specify a custom path for the compiled files. This path is then passed to `pyximport.install` as the `build_dir` argument when `__init__.py` is run, upon first import of qutip.

### Alternate Solutions

_No response_

### Additional Context

_No response_

Ericgig:
**Description**
Add `build_dir` to set the cythonize build location to the `CompilationOptions`.

**Related issues or PRs**
Fix #2230 for v5

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -245,6 +245,9 @@ class CompilationOptions(QutipOptions):
 
     clean_on_error : bool [True]
         When writing a cython file that cannot be imported, erase it.
+
+    build_dir: str [None]
+        cythonize's build_dir.
     """
     _link_flags = ""
     _compiler_flags = ""
@@ -275,6 +278,7 @@ class CompilationOptions(QutipOptions):
         "link_flags": _link_flags,
         "extra_import": "",
         "clean_on_error": True,
+        "build_dir": None,
     }
     _settings_name = "compile"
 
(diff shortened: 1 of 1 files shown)


hodgestar (approved):

qutip/core/coefficient.py
```diff
@@ -245,6 +245,9 @@ class CompilationOptions(QutipOptions):
 
     clean_on_error : bool [True]
         When writing a cython file that cannot be imported, erase it.
+
+    build_dir: str [None]
+        cythonize's build_dir.
```
hodgestar, line 250: Might be worthwhile mentioning Cython's default and why one might need to set this (e.g. that the files there are cached).

Ericgig: While the parameter exist in the source code, it is not in the [documentation](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#cythonize-arguments) (yet?)
I don't want to give more details than cython themselves give.

PR #2260, opened 2023-11-09
Bosonic fitting

gsuarezr:
**Description**

The aim of this project is to enhance the capabilities of qutip's HEOM implementation by introducing a set of tools that offer the option to simulate arbitrary spectral densities or correlation functions. 

To achieve this the main idea is to fit our wanted spectral density by a set of underdamped Drude-Lorentz  spectral densities or correlation functions. The idea is that the user will supply his desired spectral density or correlation function and the tools will automatically do the fit and construct the corresponding HEOM bath. Guesses ffor the fit and the number of terms will be decided automatically if the user does not specify them. To decided on the number of terms, we will used the normalized root mean square error as a measure of the goodness of the fit. 

I also added functions to bosonic  bath to quickly obtain the correlation function, power spectrum and spectral density for the existing  bosonic baths 

**Note**

Those tests fail because mpmath is not in the requirements, it is needed because the scipy version of the Rienmann zeta function doesn't take complex arguments (https://github.com/scipy/scipy/issues/9182).

**Related issues or PRs**

The Qutip tutorial notebooks are updated to show this new functionality in (https://github.com/qutip/qutip-tutorials/pull/72)


pmenczel (commented):
Thank you for adding some tests as well! Though I haven't looked at them yet.

@hodgestar has promised to make some more high-level comments later, and I expect the structure to change a bit. To speed things up, I'm just making some smaller comments here which I hope will remain valid regardless.

qutip/solver/heom/bofin_baths.py
```diff
@@ -8,15 +8,22 @@
...
-
 import numpy as np
-from scipy.linalg import eigvalsh
 
+try:
+    from mpmath import mp
+except ModuleNotFoundError:
+    print(
+        'The mpmath module is needed for the description of Ohmic baths')
+
```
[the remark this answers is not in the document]
mcditoos: I agree that's a better fix, when I did it I was just trying to get around the tests, didn't give it too much thought. It's great to know people in scipy are working on it, i'll keep an eye on it so that I know when it's released ,  I wasn't sure how to answer these, so I ended up doing a commit that implements most suggestions including this one.

qutip/solver/heom/bofin_baths.py
```diff
@@ -973,10 +1057,621 @@ def _calc_eps(self, Nk):
...
+            summary = gen_summary(self.fit_time, self._rmse, self.spec_n,
+                                  "The Spectral Density", lam, gamma, w0)
+            return summary
+        except NameError:
+            return "Fit correlation instance: \n No fit has been performed yet"
+
+    def summary(self):
+        print(self.__str__())
+
+    def spectral_density_approx(self, w, a, b, c):
```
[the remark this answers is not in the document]
mcditoos: I did rename spectral_density_approx -> _spectral_density_approx , but kept spectral_density -> spectral_density_approx , mainly because if I keep the name spectral_density then I can use the power_spectrum method that I put in BosonicBath, of course maybe all these methods will be moved around, but I guess I'll wait on the highlevel comments before renaming this one

qutip/solver/heom/bofin_baths.py
```diff
@@ -973,10 +1057,621 @@ def _calc_eps(self, Nk):
...
+                / (((w + c[i]) ** 2 + b[i] ** 2)
+                    * ((w - c[i]) ** 2 + b[i] ** 2))
+            )
+        return tot
+
+    def spectral_density(self, w):
+        lam, gamma, w0 = self.params_spec
+        return self.spectral_density_approx(w, lam, gamma, w0)
+
+    def get_fit(
```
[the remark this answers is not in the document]
mcditoos: You're right! I added this in the last commit

qutip/solver/heom/bofin_baths.py
```diff
@@ -973,10 +1057,621 @@ def _calc_eps(self, Nk):
...
+        vkAR = []
+        ckAI = []
+        vkAI = []
+        UD = UnderDampedBath(self.Q, lam[0], gamma[0], w0[0], self.T, self.Nk)
+        terminator = 0. * spre(self.Q)
+        # the number of matsubara expansion terms to include in the terminator:
+        terminator_max_k = 1000
+
+        for lamt, Gamma, Om in zip(lam, gamma, w0):
+            ck_real, vk_real, ck_imag, vk_imag = UD._matsubara_params(
```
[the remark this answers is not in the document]
mcditoos: That's a good suggestion! just implemented it in the last commit

qutip/solver/heom/bofin_baths.py
```diff
@@ -973,10 +1057,621 @@ def _calc_eps(self, Nk):
...
+            np.array(ckAI).flatten(),
+            np.array(vkAI).flatten(),
+        ]
+        self.ckAR = np.array(ckAR).flatten()
+        self.ckAI = np.array(ckAI).flatten()
+        self.vkAR = np.array(vkAR).flatten()
+        self.vkAI = np.array(vkAI).flatten()
+
+        self.terminator = terminator
+        self.Bath_spec = BosonicBath(
```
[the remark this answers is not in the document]
mcditoos: You're right, not instanciating it was also a easy fix, I ended up removing that line and just calling super().__init__ to create the exponents. but not sure if that's the best way to do it, though I think it's better than this one, I'll wait for Simon's comments

qutip/solver/heom/bofin_baths.py
```diff
@@ -973,10 +1057,621 @@ def _calc_eps(self, Nk):
...
+                self.fit_time_imag, self.rmse_imag, self.Ni,
+                "The Imaginary Part \n Of the Correlation Function", lam2,
+                gamma2, w02)
+
+            return summary, summary2
+
+        except NameError:
+            return "Fit correlation instance: \n No fit has been performed yet"
+
+    def summary(self):
```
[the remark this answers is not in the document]
mcditoos: Thanks :)

qutip/solver/heom/bofin_baths.py
```diff
@@ -973,10 +1057,621 @@ def _calc_eps(self, Nk):
...
+        (and the bath parameters).
+        """
+        return (
+            self.ohmic_spectral_density(w)
+            * ((1 / (np.e ** (w * self.beta) - 1)) + 1)
+            * 2
+        )
+
+
+# Fitting Functions
```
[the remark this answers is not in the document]
mcditoos: I think you're right, specially if the AAA stuff ends up working, because then code just for fitting would get longer, not sure if I should move those now, or wait for a high level review on the structure of these things

qutip/tests/solver/heom/test_bofin_baths.py
```diff
@@ -463,3 +472,247 @@ def test_create(self):
...
+    def test_power_spectrum(self):
+        self.bath.params_real = [[1] * 2] * 3
+        self.bath.params_imag = [[1] * 2] * 3
+        self.bath._matsubara_coefficients()
+        S = self.bath.power_spectrum(np.array([1]))
+        assert np.isclose(S, np.array([-2.4 + 0j]))
+
+
+@pytest.mark.skipif('mpmath' not in sys.modules,
+                    reason="requires the mpmath library")
```
pmenczel, line 662: This is a bit weird since `'mpmath'` will not be in `sys.modules` until it is imported, even if it is available. I did some googling, and the recommended way to handle such a situation seems to be `pytest.importorskip`. I don't see a convenient way to apply that to a whole class, the best way seems to be to add `pytest.importorskip("mpmath")` at the beginning of each test in this class.

pmenczel (commented):
Thank you for the updates! A few more minor comments on the new changes.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -105,6 +105,7 @@ class HierarchyADOs:
     labels: list of tuples
         A list of the ADO labels within the hierarchy.
     """
+
```
hodgestar, line 108: Could you revert all of these additional newlines? There is no need for this PR to touch this file at all.

qutip/tests/solver/heom/test_bofin_baths.py
```diff
@@ -463,3 +472,247 @@ def test_create(self):
                 sigma_bar_k_offset=-1,
                 tag="bath1",
             )
+
+
+class TestFitUtils:
+    cks = [1] * 5
```
hodgestar, line 478: This looks unused. Could you remove it? I saw @pmenczel found quite a few unused things too. Could you be a bit more careful about leaving cruft like this lying around? It's tricky for reviewers to spot.

qutip/tests/solver/heom/test_bofin_baths.py
```diff
@@ -463,3 +472,247 @@ def test_create(self):
...
+        assert np.isclose(vkAr, vkAr2).all()
+        assert np.isclose(ckAI, ckAI2).all()
+        assert np.isclose(vkAI, vkAI2).all()
+
+
+class TestFitCorr:
+    alpha = 0.005
+    wc = 1
+    T = 1
+    bath = FitCorr(sigmax())
```
hodgestar, line 612: Just move these into the tests that need them since they're only one extra line usually. Alternatively, write a pytest fixture if you need common setup for tests.

qutip/tests/solver/heom/test_bofin_baths.py
```diff
@@ -463,3 +472,247 @@ def test_create(self):
...
+                [(0.5 + 0j), (0.5 + 0j), (0.5 - 0j), (0.5 - 0j)],
+                [(-1 - 1j), (-1 - 1j), (-1 + 1j), (-1 + 1j)],
+                [-0.5j, -0.5j, 0.5j, 0.5j],
+                [(-1 - 1j), (-1 - 1j), (-1 + 1j), (-1 + 1j)],
+            ]
+        )
+        self.bath.params_real = [[1] * 2] * 3
+        self.bath.params_imag = [[1] * 2] * 3
+        self.bath._matsubara_coefficients()
+        assert np.isclose(np.array(self.bath.cvars), ans).all()
```
hodgestar, line 651: 1. Don't the params to attributes like this. If the FitCorr needs parameters, pass them when initializing the object.
2. It's not great if tests call private `_` methods like `_matsubara_coefficients()`, although occasionally it's sensible. This does not look like such a case.
3. Definitely don't have methods make attributes like `cvars` appear. If the method needs to return a result, just return it.

Probably there is a deeper issue here, so the answer might not be to address directly the comments above, but just to change something so that all of the issues go away. I see @pmenczel noted that the `.cvars` were unused and should just be removed anyway?

qutip/tests/solver/heom/test_bofin_baths.py
```diff
@@ -463,3 +472,247 @@ def test_create(self):
...
+@pytest.mark.skipif('mpmath' not in sys.modules,
+                    reason="requires the mpmath library")
+class TestOhmicBath:
+    alpha = 0.5
+    s = 1
+    wc = 1
+    T = 1
+    Nk = 4
+    bath = OhmicBath(T, sigmax(), alpha, wc, s, Nk,
+                     method="spectral", rmse=1e-4)
```
hodgestar, line 670: Same comment here -- move these into individual tests or create a pytest fixture.

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +349,64 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+                ):
+                    S += s(exp.ck, exp.vk)
+                elif exp.type == BathExponent.types['RI']:
+                    S += s(1j*exp.ck2, exp.vk)
+                else:
+                    S += s(1j*exp.ck, exp.vk)
+        return S
+
+    def correlation_function(self, t, full=1):
+        """Computes the correlation function
```
hodgestar, line 386: These and the other public facing docstrings should document the meanings and types of the arguments and the return value.

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +349,64 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+            if (
+                exp.type == BathExponent.types['R'] or
+                exp.type == BathExponent.types['RI']
+            ):
+                real += exp.ck * np.exp(-exp.vk * t)
+            if exp.type == BathExponent.types['I']:
+                imag += exp.ck * np.exp(-exp.vk * t)
+            if exp.type == BathExponent.types['RI']:
+                imag += exp.ck2 * np.exp(-exp.vk * t)
+        if full == 1:
```
hodgestar, line 403: I would get rid of the `full` argument and just always return the complex value. You can likely also drop `real` and `imag` and replace them with just `corr = np.zeros(len(t), dtype=np.complex128)`.

Note that you'll also need to take care of the case where `t` is a scalar (which is nice to support).

qutip/tests/solver/heom/test_bofin_baths.py
mcditoos: Hi, All the functions inside are still used, they are just not part of a class anymore and are now global functions,, I never separated them for tests, mainly because I saw all tests were written for classes and wasn't sure if that was part of the style. I will split the tests into individual tests now. Thanks for pointing it out :)

qutip/solver/heom/bofin_baths.py
mcditoos: Sure, just changed it according to the suggestion!

qutip/solver/heom/bofin_baths.py
mcditoos: Got it, need to improve all of those docstrings :)

qutip/tests/solver/heom/test_bofin_baths.py
mcditoos: Sorry about those, wasn't sure what was the best way to do it, I will look into pytest fictures or move it into individual tests, thank you!

qutip/tests/solver/heom/test_bofin_baths.py
mcditoos: Sure, will do!

qutip/tests/solver/heom/test_bofin_baths.py
mcditoos: 1) I understand, I did it like that because I didn't want to run a fit to test that function but rather wanted an identifiable fixed input, will rethink how to do it, or will simply fit
2)I wasn't aware, thanks !
3) understood, if the same thing is needed in many places then I just run the method in each other method that needs it, right?

Yes I guess the deeper issue was that I didn't really understand how the exponent construction and the combine works so I was using those cvars instead of the exponents, untill Paul suggested using exponents and removing cvars, however I didn't do a good job and they remained in a few places, sorry about that!

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +339,85 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+        ----------
+        t : np.array or float
+            the time at which to evaluare the correlation function
+        epsabs : float
+            Absolute error tolerance
+        epsrel : float
+            Relative error tolerance
+        """
+        def c_i(w, t): return (1/np.pi)*self.spectral_density(w)*(
+            (1/np.tanh(w/(2*self.T)))*np.cos(w*t) - 1j*np.sin(w*t))
```
nwlambert, line 359: does it make sense to also have a check for zerodivisionerror here, and if it catches it use np.sign() or something instead?

qutip/solver/heom/__init__.py
```diff
@@ -53,3 +56,9 @@
     HierarchyADOs,
     HierarchyADOsState,
 )
+
+from .bofin_fit import (
```
hodgestar, line 60: Great that these got split out into their own module.

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +346,181 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+            Energy of the mode
+
+        Returns
+        -------
+        The population of the mode with energy w
+        """
+
+        if self.T is None:
+            raise NotImplementedError(
+                "Bath temperature must be specified for this operation")
```
hodgestar, line 419:
```suggestion
            raise ValueError(
                "Bath temperature must be specified for this operation")
```
This is not technically a `NotImplementedError`. The method is implemented, it's just missing information it needs. I suggested `ValueError` but there might be other correct answers.

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +346,181 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+        if self.T == 0:
+            return np.zeros_like(w)
+
+        w = np.array(w, dtype=float)
+        with np.errstate(divide='ignore'):  # return inf if w=0
```
hodgestar, line 424: When can a division by 0 occur here? We handle the case where `T = 0` separately above.

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +346,181 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+        ----------
+        The power spectrum of the mode with energy w
+        """
+
+        # For w=0, the result would have the form 0 * inf
+        # We shift w by an epsilon to get the value in the limit w->0
+        w = np.array(w, dtype=float)
+        w[w == 0.0] += 1e-6
+
+        S = (2 * np.sign(w) * self.spectral_density(np.abs(w)) *
```
hodgestar, line 448: It's not clear to me where the `inf` would come from in the `0 * inf` mentioned in the comment. Possibly from `_bose_einstein`? But see the question higher up that points out that `_bose_einstein` currently returns zeroes when `T = 0`.

In any case, it would be nice to handle this better if we can.

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +346,181 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+
+        Returns
+        ----------
+        The power spectrum of the mode with energy w
+        """
+
+        S = np.zeros_like(w, dtype=float)
+        for exp in self.exponents:
+            ck = exp.ck or 0
+            ck2 = exp.ck2 or 0
```
hodgestar, line 499: Please check the bath type here rather than the values of `ck` and `ck2` like we do everywhere else.

qutip/solver/heom/bofin_baths.py
```diff
@@ -412,8 +594,9 @@ def terminator(self):
         delta, L = self._dl_terminator.terminator(self.exponents)
         return delta, L
 
-    def _matsubara_params(self, lam, gamma, T, Nk):
-        """ Calculate the Matsubara coefficents and frequencies. """
+    @staticmethod
```
hodgestar, line 597: It's a bad idea to make this a staticmethod unless it's called from outside the class because it limits refactorying opportunties. E.g. If this method wants to call other methods, it now has to do `ClassName._other_method(...)` instead of `self._other_method`.

Making it a classmethod is also acceptable.

qutip/solver/heom/bofin_baths.py
```diff
@@ -428,6 +611,46 @@ def _matsubara_params(self, lam, gamma, T, Nk):
 
         return ck_real, vk_real, ck_imag, vk_imag
 
+    def spectral_density(self, w):
+        """
+        Calculates the DrudeLorentz spectral density, see Eq. 15 in the BoFiN
+        paper (DOI: 10.1103/PhysRevResearch.5.013181).
```
hodgestar, line 617: Could we write the actual equation here so that users don't have to search for a paper to see what the function does?

hodgestar (commented):
Partial review, mostly of the code in `bofin_baths`.

qutip/solver/heom/bofin_baths.py
pmenczel: If omega is zero, then the denominator (`exp(w / T) - 1`) is zero.

qutip/solver/heom/bofin_baths.py
pmenczel: The Bose-Einstein factor at w=0 diverges (see above). At the same time, the spectral density at w=0 is zero.

More precisely, the Bose-Einstein factor diverges as 1/w. Therefore, the power spectrum at w=0 can be zero, finite or infinite depending on the behavior of the SD at small w (subohmic, ohmic or superohmic). Without that line 446, we would return NaN at w=0. The best idea we could come up with was evaluating the expression somewhere close to zero instead. If you have a better idea, we would be happy to hear it.

(Note that this function is only a utility function for the purpose of visualization etc, so it might be not too important to get this 100% exact.)

qutip/solver/heom/bofin_baths.py
pmenczel: It is called from outside the class. However, it would also work as a classmethod.

hodgestar: I should probably have not done reviewing so late at night. :/

pmenczel: Oh my previous comment wasn't quite correct either though, sorry. This function here does not get called from outside, but its counterpart in the underdamped bath class does. Might be best to make both into classmethods, then?

qutip/solver/heom/bofin_baths.py
mcditoos: Sorry for the delay, This is to set the value of the ck from None to 0 to use it in calculation, we do check the bath.exponent below like everywhere else, but if we just check the bath exponent then some cks will be None. Could move this bit to the exponent types that have Nones for ck or ck2

qutip/solver/heom/bofin_baths.py
mcditoos: Sure, should have done so from the beginning, I just wasn't aware of how easy it is

mcditoos:
Sorry for the delay on the docstrings and so on, I think all suggestions should be incorporated now.  However, thanks to Paul's suggestion of adapting the tutorial 1e. I realized that I made  a "mistake from the very beggining", namely the function used to fit the correlation function is only good if the imaginary part of the spectral density is 0 at t=0, because I'm using

$C(t)=\sum_{k}a_{k}e^{-b_{k} t}e^{i c_{k} t}$

so 

$C_{R}(t)=\sum_{k}a_{k}e^{-b_{k} t}cos(c_{k} t)$

$C_{I}(t)=\sum_{k}a_{k}e^{-b_{k} t}sin(c_{k} t)$

No matter how many terms I take $C_{I}(0)=0$. I should have realized before but this can be bypassed by excluding $t=0$ from the fit, and starting at a small t,  you get a decent fit, as long as you don't look at C(0), and I never check for these situations before. The results of doing this are pretty decent, but updating tutorial 1e I realized it required higher calculation times for the HEOMSolver when compared to using another ansatz for the correlation function, since solving is the expensive bit then I changed the code use another function for fitting

$C(t)=\sum_{k}(a_{k}+i d_{k})e^{-b_{k} t}e^{i c_{k} t}$

The downside is that since this function has 4 parameters, it required changing other functions as well, luckily only functions starting with _, hopefully this won't delay the PR too much. I updated the tutorials with this as well, and added a bit so that passing the parameters for the fitting is easier  (after this comment), if the change is not desirable we can just go back to the commit before this comment

qutip/solver/heom/bofin_baths.py
pmenczel: The same applies to the `spectral_density` method of the underdamped class

pmenczel: The same applies to the spectral_density method of the underdamped class

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1011 @@
...
+    @staticmethod
+    def _meier_tannor_SD(w, a, b, c):
+        r"""
+        Underdamped spectral density used for fitting in Meier-Tannor form
+        (see Eq. 38 in the BoFiN paper, DOI: 10.1103/PhysRevResearch.5.013181)
+
+        $J(\omega) = \sum_{i=1}^{k} \frac{2 \alpha_{i}^{2} \Gamma_{i} \omega
+        }{\left( \left( \omega + \Omega_{i}\right)^{2} + \Gamma_{i}^{2}
+        \right)+\left( \omega - \Omega_{i}\right)^{2} + \Gamma_{i}^{2}
+        \right)}$
```
pmenczel, line 77: This is not the formula used in the code (coupling constant not squared).
Also, we should call the parameters either a,b,c or alpha,Gamma,Omega.
(However, I don't really think we need the equation here, since the method is not public...)

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1011 @@
...
+        final_rmse : float
+            Desired normalized root mean squared error.
+        lower : list
+            lower bounds on the parameters for the fit. A list of size 3,
+            each value represents the lower bound for each parameter.
+            The order of the parameters is the same as for the function to be
+            fitted (_meier_tannor_SD). The first term describes the coupling,
+            the second the cutoff frequency,and the last one the central
+            frequency. The lower bounds are considered to be the same for all
+            N modes. for example
```
pmenczel, line 128: Thank you for the explanation. However, better remove the mention of _meier_tannor_SD, since that method will not appear in the documentation.

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1011 @@
...
+        ----------
+        t : np.array or float
+            The times at which to evaluates the correlation function.
+        a : list or np.array
+            A list describing the amplitude of the correlation approximation.
+        b : list or np.array
+            A list describing the decay of the correlation approximation.
+        c : list or np.array
+            A list describing the oscillations of the correlation
+            approximation.
```
pmenczel, line 292: `d` is missing

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1011 @@
...
+            not provided.
+        lower : list
+            lower bounds on the parameters for the fit.
+        upper: list
+            upper bounds on the parameters for the fit.
+        sigma: float
+            uncertainty in the data considered for the fit.
+        guesses : list
+            Initial guess for the parameters.
+        """
```
pmenczel, line 680: Return missing

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1011 @@
...
+            upper.
+
+        Note: If one of lower, upper, sigma, guesses is None,
+        all are discarded.
+
+        Returns
+        -------
+        * A Bosonic Bath created with the fit parameters for the original
+          spectral density function (that was provided or interpolated)
+        * A dictionary containing the following information about the fit:
```
pmenczel, line 152: These produce weird output in the docs, but I don't quite understand why :/

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+            self._w = w
+            self._J_array = J(w)
+            self._J_fun = J
+        else:
+            self._w = w
+            self._J_array = J
+            self._J_fun = InterpolatedUnivariateSpline(w, J)
+
+    @staticmethod
+    def _meier_tannor_SD(w, a, b, c):
```
pmenczel, line 69: Since Simon recommended to prefer `classmethod` over `staticmethod`.

```suggestion
    @classmethod
    def _meier_tannor_SD(cls, w, a, b, c):
```

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+        ----------
+        N : optional, int
+            Number of underdamped oscillators to use.
+            If set to None, it is determined automatically.
+        Nk : optional, int
+            Number of exponential terms used to approximate the bath
+            correlation functions, per bath. Defaults to 1.
+        final_rmse : float
+            Desired normalized root mean squared error. Defaults to
+            $5\times10^{-6}$
```
pmenczel, line 117:
```suggestion
            Desired normalized root mean squared error. Defaults to
            :math:`5\times10^{-6}`. Only used if N is set to None.
```

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+        Parameters
+        ----------
+        w : np.array
+            The frequency of the spectral density
+        a : np.array
+            Array of coupling constants
+        b : np.array
+            Array of cutoff parameters
+        c : np.array
+            Array of resonant frequencies
```
pmenczel, line 84:
```suggestion
        a : np.array
            Array of coupling constants ($\alpha_i^2$)
        b : np.array
            Array of cutoff parameters ($\Gamma'_i$)
        c : np.array
            Array of resonant frequencies ($\Omega_i$)
```

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+
+            .. math::
+                J(\omega) = \sum_{i=1}^{k} \frac{2 a_{i} b_{i} \omega
+                }{\left(\left( \omega + c_{i}\right)^{2} + b_{i}^{2}\right)
+                \left(\left( \omega - c_{i}\right)^{2} + b_{i}^{2} \right)}
+
+            The first term describes the coupling,
+            the second the cutoff frequency,and the last one the central
+            frequency. The lower bounds are considered to be the same for all
+            N modes. for example
```
pmenczel, line 132: (Just to make this a bit more concise)

```suggestion
            Lower bounds on the parameters for the fit. A list of size 3,
            containing the lower bounds for :math:`a_i` (coupling constants),
            :math:`b_i` (cutoff frequencies) and :math:`c_i`
            (resonant frequencies) in the following fit function:

            .. math::
                J(\omega) = \sum_{i=1}^{k} \frac{2 a_{i} b_{i} \omega
                }{\left(\left( \omega + c_{i}\right)^{2} + b_{i}^{2}\right)
                \left(\left( \omega - c_{i}\right)^{2} + b_{i}^{2} \right)}

            The lower bounds are considered to be the same for all N modes.
            For example,
```

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+            Initial guesses for the parameters. Same structure as lower and
+            upper.
+
+        Note: If one of lower, upper, sigma, guesses is None, all are discarded
+
+        Returns
+        -------
+        - A Bosonic Bath created with the fit parameters for the original
+          spectral density function (that was provided or interpolated)
+        - A dictionary containing the following information about the fit:
```
pmenczel, line 155: The output of this looks a bit broken (unsure how to get it right, though)

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+        N:
+            The number of terms used for the fit.
+        params:
+            The fitted parameters (3*N parameters), it contains three lists
+            one for each parameter, each list containing N terms.
+        Nk:
+            The number of exponents used to construct the bosonic bath,
+            defaults to 1. To approximate the correlation function the
+            number of exponents grow as the temperature decreases, so Nk
+            needs to be adjusted accordingly.
```
pmenczel, line 169: Does this text belong here, or rather above at the Nk parameter?

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+        Nr : optional, int
+            Number of exponents to use for the real part.
+            If set to None it is determined automatically.
+        Ni : optional, int
+            Number of exponents terms to use for the imaginary part.
+            If set to None it is found automatically.
+        final_rmse : float
+            Desired normalized root mean squared error. Only used if Ni or Nr
+            are not specified.
+        lower : list
```
pmenczel, line 344: It's good to have that `full_ansatz` as an option now! Maybe this should just be a list of length 3 if `full_ansatz` is false?

qutip/solver/heom/bofin_fit.py
```diff
@@ -0,0 +1,1095 @@
...
+
+
+
+        Note: If one of lower, upper, sigma, guesses is None, all are discarded
+
+        Returns
+        -------
+        - A Bosonic Bath created with the fit parameters from the original
+          correlation function (that was provided or interpolated).
+        - A dictionary containing the following information about the fit:
```
pmenczel, line 401: Same formatting thing here

pmenczel:
Somehow vscode published those review comments above before I was done reviewing. Only one more comment though: in the tests, you ignore some RuntimeWarnings here and there; it would be good to add a comment explaining why. Thank you for adding a good amount of tests!

If these last minor points get addressed, this looks good to me. Only those utility functions in the `bofin_fit` module are a bit of a mess - if we ever want to add more fitting methods, it will require some work. But it is non-public api and it seems to do the job for now.

qutip/solver/heom/bofin_baths.py
```diff
@@ -662,13 +903,16 @@ def __init__(
...
-        super().__init__(
-            Q, ck_real, vk_real, ck_imag, vk_imag, combine=combine, tag=tag,
-        )
+        super().__init__(Q, ck_real, vk_real, ck_imag, vk_imag,
+                         combine=combine, tag=tag, T=T)
 
-    def _matsubara_params(self, lam, gamma, w0, T, Nk):
-        """ Calculate the Matsubara coefficents and frequencies. """
+    @classmethod
+    def _matsubara_params(cls, lam, gamma, w0, T, Nk):
```
pmenczel, line 914: This method fails when called with `T=0`

qutip/solver/heom/bofin_baths.py
```diff
@@ -337,6 +346,188 @@ def combine(cls, exponents, rtol=1e-5, atol=1e-7):
...
+            The correlation function at time t as an array or float.
+        """
+
+        def integrand(w, t):
+            return self.spectral_density(w) / np.pi * (
+                (2 * self._bose_einstein(w) + 1) * np.cos(w * t)
+                - 1j * np.sin(w * t)
+            )
+
+        result = quad_vec(lambda w: integrand(w, t), 0, np.Inf, **kwargs)
```
pmenczel, line 399: AttributeError: `np.Inf` was removed in the NumPy 2.0 release. Use `np.inf` instead.

pmenczel (commented):
I tried to use this today, but ran into some problems. I know the code here is quite old by now and will probably change a lot when we add the general bath class, but I wanted to note down the bugs somewhere.

Issue #791, opened 2017-12-29
Problem when using MKL


ingstra:
Hello! I'm trying to use the steadystate solver. When not using MKL, my program runs (albeit a bit slow). When using MKL, I get the following error message. I have tried to figure it out myself, to no avail.

Traceback (most recent call last):
  File "cubicH_quantumPump_loop.py", line 54, in <module>
    rho = steadystate(H, c_ops)
  File "/home/ingrid/anaconda3/lib/python3.6/site-packages/qutip/steadystate.py", line 234, in steadystate
    return _steadystate_direct_sparse(A, ss_args)
  File "/home/ingrid/anaconda3/lib/python3.6/site-packages/qutip/steadystate.py", line 400, in _steadystate_direct_sparse
    v = mkl_spsolve(L, b, perm = in_perm, verbose = ss_args['verbose'])
  File "/home/ingrid/anaconda3/lib/python3.6/site-packages/qutip/_mkl/spsolve.py", line 384, in mkl_spsolve
    x = lu.solve(b, verbose=verbose)
  File "/home/ingrid/anaconda3/lib/python3.6/site-packages/qutip/_mkl/spsolve.py", line 167, in solve
    raise Exception(pardiso_error_msgs[str(error)])
KeyError: '[-4]'

nonhermitian:
You may want to try the development version.  However, it is saying that the MKL solver is failing with a zero pivot.  At present, it is not possible to modify the underlying parameters for the solver, but it is something I want to do.  The MKL solver is quite fast but sometimes runs into trouble.

pranavm1502:
Hi! I am also facing the same problem  with steadystate solver. I am also unaware of how to "not use" MKL for the solver. Is there some settings which can be changed to turn it off?

nonhermitian:
This should be addressed in Pull #801.  The only other reported instance of error when using the MKL solver was resolved when the new kwarg `weighted_matching=False` is used in the steady state solver.

PR #2261, opened 2023-11-09
HEOM Parity support

gsuarezr:
**Description**

This PR adds odd parity support to HEOM's fermionic baths, the user can choose whether to use odd or even parity according to his needs when construncting the right hand side of HEOM's equations

**Related issues or PRs**

Heom tutorial  for this feature https://github.com/qutip/qutip-tutorials/pull/74


hodgestar:
@mcditoos Thanks more making the PR. Would you mind reverting all the unrelated changes? We don't mind small clean ups, but big clean ups should rather go into separate PRs and it should be clear why they are being made.

Once you're ready for this to be reviewed, remove the draft status and ping me on the PR.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -602,9 +613,9 @@ class HEOMSolver(Solver):
         "state_data_type": "dense",
     }
 
-    def __init__(self, H, bath, max_depth, *, options=None):
+    def __init__(self, H, bath, max_depth, odd_parity=False, *, options=None):
```
hodgestar, line 616:
```suggestion
    def __init__(self, H, bath, max_depth, *, odd_parity=False, options=None):
```
Let's make the `odd_parity` argument keyword only too.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -745,6 +756,16 @@ def _grad_prev_bosonic(self, he_n, k):
             )
         return op
 
+    def _get_signs(self, n_excite, n_excite_before_m):
```
hodgestar, line 759: Extracting the signs out into this function destroys the readability of the logic. Functions should not be use to merely reduce repetitive, but rather to abstract out concepts that clarify understanding.

This function does the opposite -- it makes the logic *harder* to understand.

I'll suggest an alternative below.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -753,11 +774,10 @@ def _grad_prev_fermionic(self, he_n, k):
         ]
 
         n_excite = sum(he_fermionic_n)
-        sign1 = (-1) ** (n_excite + 1)
-
```
hodgestar, line 757: Instead of removing this line and hiding the mathematics, I suggest doing:

`sign1 = (-1) ** (n_excite + 1 - self.odd_partiy)`

and

`sign2 = (-1)**(n_excite_before_m + self.odd_party)`.

This makes the mathematics clearer (I think). It relies on Python numerically treating `False` and `True` as `0` and `1`, which is perhaps worth a comment which I'll suggest further up.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -602,9 +613,9 @@ class HEOMSolver(Solver):
         "state_data_type": "dense",
     }
 
-    def __init__(self, H, bath, max_depth, *, options=None):
+    def __init__(self, H, bath, max_depth, odd_parity=False, *, options=None):
         _time_start = time()
-
+        self.odd_parity = odd_parity
```
hodgestar, line 618:
```suggestion
        self.odd_parity = bool(odd_parity)  # we call bool here because odd_parity will be used in arithmetic
```

hodgestar (changes requested):
Thank you for removing all the unnecessary changes. The parity support changes are much clearer now.

I've request some small changes. In addition we will need a test for the new feature and ideally for the feature to be mentioned in the HEOM guide in the documentation (perhaps also with a small example).

qutip/solver/heom/bofin_solvers.py
mcditoos: You are right I think this indeed clearer

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -54,8 +54,8 @@ def assert_raises_steady_state_time_dependent(hsolver):
     with pytest.raises(ValueError) as err:
         hsolver.steady_state()
     assert str(err.value) == (
-       "A steady state cannot be determined for a time-dependent"
-       " system"
+        "A steady state cannot be determined for a time-dependent"
```
hodgestar, line 57: This can stay -- it's a valid indentation fix.

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -232,6 +232,7 @@ class DrudeLorentzPureDephasingModel:
     """ Analytic Drude-Lorentz pure-dephasing model for testing the HEOM
         solver.
     """
+
```
hodgestar, line 235: Remove these too.

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -9,7 +9,7 @@
 
 from qutip import (
     basis, destroy, expect, liouvillian, qeye, sigmax, sigmaz,
-    tensor, Qobj, QobjEvo
+    tensor, Qobj, QobjEvo, fcreate, fdestroy
```
hodgestar, line 12: `fcreate` appears to be unused. Could you remove it?

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -1127,6 +1131,46 @@ def test_solving_with_step(self):
...
+        resultHEOMPade = HEOMSolver(L, [bath1, bath2], depth, odd_parity=True)
+        rhoss, _ = resultHEOMPade.steady_state(use_mkl=False)
+        rhoss = rhoss.full()
+        expected_odd = np.diag([-0.18472, 0.68472, 0.68472, -0.18472])
+        expected = np.diag([0.10623, 0.39376, 0.39376, 0.10623])
+        assert np.isclose(rhoss, expected_odd,atol=1e-5).all()
+        resultHEOMPade = HEOMSolver(L, [bath1, bath2], depth, odd_parity=False)
+        rhoss, _ = resultHEOMPade.steady_state(use_mkl=False)
+        rhoss = rhoss.full()
+        assert np.isclose(rhoss, expected,atol=1e-5).all()
```
hodgestar, line 1170:
```suggestion
        solver = HEOMSolver(L, [bath1, bath2], depth, odd_parity=True)
        rhoss, _ = solver.steady_state(use_mkl=False)
        rhoss = rhoss.full()
        expected_odd = np.diag([-0.18472, 0.68472, 0.68472, -0.18472])
        np.testing.assert_allclose(rhoss, expected_odd, atol=1e-5)

        solver = HEOMSolver(L, [bath1, bath2], depth, odd_parity=False)
        rhoss, _ = solver.steady_state(use_mkl=False)
        rhoss = rhoss.full()
        expected = np.diag([0.10623, 0.39376, 0.39376, 0.10623])
        np.testing.assert_allclose(rhoss, expected, atol=1e-5)
```

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -1127,6 +1131,46 @@ def test_solving_with_step(self):
...
+
+
+
```
hodgestar, line 1173: There appear to be too many blank lines here.

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -1127,6 +1131,46 @@ def test_solving_with_step(self):
...
+        L = liouvillian(H)
+        bath1 = LorentzianPadeBath(
+            Q=d_1, gamma=2 * Gamma, w=W, mu=mu, T=1 / beta, Nk=Nk,
+            tag="Lead 1")
+        bath2 = LorentzianPadeBath(
+            Q=d_2, gamma=2 * Gamma, w=W, mu=mu, T=1 / beta, Nk=Nk,
+            tag="Lead 2")
+        resultHEOMPade = HEOMSolver(L, [bath1, bath2], depth, odd_parity=True)
+        rhoss, _ = resultHEOMPade.steady_state(use_mkl=False)
```
hodgestar, line 1162: Could you explain why you needed to pass `use_mkl=False` here? I think we would expect it to automatically do something sensible for users in all cases and we don't appear to have had to use it in any of the other HEOM solver tests yet.

hodgestar (changes requested):
Thanks for the changes. Next round of reviewing done.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -566,6 +568,17 @@ class HEOMSolver(Solver):
...
+        For fermionic baths only.  Default is "False".  If set to "True",
+        the RHS construction differs slightly (it implies the RHS is acting on
+        an initial density operator which has odd parity in terms of its
+        representation in fermionic annhilation operators). In other words, a
+        state with a well defined number of excitations has even parity, but
+        one with superpositions of the number of excitations has odd parity.
+        This is normally prohibited in 'normal' electronic systems, where
+        charge is conserved, but is useful for the construction of
+        electron spectrum (under linear response assumptions).
+
```
nwlambert, line 581: Simon asked us to help distill and clarify this a little bit; how about:

" For fermionic baths only.  Default is "False".  "Parity" refers to the parity of the initial system state used with the HEOM. An odd parity state is one that can be made from an applying an odd number of fermionic creation operators to the vacuum density operator.  Physical systems have even parity, but allowing the generalization to odd-parity states allows one to calculate useful physical quantities like the system power spectrum or density of states. The form of the HEOM differs depending on the parity of the initial system state, so if this option is set to "True", a different RHS is constructed, which can then be used with a system state of odd parity. "

qutip/tests/solver/heom/test_bofin_solvers.py
mcditoos: I used used_mkl=False, because if I don't for the test fails for python 3.10 with no cython (only in that setup it passes in all others), the test fails with Exception: Zero pivot, numerical factorization or iterative refinement problem. I found this old thread https://github.com/qutip/qutip/issues/791 suggesting that sometimes mkl solvers give some trouble so I decided to set it to false

hodgestar (approved):
Thanks @gsuarezr!

Issue #2268, opened 2023-11-24
simdiag does not always return orthonormal eigenvectors


Sola85:
### Bug Description

In some rare cases, like the example below, simdiag does not return orthonormal eigenvectors, despite this always being possible.

### Code to Reproduce the Bug

```shell
import numpy as np
from qutip import simdiag, Qobj

a = np.array([[1,  0, 1, -1, 0],
              [0,  4, 0,  0, 1],
              [1,  0, 4,  1, 0],
              [-1, 0, 1,  4, 0],
              [0,  1, 0,  0, 4]])

b = np.eye(5)

evals, evecs = simdiag([Qobj(a), Qobj(b)])
evecs = np.array([evec.full() for evec in evecs]).squeeze()

# Check that eigenvectors form an othonormal basis (<=> matrix of eigenvectors is unitary)
deviation_from_unitarity = np.linalg.norm(evecs@evecs.conj().T - np.eye(len(evecs)))
assert np.isclose(deviation_from_unitarity, 0)
```

### Code Output

```shell
AssertionError
```

### Expected Behaviour

Since by requirement of simdiag all input matrices must be hermitian and commute, the common eigenvectors should always be orthonormal.

### Your Environment

```shell
QuTiP Version:      4.7.1
Numpy Version:      1.23.5
Scipy Version:      1.10.1
Cython Version:     None
Matplotlib Version: 3.7.1
Python Version:     3.11.6
Number of CPUs:     8
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
Platform Info:      Linux (x86_64)
```

### Additional Context

The fix is simple: 
Here
```python
subspace = vecs.conj().T @ ops[i].full() @ vecs
eigvals, eigvecs = la.eig(subspace)
```
`la.eig` has to be replaced by `la.eigh`

hodgestar:
Thanks @Sola85. Would you mind trying this on 4.7.3?

Sola85:
> Thanks @Sola85. Would you mind trying this on 4.7.3?

Yes, this is reproducable in 4.7.3 and in the current master branch

Sola85:
Fixed in PR #2269

PR #2269, opened 2023-11-24
fix simdiag not returning orthonormal eigenvectors

Sola85:
**Description**
Use `scipy.linalg.eigh` instead of `scipy.linalg.eig` for `qutip.simdiag` in order to ensure returned eigenvectors are always orthonormal. Since `simdiag` requires operands to be commuting and hermitian, this change should be justified.

I did not add any new tests as this bug only occured in very rare circumstances. If desired, I could add the reproducing code from the related issue as a test case.

**Related issues or PRs**
Fixes issue #2268

### DIFF
--- a/doc/changes/2269.bugfix
+++ b/doc/changes/2269.bugfix
@@ -1,0 +1,1 @@
+Fixed simdiag not returning orthonormal eigenvectors.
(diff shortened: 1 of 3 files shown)


Ericgig (commented):
Thank you for the fix.

Could you add a test that would fail if simdiag did not return orthonormal vector?
The example in the issue would work.

Ericgig (approved):

PR #2272, opened 2023-11-28
Fix LaTeX display of Qobj state in Jupyter cell outputs

### LINKED ISSUE #2172: Quantum Object not rendered properly in Latex when using the VSCode IDE
### Bug Description

When printing a Qobj in Jupyter Notebooks of the VSCode IDE, the output is rendered as raw text rather than as Latex equation.

### Code to Reproduce the Bug

```shell
import qutip as qt
qt.fock(2,0)
```

### Code Output

```shell
Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket'\begin{equation*}\left(\begin{array}{{11}c}1\0\end{array}\right)\end{equation}
```

### Expected Behaviour

The output should instead look like:

`Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket'`

image

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+nogit
Numpy Version:      1.24.3
Scipy Version:      1.10.1
Cython Version:     0.29.33
Matplotlib Version: 3.7.1
Python Version:     3.8.16
Number of CPUs:     8
BLAS Info:          OPENBLAS
INTEL MKL Ext:      False
```

### Additional Context

I see already there are some errors in the latex code output. 
- `\end{equation}` -> `\end{equation*}`
- `{{11}c}`: is this format correct?. I would rather do: `{cc}`

ejthomas:
**Description**
When a `Qobj` is present in the output of a Jupyter cell, e.g. when running:
```python
import qutip as qt
qt.fock(2, 0)
```
the output, which contains LaTeX code, is displayed in Markdown. The aim of this PR is to display the LaTeX output correctly. Replacing the `equation` environment in the LaTeX code with `$$...$$` and changing the `array` environment alignment from `*{11}c` to `cc` causes the vectors and matrices to display in the expected mathematical typeface. A test has not yet been provided, because the change was to a stored raw string; if a test is needed then the presence of this syntax in the string could be detected.

**Related issues or PRs**
Fixes #2172

### DIFF
--- a/doc/changes/2172.bugfix
+++ b/doc/changes/2172.bugfix
@@ -1,0 +1,1 @@
+Fix LaTeX display of Qobj state in Jupyter cell outputs
--- a/doc/changes/2172.bugfix
+++ b/doc/changes/2172.bugfix
@@ -1,1 +1,0 @@
-Fix LaTeX display of Qobj state in Jupyter cell outputs
--- a/doc/changes/2272.bugfix
+++ b/doc/changes/2272.bugfix
@@ -1,0 +1,1 @@
+Fix LaTeX display of Qobj state in Jupyter cell outputs
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -603,10 +603,10 @@ class Qobj:
             cols.append(None)
             cols += list(range(n_cols - half_length, n_cols))
         # Make the data array.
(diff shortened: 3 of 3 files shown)


Ericgig (approved):
Thank you for the contribution.

Issue #2263, opened 2023-11-15
Normalization problem with choi_to_kraus()


stserkis:
### Bug Description

I have a stabilizer circuit with five qubits. A measurement is happening in one of them so we can end up with two channels for the rest of the qubits for each measurement outcome 0 and 1. For each channel I can directly extract the corresponding Kraus operators and the total set of them (including both channels) satisfies the resolution to the identity (which is expected). However, if I first extract the choi matrices for both channels and from them I then extract the Kraus operators through the function choi_to_kraus(), the resolution to the identity is not satisfied.

I attach below two txt files, that contain the choi matrices in Qobj form
[choi_even.txt](https://github.com/qutip/qutip/files/13360165/choi_even.txt)
[choi_odd.txt](https://github.com/qutip/qutip/files/13360166/choi_odd.txt)

### Code to Reproduce the Bug

```shell
kraus_list_even = choi_to_kraus(choi_even)
kraus_list_odd = choi_to_kraus(choi_odd)
print(sum(dag(op) * op for op in kraus_list_even) + sum(dag(op) * op for op in kraus_list_odd))
```

### Code Output

```shell
Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[1.00505236 0.         0.         0.        ]
[0.         1.00466051 0.         0.        ]
[0.         0.         0.99533949 0.        ]
[0.         0.         0.         0.99494764]]
```

### Expected Behaviour

Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[1. 0. 0. 0.]
[0. 1. 0. 0.]
[0. 0. 1. 0.]
[0. 0. 0. 1.]]

### Your Environment

```shell
QuTip Version: 4.7.3
Numpy Version: 1.24.4 
Python Version: 3.9.13
```

### Additional Context

I've tried changing the tolerance in the function but the normalization is still wrong.

PR #2276, opened 2023-12-05
Make `choi_to_kraus` rely on eigenstate solver for hermitian matrices if the Choi matrix is Hermitian

theodotk:
**Description**
Make `choi_to_kraus` rely on eigenstate solver for Hermitian matrices if the Choi matrix is Hermitian. Basically, this adapts what is done in qutip 5, while waiting for its release

**Related issues or PRs**
#2263


Ericgig (commented):
Thank you for the PR.

Could you fix it so the tests passes.

Ericgig (approved):
Thank you

PR #2280, opened 2023-12-22
Parallelization fixes

pmenczel:
**Background**

Since I am planning to add an mpi_parallel_map to the parallel module, I had a detailed look at the current implementations of parallel_map and loky_pmap. In the case of timeouts, errors or interrupts, I found surprising behavior. I will first summarize the current behavior in these situations, and then the behavior if my changes are included. It seems difficult to write better unit tests for such timing-sensitive behavior, but I have tested in detail on both Linux (WSL) and Windows.

**Current behavior**

&nbsp; | `parallel_map` | `loky_pmap`
---|---|---
Timeout | Finishes currently running tasks, <br> then returns results of all finished tasks. | If `fail_fast`, behaves similarly to `parallel_map`. <br> Otherwise, completes all tasks but only returns <br> results of those that were started before the timeout.
CTRL+C | First CTRL+C interrupts all tasks, but program <br> enters deadlock requiring second CTRL+C | Interrupts currently running tasks, but then still <br> executes the rest before raising `KeyboardInterrupt`.
Task raises <br> exception | `fail_fast`: finishes running tasks, <br> then raises exception. <br> `!fail_fast`: raises `MapExceptions` at the end. | `fail_fast`: finishes all tasks, then raises exception. <br> `!fail_fast`: raises `MapExceptions` at the end.
Job timeout | Ignored | Always completes all tasks anyway, see below

**New behavior**

&nbsp; | `parallel_map` | `loky_pmap`
---|---|---
Timeout | Finishes currently running tasks, <br> then returns results of all finished tasks. | Aborts currently running tasks, <br> then returns results of all finished tasks.
CTRL+C | First CTRL+C raises `KeyboardInterrupt` | First CTRL+C raises `KeyboardInterrupt`
Task raises <br> exception | `fail_fast`: finishes running tasks, <br> then raises exception. <br> `!fail_fast`: raises `MapExceptions` at the end. | `fail_fast`: finishes tasks earlier in the list, then <br> aborts remaining ones and raises exception. <br> `!fail_fast`: raises `MapExceptions` at the end.
Job timeout | Removed from documentation | Removed from documentation

**Job timeout**

Currently, the job timeout parameter is ignored by parallel_map. In loky_pmap, it is not the maximum allowed time for one job, but the maximum time between two job finishes (possibly in different processes). If this time is exceeded, all tasks will still be executed until the end; only the results of the tasks that finished too slowly will be discarded. I do not think that this was the intention of the job timeout parameter?

Unfortunately, both `ProcessPoolExecutor` (which parallel_map is based on) and its loky version do not support timeouts for single tasks, nor do they support aborting single tasks manually. (The loky one supports killing *all* worker processes at once.) If we wanted to have a job timeout parameter, we would need to either use non-public API to obtain references to the worker processes and interrupt them manually, or to completely rewrite parallel_map and base it on e.g. `multiprocessing.pool.Pool`.

Maybe better to just remove the job_timeout parameter? If you agree with that, I will then also remove it from the available options for `MultiTrajSolver` and all its subclasses.

### DIFF
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -203,11 +203,11 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
         if not future.cancelled():
             try:
                 result = future.result()
+                remaining_ntraj = result_func(future._i, result)
+                if remaining_ntraj is not None and remaining_ntraj <= 0:
+                    finished.append(True)
             except Exception as e:
                 errors[future._i] = e
-        remaining_ntraj = result_func(future._i, result)
-        if remaining_ntraj is not None and remaining_ntraj <= 0:
-            finished.append(True)
         progress_bar.update()
 
     if sys.version_info >= (3, 7):
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -208,6 +208,18 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
                     finished.append(True)
             except Exception as e:
                 errors[future._i] = e
+            except KeyboardInterrupt:
+                # When a keyboard interrupt happens, it is raised in the main
+                # thread and in all worker threads. The worker threads have
+                # already returned and the main thread is only waiting for the
+                # ProcessPoolExecutor to shutdown before exiting. If the call
+                # to `future.result()` in this callback function raises the
+                # KeyboardInterrupt again, it makes the system enter a kind of
+                # deadlock state, where the user has to press CTRL+C a second
+                # time to actually end the program. For that reason, we
+                # silently ignore the KeyboardInterrupt here, avoiding the
+                # deadlock and allowing the main thread to exit.
+                pass
         progress_bar.update()
 
     if sys.version_info >= (3, 7):
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -25,7 +25,6 @@ else:
 
 
 default_map_kw = {
-    'job_timeout': threading.TIMEOUT_MAX,
     'timeout': threading.TIMEOUT_MAX,
     'num_cpus': available_cpu_count(),
     'fail_fast': True,
@@ -165,7 +164,6 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
     map_kw: dict, optional
         Dictionary containing entry for:
         - timeout: float, Maximum time (sec) for the whole map.
-        - job_timeout: float, Maximum time (sec) for each job in the map.
         - num_cpus: int, Number of jobs to run at once.
         - fail_fast: bool, Abort at the first error.
 
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -149,15 +149,15 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
         The list or array of values for which the ``task`` function is to be
         evaluated.
     task_args : list, optional
-        The optional additional argument to the ``task`` function.
+        The optional additional arguments to the ``task`` function.
     task_kwargs : dictionary, optional
-        The optional additional keyword argument to the ``task`` function.
+        The optional additional keyword arguments to the ``task`` function.
     reduce_func : func, optional
-        If provided, it will be called with the output of each tasks instead of
-        storing a them in a list. Note that the order in which results are
+        If provided, it will be called with the output of each task instead of
+        storing them in a list. Note that the order in which results are
         passed to ``reduce_func`` is not defined. It should return None or a
-        number. When returning a number, it represent the estimation of the
-        number of task left. On a return <= 0, the map will end early.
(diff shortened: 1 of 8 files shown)


qutip/solver/parallel.py
```diff
@@ -194,20 +191,33 @@ def parallel_map(task, values, task_args=None, task_kwargs=None,
...
                 result = future.result()
+                remaining_ntraj = result_func(future._i, result)
+                if remaining_ntraj is not None and remaining_ntraj <= 0:
+                    finished.append(True)
             except Exception as e:
                 errors[future._i] = e
-        remaining_ntraj = result_func(future._i, result)
-        if remaining_ntraj is not None and remaining_ntraj <= 0:
-            finished.append(True)
+            except KeyboardInterrupt:
```
hodgestar, line 209: If raising `KeyboardInterrupt` in the main thread always causes a deadlock, then catching it here would seem to leave a small window somewhere outside of this `try ... except` and outside of this function where it might not be caught and the deadlock would happen anyway?

This definitely narrows the window where deadlock can occur, but could we do better or even eliminate the deadlock entirely?

Thank you for the great write up in the comment. I wouldn't have been able to ask this question without it.

qutip/solver/parallel.py
```diff
@@ -369,13 +383,11 @@ def loky_pmap(task, values, task_args=None, task_kwargs=None,
         [job.cancel() for job in jobs]
         raise e
 
-    except TimeoutError:
-        [job.cancel() for job in jobs]
-
     finally:
-        executor.shutdown()
+        os.environ['QUTIP_IN_PARALLEL'] = 'FALSE'
+        executor.shutdown(kill_workers=True)
```
hodgestar, line 388: What does `kill_workers=True` do?

qutip/solver/parallel.py
```diff
@@ -369,13 +383,11 @@ def loky_pmap(task, values, task_args=None, task_kwargs=None,
         [job.cancel() for job in jobs]
         raise e
 
-    except TimeoutError:
-        [job.cancel() for job in jobs]
-
     finally:
-        executor.shutdown()
+        os.environ['QUTIP_IN_PARALLEL'] = 'FALSE'
```
hodgestar, line 387: Good catch on moving this into finally. We should have seen that much earlier. Doh.

hodgestar (commented):
This looks like a big improvement to me. I asked a couple of questions about whether it can be improved more.

qutip/solver/parallel.py
pmenczel: I don't really understand why that deadlock happens, but I do believe it should be gone now. What happens is this:
1. User presses CTRL+C -> KeyboardInterrupt is immediately raised in the main process and all worker processes
2. The worker processes return, and the corresponding `Future` objects record that the tasks didn't yield a result, but raised an exception.
3. In the main process, on the way out after the KeyboardInterrupt is raised, we hit the `finally` clause that shuts down the executor. The executor waits for all worker processes to finish and still calls the "done callback" for all the tasks that were currently running.
4. When `future.result()` is invoked in the callback, it raises the KeyboardInterrupt that was recorded in step 2 in the main thread. This is what causes that weird "deadlock".

The point is that the only problematic KeyboardInterrupt is the one that is raised from `future.result()`.

qutip/solver/parallel.py
pmenczel: It kills all worker processes immediately. (That might sound worrisome since killing threads forcibly is a _bad thing_ generally, but these are actual processes, so it is okay afaik.)

In `parallel_map`, if there is an error (and `fail_fast` enabled) or if we hit the timeout, we still cannot return until all currently running tasks finish, since the standard ProcessPoolExecutor does not have such an option. The loky version offers that `kill_workers` flag, making it possible to return immediately in these cases, which seemed useful to me.

qutip/solver/parallel.py
hodgestar: Ah, then there is a small time window during which the `except KeyboardInterrupt` will catch the interrupt from the main thread instead of the raised by `future.result()`. To be completely safe we should not use only `.result()` but also check `future.exception()` (and pass if it is a KeyboardInterrupt).

qutip/solver/parallel.py
hodgestar: I went and read the loky docs and this use of it seems okay. And it is really nice to not have to wait for long worker tasks to complete.

qutip/solver/parallel.py
pmenczel: You are completely right, thank you! I have updated this part so that it checks `future.exception()`.

hodgestar: Thanks!

hodgestar (approved):

hodgestar:
@pmenczel Thanks for sorting this all out. It is much better now. Happy for you to merge it when you're ready.

PR #2283, opened 2024-01-05
Make `kraus_to_choi` faster by using numpy

theodotk:
`np.tensordot` is faster than a triple loop. On my applications the function became up to 100 times faster.

Also, the way to handle `dims` of the Choi matrix was incorrect (or I misunderstood something). It is explained in the comments.

This was coded in collab with [Rafael Haenel](https://github.com/rafaelha).

Also, please, could you make a release of the qutip 4 with all the new changes after this PR passes?

### DIFF
--- a/doc/changes/2283.misc
+++ b/doc/changes/2283.misc
@@ -1,0 +1,1 @@
+Rework `kraus_to_choi`, making it faster
--- a/qutip/superop_reps.py
+++ b/qutip/superop_reps.py
@@ -205,8 +205,8 @@ def kraus_to_choi(kraus_ops):
     r"""
     Convert a list of Kraus operators into Choi representation of the channel.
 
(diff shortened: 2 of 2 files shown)


Ericgig (changes requested):
Thank for for the optimization.

The test failure for 3.6 is due to `np.asarray(Qobj)` failing with old version of numpy.
We don't want to remove support for older version of python and numpy for v4.
Could you fix it by using `Qobj.full` as you did in the v5 version.

Other than that, it's looking good.

Other tests failure are not related to your changes.

Ericgig (approved):

theodotk:
Thanks for merging!
Is it possible to make a release of qutip 4? Or, if you have a set release calendar, when to expect it?

Ericgig:
Hi @theodotk, the 4.7.4 release is in progress.

PR #2288, opened 2024-01-08
Ignore deprecation warnings from cython 0.29

Ericgig:
**Description**
Cython 0.29 is raise a deprecation warning when used with the newest setuptools.

Ignore these warnings in tests.


hodgestar (approved):
I'm okay with ignoring these in master too, as long as we remember to take them out again at some point.

PR #2289, opened 2024-01-10
Improve docstring in states

Ericgig:
**Description**
As presented in #1722, the docstring of the states and operators functions are not consistent.

This change the docstring so that _Number of basis states in Hilbert space_ is used everywhere, instead of _fock states_ or _vector states_.

This set the variable names as:
- Single Hilbert space function : `N`
- Multiple Hilbert space function: `dimensions`
- Number of qubits: `N_qubit`

Added support for `dimensions` input to take `Space`  object.
Added sanity check in dimensions object.
Added `scalar_like` method to dimensions object. In transform any dims `[[2, 3], [2, 3]]` to it's scalar with same structure `[[1, 1], [1, 1]]`.
Added another input parameter to qzero to create rectangular operator `qzero([2],[3]).dims == [[2], [3]]`.

**Related issues or PRs**
fix #1184, replace #2075


qutip/core/dimensions.py
```diff
@@ -27,6 +28,8 @@ def flatten(l):
     [0, 1, 2]
 
     """
+    if isinstance(l, (Space, Dimensions)):
```
hodgestar, line 31: We should mention in the docstring what the argument `l` can be and what happens with them (Space and Dimension are converted to lists, scalars and other non-lists are converted to single element lists).

qutip/core/operators.py
```diff
@@ -639,25 +639,29 @@ def _implicit_tensor_dimensions(dimensions):
     dimensions : list
         Dimension list in the form required by ``Qobj`` creation.
     """
-    if not isinstance(dimensions, list):
+    if not isinstance(dimensions, (list, Space)):
```
hodgestar, line 642: If this can now accept a Space, should it not also accept a Dimensions object?

qutip/core/states.py
```diff
@@ -43,13 +44,29 @@ def _promote_to_zero_list(arg, length):
...
+    Returns
+    -------
+    space : :class:`.Space`
+    """
+    if isinstance(dimensions, Space):
+        return dimensions
+    elif isinstance(dimensions, list):
+        return Space(dimensions)
+    else:
+        return Space([dimensions])
```
hodgestar, line 60: What should happen if `dimensions` is already a Dimension object?

hodgestar (commented):
I left a couple of small questions, but otherwise this looks good.

We should document the methods of the dimension and space classes better (i.e. add docstrings which describe the arguments) but happy for that to happen sometime in a later PR.

qutip/core/operators.py
Ericgig: Dimensions object would change the input meaning causing more confusion:
`qeye([[2], [2]]) != qeye(Dimension([[2], [2]]))`
The first case create a super operator, while the second an operator...

For `qzero` it would be nice since rectangular operators are fine, but changing the input to no longer match `qeye` is probably not a good idea.

qutip/core/states.py
Ericgig: Create a super operator.

qutip/core/operators.py
```diff
@@ -670,10 +645,17 @@ def qzero(dimensions, *, dtype=None):
...
+    if dims_right is not None:
+        dims_right = Space(dims_right)
+        dims = [dims[0], dims_right]
+        size_right = dims_right.size
+    else:
+        size_right = size
     # A sparse matrix with no data is equal to a zero matrix.
-    type_ = 'super' if isinstance(dimensions[0][0], list) else 'oper'
-    return Qobj(_data.zeros[dtype](size, size), dims=dimensions,
+    return Qobj(_data.zeros[dtype](size, size_right), dims=dims,
```
hodgestar, line 658:
```suggestion
    return Qobj(_data.zeros[dtype](size_left, size_right), dims=dims,
```

hodgestar (approved):

Issue #2293, opened 2024-01-15
qutip 4.7.3 QobjEvo test failures on Python 3.12


drew-parsons:
### Bug Description

Building qutip 4.7.3 for Debian unstable, all tests are passing with python 3.11. But a dozen tests using qobjevo fail with python 3.12:
```
=========================== short test summary info ============================
FAILED qutip/tests/test_brmesolve_td.py::test_time_dependent_spline_in_c_ops
FAILED qutip/tests/test_brmesolve_td.py::test_split_operators_maintain_answer[_2_tuple_split]
FAILED qutip/tests/test_correlation.py::TestTimeDependence::test_varying_coefficient_hamiltonian_2ls[string]
FAILED qutip/tests/test_correlation.py::TestTimeDependence::test_varying_coefficient_hamiltonian_2ls[numpy]
FAILED qutip/tests/test_mesolve.py::TestMESolverConstDecay::testMEDecayAsStrList
FAILED qutip/tests/test_mesolve.py::TestMESolveTDDecay::testMETDDecayAsStrList
FAILED qutip/tests/test_mesolve.py::TestMESolveSuperInit::testMETDDecayAsStrList
FAILED qutip/tests/test_mesolve.py::TestMESolveStepFuncCoeff::test_array_str_coeff
FAILED qutip/tests/test_mesolve.py::TestMESolveStepFuncCoeff::test_array_str_py_coeff
FAILED qutip/tests/test_openmp.py::test_openmp_mesolve_td - Exception: Could ...
FAILED qutip/tests/test_propagator.py::testPropHOStrTd - Exception: Could not...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_call - Exception: Could not ...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_call_args - Exception: Could...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_mul_vec - Exception: Could n...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_mul_mat - Exception: Could n...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_expect_psi - Exception: Coul...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_expect_rho - Exception: Coul...
FAILED qutip/tests/test_qobjevo.py::test_QobjEvo_with_state - Exception: Coul...
[...]

PR #2294, opened 2024-01-15
Add 3.12 to action, but without cython support

Ericgig:
**Description**
Add python 3.12 to automated test and wheel build.
However python 3.12 removed the `imp` module used by `pyximport`. 
The fix on cython side may only be for cython 3.X, so we run without cython at runtime on 3.12.

Related to #2293


.github/workflows/tests.yml
```diff
@@ -123,6 +123,12 @@ jobs:
             python-version: "3.10"
             pytest-extra-options: "-W ignore::ImportWarning -k 'not (test_correlation or test_interpolate or test_mcsolve)'"
 
+          # Builds without Cython at runtime.  This is a core feature;
+          # everything should be able to run this.
```
hodgestar, line 127: We should update the comment to mention why we have an extra non-Cython run for 3.12.

qutip/__init__.py
```diff
@@ -54,24 +54,30 @@
...
             "Old version of Cython detected: needed {}, got {}."
             .format(_cy_require, _Cython.__version__)
         )
-    if _version2int(_Cython.__version__) >= _version2int(_cy_unsupported):
+    elif _version2int(_Cython.__version__) >= _version2int(_cy_unsupported):
         warnings.warn(
             "The new version of Cython, (>= 3.0.0) is not supported."
             .format(_Cython.__version__)
         )
+    elif _version2int(_Cython.__version__.split()[0]) >= _version2int("3.12.0"):
```
hodgestar, line 70: I'm not sure I understand this `elif` condition. We appear to be comparing a Python version to a Cython version?

qutip/tests/conftest.py
```diff
@@ -32,6 +32,10 @@ def _skip_cython_tests_if_unavailable(item):
         # importorskip does not have maxversion
         if _version2int(_Cython.__version__) >= _version2int("3.0.0"):
             pytest.skip("cython 3.0.0 not supported")
+            return
+        import qutip
+        if not qutip.settings.has_cython:
+            pytest.skip("Cython not available at runtime.")
```
hodgestar, line 38: I'm a bit uncomfortable relying on QuTiP's own detection magic to skip tests. If it ends up being wrong in some cases Cython tests will be silently skipped in CI.

Perhaps let's just check against the Python version and skip if it's Python 3.12?

qutip/tests/conftest.py
```diff
@@ -32,6 +33,10 @@ def _skip_cython_tests_if_unavailable(item):
         # importorskip does not have maxversion
         if _version2int(_Cython.__version__) >= _version2int("3.0.0"):
             pytest.skip("cython 3.0.0 not supported")
+            return
+        import qutip
```
hodgestar, line 37:
```suggestion
```
Doesn't look like we're using qutip here anymore.

hodgestar (approved):
Left one more comment, otherwise looks good.

hodgestar (approved):

Issue #2299, opened 2024-01-22
MemoryError when updating to latest qutip


Matt-Ord:
### Bug Description

Might just be me mis-using the new 5.X branch of Qutip. I get a memory error when upgrading to the new branch.

It appears to fail when performing self._to_dm(state) for each state. My states are vectors of length 800 - so it runs out of memory. Is it possible to request the solver only stores the wavefunction at each timestep? I want the wavefunction for each trajectory for each timestep. Alternatively one could introduce a new "proj" data storage to make this more efficient.

### Code to Reproduce the Bug

```shell
result = qutip.ssesolve(
        hamiltonian_qobj,
        initial_state_qobj,
        times.times,
        sc_ops=sc_ops,
        e_ops=[],
        options={
            "method": "euler",
            "progress_bar": "enhanced",
            "store_states": True,
            "keep_runs_results": True,
            "improved_sampling": False,
            "dt": times.delta_t / times.fundamental_n,
        },
        ntraj=n_trajectories,  # cspell:disable-line
    )
```

### Code Output

```shell
File "...", line 270, in solve_stochastic_schrodinger_equation
    result = qutip.ssesolve(
             ^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\qutip\solver\stochastic.py", line 461, in ssesolve
    return sol.run(
           ^^^^^^^^
  File "...\.venv\Lib\site-packages\qutip\solver\multitraj.py", line 198, in run
    map_func(
  File "...\.venv\Lib\site-packages\qutip\solver\parallel.py", line 122, in serial_map
    remaining_ntraj = reduce_func(result)
                      ^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\qutip\solver\result.py", line 645, in add   
    op(trajectory)
  File "...\.venv\Lib\site-packages\qutip\solver\result.py", line 519, in _reduce_states
    accu + self._to_dm(state)
    ~~~~~^~~~~~~~~~~~~~~~~~~~
  File "...\.venv\Lib\site-packages\qutip\core\qobj.py", line 134, in out
    return method(self, other)
           ^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\qutip\core\qobj.py", line 408, in __add__   
    return Qobj(_data.add(self._data, other._data),
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2304, opened 2024-01-23
Add error message to MemoryError

Ericgig:
**Description**
When we can't allocate enough memory for a matrix, there was no error message.
For sparse array, were the users expect the it not to take that much space, this can be confusing.
I added error message which tell the shape of the matrix and number of elements.

It won't help much, but may show that the matrix is not as sparse as expected.

**Related issues or PRs**
#2299


hodgestar (commented):
The code changes seem fine, but it looks like some tests are failing.

Could we also write some tests for these? It might be a little bit tricky in some cases, but hopefully we can at least cover the easier ones (e.g. directly creating big dense or big sparse data objects).

Ericgig:
The failing test are related to #2301 not being merged in yet.

I am not sure we want tests for this.
We have no control on the memory of the runner, which could change at any time.
There could be strange thing with virtual memory, swap, etc.

hodgestar:
Agreed that it is tricky to test. I mostly wanted the tests because the nice error messages might contain bugs. Happy to let this slide for now though.

hodgestar (approved):

PR #2308, opened 2024-01-25
Add the possibility to customize point colors as in V4

### LINKED ISSUE #1974: Bloch Sphere Colorbar not working in v5
### Bug Description

Adding points and colors to a `Bloch` object and rendering the `Bloch` object does not result in correctly colored points in QuTiP 5.
All points have the same colors. Running the same code in QuTiP 4 generates the correct output.

### Code to Reproduce the Bug

```shell
import matplotlib as mpl
import numpy as np
from matplotlib import cm
from qutip import Bloch, basis, sesolve, sigmax, sigmay, sigmaz

delta = 0.5 * 2 * np.pi
v = 2.0 * 2 * np.pi
H0 = delta / 2.0 * sigmax()
H1 = v / 2.0 * sigmaz()
H = [H0, [H1, "t"]]
psi0 = basis(2, 0)

tlist = np.linspace(-10.0, 10.0, 1500)
expt_list = sesolve(H, psi0, tlist, [sigmax(), sigmay(), sigmaz()]).expect

b = Bloch()
# normalize colors to times in tlist ##
nrm = mpl.colors.Normalize(-2, 10)
colors = cm.cool(nrm(tlist))

# add data points from expectation values ##
b.add_points([expt_list[0], expt_list[1], -expt_list[2]], "m")

# customize sphere properties ##
b.point_color = list(colors)
b.point_marker = ["o"]
b.point_size = [20]

b.zlpos = [1.1, -1.2]

b.show()
```

### Code Output

```shell
a plot (see below)
```

### Expected Behaviour

I expect the same coloring of points as with version 4. The expected plot is shown below.

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+c64f38e
Numpy Version:      1.23.0
Scipy Version:      1.8.1
Cython Version:     0.29.30
Matplotlib Version: 3.5.2
Python Version:     3.8.10
Number of CPUs:     4
BLAS Info:          OPENBLAS
INTEL MKL Ext:      False
```

### Additional Context

image

image

DnMGalan:
**Description**
Allow V5 to accept customization of point colors as it did in V4. To keep the changes implemented in V5, I have created a new variable to store the colors that are applied to the points (_inner_point_color_), so in total we now have three variables:

_inner_point_color_: Defined as an argument in the _add_points_ function, this variable takes precedence over the others.
_point_color_: This variable stores a list of colors, reminiscent of V4, and is utilized when _inner_point_color_ is not specified.
_point_default_color_: A list of colors that serves as the default when neither of the other variables is defined

**Related issues or PRs**
fix #1974

### DIFF
--- a/doc/changes/1974.bugfix
+++ b/doc/changes/1974.bugfix
@@ -1,0 +1,1 @@
+Add the possibility to customize point colors as in V4 and fix point plot behavior for 'l' style
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -794,7 +794,6 @@ class Bloch:
             dist = np.linalg.norm(points, axis=0)
             if not np.allclose(dist, dist[0], rtol=1e-12):
                 indperm = np.argsort(dist)
-                points = points[:, indperm]
             else:
                 indperm = np.arange(num_points)
 
@@ -817,9 +816,9 @@ class Bloch:
                 color = list(color)
 
             if self.point_style[k] in ['s', 'm']:
-                self.axes.scatter(np.real(points[1]),
-                                  -np.real(points[0]),
-                                  np.real(points[2]),
+                self.axes.scatter(np.real(points[1][indperm]),
+                                  -np.real(points[0][indperm]),
+                                  np.real(points[2][indperm]),
                                   s=s,
                                   marker=marker,
                                   color=color,
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -169,8 +169,10 @@ class Bloch:
         # ---point options---
         # List of colors for Bloch point markers, default = ['b','g','r','y']
         self.point_default_color = ['b', 'r', 'g', '#CC6600']
+        # Old variable used in V4 to customise the color of the points
+        self.point_color = None
         # List that stores the display colors for each set of points
-        self.point_color = []
+        self.inner_point_color = []
         # Size of point markers, default = 25
         self.point_size = [25, 32, 35, 45]
         # Shape of point markers, default = ['o','^','d','s']
@@ -360,7 +362,7 @@ class Bloch:
         self.point_style.append(meth)
         self.points.append(points)
         self.point_alpha.append(alpha)
-        self.point_color.append(colors)
+        self.inner_point_color.append(colors)
 
     def add_states(self, state, kind='vector', colors=None, alpha=1.0):
         """Add a state vector Qobj to Bloch sphere.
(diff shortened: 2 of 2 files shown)


Ericgig:
@DnMGalan 
Thank you for contributing.

Please write a towncrier entry. It's not about how small or large the issue, but a reminder for us to thank you for your contribution when doing the release.

I don't think any one opened an issue about the problem with the 'l' style, but we would be grateful if you fixed it. 
You don't need to open a issue for it. You can solve it here or in another PR at your convenience.

DnMGalan:
Hi @Ericgig, thank you for your feedback. I have already implemented the changes.

DnMGalan:
Sure! Here are the plots before the changes using the code posted in issue #1974. The bug we wanted to fix is that when choosing style 'm' and passing an array of colors as we did in V4, the scrip only plots one color, (additionally, i also plot one case with style 'l'):

M_self_test_old L_self_old

And these are the same plots after the changes using the variables `inner_point_color` (left graph) and `point_color` (right graph). As we can see for both variables we get the same result which is what we expected:

 

 

For before and after changes the plots using the default colors are the same (these are styles 'l', 's' and 'm' in that order):

  

Making more tests I found the error with style 'l' that I mentioned in issue #1974, which consisted in an erratic plotting after sorting the points. Here we can see some results of those tests before the changes (style 'l' using the `point_color` and `point_default_color` variables, plus 'm' with the `point_color` variable):

  

And as we can see after the changes, the 'l' style plotting is as expected and the 'm' style plotting has all the colors we defined:

Ericgig (commented):
Thank you, this fix the issue and the `l` points bug.

I see 2 points of improvements:
Rename `inner_point_color` to `_inner_point_color`, keeping it private so it's only set by `add_points`.

There are no check of `point_color` type. This could result in strange effects.
`b.point_color = "black"` will result in blue line "l" points  (`"black"[0] == "b"), but black "s" points.

With
```
b.add_points(_2_points, "s")
b.point_color = ["blue", "red"]
```
`scatter` will plot the "s" points with different colors.

```
b.add_points(_5_points, "m")
b.point_color = ["blue", "red"]
```
Will be an error.

Not too sure how it should be fixed.
We may just want to push it to a new issue or add to #1913.

DnMGalan:
- I agree that renaming the _inner_point_color variable is a good idea.

- The case you mention of black point_color with style 'l' also occurs in V4 and is solved by defining the variable as a list, e.g., ['black'] (which also works with the changes I proposed). So basically I assumed it was an expected use case.

- I asked this question about style 's' plotting two colors in issue #1974, since we are given the same colors for style 's' and 'm', and Neil suggested leaving it as is. The docstring in the add_points function already tells you to give a single color for style 's' and a list of colors for style 'm'.

- The error you get using style 'm' with two colors is that you need to specify a color for each point (you have to pass to the function a list of colors of the same length as the length of the points).

In view of this, I don't know which approach you prefer to apply, as it depends on the expected use case. Given that the code works, perhaps the priority would be to specify the expected use while defining as part of issue #1913 the modifications you consider appropriate (although maybe someone will catch it earlier if you define it as a new issue).

DnMGalan:
Perfect! Thank you for the review. I have uploaded the last change.

Ericgig (approved):

PR #2309, opened 2024-01-26
Update test matrix to include python 3.12

Ericgig:
**Description**
Updated the test matrix, removing test using older version of python, scipy and numpy to add new versions.
Matrix include all python versions from 3.9 to 3.12, all scipy from 1.8 to 1.12, all numpy from 1.22 to 1.26.
I keep test with no cython, cython 0.X and 3.X. With and without mkl and all 3 os.

Added python 3.12 to the list of build wheels, but removed 3.8.

Increase minimum version of scipy and numpy in requirement to match those we actually tests.


hodgestar (approved):
Thanks @Ericgig.

PR #2318, opened 2024-02-06
Add `run_from_experiment` to stochastic solvers.

Ericgig:
**Description**
Add the capacity of stochastic solver to run from know noise or measurements.
In the lab, the measurement can be obtained, but that measurements is a mix of and expectation values and noise. The stochastic solver can be used to split these components to study the system.

The new method for this is `SMESolver.run_from_experiment`. It works similarly to `run` of deterministic solver: it compute a single trajectory. It can take either `dW` or `measurement` as the noise input. (`dW` is the default). `dW` input is available with every integration method, but measurement input is limited to `euler`, `milstein`, `pred_corr`, `platen`.

The measurement is `expect(m_ops, state_t) + dW / dt`.  For `run_from_experiment` it has no choice but to take `state_t` at the start of the interval, but for the result for `smesovle`, it was taken at the end of the interval. I updated the `store_measurement` options to allow the user the choose where it is taken between `start`, `end` and `middle`. When `start` is used, the `run`'s result's measurement can be feed back to `run_from_experiment` to recompute the same trajectory. (The theory is the integral of the expectation value over the interval.)

Another change in this PR is to keep the dW in the wiener and result class. We used to sum them into `W`, but the `dW` are what we use so it just end up being extra calls to `np.cumsum` and `np.diff`.


qutip/solver/stochastic.py
```diff
@@ -626,6 +648,85 @@ def _integrate_one_traj(self, seed, tlist, result):
...
+
+        e_ops : list, optional
+            List of operators for which to evaluate expectation values.
+
+        measurement : bool, default : False
+            Whether the passed noise is the Wiener increments ``dW`` (gaussian
+            noise with standard derivation of dt**0.5), or the measurement::
+
+                noise = dW/dt * dW_factors
+                    + expect(sc_ops[i] + sc_ops[i].dag, state_t)
```
pmenczel, line 686: This is not quite right for the case of heterodyne measurement, right? It could be good to also explain that case explicitly.

qutip/solver/stochastic.py
```diff
@@ -626,6 +648,85 @@ def _integrate_one_traj(self, seed, tlist, result):
...
+
+            Note that the expectation value is usally computed at the start of
+            the step. Only available for limited integration methods.
```
pmenczel, line 689: I found this sentence a bit unclear. Why "usually"?

```suggestion
            Note that this function expects the expectation values to be taken at the start of
            the time step, corresponding to the "start" setting for the "store_measurements" option.
            Only available for limited integration methods.
```

qutip/solver/stochastic.py
```diff
@@ -626,6 +648,85 @@ def _integrate_one_traj(self, seed, tlist, result):
...
+
+        Returns
+        -------
+        result : StochasticTrajResult
+            Result of the trajectory.
+
+        Notes
+        -----
+        Only default values of `m_ops` and `dW_factors` are supported.
```
pmenczel, line 698: Should we warn if `m_ops` or `dW_factors` were changed manually?

qutip/tests/solver/test_stochastic.py
```diff
@@ -380,3 +414,91 @@ def test_small_step_warnings(method):
...
+
+
+@pytest.mark.parametrize("method", ["euler", "platen"])
+@pytest.mark.parametrize("heterodyne", [True, False])
+def test_run_from_experiment_close(method, heterodyne):
+    N = 10
+
+    H = num(N)
+    a = destroy(N)
+    sc_ops = [a, a.dag() * 0.1]
```
pmenczel, line 426: Don't these `sc_ops` lead to basically the same `m_ops`?

qutip/solver/sode/_noise.py
```diff
@@ -10,31 +10,84 @@ class Wiener:
...
+        self.noise = np.concatenate((self.noise, dW), axis=0)
 
     def dW(self, t, N):
-        if t + N * self.dt > self.t_end:
-            self._extend(t + N * self.dt)
-        idx0 = int((t - self.t0 + self.dt * 0.01) // self.dt)
-        return np.diff(self.process[idx0:idx0 + N + 1, :, :], axis=0)
+        # Find the index of t.
+        # Rounded to the closest step, but only multiple of dt are expected.
+        idx0 = int((t - self.t0 + self.dt * 0.4999) // self.dt)
```
pmenczel, line 29: Is there a reason not to use `round()`? Is this faster?

pmenczel (commented):
This is a cool new feature! I do not think that I have much to add.

Just as a side note, I used this opportunity to finally try to understand what exactly happens with the `heterodyne` option, and it took me quite a while to figure out. In the users guide, there is the definition of the homodyne current; I think it could be useful to add a definition of the heterodyne currents as well? (But I don't know if that's in the scope of this PR.)

qutip/solver/stochastic.py
pmenczel: Thank you for updating. (Looking at it again, I think in the previous text it only should have been `m_ops` instead of `sc_ops`. But anyway, it is very clear now.)

qutip/solver/stochastic.py
pmenczel: Ah, I was thinking about monitoring the setters. Okay, then it's fine.

qutip/tests/solver/test_stochastic.py
pmenczel: I see. Would it make sense to add one more version of the test including some other `sc_op`, for example a Hermitian one?

qutip/solver/sode/_noise.py
Ericgig: I always used numpy functions and never realized there was a built in...

pmenczel (approved):

Issue #2317, opened 2024-02-06
ReadTheDocs QuTiP build fails after adding MPI as a dependency.


hodgestar:
### Bug Description

The ReadTheDocs build fails with:
```
gcc -pthread -B /home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/compiler_compat -Wno-unused-result -Wsign-compare -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/include -fPIC -O2 -isystem /home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/include -fPIC -I/home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/include/python3.10 -c _configtest.c -o _configtest.o
      _configtest.c:2:10: fatal error: mpi.h: No such file or directory
          2 | #include <mpi.h>
            |          ^~~~~~~
```
See https://readthedocs.org/projects/qutip/builds/23361676/ for the full build log (not sure if this is publicly visible -- if not, I can add QuTiP admins as maintainers on readthedocs.org if they give me their username).

The QuTiP ReadTheDocs configuration is at: https://github.com/qutip/qutip/blob/master/.readthedocs.yaml.

Probably this can be fixed by adding an appropriate dependency to https://github.com/qutip/qutip/blob/master/doc/rtd-environment.yml.

### Code to Reproduce the Bug

```shell
Attempt to build the documentation on ReadTheDocs.
```

### Code Output

```shell
gcc -pthread -B /home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/compiler_compat -Wno-unused-result -Wsign-compare -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/include -fPIC -O2 -isystem /home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/include -fPIC -I/home/docs/checkouts/readthedocs.org/user_builds/qutip/conda/master/include/python3.10 -c _configtest.c -o _configtest.o
      _configtest.c:2:10: fatal error: mpi.h: No such file or directory
          2 | #include <mpi.h>
            |          ^~~~~~~
```

### Expected Behaviour

ReadTheDocs builds successfully.

### Your Environment

```shell
QuTiP version: master branch
```

[...]

PR #2319, opened 2024-02-06
Remove mpi from full and update versions

Ericgig:
**Description**
Removed `mpi4py` from `full`. 
It is hard to install locally (for the users) and in readthedocs, where it is not actually used (#2317).

I am not sure what it is the best practice for this. `full` should normally include every optional dependency, but mpi4py is too specialized and not useful for most users. If it makes qutip harder to install since `pip install qutip[full]` would fail for most, maybe it's better to leave it out.

Also updated some versions.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -10,9 +10,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: mpi4py/setup-mpi@v1
-        with:
-          mpi: 'openmpi'
 
       - uses: actions/setup-python@v4
         name: Install Python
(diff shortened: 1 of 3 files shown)


hodgestar (approved):

hodgestar:
As long as we do some test runs without MPI, I think this is fine. Perhaps we could also document somewhere that `mpi` is needed -- maybe just by raising a nice exception if it needs to be imported but isn't available?

Ericgig:
We have test cases both with and without `mpi4py`.
I believe trying to use `mpi_pmap` without `mpi4py`, it will raise the default `ImportError`. 
@pmenczel can confirm.

pmenczel (approved):
Yes, calling `mpi_pmap` without mpi4py installed will raise an ImportError. The tests involving mpi will be skipped automatically, and we install mpi4py in only one of the test workflows.

I was not aware that the build documentation workflow and the actual readthedocs build are two separate things. This fix makes sense to me, since mpi4py is a very specialized dependency. And we definitely do not need to install mpi to build the documentation.

I have removed the mpi installation also from the build documentation workflow.

PR #2314, opened 2024-02-02
Add coo format

Matt-Ord:
**Checklist**
Thank you for contributing to QuTiP! Please make sure you have finished the following tasks before opening the PR.

- [x] Please read [Contributing to QuTiP Development](http://qutip.org/docs/latest/development/contributing.html)
- [x] Contributions to qutip should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/).
You can use [pycodestyle](http://pycodestyle.pycqa.org/en/latest/index.html) to check your code automatically
- [x] Please add tests to cover your changes if applicable.
- [x] If the behavior of the code has changed or new feature has been added, please also update the documentation in the `doc` folder, and the [notebook](https://github.com/qutip/qutip-tutorials). Feel free to ask if you are not sure.
- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#changelog-generation) for more information).

Delete this checklist after you have completed all the tasks. If you have not finished them all, you can also open a [Draft Pull Request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to let the others know this on-going work and keep this checklist in the PR description.

**Description**
Add the sparse COO format.

**Related issues or PRs**
Please mention the related issues or PRs here. If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id, e.g. fix #1184

PR #2321, opened 2024-02-08
Ensure dtype in Qobj creation functions.

Ericgig:
**Description**
In #2314, it was found that Qobj creation functions could return an unexpected dtype for incomplete data layers.
For example if `mul` specialization is missing `scale * basis(..., dtype)` would not return a Qobj using dtype.
I reviewed functions in states.py, operators.py and random_object.py.


hodgestar (approved):
We should also document the default `dtypes`, but this looks like a good improvement.

PR #2325, opened 2024-02-14
Add dtype to QobjEvo and use it in `qeye_like`

Ericgig:
**Description**
`qzero_like` used on `QobjEvo` would call it (it needed `qevo(0).data`), which is both slow and could break `jax.jit`.
I added `Qobj.dtype` and `QobjEvo.dtype` to be used instead of accessing the data directly.
For `QobjEvo`, the `dtype` property still call the object when the type is unclear (mixed or function based), but it should be faster otherwise.
Tested that the example in qutip/qutip-jax#33 works.

### DIFF
--- a/qutip/core/operators.py
+++ b/qutip/core/operators.py
@@ -766,7 +766,9 @@ def qeye_like(qobj):
 
     """
     if qobj.shape[0] != qobj.shape[1]:
-        raise ValueError(f"Cannot build a {qobj.shape} identity matrix.")
+        raise ValueError(
+            "Can't create an identity matrix like a non square matrix."
+        )
     return Qobj(
         _data.identity[qobj.dtype](qobj.shape[0]), dims=qobj._dims,
         isherm=True, isunitary=False, copy=False
--- a/qutip/core/operators.py
+++ b/qutip/core/operators.py
@@ -771,7 +771,7 @@ def qeye_like(qobj):
         )
     return Qobj(
         _data.identity[qobj.dtype](qobj.shape[0]), dims=qobj._dims,
-        isherm=True, isunitary=False, copy=False
+        isherm=True, isunitary=True, copy=False
     )
 
 
--- a/qutip/core/operators.py
+++ b/qutip/core/operators.py
@@ -691,11 +691,9 @@ def qzero_like(qobj):
         Zero operator Qobj.
 
     """
-    from .cy.qobjevo import QobjEvo
-    if isinstance(qobj, QobjEvo):
-        qobj = qobj(0)
+
     return Qobj(
-        _data.zeros_like(qobj.data), dims=qobj._dims,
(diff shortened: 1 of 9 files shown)


hodgestar:
While reading https://github.com/qutip/qutip-jax/issues/33 I was wondering whether we should allowed mixed-dtype QobjEvos at all. I'm struggling to think of a case where it is what a user would want and can imagine lots of situations where it would not be what anyone wanted.

Ericgig:
It would be hard to enforce since we accept functions returning `Qobj`.
If someone passed `dtype` in `args`, there is nothing we can do.
Also since the most common use case is `part @ state`, they don't interact much together and each part can be optimized for it's sparsity without issue.

Mixed case are not that rare. Per default `sigmax` is `CSR` and `qeye` is `Dia` and I can't say what is `sigmax & qeye` without checking. So `QobjEvo([sigmax(), [qeye(2), f]])` mixes `CSR` and `Dia`.
How do we decide which one we quietly transform? Everything else just work with whatever type is used. So we can't have this raise an exception.

Personally I think we should only mix in a family of dtypes, `Dense`, `Dia` and `CSR` can be mixed, but no mixing of jax with CSR, etc. But we don't have what we need to enforce it.

qutip/core/operators.py
```diff
@@ -767,12 +765,13 @@ def qeye_like(qobj):
         Identity operator Qobj.
 
     """
-    from .cy.qobjevo import QobjEvo
-    if isinstance(qobj, QobjEvo):
-        qobj = qobj(0)
+    if qobj.shape[0] != qobj.shape[1]:
+        raise ValueError(
+            "Can't create an identity matrix like a non square matrix."
+        )
```
BoxiLi, line 771: Hasn't the error already been raised in the data layer? Such as in `identity_like_dense`

qutip/core/operators.py
```diff
@@ -767,12 +765,13 @@ def qeye_like(qobj):
...
     return Qobj(
-        _data.identity_like(qobj.data), dims=qobj._dims,
-        isherm=True, isunitary=True, copy=False
+        _data.identity[qobj.dtype](qobj.shape[0]), dims=qobj._dims,
+        isherm=True, isunitary=False, copy=False
```
BoxiLi, line 774: Why `isunitary=False`?

qutip/core/operators.py
Ericgig: It no longer use the the data layer `identity_like`.

qutip/core/operators.py
Ericgig: Cut and paste error...
Fixing.
```suggestion
        isherm=True, isunitary=True, copy=False
```

BoxiLi (approved):

PR #2327, opened 2024-02-15
Add type hints

Ericgig:
**Description**
Added hints to mesolve, sesolve, result, Qobj, QobjEvo.

The aim is not to pass static type checker, but to help the user with hints: 
- I added types to outside facing functions.
- I sometime used hints that do not match the real implementation. For example:
  -  `Qobj.__mul__` is typed to support product with scalar, not other `Qobj`, (for which `__matmul__` should be used.)
  -  `e_ops` are hinted to dict but also support single operator, list, tuple...
  - `solver.options` claim it return a dict not `_SolverOptions` (which is a subclass of dict.)
- I skipped the `None` type. (`result.final_state` type is `Qobj` event if it is `None` when not saved.)

I created alias types:
- QobjEvoLike: Stuff that can be made into a QobjEvo.
- CoefficientLike: Stuff that can be made into a Coefficient.
- LayerType: Can be understood as our data layer type.  (A dynamically generated Literal could be better.)
Maybe some alias for dimensions object could be useful?

Knowing that we have extension modules, I used numbers.Number for complex hoping it support jax or tensorflow's scalars. Also `Any` used in some place where other type support can be added such as `Qobj.__init__`.

Some of the type hint features I used were only added in python 3.10. We still tests with python 3.9, but it's old enough that we can remove it (according to [spec-0000](https://scientific-python.org/specs/spec-0000/)).


doc/development/contributing.rst
```diff
@@ -113,6 +113,20 @@ This includes using the same variable names, especially if they are function arg
...
+Type Hints
+----------
+
+Adding type hints to users facing functions is recommended.
+QuTiP's approach is such:
+
+- Type hints are *hints* for the users.
+- Type hints can show the preferred usage over real implementation, for example:
+  - ``Qobj.__mul__`` is typed to support product with scalar, not other ``Qobj``, for which ``__matmul__`` should is preferred.
+  - ``solver.options`` claim it return a dict not ``_SolverOptions`` (which is a subclass of dict).
```
hodgestar, line 125:
```suggestion
  - ``solver.options`` claims it return a dict not ``_SolverOptions`` (which is a subclass of dict).
```

hodgestar (approved):

PR #2330, opened 2024-02-21
Qobj minor fixes

pmenczel:
Minor cleanup of `Qobj` (docstrings, removed unused variable, fixed typo in setter of `data`).

Docstring fixes include: removing `type` parameter from constructor, documenting possible types of `inpt` parameter, adding a handful missing functions to the list of functions.

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -147,12 +147,10 @@ class Qobj:
 
     Parameters
     ----------
-    inpt: array_like
+    inpt: array_like, data object or :obj:`.Qobj`
         Data for vector/matrix representation of the quantum object.
     dims: list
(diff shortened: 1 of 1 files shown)


Ericgig (approved):
Thank you Paul.

qutip/core/qobj.py
```diff
@@ -162,8 +160,12 @@ class Qobj:
 
     Attributes
     ----------
-    data : array_like
-        Sparse matrix characterizing the quantum object.
+    data : object
```
hodgestar, line 163: `data` should be an instance of `qutip.core.data.Data` and not an arbitrary object (minor nitpick, but might help users understand what this is).

PR #2333, opened 2024-02-24
Fix heom steady state

nwlambert:
**Description**
Two small bugs crept in in the conversion of the steady_state solver in the HEOM method to QuTiP v5.   Firstly, the system state was returned without Fortran ordering (so it was effectively returned transposed), and the line which was supposed to enforce hermiticity used a conj() instead of an adjoint() so it just deleted the imaginary parts of the off-diagonals instead.  

I think none of the notebooks or tests were sensitive to these two compounding issues, but it is fairly easy to make an example which is. This is also added as a rudimentary test, comparing output of the long-time dynamics to the steadystate. I didn't include any parameters to change in that test, but can do so if its useful.

As a comment, the steady-state solver could probably be written to be more like the standard steadystate solver. But maybe better that is done alongside an effort to make the heom solver more data-layer flexible.

### DIFF
--- a/doc/changes/2333.bugfix
+++ b/doc/changes/2333.bugfix
@@ -1,0 +1,1 @@
+Fixed two problems with the steady_state() solver in the HEOM method.
--- a/qutip/solver/heom/bofin_solvers.py
+++ b/qutip/solver/heom/bofin_solvers.py
@@ -965,9 +965,10 @@ class HEOMSolver(Solver):
         else:
             L = L.tocsc()
             solution = spsolve(L, b_mat)
-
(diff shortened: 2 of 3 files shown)


hodgestar (approved):
The changes look good to me. I hope these weren't too difficult to find!

hodgestar:
The CI errors all appear to the result of the coveralls.io service being down. Happy for this to be merged with the failures.

Ericgig:
Is this ready to merge?

nwlambert:
Yes, I think so, thanks!

PR #2336, opened 2024-02-27
Fix `bloch.clear` breaking `point_color`

Ericgig:
**Description**
In #2308, the default for `point_color` was changed from `[]` to `None`, but `clear` still set it to an empty list.
This broke a notebook in qutip-tutorial: tutorials-v5/visualization/bloch-sphere-animation.


Ericgig:
Would also work.

hodgestar (approved):

PR #2341, opened 2024-02-29
Making ENR states work with mesolve()

nwlambert:
**Description**

With the inclusion of the new dimensions class, stored in Qobj()'s _dims property, the properties of ENR states are described by ENRspace().  However, when using ENR states with mesolve, Liouvillian, sum() and other functions these _dim properties would be discarded.  This PR is a draft attempt to get things working, largely by subbing in _dims instead of dims in various places, and a test to check ENR+mesolve() works for a simple case.

All tests pass, but since this is just some bodging to get things working, and messes a little bit with some core functions maybe its also good to check the tutorial notebooks also pass.

edit: forgot to mention, this is not exhaustive; e.g., sprepost() has not been updated.

### DIFF
--- a/qutip/core/superoperator.py
+++ b/qutip/core/superoperator.py
@@ -95,21 +95,18 @@ def liouvillian(H=None, c_ops=None, data_only=False, chi=None):
         L += sum(lindblad_dissipator(c_op, chi=chi_)
                  for c_op, chi_ in zip(c_ops, chi))
         return L
-
-    data = _data.mul(_data.kron(_data.identity_like(H.data), H.data), 
-                     -1j)
-    data = _data.add(data, _data.kron_transpose(H.data,
-                     _data.identity_like(H.data)), scale=1j)
+    spI = _data.identity_like(H.data)
+    data = _data.mul(_data.kron(spI, H.data), -1j)
+    data = _data.add(data, _data.kron_transpose(H.data, spI),
+                     scale=1j)
 
     for c_op, chi_ in zip(c_ops, chi):
         c = c_op.data
         cd = c.adjoint()
         cdc = _data.matmul(cd, c)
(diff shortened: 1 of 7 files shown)


qutip/core/superoperator.py
```diff
@@ -96,28 +96,26 @@ def liouvillian(H=None, c_ops=None, data_only=False, chi=None):
...
-    op_shape = H.shape
-    sop_dims = [[op_dims[0], op_dims[0]], [op_dims[1], op_dims[1]]]
-    sop_shape = [np.prod(op_dims), np.prod(op_dims)]
-    spI = _data.identity(op_shape[0], dtype=type(H.data))
-
-    data = _data.mul(_data.kron(spI, H.data), -1j)
-    data = _data.add(data, _data.kron_transpose(H.data, spI), scale=1j)
+    data = _data.mul(_data.kron(_data.identity_like(H.data), H.data), 
+                     -1j)
+    data = _data.add(data, _data.kron_transpose(H.data,
```
Ericgig, line 101: I would keep the `spI` definition instead of calling `identity_like` multiple times.

qutip/solver/solver_base.py
```diff
@@ -95,7 +95,7 @@ def _restore_state(self, data, *, copy=True):
         """
         Retore the Qobj state from its data.
         """
-        if self._state_metadata['dims'] == self.rhs.dims[1]:
+        if self._state_metadata['dims'].as_list() == self.rhs._dims.as_list()[1]:
```
Ericgig, line 98: Since both are Dimension object, the comparison should be able to work without the `as_list`. Did you try it?

nwlambert:
probably there are more edge case and functions that need an update (sprepost() etc), but maybe for now this is a sufficient fix?

Ericgig (approved):

Issue #2265, opened 2023-11-20
measurement_statistics_observable() returns negative probailities


adazi:
https://github.com/qutip/qutip/blob/4ce8829edf00cbcf8e60b86b6bad60d9621a64f3/qutip/measurement.py#L168

Using qutip 4.7.2, I have found that using 
```python
[eigenvalues, states, probabilities]= measurement_statistics_observable(state,operator)
```
to get the possible outcomes of a measurement and the probabilities for each is *extremely expensive* and also sometime returns (invalid/unphysical) negative numbers for some probabilities. This seems to happen with larger systems (8 qubits and above). The negative probabilities are of order $10^{-12}$, so most likely this is a numerical precision issue. 

P.S. This is my first time reporting and issue for this code, so I apologize if I'm doing so in the wrong place or if I have left out information.

hodgestar:
Welcome, and thanks for filing your first issue.

`measurement_statistics_observable` determines the eigenvalues and eigenstates of the supplied operator, so it's likely to be slow as the size of the system grows. If there is a means to speed it up, we'd be happy to give that a try. You could also try the alpha release of QuTiP 5 (`pip install qutip==5.0.0a2`) and compare performance there.

Are you only seeing the negative probabilities when `state` is a density matrix? Perhaps we can make the code a bit better and explicitly tidy up the small unphysical negative probabilities.

adazi:
The negative values problem is definitely present when `state` is a density matrix, and right now that is the only case I have tested. 

I understand it will be very slow for larger systems (exponentially so), but in my application is seems to be much slower even than the time evolution of the function, which seems like it should be more computationally complex. Given that, I suspect there is some room for optimization. 

What's the best way to tell if there are any relevant differences between 4.7.2 and the 5.0.x alpha?

hodgestar:
QuTiP 5 changed the underlying data representation of Qobjs completely.

The time evolution is potentially much less complex than the eigenstate problem. Consider that if one has the eigenstates and eigenvalues, one has essentially solved all possible time evolution problems for that operator. Again, this is not to say we couldn't do something better here.

adazi:
Thanks for clearing that up. I had tried to poke around to see what had changed, but I hadn't gone deep enough to see what changed with the qobjs themselves. 

[...]

Issue #2054, opened 2023-01-04
Measurement.measure produces wrong output density matrix in two-qubit system


RMaier1:
### Bug Description

It seems like the measurement.measure function does not produce the correct density matrix after the measurement of a two-qubit system. In the minimal example below I prepared the density matrix $\rho$ of two qubits, both in $|X\rangle$ eigenstate:
$\rho_i = |X\rangle \langle X| \otimes |X\rangle \langle X|$

If I measure the projection along the $X$-axis of the first qubit with the operator 
$\hat{O} = \sigma_x \otimes \mathbb{I}$
the measurement function produces the correct eigenvalue (which is 1), but not the correct output density matrix $\rho_e$ (which should be equal to the input density matrix $\rho_i$, since the first qubit is already in the eigenstate of $\hat{O}$)

### Code to Reproduce the Bug

```shell
import qutip

rho_1 = qutip.Qobj([[1,1],[1,1]]).unit()    # Initialize two qubits, both in |+X>-state
rho_tot = qutip.tensor(rho_1, rho_1)        # Density matrix of combined system

op = qutip.tensor(qutip.sigmax(), qutip.qeye(2))    # Measurement operator acting on first Qubit, measuring projection in x-direction

result = qutip.measurement.measure(rho_tot, op) 
print(result)
```

### Code Output

```shell
50 % of the time:

(1.0, Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[0.  0.  0.  0. ]
 [0.  0.5 0.  0.5]
 [0.  0.  0.  0. ]
 [0.  0.5 0.  0.5]])


Other 50 % of the time:

(1.0, Quantum object: dims = [[2, 2], [2, 2]], shape = (4, 4), type = oper, isherm = True
Qobj data =
[[0.5 0.  0.5 0. ]
 [0.  0.  0.  0. ]
 [0.5 0.  0.5 0. ]
 [0.  0.  0.  0. ]])
```

### Expected Behaviour

The correct result of the output matrix should be $rho_i$, as the input matrix is already an eigenstate of $\hat{O}$.
The output of the measurement shoud be:
$\rho_e = |X\rangle \langle X| \otimes |X\rangle \langle X| =$ [[0.25  0.25  0.25  0.25], [0.25  0.25  0.25  0.25], [0.25  0.25  0.25  0.25], [0.25  0.25  0.25  0.25]]

Instead the measurement output is given by:
[...]

Issue #2029, opened 2022-12-03
Different behavior of measurement for `qutip-4.7` and `dev.major`


BoxiLi:
### Bug Description

It seems that the measurement in `dev.major` is not truncating the probabilities correctly. It is probably caused by some changes in the last month and leads to [a test failing in `qutip-qip`](https://github.com/qutip/qutip-qip/actions/runs/3610116105/jobs/6083718885).

I don't see any code in the measurement that manually tidies up the state. In `qutip-4.7` it is also directly compared to `0`.
https://github.com/qutip/qutip/blob/5a47ee3b8dd817d4c6c36cc84a952c41305cdcb5/qutip/measurement.py#L101-L108

So my guess is that, in `dev.major`, the final states failed to be automatically truncated.

### Code to Reproduce the Bug

```shell
([Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket'
Qobj data =
[[1.]
 [0.]], Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket'
Qobj data =
[[0.]
 [1.]]], [0.999999999998, 1e-24])
```

### Code Output

```shell
dev.major
-------

([Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket'
Qobj data =
[[1.]
 [0.]], Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket'
Qobj data =
[[0.]
 [1.]]], [0.999999999998, 1e-24])
```

qutip4.7
-------
```
([Quantum object: dims = [[2], [1]], shape = (2, 1), type = ket
Qobj data =
[[1.]
 [0.]], None], [0.999999999998, 0.0])
```
```


### Expected Behaviour

They are expected to behave the same.

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+2ef8847
Numpy Version:      1.23.5
Scipy Version:      1.9.3
Cython Version:     0.29.32
Matplotlib Version: 3.6.2
Python Version:     3.10.0
Number of CPUs:     12
BLAS Info:          OPENBLAS
INTEL MKL Ext:      False
Platform Info:      Windows (AMD64)

And

QuTiP Version:      4.7.0
Numpy Version:      1.22.4
Scipy Version:      1.7.3
Cython Version:     0.29.30
Matplotlib Version: 3.5.2
Python Version:     3.9.0
Number of CPUs:     12
BLAS Info:          OPENBLAS
OPENMP Installed:   False
[...]

PR #2342, opened 2024-02-29
Makes measurement support degenerate states

### LINKED ISSUE #1800: Support measurments of observables with degenerate eigenvalues
Currently the `qutip.measurement.measure_observable` function always uses rank-1 projectors to determine the final state after the measurement. But for obervables with degenerate eigenvalues you might just want to project the state onto a higher-dimensional eigenspace. Eg. measuring the $Z \otimes Z$ operator on the Bell state $\ket{\psi} = 1/\sqrt{2} (\ket{00} + \ket{11})$ using qutip will return either $\ket{00}$ or $\ket{11}$ despite the state being an eigenstate of the observable.
```
import qutip as qt
import qutip.measurement as meas
import numpy as np

#Prepare Bell state
state = qt.tensor([qt.basis(2, 0)]*2) + qt.tensor([qt.basis(2, 1)]*2)
state = state.unit()

# ZZ operator
op = qt.tensor([qt.sigmaz()]*2)

result_eigenvalue, final_state = meas.measure(state, op)

#Check that the measurement didn't change the state (because it is an eigenstate of the measurement operator)
assert np.isclose(qt.fidelity(final_state, state), 1) #fails
```

I am aware that I can get my intended behaviour by explicity passing the list of projection operators to the `measure` function, but the following enhancement would be very useful:
An option `respect_degenerate_eigenvalues=False` in `qutip.measurement.measure_observable` that groups the eigenstate projectors based on eigenvalue degeneracy. In the above example the rank-2 projectors |00><00| + |11><11| and |01><01| + |10><10| would be used and the Bell state would be preseved during a ZZ measurement.

Ericgig:
**Description**
Measurement functions did not take into account degenerate states of the operator. (#1800, #2054)
Also the returned probabilities could be negative. (#2265)

I added a tolerance to these function for the minimal probability. Smaller / negative probabilities are rounded to `0`.
I also changed `measurement_statistics_observable` to support degenerate operators: It always return the projectors, with projectors and probabilities of degenerate sub-systems summed. It also only return the (eigenvalues, projector, probability) set for cases where the probability is greater than `tol`.

**Related issues or PRs**
fix #1800, #2054, #2029, #2265

### DIFF
--- a/qutip/measurement.py
+++ b/qutip/measurement.py
@@ -13,7 +13,7 @@ __all__ = [
 
 import numpy as np
 
-from . import Qobj, expect, identity, tensor
+from . import Qobj, expect, identity, tensor, settings
 
 
 def _verify_input(op, state):
@@ -36,7 +36,7 @@ def _verify_input(op, state):
         raise ValueError("state must be a ket or a density matrix")
 
 
-def _measurement_statistics_povm_ket(state, ops):
+def _measurement_statistics_povm_ket(state, ops, tol=None):
     r"""
     Returns measurement statistics (resultant states and probabilities)
     for a measurements specified by a set of positive operator valued
@@ -51,6 +51,9 @@ def _measurement_statistics_povm_ket(state, ops):
       List of measurement operators :math:`M_i` (specifying a POVM such that
       :math:`E_i = M_i^\dagger M_i`).
 
+    tol : float, optional
+        Smallest value for the probabilities. Smaller probabilities will be
+        rounded to ``0``.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


qutip/measurement.py
```diff
@@ -178,33 +197,57 @@ def measurement_statistics_observable(state, op):
...
+
     Returns
     -------
     eigenvalues: list of float
         The list of eigenvalues of the measurement operator.
 
-    eigenstates_or_projectors: list of :class:`.Qobj`
-        If the state was a ket, return the eigenstates of the measurement
-        operator. Otherwise return the projectors onto the eigenstates.
+    projectors: list of :class:`.Qobj`
```
hodgestar, line 209: Returning the eigenstates in the case of a `ket` state was useful because it directly provided the post-measurement state, but I guess this change is necessary to support degenerated eigenvalues.

qutip/measurement.py
```diff
@@ -178,33 +197,57 @@ def measurement_statistics_observable(state, op):
...
-    else:
-        projectors = [e.proj() for e in eigenstates]
-        probabilities = [expect(v, state) for v in projectors]
-        return eigenvalues, projectors, probabilities
+    # Detect groups of eigenvalues within atol of each other.
+    # A group will be [True] + [False] * N
+    groups = np.append(True, np.diff(eigenvalues) >= tol)
+
+    present_group = []
+    for i in range(len(eigenvalues)-1, -1, -1):
```
hodgestar, line 230: Could we not just append the extra True at the end of the groups and go forward through the list of eigenvalues?

hodgestar (commented):
One suggestion but otherwise looks good.

hodgestar (approved):

PR #2346, opened 2024-03-05
Prepare qutip v5.0.0b1

Ericgig:
**Description**
Merge master into qutip-5.0.X for a beta release for a full release soon.
Version and classifiers updated.


doc/changes/2289.misc
hodgestar, line 1: I'm a bit surprised a changes file got added. I would have expected them all to be deleted when the changelog was updated.

setup.cfg
```diff
@@ -31,13 +31,14 @@ packages = find:
...
-    scipy>=1.0
+    numpy>=1.22
+    scipy>=1.8,<1.12
     packaging
 setup_requires =
-    numpy>=1.13.3
-    scipy>=1.0
-    cython>=0.29.20
+    numpy>=1.19
+    scipy>=1.8
```
hodgestar, line 39: It seems a bit odd to allow `scipy>1.12` for building the package but not for installing it, so just double-checking this.

doc/changes/2289.misc
Ericgig: This was merged after I had towncrier build the changelog...
I will fix it.

setup.cfg
Ericgig: We had a few issues when 1.12 got out.
While this is probably fixed, our tests are all done with this respected. I think we should leave it as is for beta 1 and remove it for the next release

hodgestar: Sounds good.

.github/workflows/build.yml
```diff
@@ -107,11 +107,11 @@ jobs:
...
         os: [ubuntu-latest, windows-latest, macos-latest]
     env:
-      # Set up wheels matrix.  This is CPython 3.8--3.10 for all OS targets.
-      CIBW_BUILD: "cp3{8,9,10,11}-*"
+      # Set up wheels matrix.  This is CPython 3.9--3.12 for all OS targets.
+      CIBW_BUILD: "cp3{9,10,11,12}-*"
       # Numpy and SciPy do not supply wheels for i686 or win32 for
       # Python 3.10+, so we skip those:
-      CIBW_SKIP: "*-musllinux* cp3{8,9,10,11}-manylinux_i686 cp3{8,9,10,11}-win32"
+      CIBW_SKIP: "*-musllinux* cp3{10,11,12}-manylinux_i686 cp3{10,11,12}-win32"
```
hodgestar, line 114: Should we not also skip 3.9 here?

hodgestar: I think we should skip the same ones we did before plus the 3.12 ones and minus any we've removed.

Agreed the comment is now inaccurate. Perhaps we can investigate before the next release.

hodgestar (approved):

PR #2350, opened 2024-03-14
Fix reshape_dense integer overflow

Ericgig:
**Description**
`reshape_dense` use `idxint` for internal computation where the input can by `size**2` and could overflow.
A 11 qubit system in `mesolve` could overflow when the input `dm` was `C` ordered (found by @nwlambert).

This fix this overflow and speed up `column_stack_dense` for `C` ordered.


hodgestar (approved):
Would it be hard to add a test for the big overflow case (that would now pass)?

Ericgig:
Medium for that specific case.
But I expect that we could have a few bug when using matrices close to the size limit.
I want to make a proper review eventually.

PR #2351, opened 2024-03-15
add output type to qobj.eigenstate

### LINKED ISSUE #2338: Allow returning eigenstates as an operator.
### Problem Description

`qobj.eigenstates()` return the states as a list of kets, but it is something more useful to have them as a single operator where each state is a column.

### Proposed Solution

Add a keyword argument (`output_type`?) which choose whether a list of kets or an operator is to be returned.
Eigenstates are computed as an operator, we then just have to skip the step that split them into kets.
The normalisation, phase fixing steps should still be completed.

### Alternate Solutions

_No response_

### Additional Context

Idea advanced in #2328.

rochisha0:
**Description**
Add output_type to qobj.eigenstate() to add return type as both operator as well as kets

**Related issues or PRs**
fix #2338

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1539,12 +1539,14 @@ class Qobj:
                 phase = np.array([np.abs(ket[phase_fix, 0]) / ket[phase_fix, 0]
                                 if ket[phase_fix, 0] else 1
                                 for ket in ekets])
+
             return evals, ekets / norms * phase
         elif output_type == 'operator':
             ekets = Qobj(evecs)
             norm = ekets.norm()
+
             return evals, ekets/norm
-        
+ 
     def eigenenergies(self, sparse=False, sort='low',
                       eigvals=0, tol=0, maxiter=100000):
         """Eigenenergies of a quantum object.
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1541,11 +1541,17 @@ class Qobj:
                                 for ket in ekets])
 
             return evals, ekets / norms * phase
+        
         elif output_type == 'operator':
-            ekets = Qobj(evecs)
-            norm = ekets.norm()
-
-            return evals, ekets/norm
+            ekets = Qobj(evecs, dims=[new_dims[0], [evecs.shape[1]]])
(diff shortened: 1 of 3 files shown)


Ericgig (commented):
@rochisha0 
Please tell us when this is ready to review.
It will need tests with the new option and a towncrier entry.

rochisha0:
@Ericgig I have made the changes upto best of my knowledge, I am not able to figure out how to apply phase fixing to the new operator type. 

Also after a few iterations my local build seems to give cython errors, is there a wiki on how to sort that out?

Ericgig:
Cython error are quite vague, but there should not be any cython error here.
It could be errors inside a cython function. Could you send the log if it persist.

qutip/core/qobj.py
```diff
@@ -1524,18 +1527,26 @@ def eigenstates(self, sparse=False, sort='low', eigvals=0,
...
+                phase = np.array([1] * len(ekets))
+            else:
+                phase = np.array([np.abs(ket[phase_fix, 0]) / ket[phase_fix, 0]
+                                if ket[phase_fix, 0] else 1
+                                for ket in ekets])
+
+            return evals, ekets / norms * phase
+        elif output_type == 'operator':
+            ekets = Qobj(evecs)
+            norm = ekets.norm()
```
Ericgig, line 1546: Closer but still not there.
Dimensions are not set.

The output should be the same as the kets concatenated in a single matrices:
```
op = qutip.rand_herm(N)
_, kets = op.eigenstates(output_type='kets', phase_fix=M)
_, oper = op.eigenstates(output_type='operator', phase_fix=M)
qutip.Qobj(np.hstack([vec.full() for vec in kets]), dims) == oper
```
The norm of a matrix, per default it the trace norm: `tr(sqrt(A @ A.adjoint()))`, while for the ket it's l2 norm: `sqrt(sum(|a_i|**2))`. 

For normalization and phase_fix, you may want to take a look into `Qobj.inv`, `qutip.qdiags`.

Ericgig:
@rochisha0 are you still working on this?

rochisha0:
@Ericgig Yes, I'm working on this. Caught up in some tasks, will try to complete over weekend.

Ericgig:
Could you fix the conflict.

PR #2354, opened 2024-03-19
Remove pin on scipy 1.12

eendebakpt:
**Description**

We allow scipy version 1.12 (released about 2 months ago) to be used with qutip. Scipy 1.12 is already tested in the build matrix https://github.com/qutip/qutip/blob/master/.github/workflows/tests.yml#L93


Ericgig (approved):
Thank you.

PR #2359, opened 2024-03-24
Multi trajectory solver: minor internal changes

pmenczel:
A few proposed changes to internals of the MultiTrajSolver and its subclasses. The changes are (a) to increase consistency between the monte carlo solvers and the stochastic solvers, (b) to reduce repeating code from MultiTrajSolver in the subclasses, (c) to fix some documentation.

* Renamed `_MTSystem` and `_MCSystem` into `_MultiTrajRHS` and `_MCRHS` to make them analogous to `_StochasticRHS`. Removed distinction between `.system` and `.rhs` which seemed unnecessary and confusing to me (please tell me if I'm wrong).
* Made `_integrate_one_traj` use the integrator's run method (which doesn't do anything special currently, but I imagine that it could potentially be implemented more efficiently than repeated calls to integrate). Made subclasses use `_run_one_traj`, `_integrate_one_traj` etc of the parent class where possible, instead of duplicating logic.
* The docstring of MCSolver.run was not documenting its parameters, instead referring to the docstring of MultiTrajSolver.run, but MultiTrajSolver is not included in the docs.

### DIFF
--- a/qutip/solver/mcsolve.py
+++ b/qutip/solver/mcsolve.py
@@ -478,25 +478,12 @@ class MCSolver(MultiTrajSolver):
         })
         return stats
 
-    def _initialize_run_one_traj(self, seed, state, tlist, e_ops,
-                                 no_jump=False, jump_prob_floor=0.0):
-        result = self._trajectory_resultclass(e_ops, self.options)
-        generator = self._get_generator(seed)
-        self._integrator.set_state(tlist[0], state, generator,
-                                   no_jump=no_jump,
-                                   jump_prob_floor=jump_prob_floor)
-        result.add(tlist[0], self._restore_state(state, copy=False))
-        return result
-
-    def _run_one_traj(self, seed, state, tlist, e_ops, no_jump=False,
-                      jump_prob_floor=0.0):
+    def _run_one_traj(self, seed, state, tlist, e_ops, **integrator_kwargs):
         """
         Run one trajectory and return the result.
         """
-        result = self._initialize_run_one_traj(seed, state, tlist, e_ops,
-                                               no_jump=no_jump,
-                                               jump_prob_floor=jump_prob_floor)
-        seed, result = self._integrate_one_traj(seed, tlist, result)
+        seed, result = super()._run_one_traj(seed, state, tlist, e_ops,
+                                             **integrator_kwargs)
         result.collapse = self._integrator.collapses
         return seed, result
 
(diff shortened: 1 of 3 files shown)


qutip/solver/mcsolve.py
```diff
@@ -373,10 +367,6 @@ def _do_collapse(self, collapse_time, state):
     def arguments(self, args):
         if args:
             self._integrator.arguments(args)
-            for c_op in self._c_ops:
-                c_op.arguments(args)
-            for n_op in self._n_ops:
-                n_op.arguments(args)
```
Ericgig, line 379: Why did you remove the update here? The mc integrator is the one using the c_ops and n_ops so it should update them if the arguments are changed. Sure they are a reference to the ones in the MCRHS. But that would be the same for the the `rhs` in the integrator...

qutip/solver/mcsolve.py
```diff
@@ -167,36 +167,30 @@ def mcsolve(H, state, tlist, c_ops=(), e_ops=None, ntraj=500, *,
...
     """
     Container for the operators of the solver.
     """
 
-    def __init__(self, rhs, c_ops, n_ops):
-        self.rhs = rhs
+    def __init__(self, H, c_ops, n_ops):
+        self.rhs = H
+
+        self.H = H
```
Ericgig, line 178: Why both?
Here `H` is wrong here since it can be both an effective Hamiltonian or Liouvillian.

qutip/solver/multitraj.py
```diff
@@ -8,23 +8,22 @@
...
-
     def arguments(self, args):
         self.rhs.arguments(args)
 
     def _register_feedback(self, type, val):
         pass
 
     def __getattr__(self, attr):
+        if attr == "rhs":
+            raise AttributeError
```
Ericgig, line 26: Not sure why this is useful.

Ericgig (commented):
Thank you for cleaning this.
For me, rhs is the operator used in the ODE solver, system is the set of operators defining the quantum system. MC's c_ops and n_ops are part of the system, but not the rhs. That said if you find that easier to understand, we can rename them.

qutip/solver/mcsolve.py
pmenczel: I had this idea that this should not be the responsibility of the integrator. But you are right, I changed it back. I guess it doesn't hurt to call `arguments` multiple times with the same args.

qutip/solver/mcsolve.py
pmenczel: You are right, changed it back

qutip/solver/multitraj.py
pmenczel: It is necessary, otherwise the call to `self.rhs` below will call itself recursively and cause a stack overflow. (This is how it was already done in the `mcsolve` module previously.)

Ericgig (approved):

PR #2361, opened 2024-03-24
Fix failing `svd_csr` test

Ericgig:
**Description**
The sparse svd test is causing segfault in a setup that I can't reproduce.
The function `svd_csr` was never used as it could not work as a `svd_dense` replacement: `_data.svd(csr)` would use `svd_dense`.

So I simply removed the function.


Ericgig:
There is one particular setup where the sparse eigen solver segfault. This happen with pure scipy and is not related to qutip.
Looking into it, I saw that when we installed qutip, it updated scipy version. It downgraded scipy 1.12 to 1.11, thus we never tested for 1.12.
Also build isolation meant that the version of numpy, scipy and cython we set in our test matrix did not match the one at installation. This is not ideal for cython since the compilation version could have a major impact.

I set it to not use build isolation and used the matrix set version to compile qutip.
This had the side effect of not using the setup that caused the segfault and solving the failing test issue.

hodgestar (approved):

Issue #2235, opened 2023-09-29
NonMarkovianMCSolver does not support improved sampling


pmenczel:
### Problem Description

The `improved_sampling` option was recently added to `mcsolve`. It might also be useful to have in `nm_mcsolve`, but adaptations to the code are required to make it work. Currently, we have disabled this option for the `NonMarkovianMCSolver`, see [PR 2234](https://github.com/qutip/qutip/pull/2234).

Since `NonMarkovianMCSolver` inherits from `MCSolver`, one might expect this option to work "out of the box". The technical reason why it does not is that `nm_mcsolve` calculates the value of an influence martingale, which is factored into the calculation of expectation values in the custom result class `NmmcResult`. If `improved_sampling` is enabled, the result class `McResultImprovedSampling` is used instead, which is not aware of the martingale.

### Proposed Solution

Both the improved sampling algorithm and `nm_mcsolve` give trajectories a *weight*. In the former case, the weight is related to the probability of the trajectory without jumps; in the latter case it is the influence martingale. I imagine it might be useful to give `McResult` itself the capability to handle weighted trajectories.  
This capability would be useful for one additional purpose: if we want to allow mixed initial conditions for `mcsolve` in the future. In this case, the trajectories would need to be weighted with a correction factor, because the fraction $f_i$ of trajectories starting in a state $i$ will generally not be equal to the corresponding eigenvalue $p_i$ of the density matrix $\rho = \sum p_i |i \rangle\langle i|$. (Hence, the correction factor would be $p_i / f_i$.)

If people think this would be useful (including the mixed initial states), I will try to implement this proposal once I can find some time to do so.

### Alternate Solutions

Alternatively, we could leave things as they are now and not support the improved sampling for `nm_mcsolve`. Another less invasive option would be copying the improved sampling algorithm to `nm_mcsolve` (adding a `NmmcResultImprovedSampling` class).

Ericgig:
Fixed with #2369

PR #2369, opened 2024-03-29
Weighted trajectories

pmenczel:
**Description**

Introduces weights for the trajectories of trajectory solvers. These weights enter all trajectory averages. This concept is a generalization of two things that are already in qutip:
* The influence martingale in `nm_mcsolve` is simply a time-dependent weight from this point of view
* The "improved sampling" option for `mcsolve` weights the no-jump trajectory with its exact probability, and then only samples from the ensemble of trajectories with at least one jump.

Currently, these two things cannot be combined: "improved sampling" does not work for `nm_mcsolve`. With these changes, it would work.
In the future, there might be at least one more application for weighted trajectories:
* For mixed initial states in `mcsolve`, it might be useful to allow sampling of initial states with frequencies or probabilities that are different from their true prefactor in the initial state. The ratio between the sampling frequencies and true prefactors would be weights on the trajectories.

This also fixes some things in merging `McResult`s and `NmmcResult`s, and perhaps fixes the target tolerance computation for the "improved sampling" option.

**Implementation**

A difficulty in the implementation is that the weights depend on the total number of trajectories in different ways. For example, the no-jump trajectory has a fixed weight that never changes, but all other trajectories come with prefactors $(1-p_0) / (N-1)$ where $p_0$ is the probability of the no-jump trajectory, and $N$ the total number of trajectories (including the one no-jump trajectory). For this reason, I separate weights into *absolute weights* (like $p_0$) and *relative weights* (like the others). The average is performed as follows:

$$ \bar\rho = \sum_{T: \text{abs}} w_a(T) w_r(T) \rho(T) + \frac{1}{N_{\text{rel}}} \sum _{T: \text{rel}} w_r(T) \rho(T) $$

where the first sum is over all trajectories T that have absolute weights, and the second sum over all other trajectories. Here, $w_a(T)$ and $w_r(T)$ are the absolute and relative weights, $\rho(T)$ the state associated with a trajectory, and $N_{\text{rel}}$ the number of trajectories that have not been assigned absolute weights.

Intuitively, the idea is that absolute weights are for trajectories where the contribution to the full state is known exactly, and the trajectories without absolute weights are the stochastic part.

**Things to do**
- [x] Update tests and create new ones
- [x] Update photocurrent to work with weights
- [x] Update target tolerance calculation to work with weights
- [x] Think about how result objects should be added if both contain trajectories with absolute weights
- [x] Double-check that this creates no problems for stochastic solvers
- [x] Towncrier entry

**Related issues or PRs**
https://github.com/qutip/qutip/issues/2235


qutip/solver/result.py
```diff
@@ -1046,6 +1201,8 @@ def photocurrent(self):
         """
         Average photocurrent or measurement of the evolution.
         """
+        # TODO this is wrong if trajectories have weights
+        # unclear how to implement in case of time-dependent weights
```
Ericgig, line 1205: Does photocurrent make sense for nm_mcsolve?
If it does not, we could raise an error when used with time-dependent weights.

qutip/solver/result.py
```diff
@@ -894,78 +919,208 @@ def __repr__(self):
...
-            return NotImplemented
+            raise NotImplementedError("Can only add two multi traj results")
         if self._raw_ops != other._raw_ops:
             raise ValueError("Shared `e_ops` is required to merge results")
         if self.times != other.times:
             raise ValueError("Shared `times` are is required to merge results")
 
         new = self.__class__(
-            self._raw_ops, self.options, solver=self.solver, stats=self.stats
+            self.e_ops, self.options, solver=self.solver, stats=self.stats
```
Ericgig, line 929: `e_ops` are callable wrapped in `ExpectOp`, by passing them here, they get wrapped twice and raw_ops are no longer available.

qutip/solver/result.py
```diff
@@ -894,78 +919,208 @@ def __repr__(self):
 
     def __add__(self, other):
         if not isinstance(other, MultiTrajResult):
-            return NotImplemented
+            raise NotImplementedError("Can only add two multi traj results")
```
Ericgig, line 922: Why not let python raise the error when no appropriate functions is found?
Not that I think this would happen, it stop other project to develop something compatible with our results.

Ericgig (commented):
Thank you for merging the results classes.

The default seems to be a relative weight of `1`. Meaning that it should work fine in solver which does not use it (stochastic). But should they just keep using `Result` until we have a use for those weight?

It there a balancing tools for weight values?
If there are 2 runs, can both have a relative weight of 10 or 0.1 without affecting the end results? If there are 2 starting kets that we want to merge with a weight of 10:1, does the actual values matter or only the ratio?

I feel it would be safer to compute the relative weight as `W(T) / sum_T(W(T))` instead of using the number of trajectories.

I am not sure how it works with weight changing over time...

qutip/solver/result.py
pmenczel: Sorry, this was a mistake, reverted the change

qutip/solver/result.py
pmenczel: It probably does not make sense for nm_mcsolve. I have added an error in the case of time-dependent weights, and removed the tests for it.

pmenczel:
Thank you for your review.

> The default seems to be a relative weight of `1`. Meaning that it should work fine in solver which does not use it (stochastic). But should they just keep using `Result` until we have a use for those weight?

I let them use `Result` again for now. I needed to add some lines [here](https://github.com/pmenczel/qutip/blob/weighted-trajectories/qutip/solver/result.py#L710) so that the new `MultiTrajResult` can deal with it. (Which might be a good idea anyway.)

> It there a balancing tools for weight values? If there are 2 runs, can both have a relative weight of 10 or 0.1 without affecting the end results? If there are 2 starting kets that we want to merge with a weight of 10:1, does the actual values matter or only the ratio?

I have added a `merge` function that lets the user merge trajectory results with custom relative weights. I think trajectories with absolute or relative weights are handled consistently. Calling `result1 + result2` uses default values for the relative weights, chosen such that every trajectory contributes equally.

> I feel it would be safer to compute the relative weight as `W(T) / sum_T(W(T))` instead of using the number of trajectories.

Hmm unfortunately I think that this won't work. For the "improved sampling" case, the sum of the relative weights is not one.

qutip/solver/result.py
```diff
@@ -895,7 +970,31 @@ def __repr__(self):
         lines.append(">")
         return "\n".join(lines)
 
-    def __add__(self, other):
+    def merge(self, other, p=None):
```
Ericgig, line 973: The `__add__` is already in the users hand and part of the official interface.
It is documented in the [guide](https://qutip.readthedocs.io/en/qutip-5.0.x/guide/dynamics/dynamics-monte.html#changing-the-number-of-trajectories) (and maybe the tutorials).
It's new, so it's probably not used much, but still we shouldn't remove it without a deprecation warning or a major release...

It would be better to keep it as merging with the default value of `p`.

Ericgig (commented):
It's nice to merge `McResultImprovedSampling` into `McResult`, but the overall complexity has increased... I believe it would be better to move `MultiTrajResult` to it's own file.

Did you review the `dynamics-nmmonte,rst`? Adding a note/example of using the improved sampling option there would be good.

pmenczel:
> It's nice to merge `McResultImprovedSampling` into `McResult`, but the overall complexity has increased... I believe it would be better to move `MultiTrajResult` to it's own file.
> 
> Did you review the `dynamics-nmmonte,rst`? Adding a note/example of using the improved sampling option there would be good.

Thank you for the review. The complexity has increased, more than I expected when I started writing. I hope it is worth it?
I have extracted the `MultiTrajResult` and its subclasses to a new module, `qutip.solver.multitrajresult`.

I have added a mention about the improved sampling option in the guide, and I will have a look at the tutorial notebook whether it makes sense to include somewhere there.

qutip/solver/stochastic.py
```diff
@@ -165,6 +165,12 @@ def wiener_process(self):
         """
         return self._trajectories_attr("wiener_process")
 
+    def merge(self, other, p=None):
+        raise NotImplementedError("Merging results of the stochastic solvers "
+                                  "is currently not supported. Please raise "
+                                  "an issue on GitHub if you would like to "
+                                  "see this feature.")
+
```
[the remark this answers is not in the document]
pmenczel: Currently, merging the `measurement`, `dW` etc attributes of stochastic results doesn't work. That was the reason why I removed merge here, but maybe that went a bit too far.

Should I just remove this, or maybe leave in a warning in the case a user tries to merge stochastic results where `keep_measurements` was used?

qutip/tests/solver/test_results.py
```diff
@@ -352,3 +459,115 @@ def test_merge_result(self, keep_runs_results):
...
+                                                "num_collapse": 2})
+
+        for _ in range(ntraj):
+            traj = TrajectoryResult(res._raw_ops, res.options)
+            seeds = np.random.randint(10_000, size=len(tlist))
+            for t, seed in zip(tlist, seeds):
+                random_state = qutip.rand_ket(dim, seed=seed)
+                traj.add(t, random_state)
+
+            if time_dep_weights and np.random.randint(2):
```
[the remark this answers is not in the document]
pmenczel: I wanted to add weights to only some of the trajectories, to test that nothing goes wrong if some of the traj don't have weights.

Ericgig (approved):
I think the increased complexity is fine in this case.
Everything looks good. Thank you.

PR #2371, opened 2024-03-29
fix the negativity function

### LINKED ISSUE #2345: Negativity function does not work
### Bug Description

The function `negativity() ` is currently broken because of a broken import of `partial_transpose()`.   This is probably an easy fix and good first issue, but more importantly we need a test for this function, and anything else that currently is not tested in `entropy.py`.

Also, might be useful to either properly support kets in negativity (i think it just assumes at the moment the input state is a density operator, and incorrectly acts on kets), or give a warning if a state is provided, and not a density operator.

### Code to Reproduce the Bug

```shell
import numpy as np
from qutip import *

test = (tensor(basis(2,0),basis(2,1)) + tensor(basis(2,1),basis(2,0)))/np.sqrt(2)
negativity(test * test.dag(),0)
```

### Code Output

```shell
123 """
    124 Compute the negativity for a multipartite quantum system described
    125 by the density matrix rho. The subsys argument is an index that
   (...)
    130     Experimental.
    131 """
    132 mask = [idx == subsys for idx, n in enumerate(rho.dims[0])]
--> 133 rho_pt = partial_transpose(rho, mask)
    135 if method == 'tracenorm':
    136     N = ((rho_pt.dag() * rho_pt).sqrtm().tr().real - 1)/2.0

TypeError: 'module' object is not callable
```

### Expected Behaviour

should be = 0.5

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+c745ace
Numpy Version:      1.26.3
Scipy Version:      1.11.4
Cython Version:     3.0.8
Matplotlib Version: 3.8.0
Python Version:     3.12.1
Number of CPUs:     28
BLAS Info:          Generic
INTEL MKL Ext:      False
```

### Additional Context

_No response_

vikas-chaudhary-2802:
**Description**
The import statement was added to import the partial_transpose function directly from the qutip module. This was done to fix the TypeError: 'module' object is not callable error.

Added a condition that check if the input rho is a ket (a type of quantum state) and it also ensures that the negativity function can handle both kets and density operators as input.

Fix #2345


qutip/entropy.py
```diff
@@ -4,7 +4,8 @@
 
 from numpy import conj, e, inf, imag, inner, real, sort, sqrt
 from numpy.lib.scimath import log, log2
-from . import (ptrace, ket2dm, tensor, sigmay, partial_transpose,
+from qutip import partial_transpose, ket2dm
```
nwlambert, line 7: maybe better to use `from .partial_transpose import partial_transpose`  and keep ket2dm in the lower input line?

qutip/entropy.py
```diff
@@ -129,6 +130,8 @@ def negativity(rho, subsys, method='tracenorm', logarithmic=False):
 
         Experimental.
     """
+    if rho.isket:
```
nwlambert, line 133: how about ` if rho.isket or rho.isbra:`

nwlambert (commented):
Thanks for the fixes, could you also add a test?  just something checking that the negativity of a maximally entangled two-qubit state is close to 0.5, and check the operator, ket and bra inputs?

vikas-chaudhary-2802:
Due to my exam collision, I was a little bit late. I apologize for that. Currently, I am working on a test case to check whether the negativity of a maximally entangled two-qubit state is close to 0.5, and I'm also verifying the operator, ket, and bra inputs. Once it's done, I will make a PR.

ytdHuang:
For `method == 'eigenvalues'`, the more fundamental way is to calculate $N(\rho^{T_A})=\left|\sum_{\lambda_i < 0}\lambda_i\right|$, where $T_A$ means the partial transpose of party $A$, and $\lambda_i$ are the corresponding eigenvalues.

$N=\sum_i \frac{|\lambda_i|-\lambda_i}{2}$ is only true if the trace of $\rho^{T_A}$ equals to `1`. Although, most of the cases should be true, but why don't we use the more fundamental way to calculate it.

nwlambert:
@vikas-chaudhary-2802 can you add the test to this PR?  it can be very simple.  if you feel like it will take some time, i can merge this  and you can do a separate one later.

@ytdHuang thanks for the suggestion, do you think it would be better to use that as the default method?  in any case, lets do that in a separate issue/PR

qutip/entropy.py
```diff
@@ -4,7 +4,9 @@
 
 from numpy import conj, e, inf, imag, inner, real, sort, sqrt
 from numpy.lib.scimath import log, log2
-from . import (ptrace, ket2dm, tensor, sigmay, partial_transpose,
+from numpy import np
```
nwlambert, line 7: please remove this, it is failing the tests (it should be import numpy as np, but np is not used anywhere? so better just to remove)

ytdHuang:
@nwlambert 
I think it depends on whether `sqrtm()` or `eigenenergies()` performs better (or less issues in general case) in `Qobj`.

vikas-chaudhary-2802:
Hi, @nwlambert 1 test is failing , can you please suggest me how can i fix it ?

nwlambert:
Hi Vikas, that is just the towncrier check, towncrier is a method to track your contribution.  You can follow the instructions here in the changelog generation section https://qutip.readthedocs.io/en/stable/development/contributing.html

Once you do that it should pass

vikas-chaudhary-2802:
Hi, @nwlambert Please review this PR

doc/changes/2371.bugfix
```diff
@@ -0,0 +1,4 @@
+Add your info here
```
nwlambert, line 1: sorry, please remove this line, then i will merge this.

nwlambert (commented):
thanks for the contribution, one more minor thing and we will merge this.  we can add the test when resolving yi-te's issue.

nwlambert (approved):

vikas-chaudhary-2802:
Hi, @nwlambert thank you for your support. I have made the changes as you instructed.

PR #2380, opened 2024-04-03
Use CSR as the default for expand_operator

BoxiLi:
**Description**
Use CSR as the default for expand_operator because the output is usually sparse.

### DIFF
--- a/qutip/core/tensor.py
+++ b/qutip/core/tensor.py
@@ -437,9 +437,9 @@ def expand_operator(oper, dims, targets, dtype=None):
     targets : int or list of int
         The indices of subspace that are acted on.
     dtype : str, optional
-        Data type of the output :class:`.Qobj`. By default it uses the data type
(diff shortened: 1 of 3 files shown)


qutip/core/tensor.py
```diff
@@ -435,13 +436,17 @@ def expand_operator(oper, dims, targets):
         E.g ``[2, 3, 2, 3, 4]``.
     targets : int or list of int
         The indices of subspace that are acted on.
+    dtype : str, optional
+        Data type of the output `Qobj`.
```
AGaliciaMartinez, line 440:
```suggestion
        Data type of the output `Qobj`. By default it uses the data type specified  in settings. If not data type is specified in settings it uses `CSR` data type.
```

Ericgig (approved):
Looks good.
Just fix the line length if you merge Asier's suggestion.

AGaliciaMartinez (approved):

PR #2382, opened 2024-04-04
fix end condition on mcsolve when using target tolerance

magzpavz:
**Description**
Previously, if you specified a target tolerance when using `mcsolve`, the result's `end_condition` would show "target tolerance reached" regardless of whether this was true. I added a line to check this and set `end_condition` to "ntraj reached" when appropriate.


Ericgig:
Thank you @magzpavz for catching this.

Could you add some tests so the error doesn't reappears.

pmenczel (approved):
Looks good to me as well

PR #2383, opened 2024-04-04
Set requirements to refuse scipy 1.13

Ericgig:
**Description**
Scipy 1.13 change the way sparse operations are done.
Some changes to scipy fix most issues, but I get some segfault with the sparse eigen solver and I am unable to run all tests.
1.13 work with v5 so I changed the requirements to exclude that version in v4.7.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
         # matrix size; make sure to test all supported versions in some form.
         python-version: ["3.9"]
         case-name: [defaults]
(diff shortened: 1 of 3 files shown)


AGaliciaMartinez:
Im +1 on restricting scipy version.

hodgestar (approved):

lazyoracle:
Hey, thanks for the workaround with pinning scipy. How long does it typically take for a qutip version to show up on conda-forge after it is already released by you folks and made available on pypi?

PR #2388, opened 2024-04-09
Improve Qobj creation function metadata setting

Ericgig:
**Description**
Many of the operator functions did not set the `isherm` flag.
Not having the flag set would cause issue with `jax.jit`.

- Set the `isherm` flag for almost every functions in `operators.py` and `gates.py`. 
  - `qdiags`  flags are only set when only one diagonal in used.
  - I did not touch functions creating operators from other operators ( `commutator`, `squeezing`)
- Set the `isunitary` flags for all these also.
- Updated the gates `dtype` parameter to use the `default_dtype` settings.
- Updated the sigma operators to have a `dtype` parameter.
- Added tests for all of these.

### DIFF
--- a/qutip/core/gates.py
+++ b/qutip/core/gates.py
@@ -370,7 +370,9 @@ def cphase(theta, *, dtype=None):
         Quantum object representation of controlled phase gate.
     """
     dtype = dtype or settings.core["default_dtype"] or _data.CSR
-    return qdiags([1, 1, 1, np.exp(1.0j * theta)], dims=_DIMS_2_QB, dtype=dtype)
+    return qdiags(
+        [1, 1, 1, np.exp(1.0j * theta)], dims=_DIMS_2_QB, dtype=dtype
+    )
 
 
 def cnot(*, dtype=None):
@@ -756,10 +758,15 @@ def hadamard_transform(N=1, *, dtype=None):
     """
     dtype = dtype or settings.core["default_dtype"] or _data.Dense
     data = 2 ** (-N / 2) * np.array(
-        [[(-1) ** _hamming_distance(i & j) for i in range(2**N)] for j in range(2**N)]
+        [
+            [(-1) ** _hamming_distance(i & j) for i in range(2**N)]
+            for j in range(2**N)
+        ]
     )
 
-    return Qobj(data, dims=[[2] * N, [2] * N], isherm=True, isunitary=True).to(dtype)
+    return Qobj(data, dims=[[2] * N, [2] * N], isherm=True, isunitary=True).to(
+        dtype
+    )
 
 
 def _powers(op, N):
--- a/qutip/core/gates.py
+++ b/qutip/core/gates.py
@@ -42,6 +42,10 @@ __all__ = [
 ]
 
 
+_DIMS_2_QB = Dimensions([[2, 2], [2, 2]])
+_DIMS_3_QB = Dimensions([[2, 2, 2], [2, 2, 2]])
+
+
 def cy_gate(*, dtype=None):
     """Controlled Y gate.
 
@@ -59,7 +63,7 @@ def cy_gate(*, dtype=None):
     dtype = dtype or settings.core["default_dtype"] or _data.CSR
     return Qobj(
         [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, -1j], [0, 0, 1j, 0]],
-        dims=dims_2_qb,
+        dims=_DIMS_2_QB,
         isherm=True,
         isunitary=True,
     ).to(dtype)
@@ -80,7 +84,7 @@ def cz_gate(*, dtype=None):
         Quantum object for operator describing the rotation.
     """
     dtype = dtype or settings.core["default_dtype"] or _data.CSR
-    return qdiags([1, 1, 1, -1], dims=dims_2_qb, dtype=dtype)
+    return qdiags([1, 1, 1, -1], dims=_DIMS_2_QB, dtype=dtype)
 
 
 def s_gate(*, dtype=None):
@@ -119,7 +123,7 @@ def cs_gate(*, dtype=None):
 
     """
     dtype = dtype or settings.core["default_dtype"] or _data.CSR
-    return qdiags([1, 1, 1, 1j], dims=dims_2_qb, dtype=dtype)
+    return qdiags([1, 1, 1, 1j], dims=_DIMS_2_QB, dtype=dtype)
 
 
 def t_gate(*, dtype=None):
@@ -159,7 +163,7 @@ def ct_gate(*, dtype=None):
     dtype = dtype or settings.core["default_dtype"] or _data.CSR
     return qdiags(
         [1, 1, 1, np.exp(1j * np.pi / 4)],
-        dims=dims_2_qb,
+        dims=_DIMS_2_QB,
         dtype=dtype,
     )
 
@@ -348,9 +352,6 @@ def qrot(theta, phi, *, dtype=None):
 #
 
 
-dims_2_qb = Dimensions([[2, 2], [2, 2]])
-
-
 def cphase(theta, *, dtype=None):
     """
     Returns quantum object representing the controlled phase shift gate.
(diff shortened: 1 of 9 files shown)


qutip/core/operators.py
```diff
@@ -334,10 +345,11 @@ def sigmam():
      [ 1.  0.]]
 
     """
-    return _SIGMAM.copy()
+    dtype = dtype or settings.core["default_dtype"] or _data.CSR
```
hodgestar, line 348: Perhaps we should have a function for this line? E.g. Perhaps `.to` could take a `default_dtype` parameter or similar. Depends whether it would help in enough cases.

hodgestar (changes requested):

qutip/core/operators.py
```diff
@@ -318,10 +328,11 @@ def sigmap():
      [ 0.  0.]]
 
     """
-    return _SIGMAP.copy()
+    dtype = dtype or settings.core["default_dtype"] or _data.CSR
```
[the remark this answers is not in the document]
Ericgig: If used by themselves, dense would be right, but they are often used as building block of larger operator and the dense representation is sticky: binary operation between dense and sparse usually return dense. I though this would make a lot of system dense unless the user does some conversions, which is not what we want.
CSR felt safer in that case.

qutip/core/operators.py
```diff
@@ -615,7 +630,10 @@ def _f_op(n_sites, site, action, dtype=None):
 
     eye = identity(2, dtype=dtype)
     opers = [s_z] * site + [operator] + [eye] * (n_sites - site - 1)
-    return tensor(opers).to(dtype)
+    out = tensor(opers).to(dtype)
+    out.isherm = False
+    out._isunitary = False
```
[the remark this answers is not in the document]
Ericgig: `isunitary` property does not have a setter. We could set `_isherm` instead if we really wanted symmetry.
It does not bother me so I will let you fix it if it bother you.

qutip/core/qobj.py
```diff
@@ -391,7 +394,9 @@ def to(self, data_type):
             converter = _data.to[data_type]
         except (KeyError, TypeError):
             raise ValueError("Unknown conversion type: " + str(data_type))
-        if type(self._data) is data_type:
+        if type(self._data) is data_type and copy:
```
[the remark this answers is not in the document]
Ericgig: While you say this, I just realized `data_type` could be an alias instead of a type, but it's not documented and the check for no change is incomplete...
This feels outside the PR but I will fix it while we are at it.

qutip/tests/core/test_gates.py
```diff
@@ -142,3 +150,49 @@ def test_gate_normalises_pauli_group(self, gate):
                     del pauli_gates[i]
                     break
         assert len(pauli_gates) == 0
+
+
+@pytest.mark.parametrize("dtype", [qutip.data.Dense, qutip.data.CSR])
```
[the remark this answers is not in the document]
Ericgig: I choose 2 types so we don't only get the default, but not all type since we are not testing the conversion code, just that it is used. Adding `dia` would just add more test without testing anything new.

qutip/core/gates.py
```diff
@@ -738,10 +815,14 @@ def qubit_clifford_group(*, dtype="dense"):
...
-    return [
+    gates = [
         op.to(dtype)
         for op in map(
             partial(reduce, mul),
             product(_powers(E, 3), _powers(X, 2), _powers(S, 4)),
         )
     ]
+    for gate in gates:
+        gate.isherm
```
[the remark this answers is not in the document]
Ericgig: Set the flags, since there are 24 of them with some being hermitian other not, I call the method compute them. Since they are all 2x2, it does not that a significant time.

qutip/core/gates.py
```diff
@@ -52,13 +60,16 @@ def cy_gate(*, dtype="csr"):
     result : :class:`.Qobj`
         Quantum object for operator describing the rotation.
     """
+    dtype = dtype or settings.core["default_dtype"] or _data.CSR
```
BoxiLi, line 63: Just a comment, given that this line is everywhere in qutip core, is there a way to simplify it?

qutip/core/gates.py
```diff
@@ -167,19 +186,25 @@ def rx(phi, *, dtype="dense"):
...
 
     """
+    dtype = dtype or settings.core["default_dtype"] or _data.Dense
     return Qobj(
         [
             [np.cos(phi / 2), -1j * np.sin(phi / 2)],
             [-1j * np.sin(phi / 2), np.cos(phi / 2)],
-        ]
+        ],
+        isherm=(phi % (2 * np.pi) <= settings.core["atol"]),
```
BoxiLi, line 195: This is nice but I fear that it may again cause problems with jit. I'm fine just leave it None here. The same for other rotations.

qutip/core/gates.py
```diff
@@ -72,10 +83,11 @@ def cz_gate(*, dtype="csr"):
     result : :class:`.Qobj`
         Quantum object for operator describing the rotation.
     """
-    return qdiags([1, 1, 1, -1], dims=[[2, 2], [2, 2]], dtype=dtype)
+    dtype = dtype or settings.core["default_dtype"] or _data.CSR
+    return qdiags([1, 1, 1, -1], dims=_DIMS_2_QB, dtype=dtype)
```
BoxiLi, line 87: How does qdiags handle the isherms property? Will it check if the input is real? If so, maybe for those we could manually set the isherms and isunitary.

Ericgig: When there is only one diagonal, it sets it automatically. 
There is only one case in our code base where this is not the case (and it's not in the tests...).
Improving tests for it and adding manual flag setting could make a good first issue.

qutip/core/operators.py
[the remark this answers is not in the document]
hodgestar: Sounds okay to me. Users can always override the default with `settings.core["default_dtype"]`.

qutip/core/operators.py
hodgestar: From discussion with @Ericgig: Happy to leave this as is pending further conversations around better default handling (e.g. having both a dense and a sparse default).

qutip/core/operators.py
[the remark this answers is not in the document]
hodgestar: Also happy for this to be left for later. We need to clean up `iscp`, `istp` and `iscptp` too.

qutip/core/qobj.py
```diff
@@ -371,33 +371,33 @@ def to(self, data_type):
...
-        store in the new type.
-
         Parameters
         ----------
-        data_type : type
-            The data-layer type that the data of this :class:`Qobj` should be
-            converted to.
+        data_type : type, str
+            The data-layer type or it's string alias that the data of this
+            :class:`Qobj` should be converted to.
```
hodgestar, line 378:
```suggestion
            The data-layer type or its string alias that the data of this
            :class:`Qobj` should be converted to.
```

qutip/entropy.py
```diff
@@ -372,5 +372,6 @@ def entangling_power(U):
     a = tensor(U, U).dag() * swap13 * tensor(U, U) * swap13
     Uswap = swap() * U
     b = tensor(Uswap, Uswap).dag() * swap13 * tensor(Uswap, Uswap) * swap13
+    print(a.tr(), b.tr())
```
hodgestar, line 375: Debugging print.

hodgestar, line 375:
```suggestion
```

hodgestar (approved):
Other than my two small suggestions, looks good.

qutip/tests/test_entropy.py
```diff
@@ -209,4 +209,6 @@ def test_triangle_inequality_4_qubits(self):
                  np.sin(np.pi*_alpha)**2 / 6, id="SWAP(alpha)"),
 ])
 def test_entangling_power(gate, expected):
+    print("_alpha", _alpha)
+    print("gate", gate)
```
Ericgig, line 213:
```suggestion
    print("_alpha", _alpha)
    print("gate", gate)
```

PR #2391, opened 2024-04-10
Add kron specialisation for sparse and dense mixes.

Ericgig:
**Description**
In general, operation between sparse and dense return dense. This is not ideal for the kron product which output would be sparser than the sparse input. The dispatcher does not have fine control for these cases, (Jake plan to add it but never had the time to get to it), so I added wrapper specialisation for these cases. 

I had to patch `expand_operator` since it used `dense` priority to work.


qutip/core/data/kron.pyx
```diff
@@ -159,6 +177,22 @@ cpdef Dia kron_dia(Dia left, Dia right):
     return out
 
 
+cpdef Dia kron_dia_dense_dia(Dia left, Dense right):
+    # The dispatcher would use kron_dense, but the output is at least as sparse
+    # as the sparse input. Since the dispatcher does not have precise control
+    # on which function to use when the signature is missing. We add
+    # then like this.
```
nwlambert, line 184: just a suggestion here; how about "Since the dispatcher does not have precise control on which function to use when the signature is missing, we specify the output to be the sparse format of the input."  or something like that (the original is a bit opaque)

qutip/core/data/kron.pyx
```diff
@@ -159,6 +177,22 @@ cpdef Dia kron_dia(Dia left, Dia right):
...
+    return kron_dia(left, _to(Dia, right))
+
+
+cpdef Dia kron_dense_dia_dia(Dense left, Dia right):
+    # The dispatcher would use kron_dense, but the output is at least as sparse
+    # as the sparse input. Since the dispatcher does not have precise control
+    # on which function to use when the signature is missing. We add
+    # then like this.
```
nwlambert, line 192: same as above

qutip/core/tensor.py
```diff
@@ -467,4 +467,4 @@ def expand_operator(oper, dims, targets, dtype=None):
     for i, ind in enumerate(rest_pos):
         new_order[ind] = rest_qubits[i]
     id_list = [identity(dims[i]) for i in rest_pos]
-    return tensor([oper] + id_list).permute(new_order)
+    return tensor([oper] + id_list).permute(new_order).to(dtype)
```
nwlambert, line 470: one question;  is there a reason the dtype defaults to CSR if its not specified or set in settings?  why not default to the dtype of `oper`?

nwlambert (approved):

qutip/core/tensor.py
Ericgig: This is what we want to avoid.
If `oper` is dense, once extended, the output will be quite sparse:
```
oper = rand_herm(2, dtype="Dense")
expand_operator(oper, [2, 2, 2, 2, 2], [2])
```
The output will be a 32*32 matrix with 4 non-zeros elements...

nwlambert: ah ok, that's a bit clearer, thanks!

Issue #2315, opened 2024-02-03
Speedup of `_correlation_3op_dm`


theodotk:
### Describe the Issue!

Calculation of 2-time correlation is understandably slow. There are two things that can improve `_correlation_3op_dm(solver, state0, tlist, taulist, A, B, C)` a bit.

1. In my experience `tlist` is usually chosen so that at `tlist[-1]` all correlations are already gone. Calculate them up to  `tlist[-1] + taulist[-1]` is often superfluous. An option can be passed to make last `tau` dependent on `t` to make each subsequent simulation shorter. An example could be something like this (but with more thinking about edge cases):
```python
_correlation_3op_dm(solver, state0, tlist, taulist, A, B, C, shorten_tau=False):
    ...
    for t_idx, rho in enumerate(rho_t):
            t = tlist[t_idx]
            if shorten_tau:
                last_tau_index = sum(np.array(taulist) <= t)
                this_step_taulist = taulist[last_tau_index:]
            else:
                this_step_taulist = taulist + t
            corr_mat[t_idx, :] = solver.run(
                C(t) @ rho @ A(t),
                this_step_taulist,
                e_ops=B
            ).expect[0]
            progress_bar.update()
        progress_bar.finished()
```
2. Even better way would be to make simulations along `taulist` in parallel. It seems that it's possible to have different runs of mcsolve at the same time, so here the same framework could be used.

Ericgig:
Thank you for the suggestions.

(2) It should be easy. We could reuse the `parallel_map` as you said.  I would make a good first contribution.  

(1) We need to be careful about this. 
I can see case where the user think the correlation is gone by `tlist[-1]` but it is not, so it should be clear which output are `0` and which are not computed.
Also if we do that, maybe finer control could be given. We could have a variable `max_t_plus_tau` with default `np.inf`.

Muhammad-Rebaal:
Hi @theodotk @Ericgig !
Hope you are doing well !
Could you please assign this issue to me ? I'd like to solve this one .
Thank You !

Ericgig:
Sure you can have a go at it. We will be waiting for you PR.

ATHARVA-JADHAV-AJ:
[...]

PR #2395, opened 2024-04-16
WIP: Speedup of _correlation_3op_dm #2315 through parallel taulist simulations

valleyofblackpanther:
### Objective
The primary objective was to improve the computational efficiency of the `correlation_2op_1t` function in QuTiP. This function computes two-time correlations for quantum systems, which is computationally intensive, especially for large systems or long time arrays. The goal was to enhance performance by parallelizing these computations.

### Enhancements Overview

1. **Parallel Computing**:
   - **Implementation**: Utilized Python’s `concurrent.futures` module, specifically `ProcessPoolExecutor`, to parallelize the computation of correlation values over different tau values (`taulist`). Each tau value computation is independent of others, making this a suitable case for parallel execution.
   - **Function Splitting**: Created a helper function `compute_single_tau_correlation` to handle the computation for a single tau. This function manages the setup of the quantum system, computes the correlation for its specific tau, and handles any exceptions that may occur, logging them accordingly.
   - **Resource Management**: Used the `cpu_count` from the `os` module to dynamically determine the number of available processors and limited the number of worker processes to avoid overwhelming the host system. This approach helps in balancing load and performance.

2. **Error Handling and Robustness**:
   - Enhanced the resilience of the parallel computation by encapsulating the worker function’s body in a try-except block, ensuring that failures in one computation do not impact the overall execution.
   - Ensured the stability of the test environment by restoring the original function state after the test execution, preventing side effects on subsequent tests.

3. **Testing**:
   - **Extended Test Coverage**: Developed new test cases to validate both the parallel and sequential execution paths of the function. Tests ensure that both modes produce equivalent results, thereby confirming that the parallelization did not introduce computational errors.
   - **Integration with Existing Tests**: Integrated the new tests into the existing pytest framework used for QuTiP, making use of fixtures and parameterization to test different scenarios and system setups effectively.

### Implementation Details

- **Concurrency**: The parallel execution is managed through a pool of processes, where each process is responsible for a subset of the total computations. This method is particularly effective for operations that can be divided into independent, smaller tasks.
- **Compatibility**: Ensured that the enhancements are backward-compatible with existing code by modifying only the internal execution strategy of the `correlation_2op_1t` function and by maintaining the original API signature.
- **Performance Considerations**: The parallel implementation can significantly reduce computation times, particularly for larger systems or longer `taulist` arrays, by leveraging multiple CPU cores concurrently.

### Conclusion

The enhancements to the `correlation_2op_1t` function in QuTiP are designed to harness the capabilities of modern multi-core processors, thereby improving the efficiency of quantum correlation computations. This is still a WIP enhancement.

### DIFF
--- a/qutip/solver/correlation.py
+++ b/qutip/solver/correlation.py
@@ -18,6 +18,8 @@ from .heom.bofin_solvers import HEOMSolver
 
 from .steadystate import steadystate
 from ..ui.progressbar import progress_bars
+from concurrent.futures import ProcessPoolExecutor
+from os import cpu_count
 
 # -----------------------------------------------------------------------------
 # PUBLIC API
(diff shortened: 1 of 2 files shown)


pmenczel:
Hi and thank you for your contribution! I see that this is still a draft, but I wanted to quickly point out qutip's existing `parallel` module: https://qutip.readthedocs.io/en/master/apidoc/functions.html#module-qutip.solver.parallel
The module offers functions which can either execute tasks serially (`serial_map`) or in parallel (`parallel_map`, `loky_pmap`, `mpi_pmap`). For example in the Monte Carlo solver, the `map` option decides which of these functions is used, allowing users to switch easily between serial and parallel execution. These functions also come with error handling already, and they select a default number of processes based on the CPU count. It would be nice to have such a `map` option for the calculation of correlation functions.
Are you planning to make similar modifications also to the other types of correlation function in the same module?

PR #2401, opened 2024-04-23
Fix mistake in doc

PositroniumJS:
**Description**
Fix a mistake in the doc.

### DIFF
--- a/doc/changes/2401.doc
+++ b/doc/changes/2401.doc
@@ -1,0 +1,1 @@
+Correct a mistake in the doc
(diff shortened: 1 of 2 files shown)


Ericgig (commented):
Thank you for the fix.
Could you add a towncrier entry so we don't forget your contribution during new release.

PositroniumJS:
> Thank you for the fix. Could you add a towncrier entry so we don't forget your contribution during new release.

I had forgotten. Thank you, it is added!

Ericgig (approved):
Thank you.

PR #2403, opened 2024-04-23
Improve setting explanation in the guide

### LINKED ISSUE #1524: Document changed and new settings

Ericgig:
**Description**
Completed the guide page of the settings to include missing entries as well as options included in `settings.core` and `settings.compile`. Added `CoreOptions` and `CompileOptions` to the apidoc and completed the docstrings.

I also removed settings that are no longer used anywhere.

Lastly I changed the default for `accept_float` compile option.

Presently, if an args is a float it cannot be updated to a complex:
```
>>> a = qutip.coefficient("A", args={"A": 1.})
>>> a(0, A=1j)
TypeError: must be real number, not complex
```
Creating a new coefficient work would be needed.

This will certainly cause an issue eventually, but updating it to complex automatically would cause some error with brmesolve spectra: (`w > w0`) as complex cannot be compared.

Without types, string are not any better than functions so removing them altogether is not ideal.

So I changed it to use complex when no comparison are done and float otherwise.

**Related issues or PRs**
fix #1524


hodgestar (approved):
Left some documentation suggestions, but otherwise looks good. Happy for this to be merged once you're happy with it and tests pass.

PR #2408, opened 2024-04-24
Faster dm norm in solvers

### LINKED ISSUE #2406: Performance drop in QuTiP 5.0.1 vs QuTiP 4.7.6
### Bug Description

While benchmarking [dynamiqs](https://github.com/dynamiqs/dynamiqs) vs QuTiP, I've noticed a large performance drop of `qutip.mesolve` when going from v4.7.6 to v5.0.1. On the example below, the benchmarks show:
```
# QuTiP 4.7.6
# 1.42 s ± 3.17 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# QuTiP 5.0.1
# 14.5 s ± 2.32 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Found similar behavior on two different CPUs (mac M2, and AMD Ryzen 7).

### Code to Reproduce the Bug

```python
import timeit

import numpy as np
import qutip as qt


def init(
    kappa_2: float = 1.0,
    g_cnot: float = 0.3,
    nbar: float = 4.0,
    num_tsave: int = 100,
    N: int = 16,
):
    # time evolution
    alpha = np.sqrt(nbar)
    gate_time = np.pi / (4 * alpha * g_cnot)
    tlist = np.linspace(0.0, gate_time, num_tsave)

    # operators
    ac = qt.tensor(qt.destroy(N), qt.qeye(N))
    nt = qt.tensor(qt.qeye(N), qt.num(N))

    # Hamiltonian
    H = g_cnot * (ac + ac.dag()) * (nt - nbar)

    # collapse operators
    c_ops = [np.sqrt(kappa_2) * (ac**2 - nbar)]

    # initial state
    plus = (qt.coherent(N, alpha) + qt.coherent(N, -alpha)).unit()
    psi0 = qt.tensor(plus, plus)

    kwargs = {'H': H, 'rho0': psi0, 'tlist': tlist, 'c_ops': c_ops}
    return kwargs

kwargs = init()
%timeit qt.mesolve(**kwargs)
```

### Your Environment

```shell
Numpy Version:      1.25.2
Scipy Version:      1.11.2
Cython Version:     0.29.37
Matplotlib Version: 3.7.2
Python Version:     3.11.4
Number of CPUs:     8
BLAS Info:          OPENBLAS
INTEL MKL Ext:      False
```

Ericgig:
**Description**
The trace norm was used to normalized dm in mesolve and other open solvers.
This is computed as `tr( sqrtm(op @ op.dag()) )` which is quite slow.
For density matrices, this is equivalent to the normal trace.
By using the trace, `mesolve` can be twice as fast.

**Related issues or PRs**
fix #2406

ps. While we check if the input is a dm to set the `isherm` flag, we don't check the liouvillian.
Since we accept any super-operator as Liouvillian, we could wrongly set the flag in some cases.

### DIFF
--- a/qutip/solver/solver_base.py
+++ b/qutip/solver/solver_base.py
@@ -102,7 +102,10 @@ class Solver:
             state = Qobj(data, **self._state_metadata, copy=copy)
 
         if data.shape[1] == 1 and self._options['normalize_output']:
-            state = state * (1 / state.norm())
+            if state._isherm:
+                state = state * (1 / state.tr())
+            else:
+                state = state * (1 / state.norm())
 
         return state
 
--- a/qutip/tests/solver/test_mesolve.py
+++ b/qutip/tests/solver/test_mesolve.py
@@ -206,7 +206,7 @@ class TestMESolveDecay:
     def test_mesolve_normalization(self, state_type):
         # non-hermitean H causes state to evolve non-unitarily
         H = qutip.Qobj([[1, -0.1j], [-0.1j, 1]])
-        H = qutip.sprepost(H, H) # ensure use of MeSolve
+        H = qutip.spre(H) + qutip.spost(H.dag()) # ensure use of MeSolve
         psi0 = qutip.basis(2, 0)
         options = {"normalize_output": True, "progress_bar": None}
 


qutip/solver/solver_base.py
```diff
@@ -102,7 +102,10 @@ def _restore_state(self, data, *, copy=True):
             state = Qobj(data, **self._state_metadata, copy=copy)
 
         if data.shape[1] == 1 and self._options['normalize_output']:
-            state = state * (1 / state.norm())
+            if state._isherm:
```
hodgestar, line 105: This is meant to check whether the state is a density matrix, yes? Surely there must be a better way if this doesn't always work. Also definitely worth a comment if the check remains a bit obscure.

pmenczel:
Are there any use cases where the state is an operator and we want to use norm instead of tr? If no, we could just do `isoper(state)` instead of `state._isherm`.

As a side note, in cases with non-Hermitian "density matrices", it probably does not make much sense to normalize the output in any way. That is, I don't think it matters too much whether we use norm or tr in such a case, because probably neither is intended by the user...

Ericgig:
The only operator we normalise are density matrices, so `isoper` work as well. Propagator computation turn the normalisation off. Other cases should only happen when the user pass an arbitrary super operator. I which case they probably don't want any normalisation on our side.

My issue is that in these cases, we still set the `isherm` flag while not having any ideas what the user is trying to do:
```
>>> H = qt.qeye(2) * 1j
>>> L = qt.spre(H)
>>> qt.mesolve(L, qt.basis(2,1), [0, 1]).final_state
Qobj data =
[[0.+0.j         0.+0.j        ]
 [0.+0.j         1.+1.55740849j]]
```

pmenczel:
> My issue is that in these cases, we still set the `isherm` flag while not having any ideas what the user is trying to do:
> 
> We don't set the flag when the input are super operators? We try to check if the Liouvillian preserve hermiticity (not easy for time dependant operator)? Junk in, junk out?

A problem can also happen if the input is not a superoperator:

```
H = qt.sigmaz() + 1j * qt.sigmax()
initial = qt.ket2dm(qt.basis(2, 0))
qt.mesolve(H, initial, [0, 1], c_ops=[qt.sigmam()]).final_state
> Quantum object: dims=[[2], [2]], shape=(2, 2), type='oper', dtype=Dense, isherm=True
> Qobj data =
> [[ 0.51118445+0.j         -0.11358048+0.24945183j]
>  [ 0.11358048+0.24945183j  0.32517255+0.j        ]]
```

pmenczel (approved):
I think the `isherm` issue is separate from this PR. I created issue https://github.com/qutip/qutip/issues/2410 about that.

PR #2416, opened 2024-05-05
Improve dnorm efficiency in unitary case

owenagnel:
## Preliminaries
The diamond norm [1] is a commonly used metric in quantum information theory for calculating the distance between two quantum channels. It has a number of useful properties making it the gold standard [2] in the field for several applications. In general, a complex semidefinite program is required to calculate the diamond norm. Although elegant, this approach is very inefficient. Unfortunately, no alternative method has been discovered for calculating the diamond norm in the general case of CPTP channels.

However, in the special case where we are trying to calculate the difference between two unitary channels, a very efficient implementation exists. This makes use of an unproved theorem on page 29 of [1]. I have proved this theorem and elaborated an efficient algorithm to calculate the diamond distance between two unitaries as part of my masters thesis. 

The current qutip implementation makes use of the semi-definite program formulation in [3] and only uses a simplified calculation on 2 qubit unitary differences.

The implementation of this novel approach is very simple - the hardest step involves diagonalising a unitary. Although time complexity is still exponential in the number of qubits, this implementation is far more efficient than the more general implementation. The Choi representation of the quantum channel isn't used and there is no need to solve a complicated semi-definite program (meaning I can do away with the `cvxpy` dependency).

## Empirical testing
Results of empirical testing on my machine are reported below

|                                 | 3 qubit | 4 qubit  |
|---------|---------|----------|
| current implementation           | 5.22 s  | 3min 21s |
| hyper-efficient implementation  | 924 µs  | 1.11 ms  |

## Proposition
Given the popularity of the circuit model and unitary-based quantum computation, I believe a very efficient implementation of the diamond distance for unitaries would be incredibly valuable for the research community. Given how simple the change is (current tests already cover the test case), I think it would be a simple and worthwhile addition to qutip.

## Citations

[1] D. Aharonov, A. Kitaev, and N. Nisan, “Quantum circuits with mixed states,” in Proceedings of the thirtieth annual ACM symposium on Theory of computing, pp. 20–30, 1998.
[2] A. Gilchrist, N. K. Langford, and M. A. Nielsen, “Distance measures to compare real and ideal quantum processes,” Physical Review A, vol. 71, no. 6, p. 062310, 2005
[3] J. Watrous, “Simpler semidefinite programs for completely bounded norms,” arXiv preprint arXiv:1207.5726, 2012.

### DIFF
--- a/doc/changes/2416.feature
+++ b/doc/changes/2416.feature
@@ -1,0 +1,2 @@
+Updated `qutip.core.metrics.dnorm` to have an efficient speedup when findind the difference of two unitaries. We use a result on page 19 of 
+D. Aharonov, A. Kitaev, and N. Nisan, (1998).
--- a/doc/changes/2416.feature
+++ b/doc/changes/2416.feature
@@ -1,2 +1,2 @@
-Updated `qutip.core.metrics.dnorm` to have an efficient speedup when finding the difference of two unitaries. We use a result on page 19 of 
+Updated `qutip.core.metrics.dnorm` to have an efficient speedup when finding the difference of two unitaries. We use a result on page 18 of 
 D. Aharonov, A. Kitaev, and N. Nisan, (1998).
--- a/doc/changes/2416.feature
+++ b/doc/changes/2416.feature
@@ -1,2 +1,2 @@
-Updated `qutip.core.metrics.dnorm` to have an efficient speedup when findind the difference of two unitaries. We use a result on page 19 of 
+Updated `qutip.core.metrics.dnorm` to have an efficient speedup when finding the difference of two unitaries. We use a result on page 19 of 
 D. Aharonov, A. Kitaev, and N. Nisan, (1998).
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -440,6 +440,9 @@ def dnorm(A, B=None, solver="CVXOPT", verbose=False, force_solve=False,
 
     The diamond norm SDP is solved by using `CVXPY <https://www.cvxpy.org/>`_.
 
+    If B is provided and both A and B are unitaries, a special optimised
+    case of the diamond norm is used.
+
     Parameters
     ----------
     A : Qobj
@@ -477,24 +480,16 @@ def dnorm(A, B=None, solver="CVXOPT", verbose=False, force_solve=False,
     # for instance, by both pyGSTi and SchattenNorms.jl. (By contrast,
     # QETLAB uses the dual problem.)
 
-    # Check if A and B are both unitaries. If so, then we can without
-    # loss of generality choose B to be the identity by using the
-    # unitary invariance of the diamond norm,
-    #     || A - B ||_♢ = || A B⁺ - I ||_♢.
-    # Then, using the technique mentioned by each of Johnston and
-    # da Silva,
-    #     || A B⁺ - I ||_♢ = max_{i, j} | \lambda_i(A B⁺) - \lambda_j(A B⁺) |,
-    # where \lambda_i(U) is the ith eigenvalue of U.
+    # Check if A and B are both unitaries. If so we can use a trick
+    # discussed in D. Aharonov, A. Kitaev, and N. Nisan. (1998).
+    # In general we can find the eigenvalues of AB⁺ and the distance
+    # d between the origin and compelx hull of these. Plugging
+    # this into 2√1-d² gives the diamond norm.
 
-    # There's a lot of conditions to check for this path.  Only check if they
-    # aren't superoperators.  The difference of unitaries optimization is
-    # currently only implemented for d == 2. Much of the code below is more
-    # general, though, in anticipation of generalizing the optimization.
     if (
         not force_solve
         and B is not None
         and A.isoper and B.isoper
-        and A.shape[0] == 2
     ):
         # Make an identity the same size as A and B to
         # compare against.
@@ -507,17 +502,11 @@ def dnorm(A, B=None, solver="CVXOPT", verbose=False, force_solve=False,
             (A * A.dag() - I).norm() < 1e-6
         ):
             # Now we are on the fast path, so let's compute the
-            # eigenvalues, then find the diameter of the smallest circle
-            # containing all of them.
-            #
-            # For now, this is only implemented for dim = 2, such that
-            # generalizing here will allow for generalizing the optimization.
-            # A reasonable approach would probably be to use Welzl's algorithm
-            # (https://en.wikipedia.org/wiki/Smallest-circle_problem).
+            # eigenvalues, then find the distance between origin and hull
             U = A * B.dag()
             eigs = U.eigenenergies()
-            eig_distances = np.abs(eigs[:, None] - eigs[None, :])
-            return np.max(eig_distances)
+            d = _find_poly_distance(eigs)
+            return 2 * np.sqrt(1 - d**2)  # plug into formula
 
     # Force the input superoperator to be a Choi matrix.
     J = to_choi(A)
@@ -583,3 +572,29 @@ def unitarity(oper):
     """
     Eu = _to_superpauli(oper).full()[1:, 1:]
     return np.linalg.norm(Eu, 'fro')**2 / len(Eu)
+
+
+def _find_poly_distance(eigenvals: np.ndarray) -> float:
+    """Function to find the distance between origin and the
+    convex hull of eigenvalues."""
+    phases = np.angle(eigenvals)
+    pos_max = phases.max()
+    neg_min = phases.min()
+    pos_min = np.where(phases > 0, phases, np.inf).min()
+    neg_max = np.where(phases <= 0, phases, -np.inf).max()
+
+    if neg_min > 0:  # all eigenvals have positive phase, hull is above x axis
+        return np.cos((pos_max - pos_min) / 2)
+
+    if pos_max <= 0:  # all eigenvals have negative phase, hull is below x axis
+        return np.cos((np.abs(neg_min) - np.abs(neg_max)) / 2)
+
+    big_angle = pos_max - neg_min
+    small_angle = pos_min - neg_max
+    if big_angle >= np.pi:
+        if small_angle <= np.pi:  # hull contains the origin
+            return 0
+        else:  # hull is left of y axis
+            return np.cos((2 * np.pi - small_angle) / 2)
+    else:  # hull is right of y axis
+        return np.cos(big_angle / 2)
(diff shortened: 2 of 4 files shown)


owenagnel:
Note to reviewers:
- Code climate is not passing because too many return statements... It's any easy fix but is it worth it?

doc/changes/2416.feature
```diff
@@ -0,0 +1,2 @@
+Updated `qutip.core.metrics.dnorm` to have an efficient speedup when findind the difference of two unitaries. We use a result on page 19 of
```
hodgestar, line 1:
```suggestion
Updated `qutip.core.metrics.dnorm` to have an efficient speedup when finding the difference of two unitaries. We use a result on page 19 of 
```

qutip/core/metrics.py
```diff
@@ -440,6 +440,9 @@ def dnorm(A, B=None, solver="CVXOPT", verbose=False, force_solve=False,
 
     The diamond norm SDP is solved by using `CVXPY <https://www.cvxpy.org/>`_.
 
+    If B is provided and both A and B are unitaries, a special optimised
+    case of the diamond norm is used.
```
hodgestar, line 444: Please add the reference to the papers, etc with the method here too.

hodgestar:
Thanks for tackling this (I think) long-standing too. I haven't done a thorough check but on a first pass it looks like a great improvement.

doc/changes/2416.feature
owenagnel: Will commit this right away

qutip/core/metrics.py
owenagnel: Sounds good!

owenagnel:
Thanks for reviewing! Sounds good, will make those changes ASAP. 

Quick disclaimer, I can't seem to install `cvxpy` on my machine without using anaconda, so I haven't run the qutip tests personally. I assume these run automatically in the pipeline. I have done a lot of testing independently though and the implementation seems to work very reliably (often giving more accurate results than the semidefinite program).

hodgestar:
> Quick disclaimer, I can't seem to install `cvxpy` on my machine without using anaconda, so I haven't run the qutip tests personally. I assume these run automatically in the pipeline.

That's fine. The CI pipeline does test with cvxpy.

owenagnel:
Made another quick change. We know from Prop. 3.44 of (Watrous, 2018) that the diamond norm of CPTP channel is 1. So I added a quick `if` statement that checks this. Also refactored so there's only one return statement to make code climate happy, but seems like it's unhappy with something else and maybe made the logic harder to follow. Please advise if you have a preference.

qutip/core/metrics.py
```diff
@@ -486,73 +489,70 @@ def dnorm(A, B=None, solver="CVXOPT", verbose=False, force_solve=False,
...
-                                  J_dat.indptr),
-                                 shape=J_dat.shape).toarray()
-    else:
-        problem = dnorm_sparse_problem(dim, J_dat)
-
-    problem.solve(solver=solver, verbose=verbose)
-
-    return problem.value
+        and check_unitary(A)
+        and check_unitary(B)
```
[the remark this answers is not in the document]
owenagnel: Thanks! All fixed

Ericgig (approved):
I will let @hodgestar re-review and merge, but looks good to me. 
Thank you.

owenagnel:
> I will let @hodgestar re-review and merge, but looks good to me.
> Thank you.

Thanks for reviewing!

qutip/core/metrics.py
```diff
@@ -583,3 +566,29 @@ def unitarity(oper):
     """
     Eu = _to_superpauli(oper).full()[1:, 1:]
     return np.linalg.norm(Eu, 'fro')**2 / len(Eu)
+
+
+def _find_poly_distance(eigenvals: np.ndarray) -> float:
+    """Function to find the distance between origin and the
+    convex hull of eigenvalues."""
```
hodgestar, line 573:
```suggestion
    """
    Return the distance between the origin and the convex hull of eigenvalues.

    The complex eigenvalues must have unit length (i.e. lie on the circle about the origin).
    """
```

qutip/core/metrics.py
```diff
@@ -583,3 +566,29 @@ def unitarity(oper):
...
+
+    big_angle = pos_max - neg_min
+    small_angle = pos_min - neg_max
+    if big_angle >= np.pi:
+        if small_angle <= np.pi:  # hull contains the origin
+            return 0
+        else:  # hull is left of y axis
+            return np.cos((2 * np.pi - small_angle) / 2)
+    else:  # hull is right of y axis
+        return np.cos(big_angle / 2)
```
hodgestar, line 594:
```suggestion
    phase_max = phases.max()
    phase_min = phases.min()

    if phase_min > 0:  # all eigenvals have positive phase, hull is above x axis
        return np.cos((phase_max - phase_min) / 2)
    if phase_max <= 0:  # all eigenvals have negative phase, hull is below x axis
        return np.cos((np.abs(phase_min) - np.abs(phase_max)) / 2)

    pos_phase_min = np.where(phases > 0, phases, np.inf).min()
    neg_phase_max = np.where(phases <= 0, phases, -np.inf).max()

    big_angle = phase_max - phase_min
    small_angle = pos_phase_min - neg_phase_max

    if big_angle >= np.pi:
        if small_angle <= np.pi:  # hull contains the origin
            return 0
        else:  # hull is left of y axis
            return np.cos((2 * np.pi - small_angle) / 2)
    else:  # hull is right of y axis
        return np.cos(big_angle / 2)
```
I renamed the variables to make it a little clearer to me what they represent and moved the faster paths a bit further up so we can return early (maybe a bit unnecessary since this function is not the slow part).

hodgestar (changes requested):
Thanks again for this. I made some suggestions largely to clarify things.

Could you also add a test to the dnorm tests in `test_metrics` that explicitly tests the unitary case using `rand_unitary` to generate A and B?

owenagnel:
> Please also check my suggestions -- I am definitely not infallible!

Should have addressed all these, all very good suggestions! Thanks for reviewing

doc/changes/2416.feature
```diff
@@ -0,0 +1,2 @@
+Updated `qutip.core.metrics.dnorm` to have an efficient speedup when finding the difference of two unitaries. We use a result on page 19 of
```
hodgestar, line 1:
```suggestion
Updated `qutip.core.metrics.dnorm` to have an efficient speedup when finding the difference of two unitaries. We use a result on page 18 of 
```

qutip/core/metrics.py
```diff
@@ -538,6 +522,9 @@ def dnorm(A, B=None, solver="CVXOPT", verbose=False, force_solve=False,
         # The 2-norm was not implemented for sparse matrices as of the time
         # of this writing. Thus, we must yet again go dense.
         return la.norm(op.full(), 2)
+    if not force_solve and J.iscptp:
```
hodgestar, line 525: Let's do the `iscptp` check before the `iscp` one since it's a less general case and faster to calculate.

hodgestar (commented):
Two tiny suggestions and then it looks good.

hodgestar (approved):

qutip/tests/core/test_metrics.py
```diff
@@ -453,6 +453,15 @@ def test_force_solve(self, dimension, generator):
...
 
+    @pytest.mark.repeat(3)
+    def test_unitary_case(self, dimension):
+        """Check that the diamond norm is one for unitary maps."""
+        A, B = rand_unitary(dimension), rand_unitary(dimension)
+        assert (
+            dnorm(A, B)
+            == pytest.approx(dnorm(A, B, force_solve=True), abs=1e-5)
+        )
+
```
Ericgig, line 464: This is redundant to `test_force_solve`...

hodgestar: Sorry I missed this. Perhaps `force_solve=True` should be left out?

Ericgig: I think the unitary case is already tested enough and no new tests are needed.

hodgestar: I agree to some extent, but it would be nice to have a test explicitly for this case since now it is treated with a separate code path.

owenagnel: Regarding this, I think having a unitary test is probably a good idea since we now do have an explicit optimisation for a difference of unitaries. Regarding `test_force_solve`, I also think it's redundant. To be honest I think we should just get rid of `test_force_solve` as it doesn't really accomplish anything. For the `rand_unitary` generator it's exactly the same as the new `test_unitary_case`. For the `rand_super_bcsz` generator, I'm sort of confused what's being tested in the first place. There is not optimisation for a difference of CPTP channels, the semidefinite program is run whether `test_force_solve` is set to `True` or not (the difference of CPTP channels is not completely positive in general.)

owenagnel: I think we ought to keep `test_unitary_case` and get rid of `test_force_solve`. Perhaps it's also worth writing a test for the completely positive channel optimisation.

hodgestar: +1 on adding a `test_completely_positive_case` or `test_cp_case` and `test_cptp_case`.

I didn't look at `test_force_solve` but I expect it to pass `force_solve=True` while I think `test_unitary_case` should pass `force_solve=False`.

@Ericgig Thoughts?

owenagnel: Great, `test_cptp` already exists, although renaming to `test_cptp_case` for consistency is maybe worthwhile. 

I need to figure out how best to implement `test_cp_case`, any suggestions would be great.

`test_force_solve` actually just compares the result of `dnorm` when  `force_solve` is set to `True` and when it's set to `False`. All `force_solve` does is force `dnorm` to solve the semidefinite program rather than choose an optimisation if it's available, so it's a good baseline to test against.

As things stand, in `test_unitary_case` I compare the result of `dnorm` with unitary difference optimisation to the baseline semidefinite program (`force_solve = True`). I think this is the correct way to run the test in the absence of any analytical formulas.

Ericgig: I am for removing `test_force_solve` and splitting each case in it's own test.
Using `force_solve=True` in `test_unitary_case` and other tests to ensure that the optimised algorithm give the same result as the normal one is the way to go.

owenagnel: Have committed these changes @Ericgig. Let me know if it is worth having a `test_force_solve` case regardless.

Ericgig (approved):

hodgestar (approved):

hodgestar:
@owenagnel Thank you for an excellent contribution. Merged to master.

PR #2421, opened 2024-05-08
Add support for numpy 2

Ericgig:
**Description**
Add support for the up comming numpy v2:

- Allows qutip to be compiled with numpy 2, and pass tests.
- Test that binaries compiled with numpy 2 are compatible with numpy 1 at runtime.

I did not update the requirement yet, it must be installed in `--no-build-isolation`.
When it's officially released, we will need to update `pyproject.toml` and `setup.cfg`, and possibly add test for np2 on windows and mac.

Here are the changes that affect us:
- `np.array(.., copy=False)` no longer mean copy only if needed. `copy=None` was added for that.
- `np.trapz` was replaced by `np.trapezoid`. Since there are no overlaps, I opted to use `scipy.integrate.trapezoid` instead.
- `np.array(1, np.int32) * 2**33` will enforce the integer type and overflow.
- `Py_intptr_t` from `python.h` is no longer available through numpy cython interface.


qutip/core/data/make.py
```diff
@@ -119,7 +119,11 @@ def one_element_dia(shape, position, value=1.0):
     data = np.zeros((1, shape[1]), dtype=complex)
     data[0, position[1]] = value
     offsets = np.array([position[1]-position[0]])
-    return Dia((data, offsets), copy=False, shape=shape)
+    if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+        copy_if_needed = None
+    else:
+        copy_if_needed = False
+    return Dia((data, offsets), copy=copy_if_needed, shape=shape)
```
hodgestar, line 126: Our interface shouldn't depend on the underlying version of numpy, so perhaps this logic needs to live in `Dia.__init__`?

qutip/utilities.py
```diff
@@ -108,7 +108,7 @@ def clebsch(j1, j2, j3, m1, m2, m3):
     C = np.sqrt((2.0 * j3 + 1.0)*_to_long(c_factor))
 
     s_factors = np.zeros(((vmax + 1 - vmin), (int(j1 + j2 + j3))), np.int32)
-    sign = (-1) ** (vmin + j2 + m2)
+    sign = int((-1) ** (vmin + j2 + m2))
```
hodgestar, line 111: Why did this need to be wrapped with `int`?

hodgestar (commented):
Woot for starting to get things ready for numpy 2. Did a quick review of the changes so far. Most look good.

qutip/utilities.py
Ericgig: `clebsch` is computed using large integer over 32 and sometime 64 bits.
If the sign is a 0d numpy array it overflow.

qutip/core/data/make.py
Ericgig: I would prefer to just let the back-ends manage it.
Only this call need fix for our tests, but we use `np.array` close to a hundred times and we have to think about jax, cupy, etc. too. I don't think we should fight to make them all works the same...

qutip/utilities.py
hodgestar: Ah. Perhaps add a comment above this line with that explanation?

qutip/core/data/make.py
hodgestar: By "our interface" I meant just "Dia.__init__" not the dataclass interface as a whole. If we don't change this inside `__init__` then our users also need to do this check if they call `Dia(...)`.

hodgestar: From discussion with @Ericgig: Perhaps we can do something like:
```
if numpy < 2:
    copy = bool(copy)
```
in the `__init__` for CSR, Dense and Dia.

qutip/utilities.py
```diff
@@ -108,7 +108,8 @@ def clebsch(j1, j2, j3, m1, m2, m3):
     C = np.sqrt((2.0 * j3 + 1.0)*_to_long(c_factor))
 
     s_factors = np.zeros(((vmax + 1 - vmin), (int(j1 + j2 + j3))), np.int32)
-    sign = (-1) ** (vmin + j2 + m2)
+    # `S` and `C` are large integer, if `sign` is a np.int32 it could oveflow
```
Ericgig, line 111:
```suggestion
    # `S` and `C` are large integer,s if `sign` is a np.int32 it could oveflow
```

qutip/core/data/make.py
hodgestar: Thanks for implementing, @Ericgig.

hodgestar:
Numpy is maintaining a list of project versions that support numpy 2 -- https://github.com/numpy/numpy/issues/26191. They're waiting of downstream packages to be ready before release. We could add ourselves there if we liked.

.github/workflows/tests.yml
```diff
@@ -37,6 +37,24 @@ jobs:
         # the lack of a variable is _always_ false-y, and the defaults lack all
         # the special cases.
         include:
+          # Test with the version 2 of numpy comming soon.
+          - case-name: numpy2
+            os: ubuntu-latest
+            python-version: "3.12"
+            scipy-requirement: ""
+            numpy-requirement: "==2.0.0rc1"
```
eendebakpt, line 45:
```suggestion
            numpy-requirement: "==2.0.0rc2"
```
There is a new release candidate available

.github/workflows/tests.yml
```diff
@@ -37,6 +37,24 @@ jobs:
...
+            pypi: 1
+
+          # Binaries compiled with numpy 2 should be compatible when using
+          # numpy 1.X at runtime.
+          - case-name: numpy2_to_1
+            os: ubuntu-latest
+            python-version: "3.10"
+            scipy-requirement: ""
+            numpy-requirement: "==2.0.0rc1"
```
eendebakpt, line 54:
```suggestion
            numpy-requirement: "==2.0.0rc2"
```

eendebakpt (commented):
Numpy 2.0 is scheduled to be released June 6th (see https://github.com/numpy/numpy/issues/24300#issuecomment-2125643448)

hodgestar (approved):

PR #2423, opened 2024-05-09
Stochastic solvers small bugfix

pmenczel:
When calling `ssesolve` or `smesolve` without any stochastic collapse operators but `store_measurement` set to True, there is an IndexError. This PR fixes the error and adds a test for this situation.

Example code:
```python
import qutip as qt
r = qt.ssesolve(qt.sigmaz(), qt.basis(2, 1), [0, 1, 2], options={'store_measurement': True}, ntraj=15)
```

### DIFF
--- a/qutip/solver/stochastic.py
+++ b/qutip/solver/stochastic.py
@@ -83,6 +83,11 @@ class StochasticTrajResult(Result):
         """
         if not self.options["store_measurement"]:
             return None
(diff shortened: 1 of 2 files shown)


Ericgig (approved):
Nice catch.

Without `sc_ops` the stochastic solver is simply `me/sesolve`, maybe we should also add a fallback like `mcsolve`.

Issue #2422, opened 2024-05-09
Bug in mcsolve with CollapseFeedback when using serial map


pmenczel:
### Bug Description

When calling `mcsolve` with operators that have a collapse feedback, everything works as it should on the first trajectory. On all other trajectories, the collapse feedback parameter will still contain the collapses of the first trajectory (and only those).

### Code to Reproduce the Bug

```shell
import qutip as qt
import numpy as np

def hamiltonian(t, jumps):
    print(t, ', ', jumps)
    return qt.sigmax() * (len(jumps) + 1)

tlist = np.linspace(0, 10, 500)
H = qt.QobjEvo(hamiltonian, args={'jumps': qt.MCSolver.CollapseFeedback()})

result = qt.mcsolve(H, qt.basis(2, 1), tlist,
                    c_ops=[qt.sigmam()], e_ops=[qt.sigmaz()],
                    ntraj=2, options={'keep_runs_results': True, 'map': 'serial'})
```

### Code Output

```shell
##### FIRST TRAJECTORY

0.0 ,  []
2.109452934261477e-12 ,  []
...
0.6212424849699398 ,  []
0.6086836025648399 ,  [(0.6086836025648399, 0)]
...
1.6032064128256511 ,  [(0.6086836025648399, 0)]
1.6024822039345563 ,  [(0.6086836025648399, 0), (1.6024822039345563, 0)]
...
10.0 ,  [(0.6086836025648399, 0), (1.6024822039345563, 0), (5.151302705262849, 0)]

##### SECOND TRAJECTORY

0.0 ,  [(0.6086836025648399, 0), (1.6024822039345563, 0), (5.151302705262849, 0)]
1.0547264671307385e-12 ,  [(0.6086836025648399, 0), (1.6024822039345563, 0), (5.151302705262849, 0)]
...
10.0 ,  [(0.6086836025648399, 0), (1.6024822039345563, 0), (5.151302705262849, 0)]
```

### Expected Behaviour

On the second trajectory, the list of collapses should have been reset to `[]` and then updated during the run of that trajectory.

### Your Environment

```shell
QuTiP Version:      5.1.0.dev0+d8c6584
Numpy Version:      1.26.3
Scipy Version:      1.12.0
Cython Version:     3.0.8
Matplotlib Version: 3.8.2
Python Version:     3.12.1
Number of CPUs:     20
BLAS Info:          Generic
INTEL MKL Ext:      False
Platform Info:      Linux (x86_64)
```

[...]

PR #2424, opened 2024-05-09
Ensure feedback work for all trajectories

Ericgig:
**Description**
Fix the bug in collapse feedback that made it work properly only for the first trajectories. (#2422)
Also added a check in the feedback test to ensure the collapse argument is reset for each trajectories.

### DIFF
--- a/qutip/core/cy/qobjevo.pyx
+++ b/qutip/core/cy/qobjevo.pyx
@@ -482,7 +482,6 @@ cdef class QobjEvo:
                     f"Desired feedback {key} is not available for the {solver}."
                 )
             new_args[key] = solvers_feeds[feed]
-        # self.arguments(new_args)
 
         if new_args:
             cache = []
--- a/qutip/core/cy/qobjevo.pyx
+++ b/qutip/core/cy/qobjevo.pyx
@@ -482,7 +482,16 @@ cdef class QobjEvo:
                     f"Desired feedback {key} is not available for the {solver}."
                 )
             new_args[key] = solvers_feeds[feed]
-        self.arguments(**new_args)
+        # self.arguments(new_args)
+
+        if new_args:
+            cache = []
+            self.elements = [
+                element.replace_arguments(new_args, cache=cache)
+                for element in self.elements
+            ]
+
+
 
     def _update_feedback(QobjEvo self, QobjEvo other=None):
         """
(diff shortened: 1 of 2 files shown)


qutip/core/cy/qobjevo.pyx
```diff
@@ -482,7 +482,16 @@ cdef class QobjEvo:
                     f"Desired feedback {key} is not available for the {solver}."
                 )
             new_args[key] = solvers_feeds[feed]
-        self.arguments(**new_args)
+        # self.arguments(new_args)
```
hodgestar, line 485: How did calling `.arguments(...)` wipe out the feedback items? What is different about calling `.replace_arguments`?

I guess this commented out line should now be deleted.

Ericgig: It is set so that `argument` can change from a normal args to a feedback and vise-versa. 
```
qevo = QobjEvo(..., args={"A": MESolver.ExpectFeedback(op))
solver = MESolver(qevo, c_ops)
solver.run(...) # Run with feedback
solver.run(..., args={"A": 1.) # Rerun using the same QobjEvo without feedback
solver.run(..., args={"A": MESolver.StateFeedback(rho0)) # Rerun using new feedback
```
Therefore when `argument` is called, it update the `_feedback_functions` and `_solver_only_feedback` attributes to reflect the change between normal and dynamic args.

`replace_arguments` apply the new values of the args on the `coefficients`, but do not change attributes of the `QobjEvo`.

Ericgig:
```suggestion
```

hodgestar: Thanks, that makes sense.

qutip/core/cy/qobjevo.pyx
```diff
@@ -482,7 +482,15 @@ cdef class QobjEvo:
...
-        self.arguments(**new_args)
+
+        if new_args:
+            cache = []
+            self.elements = [
+                element.replace_arguments(new_args, cache=cache)
+                for element in self.elements
+            ]
+
+
```
hodgestar, line 493:
```suggestion
```
Remove extra blank lines.

hodgestar, line 493:
```suggestion
            ]
```
Remove extra blank lines.

hodgestar, line 493: These extra blank lines should be removed but GitHub's suggestion syntax doesn't render the removal of them nicely.

hodgestar (approved):

qutip/core/cy/qobjevo.pyx
Ericgig:
```suggestion
```

Issue #2415, opened 2024-05-03
Same code with different ouputs in qutip 4.7.2 and 5.0.1


CrazyCakes:
### Bug Description

I conducted a quantum gate simulation and calculated the fidelity using QobjEvo and mesolve in qutip 4.7.2. However, when running the same code in qutip 5.0.1, the fidelity results were dramatically different. I've experimented with various environments, but the issue persists.

### Code to Reproduce the Bug

```shell
import numpy as np
import qutip as qt

H=qt.tensor(qt.sigmaz(), qt.sigmaz())

H_x=qt.tensor(qt.qeye(2), qt.sigmax())
def a(t, args):
    return 1
H_y=qt.tensor(qt.qeye(2), qt.sigmay())
def b(t, args):
    return 0


U=qt.tensor(qt.basis(2, 0), qt.basis(2, 0))*qt.tensor(qt.basis(2, 0), qt.basis(2, 0)).dag()+\
  qt.tensor(qt.basis(2, 0), qt.basis(2, 0))*qt.tensor(qt.basis(2, 1), qt.basis(2, 1)).dag()+\
  qt.tensor(qt.basis(2, 1), qt.basis(2, 1))*qt.tensor(qt.basis(2, 0), qt.basis(2, 0)).dag()+\
  qt.tensor(qt.basis(2, 1), qt.basis(2, 1))*qt.tensor(qt.basis(2, 1), qt.basis(2, 1)).dag()

options=qt.Options(store_final_state=True)
Omaga=1/32
T=np.pi/(2*Omaga)
times=np.linspace(0, T, 10)
S=qt.QobjEvo([1/2*H, [Omaga*H_x, a], [Omaga*H_y, b]], tlist=times) 
result=qt.mesolve(S, U, times, options=options)

print((result.final_state*U).tr()/4)
```

### Code Output

```shell
4.7.2:
0.9999903297580207
5.0.1:
0.4999951648790101+1.261294352485955e-19j
```

### Expected Behaviour

The outputs of versions 4.7.2 and 5.0.1 should be identical or at least exhibit minor differences.

### Your Environment

```shell
QuTiP Version:      4.7.2 or 5.0.1
Numpy Version:      1.26.0
Scipy Version:      1.11.4
Cython Version:     None
Matplotlib Version: 3.8.2
Python Version:     3.11.5
Number of CPUs:     16
BLAS Info:          Generic
INTEL MKL Ext:      False
Platform Info:      Windows (AMD64)
```

[...]

PR #2427, opened 2024-05-16
Only normalize if initial state is normalized.

### LINKED ISSUE #2426: mesolve produces unphysical result in qutip v5.0.1 (and not in v4.6)
### Bug Description

When computing the evolution of a density matrix with off-diagonal elements, in presence of decoherence, the matrix is left un-changed in qutip v5.0.1. In qutip v4.6, I obtain the expected behavior: the off-diagonal elements decay.

### Code to Reproduce the Bug

```shell
import qutip as qt
rho_final = qt.mesolve(
                0*qt.sigmaz(), # Hamiltonian set to 0
                rho0=qt.sigmax(), # Initial state with off-diagonal elements
                tlist=[0,1], # a.u. for time
                c_ops=[qt.sigmaz()] # Large amount of decoherence
                ).states[-1]
print(rho_final)
```

### Code Output

```shell
Qobj data =
[[0.  0.5]
 [0.5 0. ]]
```

### Expected Behaviour

The same code in Qutip v4.6 produces the expected outcome:

```
Quantum object: dims = [[2], [2]], shape = (2, 2), type = oper, isherm = True
Qobj data =
[[0.         0.13533571]
 [0.13533571 0.        ]]
```

### Your Environment

```shell
QuTiP Version:      5.0.1
Numpy Version:      1.26.2
Scipy Version:      1.11.4
Cython Version:     0.29.24
Matplotlib Version: 3.4.3
Python Version:     3.9.7
Number of CPUs:     12
BLAS Info:          Generic
INTEL MKL Ext:      False
```

### Additional Context

_No response_

Ericgig:
**Description**
We had a few issues (#2415, #2426) where an initial state that has not a trace of `1` was use and the new normalization of density matrices resulted in wrong results.

This change the normalization to happen only when the initial state is already normalized, expecting that when a non-normalized initial density matrix in used in `mesolve`, etc., it is a conscious choice and the solver should not fight it.

**Related issues or PRs**
fix #2426

### DIFF
--- a/qutip/solver/brmesolve.py
+++ b/qutip/solver/brmesolve.py
@@ -102,7 +102,8 @@ def brmesolve(H, psi0, tlist, a_ops=(), e_ops=(), c_ops=(),
             On `None` the states will be saved if no expectation operators are
             given.
         - | normalize_output : bool
-          | Normalize output state to hide ODE numerical errors.
+          | Normalize output state to hide ODE numerical errors. Only normalize 
+            the state if the initial state is already normalized.
         - | progress_bar : str {'text', 'enhanced', 'tqdm', ''}
           | How to present the solver progress.
             'tqdm' uses the python module of the same name and raise an error
--- a/qutip/solver/floquet.py
+++ b/qutip/solver/floquet.py
@@ -535,7 +535,8 @@ def fsesolve(H, psi0, tlist, e_ops=None, T=0.0, args=None, options=None):
             On `None` the states will be saved if no expectation operators are
             given.
         - | normalize_output : bool
-          | Normalize output state to hide ODE numerical errors.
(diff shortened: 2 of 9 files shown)


hodgestar (approved):

BoxiLi:
I actually have a question about this automatic normalization.

Do we really need this to hide the numerical errors? Wouldn't it be better to just expose the error since it will be proportional to `atol` anyway (and if not, something went wrong)? Which group of users will this auto-normalization benefit?

Ericgig:
We mostly added it to `mesolve` since we always had it for `sesolve`.
Numerical error is not well understood by all users and we can get issues that the solver are wrong because the trace, norm, populations etc. is not constant.

PR #2437, opened 2024-05-27
Monte-Carlo simulations with mixed initial states

pmenczel:
**Description**
Enables use of `mcsolve` and `nm_mcsolve` with mixed initial conditions.

Added functions `run_mixed` to the MCSolver and NonMarkovianMCSolver classes. Instead of an initial pure state, these functions take either a density matrix or a list of `(psi_i, p_i)` where `psi_i` are pure initial states and `p_i` are the weights for the average over the initial states. In the latter case, the `ntraj` parameter can be a list specifying the number of trajectories to be used for each initial state. The trajectories automatically obtain correction weights if the fraction of trajectories starting in `psi_i` does not match `p_i`.

The functions `mcsolve` and `nm_mcsolve` take either a pure state or a density matrix as the initial state, and call either `run` or `run_mixed` of the solver, respectively.

The stochastic solvers also inherit the `run_mixed` functions, but I don't know if there is any use case for that.

The most complicated issue here was the combination of a mixed initial state with "improved sampling". In this case, run_mixed first runs the no-jump trajectories for all initial states (potentially using `parallel_map` etc) and then runs all other trajectories (using `parallel_map` again, this means that there are two progress bars...)

**Todo**
- [X] Please add tests to cover your changes if applicable.
- [x] If the behavior of the code has changed or new feature has been added, please also update the documentation in the `doc` folder, and the [notebook](https://github.com/qutip/qutip-tutorials). Feel free to ask if you are not sure.
- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>`

### DIFF
--- a/qutip/solver/mcsolve.py
+++ b/qutip/solver/mcsolve.py
@@ -6,7 +6,7 @@ __all__ = ['mcsolve', "MCSolver"]
 import numpy as np
 from numpy.typing import ArrayLike
 from numpy.random import SeedSequence
-from ..core import QobjEvo, spre, spost, Qobj, unstack_columns
+from ..core import QobjEvo, spre, spost, Qobj, unstack_columns, qzero_like
 from ..typing import QobjEvoLike
 from .multitraj import MultiTrajSolver, _MultiTrajRHS, _InitialConditions
 from .solver_base import Solver, Integrator, _solver_deprecation
@@ -528,9 +528,25 @@ class MCSolver(MultiTrajSolver):
         """
         Run one trajectory and return the result.
         """
+        jump_prob_floor = integrator_kwargs.get('jump_prob_floor', 0)
+        if jump_prob_floor == 1:
+            # The no-jump probability is one, but we are asked to generate
+            # a trajectory with at least one jump.
+            # This can happen when a user uses "improved sampling" with a dark
+            # initial state, or a mixed initial state containing a dark state.
+            # Our best option is to return a trajectory result containing only
+            # zeroes. This also ensures # that the final multi-trajectory
+            # result will contain the requested number of trajectories.
+            state_qobj = self._restore_state(state, copy=True)
+            zero = qzero_like(state_qobj)
+            result = self._trajectory_resultclass(e_ops, self.options)
+            result.collapse = []
+            for t in tlist:
+                result.add(t, zero)
+            return seed, result
+
         seed, result = super()._run_one_traj(seed, state, tlist, e_ops,
                                              **integrator_kwargs)
-        jump_prob_floor = integrator_kwargs.get('jump_prob_floor', 0)
         if jump_prob_floor > 0:
             result.add_relative_weight(1 - jump_prob_floor)
         result.collapse = self._integrator.collapses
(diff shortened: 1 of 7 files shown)


qutip/solver/mcsolve.py
```diff
@@ -499,13 +507,45 @@ def _initialize_stats(self):
...
         Run one trajectory and return the result.
         """
+        jump_prob_floor = integrator_kwargs.get('jump_prob_floor', 0)
+        if jump_prob_floor == 1:
+            # The no-jump probability is one, but we are asked to generate
+            # a trajectory with at least one jump.
+            # This can happen when a user uses "improved sampling" with a dark
+            # initial state, or a mixed initial state containing a dark state.
+            # Our best option is to return a trajectory result containing only
+            # zeroes. This also ensures # that the final multi-trajectory
```
hodgestar, line 538:
```suggestion
            # zeroes. This also ensures that the final multi-trajectory
```

qutip/solver/mcsolve.py
```diff
@@ -561,6 +598,158 @@ def run(
         result.stats['run time'] = time() - start_time
         return result
 
+    def run_mixed(
```
hodgestar, line 601: Question: Our currently solver classes have a single `run` method (with some exceptions). It would be nice if we could maintain that here. Is it possible to have a `run` method that calls either `run_mixed` or `run_pure`? Or two have two solvers or combine the two methods? Not sure if any of these options are sensible.

pmenczel: I was waiting for this review comment since I wasn't sure how to go about it :) Yes, it would be possible to have a single `run` method. However, some arguments have different types for run_mixed, and run_mixed does not accept the `target_tol` parameter. I thought that might be a bit confusing, but if you think it is better, I will change it to a single method (which would call either `_run_pure` or `_run_mixed` or so).

hodgestar: Calling `_run_pure` and `_run_mixed` sounds good. I guess it depends how different the arguments are. One parameter not accepted doesn't seem like the end of the world. What are the different types?

hodgestar: Thanks for making the change. The big picture structure looks good to me now. I'll leave the detailed review to @Ericgig since he knows `mcsolve` much better than I do.

pmenczel:
Hm I'll have a look at those failing tests next week. But probably just too few times in `tlist`.

pmenczel:
This should be ready for review now.

Ericgig (commented):
Work well.

- Should we stop the user from passing a list of kets to `mcsolve` directly?
  Presently it work but it's not documented (and list `ntraj` don't.)
- It would be nice to have the trajectories per initial state information stored in the result somewhere.
- When `H` is a super operator, it can evolve density matrix directly. I expect it to give the same result as considering it a mixed state and splitting into kets. In that case, dm input states should not use the mixed stats path. And maybe there is a test there, `mcsolve` with Liouvillian evolving a dm vs mixed state.

qutip/solver/mcsolve.py
```diff
@@ -499,51 +503,201 @@ def _initialize_stats(self):
...
+        # whether "improved sampling" is turned on, and whether the initial
+        # state is mixed.
+        if isinstance(state, (list, tuple)):
+            is_mixed = True
+        else:  # state is Qobj, either pure state or dm
+            if isinstance(ntraj, (list, tuple)):
+                raise ValueError('The ntraj parameter can only be a list if '
+                                 'the initial conditions are mixed and given '
+                                 'in the form of a list of pure states')
+            is_mixed = not state.isket
```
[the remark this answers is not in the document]
pmenczel: Oh, I was somehow not aware of that. I think it is right now.

pmenczel:
Thanks for the review!

> * Should we stop the user from passing a list of kets to `mcsolve` directly?
>   Presently it work but it's not documented (and list `ntraj` don't.)
You are right, done. Of course we could also allow the same arguments for `mcsolve` as for `run`, but I thought it might be good to keep the `mcsolve` interface simpler.

> * It would be nice to have the trajectories per initial state information stored in the result somewhere.
I have added that information and checking it in a test.

Ericgig (approved):

Issue #2389, opened 2024-04-10
Gates in qutip-qutip


ajgpitch:
### Problem Description

Now that qutip-qip has moved to another package, gates are not available in the main repo. This is inconvenient for examples such as the control optimization ones, where one would need to install the qutip-qip package just to access say the Hadamard or CNOT for instance.

### Proposed Solution

The easiest solution would be just to copy the `gates.py` file from the QIP package into the main repo. Perhaps might confuse imports in qip? Also, I think that file has func defs that refer to 'controls' that might not make sense in this repo

### Alternate Solutions

Have a file called `operations.py`. Just have a subset of the gates (operations), the most typical named unitaries - sounds less QIP like and would avoid import confusion. I think also `hadamard_transform` could be shortened to `hadamard`, but also include the original name.

BoxiLi:
We have kept many gates in qutip/core/gates. Not all of them but surely for cnot, hadamard etc.

BoxiLi:
It is actually very similar to your solution. We should add this to the migration guide.

ajgpitch:
Thanks @BoxiLi 
Okay, embarrassingly, I was still in my `update-4,x` branch, so hence didn't find the file.

Still, I did try a lot of stuff like `from qutip import hadamard_transform` (in other envs) that didn't work, before I found (with some help) that `from qutip import gates` is a good option. It seems like there was a deliberate choice not to expose the function names individually.

A search for "hadamard_transform" or "gates" does not return anything helpful. Does it need pointing at in the docs build?

BoxiLi:
That seems true, gates are missing from the documentation. This could be a Good first issue. To add them back to the documentation.

alan-nala:
Hi, I would like to contribute to this issue as part of the UnitaryHack2024. I will incorporate basic use of gates module to guide-basics.rst and a table of the available gates.

PR #2441, opened 2024-05-30
Gates in qutip-v5 User Guide

alan-nala:
**Description**

Hi, I would like to contribute to the issue https://github.com/qutip/qutip/issues/2389 as part of the UnitaryHack2024. I have incorporated a basic use of gates module to guide-basics.rst and a table of the available gates, with and example of how to use a Hadamard gate. Let me know if there is anything else that have to be done.

**Related issues or PRs**
 https://github.com/qutip/qutip/issues/2389


BoxiLi:
Hi, thanks for working on this!
- This table looks nice. The first column shouldn't be "Gate Name" but "Gate functions" and the content should be the name of python functions in `qutip.core.gates.py`.
- Each function should also have its own API documentation. You can create a `quantum gates` section under [Manipulation and Creation of States and Operators](https://qutip.readthedocs.io/en/qutip-5.0.x/apidoc/functions.html#manipulation-and-creation-of-states-and-operators) in https://github.com/qutip/qutip/blob/master/doc/apidoc/functions.rst
- For each function, a latex-rendered matrix should be included as a definition of the gate matrix in the docstring, like in https://qutip-qip.readthedocs.io/en/stable/apidoc/qutip_qip.operations.html#qutip_qip.operations.MS. You can copy them from those in qutip-qip, but some definitions are missing I think.
- The name of functions in the table should be clickable and lead to the function API doc, like the one in https://qutip-qip.readthedocs.io/en/stable/apidoc/qutip_qip.operations.html#qutip_qip.operations.X

hodgestar:
Could we also put this into its own section and perhaps just reference it from the basic introduction? And make sure it gets added to the reference documentation if it isn't there already.

BoxiLi:
Good point, I didn't notice that it is under the basic guide section. What about add it as a subsection (Quantum Gates) after [Qubit (two-level) systems](https://qutip.readthedocs.io/en/qutip-5.0.x/guide/guide-states.html#qubit-two-level-systems).

alan-nala:
Hi, thanks for the feedback. I will try to update it taking these comments into account.

The names of the gates in the table are the same that appear in https://github.com/qutip/qutip/blob/master/qutip/core/gates.py,  I can put them in lowercase to exactly match the functions in the code.

I also see that the module "qutip.core.gates" is missing in [ https://qutip.readthedocs.io/en/master/_modules/index.html](url)  so I will also added there.

BoxiLi:
Great! Yes, lowercase is better and it would be best if they link to the actual API documentation, like in the link I provided. And for that indeed `qutip.core.gates` needs to be added to the API doc first.

alan-nala:
Hi, is the first time that I use sphinx and I am having trouble generating the apidoc. I understand that I have to run sphinx-apidoc to generate the .rst files from /qutip but I don't know where to put  them, so they can be detected by autodoc. I tried but every time I run "make html" I get "autodoc: failed to import module".

BoxiLi:
I think you don't need to generate the apidoc manually. To add the API docs, all you need is to add the path `qutip.core.gates` to https://github.com/qutip/qutip/blob/master/doc/apidoc/functions.rst#quantum-objects. `make html` should then automatically read all the functions in the file and generate the API doc. To modify the API doc, e.g. adding new things, one simply edits the docstrings in each of the Python function.

alan-nala:
> I think you don't need to generate the apidoc manually. To add the API docs, all you need is to add the path `qutip.core.gates` to https://github.com/qutip/qutip/blob/master/doc/apidoc/functions.rst#quantum-objects. `make html` should then automatically read all the functions in the file and generate the API doc. To modify the API doc, e.g. adding new things, one simply edits the docstrings in each of the Python function.

Yes, that was the first thing that I have tried but I think there is a problem with the env and the python version. I did the same with qutip-qip docs with no problems.

BoxiLi:
Could you push the code you have right now? We have CI that builds documentation, which will reveal the error you mentioned in more detail. Maybe I can help then.

BoxiLi:
The CI is still running but have you tried to align with other qutip API docs like this?
```
.. automodule:: qutip.core.operators
    :members: charge, commutator, create, destroy, displace, fcreate, fdestroy, jmat, num, qeye, identity, momentum, phase, position, qdiags, qutrit_ops, qzero, sigmam, sigmap, sigmax, sigmay, sigmaz, spin_Jx, spin_Jy, spin_Jz, spin_Jm, spin_Jp, squeeze, squeezing, tunneling, qeye_like, qzero_like
```
I am not sure if autosummary is fully configured in qutip

BoxiLi:
According to the error message
```
WARNING: invalid signature for automodule ('qutip.core.gates\n :members:\n :show-inheritance:\n :imported-members:')
WARNING: don't know which module to import for autodocumenting 'qutip.core.gates\n :members:\n :show-inheritance:\n :imported-members:' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
```
That should be the error source. Lets use
```
.. automodule:: qutip.core.gates
    :members: ...
```
first and see if it works.

alan-nala:
> According to the error message
> 
> ```
> WARNING: invalid signature for automodule ('qutip.core.gates\n :members:\n :show-inheritance:\n :imported-members:')
> WARNING: don't know which module to import for autodocumenting 'qutip.core.gates\n :members:\n :show-inheritance:\n :imported-members:' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
> ```
> 
> That should be the error source. Lets use
> 
> ```
> .. automodule:: qutip.core.gates
>     :members: ...
> ```
> 
> first and see if it works.

Thank you, lets try that.

BoxiLi:
Any new progress on this? The API doc is generated successfully. Although it would be nice to have more detailed description to each functions.

The rendered user guide now reads like
image
with names as strings.

It would be nice to make them clickable object that links to the function signatures. Like the `tensor` in https://qutip.readthedocs.io/en/qutip-5.0.x/guide/guide-tensor.html#example-constructing-composite-hamiltonians

image

alan-nala:
> Any new progress on this? The API doc is generated successfully. Although it would be nice to have more detailed description to each functions.
> 
> The rendered user guide now reads like image with names as strings.
> 
> It would be nice to make them clickable object that links to the function signatures. Like the `tensor` in https://qutip.readthedocs.io/en/qutip-5.0.x/guide/guide-tensor.html#example-constructing-composite-hamiltonians
> 
> image

Thank you for being so clear with the explanation, now I am able to build the html. I have done this update.

BoxiLi (approved):

alan-nala:
Hi, I would like to know if it is complete in order to collect the bounty, the Hackathon finishes tomorrow. Thanks.

BoxiLi:
Thank you! @alan-nala

PR #2445, opened 2024-06-03
Bloch `add_states` color argument dim bug

PositroniumJS:
**Description**
- Fix a dimension problem for the argument `color` of `Bloch.add_states`
  ```python
  import qutip as qt
  bloch.add_states([qt.basis(2,0), qt.basis(2,1)], colors=["crimson", "gold"]) # fails whereas it should work
  bloch.add_states([qt.basis(2,0), qt.basis(2,1)], colors=[["crimson"], ["gold"]]) # what works
  ```
- Clean-up of the code in `Bloch.add_states` as asked in #1913 
- Add `Bloch.add_arc` and `Bloch.add_line` in the guide on Bloch class

### DIFF
--- a/doc/guide/guide-bloch.rst
+++ b/doc/guide/guide-bloch.rst
@@ -29,7 +29,7 @@ Before getting into the details of these objects, we can simply plot the blank B
 .. plot::
     :context:
 
-    b.make_sphere()
+    b.render()
 
 In addition to the ``show`` command, see the API documentation for :class:`~qutip.bloch.Bloch` for a full list of other available functions.
 As an example, we can add a single data point:
@@ -107,6 +107,15 @@ a similar method works for adding vectors:
     b.clear()
     vec = [[1, 0, 0], [0, 1, 0], [0, 0, 1]]
     b.add_vectors(vec)
+    b.render()
+
+You can also add lines and arcs:
+
+.. plot::
+    :context: close-figs
+
+    b.add_line(x, y)
+    b.add_arc(y, z)
     b.render()
 
 Adding multiple points to the Bloch sphere works slightly differently than adding multiple states or vectors.  For example, lets add a set of 20 points around the equator (after calling `clear()`):
@@ -332,7 +341,7 @@ The code to directly generate an mp4 movie of the Qubit decay is as follows ::
       sphere.clear()
       sphere.add_vectors([np.sin(theta), 0, np.cos(theta)], ["r"])
       sphere.add_points([sx[:i+1], sy[:i+1], sz[:i+1]])
-      sphere.make_sphere()
+      sphere.render()
       return ax
 
    ani = animation.FuncAnimation(fig, animate, np.arange(len(sx)), blit=False, repeat=False)
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -1,6 +1,7 @@
 __all__ = ['Bloch']
 
 import os
+from typing import Literal
 
 import numpy as np
 from numpy import (outer, cos, sin, ones)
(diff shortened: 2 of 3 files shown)


PositroniumJS:
Hello @Ericgig,
Do you think that it is ready to merge?

doc/guide/guide-bloch.rst
```diff
@@ -29,7 +29,7 @@ Before getting into the details of these objects, we can simply plot the blank B
 .. plot::
     :context:
 
-    b.make_sphere()
+    b.render()
```
Ericgig, line 32: Why change this?
Both work.

qutip/bloch.py
```diff
@@ -393,12 +393,12 @@ def add_states(self, state, kind='vector', colors=None, alpha=1.0):
 
             if kind == 'vector':
                 if colors is not None:
-                    self.add_vectors(vec, colors=colors[k], alpha=alpha)
+                    self.add_vectors(vec, colors=[colors[k]], alpha=alpha)
```
Ericgig, line 396: Would this break code of those who already implemented the list of list method?
Could we support both, or add a custom error message to update them to the change?

doc/guide/guide-bloch.rst
PositroniumJS: Because of

> We have `render` and `make_sphere` methods that do exactly the same (`make_sphere` calls render) I would vote to deprecate `make_sphere` for QuTiP 5.0.

from issue #1913

qutip/bloch.py
PositroniumJS: It will break the list of list method but [according to the doc,](https://qutip.readthedocs.io/en/qutip-5.0.x/apidoc/classes.html#qutip.bloch.Bloch.add_states), it should have never worked like this.

But I agree that it can be convenient to add a check in the method

PositroniumJS: I have added a check and I have cleaned the function as asked in #1917

PositroniumJS:
@Ericgig I think it is ready to merge. Thank you 🙏

Ericgig (approved):

Ericgig:
Thank you

lukaspostler:
Hello @PositroniumJS ,
I used to specify colors for states on the Bloch sphere as four-tuples. It seems like this is not possible anymore in Qutip 5.1.1.
Do I get it right, that only a list/array of strings with the same length as the argument `state` can be specified for `color`?

Here is a minimal working example that used to work with with Qutip 5.0.4 and does no work anymore with Qutip 5.1.1:
```
import qutip as qt
import numpy as np
from numpy import pi as pi
import matplotlib.cm as cm

state_zero = qt.basis(2, 0)
state_one = qt.basis(2, 1)

angles = np.linspace(0, pi, 100)
colors = cm.rainbow(np.linspace(0, 1, len(angles)))
colors = [tuple(color) for color in colors]
colors = np.asarray(colors)
quantum_states = [np.cos(angle) * state_zero + np.sin(angle) * state_one for angle in angles]

bloch = qt.Bloch()
bloch.add_states(quantum_states, kind='point', colors=colors)
bloch.show()
plt.show()
```

PositroniumJS:
> Hello @PositroniumJS , I used to specify colors for states on the Bloch sphere as four-tuples. It seems like this is not possible anymore in Qutip 5.1.1. Do I get it right, that only a list/array of strings with the same length as the argument `state` can be specified for `color`?

Hello @lukaspostler,
Sorry, I was not aware of this possibility. It requests only a small change [here](https://github.com/qutip/qutip/blob/master/qutip/bloch.py#L409) to work again: replace `if colors.shape != state.shape:` by `if len(colors) != len(state):`

lukaspostler:
Hello @PositroniumJS,
thank you for looking into and fixing this!

PR #2448, opened 2024-06-04
Ensure `_prepare_state` support jax

Ericgig:
**Description**
The check that the initial state was already normalised for normalising output would branch over values, which breaks `jax.jit` support... Moved the option check first.

Also changed the `isoper` check to filter out rectangular operator.
Rectangular state in solver are tolerated, but we don't have tests for it. I did not add any here, but we should 

**Related issues or PRs**
qutip/qutip-jax#49 qutip/qutip-tutorials#100


pmenczel (approved):
Looks good to me, and I think this should fix the issues in qutip-qoc

PR #2452, opened 2024-06-12
Add dispatcher for sqrtm

rochisha0:
**Description**
Add dispatcher for `sqrtm` in `qobj.py`. This will help us to create a similar function in `jax` for `sqrtm` specifically.

This PR creates a dispatcher function for `sqrtm` that uses `scipy.linalg.sqrtm` to compute square root of a matrix


Ericgig (commented):
The dispatcher and the tests are good.
Seems to have some mix up with version in qobj.py...
Could you merge / rebase your branch to fix this.

PR #2455, opened 2024-06-12
Add merge for stochastic and fix multitrajresult issues.

Ericgig:
**Description**

Add capacity to merge stochastic solver results.

While adding it found a few bugs/ issues with the merge and weight that I fixed:

- When one result had trajectories and the other did not, but both stored the states, the merged result had no states.
- When merging 2 results with default ratio, that ratio would depend on whether the `improved_sampling` options was used.
- When the no-jump probability was 1, the total weight as measured by `sum(res.runs_weight)` would not be one.
- A no-jump probability over 1 by numerical error would break the evolution.
- Error in the weight application in the `_target_tolerance_end`.

The last 3 are easy fix.

For the default weight when merging, using the `num_trajectories` instead of `_num_rel_trajectories` is enough.
But I tried the approach of not counting the no-jumb evolution as a trajectory. This has many small advantages:
- All `result.trajectories` are the same, the first few are not special. 
- All seeds are used.
- No confusing on the merging ratio.
- Keeping the absolute and relative weight apart simply part of the merging.
- The progress bar run up to the expected number.

I also tired to simplify the weight management by accumulating per data then split by abs, rel. This makes it easier to add averaged data (nmmc's trace). Seeing the change in lines number in `multitrajresults.py` it seems to be about the same.

I am also thinking of removing the weight from trajectories and making it only a multitraj property:
`multitraj.add((seed, traj, rel_weight))`
`multitraj.add_deterministic(traj, abs_weight)`

The marginal / trace would stay in the trajectory, but not merging those with weight feel simpler to me.

@pmenczel 
This PR larger than I expected, do you want me to break it into smaller ones for review or are you fine like this.
What do you think about splitting the no-jump / abs traj from the normal ones?
What about having the weights in the `add` instead of the trajectories?

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -384,6 +384,7 @@ class Qobj:
     def __radd__(self, other: Qobj | numbers.Number) -> Qobj:
         return self.__add__(other)
 
+    @_require_equal_type
     def __iadd__(self, other: Qobj | numbers.Number) -> Qobj:
         self.data = (self + other).data
         self._isherm = (self._isherm and other._isherm) or None
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -384,6 +384,11 @@ class Qobj:
     def __radd__(self, other: Qobj | numbers.Number) -> Qobj:
         return self.__add__(other)
 
+    def __iadd__(self, other: Qobj | numbers.Number) -> Qobj:
+        self.data = (self + other).data
+        self._isherm = (self._isherm and other._isherm) or None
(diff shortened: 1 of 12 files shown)


pmenczel:
Thank you for these fixes! Sorry, it will take me a bit until I can have a proper look (vacation next week). It would help if you could break the PR down a bit. Perhaps into things related to the general multitrajresult, and things related to the stochastic solvers?

Ericgig:
Most of it merged through #2502, #2474 and #2459.

PR #2457, opened 2024-06-13
Update requirement to include numpy 2

Ericgig:
Last week we added numpy 2 support but did not update the requirements in `setup.cfg`.

This update dependency to build using numpy 2 instead of oldest-supported-numpy.
I also increased the minimum requirements for matplotlib and scipy to version tested.

I added more test using numpy 2, only one tests build and run using numpy 1.X.
I added more windows / mac tests. Using these os for testing numpy versions and no cython.
`cvxopt` (`dnorm`) does not run with numpy 2 yet, so it's only tested with older numpy version.

Fix sqrtm needing copy. This mean that the test intended to run with numpy 2 previouly did not...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,8 +7,8 @@ requires = [
     "cython>=0.29.20,<3.0.0; python_version<='3.9'",
     # See https://numpy.org/doc/stable/user/depending_on_numpy.html for
     # the recommended way to build against numpy's C API:
-    "oldest-supported-numpy",
-    "scipy>=1.8",
+    "numpy>=2.0.0rc2",
+    "scipy>=1.9",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ requires = [
     "cython>=0.29.20,<3.0.0; python_version<='3.9'",
     # See https://numpy.org/doc/stable/user/depending_on_numpy.html for
     # the recommended way to build against numpy's C API:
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -7,8 +7,8 @@ requires = [
     "cython>=0.29.20,<3.0.0; python_version<='3.9'",
     # See https://numpy.org/doc/stable/user/depending_on_numpy.html for
     # the recommended way to build against numpy's C API:
-    "oldest-supported-numpy",
-    "scipy>=1.8",
+    "numpy>=2.0.0",
```
hodgestar, line 10: Is building with the latest numpy2 the correct thing to do here? Or should we build with an old numpy2 like with oldest-supported-numpy for numpy 1.x?

hodgestar (approved):
Looks okay to me. I left one question about whether building with the latest numpy2 is the correct approach now.

pyproject.toml
Ericgig: This is how I understand their migration guide. There is no `oldest-supported-numpy2` and that package has not been updated for the release.

rgommers:
Will the NumPy 2.0 changes land in QuTip 5.1.0, or some other version?

pyproject.toml
hodgestar: Thanks @rgommers !

hodgestar:
@rgommers I believe @Ericgig is planning to release them in 5.0.3.

rgommers:
Great, thanks!

PR #2459, opened 2024-06-18
Multi Trajectory solver bugfixes

Ericgig:
**Description**
Fix issues with multi- trajectory results:
- When one result had trajectories and the other did not, but both stored the states, the merged result had no states.
- When merging 2 results with default ratio, that ratio would depend on whether the `improved_sampling` options was used. 
- When the no-jump probability was 1, the total weight as measured by sum(res.runs_weight) would not be one.
- A no-jump probability over 1 by numerical error would break the evolution.
- Fix weight application in the `_target_tolerance_end`. (At `abs_weight == 1`, the error should be 0).
- `average_final_state` not being computed properly when runs were stored.

This PR is a part of #2455. (1/4?)
The absolute weighted trajectories are still mixed with the other. (#2455 fixed the issues by not counting the no-jump evolution are a trajectory.)

Tests need #2457 to pass.

### DIFF
--- a/qutip/solver/multitrajresult.py
+++ b/qutip/solver/multitrajresult.py
@@ -337,13 +337,16 @@ class MultiTrajResult(_BaseResult):
             # We only include traj. without abs. weights in this calculation.
             # Since there are traj. with abs. weights., the weights don't add
             # up to one. We have to consider that as follows:
-            #   <(x - <x>)^2> / <1> = <x^2> / <1> - <x>^2 / <1>^2
+            # err = (std * <w>**2 / (N-1)) ** 0.5
+            # avg = <x * w>
+            # avg2 = <x**2 * w>
+            # std * <w>**2 = (<x**2> - <x>**2) * <w>**2
+            #              = avg2 * <w> - avg**2
             # and "<1>" is one minus the sum of all absolute weights
             one = one - self._total_abs_weight
 
-        std = avg2 - abs(avg)**2
-        target_ntraj = np.max(std / target**2) * one**2 + 1
-
+        std = avg2 * one - abs(avg)**2
+        target_ntraj = np.max(std / target**2) + 1
         self._estimated_ntraj = min(target_ntraj - self._num_rel_trajectories,
                                     self._target_ntraj - self.num_trajectories)
         if self._estimated_ntraj <= 0:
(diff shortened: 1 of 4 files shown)


qutip/solver/multitrajresult.py
```diff
@@ -337,13 +337,16 @@ def _target_tolerance_end(self):
...
+            # std * <w>**2 = (<x**2> - <x>**2) * <w>**2
+            #              = avg2 * <w> - avg**2
             # and "<1>" is one minus the sum of all absolute weights
             one = one - self._total_abs_weight
 
-        target_ntraj = np.max((avg2 / one - (abs(avg) ** 2) / (one ** 2)) /
-                              target**2 + 1)
-
+        std = avg2 * one - abs(avg)**2
+        target_ntraj = np.max(std / target**2) + 1
```
pmenczel, line 349: Thank you for fixing this calculation. I did my best to understand how the target tolerance calculation works, but I don't think I really understood it. (That is, even without weights.)

If there is a reference explaining the target tolerance calculation, it might be useful linking to it in the users guide.

In line 345, should that `"<1>"` be a `"<w>"` now?

qutip/solver/multitrajresult.py
```diff
@@ -635,8 +656,7 @@ def merge(self, other, p=None):
 
         where p is a parameter between 0 and 1. Its default value is
         :math:`p_{\textrm{def}} = N / (N + N')`, N and N' being the number of
-        trajectories in the two result objects. (In the case of weighted
-        trajectories, only trajectories without absolute weights are counted.)
+        trajectories in the two result objects.
```
pmenczel, line 659: Of course, there is no "right" or "wrong" for the default value of `p`, we can choose what we think is most useful.

Why I thought it useful to only count the relative trajectories here: when `p` is not specified, I wanted every randomly generated trajectory to have the same importance. In my eyes that means that they should not pick up weights. If the value is `p = p_equal`, then they do not pick up weights.

If you think that `p = self.num_trajectories / new.num_trajectories` is better, I am okay with that; I just wanted to briefly explain why I chose the other value. Your choice is maybe less surprising to users.

qutip/solver/multitrajresult.py
```diff
@@ -804,6 +846,11 @@ def _initialize_sum_states(self, example_trajectory):
         self.sum_states = [
             qzero_like(_to_dm(state)) for state in example_trajectory.states]
 
+    def _initialize_sum_finalstate(self, example_trajectory):
+        self.sum_final_state = qzero_like(
+            _to_dm(example_trajectory.final_state)
+        )
```
pmenczel, line 852: Thank you for sorting out this final state mess. I would suggest changing also lines 833-834 into
```
        if example_trajectory.final_state and store_final_state:
            self._initialize_sum_finalstate(example_trajectory)
```

pmenczel (commented):
Thank you for splitting up the PR, and thank you for the bug fixes. I am sorry for missing those edge cases in the tests!

qutip/solver/multitrajresult.py
Ericgig: If we don't count the trajectory with no jump, then there is no confusion. I believe that would be the best way to do it.

pmenczel (approved):

PR #2460, opened 2024-06-18
Add inplace operators for Qobj

Ericgig:
**Description**
Add inplace operations `+=`, `-=`, `*=`, `@=`, `/=` for Qobj.
These operations could already be used as python fallback on normal operator.
But cases like:
```
for op in operators:
    op *= 2
```
in place operations are needed.

Added a tests and checked cases where in place operations were used on users input.

### DIFF
--- a/qutip/core/gates.py
+++ b/qutip/core/gates.py
@@ -778,7 +778,7 @@ def _powers(op, N):
     yield acc
 
     for _ in range(N - 1):
-        acc *= op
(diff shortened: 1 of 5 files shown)


hodgestar:
I'm -1 on allowing Qobjs to be mutated like this. It's a big footgun for ourselves and users and the benefit of being able to mutate list items in place seems small.

PR #2461, opened 2024-06-19
add support for jit and grad

rochisha0:
**Description**
This PR adds backend support to `metrics.py`. This change will help us to change backend to `jax.numpy`

**Related issues or PRs**
A PR would be created to add the changes to `qutip-jax`

**Note**
`'fidelity', 'tracedist', 'bures_dist', 'bures_angle', 'hellinger_dist', 'hilbert_dist'` are tested and working with `jit`.
`'average_gate_fidelity', 'process_fidelity', 'unitarity', 'dnorm'` are not working as of now. I am finding some issues with testing them.

Only `trace_dist` and `hilbert_dist` are working with `grad` as there is no support for `sqrtm` in grad.

### DIFF
--- a/doc/changes/2461.feature
+++ b/doc/changes/2461.feature
@@ -1,0 +1,1 @@
+Add support for `jit` and `grad` in qutip.core.metrics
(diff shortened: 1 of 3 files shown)


Ericgig (approved):
Good to see we can `jit` and `grad` with little change.
Ideally we don't want to read the setting in every qutip function, so lets merge this into `dev.major` until we have a cleaner way to do this.

PR #2462, opened 2024-06-19
add dynamic backend selection for entropy.py

rochisha0:
**Description**
This PR introduces support for different numerical backends in entropy.py. By leveraging dynamic backend selection, we enable compatibility with both NumPy and JAX

**Related issues or PRs**
This PR builds upon the groundwork laid in PR #2461, further extending the capability of dynamic backend selection.

### DIFF
--- a/qutip/entropy.py
+++ b/qutip/entropy.py
@@ -2,8 +2,8 @@ __all__ = ['entropy_vn', 'entropy_linear', 'entropy_mutual', 'negativity',
            'concurrence', 'entropy_conditional', 'entangling_power',
            'entropy_relative']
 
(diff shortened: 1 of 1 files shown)


qutip/entropy.py
```diff
@@ -113,11 +116,11 @@ def concurrence(rho):
     evals = rho_tilde.eigenenergies()
 
     # abs to avoid problems with sqrt for very small negative numbers
-    evals = abs(sort(real(evals)))
+    evals = abs(np.sort(np.real(evals)))
```
Ericgig, line 119: Should the `abs` use numpy's?

rochisha0: It wasn't using earlier and doesn't cause any issue with `jax`

PR #2466, opened 2024-06-21
Fixed Dicke Entropy returns -Inf #1919

### LINKED ISSUE #1919: Dicke Entropy returns -Inf
### Bug Description

This issue originates from a [problem in the Help Group](https://groups.google.com/g/qutip/c/F7tOoZzMjto).

Calculating the entropy of a density matrix in the Dicke basis leads to -Inf, while the standard calculation with "entropy_vn" returns a finite (more logical) result. 
I already tracked down the problem to the calculation of eigenvalues in `dicke_trace_function`. Due to numerical behaviour some eigenvalues have very small negative values, which lead to -Inf entropy. If these values are rounded to zero, the problem disappears. However, I am not sure, which way is the best way to perform this rounding. 
@nathanshammah @quantshah 

### Code to Reproduce the Bug

```shell
from qutip import *
from qutip.piqs import *
import numpy as np
import numpy.random
from scipy.linalg import eig
from scipy.sparse import block_diag

N = 10
system = Dicke(N)
[jx, jy, jz] = jspin(N,basis="dicke")
jx=jx*2
jy=jy*2
jz=jz*2

V=3
g=1
system.emission = 1
system.hamiltonian = V/(2*(N-1)) * jz*jz + g/2 * jx
D_tls = system.liouvillian()
rho_ss = steadystate(D_tls,method='eigen',)
vn_dicke=entropy_vn_dicke(rho_ss)
vn_dicke_over=entropy_vn(Qobj(block_diag(dicke_blocks_full(rho_ss))))

print(vn_dicke,vn_dicke_over)
```

### Code Output

```shell
-inf 0.006061297461844266
```

### Expected Behaviour

The value for the entropy should not depend on the way it is calculated. I.e., we should have `vn_dicke == vn_dicke_over` up to numerical uncertainty.

### Your Environment

```shell
uTiP Version:      5.0.0.dev0+ee51e50
Numpy Version:      1.21.0
Scipy Version:      1.6.2
Cython Version:     0.29.23
Matplotlib Version: 3.3.4
Python Version:     3.8.10
Number of CPUs:     4
BLAS Info:          OPENBLAS
OPENMP Installed:   False
INTEL MKL Ext:      False
```

### Additional Context

_No response_

pyukey:
**Description**
Dicke density state matrix should only have positive eigenvalues, but rounding errors causes eigenvalues of 0 to instead be represented as a very small negative number.
This causes `dicke_trace_function` to return -Inf, since `scipy.special.entr` returns -Inf for any values < 0.
This issue was fixed by simply taking the absolute value of the eigenvalues in dicke_trace_function, which we are allowed to do since the density state matrix should only have positive eigenvalues.

**Related issues or PRs**
fix #1919

### DIFF
--- a/doc/changes/2466.bugfix
+++ b/doc/changes/2466.bugfix
@@ -1,4 +1,1 @@
-Fixed issue #1919
-Dicke density state matrix should only have positive eigenvalues, but rounding errors causes eigenvalues of 0 to instead be represented as a very small negative number.
-This causes dicke_trace_function to return -Inf, since scipy.special.entr returns -Inf for any values < 0.
-This issue was fixed by simply taking the absolute value of the eigenvalues in dicke_trace_function.
+Fixed rounding error in dicke_trace_function that resulted in negative eigenvalues.
--- a/doc/changes/2466.bugfix
+++ b/doc/changes/2466.bugfix
@@ -1,4 +1,4 @@
 Fixed issue #1919
-Dicke density state matrix should only have positive eigenvalues, but rounding errors causes eigenvalues of 0 to isntead be represented as a very small negative number.
+Dicke density state matrix should only have positive eigenvalues, but rounding errors causes eigenvalues of 0 to instead be represented as a very small negative number.
 This causes dicke_trace_function to return -Inf, since scipy.special.entr returns -Inf for any values < 0.
 This issue was fixed by simply taking the absolute value of the eigenvalues in dicke_trace_function.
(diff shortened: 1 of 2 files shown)


doc/changes/2466.bugfix
```diff
@@ -0,0 +1,4 @@
+Fixed issue #1919
+Dicke density state matrix should only have positive eigenvalues, but rounding errors causes eigenvalues of 0 to instead be represented as a very small negative number.
+This causes dicke_trace_function to return -Inf, since scipy.special.entr returns -Inf for any values < 0.
+This issue was fixed by simply taking the absolute value of the eigenvalues in dicke_trace_function.
```
Ericgig, line 4: The towncrier entry is used to create the changelog. It does not need technical details or reference to issues numbers.

Ericgig (commented):
The code change look good.
A shorter towncrier entry would be better.

doc/changes/2466.bugfix
pyukey: Thank you for the comment! Hopefully the towncrier entry is good now

Ericgig (approved):

Ericgig:
Thank you.

PR #2467, opened 2024-06-24
Plot arcs after front wireframe in Bloch

### LINKED ISSUE #2444: Arc color on Bloch sphere is changed by front wireframe
### Bug Description

When plotting an arc on the Bloch sphere over the front wireframe, it becomes darker and is less visible
example_old_order

It happens because front wireframe is plotted last, if I change the order [here](https://github.com/qutip/qutip/blob/master/qutip/bloch.py#L681) to:
```python
        self.plot_front()
        self.plot_arcs()
```
then I get something looking better:
example_new_order

Should we apply this change or even plot front and back before all the other elements?

### Code to Reproduce the Bug

```shell
import qutip as qt

bloch = qt.Bloch()
xplus = (qt.basis(2, 0) + qt.basis(2, 1)).unit()
xmoin = (qt.basis(2, 0) - qt.basis(2, 1)).unit()
yplus = (qt.basis(2, 0) + 1j * qt.basis(2, 1)).unit()
ymoin = (qt.basis(2, 0) - 1j * qt.basis(2, 1)).unit()
bloch.add_arc(xmoin, ymoin, fmt="crimson")
bloch.add_arc(ymoin, xplus, fmt="crimson")
bloch.add_arc(xplus, yplus, fmt="gold")
bloch.add_arc(yplus, xmoin, fmt="gold")
bloch.save("example.png")
```

### Code Output

_No response_

### Expected Behaviour

The color of the arc should not depend on its position on the Bloch sphere

### Your Environment

```shell
QuTiP Version:      5.0.2
Numpy Version:      1.26.4
Scipy Version:      1.13.1
Cython Version:     None
Matplotlib Version: 3.9.0
Python Version:     3.11.6
Number of CPUs:     4
BLAS Info:          Generic
INTEL MKL Ext:      False
```

### Additional Context

_No response_

PositroniumJS:
**Description**
Plot arcs after front wireframe in Bloch

**Related issues or PRs**
Fix #2444


PositroniumJS:
@Ericgig I have followed what stated in the issue!

PositroniumJS:
@Ericgig happy new year!

Ericgig:
@PositroniumJS 
There was some discussion about this.  
With this, the arcs that are along the back of the sphere appear in front of the front grid, so it's exchanging one annoyance for another and not everyone agrees it's better.  Ideally we should cut the arcs on the back and front and plot them separately, but it's not clear how easy it is to implement and if it's too much to ask from you...

PositroniumJS:
> @PositroniumJS There was some discussion about this. With this, the arcs that are along the back of the sphere appear in front of the front grid, so it's exchanging one annoyance for another and not everyone agrees it's better. Ideally we should cut the arcs on the back and front and plot them separately, but it's not clear how easy it is to implement and if it's too much to ask from you...

I have updated my code to fix that and an equator problem also related (on back wirefram, on equator was from -x to +x instead of -z to +z).

However, I have found a weird bug for a special initial and final states (not related to my changes), I do not know if we open an issue especially for that and merge this PR

```python
import numpy as np
import qutip as qt
start = [np.cos(np.pi/3), np.sin(np.pi/3), 0]
end = [np.cos(-2*np.pi/3), np.sin(-2*np.pi/3), 0]
b = qt.Bloch()
b.add_arc(end, start)
b.save(dpin=200)
```

With previous code
bloch_0

With my code
bloch_bug

Ericgig:
When the line between the start and end passes through the origin, their is no unique arc.
There is a check that this points are not on the opposite, but it's not sensible enough to detect your case.

qutip/bloch.py
```diff
@@ -569,8 +569,29 @@ def add_arc(self, start, end, fmt="b", steps=None, **kwargs):
...
         # Normalize all the points in the line so that are distance len1 from
         # the origin.
-        arc = line * len1 / np.linalg.norm(line, axis=0)
-        self._arcs.append([arc, fmt, kwargs])
+        arc = (line * len1 / np.linalg.norm(line, axis=0)).T
+
+        # We separate the arc between what is in front and what is in back
+        front = arc[0][0] >= 0  # is it a front point?
+        part = []
+        for point in arc:
```
Ericgig, line 577: Not sure if this is the right way.
The camera is not aligned with an axis so the check seems inexact.
Also for arcs that are not on the surface of the sphere, but inside the sphere, should not be on front...

Doing it right seems quite complex with matplotlib limited 3d tools.
For now I would propose to simply add a `layer` option and let the user decide if it should be printed in back on front.

PositroniumJS: > Not sure if this is the right way. The camera is not aligned with an axis so the check seems inexact. 
What matters is whether we plot it before or after the back and front wireframes, not relative to the camera, but relative to which line has to cover which. Since the two half-spheres are separated by the yOz plane, we know whether it is in front or not depending on the x coordinate.

> Also for arcs that are not on the surface of the sphere, but inside the sphere, should not be on front...
Right, I can add a condition on the norm of start/end point (<1, ==1 or >1) 

> Doing it right seems quite complex with matplotlib limited 3d tools. For now I would propose to simply add a `layer` option and let the user decide if it should be printed in back on front.
It is not possible to rotate the Bloch sphere, is it? Otherwise we would also have to change the `plot_back()` and `plot_front()` functions. We would just have to rotate the axis Ox according to the rotation and look for the coordinates of the points along this new axis, but I think it goes further than this issue/PR

Ericgig: The `w` you added to `plot_front` isn't to cut the wire frame more in line with the camera than simply the yz plane? 

Yes a condition for `norm >=1` could work.

There is a `view` parameter in `__init__` to rotate the camera, but the layer stacking is not updated according to it.

Ericgig: Yes, please add the check for the norm.
There is a lot that do not consider the `view` parameter, you can ignore it.

PositroniumJS: I am done, I had to create a lot of subfunctions to reduce the "Cognitive complexity" of the function and satisfy Code Climate

doc/changes/2467.bugfix
```diff
@@ -0,0 +1 @@
+Fix #2444
```
Ericgig, line 1:
```suggestion
Fix arc behind hidden behind the wireframe. 
```

Ericgig (approved):
Thank you.

PR #2473, opened 2024-06-26
Types.creators

Ericgig:
**Description**
Add type hints to function that create `Qobj` (`states.py`, `operators.py`, `gates.py` and `random_object.py`).
I also ran `mypy` on these files and fixed easy issues (`Qobj * Qobj` -> `Qobj @ Qobj`).

Some points to look at more closely: 
- We have a lot of `int, list of int, list of list of int, Space` etc. But what is actually accepted can change:
  - `basis` can take super operators dimension to create an `operator-ket`, but not if `offset` is used.
    Should we raise an error and officially not support these case or arrange it so offset is `supported`?
  - Is `SpaceLike` an alias useful for these case or just confusing?
- Array of Qobj output have been changed to list of Qobj.
- I did not update most docstring. `qdiags` dims are `DimensionLike` in type hints, but `list` in the docstring.
- `jmat` return a single `Qobj` or a list of `Qobj`... For typing, it would be nicer if it always returned a single operator. There is also no advantage to computing the operators together.
- `numbers.Number` is not a valid type according to mypy etc. (`Qobj.__mul__` is always raising errors...)

### DIFF
--- a/qutip/core/gates.py
+++ b/qutip/core/gates.py
@@ -169,7 +169,7 @@ def ct_gate(*, dtype: LayerType = None) -> Qobj:
     )
 
 
-def rx(phi, *, dtype: LayerType = None) -> Qobj:
+def rx(phi: float, *, dtype: LayerType = None) -> Qobj:
     """Single-qubit rotation for operator sigmax with angle phi.
 
     Parameters
--- a/qutip/core/gates.py
+++ b/qutip/core/gates.py
@@ -819,7 +819,7 @@ def qubit_clifford_group(*, dtype: LayerType = None) -> list[Qobj]:
 
     X = sigmax()
     S = phasegate(np.pi / 2)
-    E = H @ (S**3) @ w**3
+    E = H @ (S**3) * w**3
 
     # partial(reduce, mul) returns a function that takes products
     # of its argument, by analogy to sum. Note that by analogy,
--- a/qutip/core/gates.py
+++ b/qutip/core/gates.py
@@ -8,6 +8,7 @@ from . import Qobj, qeye, sigmax, fock_dm, qdiags, qeye_like
 from .dimensions import Dimensions
 from .. import settings
 from . import data as _data
+from ..typing import LayerType
 
 
 __all__ = [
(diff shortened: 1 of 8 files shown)


qutip/typing.py
```diff
@@ -35,3 +35,14 @@ def __call__(self, t: Real, **kwargs) -> Number:
...
+
+SpaceLike = Union[int, list[int], list[list[int]], "Space"]
+
+
+DimensionLike = Union[
+    list[list[int], list[int]],
+    list[list[list[int]], list[list[int]]],
+    list["Space", "Space"],
+    "Dimensions",
+]
```
hodgestar, line 48:
```suggestion
DimensionLike = Union[
    list[SpaceLike, SpaceLike],
    "Dimensions",
]
```
Would this work too? I'm not completely sure it's better since the reader now has to look up what SpaceLike is but on the other hand it makes the relationship between SpaceLike and DimensionLike clear.

qutip/core/gates.py
```diff
@@ -168,7 +169,7 @@ def ct_gate(*, dtype=None):
     )
 
 
-def rx(phi, *, dtype=None):
+def rx(phi, *, dtype: LayerType = None) -> Qobj:
```
hodgestar, line 172:
```suggestion
def rx(phi: float, *, dtype: LayerType = None) -> Qobj:
```
I guess we should also add types to the other arguments if we're doing that for some gates?

qutip/core/gates.py
```diff
@@ -590,16 +591,14 @@ def sqrtiswap(*, dtype=None):
...
-def molmer_sorensen(theta, *, dtype=None):
+def molmer_sorensen(theta: float, *, dtype: LayerType = None) -> Qobj:
     """
     Quantum object of a Mølmer–Sørensen gate.
 
     Parameters
     ----------
     theta: float
         The duration of the interaction pulse.
-    N: int
```
hodgestar, line 601: Woops. Good catch.

hodgestar (approved):

qutip/core/gates.py
Ericgig: Yes, I forgot those. Adding them.

hodgestar (approved):

PR #2476, opened 2024-06-28
Adapt to download / upload artifacts v4

Ericgig:
**Description**
`actions/upload-artifact@v4` and `actions/download-artifact@v4` have breaking change cause the 5.0.3 release to miss wheels on pypi.

Different jobs in one workflow can't create artifacts with the same name. (but not on retry).
Download artifact only download artifact from one job without the proper options.
(https://github.com/actions/upload-artifact/issues/478)

- Upload to different artifacts per OS.
- Merge artifacts when downloading.
- Add test that all wheels and source are available before pypi release.
- For test build, built artifacts are listed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,7 +139,7 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
-          name: wheels
+          name: wheels-${{ matrix.os }}
           path: ./wheelhouse/*.whl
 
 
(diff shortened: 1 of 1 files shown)


AGaliciaMartinez:
Sounds good to me, perhaps a few comments on if statements explaining why they are there can help as it took me some thinking to figure out.

AGaliciaMartinez (approved):

AGaliciaMartinez:
The checks fail because of towncrier requiring a changelog. Should we add it before merging for this PR?

PR #2477, opened 2024-06-28
Adapt to download / upload artifacts v4

Ericgig:
**Description**
`actions/upload-artifact@v4` and `actions/download-artifact@v4` have breaking change cause the 5.0.3 release to miss wheels on pypi.

Different jobs in one workflow can't create artifacts with the same name. 
Download artifact only download artifact from one job without the proper options.
(https://github.com/actions/upload-artifact/issues/478)

I also updated cibuildwheel version, we did it in qutip-5.0.X to build wheels for py3.12, but did not push it yet to master.

I also merged the jobs to publish or pypi and testpypi, only some environment variables changed and they could get out of synchronization.

Lastly I added a check that all wheels are available before uploading to pypi. With 5.0.3's first try, only linux's wheel were published,

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,7 +111,7 @@ jobs:
       CIBW_BUILD: "cp3{10,11,12}-*"
       # Numpy and SciPy do not supply wheels for i686 or win32 for
       # Python 3.10+, so we skip those:
-      CIBW_SKIP: "*-musllinux* cp3{10,11,12}-manylinux_i686 cp3{10,11,12}-win32"
+      CIBW_SKIP: "*-musllinux* *-manylinux_i686 *-win32"
       OVERRIDE_VERSION: ${{ github.event.inputs.override_version }}
 
     steps:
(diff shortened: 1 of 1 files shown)


Ericgig:
@AGaliciaMartinez I will add comment here in the for future release.

.github/workflows/build.yml
```diff
@@ -177,13 +186,29 @@ jobs:
...
+            export TWINE_REPOSITORY=pypi
+            export TWINE_PASSWORD=${{ secrets.PYPI_TOKEN }}
+          else
+            # Exit without deploying
+            echo "Don't update wheels"
+            exit 0
+          fi
+          echo "Deploy to $TWINE_REPOSITORY"
+          echo "python -m pip install "twine""
+          echo "python -m twine upload --verbose wheels/*.whl sdist/*.tar.gz"
```
AGaliciaMartinez, line 206: The pipeline seems to still be faling. based on the error message and the previous check I think we need to change `sdist` to `wheels`

AGaliciaMartinez:
Hey @Ericgig, if I understand correctly, you want to remove TestPyPi in this PR. Is that correct?

Ericgig:
No, if `test` is used for the branch name, it will still deploy to testpypi. Both `deploy to ...` jobs were cut and pasted with only 2 lines changed so I merged them so we don't forget to update one without the other. I also added more feedback to the job that would show uploaded files even for a dry run.

hodgestar (approved):

PR #2484, opened 2024-07-08
fix expm, cosm, sinm to work with jax

rochisha0:
**Description**
In accordance with our ongoing effort to make `qobj` work with `jax`. This change makes `expm`, `cosm`, `sinm` work with `jax`.

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -822,6 +822,11 @@ class Qobj:
         """
         if not self._dims.issquare:
             raise TypeError("expm is only valid for square operators")
+        if isinstance(self.data, _data.CSR):
+            return Qobj(_data.expm(self._data, dtype=dtype),
+                    dims=self._dims,
+                    isherm=self._isherm,
+                    copy=False)
         return Qobj(_data.expm(self._data, dtype=self.dtype),
                     dims=self._dims,
                     isherm=self._isherm,
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -828,7 +828,7 @@ class Qobj:
                         dims=self._dims,
                         isherm=self._isherm,
                         copy=False)
-        return Qobj(_data.expm(self._data, dtype=dtype),
+        return Qobj(_data.expm(self._data, dtype=self.dtype),
                     dims=self._dims,
                     isherm=self._isherm,
                     copy=False)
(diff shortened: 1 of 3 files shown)


rochisha0:
`logm` still does not work, I am unable to figure out the reason.

qutip/core/qobj.py
```diff
@@ -822,7 +822,7 @@ def expm(self, dtype: LayerType = _data.Dense) -> Qobj:
         """
         if not self._dims.issquare:
             raise TypeError("expm is only valid for square operators")
-        return Qobj(_data.expm(self._data, dtype=dtype),
```
Ericgig, line 825: You found the cause of the issue, but the default is dense for a reason. Exponential of sparse matrix are often dense and the computation is inefficient. This would keep sparse matrices' output sparse which is not what we want.

Ericgig (changes requested):

qutip/core/qobj.py
rochisha0: So should I write an `if` statement for CSR matrices?

Ericgig: I leave the method to fix the issue up to you.

Ericgig: What about `Dia` format? `JaxDia`?

Ericgig (changes requested):
Getting close, here is a few improvements suggestion:

- The input `dtype` of `Qobj.expm` is ignored.
- `isinstance` can check for more than one class at once: `isinstance(instance, (class1, class2))`.
- `dtype=self.dtype` would not be ideal if `expm` does not support that type, `dtype=None` would be better.
- It the code would be cleaner if there was only one `return Qobj(...)`, both calls are the same except for one parameter.

Ericgig (changes requested):
A lot cleaner.

But it ignore the user input `dtype` when the type is not sparse.
If an input is sometime used, some time ignored, without any indication to the user, that's bad. The users can see the default. If it's being ignored, it's only confusing.

rochisha0:
So would it be best to include `dense` there too?

Ericgig:
No, `dtype` should always be respected, even for jax.
So the default need to be changed.
One options would be to set `dtype=None` for the default and then if it's `None` and sparse, then use dense. (+ update the docstring and tests.)
Another options would be to remove the `dtype` parameter and replace it with `sparse` like `eigenenergies`. But I would like to keep support for both for a while has we should no introduce breaking change outside of major release which won't come anytime soon (detect if the user input a dtype and warn it in that case, but still give him his intended behaviour.)

Ericgig (commented):
Look good.

Now, we just need the tests to passes.

PR #2485, opened 2024-07-08
Add types hints qutip/core

### LINKED ISSUE #2456: `qutip.expect` does not support numpy array as input
### Bug Description

In documentation of `qutip.expect` both NumPy array and Qobj are allowed as input, but it is not.

### Code to Reproduce the Bug

```shell
import qutip
qutip.expect(qutip.sigmaz().full(), qutip.ket([0],[2]).full())
```

### Code Output

```shell
AttributeError: 'numpy.complex128' object has no attribute 'isherm'
```

### Expected Behaviour

Either fix the documentation or fix the code to support NumPy array. It would be nice to keep the NumPy support though

### Your Environment

```shell
QuTiP Version:      5.1.0.dev0+0aee93e
Numpy Version:      1.26.4
Scipy Version:      1.12.0
Cython Version:     None
Matplotlib Version: 3.9.0
Python Version:     3.11.9
Number of CPUs:     40
BLAS Info:          Generic
INTEL MKL Ext:      False
```

### Additional Context

_No response_

### LINKED ISSUE #2480: Qobj docstring doesn't match `init` signature
### Bug Description

The Qobj class documentation states:

```
class Qobj:
    """
    A class for representing quantum objects, such as quantum operators and
    states.

    The Qobj class is the QuTiP representation of quantum operators and state
    vectors. This class also implements math operations +,-,* between Qobj
    instances (and / by a C-number), as well as a collection of common
    operator/state operations.  The Qobj constructor optionally takes a
    dimension ``list`` and/or shape ``list`` as arguments.

    Parameters
    ----------
    inpt: array_like, data object or :obj:`.Qobj`
        Data for vector/matrix representation of the quantum object.
    dims: list
        Dimensions of object used for tensor products.
    shape: list
        Shape of underlying data structure (matrix shape).
    copy: bool
        Flag specifying whether Qobj should get a copy of the
        input data, or use the original.
```

but the actual `init` function has the form
```
def __init__(self, arg=None, dims=None,
                 copy=True, superrep=None, isherm=None, isunitary=None):

```
so the arguments `inpt` and `shape` aren't recognized.

### Code to Reproduce the Bug

```shell
qt.Qobj(inpt=np.random.randn(4,4),dims=[[2,2],[2,2]])
```

### Code Output

```shell
Cell In[10], line 1
----> 1 qt.Qobj(inpt=np.random.randn(4,4),dims=[[2,2],[2,2]])

TypeError: Qobj.__init__() got an unexpected keyword argument 'inpt'
```

### Expected Behaviour

If I hover on `qt.Qobj` in VS Code, the signature actually shows `inpt` and not `arg`.  It seems like there is some inconsistency in the name of this keyword?  No error is expected.

### Your Environment

```shell
Current admin team: Alexander Pitchford, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, Boxi Li, Jake Lishman, Simon Cross and Asier Galicia.

QuTiP Version:      5.0.2
Numpy Version:      1.26.4
Scipy Version:      1.13.0
Cython Version:     None
Matplotlib Version: 3.8.4
Python Version:     3.12.3
Number of CPUs:     10
BLAS Info:          Generic
INTEL MKL Ext:      False
Installation path:  /Users/joey/miniconda3/envs/qutip5/lib/python3.12/site-packages/qutip
```

### Additional Context

_No response_

Ericgig:
**Description**
Add hints for most of the remaining user facing function in `qutip/core`.
Fixed the docstring of `Qobj`: fix #2480
Improved the docstring of `expect` to fix the confusion with array input: fix #2456

I changed `numbers.Number` for `complex` as the later is not recognized by static type checkers.
I also changed the way `qobj.py` handle circular import. This makes it easier to import it and `qobjevo.py` without causing circular import elsewhere.

Changed the type of expect output. It was

|  | qobj oper | list oper |
|---|---|---|
| qobj state | scalar | Array |
| list state | Array | list[Array]  |

and now is

|  | qobj oper | list oper |
|---|---|---|
| qobj state | scalar | list |
| list state | Array | list[Array]  |

With multiple operators, the output dtype, float or complex, could change, with multiple states, a list of array was used to the right type was applied used, but for multiple operator, single states, they were forced to use the same dtype.

### DIFF
--- a/qutip/core/blochredfield.py
+++ b/qutip/core/blochredfield.py
@@ -7,14 +7,20 @@ from . import Qobj, QobjEvo, liouvillian, coefficient, sprepost
 from ._brtools import SpectraCoefficient, _EigenBasisTransform
 from .cy.coefficient import InterCoefficient, Coefficient
 from ._brtensor import _BlochRedfieldElement
-
+from ..typing import CoeffProtocol
 
 __all__ = ['bloch_redfield_tensor', 'brterm']
 
 
-def bloch_redfield_tensor(H, a_ops, c_ops=[], sec_cutoff=0.1,
-                          fock_basis=False, sparse_eigensolver=False,
-                          br_dtype='sparse'):
+def bloch_redfield_tensor(
+    H: Qobj | QobjEvo,
+    a_ops: list[tuple[Qobj | QobjEvo, Coefficient | str | CoeffProtocol]],
+    c_ops: list[Qobj | QobjEvo] = None,
+    sec_cutoff: float = 0.1,
+    fock_basis: bool = False,
+    sparse_eigensolver: bool = False,
+    br_dtype: str = 'sparse',
+):
     """
     Calculates the Bloch-Redfield tensor for a system given
     a set of operators and corresponding spectral functions that describes the
@@ -46,10 +52,10 @@ def bloch_redfield_tensor(H, a_ops, c_ops=[], sec_cutoff=0.1,
         .. code-block::
 
             a_ops = [
-                (a+a.dag(), ('w>0', args={"w": 0})),
+                (a+a.dag(), coefficient('w>0', args={"w": 0})),
                 (QobjEvo(a+a.dag()), 'w > exp(-t)'),
                 (QobjEvo([b+b.dag(), lambda t: ...]), lambda w: ...)),
-                (c+c.dag(), SpectraCoefficient(coefficient(array, tlist=ws))),
+                (c+c.dag(), SpectraCoefficient(coefficient(ws, tlist=ts))),
             ]
 
 
@@ -102,8 +108,15 @@ def bloch_redfield_tensor(H, a_ops, c_ops=[], sec_cutoff=0.1,
         return R, H_transform.as_Qobj()
 
 
-def brterm(H, a_op, spectra, sec_cutoff=0.1,
-           fock_basis=False, sparse_eigensolver=False, br_dtype='sparse'):
+def brterm(
+    H: Qobj | QobjEvo,
+    a_op: Qobj | QobjEvo,
+    spectra: Coefficient | CoeffProtocol | str,
+    sec_cutoff: float = 0.1,
+    fock_basis: bool = False,
+    sparse_eigensolver: bool = False,
+    br_dtype: str = 'sparse',
+):
     """
     Calculates the contribution of one coupling operator to the Bloch-Redfield
     tensor.
--- a/qutip/core/blochredfield.py
+++ b/qutip/core/blochredfield.py
@@ -1,8 +1,9 @@
 import os
 import inspect
 import numpy as np
-from qutip.settings import settings as qset
+from typing import overload
 
+from qutip.settings import settings as qset
 from . import Qobj, QobjEvo, liouvillian, coefficient, sprepost
 from ._brtools import SpectraCoefficient, _EigenBasisTransform
 from .cy.coefficient import InterCoefficient, Coefficient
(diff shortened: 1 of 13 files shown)


qutip/core/blochredfield.py
```diff
@@ -1,20 +1,49 @@
...
 
 __all__ = ['bloch_redfield_tensor', 'brterm']
 
 
-def bloch_redfield_tensor(H, a_ops, c_ops=[], sec_cutoff=0.1,
-                          fock_basis=False, sparse_eigensolver=False,
-                          br_dtype='sparse'):
+@overload
+def bloch_redfield_tensor(
+    H: Qobj,
```
hodgestar, line 18: It's unclear to me how to read these overloads and to tell what they allow.

The three cases chosen seem to be a subset of the full cartesian product of possible types and their expected return type (please don't add all possible cases).

In the first overload, it's unclear to me how the return can be a `Qobj` if the `a_ops` or `c_ops` are time-dependent.

Passing `Qobj` for `H` or `c_ops` seems to be an allowed type for all three overloads, but the allowed return types are different in each case.

The last two overloads seem to be identical except that the third allows a return of `Qobj | QobjEvo` making the second redundant.

Probably `@overload` is good in some cases of intermediate complexity, but probably not a good fit for cases like this where the output type depends on the types of three other parameters.

Ericgig: Returning multiple types `Qobj | QobjEvo` is causing more issues than helping. Static type checker will cause error if you try to use the result any way that is not supported by all return types. In this case we there is `@overload` for complex case and `TypeVar` for simple ones. 

In the first case, `c_ops` accepting `QobjEvo` is an error. Thank you for catching. The coefficient in `a_ops` is for the spectrum function only and cannot be time dependent if the operator is a `Qobj`.

The second case is any other accepted case.

The last case not an overload and is not used by type checker. I could remove types there if it remove confusion.

hodgestar: Hmm. I don't have a great suggestion for how to resolve this.

When we started with type hints for QuTiP we wanted to create simple hints that would help our users understand what kinds of objects could be passed and help us be consistent in what kinds of objects we supported.

These complex overload annotations are not that.

Since I don't have a great suggestion, let me review the rest and let you decide how to proceed here.

qutip/core/dimensions.py
```diff
@@ -449,7 +455,7 @@ def __init__(self, dims):
         self._pure_dims = True
         self.__setitem__ = _frozen
 
-    def __eq__(self, other):
+    def __eq__(self, other) -> bool:
```
hodgestar, line 458: Do we need to typehint `__eq__` and similar dunder methods that already have expected types in the language?

qutip/core/expect.py
```diff
@@ -16,17 +38,18 @@ def expect(oper, state):
...
         A single or a `list` of quantum states or density matrices.
 
     Returns
     -------
-    expt : float/complex/array-like
-        Expectation value.  ``real`` if ``oper`` is Hermitian, ``complex``
-        otherwise. A (nested) array of expectaction values if ``state`` or
+    expt : float / complex / list / array
+        Expectation value(s).  ``real`` if ``oper`` is Hermitian, ``complex``
+        otherwise. If multiple ``oper`` are passed, a list of array
```
hodgestar, line 51:
```suggestion
        otherwise. If multiple ``oper`` are passed, a list of array.
```

hodgestar (approved):
Left some comments, but happy for you to make the final calls.

qutip/core/dimensions.py
hodgestar: That is true.

PR #2488, opened 2024-07-11
More types hints for solvers

Ericgig:
**Description**
With this most solver have hints. 

I changed the type for `e_ops` to accept single, list and dict instead of just the dict. We almost never use dict in documentation and it was missing in many docstring, so having it as the only hinted type felt wrong.

I also removed the `args` from the `NonMarkovianMCSolver` to make it match the other solvers. It was not used for the Hamiltonian as was indicated in the docstring.

### DIFF
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -53,6 +53,7 @@ coefficient_builders = {
     np.ndarray: InterCoefficient,
     scipy.interpolate.PPoly: InterCoefficient.from_PPoly,
     scipy.interpolate.BSpline: InterCoefficient.from_Bspline,
+    numbers.Number: ConstantCoefficient,
 }
 
 
--- a/qutip/core/cy/coefficient.pyx
+++ b/qutip/core/cy/coefficient.pyx
@@ -745,7 +745,7 @@ cdef class ConstantCoefficient(Coefficient):
     """
     cdef complex value
 
-    def __init__(self, complex value):
+    def __init__(self, complex value, **_):
         self.value = value
 
     def replace_arguments(self, _args=None, **kwargs):
--- a/qutip/solver/floquet.py
+++ b/qutip/solver/floquet.py
@@ -530,7 +530,7 @@ def floquet_tensor(
 
 
 def fsesolve(
-    H: QobjEvoLike,
+    H: QobjEvoLike | FloquetBasis,
     psi0: Qobj,
     tlist: ArrayLike,
     e_ops: EopsLike | list[EopsLike] | dict[Any, EopsLike] = None,
@@ -613,12 +613,12 @@ def fsesolve(
 
 
 def fmmesolve(
-    H: QobjEvoLike,
+    H: QobjEvoLike | FloquetBasis,
     rho0: Qobj,
     tlist: ArrayLike,
     c_ops: list[Qobj] = None,
     e_ops: EopsLike | list[EopsLike] | dict[Any, EopsLike] = None,
-    spectra_cb: list[Callable[[float], complex]]= None,
+    spectra_cb: list[Callable[[float], complex]] = None,
     T: float = 0.0,
     w_th: float = 0.0,
     args: dict[str, Any] = None,
@@ -875,7 +875,7 @@ class FMESolver(MESolver):
         if args:
             raise ValueError("FMESolver cannot update arguments")
 
-    def start(self, state0: Qobj, t0: float, *, floquet: bool=False) -> None:
+    def start(self, state0: Qobj, t0: float, *, floquet: bool = False) -> None:
         """
         Set the initial state and time for a step evolution.
         ``options`` for the evolutions are read at this step.
--- a/qutip/solver/floquet.py
+++ b/qutip/solver/floquet.py
@@ -6,7 +6,9 @@ __all__ = [
     "FMESolver",
 ]
 
+from typing import Any, overload, TypeVar, Literal, Callable
 import numpy as np
+from numpy.typing import ArrayLike
 from qutip.core import data as _data
 from qutip.core.data import Data
 from qutip import Qobj, QobjEvo
@@ -17,6 +19,7 @@ from .integrator import Integrator
 from .result import Result
 from time import time
 from ..ui.progressbar import progress_bars
+from ..typing import EopsLike, QobjEvoLike
 
 
 class FloquetBasis:
@@ -37,13 +40,13 @@ class FloquetBasis:
 
     def __init__(
         self,
-        H,
-        T,
-        args=None,
-        options=None,
-        sparse=False,
-        sort=True,
-        precompute=None,
+        H: QobjEvoLike,
+        T: float,
+        args: dict[str, Any] = None,
+        options: dict[str, Any] = None,
+        sparse: bool = False,
+        sort: bool = True,
+        precompute: ArrayLike = None,
     ):
         """
         Parameters
@@ -120,7 +123,13 @@ class FloquetBasis:
             for ket in _data.split_columns(kets_mat)
         ]
 
-    def mode(self, t, data=False):
+    @overload
+    def mode(self, t: float, data: Literal[False]) -> Qobj: ...
+
+    @overload
+    def mode(self, t: float, data: Literal[True]) -> Data: ...
+
+    def mode(self, t: float, data: bool = False):
         """
         Calculate the Floquet modes at time ``t``.
 
@@ -151,7 +160,13 @@ class FloquetBasis:
         else:
             return self._as_ketlist(kets_mat)
 
-    def state(self, t, data=False):
+    @overload
+    def state(self, t: float, data: Literal[False]) -> Qobj: ...
+
+    @overload
+    def state(self, t: float, data: Literal[True]) -> Data: ...
+
+    def state(self, t: float, data: bool = False):
         """
         Evaluate the floquet states at time t.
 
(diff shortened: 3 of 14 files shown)


qutip/core/cy/coefficient.pyx
```diff
@@ -745,7 +745,7 @@ cdef class ConstantCoefficient(Coefficient):
     """
     cdef complex value
 
-    def __init__(self, complex value):
+    def __init__(self, complex value, **_):
```
hodgestar, line 748: What is the `**_` for here? The other coefficient subclasses don't accept arbitrary additional keyword arguments.

qutip/solver/floquet.py
```diff
@@ -180,7 +195,13 @@ def state(self, t, data=False):
         else:
             return self._as_ketlist(states_mat)
 
-    def from_floquet_basis(self, floquet_basis, t=0):
+    QobjOrData = TypeVar("QobjOrData", Qobj, Data)
```
hodgestar, line 198: `QobjOrData` seems common enough that it could migrate to the qutip typing module.

qutip/solver/nm_mcsolve.py
```diff
@@ -25,9 +29,21 @@
...
-               timeout=None):
+def nm_mcsolve(
+    H: QobjEvoLike,
+    state: Qobj,
+    tlist: ArrayLike,
+    ops_and_rates: list[tuple[Qobj, float | CoefficientLike]] = (),
+    e_ops: EopsLike | list[EopsLike] | dict[Any, EopsLike] = None,
+    ntraj: int = 500,
+    *,
+    args: dict[str, Any] = None,
```
hodgestar, line 40: Should `args` be removed here too?

hodgestar (approved):

qutip/core/cy/coefficient.pyx
Ericgig: Yes, they all do.

It's for the coefficient function. It call every `???Coefficient.__init__` with the same call: https://github.com/qutip/qutip/blob/418e33807b3f63a6d2879dad792f85b500dfc855/qutip/core/coefficient.py#L189

qutip/solver/nm_mcsolve.py
Ericgig: No, the `??solver` functions take operators as list, but the class interface takes QobjEvo. It does not have a `tlist` needed for some coefficient, it's also a lot cleaner.

qutip/solver/nm_mcsolve.py
```diff
@@ -359,12 +372,16 @@ class NonMarkovianMCSolver(MCSolver):
     }
 
     def __init__(
-        self, H, ops_and_rates, args=None, options=None,
+        self,
+        H: Qobj | QobjEvo,
+        ops_and_rates = Sequence[tuple[Qobj, float | Coefficient]],
+        *,
+        options: dict[str, Any] = None,
```
pmenczel, line 379: I am unsure about removing the `args` parameter here. It seems reasonable to me that anywhere that the user is allowed to pass a pure python function to describe time dependence, they should also be able to pass `args`.

Also, the type hint for `ops_and_rates` doesn't correspond to the actual code. What we actually accept, currently, is `Sequence[tuple[Qobj, float | CoefficientLike]]`. Therefore, users are able to pass callables there, and I would expect `args` to be possible.

Note that for the constructor of MCSolver, the type hint for `H` is `QobjEvoLike`, but I think currently it only accepts `Qobj | QobjEvo` (because of [this issuper](https://github.com/qutip/qutip/blob/master/qutip/solver/mcsolve.py#L467)). Therefore, an `args` parameter is unneccessary. If the code is changed such that any `QobjEvoLike` is accepted, I would argue to add an `args` parameter also there.

Ericgig: All `??Solver` do not take `QobjEvoLike`, only pre-made `QobjEvo` by design. Thank you for catching.

The `args` description was wrong. It says that it was used for `Arguments for time-dependent Hamiltonian and collapse operator terms.`. But it these operators should already be `QobjEvo` and do not use this `args`. 

`CoefficientLike` support array format, thus need the `tlist`, which does not belong there.  In the `??solve` functions, we accept anything an try to do our best to make sense of it, but it's messy... With the class interface, I try to be more strict and avoid these cases.

qutip/solver/mcsolve.py
```diff
@@ -562,8 +563,8 @@ def run(
         ntraj: int | list[int] = None,
         *,
         args: dict[str, Any] = None,
-        e_ops: dict[Any, Qobj | QobjEvo | Callable[[float, Qobj], Any]] = None,
-        target_tol: float = None,
+        e_ops: EopsLike | list[EopsLike] | dict[Any, EopsLike] = None,
+        target_tol: float | tuple[float, float] | list[tuple[float, float]] = None,
```
pmenczel, line 567: Should the `target_tol` also be updated in MultiTrajSolver.run?

qutip/core/coefficient.py
```diff
@@ -53,6 +53,7 @@ def _return(base, **kwargs):
     np.ndarray: InterCoefficient,
     scipy.interpolate.PPoly: InterCoefficient.from_PPoly,
     scipy.interpolate.BSpline: InterCoefficient.from_Bspline,
+    numbers.Number: ConstantCoefficient,
```
pmenczel, line 56: Should we then also include `float` in the definition of `CoefficientLike`? (Then, in `nm_mcsolve`, we could remove some "`float |`")

Also, the `if isinstance(rate, numbers.Number)` branching in `nm_mcsolve._parse_op_and_rate` shouldn't be necessary any more with this, right?

qutip/solver/nm_mcsolve.py
```diff
@@ -541,7 +570,15 @@ def _run_one_traj(self, seed, state, tlist, e_ops, **integrator_kwargs):
         result.trace = martingales
         return seed, result
 
-    def run(self, state, tlist, ntraj=1, *, args=None, **kwargs):
+    def run(
+        self,
+        state: Qobj,
+        tlist: Sequence[float],
```
pmenczel, line 576: In other places, I saw `tlist` hinted as `ArrayLike`. (I don't know which one is better / correct, but it should be consistent, I think.)

qutip/solver/stochastic.py
```diff
@@ -1,17 +1,24 @@
 __all__ = ["smesolve", "SMESolver", "ssesolve", "SSESolver"]
 
+import numpy as np
+from numpy.typing import ArrayLike
+from numpy.random import SeedSequence
```
pmenczel, line 5:
```suggestion
import numpy as np
from numpy.random import SeedSequence
```

qutip/solver/stochastic.py
```diff
@@ -192,7 +199,7 @@ def wiener_process(self):
         """
         return self._trajectories_attr("wiener_process")
 
-    def merge(self, other, p=None):
+    def merge(self, other: "StochasticResult", p=None) -> "StochasticResult":
```
pmenczel, line 202:
```suggestion
    def merge(self, other: "StochasticResult", p: float = None) -> "StochasticResult":
```

qutip/solver/floquet.py
```diff
@@ -37,13 +40,13 @@ class FloquetBasis:
...
     def __init__(
         self,
-        H,
-        T,
-        args=None,
-        options=None,
-        sparse=False,
-        sort=True,
-        precompute=None,
+        H: QobjEvoLike,
```
pmenczel, line 43:
```suggestion
        H: Qobj | QobjEvo,
```

qutip/solver/floquet.py
```diff
@@ -496,7 +527,15 @@ def floquet_tensor(H, c_ops, spectra_cb, T=0, w_th=0.0, kmax=5, nT=100):
     return Qobj(r, dims=[dims, dims], superrep="super", copy=False)
 
 
-def fsesolve(H, psi0, tlist, e_ops=None, T=0.0, args=None, options=None):
+def fsesolve(
+    H: QobjEvoLike,
```
pmenczel, line 531: There might be conversion `QobjEvoLike` -> `QobjEvo` missing in the function (and also `fmesolve`)

qutip/solver/nm_mcsolve.py
pmenczel: Okay, that makes sense to me. I went to the changes and found a few more small inconsistencies, I think. Overall, very nice to have some type hints, thank you!

qutip/solver/nm_mcsolve.py
Ericgig: Neither is exact, `Sequence` description match what we expect, but does not include numpy arrays.
ArrayLike does not limit the type of element or number of dimensions.
I am not sure which we should use...

pmenczel: Since we're not aiming to be 100% exact with the hints (as far as I understood), I would think that either is fine as long as it is used consistently.

pmenczel (commented):
Looks quite good to me now. Just two final questions, sorry:
* In fmmesolve and fsesolve, should `H` perhaps be `QobjEvoLike | FloquetBasis`?
* Lines 215-218 [here](https://github.com/Ericgig/qutip/blob/types_solver/qutip/solver/nm_mcsolve.py#L215) could now be replaced by just
```
rate = coefficient(rate, **kw)
```
right?

pmenczel (approved):

PR #2490, opened 2024-07-12
[feature] Create dynamic numpy backend selection class 

rochisha0:
**Description**
This pull request introduces a new `NumpyBackend `class that enables dynamic selection of the numpy_backend used in `qutip`. The class facilitates switching between different numpy implementations ( `numpy` and `jax.numpy` mainly) based on the configuration specified in the `settings.core` dictionary. 

Made changes in `entropy` and `metrics` to make them `jax` enabled


qutip/entropy.py
```diff
@@ -2,8 +2,8 @@
            'concurrence', 'entropy_conditional', 'entangling_power',
            'entropy_relative']
 
-from numpy import conj, e, inf, imag, inner, real, sort, sqrt
-from numpy.lib.scimath import log, log2
+from .backend import np
+from math import e
```
Ericgig, line 6: This `e` is not used...

Ericgig (changes requested):
This would work. Did you test if this affect the 

- Could you make the PR to the `dev.major` branch instead of `master`?
- Let's name the new file `qutip/core/numpy_backend.py`.
- Files should end with an empty line.
- It would be good to have a test for this.

qutip/entropy.py
rochisha0: It is used in function definition of `entropy_relative`

rochisha0:
Yes, I have tested and it works both with `numpy` and `jax.numpy`

Ericgig:
Yes, please change the base branch to `dev.major` (and resolve any conflict if present).

Successful importation is not sufficient. It need to show that the numpy implementation can be changed. Normal usage only show that it can be used as the original library, not that it's changeable.

qutip/entropy.py
Ericgig: Why use the one from math instead of numpy?
Is using the numpy wrapper in a function definition causing bugs?

rochisha0: I did this to future proof with any error that may occur.

rochisha0:
@Ericgig I'm thinking of converting this as draft and see if we can make `measurement` , `random_object` and others `jax` enabled with this change. What do you think?

Ericgig:
No, let's add those in other PRs. 
There is always more than can be added, but it does not have to be done all at once.

Ericgig:
Commits behind are not an issue. Commits ahead will cause conflict that you will have to fix.
No need to rebase (`git rebase`), git merging tools should be good enough.

`dev.major` is not that much behind master. Is your local copy of it up to date?

qutip/numpy_backend.py
Ericgig, line 1: Please move this file into `qutip/core`.

qutip/tests/test_numpy_backend.py
```diff
@@ -0,0 +1,47 @@
...
+        np_backend = NumpyBackend()
+
+        backend.sum = MagicMock(return_value="sum_result")
+        result = np_backend.sum([1, 2, 3])
+        assert result == "sum_result"
+        backend.sum.assert_called_with([1, 2, 3])
+
+
+@pytest.fixture(scope='module')
+def mock_settings():
```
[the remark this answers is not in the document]
rochisha0: It is used to temporarily modify `numpy_backend` in settings with mock backend.

rochisha0:
After shifting to `core` the tests have started interfering with each other. It would be better if we could keep the `numpy_backend` class to qutip not `core`.

qutip/tests/core/test_numpy_backend.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+import numpy as np
+from unittest.mock import patch, MagicMock
+
+from ...settings import settings
+from ...core.numpy_backend import NumpyBackend
```
Ericgig, line 6: No relative import in tests.

qutip/tests/core/test_numpy_backend.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestNumpyBackend:
+    def test_backend_property(self):
+        settings.core["numpy_backend"] = mock_np
+        np_backend = NumpyBackend()
+        assert np_backend.backend is mock_np
+
+        settings.core["numpy_backend"] = mock_jax
+        assert np_backend.backend is mock_jax
+        settings.core["numpy_backend"] = np
```
Ericgig, line 21: If the test fail, this would never be called.
Qutip's options support context:
```
with qutip.CoreOptions(numpy_backend=...):
    assert ...
```

qutip/tests/core/test_numpy_backend.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Mocking JAX and NumPy to demonstrate backend switching
+mock_jax = MagicMock()
+mock_np = MagicMock()
+
+
+class TestNumpyBackend:
+    def test_backend_property(self):
+        settings.core["numpy_backend"] = mock_np
+        np_backend = NumpyBackend()
```
Ericgig, line 16: We use it as `from .core.numpy_backend import np`, so it should be tested that way.

qutip/tests/core/test_numpy_backend.py
Ericgig, line 1: At least one test should use the normal numpy.

qutip/core/metrics.py
```diff
@@ -80,7 +79,7 @@ def fidelity(A, B):
     # We also truncate negative eigenvalues to avoid nan propagation;
     # even for positive semidefinite matrices, small negative eigenvalues
     # can be reported.
-    eig_vals = (sqrtmA * B * sqrtmA).eigenenergies()
+    eig_vals = (sqrtmA * B * sqrtmA).eigenenergies()
```
Ericgig, line 82:
```suggestion
    eig_vals = (sqrtmA * B * sqrtmA).eigenenergies()
```

qutip/tests/core/test_numpy_backend.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Mocking JAX to demonstrate backend switching
+mock_jax = Mock()
+mock_np = numpy
+
+
+class TestNumpyBackend:
+    def test_getattr_numpy(self):
+        with CoreOptions(numpy_backend=mock_np):
+            assert np.sum([1, 2, 3]) == numpy.sum([1, 2, 3])
```
Ericgig, line 16: Adding one strict assert, but otherwise good.
```suggestion
            assert np.sum([1, 2, 3]) == numpy.sum([1, 2, 3])
            assert np.sum is numpy.sum
```

Ericgig (approved):
Look good.
I made two minor suggestion that I will merge immediately, and merge this today.

PR #2492, opened 2024-07-15
Fix bug in stochastic step.

### LINKED ISSUE #2491: step does not work for SMESolver
### Bug Description

I wanted to use the new `step` method for the stochastic solver class `SMESolver`, but it always breaks down. I also tried it for `SSESolver`, but the same error message occurs. I use qutip 5.0.3. I took the example from the documentation as a minimal working example. I also tried `step` for the `MESolver` class and there it works. I think it should also not only output the state but also the measurement output if possible.

As a side note, I also wanted to know if there is a particular reason why the argument for the initial state in the `start` method is called `state0` for `MESolver`, and `state` for `SMESolver`.

### Code to Reproduce the Bug

```shell
import numpy as np
from qutip import destroy, coherent, SMESolver, expect

# parameters
DIM = 20               # Hilbert space dimension
DELTA = 5 * 2 * np.pi  # cavity detuning
KAPPA = 2              # cavity decay rate
INTENSITY = 4          # intensity of initial state
NUMBER_OF_TRAJECTORIES = 500

# operators
a = destroy(DIM)
x = a + a.dag()
H = DELTA * a.dag() * a

rho_0 = coherent(DIM, np.sqrt(INTENSITY))
times = np.arange(0, 1, 0.0025)

stoc_solver = SMESolver(
    H,
    c_ops=[],
    sc_ops=[np.sqrt(KAPPA) * a],
    heterodyne=False,
    options={"dt": 0.00125, "store_measurement": True,}
)
data = [4.]
stoc_solver.start(state=rho_0, t0=times[0])
for t in times[1:]:
    psi_t = stoc_solver.step(t)
    data.append(expect(x, psi_t))
```

### Code Output

```shell
Cell In[4], line 29
     27 stoc_solver.start(state=rho_0, t0=times[0])
     28 for t in times[1:]:
---> 29     psi_t = stoc_solver.step(t)
     30     data.append(expect(x, psi_t))

File /user/anaconda/envs/myenv/lib/python3.9/site-packages/qutip/solver/multitraj.py:129, in MultiTrajSolver.step(self, t, args, copy)
    127     raise RuntimeError("The `start` method must called first.")
    128 self._argument(args)
--> 129 _, state = self._integrator.integrate(t, copy=False)
    130 return self._restore_state(state, copy=copy)

ValueError: too many values to unpack (expected 2)
```

### Expected Behaviour

It should have returned the state at time step t instead of failing. It would also be nice if the method would return the output of the measurement current as well.

### Your Environment

```shell
QuTiP Version:      5.0.3
Numpy Version:      1.26.4
Scipy Version:      1.13.1
Cython Version:     None
Matplotlib Version: 3.6.0
Python Version:     3.9.19
Number of CPUs:     16
BLAS Info:          Generic
INTEL MKL Ext:      False
```

### Additional Context

_No response_

Ericgig:
**Description**
Fix a bug in the stochastic solver step function.

Also add the wiener increment as an optional output.

**Related issues or PRs**
fix #2491

### DIFF
--- a/qutip/solver/stochastic.py
+++ b/qutip/solver/stochastic.py
@@ -1,6 +1,6 @@
 __all__ = ["smesolve", "SMESolver", "ssesolve", "SSESolver"]
 
-from typing import Any
+from typing import Any, Literal, overload
 from .multitrajresult import MultiTrajResult
 from .sode.ssystem import StochasticOpenSystem, StochasticClosedSystem
 from .sode._noise import PreSetWiener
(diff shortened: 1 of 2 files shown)


qutip/tests/solver/test_stochastic.py
```diff
@@ -562,3 +562,29 @@ def test_merge_results(store_measurement, keep_runs_results):
             w.shape == result_merged.wiener_process[0].shape
             for w in result_merged.wiener_process
         )
+
+@pytest.mark.parametrize("open", [True, False])
```
hodgestar, line 566: It would be nice to have a test to cover the `heterodyne=False` case.

qutip/solver/stochastic.py
```diff
@@ -792,6 +793,55 @@ def run_from_experiment(
...
+        args : dict, optional
+            Update the ``args`` of the system.
+            The change is effective from the beginning of the interval.
+            Changing ``args`` can slow the evolution.
+
+        copy : bool, default: True
+            Whether to return a copy of the data or the data in the ODE solver.
+
+        wiener_increment: bool, default: False
+            Whether to return ``dW`` with for the step with the state.
```
hodgestar, line 832:
```suggestion
            Whether to return ``dW`` in addition to the state.
```

hodgestar (approved):

PR #2503, opened 2024-07-29
Update solver's signature to take `e_ops` and `args` as keyword only.

### LINKED ISSUE #2489: solve function signature don't match
### Problem Description

Some the the `??solve` functions inconsistent signature:
Almost all of them start with `H, state0, tlist`, the collapse if used, expectation, `args` and finally options.

But 
- For `brmesolve`, `c_ops` comes after `e_ops`.
- For HEOM: the bath is between the Hamiltonian and the initial state.
- Stochastic solver have `e_ops` `options` and `args` as keyword only.
- `ntraj` is positional for `mcsolve`, but keyword only for stochastic solver.
- Solver specific parameters are between `e_ops` and `args` for floquet, between `args` and `options` for bloch redfield and after `sc_ops` in stochastic.

### Proposed Solution

Make `e_ops`, `args`, `options` and `ntraj` keyword only from 5.3, with a warning message from 5.1.

This would make is all signatures, except HEOM, match.

For HEOM, the `bath` and `max_depth` should be swapped for the `state` and `tlist`.

### Alternate Solutions

_No response_

### Additional Context

_No response_

Ericgig:
**Description**
The signatures of the `__solve` functions are not consistent. Most time the `e_ops` comes after the `c_ops`, but this is not always the case.
Making `e_ops`, `args` and `options` keyword fix this issues. (see #2489).
Using `e_ops`, etc. positionally will raise a `FutureWarning`, with the full removal planed for qutip 5.3.

For `brmesolve`, the transition is not clean as `e_ops` was before `c_ops`. Both can take the same input, so that no confusion happen between the two, in that case `c_ops` was also made keyword only. For that solver, the `a_ops` replace the `c_ops` as the main mixing with the environment and should not be used that much.

**Related issues or PRs**
fix #2489

### DIFF
--- a/qutip/solver/brmesolve.py
+++ b/qutip/solver/brmesolve.py
@@ -27,7 +27,7 @@ def brmesolve(
     tlist: ArrayLike,
     a_ops: list[tuple[QobjEvoLike, CoefficientLike]] = None,
     sec_cutoff: float = 0.1,
-    *pos_args,
+    *_pos_args,
     c_ops: list[QobjEvoLike] = None,
     e_ops: EopsLike | list[EopsLike] | dict[Any, EopsLike] = None,
     args: dict[str, Any] = None,
@@ -93,6 +93,13 @@ def brmesolve(
         Cutoff for secular approximation. Use ``-1`` if secular approximation
         is not used when evaluating bath-coupling terms.
 
+    *_pos_args
+        Temporary shim to update the signature from
+        (..., a_ops, e_ops, c_ops, args, sec_cutoff, options)
+        to (..., a_ops, sec_cutoff, *, e_ops, c_ops, args, options)
+        making e_ops, c_ops, args and options keyword only parameter from
+        qutip 5.3.
+
     c_ops : list of (:obj:`.QobjEvo`, :obj:`.QobjEvo` compatible format), optional
         List of collapse operators.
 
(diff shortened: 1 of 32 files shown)


qutip/solver/brmesolve.py
```diff
@@ -25,10 +26,11 @@ def brmesolve(
     psi0: Qobj,
     tlist: ArrayLike,
     a_ops: list[tuple[QobjEvoLike, CoefficientLike]] = None,
-    e_ops: EopsLike | list[EopsLike] | dict[Any, EopsLike] = None,
+    sec_cutoff: float = 0.1,
+    *pos_args,
```
hodgestar, line 30: Could we add `pos_args` to the argument list and use it to document that the parameters following it are now keyword only and what the previous ordering was.

I would also be happy for the order to be finally changed as soon as possible -- e.g. 5.2. Having this clumsy signature is probably more detrimental than small breakages.

hodgestar (approved):
Reducing the number of positional arguments to the solvers is a great clarifying change. It is also unfortunately a tricky change. I would be in favour of reducing the amount of time the code tries to cater to both cases as much as possible.

qutip/solver/brmesolve.py
Ericgig: It depend on the pace of releases...
We should not have breaking changes in minor release without deprecation warning. They usually come with a 1 year head up in other projects. 2 minor versions in a year feels a reasonable pace, but we don't have set schedule making it difficult...

qutip/solver/solver_base.py
```diff
@@ -569,3 +569,13 @@ def _solver_deprecation(kwargs, options, solver="me"):
     if kwargs:
         raise TypeError(f"unexpected keyword argument {kwargs.keys()}")
     return options
+
+
+def _kwargs_migration(position, keyword, name):
+    if position is not None:
+        warnings.warn(
+            f"{name} will be positional only from qutip 5.3 for all solver",
```
pmenczel, line 577: This should read "keyword only", right?

Ericgig: Yes, thank you for catching it.

qutip/solver/brmesolve.py
```diff
@@ -79,19 +81,26 @@ def brmesolve(
...
+        Cutoff for secular approximation. Use ``-1`` if secular approximation
+        is not used when evaluating bath-coupling terms.
+
+    *_pos_args :
+        | Temporary shim to update the signature from
+        | ``(..., a_ops, e_ops, c_ops, args, sec_cutoff, options)``
+        | to
+        | ``(..., a_ops, sec_cutoff, *, e_ops, c_ops, args, options)``
+        | making ``e_ops``, ``c_ops``, ``args`` and ``options`` keyword only
+          parameter from qutip 5.3.
```
pmenczel, line 103:
```suggestion
```

I don't know the word "shim", and I don't think these temporary parameters are documented in the other solvers, so we could consider removing this

qutip/solver/floquet.py
```diff
@@ -588,6 +590,23 @@ def fsesolve(
         contains either an *array* of expectation values or an array of
         state vectors, for the times specified by `tlist`.
     """
+    if pos_args or not isinstance(T, (int, float)):
+        # Old signature used
+        warnings.warn(
+            f"e_ops, args and options will be keyword only"
+            " for all solver from qutip 5.3",
```
pmenczel, line 597:
```suggestion
            " for all solvers from qutip 5.3",
```

pmenczel (approved):
Looks good, I only left two very minor comments on the text. Are any changes required in the tutorial notebooks?

qutip/solver/brmesolve.py
Ericgig: I added it after a comment from Simon. This change is more complex than others so it would help to have a reminder of what the signature previously was somewhere.

pmenczel: I see, okay

PR #2504, opened 2024-07-30
Add familly package tests

Ericgig:
**Description**
Add test jobs for the familly packages (`qutip-qip`, `qutip-qoc`, `qutip-jax`) for PR in `qutip/qutip`.
Tests the development version or could be ran manually to test any version (tags / branch).

This will cause a lot of PR to look bad with error outside their scope, but if we keep the other modules working, then this could be a useful tool.


hodgestar:
I think this is a great idea. Do you see any issues with merging it?

Ericgig:
It will cause a lot of tests failure not related to changes, We will get tests failure due to new version of jax or a deprecation in numpy /scipy will have to be fixed in multiple repos for the tests to work as we are seen with the present failure.

It require commitment to keep everything working and up to date so I would like to hear from the maintainers of the other projects before merging. 
Also should other projects such as `qutip-cupy`, `qutip-qtrl` be included?

hodgestar:
For `qutip-cupy` I would say no for the moment -- it is unmaintained. For `qutip-qtrl` probably yes?

Let's add this to the agenda for the next QuTiP admin meeting. We will need some agreement that maintainers of other repositories have time to fix issues that come up.

Ericgig:
@hodgestar
Could you review?

.github/workflows/tests_env.yml
```diff
@@ -0,0 +1,61 @@
+name: Tests familly project
```
hodgestar, line 1:
```suggestion
name: Test family projects
```

.github/workflows/tests_env.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: actions/checkout@v4
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          python-version: 3.11
+
+      - name: Install QuTiP and dependencies
+        run: |
+          # Install the extra requirement
+          python -m pip install pytest pytest-rerunfailures  # tests
```
hodgestar, line 37: What do we need pytest-rerunfailures for?

.github/workflows/tests_env.yml
```diff
@@ -0,0 +1,61 @@
...
+          python -m pip install numpy scipy cython filelock
+          python -m pip install -e . -v --no-build-isolation
+      - name: Install qutip-plugin
```
hodgestar, line 40:
```suggestion
      - name: Install QuTiP family package
```

hodgestar (commented):
Overall looks good to me. There are some assumptions about how the other repositories use tags and run tests, but I think that's okay.

I would prefer not to re-run failures unless there is some good reason (e.g. maybe one the family package repositories requires it?)

.github/workflows/tests_env.yml
Ericgig: It's a requirement for tests in `setup.cfg`.
It add the pytest mark `flaky` used for metric and simdiag tests.

hodgestar: Ah. It would be nice to remove those eventually.

hodgestar (approved):

Issue #2175, opened 2023-06-13
steadystate() cannot handle Liouvillians with non-unique steady states


jonahpeter:
### Bug Description

The function steadystate() does not seem capable of handling Liouvillians with non-unique steady states. This is immediately apparent because steadystate() does not contain an argument for an initial state at time zero. Instead, it returns a state (presumably some linear combination of the null-space operators) that does not match the long time output from mesolve(), which varies based on initial condition.

There are many Liouvillians which do not admit unique steady states, but a simple example can be found in Section 3.3 of [this tutorial](https://doi.org/10.48550/arXiv.2303.16449).

### Code to Reproduce the Bug

```shell
The bug is apparent from the documentation, as steadystate() does not contain an argument for an initial state.
```

### Expected Behaviour

For Liouvillians that admit non-unique steady states, I expected steadystate() to return the long time limit of mesolve() given a particular initial condition.

### Your Environment

```shell
QuTiP: Quantum Toolbox in Python
================================
Copyright (c) QuTiP team 2011 and later.
Current admin team: Alexander Pitchford, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, Boxi Li, Jake Lishman and Simon Cross.
Board members: Daniel Burgarth, Robert Johansson, Anton F. Kockum, Franco Nori and Will Zeng.
Original developers: R. J. Johansson & P. D. Nation.
Previous lead developers: Chris Granade & A. Grimsmo.
Currently developed through wide collaboration. See https://github.com/qutip for details.

QuTiP Version:      4.6.2
Numpy Version:      1.21.2
Scipy Version:      1.7.1
Cython Version:     0.29.24
Matplotlib Version: 3.5.0
Python Version:     3.7.11
Number of CPUs:     2
BLAS Info:          INTEL MKL
OPENMP Installed:   False
INTEL MKL Ext:      True
Platform Info:      Darwin (x86_64)
Installation path:  /opt/anaconda3/lib/python3.7/site-packages/qutip
================================================================================
Please cite QuTiP in your publication.
================================================================================
For your convenience a bibtex reference can be easily generated using `qutip.cite()`
```

nwlambert:
Yeah in my experience this is something one has to be careful of with the steady-state solver as it is. Depending on the method chosen it can either fail, return one of the possibilities, or some linear combination of them.  The default one (direct) tends to fail, which at least sometimes lets you know you have this issue. 

[...]

PR #2508, opened 2024-08-09
Add propagator method to steadystate

Ericgig:
**Description**
Add a new method for `steadystate` that can take an initial state. Useful for system that can have multiple steadystate (#2175)

### DIFF
--- a/qutip/solver/steadystate.py
+++ b/qutip/solver/steadystate.py
@@ -1,5 +1,5 @@
 from qutip import liouvillian, lindblad_dissipator, Qobj, qzero_like, qeye_like
-from qutip import vector_to_operator, operator_to_vector
+from qutip import vector_to_operator, operator_to_vector, hilbert_dist
 from qutip import settings
 import qutip.core.data as _data
 import numpy as np
@@ -56,6 +56,7 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
         - "eigen" : Eigenvalue problem
         - "svd" : Singular value decomposition
         - "power" : Inverse-power method
+        - "propagator" : Apply the propagator in repetition
 
     solver : str, optional
         'direct' and 'power' methods only.
@@ -106,6 +107,12 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
         (default sparse).  With "direct" and "power" method, when the solver is
         not specified, it is used to set whether "solve" or "spsolve" is
         used as default solver.
+
+    rho: Qobj, default: None
+        Initial state of the propagator method.
+
+    propagator_tol: float, default: 1e-5
+        Tolerance for propagator method convergance.
 
     **kwargs :
         Extra options to pass to the linear system solver. See the
@@ -184,6 +191,9 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
         kwargs.pop("weight", 0)
         kwargs.pop("sparse", 0)
         return _steadystate_power(A, method=solver, **kwargs)
+
+    elif method == "propagator":
+        return _steadystate_expm(A, **kwargs)
     else:
         raise ValueError(f"method {method} not supported.")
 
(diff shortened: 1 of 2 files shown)


qutip/solver/steadystate.py
```diff
@@ -268,6 +278,28 @@ def _steadystate_svd(L, **kw):
...
+    niter = 0
+    while hilbert_dist(rho, rho_prev) > propagator_tol and niter < 30:
+        rho_prev = rho
+        rho = prop(rho)
+        prop = prop @ prop
+        rho = (rho + rho.dag()) / (2 * rho.tr())
+        niter += 1
+
+    if niter == 30:
+        raise Exception("Did not converge to a steadystate.")
```
hodgestar, line 298: It might be true that the distance is already less than `propagator_tol` here and we raise an unnecessary error.

qutip/solver/steadystate.py
```diff
@@ -268,6 +278,28 @@ def _steadystate_svd(L, **kw):
...
+    prop = (100 * L).expm()
+
+    rho_prev = qzero_like(rho)
+    niter = 0
+    while hilbert_dist(rho, rho_prev) > propagator_tol and niter < 30:
+        rho_prev = rho
+        rho = prop(rho)
+        prop = prop @ prop
+        rho = (rho + rho.dag()) / (2 * rho.tr())
+        niter += 1
```
hodgestar, line 295:
```suggestion
    dist = 1.0 + propagator_rol
    niter = 0
    while niter < 30:
        rho_next = prop(rho)
        rho_next = (rho_next + rho_next.dag()) / (2 * rho_next.tr())
        if hilbert_dist(rho_next, rho) <= propagator_tol:
            return rho_next
        rho = rho_next
        prop = prop @ prop
        niter += 1
```
While the original is terse it calculates a `prop` that is ununsed and starts the first iteration with a `hilbert_dist(...)` comparison that always fails. There are probably further small improvements that can be made to the above.

hodgestar (commented):
I think this is a nice simple method that is worthwhile adding. If we're intending it to resolve the issue of finding multiple steadystates we need to document the procedure for using it that way. In either case we need to explain a bit more the details of what this method does.

I assume the procedure is something like: run `steadystate(..., method="propagator")` a number of times equal to the dimension of the density matrix space and then find the subspace spanned by the results? Or is there something better one can do?

qutip/solver/steadystate.py
```diff
@@ -56,6 +56,7 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
         - "eigen" : Eigenvalue problem
         - "svd" : Singular value decomposition
         - "power" : Inverse-power method
+        - "propagator" : Apply the propagator in repetition
```
hodgestar, line 59:
```suggestion
        - "propagator" : Find the steady state by repeatedly applying the propagator
```

qutip/solver/steadystate.py
```diff
@@ -107,6 +108,22 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
         not specified, it is used to set whether "solve" or "spsolve" is
         used as default solver.
 
+    rho: Qobj, default: None
+        Initial state of the propagator method.
```
hodgestar, line 112:
```suggestion
        Initial state for the "propagator" method.
```

qutip/solver/steadystate.py
```diff
@@ -107,6 +108,22 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
...
+
+    propagator_T: float, default: 10
+        Initial time step of the propagator method. The time step double each
+        iteration.
```
hodgestar, line 116:
```suggestion
    propagator_T: float, default: 10
        Initial time step for the propagator method. The time step is doubled each
        iteration.
```

qutip/solver/steadystate.py
```diff
@@ -107,6 +108,22 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
...
+
+    propagator_tol: float, default: 1e-5
+        Tolerance for propagator method convergance. If the hilbert distance
+        between the states of a step is less than this tolerance. The states is
+        considered to have converged to the steady state.
```
hodgestar, line 121:
```suggestion
    propagator_tol: float, default: 1e-5
        Tolerance for propagator method convergence. If the Hilbert distance
        between the states of a step is less than this tolerance, the state is
        considered to have converged to the steady state.
```

qutip/solver/steadystate.py
```diff
@@ -268,6 +290,27 @@ def _steadystate_svd(L, **kw):
...
+    while niter < max_iter:
+        rho_next = prop(rho)
+        rho_next = (rho_next + rho_next.dag()) / (2 * rho_next.tr())
+        if hilbert_dist(rho_next, rho) <= propagator_tol:
+            return rho_next
+        rho = rho_next
+        prop = prop @ prop
+        niter += 1
+
+    raise RuntimeError("Did not converge to a steadystate.")
```
hodgestar, line 311:
```suggestion
    raise RuntimeError(f"Did not converge to a steadystate after {max_iter} iterations.")
```

PR #2509, opened 2024-08-16
Optimize `numpy_backend`

Ericgig:
**Description**
Change numpy_backend to contain the numpy implementation in an attribute instead of reading it from the settings each time. This is done by making the options work as a property, when it is set, it update the `np` instance.

### DIFF
--- a/qutip/core/numpy_backend.py
+++ b/qutip/core/numpy_backend.py
@@ -1,14 +1,11 @@
 from ..settings import settings
 
-
 class NumpyBackend:
-    @property
-    def backend(self):
-        return settings.core["numpy_backend"]
+    def _qutip_setting_backend(self, np):
+        self.np = np
 
     def __getattr__(self, name):
-        backend = object.__getattribute__(self, 'backend')
-        return getattr(backend, name)
+        return getattr(self.np, name)
 
 
 # Initialize the numpy backend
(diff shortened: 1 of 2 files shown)


Ericgig:
@rochisha0, could you look at this PR and comment?
I don't think you can do an official review.

hodgestar (approved):

hodgestar:
@Ericgig When I read the description of this PR I thought you might have implemented something that might be even faster. Instead of having the numpy proxy have dynamic access via `__getattr__` we could do either of the following:

* When the numpy backend is set, actually copy the attributes across to the proxy objects so that they are looked up in `__dict__` in the ordinary way without having to fall back to `__getattr__`.
* Do the same as the above but define the attributes as slots so that the `__dict__` lookup is avoided.
* Or implement `__getattribute__` so that the original `__dict__` lookup is never checked.

rochisha0:
@Ericgig . This looks good to me.

Ericgig:
@hodgestar 
I never heard of `__slots__`, interesting. 
I though about copying values in `__dict__`, but we need this quickly for rochisha's project and this is not a method that can be rushed. For floquet, we will need to find a way to map `np.add.outer` to `jax.vmap(jax.numpy.add)`. Having that use case in mind, I was probably over thinking it...

I will merge it as is and make a PR to merge `dev.major` into `master`, but will come back to it.

Issue #2328, opened 2024-02-18
Enforce CoreOptions(default_dtype) more broadly?


nwlambert:
### Bug Description

When setting `CoreOptions(default_dtype)` I expect all quantum objects to inherit the specified dtype. States and operators defined in states.py and operators.py do so but some functions return other data types.  like `eigenstates()`.  This can lead to some unexpected behavior (and very inefficient code).

Seems like a good first issue for someone to go around and find functions where this should be enforced and do so?  

Alternatively, would forcing CoreOptions(default_dtype) to be used in `Qobj` creation itself be easier? Though I kinda feel it might cause secondary problems, like if  you wanted largely to use a default dtype but then have special cases where you don't, etc etc.

### Code to Reproduce the Bug

```shell
with CoreOptions(default_dtype="CSR"):
    H = sigmaz() + sigmax()
    print(H.eigenstates()[1][0].data)
```

### Code Output

```shell
Dense(shape=(2, 1), fortran=True)
```

### Expected Behaviour

Return sparse states.

### Your Environment

```shell
QuTiP Version:      5.0.0a2
Numpy Version:      1.26.0rc1
Scipy Version:      1.10.1
Cython Version:     0.29.33
Matplotlib Version: 3.6.3
Python Version:     3.11.7
Number of CPUs:     28
BLAS Info:          Generic
INTEL MKL Ext:      False
Platform Info:      Linux (x86_64)
```

hodgestar:
@nwlambert Can you say a bit more about the situation where this arose?

In the example you give, the eigenstates really are dense (although, small, of course), and I think that generically that is true for many states.

Perhaps we need a `default_state_dtype` and `default_oper_dtype`?

nwlambert:
Yeah, on the practical side I was using those states to construct collapse operators, and then using these in steadystate() for a large range of parameters.  Bit surprised to see what took 20 minutes in 4.7 took >5 hours in in v5 :0   I guess this is a combo of Liouvillian construction taking much longer as it then converts the Hamiltonian to Dense as well, and steady-state solving taking a bit longer too.

Of course I can manually make those operators sparse (and then v5 just takes 7 mins!), but more generally, I think the logic of default_dtype is nice;  essentially allows you to run QuTiP in that 'mode'. Also useful for cases like making jax objects, etc. 

[...]

PR #2519, opened 2024-08-27
Apply `default_dtype` in Qobj and operation.

Ericgig:
**Description**
Add a new options that control where the `default_dtype` is taken into account: `default_dtype_range` (feel free to propose better name.)
It has 3 possible values:
- "creation": At `Qobj` creation functions, behaviour we presently have.
- "missing": Missing specialization will output that type: `Dense + Jax` is not defined so will default to that type.
- "full": Every operations return that type: `Dense + Dense` would convert the result to the `default_dtype`. Unary operations will also convert the output type. It could break some functions that explicitly set the date type such as `steadystate` or `HEOM`...

Also added control of the created format in `Qobj.__init__`: added `dtype`, when the initial data is a list of list, use `default_dtype`, with `"full"` also use `default_dtype`. Only using the default value with list normally is so solver states are not converted automatically which would slow the solver unexpectedly. 

I am looking into creating alias for a group of data layer. This would allow to set the default to jax and jaxdiag at once. Used with `"full"`, this would ensure only jax compatible data layers are used while not forcing the sparcity. This would hopefully make `"full"` usable.

**Related issues or PRs**
#2328


Ericgig:
@nwlambert 
I tried to address the issue of defautl dtype not being apply where expected here.
It probably won't affect the brmesolve issue but I would like you opinion on the approach.

nwlambert:
thanks eric, I like the option of creation and full. Took me a moment to understand the meaning of the ''missing'' option.  is there some clear situation where someone would use this?  perhaps there is a better name too?  

for `default_dtype_range`, how about `default_dtype_scope`?

i will try and play with the PR, see if it does what i would expect

Ericgig:
I don't think either "missing" or "full" should be used much with pure qutip. Only with plugins like qutip-jax it makes any sense to me. With "missing", you could set it so it will end up all in jax format and not have too much issues. 

"full" will probably break HEOM, some steadystate feature, slow down scipy integrators, etc. "missing? should be safe.

Ok for `default_dtype_scope`.

nwlambert:
I had a quick play with the feature, for the cases I had stumbled on before, with `eigenstates()' ignoring default_dtype, using the scope of 'full' covers it now. I understand having this work with just 'creation' scope is problematic because you want to make sure the dtype is not used on results lists.

heom+full breaks with dia and dense, as expected!

speaking of heom, manually converting the RHS to jaxdia and doing stuff it worked fine. i think generalizing the RHS construction in heom away from csr is probably very difficult without huge performance drop in the construction, so perhaps some kind of wrapper or option to do conversion of types after construction would be fine?

Ericgig:
I tried running tests with scope=full and it has priority over manually entered dtype values...
Not that using object in those other type is really possible with that options, it still not great.

hodgestar:
For the HEOM, perhaps the HEOM solver could gain an option to select the dtype to convert the RHS to?

nwlambert (approved):
I like the fixes for heom and steadystate.  Personally the matmal dense sparse fix in #2611 fixed my main issue with unexpected dtypes, but I think this can still be useful for people looking for performance bugs and wanting to enforce default dtype more broadly.  I will approve, but @Ericgig if you are having second thoughts on it, will leave it up to you.

also, is it my imagination or does this PR slightly slow down the tests by 1-2 minutes? maybe acceptable, and perhaps just my imagination.

PR #2523, opened 2024-09-04
Improve and update documentation

Ericgig:
**Description**
- Update versions in requirements.txt
- Update the contributor list and add Paul to the developers team.
  - I used the list generated by git, thus only authors of merged commit in `master` are present, with their `git.config` name instead of github name.
- Unified changelog title format.
  - Before  towncrier, the title used `Version X.X.X`, which was changed to `QuTiP X.X.X`. 
  - Also the title formating (`+++`) was not caught as a section and did not appear into the table of content.
- Removed the gallery: it was empty zip files.
- Removed the `qutip_tree` image. It was fancy, but useless and was still at 4.6.
- Improved look of the apidoc.
  - Ensure each property take a line, not side by side.
  - Added type hints aliases (`DimensionsLike`, ...)
  - Added support for aliases in more files (`import annotations`)
  - Split long signature into multiple lines, one per parameter. We have some long signature, made longer with type hints, this make them a lot more readable. (The readthedocs style was not great for that, so it was patched in site.ccs.)
    - Another options would be not to include the hints in the signature of the apidoc, they are already listed in each entry of the parameter section.


pmenczel (approved):
Everything looks quite good to me.

Looking at the documentation I just noticed that the "family packages" are not mentioned or linked anywhere, and that the optimal control part is out of date. But that's probably out of scope of this PR, I have opened an issue for now: https://github.com/qutip/qutip/issues/2524

doc/guide/guide-overview.rst
```diff
@@ -4,25 +4,7 @@
...
 ******************
 
-The goal of this guide is to introduce you to the basic structures and functions that make up QuTiP. This guide is divided up into several 
-sections, each highlighting a specific set of functionalities. In combination with the examples that can be found on the project 
-web page `https://qutip.org/tutorials.html <https://qutip.org/tutorials.html>`_, this guide should provide a more or less complete overview 
+The goal of this guide is to introduce you to the basic structures and functions that make up QuTiP. This guide is divided up into several
+sections, each highlighting a specific set of functionalities. In combination with the examples that can be found on the project
+web page `https://qutip.org/tutorials.html <https://qutip.org/tutorials.html>`_, this guide should provide a more or less complete overview
 of QuTip. We also provide the API documentation in :ref:`apidoc`.
-
```
hodgestar, line 11: It would be nice to keep the overview and there is code for generating it in `doc/QuTiP_tree_plot`. Perhaps that could be updated and perhaps automated a bit more?

hodgestar:
Thanks @Ericgig. The clean up was long overdue. There is also some strange stuff in `doc/contrib`. Perhaps it could be removed too?

doc/guide/guide-overview.rst
Ericgig: This circular format might have been nice 10 years ago, but with the size that qutip is now, it's an unreadable mess... Just updating it is not enough to make it worth keeping.

Ericgig:
The stuff in `doc/contrib` does not seems to have been used anywhere. I erased it and not error/warnings appeared, but can't tell you what it was used for.

doc/guide/guide-overview.rst
hodgestar: Then perhaps remove the rest of it? I.e. `doc/QuTiP_tree_plot/`.

hodgestar: Thanks!

hodgestar (approved):

PR #2526, opened 2024-09-06
update version in rtd-environmnent

Ericgig:
**Description**
In #2523, I updated versions for the sphinx build in `doc/requirements.txt`, but readthedocs has it's own version requirements in `doc/rtd-environment.yml` which I forgot to update.

PR #2527, opened 2024-09-06
Use versions available on conda

Ericgig:
**Description**
When building the documentation for the action (on locally on my side), we install with `pip`, however with readthedocs, we use manba. Some packages versions are not available on conda-forge, just pypi, causing the rtd build to fall with #2526...


hodgestar:
If needed, you can also install some things with `pip` in the `pip:` section at the bottom.

hodgestar (approved):

PR #2528, opened 2024-09-10
QobjEvo support in expect

Ericgig:
**Description**
Looking at #2511, we would need to compute the expectation values for time dependent operator and their derivative.

Here I add support for `QobjEvo` in `expect`: `expect(op: QobjEvo, state: QobjEvo) -> Coefficient`.
(List of operators or states not supported.)

### DIFF
--- a/qutip/core/expect.py
+++ b/qutip/core/expect.py
@@ -159,6 +159,82 @@ def _single_qobjevo_expect(oper, state):
     return out_coeff
 
 
+def _single_qobjevo_expect(oper, state):
+    oper = QobjEvo(oper)
+    state = QobjEvo(state)
+    if not isoper(state):
+        state = state @ state.dag()
+
+    op_list = oper.to_list()
+    state_list = state.to_list()
+
+    out_coeff = coefficient(0.)
+
+    for op, rho in itertools.product(op_list, state_list):
+        if isinstance(op, Qobj):
+            op = [op, coefficient(1.)]
+        if isinstance(rho, Qobj):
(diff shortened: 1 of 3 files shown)


qutip/core/expect.py
```diff
@@ -98,6 +112,129 @@ def _single_qobj_expect(oper, state):
...
+
+            def _qevo_both_expect(t):
+                return expect(op[0](t, **op[1]), rho[0](t, **rho[1]))
+
+            out_coeff = out_coeff + coefficient(_qevo_both_expect)
+
+    return out_coeff
+
+
+def _single_qobjevo_expect(oper, state):
```
hodgestar, line 162: This function seems to appear twice?

qutip/core/expect.py
```diff
@@ -98,6 +112,129 @@ def _single_qobj_expect(oper, state):
...
+
+        if isinstance(op[0], Qobj) and isinstance(rho[0], Qobj):
+            out_coeff = out_coeff + coefficient(
+                _single_qobj_expect(op[0], rho[0])
+            ) * op[1] * rho[1]
+
+        # One of the QobjEvo is in the function format: QobjEvo(lambda t, **kw: Qobj(...)
+        elif isinstance(rho[0], Qobj):
+
+            def _qevo_oper_expect(t):
```
hodgestar, line 140: Can `QobjEvo` arguments be handled here somehow?

qutip/core/expect.py
Ericgig: Sorry.

PR #2497, opened 2024-07-22
Better mkl lookup

Ericgig:
**Description**
The mkl finding was sometime missing the library even when available.
Sometime the library can has a version number in the file name:
`libmkl_rt.so.2`, `mkl_rt.2.dll`, with or without a symbolic link to the version less name.
Also it ouly look for it if it's found in numpy's config, but the module mkl is available an could be installed even if numpy does not link to it. Our use using ctypes does not require numpy to be using it to work.

So I changed the lookup function to look for the library without checking is numpy uses it. I also added a new setting, `settings.mkl_lib_location`, which is not read only and allow users to enter the path to their mkl implementation. We only look inside the python library path, this would allow to use libraries installed at the OS level.

Also fixed a bug in mkl spsolve with scipy 1.14 that was missed since mkl tests were skipped.

**Related issues or PRs**
fix #2496

PR #2529, opened 2024-09-10
Fix failing test with mkl windows

Ericgig:
**Description**
Since merging #2497, mkl tests on windows fails from segfault.
I am not sure of the source of the error:
Since the library was not found before the PR, the bug can have been there for a long time, but the tests passed in #2497.
Just settings mkl to verbose is enough for the tests to passes...

I don't think relying on the verbose options is a good idea so I made it unable to find mkl on windows for this PR.

We will need to find a more reliable way to use it, `ctypes` does not look at the header file so it guess the types from our call. Maybe some version of mkl are 32 bit and other 64 bit and we are not adapting.

 [pydiso](https://github.com/simpeg/pydiso) looks promising.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -211,6 +211,7 @@ jobs:
         # to see if something hung.
         timeout-minutes: 60
         run: |
+          export MKL_VERBOSE=2
           if [[ -n "${{ matrix.openmp }}" ]]; then
             # Force OpenMP runs to use more threads, even if there aren't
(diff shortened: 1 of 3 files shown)


hodgestar (approved):

hodgestar:
I would agree that ctypes is the likely cause of the segfault. It doesn't examine the headers so there can easily be a mismatch between the types it guessed and the types used in the shared library. Pydiso sounds like a reasonable approach.

PR #2531, opened 2024-09-12
Reorder api-doc

Ericgig:
**Description**
The api splitting functions and classes is not very intuitive. 
I reordered it so that similar feature are grouped together: `MESolver` follows `mesolve`, etc.


hodgestar (approved):
This looks great, @Ericgig.

Issue #1278, opened 2020-06-01
Qobj Data Layer Specification and Tracking


jakelishman:
_Last updated: 2020-08-12_

This is an issue for tracking and publicising the development status of a separated data layer.  This is a major breaking change to be introduced in QuTiP 5, and will change how library functions and users access the underlying data structures of `Qobj`.  This document and issue will change over time.

Development is taking place on the [`dev.major` branch](https://github.com/qutip/qutip/tree/dev.major) in `qutip/qutip`.  Pull requests implementing new features:
- #1282 separating out the core QuTiP operations into a (logical but user-transparent) package `qutip.core`.
- #1296 implementing the first parts of the data multiple dispatch and new data types.
- #1332 replacing the `fast_csr_matrix` with the new Cython type `CSR` all across QuTiP
- #1338 implementing the data-layer creation, conversion and dispatcher routines.

We are writing a new "developers' guide" to explain the logic behind the design decisions, and help new QuTiP developers work on the code.  You can follow initial development hosted as [a repo in my user account](https://github.com/jakelishman/qutip-devguide) until we decide how this will be added to the main documentation.  You can view a rendered version of the documentation at https://jakelishman.github.io/qutip-devguide.

Some design documents:
- [data layer](https://github.com/jakelishman/qutip-doc/blob/design-data-layer/design/01-data-time-layer-separation.md)
- [data structure implementation choices](https://github.com/jakelishman/qutip-doc/blob/design-data-layer/design/02-data-structures.md)
- [original GSoC proposal (PDF)](https://www.binhbar.com/posts/2020/05/welcome-to-google-summer-of-code-2020/proposal.pdf)

New design elements under consideration:
 - [improved `'type'` and `'dims'` handling for tensor-product spaces](https://github.com/qutip/qutip/issues/1320)

Major breaking changes to QuTiP:
 - `Qobj.data` is no longer an instance of `scipy.sparse.spmatrix`, but the new abstract "data-layer type" `core.data.Data`
 - `bra * ket` now returns a scalar, not a `Qobj` (no need to do `(bra * ket).data[0, 0]` or `(bra * ket).tr()`) any more
 - 1D subspaces in `Qobj.dims` are now never contracted silently (they were, but inconsistently before), but the new function `Qobj.contract()` can be called to do this explicitly
[...]

PR #2533, opened 2024-09-18
Add __array__ method to Qobj

gautierronan:
Proposal to add the `__array__` method to Qobj, such that one can call `np.asarray(some_qobj)`.


Ericgig:
We had it before and it introduced issues that lead us to remove it, see #1278, #1433, #1611...

It also limits us to what is a `Qobj` which we have ideas extend.

PR #2534, opened 2024-09-18
Adding classes to characterize Bosonic Enviroments

gsuarezr:
**Description**
The PR provides a way to characterize thermal environments for open quantum systems, The user can now construct Thermal environments by specifying The spectral density, or the correlation function , or the power spectrum and the Temperature. Once specified the other two are calculated.

We hope once integrated with the solvers, this PR will allow for a more systematic way of comparing different approaches, and easier simulations by quickly obtaining the quantities needed from the bath objects.

The PR also provides exponential approximations of the environment via a fit of the correlation function, and a fit of the spectral density in terms of underdamped modes, the PR allows for incorporation of more approximation methods, which will be added in the future. 

**Related issues or PRs**
This PR is a generalization of #2260  . Special Thanks to @pmenczel  for improving the code significantly and valuable suggestions!

**TODO:**

I think perhaps those are better finished once we have decided that this structure for the environments is the way to go. However, the most important missing bits are 

- [x] Finish documentation in the User's guide
- [x] Add relevant tests for the HEOM solver using the enviroment class
- [x] Finish tests for environment.py


pmenczel:
Thank you for all your work @gsuarezr 

Just for a bit more context. Qutip currently contains in the HEOM module "BosonicBath / FermionicBath" classes that describe a bath with a multi-exponential decomposition of the correlation function. This new environment module is supposed to be a more general framework to characterize thermal environments. Environments with multi-exponential decompositions are included as `ExponentialBosonicEnvironment`; this class is supposed to mostly replace the old HEOM `BosonicBath`. We now use the term "environment" instead of "bath" mostly to avoid having the same name for the new classes as in the HEOM module.

There is a little difference between the old and the new class, namely that the system coupling operator is a part of BosonicBath (i.e., an attribute of BosonicBath and required in the constructor), but we have decided for the new module that system operators should not be part of the environment. This means that to create a HEOM solver, one will need to specify environments and coupling operators, somewhat similar to the `brmesolve` solver where one specifies both power spectrum and `a_op`. (Currently, one just specifies a list of "baths" in the HEOM solver). This will require some small changes to the HEOM solver interface that we still have to implement before merging this PR.

Another thing missing from this PR are fermionic environments. I would have liked to have them, but it was becoming too much work and a too large PR, so fermionic environments will have to be added a bit later.

Some more advantage of our changes are:
* The new environment classes could be used directly as inputs for solvers other than the HEOM solver, for example for `mesolve` or `brmesolve`. These solvers could compute their jump operators and rates from the environment power spectrum and system operators. (That is however also left for a future PR, since it would require some thinking about how environments can best be passed to those solvers.) Similarly, QuTiP could automatically compute Markovian embeddings with pseudomodes or reaction coordinates.
* The module provides a framework for the idea of approximating one environment by another (multi-exponential) one, and more approximation methods could be added in the future. For example, Gerardo has been working on a method fitting the power spectrum using the so-called AAA algorithm.

qutip/core/environment.py
```diff
@@ -1819,22 +1819,23 @@ def _fft(f, wMax, tMax):
 
 
 def _cf_real_fit_model(tlist, a, b, c, d=0):
-    return np.real((a + 1j * d) * np.exp((b + 1j * c) * tlist))
+    return np.real((a + 1j * d) * np.exp((b + 1j * c) * np.abs(tlist)))
 
 
 def _cf_imag_fit_model(tlist, a, b, c, d=0):
-    return np.imag((a + 1j * d) * np.exp((b + 1j * c) * tlist))
+    return np.imag((a + 1j * d) * np.exp((b + 1j * c) * np.abs(tlist)))
```
pmenczel, line 1826: Shouldn't it be
```suggestion
    return np.sign(tlist) * np.imag(
        (a + 1j * d) * np.exp((b + 1j * c) * np.abs(tlist))
    )
```

qutip/utilities.py
```diff
@@ -404,6 +404,12 @@ def iterated_fit(
         `[[p11, ..., p1n], [p21, ..., p2n], ..., [pN1, ..., pNn]]`.
     """
 
+    # Check if array is constant
+    # if (ydata == ydata[0]).all():
+    #     if num_params == 3:
+    #         return [ydata[0]], [0], [0]
+    #     else:
+    #         return [ydata[0]], [0], [0], [0]
```
pmenczel, line 412: What is this doing here?

qutip/utilities.py
```diff
@@ -465,6 +471,8 @@ def _rmse(fun, xdata, ydata, params):
     parameters. (The closer to zero = the better the fit.)
     """
     yhat = _evaluate(fun, xdata, params)
+    if (np.max(ydata) - np.min(ydata)) == 0.0:
+        return 0
```
pmenczel, line 475: What are you trying to do here? The rmse is not necessarily zero if the data is constant

qutip/core/environment.py
```diff
@@ -0,0 +1,2015 @@
...
+
+    def correlation_function(self, t, **kwargs):
+        t = np.array(t, dtype=float)
+        result = np.zeros_like(t, dtype=complex)
+        positive_mask = (t >= 0)
+        non_positive_mask = np.invert(positive_mask)
+
+        result[positive_mask] = self._cf(t[positive_mask])
+        result[non_positive_mask] = np.conj(
+            self._cf(-t[non_positive_mask])
```
nwlambert, line 703: just a reminder, this still breaks from_correlation_function() baths defined only for positive times

qutip/core/environment.py
```diff
@@ -0,0 +1,2015 @@
...
+
+    @classmethod
+    def from_correlation_function(
+        cls,
+        C: Callable[[float], complex] | ArrayLike,
+        tlist: ArrayLike = None,
+        tMax: float = None,
+        *,
+        T: float = None,
+        tag: Any = None
```
nwlambert, line 171: would be nice to be able to pass combine=True/False here too

nwlambert:
Just another small issue, I found that if I use UnderDampedEnvironment.spectral_density() with brmesolve(), it fails in numpy 2.1.1 unless I wrap the SD function in another function.  Strangely works in numpy 1.26.4.     Seems to some difference in the way zero-dimensional arrays are treated, but not super sure.

qutip/core/environment.py
```diff
@@ -0,0 +1,2015 @@
...
+                      tMax, tMax=wMax)
+        return result(t) / (2 * np.pi)
+
+    # --- fitting
+
+    def approx_by_cf_fit(
+        self, tlist: ArrayLike, target_rsme: float = 2e-5, Nr_max: int = 10,
+        Ni_max: int = 10, guess: list[float] = None, lower: list[float] = None,
+        upper: list[float] = None, full_ansatz: bool = False, tag: Any = None
+    ) -> tuple[ExponentialBosonicEnvironment, dict[str, Any]]:
```
nwlambert, line 351: can we support passsing more options to curve_fit, particularly sigma and max_fev?

qutip/core/environment.py
pmenczel: Did you mean to comment this somewhere else?

qutip/core/environment.py
pmenczel: Ah I left out that `sigma` because there were so many parameters already and I couldn't imagine it being useful, but if it is, sure, we can add it back in!

gsuarezr: Sure, I'll add `sigma` back, and `add max_fev`, which it's just hardcoded now :)

qutip/core/environment.py
gsuarezr: That's True!

qutip/utilities.py
gsuarezr: Sorry, Forgot to delete the comment 😞  , will delete in the next commit

qutip/core/environment.py
nwlambert: Ah yeah, I meant for the cf_from_fit function, and anything else that returns an exponential environment.  Though I imagine we might adjust that at some point, but I like having control of the combine parameter in heom stuff

qutip/core/environment.py
nwlambert: Yeah, in some cases I found controlling sigma pretty useful

gsuarezr:
> Just another small issue, I found that if I use UnderDampedEnvironment.spectral_density() with brmesolve(), it fails in numpy 2.1.1 unless I wrap the SD function in another function. Strangely works without needing to wrap it in numpy 1.26.4. Seems to some difference in the way zero-dimensional arrays are treated, but not super sure.

I noticed that as well,.I didn't fix it because we are not integrating with the solvers yet, but could check if the argument of correlation function is an array, and if it's not do the .item() bit, just so that brmesolve works smoothly

qutip/core/environment.py
gsuarezr: Hey, 

Maybe we could discuss this point! I think the issue you were facing was related to the fitting when the function you pass is all zeros and not this bit.  I sent a notebook with that example working after the previous commit through email, it's the email where I mentioned I had a problem with Bloch-Redfield in the examples and that I forgot to add combine to functions in this commit

Both Paul's previous notebook and the tests here are done for positive times only! I did say that may be the issue when you mentioned it on line, but I should have thought a bit more before replying

If this is still an issue let me know!

qutip/utilities.py
gsuarezr: Just avoiding outputting rmse being nan.  It is the case for the fittings considered, but I can just change it to  

```
if (yhat == ydata).all(): 
    return 0
```
To keep _rmse fully general

pmenczel: Aah no, I guess it is fine as it is, thank you for clarifying!

qutip/core/environment.py
```diff
@@ -0,0 +1,2017 @@
...
+    t0 = tlist[np.argmax(corr_abs)]
+
+    # Checks if constant array, and assigns zero
+    if (clist == clist[0]).all():
+        if full_ansatz:
+            return [[0] * 4]*3
+        return [[0] * 3]*3
+
+    if full_ansatz:
+        lower = [-100 * corr_max, -np.inf, -1, -100 * corr_max]
```
nwlambert, line 1843: just remembered, some of these bounds are a bit odd, why is c bound to +- 1?  the others are kinda reasonable in full_ansatz i guess

qutip/core/environment.py
```diff
@@ -0,0 +1,2017 @@
...
+        guess = [corr_max, -100*corr_max, 0, 0]
+        upper = [100*corr_max, 0, 1, 100*corr_max]
+    else:
+        lower = [-20 * corr_max, -np.inf, 0]
+        guess = [corr_max, -t0, t0]
+        upper = [20 * corr_max, 0.1, np.inf]
```
nwlambert, line 1849: did you have some intuition for using t0 as a guess here?   for some examples I tried on monday it would make the the fit unfeasible

gsuarezr: Didn't notice $w_{c}$ got renamed to $t_{0}$. The intuition, is that the parameter corresponds to the inverse of the cutoff time (inverse of the cutoff frequency). Not sure why it would be  unfeasible, I think it should always be feasible. Unless it was with negative times for which the model functions had divergent exponents, which should be fixed now ! 

But if you share one of those examples I can take a look!

qutip/core/environment.py
gsuarezr: No particular reason, it should be $\pm \pi$ . will fix it on the next push. The same is true where `full_ansatz=False`. The more sensible bound is  $\pm \pi$

doc/apidoc/classes.rst
```diff
@@ -45,6 +45,37 @@ Distributions
...
+------------
+
+.. autoclass:: qutip.core.BosonicEnvironment
+    :members:
+
+.. autoclass:: qutip.core.DrudeLorentzEnvironment
+    :members:
+    :inherited-members:
+    :show-inheritance:
+    :exclude-members: from_correlation_function, from_power_spectrum, from_spectral_density
```
hodgestar, line 58: Small question -- why are these alternative constructors being excluded? Wouldn't users want to see them? Are they shown somwhere else perhaps?

hodgestar (commented):
@pmenczel @gsuarezr The structure looks great now. Thank you to both of you for all the work you've put in to make the bath classes happen.

doc/apidoc/classes.rst
pmenczel: They are shown in the base class, `BosonicEnvironment`. It would not make much sense to call them on the specialized subclasses like `DrudeLorentzEnvironment`, where the correlation function, power spectrum and spectral density are all already determined. (Does that mean these alternative constructors should be `staticmethod` rather than `classmethod`?)

hodgestar: Regarding the docs: Ah, I see.

Its traditional to make these alternative constructors class methods, because they usually return an new instance of the class they're attached to. The situation here is a bit odd because these helper methods return instances of sub-classes (`_BosonicEnvironment_fromSD` and friends). It's also odd because these methods will also appear on all the sub-classes too.

I don't think it's necessarily bad how it is, perhaps just a bit surprising.

qutip/core/environment.py
```diff
@@ -0,0 +1,2722 @@
...
+    corr_max = np.max(np.abs(clist))
+
+    # Checks if constant array, and assigns zero
+    if (clist == clist[0]).all():
+        if full_ansatz:
+            return [[0] * 4]*3
+        return [[0] * 3]*3
+
+    if full_ansatz:
+        lower = [-100 * corr_max, -np.inf, -1, -100 * corr_max]
```
nwlambert, line 1902: the frequency bounds here need changing too?

pmenczel:
Thank you for the updates, @gsuarezr !

To give a brief summary of the progress: on our todo-list we have the following items left:
* Neill's comment above and one TODO comment I left somewhere in the code earlier
* Finishing tests for the environment module and the fitting utility function
* Adding tests for the HEOM solver using the new API
* Updating (at least some of) the HEOM tutorial notebooks to show off the new features

The main code and documentation is converging from our side, and I think ready for more detailed reviews.

qutip/utilities.py
```diff
@@ -331,3 +342,178 @@ def _version2int(version_string):
...
+#
+
+def iterated_fit(
+    fun: Callable[..., complex], num_params: int,
+    xdata: ArrayLike, ydata: ArrayLike,
+    target_rmse: float = 1e-5,
+    guess: ArrayLike | Callable[[int], ArrayLike] = None,
+    Nmin: int = 1, Nmax: int = 10,
+    lower: ArrayLike = None, upper: ArrayLike = None,
+    sigma: float | ArrayLike = None, maxfev: int = None
```
nwlambert, line 358: just a minor thing, previously maxfev was hardcoded.  i think its better as here, to use default and allow user to define, but  the default is insufficient for some of the paper examples, so just a note that when updating those examples we have to up it

qutip/solver/brmesolve.py
```diff
@@ -198,17 +199,23 @@ def brmesolve(
...
         if isinstance(spectra, str):
             new_a_ops.append(
-                (a_op, coefficient(spectra, args={**args, 'w':0})))
+                (a_op, coefficient(spectra, args={**args, 'w': 0})))
         elif isinstance(spectra, InterCoefficient):
             new_a_ops.append((a_op, SpectraCoefficient(spectra)))
         elif isinstance(spectra, Coefficient):
             new_a_ops.append((a_op, spectra))
+        elif (isinstance(spectra, BosonicEnvironment) or
+              isinstance(spectra, FermionicEnvironment)):
```
nwlambert, line 208: maybe better not to support fermionic environments here yet (brmesolve needs hermitian coupling operators, and fermionicenvironments have two different power spectra)

qutip/core/environment.py
```diff
@@ -850,12 +850,14 @@ class DrudeLorentzEnvironment(BosonicEnvironment):
     """
 
     def __init__(
-        self, T: float, lam: float, gamma: float, *, tag: Any = None
+        self, T: float, lam: float, gamma: float, *, tag: Any = None,
+        Nk: int = None
```
pmenczel, line 854: Thank you, it is a good solution to be able to set `Nk` here. But a few things for polishing:

- Could we please document the parameter `Nk` in the docstring above?
- In `correlation_function`, if `self.Nk` is not None, the parameter `Nk` will now be ignored. Two options: (1) Remove the `Nk` parameter there. (2) If we have that parameter in both places, if the user passes `Nk` to the function, that should always take precedence over `self.Nk`.
- And could we modify the `correlation_function` docstring accordingly?
- All of this should be changed analogously in the underdamped and Lorentzian environments.

qutip/solver/brmesolve.py
gsuarezr: That's True, I meant that last commit to find out which version of numpy needed the .item() bit, I'll try adapting it to properly do fermionic this weekend, if not I'll keep the BosonicEnvironment only (Possibly also worth accepting bosonic baths as well ?)

qutip/core/environment.py
gsuarezr: Thanks for the review! 

I think the underdamped doesn't need it because there we obtain the correlation function from the power spectrum. But I did include it in docstrings and modified it for the Lorentzian one.

pmenczel: Oh, you are right about the underdamped of course.

About the number of parameters, I was thinking about that as well and I am a bit unsure. But I guess if we know exactly which parameters we accept, it is better not to use `kwargs`, even if the parameter list becomes long. Because with `kwargs`, we don't have autocomplete / information from the IDE. (Also, I don't know how to make the documentation look nice)

gsuarezr: In that case I'll just undo the kwargs bit. I just thought it would be nice to have because codeclimate complains about it constantly. 

The only other bit that I wanted to modify while updating the tutorials is the fact that the from_spectral_density constructor (correlation function and power spectrum as well), accept a function with only one parameter w (or t) and I thought it would be nice if they have an args argument, so that we could pass the other parameters to it.

Something like 

def ohmic_spectral_density(w, alpha, wc):
    """The Ohmic bath spectral density as a function of w
    (and the bath parameters).
    """
    return w * alpha * np.exp(-w / wc)

sd_env2=BosonicEnvironment.from_spectral_density(J=ohmic_spectral_density,wlist=w,args=(alpha,wc))

Though again not sure, if that's wanted. I find that to be nice though

I'm pretty much done with the tutorials, I guess that If I go through the User's guide and don't have anythinf to add I should just mark this as "ready for review" so that it is no longer a draft PR?

pmenczel: Adding an args parameter there would be nice! :)

> I'm pretty much done with the tutorials, I guess that If I go through the User's guide and don't have anythinf to add I should just mark this as "ready for review" so that it is no longer a draft PR?

Sounds good 🥳

qutip/solver/brmesolve.py
```diff
@@ -198,17 +199,22 @@ def brmesolve(
...
             new_a_ops.append(
-                (a_op, coefficient(spectra, args={**args, 'w':0})))
+                (a_op, coefficient(spectra, args={**args, 'w': 0})))
         elif isinstance(spectra, InterCoefficient):
             new_a_ops.append((a_op, SpectraCoefficient(spectra)))
         elif isinstance(spectra, Coefficient):
             new_a_ops.append((a_op, spectra))
+        elif isinstance(spectra, BosonicEnvironment):
+            spec = SpectraCoefficient(
+                coefficient(lambda w: spectra.power_spectrum(w))
```
hodgestar, line 209:
```suggestion
                coefficient(spectra.power_spectrum)
```
Is there something that requires the use of a lambda here?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -726,6 +752,39 @@ def _combine_bath_exponents(self, bath):
...
+        contains_bath = any(
+            isinstance(elem, (Bath, BosonicBath, FermionicBath))
+            for elem in bath_spec
+        )
+        if contains_bath:
+            # Potentially confused user, passed mixed list of environments
+            # and baths. Error will be raised elsewhere.
+            return False
+
+        return True
```
hodgestar, line 775:
```suggestion
        if not isinstance(bath_spec, (list, tuple)) or len(bath_spec) != 2:
            return False
        env, Q = bath_spec

        if not isinstance(env, (BosonicEnvironment, FermionicEnvironment)):
            return False

        if not isinstance(q, (Qobj, QobjEvo):
            return False

        return True
```
I'm not sure if there needs to be a slightly more complex check for whether the `env` is the right kind of bath or not, but the length and ordering of the expected pair should definitely be checked, yes?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -712,12 +729,21 @@ def _initialize_stats(self):
...
+        exponents = []
+        for b in bath:
+            if self._is_environment_api(b):
+                b = self._env_to_bath(b)
+            if (isinstance(b, (BosonicEnvironment, FermionicEnvironment))
+                    and not isinstance(b, (BosonicBath, FermionicBath))):
+                raise ValueError("Environments must be passed with their"
+                                 " corresponding coupling operator as a list"
+                                 " or tuple (env, Q)")
+            exponents.extend(b.exponents)
```
hodgestar, line 745:
```suggestion
        # Only one bath provided, not a list of baths
        if (not isinstance(bath, (list, tuple)) or self._is_environment_api(bath)):
            bath = [bath]
        bath = [self._to_bath(b) for b in bath]
        exponents = []
        for b in bath:
            exponents.extend(b.exponents)
```

qutip/solver/heom/bofin_solvers.py
```diff
@@ -726,6 +752,39 @@ def _combine_bath_exponents(self, bath):
...
+    def _env_to_bath(self, bath_spec):
+        if isinstance(bath_spec[0], ExponentialBosonicEnvironment):
+            return BosonicBath.from_environment(*bath_spec)
+        if isinstance(bath_spec[0], ExponentialFermionicEnvironment):
+            return FermionicBath.from_environment(*bath_spec)
+        else:
+            raise ValueError("The HEOM solver requires the environment to have"
+                             " a multi-exponential correlation function. Use"
+                             " one of the `approx_by_` functions to generate a"
+                             " multi-exponential approximation.")
```
hodgestar, line 786:
```suggestion
    def _to_bath(self, bath_spec):
        if isinstance(bath_spec, Bath):
            return bath_spec
        if not self._is_environment_api(bath_spec):
            raise ValueError(
                "Environments must be passed as either Bath instances or"
                " as a tuple or list corresponding to an environment and a"
                " coupling operator, (env, Q)"
            )
        env, Q = bath_spec
        if isinstance(env, ExponentialBosonicEnvironment):
            return BosonicBath.from_environment(env, Q)
        if isinstance(env, ExponentialFermionicEnvironment):
            return FermionicBath.from_environment(env, Q)
          raise ValueError("The HEOM solver requires the environment to have"
                           " a multi-exponential correlation function. Use"
                           " one of the `approx_by_` functions to generate a"
                           " multi-exponential approximation.")
```

hodgestar: These suggestions are to try simplify and tighten up the checking. Previously it was not clear to me that strange things wouldn't happen if a user passed empty, 0, 1 or 3+ element lists or lists with the env and Q in unexpected places.

Not sure if I missed any details. Hopefully the suggestion helps.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -1186,11 +1245,6 @@ class HSolverDL(HEOMSolver):
     See :class:`HEOMSolver` and :class:`DrudeLorentzBath` for more
     descriptions of the underlying solver and bath construction.
 
-    An exact copy of the QuTiP 4.6 HSolverDL is provided in
-    ``qutip.nonmarkov.dlheom_solver`` for cases where the functionality of
-    the older solver is required. The older solver will be completely
-    removed in QuTiP 5.
```
hodgestar, line 1192: Whoops. Should have removed this not in QuTiP 5. Thanks!

qutip/utilities.py
hodgestar: These parameters should also be on separate lines. It took me a few seconds to find `maxfev`.

hodgestar (approved):
I left some suggestions but this looks great. Thanks @pmenczel and @gsuarezr.

qutip/solver/brmesolve.py
pmenczel: I don't think so, changed

qutip/solver/heom/bofin_solvers.py
pmenczel: Thank you, I have implemented your suggestion, makes it clearer. But to clarify, I intentionally allowed lists with more than 2 entries, and I'm passing extra arguments to `BosonicBath.from_environment`. The only purpose of this is, currently, that one can pass the `dim` parameter as the third entry. (This isn't documented since a) I don't really know in which situation one needs that parameter and b) I expect this will rather be used internally.)

PR #2542, opened 2024-10-10
Fix towncrier check

Ericgig:
**Description**
The towncrier check was failing when installing the module since we used the OS python and which is not managed by pip:
```
python -m pip install towncrier
error: externally-managed-environment
```


.github/workflows/tests.yml
```diff
@@ -266,6 +266,8 @@ jobs:
 
       - name: Install Towncrier
         run: |
+          python -m venv towncrier_check
+          source towncrier_check/bin/activate
```
hodgestar, line 270: Would `pip install --user towncrier` work?

Ericgig: I did not try, but from what I understand in https://github.com/pypa/pip/issues/12085, it should not work.
`--break-system-packages` would work and towncrier should be safe, but environment is the official solution.

hodgestar (approved):
Thanks, @Ericgig.

PR #2545, opened 2024-10-14
Add einsum

### LINKED ISSUE #2340: Add `einsum` for `Qobj`.
### Problem Description

Having a version of `einsum` working on `Qobj`'s dimensions would be useful (qutip/qutip-qip#225).
A dense implementation should be easy enough to develop using `qutip.core.dimension.to_tensor_rep`.

### Proposed Solution

Use `to_tensor_rep`, `from_tensor_rep` with `np.einsum` to create qutip version of einsum.

### Alternate Solutions

If would be even better to develop it without conversion to numpy array to allow support for sparse, jax, cupy, etc. matrices.
However this would probably not be an easy task doable in the scope of `good first issue`.

### Additional Context

_No response_

fmayo96:
**Description**
Added a einsum, a numpy.einsum (Evaluates the Einstein summation convention on the operands.) version for Qobj dimensions.

**Related issues or PRs**
fixes #2340

** Question: Not sure if I should update the documentation and how. Just in case I did not modify the docs yet.

### DIFF
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -351,6 +351,11 @@ def from_tensor_rep(tensorrep, dims):
 def _frozen(*args, **kwargs):
     raise RuntimeError("Dimension cannot be modified.")
 
+def einsum(subscripts, *operands, out=None, dtype=None, order='K', casting='safe', optimize=False):
+    operands_array = [to_tensor_rep(op) for op in operands]
+    result = np.einsum(subscripts, *operands_array)
+    dims = [[d] for d in result.shape]
+    return from_tensor_rep(result, dims)
 
 class MetaSpace(type):
     def __call__(cls, *args: SpaceLike, rep: str = None) -> "Space":
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -369,7 +369,7 @@ def einsum(subscripts, *operands):
     Qobj (numpy.complex128)
         Result of einsum as Qobj (numpy.complex128 if result is scalar)
     """
-    operands_array = [to_tensor_rep(op) for op in operands]
+    operands_array = [to_tensor_rep(op).reshape(op.shape) for op in operands]
     result = np.einsum(subscripts, *operands_array)
     if result.shape == ():
         return result
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -354,7 +354,7 @@ def _frozen(*args, **kwargs):
 def einsum(subscripts, *operands):
     """
     Implementation of numpy.einsum for Qobj.
-    Returns the result of einsum as a Qobj (or numpy.complex128 if the result is a scalar).
+    Evaluates the Einstein summation convention on the operands.
     Parameters
     ----------
     subscripts: str
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -351,9 +351,29 @@ def from_tensor_rep(tensorrep, dims):
 def _frozen(*args, **kwargs):
     raise RuntimeError("Dimension cannot be modified.")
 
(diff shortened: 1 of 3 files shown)


qutip/core/dimensions.py
```diff
@@ -352,6 +352,31 @@ def _frozen(*args, **kwargs):
...
+    Returns
+    -------
+    Qobj (numpy.complex128)
+        Result of einsum as Qobj (numpy.complex128 if result is scalar)
+    """
+    operands_array = [to_tensor_rep(op) for op in operands]
+    result = np.einsum(subscripts, *operands_array)
+    if result.shape == ():
+        return result
+    dims = [[d] for d in result.shape]
```
Ericgig, line 376: Not sure if this work well with compound dimensions.
Could you add tests with multiple qubits systems.

Ericgig (commented):
Thank you for tackling this issue.
Other than confirming the dims are good in multiple qubits systems, it looks good.

hodgestar:
Thanks @fmayo96.

Question: Should the `einsum` function be in `dimensions.py`? And should it be internal or exposed via the public API? It looks generically useful but what specific use cases do we have in mind?

BoxiLi:
The simplest use case is the circuit simulation, where e.g. single qubit unitaries are multiplied to multiqubit states. For Hamiltonian dynamics I cannot immediately think of an application cause usually you need to sum many terms. Unless we develop a troterrizatiom solver.

fmayo96:
I have corrected the code for compound systems and added tests for two and three qubits. I am not sure about which is the best place for this function, I made it to solve the Issue #2340. If you consider it fits better in a different place I will be happy to modify it

qutip/core/dimensions.py
Ericgig: Sometime like 
`dims = [[d for d in result.shape[:result.ndim // 2], [d for d in result.shape[result.ndim // 2:]]`
would keep the compound dimensions.
We should raise warning or error if the output ndim is odd.

Ericgig (commented):
Sorry, but the first attempt was better.
We want the `einsum` on the dimensions, not the shape. 
With a 2 qubits operator, there should be 4 indices, 6 indices for 3 qubits, etc.
The output for compound system should also be compound.
We want support for operations that could not be done with any mix of `matmul`, `trace`, etc.

Here are some example of use:

```
einsum("ikjl,jm->ikml", [tensor(sigmaz(), sigmaz()), sigmaz()]) == tensor(qeye(2), sigmaz())
einsum("ijkl->kjil", [tensor(sigmam(), sigmaz())]) == tensor(sigmap(), sigmaz())
```

There are no place for it, a new file (`qutip/core/einsum.py`) would be good, but it's fine to keep in `dimensions.py` for now. We will move it later.

fmayo96:
Oh, I am sorry, I have misunderstood what you wanted. I think that now I have fixed it. Thank you very much for the feedback!

qutip/core/dimensions.py
```diff
@@ -352,6 +352,32 @@ def _frozen(*args, **kwargs):
...
+    -------
+    Qobj (numpy.complex128)
+        Result of einsum as Qobj (numpy.complex128 if result is scalar)
+    """
+    operands_array = [to_tensor_rep(op) for op in operands]
+    result = np.einsum(subscripts, *operands_array)
+    if result.shape == ():
+        return result
+    dims = [d for d in result.shape[:result.ndim // 2]],
+    [[d for d in result.shape[result.ndim // 2:]]]
```
Ericgig, line 377:
```suggestion
    dims = [
        [d for d in result.shape[:result.ndim // 2]],
        [d for d in result.shape[result.ndim // 2:]]
    ]
```

Ericgig (approved):
Thank you

PR #2554, opened 2024-11-15
c_{k,real} should not have -i in the math section

ausstein:
**Description**
In the guide for using HEOM with a bosonic bath, in the description of the Drude Lorentz BCF the Matsubara terms are split between real and imaginary parts. However, in the maths section of the real part, $c_{k,real}$ the $-i$ term remains for $k=0$.  In the python section it is already correctly removed.
This was fixed

for k=0 changed  $` c_{k,real} = \lambda \gamma   [\cot(\beta \gamma / 2) - i] `$ to  $` c_{k,real} = \lambda \gamma   [\cot(\beta \gamma / 2)] `$

### DIFF
--- a/doc/guide/heom/bosonic.rst
+++ b/doc/guide/heom/bosonic.rst
@@ -271,7 +271,7 @@ coefficients of this expansion are, for the real part, :math:`C_{real}(t)`:
     \end{cases}
 
     c_{k,real} &= \begin{cases}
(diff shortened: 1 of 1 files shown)


pmenczel (approved):
Hi, thank you for contributing and for spotting this typo. You are right, that "-i" should not be there.

I should let you know that we are currently reworking this part of the documentation as part of PR #2534. But I will merge your change so that it is corrected in the docs already now.

PR #2565, opened 2024-12-02
Small tests fixes

Ericgig:
**Description**
- `mpmath` is optional, so skip tests that uses it when not installed.
This skip all tests in that file when missing. If some bath feature should be tested even when missing, we should skip only these tests .

- Remove `ket2dm(basis(...))` from `parametrize` in visualization tests. Function in parametrize decorator are called when tests are collected, while inside fixture they are call when tests are run.
Also use a larger state for the density matrix tests. `basis(1)` can be both `ket`, `bra` and `operator`... 

- Use `full` in distribution. When using a `fock_dm` instead of a `ket2dm(basis(...))`, `data_as` returned a `dia_matrix` which did not works in that case.


pmenczel (approved):
Thank you. Sorry for missing that the mpmath tests should be skipped, I meant to add that. I have changed it so that only the tests that require mpmath are skipped, hope that's okay.

PR #2568, opened 2024-12-05
Corrected column titles in SD fit info

pmenczel:
Mini fix to correct the column titles of the fit summary in spectral density fits.


nwlambert (approved):
thanks for the fix!

Issue #2567, opened 2024-12-04
Inefficient kraus_to_super implementation


Sola85:
### Problem Description

The implementation of `qutip.kraus_to_super` seems very inefficient as conversion happens via an intermediate Choi matrix, which is stored as a dense matrix. 

Is there any reason against using the below implementation (which seems much faster and does not convert to dense)?

### Proposed Solution

```python
import qutip as qt
from time import perf_counter

kraus_ops = [qt.rand_herm(2**6, density=0.1) for _ in range(5)]

t0 = perf_counter()
channel1 = qt.kraus_to_super(kraus_ops)
t1 = perf_counter()
channel2 = qt.Qobj(sum(qt.tensor(k.conj(), k) for k in kraus_ops), dims=channel1.dims)
t2 = perf_counter()

assert abs((channel1 - channel2).norm("fro")) < 1e-10

print(f"qutip.kraus_to_super:\t", t1 - t0)
print(f"custom kraus_to_super:\t", t2 - t1)
```

In this simple benchmark I get something like:
```
qutip.kraus_to_super:	 1.7386491200013552
custom kraus_to_super:	 0.03996859999460867
```
which is a ~45x speedup.

Ericgig:
Our support of kraus map (and other super representations) is somewhat lacking and there is a lot to improve. 
I can only speculate the reason `choi` is used as an intermediate representation, but there are no good reason that force us to keep using it.

PR #2569, opened 2024-12-05
Add sparse option to kraus_to_super

Sola85:
**Description**
Add an alternate implementation for `kraus_to_super`, which preserves sparsity. 
Contrary to what I said in the linked issue, the sparse option is not always faster than the existing implementation, so instead of replacing the implementation, I added a `sparse` keyword argument to `kraus_to_super`, which defaults to the old behaviour.

**Related issues or PRs**
Fixes https://github.com/qutip/qutip/issues/2567


Ericgig (approved):
Thank you!
You could also have removed the old method.

PR #2571, opened 2024-12-10
Fixed the sqrtiswap logic gate error

YangZ926:
**Description**
Fixed the sqrtiswap logic gate error. 

$$
\sqrt{i \text { SWAP }} = 
\left[\begin{array}{cccc}
1 & 0 & 0 & 0 \\
0 & 1 / \sqrt{2} & -i / \sqrt{2} & 0 \\
0 & -i / \sqrt{2} & 1 / \sqrt{2} & 0 \\
0 & 0 & 0 & 1
\end{array}\right]
$$

### DIFF
--- a/doc/changes/2571.bugfix
+++ b/doc/changes/2571.bugfix
@@ -1,0 +1,1 @@
+Fixed the sqrtiswap logic gate error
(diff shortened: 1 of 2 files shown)


Ericgig:
The definition I have seen does not have the negative sign...
It also does not square to iswap.

PR #2582, opened 2024-12-13
Check for integer overflow when computing Dense data size

### LINKED ISSUE #2575: Segfaults in various functions when dimension is too large
### Bug Description

Some functions segfault when operating on data with too large dimensions. The functions in the below example result in a `MemoryError: Could not allocate memory to create a (65535, 65535) Dense matrix`  (as they should) when `n < 2**16`, but when `n >= 2**16` they result in a segfault.

### Code to Reproduce the Bug

```shell
import qutip
import numpy as np

#n = 2**16 - 1 # MemoryError -> ok
n = 2**16 # segfault


state = qutip.Qobj(np.random.random((n, 1)), dims=[[n], [1]])

#each of the following lines individually cause a segfault when n >= 2**16
state = qutip.rand_ket(n, density=0.001)
state*state.dag()
qutip.ket2dm(state)
```

### Code Output

```shell
MemoryError: Could not allocate memory to create a (65535, 65535) ... 

or 

zsh: segmentation fault (core dumped)  python qutip_debugging.py
```

### Expected Behaviour

`MemoryError`'s should be raised in every situation.

### Your Environment

```shell
QuTiP Version:      5.0.4                                                                                                                                                                                
Numpy Version:      1.26.4                                                                                                                                                                               
Scipy Version:      1.12.0                                                                                                                                                                               
Cython Version:     0.29.28                                                                                                                                                                              
Matplotlib Version: 3.5.1                                                                                                                                                                                
Python Version:     3.9.12                                                                                                                                                                               
Number of CPUs:     24                                                                                                                                                                                   
BLAS Info:          Generic                                                                                                                                                                              
INTEL MKL Ext:      False                                                                                                                                                                                
```

### Additional Context

_No response_

Ericgig:
**Description**
We use int32 per default for the shape of the data.
When allocating memory ourselves for dense layer, the number of elements was computed using these int32 an could overflow:
allocating a `2**16 x 2**16` matrix would allocate a matrix of `0` elements and segfault instead of raising a MemoryError.

**Related issues or PRs**
fix #2575

### DIFF
--- a/qutip/core/data/dense.pyx
+++ b/qutip/core/data/dense.pyx
@@ -127,7 +127,7 @@ cdef class Dense(base.Data):
         """
         cdef Dense out = Dense.__new__(Dense)
         cdef size_t size = (
-            _mul_checked(self.shape[0], self.shape[1]) * sizeof(double complex)
+            _mul_checked(_mul_checked(self.shape[0], self.shape[1]), sizeof(double complex))
         )
         cdef double complex *ptr = <double complex *> PyDataMem_NEW(size)
         if not ptr:
@@ -176,7 +176,7 @@ cdef class Dense(base.Data):
         will not affect the original data structure.
         """
         cdef size_t size = (
-          _mul_checked(self.shape[0], self.shape[1]) * sizeof(double complex)
+          _mul_checked(_mul_checked(self.shape[0], self.shape[1]), sizeof(double complex))
         )
         cdef double complex *ptr = <double complex *> PyDataMem_NEW(size)
         if not ptr:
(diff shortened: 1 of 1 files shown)


qutip/core/data/dense.pyx
```diff
@@ -121,7 +126,9 @@ cdef class Dense(base.Data):
         low-level C code).
         """
         cdef Dense out = Dense.__new__(Dense)
-        cdef size_t size = self.shape[0]*self.shape[1]*sizeof(double complex)
+        cdef size_t size = (
+            _mul_checked(self.shape[0], self.shape[1]) * sizeof(double complex)
```
hodgestar, line 130: Can the multiplication by `sizeof(double complex)` also overflow?

Ericgig: `size_t` is can be `uint32` on older system so I added the check there also.

hodgestar (approved):
Is it worthwhile having a `_mul_checked` with two arguments and one with three arguments to minimize the number of function calls that do overflow checks?

PR #2583, opened 2024-12-17
Build wheels for mac with intel chips

Ericgig:
**Description**
From a discussion on discord between Jake and Simon.
We are no longer building wheels for mac with intel chips.
`macos-13`: intel chips
`macos-latest`: apple silicon

Changed the tests so both mac versions are covered.


hodgestar (approved):
Thanks, @Ericgig!

PR #2585, opened 2024-12-18
Update docstring of simdiag

Sola85:
The ordering of the return type is (eigenvalues, eigenvectors), not the other way round.


Ericgig (approved):
Thank you

PR #2589, opened 2024-12-19
Run some tests on python 3.13

Ericgig:
**Description**
I though we already ran some test with python 3.13...
Add some for all OS.


Ericgig:
For python 3.13, tests passes, but the upload to coveralls fails.
coveralls does not support it yet: https://github.com/TheKevJames/coveralls-python/pull/542

hodgestar (approved):

hodgestar:
I would not be against ditching coveralls and saving our own coverage artifacts that we can download and look at. Thoughts?

Issue #2580, opened 2024-12-13
Environments should work with more solvers


pmenczel:
### Problem Description

We recently added `qutip.BosonicEnvironment` and `qutip.FermionicEnvironment`, which are general descriptions of open quantum system environments. This issue is to track how well this is integrated with QuTiP's solvers.

* Currently, environments can be used with the HEOM solvers.
* Currently, bosonic environments can be used with the Bloch-Redfield solver (which just uses the power spectrum of the environment for the `a_ops`). However, that is completely undocumented.

### Proposed Solution

* Document that bosonic environments can be used as `a_ops` with the BR solver.
* Make fermionic environments usable with the BR solver. That would involve first going through the theory to see what the BR equation looks for the coupling term that we assume in the fermionic environment, and perhaps adapting the BR solver to it. (The [docs](https://qutip.readthedocs.io/en/master/guide/dynamics/dynamics-bloch-redfield.html#brief-derivation-and-definitions) say that the BR solver can only be used with Hermitian coupling operators, but @nwlambert says he has used it with non-Hermitian coupling operators and it worked.)
* Make (bosonic and fermionic) environments usable with `mesolve`. As a user, I would expect being able to pass an environment to `mesolve` and it should do the same thing as `brmesolve` with full secular approximation. In practice, it might be best to just delegate to the BR solver, which is optimized for diagonalizing the Hamiltonian in every time step (which would be necessary here as well).
* Think about whether environments should be integrated with any other solvers.

gsuarezr:
Hey, sorry for taking so long to reply here! hope you enjoyed the holidays!

I have been playing around with the new environments, but mainly adding more fitting methods and struggling to generalize the fermionic bit, I hope I can tackle the BR after I get a better hold of it, but I am being really slow on it,. For the fermionic environments, I guess we want to avoid numerical integration and would like something like the fft we are using here. I was playing around with something called the [fast Laplace transform](https://www.sciencedirect.com/science/article/pii/S2772415823000093) to obtain the  correlation function (using [this code ](https://github.com/lohyenlee/flt))


Basically, I would like to reproduce [this paper](https://arxiv.org/abs/2204.06875) with the qutip environments, that's why I've been playing around with fermions but not sure how good it is or whether it's overkill and anyway I know it might be a lot of work again so I'll mention it again when I have something actually working nicely 

[...]

PR #2594, opened 2025-01-07
Adding extra fitting methods for quantum environments

gsuarezr:
**Description**
This PR aims to add other methods used to generate decaying exponential decompositions of environments. The methods added in this PR have been applied to fermionic environments in the literature, while the current draft PR only implements these methods for  bosonic environments, hopefully they can be used in fermionic environments soon. The key issue there is how to create the interface for the creation of arbitrary fermionic environments. 

The methods this PR aims to add are : 

- [x] AAA 
- [x] Prony
- [x] Matrix Pencil
- [x] Esprit
- [x] Espira (I and II)
- [x]  Power Spectrum NLSQ

**Checklist**

- [x] Refactor fitting methods
- [x] Add unit tests for fitting methods
- [x] Include the changelog 
- [x] Document characteristics and when to use each method

NOTE: I decided to not do the balanced truncation method in this PR, it works to reduce exponents in all of them, but I probably need some time to understand it

**Related issues or PRs** 
#2580 

I would like somebody else's opinion on whether or not to keep mp, and espira-I, while they have their uses in signal analysis, maybe not so much here.  In  qutip/qutip-tutorials#72  1d all of them are used. It needs work as a tutorial, but maybe someone can run it and emit their opinion


qutip/core/environment.py
```diff
@@ -767,6 +768,133 @@ def approx_by_sd_fit(
             ckAR, vkAR, ckAI, vkAI, combine=combine, T=self.T, tag=tag)
         return approx_env, fit_info
 
+    def approx_by_aaa(
```
hodgestar, line 771: It looks like we might have an every growing number of such methods so perhaps we should have one method named `approximate` that accepts a method name and dispatches the rest of the arguments to the implementing functions. For example, in a simple implementation `.approximate("aaa", ...)` might end up calling `_approx_by_aaa(...)`.

gsuarezr: Hi Simon, 

I wasn´t expecting any comment so soon, thanks a lot!

Indeed, there seems to be a growing number of methods appearing in the literature. I added the ones in [this review article](https://pubs.aip.org/aip/jcp/article-abstract/160/20/204105/3294390/High-accuracy-exponential-decomposition-of-bath?redirectedFrom=fulltext) (sorry to attach a paper behind a paywall sadly no arxiv 😢 ). If @pmenczel  agrees then I think that's probably the best solution, I guess the main issue would be that some methods don´t take in the same parameters so it might be confusing for a user but I think that's fine 😄 

To be honest for my purposes only Prony and AAA are of importance, but since the other ones are similar I included them as well, but not 100% sure all should be included in the end.

pmenczel: In principle I would like an implementation like `.approximate("aaa", ...)`. That is actually how I implemented it at first. However, I found it very annoying to use since all the methods require different parameters, and the IDE then didn't show a tooltip listing the required parameters. That is why I went with having different functions.

Perhaps that could be fixed using `@overload`, though?

gsuarezr: I played around with overload in the last week, and I could get the hinting working for simple cases, but when doing it similarly in the code we already have for some reason it no longer works, at least not in vscode, I'm still looking into it though! 
Sorry for not replying before 😢

pmenczel:
Hi, thank you for all your work. I just pushed a commit without major changes to code or text, but moving a lot of documentation around. Take a look to see if you like it. The problem was that `@overload` versions of functions are not supposed to have docstrings, and these docstrings are ignored in the automatically generated API documentation. So I created a central place in the docs to document all of the approximation methods.

In addition, I thought it would make sense to have the matsubara and pade approximations also available through `approximate`. Unfortunately, if a subclass adds new @overload versions of a functions, all previous ones must also be copied into the subclass. Codeclimate is a bit unhappy about that, but it seems unavoidable with the single-approximate-function approach.

A few other points:
* I left two TODOs in the users guide
* I am unsure about `mp` and `espira-I`. If even you don't know a use case, we could consider dropping it to reduce the amount of code and not overwhelm the user with choices. On the other hand, you have already written the code, so it would seem a bit stupid to delete it now. I would say, it is up to you.
* Regarding the AAA, we could add a little comment in the code that our own implementation can perhaps be removed once we support only scipy 1.15+.
* I am unsure about that $r^2$ coefficient instead of least squares. Are you saying that it improves the results significantly?
    - Perhaps we could have a flag to decide whether to do $r^2$ or least squares?
    - There are many places where we refer to least squares, like docstrings, name of `rmse` variable, referring to fitting functions as "NLSQ", ...
    - Double check that the description of the `fit_info` output is correct everywhere in the new section in the Users Guide. By the way, it seems to be missing in the `approximate("aaa", ...)` docstring.
* Let's also double check if stuff has to be updated in the HEOM guide.
* I haven't touched any fermions stuff, I'm leaving that for the follow-up fermion fitting PR that you mentioned.

Note to myself: I still have to have a closer look at the tutorial notebooks

gsuarezr:
Thanks ! and thak you for such a quick review! Indeed it's a shame that codeclimate doesn't like it. probably going to be repeated when fermionic things are added, but I guess we can ignore codeclimate in this particular case?

> I left two TODOs in the users guide

Cool I'll take a look at them 

> I am unsure about mp and espira-I. If even you don't know a use case, we could consider dropping it to reduce the amount of code and not overwhelm the user with choices. On the other hand, you have already written the code, so it would seem a bit stupid to delete it now. I would say, it is up to you.

Yeah, I think more than the amount of code is too many choices for the user, right now I think we can just delete mp, and keep espira-I because it seems to be working nicely with fermions, though I need to take a look at some more examples, of course as I mentioned on line, whenever pade is available pade is better, a lot of the time by a lot

> Regarding the AAA, we could add a little comment in the code that our own implementation can perhaps be removed once we support only scipy 1.15+.

Yes, we definitely should. The main problem is that the current implementation of scipy does not return what I called iindices, so we cannot construct the objects needed for espira from it. But I guess we can ask for it! perhaps They'll add esprit and those prony based methods in the future, I asked about it on the scientific python discord, I'm waiting for a reply


> I am unsure about that $1-r^{2}$ coefficient instead of least squares. Are you saying that it improves the results significantly?

it's supposed to replace rmse,  both are measures of the goodness of fit. That one just can be kept as it is, $r^{2}$ just seems to be a more common measure so I wanted to replace rmse, but probably did not document that correctly, it doesn't change the results at all.

On that hand I was thinking maybe, the goodness of fit is not so informative sometimes, because you might have a good fit for the correlation function, but that does not reproduce the power spectrum so well (The bit about problems keeping kms with these approximations, I think that was briefly discussed in Neill's pseudomode paper). So perhaps goodness of fit to both should be reported? Not sure. Maybe it's just better to keep it as it was.

I will just delete the $r^{2}$ stuff


> Double check that the description of the fit_info output is correct everywhere in the new section in the Users Guide. By the way, it seems to be missing in the approximate("aaa", ...) docstring. Let's also double check if stuff has to be updated in the HEOM guide.

Sure I'll check the guides and dosctrings, I didn't check them too much in case many changes were required. I think the notation in the summaries is also not uniform right now, I'll just keep an eye on it to make it uniform and check stuff in the user's guide

> I haven't touched any fermions stuff, I'm leaving that for the follow-up fermion fitting PR that you mentioned.

Sure, I'll just create that one once this one is merged.


> Note to myself: I still have to have a closer look at the tutorial notebooks

If notation, is going to be kept (up to minor details, like I just noticed sometimes is Nmax or N_max in different methods), then I will take another look at the tutorials and send you a message on line once they're ready

pmenczel:
> Yeah, I think more than the amount of code is too many choices for the user, right now I think we can just delete mp, and keep espira-I because it seems to be working nicely with fermions, though I need to take a look at some more examples, of course as I mentioned on line, whenever pade is available pade is better, a lot of the time by a lot

Sounds reasonable :)

> Yes, we definitely should. The main problem is that the current implementation of scipy does not return what I called iindices, so we cannot construct the objects needed for espira from it. But I guess we can ask for it! perhaps They'll add esprit and those prony based methods in the future, I asked about it on the scientific python discord, I'm waiting for a reply

Oh in that case, let's definitely keep your implementation for now, and if the scipy implementation can't do everything we need, there's no need to add a comment. But if they reply and eventually change things, that would be nice!

> it's supposed to replace rmse, both are measures of the goodness of fit. That one just can be kept as it is, r2 just seems to be a more common measure so I wanted to replace rmse, but probably did not document that correctly, it doesn't change the results at all.
> 
> On that hand I was thinking maybe, the goodness of fit is not so informative sometimes, because you might have a good fit for the correlation function, but that does not reproduce the power spectrum so well (The bit about problems keeping kms with these approximations, I think that was briefly discussed in Neill's pseudomode paper). So perhaps goodness of fit to both should be reported? Not sure. Maybe it's just better to keep it as it was.
> 
> I will just delete the r2 stuff

Ah. Well if r2 is a more common thing to do in fitting, it's okay to keep it. I just didn't know, I thought least squares was the "normal" thing. Just make sure that everything is named and documented right.

I think if we fit the power spectrum, we only need to include information about the goodness of how well it fits the power spectrum; if we fit the CF we only need to include the goodness of how it fits the CF, etc.

gsuarezr:
In the end I removed $r^{2}$, it is more common but having one of the two is enough since [they both agree on what the better model is](https://datascience.stackexchange.com/questions/100605/rmse-vs-r-squared ). I also removed mp, and did the todos in the user guide. I left a comment there in case what I wrote wasn't clear

> I think if we fit the power spectrum, we only need to include information about the goodness of how well it fits the power spectrum; if we fit the CF we only need to include the goodness of how it fits the CF, etc.

Sure, that's sort of the logical thing, I was just thinking about it because I've found that these goodness of fit measures are not too useful to know which fit is better in many cases because of the KMS stuff

qutip/utilities.py
```diff
@@ -523,3 +529,414 @@ def _fit(fun, num_params, xdata, ydata, guesses, lower, upper, sigma,
...
+
+# Prony methods Fitting
+
+def _prony_model(orig, amp, phase):
+    # It serves to compute rmse, a single term of the prony
+    # polynomial form https://doi.org/10.1093/imanum/drab108 using phases
+    return amp * np.power(phase, np.arange(len(orig)))
+
+
+def prony_methods(method: str, signal: ArrayLike, n: int):
```
pmenczel, line 796:
```
method: Literal["prony", "esprit"]
```

Also, the formatting of the docstring below is off. Same for espira.

qutip/utilities.py
```diff
@@ -523,3 +529,414 @@ def _fit(fun, num_params, xdata, ydata, guesses, lower, upper, sigma,
...
+    """
+    Estimate amplitudes and frequencies using prony methods.
+    Based on the description in https://doi.org/10.1093/imanum/drab108
+    and their matlab implementation
+    Args:
+        method (str): The method to obtain the roots of the prony polynomial
+        can be prony, and Estimation of signal parameters
+        via rotational invariant techniques (esprit)
+        signal (np.ndarray): The input signal (1D complex array).
+        n (int): Desired number of modes to  use as estimation
```
pmenczel, line 806: I'm not so familiar with this fitting stuff. For prony and espira, the sampling points are always assumed to be evenly spaced?

qutip/tests/test_utilities.py
```diff
@@ -127,19 +127,43 @@ class TestFitting:
     def model(self, x, a, b, c):
         return np.real(a * np.exp(-(b + 1j * c) * x))
 
-    def test_same_model(self):
+    @pytest.fixture
+    def generate_data(self):
```
pmenczel, line 131: Now that the fitting is moving a bit more in the spotlight, it might be good to test a little better. How about something like

```suggestion
    @pytest.fixture
    @pytest.mark.parametrize("noisy", [False, True])
    def generate_data(self, noisy):
```

and add some small random noise to `y`?

qutip/tests/test_utilities.py
```diff
@@ -127,19 +127,43 @@ class TestFitting:
...
         assert (np.all(np.isclose(params, [fparams1, fparams2], atol=1e-3)) or
-                np.all(np.isclose(params, [fparams2, fparams1], atol=1e-3)))
\ No newline at end of file
+                np.all(np.isclose(params, [fparams2, fparams1], atol=1e-3)))
+
+    def test_aaa(self, generate_data):
+        x, y, _, _ = generate_data
+        result = utils.aaa(y, x, tol=1e-13, max_iter=4)
+        rmse = result["rmse"]
+        assert rmse < 1e-3
```
pmenczel, line 153: For all the new fitting methods: in addition to checking that the returned rmse is small enough, it would be good to also check that the result is actually close to the original data (`np.assert_allclose`)

qutip/tests/core/test_environment.py
```diff
@@ -624,6 +625,71 @@ def test_dynamic_sd_fit(self, reference, wMax, tol, params):
...
 
+    @pytest.mark.parametrize(["reference", "tMax", "tol"], [
+        pytest.param(OhmicReference(3, .75, 10, 1),
+                     15, 1e-3, id="Ohmic Example"),
+        pytest.param(UDReference(1, .5, .1, 1), 2, 1e-3, id='UD Example'),
+    ])
+    def test_fixed_prony_fit(self, reference, tMax, tol):
+        # TODO: come up with a better test, for now I was just fitting into this
+        # style but those support points are not good for a low number of
+        # exponents
```
pmenczel, line 636: TODO here. Also, tests for ESPIRA fits are missing, right?

doc/guide/environments/approximations.rst
```diff
@@ -0,0 +1,666 @@
...
+   * - Allows Constraints
+     - Yes
+     - No
+     - Partially
+   * - Stable
+     - No
+     - Partially
+     - Yes
+
+In the table requires extra information refers to whether we need to input
```
pmenczel, line 148:
```suggestion
In the table, "Requires Extra Information" refers to whether we need to input 
```
(etc)

pmenczel (commented):
Thank you for your changes! I think we are getting very close. I have some comments with suggestions on naming and some docstrings. Other than that, only:

* There are still some places in docstrings etc that refer to the "approx_by_" functions, we should update to "approximate"
* Same for the HEOM users guide maybe?
* Out of curiosity, did you ever hear back from the scipy discord about AAA?
* Again, note to myself to look at the notebook 👀

gsuarezr:
> Out of curiosity, did you ever hear back from the scipy discord about AAA?

No, they said they'd look into it 😢

qutip/utilities.py
gsuarezr: Yes!

qutip/utilities.py
```diff
@@ -523,3 +529,414 @@ def _fit(fun, num_params, xdata, ydata, guesses, lower, upper, sigma,
...
+        U1, _, _ = svd(hankel0)
+        pencil_matrix = np.linalg.pinv(U1.T @ hankel0) @ (U1.T @ hankel1)
+        phases = eigvals(pencil_matrix)
+    generation_matrix = np.array(
+        [[phase**k for phase in phases] for k in range(len(signal))])
+    amplitudes = lstsq(generation_matrix, signal)[0]
+    params = _unpack(
+        np.array([val for pair in zip(amplitudes, phases) for val in pair]), 2)
+
+    rmse = _rmse(_prony_model, signal, signal, params)
```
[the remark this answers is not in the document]
gsuarezr: I wasn't sure which bit looked fishy

The first two lines it's just because I have  two lists $[c_{1},c_{2},....]$ and $[p_{1},p_{2},....]$ so the [val for pair in zip(amplitudes, phases) for val in pair] is to have $[c_{1},p_{1},c_{2},p_{2}....]$ and the unpack to have $[[c_{1},p_{1}],[c_{2},p_{2}]....]$. Mostly so params has the same structure as the least squares stuff. 

if it's the fact that signal appears twice, it's just because the length of the original signal is needed to compute the approximated one, because the date is evenly spaced. I could have passed len(signal) as xdata and modify the function instead from 
```python
 def _prony_model(orig, amp, phase):
      return amp * np.power(phase, np.arange(len(orig)))
```
to
```python
 def _prony_model(n, amp, phase):
      return amp * np.power(phase, np.arange(n))
```

I guess that way is more clear

qutip/utilities.py
```diff
@@ -361,7 +363,7 @@ def iterated_fit(
     upper: ArrayLike = None,
     sigma: float | ArrayLike = None,
     maxfev: int = None
-) -> tuple[float, ArrayLike]:
+) -> tuple[float, float, ArrayLike]:
```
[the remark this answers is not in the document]
gsuarezr: True, forgot to delete it when I removed $r^{2}$

qutip/tests/test_utilities.py
gsuarezr: Sounds Good!

qutip/tests/test_utilities.py
gsuarezr: Sure, will check for that as well

qutip/tests/core/test_environment.py
gsuarezr: yes, the test need to be better I just added more exponents because I didn't want to test those examples at the time. I was planning to do it later on in case names or arguments changed. I will delete the TODO and add the ones for ESPIRA

qutip/utilities.py
```diff
@@ -523,3 +529,414 @@ def _fit(fun, num_params, xdata, ydata, guesses, lower, upper, sigma,
...
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+
+def aaa(func: Callable[..., complex], z: ArrayLike,
+        tol: float = 1e-13, max_iter: int = 100):
```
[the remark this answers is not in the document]
gsuarezr: Could we avoid adding the _fit?  I sort of don't like it because except for AAA all I did in utilities is phase estimation I get the exponent later in the environment file.  I also think when you add fit people immediately think minimization and that's not it here

qutip/utilities.py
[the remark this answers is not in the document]
pmenczel: Ah sorry, yes I was referring to the double "signal". I only meant to quote that line, but I think github always includes at least 4 lines. Thank you for the change!

gsuarezr:
I changed the Ohmic environment reference because the value we were using at zero temperature was only ok for Ohmic, but not super or sub Ohmic. The problem was we were using 

$C(t)=\alpha \omega_{c}^{s+1} \Gamma(s+1) (1+ i \omega_{c} t)^{-(s+1)}$

But the right one is 

$C(t)=\alpha \omega_{c}^{2} \Gamma(s+1) (1+ i \omega_{c} t)^{-(s+1)}$

so it was only good for $s=1$, Ohmic, whenever $s \neq 1 $. The correlation function for $T=0$ and $T=1\times 10^{-30}$ would differ significantly, now that's fixed :)

The other unrelated thing I did was add a method called rescale. The method multiplies the coefficients of the environment times a constant. I think this is useful when running multiple simulations that only differ by the coupling constant, so that fits do not need to be performed several times, but if it feels weird, the implementation is not good, or it seems confusing feel free to delete it

PR #2599, opened 2025-01-09
Docstrings and type hints  for distributions

MathiB123:
**Description**
Added docstrings and type hints for functions in `distributions.py`.

### DIFF
--- a/qutip/distributions.py
+++ b/qutip/distributions.py
@@ -246,7 +246,7 @@ class WignerDistribution(Distribution):
     extent : list, default : [[-5, 5], [-5, 5]]
         List of arrays with the bounds [a, b] for each coordinate.
     steps : int, default : 250
-        The number of data points generated between the bounds for 
+        The number of data points generated between the bounds for
         each coordinate.
 
     """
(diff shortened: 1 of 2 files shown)


qutip/distributions.py
```diff
@@ -460,10 +558,9 @@ def __init__(self, rho=None, omega=1.0, extent=[-5, 5], steps=250):
         if rho:
             self.update(rho)
 
-    def update(self, rho):
-        """
-        Calculate the probability function for the given state of an harmonic
-        oscillator (as density matrix)
+    def update(self, rho: Qobj):
+        """Calculates the probability function for the given state of an
+        harmonic oscillator (as density matrix).
```
Ericgig, line 563: You added `Parameter` to the other `update` method. Let's add it here too.

Ericgig (commented):
Why did you add type hints to only the `update` function?
The `__init__` and the `visualize` are also user facing and would benefit from hints to.

qutip/distributions.py
MathiB123: All modifications were made

Ericgig (approved):
Thank you.

PR #2607, opened 2025-01-13
Removed certain distributions already in visualization

MathiB123:
Removed WignerDistribution, QDistribution from distributions.py and their tests. 

Added Harmonic Oscillator to API doc.

### DIFF
--- a/qutip/animation.py
+++ b/qutip/animation.py
@@ -3,11 +3,11 @@ Functions to animate results of quantum dynamics simulations,
 """
 __all__ = ['anim_wigner_sphere', 'anim_hinton', 'anim_sphereplot',
            'anim_matrix_histogram', 'anim_fock_distribution', 'anim_wigner',
-           'anim_spin_distribution', 'anim_qubism', 'anim_schmidt']
+           'anim_spin_distribution', 'anim_qubism', 'anim_schmidt', 'anim_qfunc']
 
 from . import (plot_wigner_sphere, hinton, sphereplot, matrix_histogram,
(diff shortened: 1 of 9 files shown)


Ericgig (commented):
Looking good for the removals.

`plot_qfunc` looks good, but it need to be integrated in the rest of the code more: apidoc, `__all__`, tests, animations, towncrier... You can search for `plot_wigner` in the full project, for a list places.

Ericgig (commented):
Also changes in #2599 broke qutip-jax tests.
qutip should work when matplotlib it not installed, but types hints in distribution don't work without it. 
Could you fix it in this PR.

qutip/animation.py
```diff
@@ -3,11 +3,11 @@
 """
 __all__ = ['anim_wigner_sphere', 'anim_hinton', 'anim_sphereplot',
            'anim_matrix_histogram', 'anim_fock_distribution', 'anim_wigner',
-           'anim_spin_distribution', 'anim_qubism', 'anim_schmidt']
+           'anim_spin_distribution', 'anim_qubism', 'anim_schmidt', 'anim_qfunc']
```
Ericgig, line 6:
```suggestion
           'anim_spin_distribution', 'anim_qubism', 'anim_schmidt',
           'anim_qfunc']
```

Ericgig (commented):
Could you add `anim_qfunc` in the apidoc.
Also move the distribution section as an automodule under "Graphs and Visualization".

Ericgig (approved):

PR #2610, opened 2025-01-15
Create a `A @ B.dag()` operation

Ericgig:
**Description**
Create function to compute `A @ B.dag() ` in one operation.
This will help with super operators not using super representation.

For example kraus operators: `A @ rho @ A.dag()`.

For CSR operators, `CSR @ Dense` is fast, but `Dense @ CSR` is slow. (We have to convert one of the matrices.)

However `Dense @ CSR.dag == ( CSR.conj @ Dense.T ).T`. Transposing dense array is free since we support both c and fortran ordered matrices, so we can do that operation almost as fast as the usual `CSR @ Dense` operation.

For `Dense @ Dense` operation, lapack matmul function support `A @ B.dag`. So the new operation allows to do the matmul without creating the temporary dag matrix.


Ericgig:
With this PR, it will need to be used manually.
It will be useful for some specific physic based integrator method.
We also played with a split pre-post QobjEvo for dense representation. I would like to make this official with a KrausMap class.
(I made a draft in my fork: https://github.com/Ericgig/qutip/pull/14)

qutip/core/data/matmul.pyx
```diff
@@ -42,6 +46,10 @@ cdef extern from "src/matmul_csr_vector.hpp" nogil:
         double complex *data, T *col_index, T *row_index,
         double complex *vec, double complex scale, double complex *out,
         T nrows)
+    void _matmul_dag_csr_vector[T](
```
pmenczel, line 49: As someone unfamiliar with this stuff, this is a bit mysterious. There is no documentation for this function, and why does it have to be written in C++ instead of cython? (I do see that it is the same for regular `matmul`)

qutip/core/data/matmul.pyx
```diff
@@ -519,6 +528,173 @@ cpdef Dense matmul_dense_dia_dense(Dense left, Dia right, double complex scale=1
     return out
 
 
+cpdef Dense matmul_dag_dense_csr_dense(
+    Dense left, CSR right,
+    double complex scale=1, Dense out=None
+):
+    """
+    Perform the operation
+        ``out := scale * (left @ right) + out``
```
pmenczel, line 537: Is there a dagger missing?

qutip/core/data/src/matmul_csr_vector.cpp
```diff
@@ -52,8 +52,53 @@ void _matmul_csr_vector(
         _mm_storeu_pd((double *)&out[row], num3);
     }
 }
+/*
```
pmenczel, line 55: What is the meaning of this?

pmenczel (commented):
It is difficult for me to follow the details, but the tests look good and they pass.

qutip/core/data/matmul.pyx
Ericgig:
```suggestion
        ``out = scale * (left @ right.dag()) + out``
```

qutip/core/data/matmul.pyx
Ericgig: I copied (half successfully) the c++ version we have for matmul.
From my test years ago, it's faster that what we can do in cython and what scipy does (did?), by about 30%.

Ericgig:
@pmenczel 
Do you want me to compare a cython, simple c++, hand made SIMD (what we have now) and SIMD openmp implementations,
for both this and the normal matmul operation?
I haven't benchmarked them recently, but my guess is that the hand made version we have is no longer the best. 
It's written using low level instructions for 2011 cpus and the compiler should be able to do a better job if we allow it to target recent hardware.

pmenczel (approved):
Okay, I understand. Interesting, all the kinds of stuff that are somewhere deep in qutip. Before merging, perhaps you could add a few lines of comment about the commented out code? Just for the (unlikely?) case that someone other than you will try to touch it some day.

Regarding benchmarks, if this could be an easy way to slightly speed up matmul, it sounds like a good idea to me. But perhaps not the #1 priority of things to do.

PR #2611, opened 2025-01-22
Change operator created with `|a><b|` default `dtype`.

Ericgig:
**Description**

It is common for users to create operators using `|a><b|`. 
But our default matrix types will end up creating a dense operator, and with the weight we have, dense is set have priority in many binary operations.

To prevent this, this PR change the operation use for `|a><b|` to create sparse matrices more easily.
`matmul_outer` does the same operation as `matmul`, but will default to sparse in mix cases.
It also look at the output density and swap between dense and sparse automatically. (cutoff 30% fill chosen so `|a><b|` create sparse matrices when input are fock state even for qubit dimensions.

With #2391, it should help keeping large operators sparse as expected.

Alternatives
- We could have every data layer matrix keep an internal `nnz` or density value and have matmul function chose the output accordingly.
- If/when we make the symbolic data layer, we could detect operators created from ket and act accordingly.
- Set the default dtype for ket as sparse. This would slow down `op @ ket`, but we can still keep the dense states as default in solvers.

(Also fix issues with epsilon in the specialization finding algorithm.)


Ericgig:
@nwlambert Could you review.

nwlambert (approved):
Thanks eric, this seems to resolve the problems I had in #2328 

I approved, implementation seemed good to me.  One niave question, I guess  nnz is only called when doing the matmul outer, so shouldn't cause a performance hit anywhere important (presuming there is a cost to nnz, maybe its negligible?)

Ericgig:
Yes `nnz` will only be called in `matmul_outer` and should not affect any time other than the specific case of `ket @ bra`. In that case, since they are vectors, `nnz` time should be quite fast.

Issue #2618, opened 2025-01-30
Pretty-printing of basis expansion for states


Sola85:
### Problem Description

The current string representation of `Qobj`'s is often not very useful, in particular when dealing with states that are a superposition of just a few basis elements. E.g.

```python
ket = 0.3*tensor(basis(8, 5), basis(2, 1), basis(3, 0)) + 0.5*tensor(basis(8, 3), basis(2, 0), basis(3, 2))
print(ket)
```
This spams my console with 50 (mostly empty) lines. Having the option to print this as `0.3 |5, 1, 0> + 0.5 |3, 0, 2>` would be much nicer.

### Proposed Solution

If there is interest for this, I could open a PR to add something like `Qobj.print_basis_expansion()` or similar (I'd be open for suggestions for a better name).

This is the implementation I've been using so far.

```python
def print_basis_expansion(state: qutip.Qobj, decimal_places: int = 14, dim_separator: str = "auto", 
                          term_separator: str = " +\n", sort: str = "largest-first") -> str:
    """
        Return a string-representation of the basis expansion for a ket or a bra, e.g.

        "0.5 |010> + 0.25 |000> + ..."

        Parameters
        ----------
        decimal_places : int
            The number of decimal places to show for the coefficients.

        dim_separator: str
            Separator string between dimension indices. E.g. use ", " to print as "|0, 0, 0>". 
            If dim_separator="auto", then "" is used for qubits and ", " is used otherwise.

        term_separator: str
            Separator string between terms. E.g. use " + "
            to print in a single line as "|000> + |111>"

        sort: str
            Show largest elements (by magnitude) first if "largest-first".
            Unsorted otherwise.

        Returns
        -------
        basis_expansion : str
            The requested string representation.
    """

    if not (state.isket or state.isbra):
        raise ValueError(f"Only implemented for states, not {state.type}.")
    dims = state.dims[0 if state.isket else 1]

    if dim_separator == "auto":
        # No separator for pure qubit states, but comma-separator otherwise,
        # since bitstrings are nice, but e.g. |153> would be ambiguous. 
        dim_separator = ", " if any([dim > 2 for dim in dims]) else ""

    template = "{} |{}>" if state.isket else "{} <{}|"
    
    ket_full = np.round(state.full(squeeze=True), decimals=decimal_places)

    indices = range(len(ket_full))
    if sort == "largest-first":
        indices = np.argsort(np.abs(ket_full))[::-1]

    parts = []
    for i in indices:
[...]

PR #2620, opened 2025-01-31
Support pretty-printing of basis expansion for states

Sola85:
**Description**
Added `Qobj.print_basis_expansion()` which returns string representations for states of the form "(0.24+0.3j) |0010> + ...", i.e. essentially the inverse of `qutip.ket()`/`qutip.bra()`.
This works for any dimensions, i.e. `qutip.basis([5, 5, 5], [4, 0, 2]).print_basis_expansion()` would return  `"(1+0j) |4, 0, 2>"`.

I'd be happy for suggestions for a better name for this function, since it doesn't actually print anything to console, but I found this name less unwieldy than e.g. `.basis_expansion_string()` and clearer than just `.basis_expansion()`.

**Related issues or PRs**
Fixes https://github.com/qutip/qutip/issues/2618


Ericgig (approved):
Look good, thank you.
Very clean representation of states.
I would go with `basis_expansion`.

Sola85:
> I would go with `basis_expansion`.

Done!

Issue #2617, opened 2025-01-29
Add visualisations method to results


Ericgig:
### Problem Description

It is useful to use plots to look at results from evolution. There are functions for that (`plot_expectation_values`, `anim_...`,) but even tutorials often use matplotlib directly. It would be nice to have them directly in the `Result` class.

### Proposed Solution

Create `Result.plot_expect` that does just that.
Since `e_ops` can be a dictionary, it should take label from that.
And maybe use the solver name for the title.

It would be nice to have more options for `mcsolve` results: plot the average of all trajectories and/or the expectation values of the N first trajectories.
It could also plot the photocurrent.

vikas-chaudhary-2802:
I love to work on this issue.

lucifer4073:
Hi @Ericgig , Can I work on this issue ?

ANUKOOL324:
hi @Ericgig allow me to work on this issue ?

Ericgig:
@ANUKOOL324, the previous PR seems to have been abandoned...
You can have a go at it.

famous111:
Hi @Ericgig, would you like to have a look at the new PR?

AnujRewar:
Please review my PR

famous111:
Hi @AnujRewar, thank you for your contribution! I wanted to mention that I have already submitted a PR #2724 and Ericgig started to review it on August 12th. I plan to continue working on this issue and will keep an eye on the review process.

Ericgig:
@famous111, @AnujRewar  
Are you doing this for a class?
Now that the summer is ending, we have a few new contributors coming all at once, when we usually only have such interest during GSoC application period. Either way, thank you for you interest. 

Have you looked at `visualization.py`, we have unified the interface of all plotting function in qutip to have keyword only `fig` and `axes` and I would expect something similar here.

AnujRewar:
Hi @Ericgig, thank you for your message!

I’m new to contributing to open-source projects, and I’m not doing this for a class. I’m really excited to get involved with QuTiP and contribute more. I also plan to participate in GSoC in the future, so this is a great opportunity for me to learn and get familiar with the project.

I’ve looked at visualization.py and noticed the unified interface using fig and axes as keyword arguments. I’ll make sure to follow this structure when working on the plot_expect implementation to keep things consistent.

Thank you for the opportunity to contribute — I’m eager to learn and improve with the team

famous111:
Hi, @Ericgig
[...]

PR #2622, opened 2025-02-02
plot expect function added to result object

lucifer4073:
**Checklist**

- [X] Please read [Contributing to QuTiP Development](http://qutip.org/docs/latest/development/contributing.html)
- [X] Contributions to qutip should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/).

**Description**

The given PR adds the plot_expect function to the result class, plotting the required expectation values of `e_ops`. 

**Related issues or PRs**

The PR contributes to #2617.

**Additonal Comments**

- Additional options are such as options for mcsolve results or plotting photocurrent are yet to be added.
- This maybe added in another PR, or in the same PR, once the current changes are approved.
- Not sure on whether I need to create another function or add some extra parameters in the plot_expect to implement the additonal requirements.

### DIFF
--- a/qutip/solver/result.py
+++ b/qutip/solver/result.py
@@ -7,6 +7,7 @@ from typing import TypedDict, Any, Callable
 from ..core.numpy_backend import np
 from numpy.typing import ArrayLike
 from ..core import Qobj, QobjEvo, expect
+import matplotlib.pyplot as plt
 
 __all__ = ["Result"]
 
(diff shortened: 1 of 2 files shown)


qutip/solver/result.py
```diff
@@ -296,6 +297,42 @@ def _pre_copy(self, state):
...
+                "No expectation values found in the result object.")
+
+        if labels is None:
+            if hasattr(self, 'e_ops') and isinstance(self.e_ops, dict):
+                labels = list(self.e_ops.keys())
+            else:
+                labels = [f'Expectation {i}' for i in range(len(self.expect))]
+
+        plt.figure()
+        for i, (label, values) in enumerate(zip(labels, self.expect)):
```
Ericgig, line 325: `self.e_data` is a dict with the key of `e_ops` as keys (or 0, 1, ...) if e_ops was originally a list. And the expectations values as values. You could use it directly.

Ericgig (commented):
Thank you for your interest in the project.

I would recommend you to look into the `Result` object a little more.
`Result.e_ops` is not the user provided `e_ops` and expect is `None` when no `e_ops` are provided, not undefined...

requirements.txt
```diff
@@ -2,3 +2,4 @@ cython>=0.29.20
 numpy>=1.22
 scipy>=1.8
 packaging
+matplotlib
```
pmenczel, line 5: I think this should not be here, since matplotlib should remain an optional dependency

pmenczel (changes requested):
Thank you for your contribution!

In `bloch.py`, we have [logic](https://github.com/qutip/qutip/blob/ee2bc3de26aa2efb7c950f3fdfe91b6d2522ad82/qutip/bloch.py#L912) that replaces `show` with `Ipython.display` when inside ipython. I wonder if we should do the same here?

ANUKOOL324:
allow me to work on the issue @Ericgig ?

PR #2623, opened 2025-02-05
Set `auto_tidyup_dims` default to `True`

Ericgig:
**Description**

Scalar dimensions (ket's `[1]`) sometime keep the tensor structure, sometime not.
Change the default for the `auto_tidyup_dims` to True so they are never kept.

Tensored ket will have dims: `[[a, b, c, ...], [1]]`
Operator-ket/bra will have dims  `[ [[a, b], [c, d]], [1] ]`.


pmenczel:
Thank you Eric! I was wondering where the `[[1], [1]]` for operator-ket / bra comes from, it feels a bit arbitrary to me and like it might confuse people. Is it just a way of preventing `bra @ operator-ket` or does it make sense mathematically in a way that I am missing?

Ericgig:
Yes it's to prevent `operator-ket @ bra`. `[[1], [1]]` is the dimensions for a scalar column stacked operator. So `operator-ket` are map from scalar operator to operator.

pmenczel:
Okay, thinking about it, it might make sense. There seems to be a problem though, since I just checked out the PR and I am getting the following:

```python
In [3]: foo = qt.operator_to_vector(qt.sigmax())
In [4]: foo.dims
Out[4]: [[[2], [2]], [1]]
```

Also I was wondering: I think you said there are still places where the tidy-up is not applied, would that be difficult to fix?

Ericgig:
Collapse from `[1, 1, 1,...]` to `[1]` and collapse from `[[1, 1, ...],[1, 1, ...]]` to `[[1], [1]]` is now automatic everywhere the Dimensions class is used.
I could add conversion from `[1]` to `[[1], [1]]` if the other dims is super there also. Or remove the split with super scalar and normal scalar dims and add something in `Qobj.__matmul__`...

pmenczel:
I just got thinking, in principle, it would be consistent to also tidy up something like `[2, 3, 1, 1]` to `[2, 3]`, right? For example, to make an outer product, we usually write `ket1 @ ket2.dag()` but one could expect that `ket1 & ket2.dag()` would give the same result. However, the dims in the first case are `[[n], [m]]` and in the second case `[[n, 1], [1, m]]`.

(Other than this, I am ready to approve the PR.)

Ericgig:
Is that safe?
`bra & oper & ket` would have dims `[[m, o], [n, m]]`, the `m` position are no longer matching.

pmenczel (approved):

PR #2624, opened 2025-02-05
Apply tidyup options in `csr.from_dense`

Ericgig:
**Description**
Conversion from csr to dense keeps all non strictly `0` entries.
This could cause strange result: `oper.expm().to("csr")` would not round, but `oper.expm(dtype="csr")` would.

With this change, the `auto_tidyup` options is used to round off small values.
The default for `auto_tidyup_atol` is `1e-14` to 0.


pmenczel (approved):
Looks good to me

PR #2625, opened 2025-02-05
Use `numpy._core` in tests.

Ericgig:
**Description**
We monkey patch a numpy function in our tests. 

It import another function from numpy's internals, which is being moved from `numpy.core` to `numpy._core`.
This cause the patched `np.testing.assert_...` to raise a warning and we end up with only the warning message loosing the original error...


hodgestar (commented):
I would vote to just remove the monkeypatching entirely. There are other ways to debug and this is annoying to maintain;

PR #2626, opened 2025-02-06
Just remove the monkeypatch

Ericgig:
**Description**
Alternative to #2625, just removed the monkey-patched numpy function instead of fixing it.
The original function limit the diff of array to 3 lines, we patched it to print the whole array.
But we have to use numpy internal to do it and they can change it anytime without warnings.

With 3 lines of diff, there is not enough to see what's wrong, but with 100, there is too much logs and it's not that useful...


hodgestar (approved):
Happy for this to be merged. If we want something nicer perhaps a pytest assertrepr hook (https://docs.pytest.org/en/latest/how-to/assert.html#defining-your-own-explanation-for-failed-assertions) could help?

PR #2628, opened 2025-02-07
Add support for FermionicEnvironment in Bloch RedFieldTensor.

Ericgig:
**Description**
- Add support for `FermionicEnvironment` in Bloch `brmesolve`, `BRSolver` and `bloch_redfield_tensor`.
- Any environment are valid spectra for the above function.
- Add a new function: `brcrossterm`, compute cross term in equation (6) of our [guide](https://qutip.readthedocs.io/en/latest/guide/dynamics/dynamics-bloch-redfield.html).
- Improve documentation for the `br_dtype` parameter in `bloch_redfield_tensor`.
- Fix the equations in the documentation.

An example of fermion environment in the guide could be good.

### DIFF
--- a/qutip/core/_brtensor.pyx
+++ b/qutip/core/_brtensor.pyx
@@ -10,6 +10,7 @@ from qutip.core._brtools cimport _EigenBasisTransform
 from qutip.core.qobj import Qobj
 
 import numpy as np
+import itertools
 from libcpp.vector cimport vector
 from libc.math cimport fabs, fmin
 
@@ -221,7 +222,7 @@ cdef class _BlochRedfieldElement(_BaseElement):
         self.sec_cutoff = sec_cutoff
         self.eig_basis = eig_basis
         self.nrows = a_op.shape[0]
-        self.dims = [a_op.dims, a_op.dims]
+        self.dims = [a_op._dims, a_op._dims]
 
         dtype = dtype or ('dense' if sec_cutoff >= np.inf else 'sparse')
         self.tensortype = {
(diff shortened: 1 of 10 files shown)


pmenczel (commented):
Only a few comments:
* The docstrings in `qutip.core.blochredfield` don't mention where environment objects are accepted.
* Actually, the functions in `qutip.core.blochredfield` are not in the API documentation. Is that intentional?
* You write "An example of fermion environment in the guide could be good." Additionally, it would be good to update the bosonic examples there to show how BosonicEnvironment objects can be used.

I am not sure whether these points are within the scope of this PR. Other than this, looks great to me, but waiting for @nwlambert tests before approving

qutip/core/_brtensor.pyx
```diff
@@ -329,8 +332,293 @@ cdef class _BlochRedfieldElement(_BaseElement):
...
+        A^b -> b_op
+        spectra -> S_ab
+
+    The tensor is computed in the ``eig_basis``, but is returned in the outside
+    basis unless ``eig_basis`` is True. The ``matmul_data_t`` method also act
+    on a state expected to be in that basis. It transform the state instead of
+    the tensor as it is usually faster, the state being a density matrix in
+    most cases.
+
+    Diffenrent term can share a same instance of the eigen transform tool
```
nwlambert, line 526: Small typo (different)

nwlambert (approved):
works great on all examples I tried, thanks eric!

qutip/core/_brtensor.pyx
Ericgig:
```suggestion
    Different term can share a same instance of the eigen transform tool
```

qutip/core/blochredfield.py
```diff
@@ -4,48 +4,55 @@
...
+from .environment import BosonicEnvironment, FermionicEnvironment
 
-__all__ = ['bloch_redfield_tensor', 'brterm']
+
+__all__ = ['bloch_redfield_tensor', 'brterm', 'brcrossterm']
+
+SpectrumType = Union[
+    Coefficient, str, CoeffProtocol,
+    BosonicEnvironment, FermionicEnvironment
+]
```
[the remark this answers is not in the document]
Ericgig: I added the function to the apidoc and added environment in the spectra docstring.

Not sure about having this as an alias.
It's its only use. ( the spectra in floquet is not the same.)

pmenczel (approved):

Issue #1998, opened 2022-09-22
Sparse Eigensolver issue.


Ericgig:
### Bug Description

The sparse eigensolver can fail when eigenvalues are purely imaginary.

### Code to Reproduce the Bug

```shell
import qutip 
(qutip.num(4) * 1j).eigenenergies(sparse=True)
```

### Code Output

```shell
array([-3.33066907e-16+3.j, -1.06901857e-16+1.j,  1.03300843e-16+1.j,
        4.44089210e-16+3.j])
```

### Expected Behaviour

All unique eigenstates should be returned.

### Your Environment

```shell
Present in both `master` and `dev.major`.
```

pschindler:
I see a similar issue with just calculating the eigenstates of a simple $\sigma_x \otimes \sigma_x$ operator

``` python
import qutip as q
import scipy

op = q.tensor(q.sigmax(), q.sigmax())
eigv_sp, eigs_sp = op.eigenstates(sparse=True)
eigv_full, eigs_full = op.eigenstates(sparse=False)
eigv_scipy, eigs_scipy = scipy.linalg.eigh(op.full())
print(eigv_sp, eigs_sp)
print(eigv_full, eigs_full)
print(eigv_scipy, eigs_scipy)
```

The sparse=True gives eigenstates 

```
[Quantum object: dims = [[2, 2], [1, 1]], shape = (4, 1), type = ket
 Qobj data =
 [[ 0.36754009-0.27845703j]
  [-0.49166296+0.21364338j]
  [ 0.49166296-0.21364338j]
  [-0.36754009+0.27845703j]]
````
whereas `sparse=False` and the `scipy.linalg.eigh` give the correct states

```
Quantum object: dims = [[2, 2], [1, 1]], shape = (4, 1), type = ket
 Qobj data =
 [[-0.70710678]
  [ 0.        ]
  [ 0.        ]
  [-0.70710678]]
```

AGaliciaMartinez:
Hi @pschindler, thanks for reporting this! Although the eigenenstates are different for both cases (`sparse=True` and `sparse=False`) I believe that they are both equally valid since the operator has degenerated eigenvalues. The approach for the computation of the eigenstates for both `sparse=True ` and `sparse=False` is different so that may explain why you see different results.

[...]

PR #2629, opened 2025-02-07
Use less the sparse eigen solver

Ericgig:
**Description**
The sparse eigen solver got more finicky since it was updated with last scipy version.
It makes our tests randomly fails quite often:
https://github.com/qutip/qutip/actions/runs/13191102536/job/36824098439?pr=2628
https://github.com/qutip/qutip/actions/runs/13169799701/job/36757919548?pr=2626
https://github.com/qutip/qutip/actions/runs/13151403307/job/36699388137#step:7:14687

We use it wrong as it was never meant to be used to compute all eigen values / vector as once.
We call scipy sparse solver twice and stitch the result together but it's not perfect (#1998).

This change the tests so we use the sparse solver mostly for a limited number of eigen states and make it clean in the docstring.
Also remove `eigs_csr` from the dispatched function. It can be still be used with `Qobj.eigenstates`, but it will not be used without an explicit `sparse` parameter.

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1563,7 +1563,8 @@ class Qobj:
         -----
         The sparse eigensolver is much slower than the dense version.
         Use sparse only if memory requirements demand it.
-
+        It is designed to compute only a few eigenvalues and can give wrong
+        results when all eigenvalues are asked.
         """
         if isinstance(self.data, _data.CSR) and sparse:
             evals, evecs = _data.eigs_csr(self.data,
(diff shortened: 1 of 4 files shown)


qutip/core/qobj.py
```diff
@@ -1682,6 +1683,8 @@ def groundstate(
         -----
         The sparse eigensolver is much slower than the dense version.
         Use sparse only if memory requirements demand it.
+        It is designed to compute only a few eigenvalues and can give wrong
+        results when all eigenvalues are asked.
         """
```
pmenczel, line 1688: Was this meant to go in the `eigenstates` docstring instead? Because in `groundstate`, we are always just asking for 1-2.

Ericgig: Yes, thank you.
For `groundstate`, the sparse solver is actually a good choice and is faster than the dense one from a size of about ~200.

pmenczel (approved):
Thank you, looks good to me

PR #2633, opened 2025-02-12
Improve McResult apidoc

Ericgig:
**Description**
Trying to help someone with mc output, I saw that the McResult apidoc was not great.
It list the parameters needed to create one, but not the output the user has access to:
https://qutip.readthedocs.io/en/latest/apidoc/solver.html#qutip.solver.multitrajresult.McResult

Listing the inherited member is better, but some of the output (`collapse`, `runs_trace`, ...) are attributes.
We also have half public method such as `add` and `add_processor` that are not useful for users.
I rewrote (cut pasted) the docstring to focus on attributes useful for the user and only added specific methods.

I also added `mpmath` to the doc requirement file since I needed it to build the documentation.

### DIFF
--- a/doc/apidoc/solver.rst
+++ b/doc/apidoc/solver.rst
@@ -43,6 +43,7 @@ Monte Carlo Evolution
 .. autoclass:: qutip.solver.mcsolve.MCSolver
     :members:
     :inherited-members:
+    :member-order: bysource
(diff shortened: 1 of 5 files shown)


pmenczel (approved):
This is a great improvement!

PR #2634, opened 2025-02-12
allow np.int as targets in expand_operator

Sola85:
Allow `expand_operator()` to work with `np.int`'s as targets, e.g. in `expand_operator(op, dims, np.ravel_multi_index(...))`.
Currently this would raise `"targets should be an integer or a list of integer"`.


Sola85:
I already contributed `2620.feature` since the last release, but I added an entry for this fix anyway :)

PR #2636, opened 2025-02-14
Add concurrency limit to action

Ericgig:
**Description**
Github action have a `concurrency` clause that control limit the number of job in a group that can run at once.
In a PR late stage, it is not unusual to make a lot of small commit in quick succession (pep8, towncrier, merge suggestions...).
In these case, usually only the latest run is useful.

This `concurrency` will cancel previous jobs and start the latest one quicker if waiting for resources. 
It's set to not cancel jobs in the master branch, but will force them to run one at a time.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -2,6 +2,9 @@ name: Build documentation
 
 on:
   [push, pull_request]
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Thanks!

PR #2643, opened 2025-02-26
Optimize solvers

Ericgig:
**Description**
Benchmarks for QuantumToolbox showed that the difference between us and them was particularly bad for `mcsolve`.
After some looking around, the `adams` method is bad with quantum jumps. It's an ODE method that remembers previous steps and accelerate. Being reset at each jumps is slow when there are a lot of them. Using runge-kutta methods (`vern7`, `dop853`, ...) are therefore more appropriate for mcsolve and I changed the default.

I also checked code that took more time than expected:

- I removed extra copies of the state.
- `np.cumsum` used to look at which jump happen took as much time as a ODE step. Pure python is faster.
- Comparing list format dims is slow.

Overall, mcsolve is 20% faster with the same method, 3x with the change of method for the benchmark used for QuantumToolbox paper, but this is quite problem dependant.

### DIFF
--- a/qutip/solver/integrator/explicit_rk.pyx
+++ b/qutip/solver/integrator/explicit_rk.pyx
@@ -221,7 +221,7 @@ cdef class Explicit_RungeKutta:
         self._norm_front = self._norm_prev
 
         #prepare the buffers
-        for i in range(self.rk_extra_step):
+        for i in range(len(self.k), self.rk_extra_step):
             self.k.append(self._y.copy())
         self._y_temp = self._y.copy()
         self._y_front = self._y.copy()
@@ -288,8 +288,6 @@ cdef class Explicit_RungeKutta:
             self._status = Status.NOT_INITIATED
             return
 
-        self._status = Status.NORMAL
-
         if t == self._t:
             return
 
@@ -298,14 +296,19 @@ cdef class Explicit_RungeKutta:
             return
 
         if self.interpolate and t < self._t_front:
-             self._y = self._interpolate_step(t, self._y)
-             self._t = t
+            if self._status != Status.INTERPOLATED:
+                self._prep_dense_out()
+                self._status = Status.INTERPOLATED
+            self._y = self._interpolate_step(t, self._y)
+            self._t = t
+
+        self._status = Status.NORMAL
 
         if step and self._t < self._t_front and t > self._t_front:
             # To ensure that the self._t ... t_out interval can be covered.
             t = self._t_front
 
-        while self._t_front < t:
+        while self._t_front < t and self._status >= 0:
             self._y_prev = copy_to(self._y_front, self._y_prev)
             self._t_prev = self._t_front
             self._norm_prev = self._norm_front
--- a/qutip/solver/integrator/explicit_rk.pyx
+++ b/qutip/solver/integrator/explicit_rk.pyx
@@ -302,6 +302,7 @@ cdef class Explicit_RungeKutta:
                 self._status = Status.INTERPOLATED
             self._y = self._interpolate_step(t, self._y)
             self._t = t
+            return
 
         self._status = Status.NORMAL
 
--- a/qutip/solver/integrator/explicit_rk.pyx
+++ b/qutip/solver/integrator/explicit_rk.pyx
@@ -221,7 +221,8 @@ cdef class Explicit_RungeKutta:
         self._norm_front = self._norm_prev
 
         #prepare the buffers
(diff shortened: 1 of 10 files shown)


Ericgig:
Changing the default method revealed a few issues:
- `Explicit_RungeKutta` could prepare unlimited buffers for `k` if initiated multiple times as done in `mcsolve`.
- `Explicit_RungeKutta` now check if the dense output is prepared before using it.
- Removed `MKL_VERBOSE` as it use too much ram that it would affect the tests.
- `vern7` / `vern9` will default to dense for the state. It support any data layers, but would be so slow tests would timeout.
- `vern7` step are too long to find in the collapse in 5 steps. When unlucky, worst cases can be very long, an options was added to avoid them.

qutip/solver/mcsolve.py
```diff
@@ -376,15 +385,19 @@ def _do_collapse(self, collapse_time, state):
...
-            probs = np.zeros(len(self._n_ops))
-            for i, n_op in enumerate(self._n_ops):
-                probs[i] = n_op.expect_data(collapse_time, state).real
-            probs = np.cumsum(probs)
-            which = np.searchsorted(probs,
-                                    probs[-1] * self._generator.random())
+            probs = [
+                n_op.expect_data(collapse_time, state).real
+                for i, n_op in enumerate(self._n_ops)
+            ]
```
pmenczel, line 395:
```suggestion
            probs = [
                n_op.expect_data(collapse_time, state).real
                for n_op in self._n_ops
            ]
```

qutip/solver/mcsolve.py
```diff
@@ -846,9 +860,14 @@ def options(self) -> dict:
...
 
-        norm_steps: int, default: 5
+        norm_steps: int, default: 25
             Maximum number of tries to find the collapse.
 
+        norm_min_step: float, default: 0.10
+            Minimum split used when finding jump location.
+            A small non-zero value can help avoid the worst cases convergence
+            when finding jumb at the cost of increased average steps.
+
```
pmenczel, line 870:
```suggestion
        norm_min_step: float, default: 0.10
            Minimum step used when finding the collapse time, given as a fraction
            of the search interval. Must be between 0 and 0.5.
            A small non-zero value can help avoid the worst cases of
            convergence at the cost of increased average steps required to find
            the collapse.

```

qutip/solver/nm_mcsolve.py
```diff
@@ -653,6 +654,11 @@ def options(self) -> dict[str, Any]:
         norm_steps: int, default: 5
             Maximum number of tries to find the collapse.
 
+        norm_min_step: float, default: 0.0
+            Minimum split used when finding jump location.
+            A small non-zero value can help avoid the worst cases convergence
+            when finding jumb at the cost of increased average steps.
```
pmenczel, line 660: Same as in MCSolver for the text. I was wondering why you set it to zero here?

qutip/solver/result.py
```diff
@@ -92,6 +92,10 @@ def __init__(self, options, *, solver=None, stats=None):
         if hasattr(options_copy, "_feedback"):
             options_copy._feedback = None
         self.options = options_copy
+        # Almost all integrators already return a copy that is safe to use.
+        self._integrator_return_not_copy = options.get("method", None) in [
+            "vern7", "vern9"
+        ]
```
pmenczel, line 98: Would it be safer if we list the integrators that do return a copy? In case a new integrator is added that doesn't copy and forgotten to add here...

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -298,14 +297,19 @@ cdef class Explicit_RungeKutta:
...
         if self.interpolate and t < self._t_front:
-             self._y = self._interpolate_step(t, self._y)
-             self._t = t
+            if self._status != Status.INTERPOLATED:
+                self._prep_dense_out()
+                self._status = Status.INTERPOLATED
+            self._y = self._interpolate_step(t, self._y)
+            self._t = t
+
+        self._status = Status.NORMAL
```
pmenczel, line 307: Will this not always overwrite the `self._status = Status.INTERPOLATED` a few lines above?

qutip/solver/nm_mcsolve.py
Ericgig: It's not needed with adams method and i did not change the default for nm_mcsolve.
I did not tests if verner is better for it.

pmenczel (approved):

PR #2647, opened 2025-03-06
Add group alias for data types

Ericgig:
**Description**
Add group alias.
This will allow to set the default to `"AnyJax"` and it will automatically choose between JaxArray and JaxDia.
With the option to apply default dtype more strongly added in #2519, it would allow to more easily change backends. 

For now dense, csr, and dia make the `core` or `cython` group.

`CoreOptions(default_dtype="AnyJax", default_dtype_scope="full")`

(ps. `"AnyJax"` alias does not exist yet. It will probably be just `"jax"` which is now associated with dense jax array.)


qutip/core/data/convert.pyx
```diff
@@ -15,13 +15,13 @@ initialisation of the `data` module.
...
 import numbers
-
+from collections import namedtuple
 import numpy as np
 from scipy.sparse import dok_matrix, csgraph
 cimport cython
 from qutip.core.data.base cimport Data
 
-__all__ = ['to', 'create']
+__all__ = ['to', 'create', '_parse_defaut_dtype']
```
nwlambert, line 24: just a spelling error here and elsewhere:
`_parse_default_dtype`

qutip/core/data/__init__.py
```diff
@@ -34,18 +34,19 @@
...
     (Dia, Dense, dia.from_dense, 1.4),
     (Dense, Dia, dense.from_dia, 1.2),
     (Dia, CSR, dia.from_csr, 1),
     (CSR, Dia, csr.from_dia, 1),
-])
+], _defer=True)
 to.register_aliases(['csr', 'CSR'], CSR)
 to.register_aliases(['Dense', 'dense'], Dense)
 to.register_aliases(['DIA', 'Dia', 'dia', 'diag'], Dia)
+to.register_group(['core', 'cython'], dense=Dense, sparse=CSR, diagonal=Dia)
```
nwlambert, line 49: do you want to catch capitalizations of these as well?

nwlambert (approved):
looks good, I think I follow most of the logic.

Just to make sure I understand, now the dtype of prebuilt objects will be assigned from the slightly general names of 'dense, sparse, diagonal', and then if there is  a default group set the appropriate dtype from that group for that name format will be used?

is there any issue if there ends up being more than those 3 names needed in the future? (some new sparse format for example?)

very minor question, if you register a new group with _defer = True, the group wont work until you manually rebuild the lookup?

Ericgig:
If the passed `dtype` (as parameter or settings) is a group alias, the appropriate format will be chosen from the set }dense, sparse, diagonal}.

If another format of is added, we will see then. Mostly "sparse" could be any non-structured sparse format: COO, CSR, CSC, etc are all appropriate for it and I don't think we would need many of these.
New dtype can be added without belonging to any group.

Yes if you build a group with `_defer = True` it will not work until something else trigger the rebuild.

nwlambert (approved):

PR #2648, opened 2025-03-07
Dysolve

MathiB123:

**Description**
Add Dysolve as a feature to compute time propagators. Work in progress.


qutip/solver/dysolve_propagator.py
```diff
@@ -0,0 +1,390 @@
...
+        self.omega = omega
+        self.t_i = None
+        self.t_f = None
+        self.dt = None
+        self.times = None
+        self.a_tol = a_tol
+        self._Sns = None
+        self.Us = None
+
+    def __call__(self, t_i: float, t_f: float, dt: float):
```
Ericgig, line 62: This method should return the propagator at the given time, or between the given times:
```
U = DysolvePropagator(...)
psi(t=2) = U(2) @ psi(t=0)
psi(t=4) = U(4, 2) @ psi(t=2)
psi(t=0) = U(0, 4) @ psi(t=4)
U(t, t) == qeye(N)
```

doc/guide/guide-dynamics.rst
```diff
@@ -9,6 +9,7 @@ Time Evolution and Quantum System Dynamics
 
    dynamics/dynamics-intro.rst
    dynamics/dynamics-data.rst
+   dynamics/dynamics-dysolve-propagator.rst
```
Ericgig, line 12: Maybe move it after `dynamics/dynamics-propagator.rst`?
But it does not belong before `dynamics/dynamics-master.rst`.

qutip/solver/dysolve_propagator.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+
+    def __init__(
+        self,
+        max_order: int,
+        H_0: Qobj,
+        X: Qobj,
+        omega: float,
+        a_tol: float = 1e-10,
+    ):
```
Ericgig, line 48: Lets try to match the existing one closer:
The Hamiltonian first.
`a_tol`, `max_order` and `dt` in options. (keyword only parameter)

Ericgig (commented):
I took a first look, let's match the user interface with the one of the existing functions.

It would be good to have a non-Qubit test.
Something simple like: H = qeye, X = destroy + create, w=1.
This would also get a lot of special case of `w=0` in the integration.

qutip/solver/dysolve_propagator.py
```diff
@@ -0,0 +1,399 @@
...
+        self,
+        H_0: Qobj,
+        X: Qobj,
+        omega: float,
+        options: dict[str] = None,
+    ):
+        # System
+        self.H_0 = H_0
+        self.eigenenergies = H_0.diag()
+        self.X = X.transform(H_0)
```
Ericgig, line 49: This is not how transform work...
The input should be a basis: all ket in of the new basis.
If `H_0` is diagonal, it should be `X.transform(qeye_like(H_0))`.

Usage:
```
        self.eigenenergies, self.basis = H_0.eigenstates()
        self.X = X.transform(self.basis)
```

qutip/solver/dysolve_propagator.py
```diff
@@ -0,0 +1,399 @@
...
+
+        """
+        matrix_elements_products = np.ones(
+            (i_j_indices.shape[0]), dtype=np.complex128)
+
+        for i in range(i_j_indices.shape[1] - 1):
+            matrix_elements_products *= self.X[
+                i_j_indices[:, 0 + i: 2 + i][:, 0],
+                i_j_indices[:, 0 + i: 2 + i][:, 1]
+            ].A1
```
Ericgig, line 214: `A1` does not always exist.
`ravel` can be using instead.

qutip/tests/solver/test_dysolve_propagator.py
```diff
@@ -0,0 +1,127 @@
...
+    (lambda eff_omega: np.array([eff_omega]),
+     lambda eff_omega, dt: (-1j/eff_omega) * (np.exp(1j*eff_omega*dt) - 1)),
+    (np.array([0, 0, 0, 0, 0]), lambda _, dt: (dt**5) / factorial(5)),
+    (np.array([1e-12, 1e-12, 1e-12]), lambda _, dt: (dt**3) / factorial(3)),
+    (lambda eff_omega: np.array([eff_omega, 0]),
+     lambda eff_omega, dt: (-1j / eff_omega) * ((-1j / eff_omega)
+                                                * (np.exp(1j*eff_omega*dt)-1) - dt)),
+    (lambda eff_omega: np.array([0, eff_omega]), lambda eff_omega, dt:
+     (-1j*dt/eff_omega) * np.exp(1j*eff_omega*dt) -
+     ((1j/eff_omega)**2) * (np.exp(1j*eff_omega*dt)-1))
```
Ericgig, line 28: It's a bit confusing what par is a `ws` and which is an `answer`...

PR #2654, opened 2025-03-19
Safer metadata import for older python

Ericgig:
**Description**
In older version of python `importlib`'s `metadata` must be imported directly.


Ericgig:
@Langhaarzombie, @ytdHuang, this should fix the bug you have with in `about`.

ytdHuang (approved):

PR #2655, opened 2025-03-19
Bugfix dimensions.Compound.remove

pmenczel:
I just came across this in `dimensions.py`. The current code looks definitely wrong. In the docstring it says
```Space([2, 3, 4]).remove(1) == Space([2, 4])```
but currently
``Space([2, 3, 4]).remove(1)``
gives an error because the argument is not a list, and
``Space([2, 3, 4]).remove([1])``
gives an error as well.

I am not sure if this is even used anywhere, but I have tried to fix it.

### DIFF
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -534,26 +534,6 @@ class Space(metaclass=MetaSpace):
         """ Dimensions as a flat list. """
         return [self.size]
 
-    def remove(self, idx: int):
-        """
-        Remove a Space from a Dimensons or complex Space.
-
-        ``Space([2, 3, 4]).remove(1) == Space([2, 4])``
-        """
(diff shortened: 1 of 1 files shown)


Ericgig:
Since, in theory, nothing stop us from having compound space with other space that have structure (Super, ENR, ...) it should be recursive to be right...

Since you have found the bug, did you have a use case for it.
All I can think would be to remove traced dims when using `ptrace`. Presently it does not support any fancy dimensions and use list directly.

pmenczel:
I do not have a use case for it. As discussed, since the `remove` and `replace` functions appear to be unused and mostly buggy (and they have a bit unclear semantics, at least to me), it might be better to just remove these functions completely.

Ericgig (approved):

PR #2657, opened 2025-03-20
Add faster implementation of ptrace for kets

Sola85:
**Description**
For my application, `Qobj.ptrace()` was a bottleneck, so I added a specialization for kets that is significantly faster.
The new implementation uses `np.einsum` to combine the computation of the outer product and the partial trace into one operation, instead of doing these two operations sequentially.

```python
from qutip import rand_ket
from time import perf_counter

state = rand_ket([2, 3, 4, 5]*2, distribution="fill")
keep_dims = [1, 3, 7]

t0 = perf_counter()
reduced_dm2 = state.ptrace(keep_dims) # Uses new implementation
t1 = perf_counter()
print("New impementation:", t1 - t0) # ~ 0.01s on 8th-gen Intel quad-core


t0 = perf_counter()
reduced_dm1 = state.proj().ptrace(keep_dims) # Old implementation
t1 = perf_counter()
print("Old impementation:", t1 - t0) # ~ 6s on 8th-gen Intel quad-core

assert (reduced_dm1 - reduced_dm2).norm("tr") < 1e-10
```

In this example, the new implementation is around 600x faster on my machine. 

The new implementation can also handle significantly larger states:
`rand_ket([2, 3, 4, 5]*3, distribution="fill").ptrace(keep_dims)` fails using the old implementation because the intermediate density matrix is too large, but the new implementation computes the result in less than 0.1s.

Note: I did not use `qutip.einsum` because this only seems to wrap the string-indexed version of `np.einsum` and not the array-indexed version. The latter is more useful is cases like the present one, since its easier do generate the einsum subscripts programatically.


qutip/tests/core/test_ptrace.py
```diff
@@ -130,3 +130,20 @@ def test_ptrace_operator(dtype, sel):
...
     assert A.ptrace(sel) == expected(A, sel)
+
+@pytest.mark.parametrize('dims, sel',
+                         [
+                             ([5, 2, 3], [2, 1]),
+                             ([5, 2, 3], [0, 2]),
+                             ([5, 2, 3], [0, 1]),
+                             ([2]*6, [3, 2]),
+                             ([2]*6, [0, 2]),
+                             ([2]*6, [0, 1]),
```
Ericgig, line 141: These pattern seems redundant. 
Why not also vary the number of kept indices?

qutip/core/tensor.py
```diff
@@ -506,3 +506,48 @@ def expand_operator(
...
+def _ptrace_specialization_kets(state: Qobj, keep_dims: list[int]) -> Qobj:
+    """ Faster implementation of ptrace for kets.
+
+    Example:
+        If ``|psi> = s_ijk |ijk>`` is a ket with three dimensions,
+        this code directly computes e.g. ``\\sum_j s_ijk s*_njm`` without first
+        having to compute ``|psi><psi|``, which can be prohibitively large.
+
+        The comments below refer an example with ``state.dims = [12, 13, 14]``
+        and ``keep_dims = [0, 2]`` (i.e. the dimensions with size 12 and 14)
```
Ericgig, line 520: I feel this should be a comment and not part of the docstring.

Ericgig (commented):
Thank you.
This look quite useful.

The function does not belong in `tensor.py`...
I know `qobj.py` is already too big, but it should probably come after the `ptrace` function already there for now. This would avoid adding a circular import.

qutip/tests/core/test_ptrace.py
Sola85: I had just copied an existing test without looking too much at it :D I now added some more relevant test cases

Sola85:
Thanks for the comments. I moved the implementation to `qobj.py`, added more interesting test cases and fixed the docstring.

Ericgig (approved):
Thank you!

Issue #2651, opened 2025-03-14
Implementing a commutator superoperator for `QobjEvo`


ottosalmenkivi:
### Problem Description

Assuming I want to create a commutator superoperator $\mathcal{H(t)}\rho = [H(t), \rho] = H(t)\rho - \rho H(t)$, where $H$ is a time-dependent operator while $\rho$ does not depend on time. With qutip I can do it with `qt.spre` and `qt.spost`. Since $H(t)$ is described by a `QobjEvo` object, the output of the superoperator generators will also be a `QobjEvo`. The bug is that summing the pre- and postmultiplication parts outputs a `QobjEvo` with double the elements as compared to $H(t)$. I figure this is how `QobjEvo` summation is supposed to work. However, since the time-dependence for the pair is the same, the matrices could be added together but such functionality does not exist. To my understanding `QobjEvo.compress()` does not recognize identical time-dependence. The superoperator is physically correct, but could be equivalently desribed by fewer terms.

In a nutshell, taking the commutator superoperator doubles the number of elements in a `QobjEvo`, because information of identical time-dependence is not taken into account. 

A simple example can be reproduced with
```
import qutip as qt
ham = qt.QobjEvo([qt.sigmaz(), [qt.sigmax(), 'sin(t)']])

scom = qt.spre(ham) - qt.spost(ham)
scom.compress() # We can compress the contant Hamiltonian terms
print(scom)
``` 
that outputs a `QobjEvo` with three elements:
```
> QobjEvo: dims = [[[2], [2]], [[2], [2]]], shape = (4, 4), type = super, superrep = super, isconstant = False, num_elements = 3
```

### Proposed Solution

One could simply create new superoperator generator functions for the commutation and anti-commutation relations, so that apply that to the individual terms in a `QobjEvo`. This would preserve the number of terms.

```
@qt.superoperator._map_over_compound_operators
def scommutator(A):
    return qt.spre(A) - qt.spost(A)

@qt.superoperator._map_over_compound_operators
def santicommutator(A):
    return qt.spre(A) + qt.spost(A)
```

### Alternate Solutions

One could think about, in increasing order of complexity:
- combining the two suggested functions above and triggering anticommutator as a Boolean flag,
- creating special logic in functions where (anti)commutator superoperators are calculated to recombine the time-dependences, or
- add logic to `QobjEvo` that recognises identical time-dependence.

### Additional Context

[...]

PR #2658, opened 2025-03-21
Improve redundancy detection in QobjEvo

Ericgig:
**Description**
Improve `QobjEvo.compress` to also merge identical coefficient.

This allow for the like of Lindblad Dissipator of time-dependant system to be a single term in the operator instead of 3.

- An `__eq__` method was added to `Coefficient`.
- `compress` is automatically called in Liouvillian and Lindblad Dissipator.
- Improved coefficient operation to avoid creating new one when possible:
  - `ConstantCoefficient` product, addition and conjugate return a `ConstantCoefficient` and not `ConjCoefficient(ConstantCoefficient)`, etc.
  - Similar for spline interpolation.
  - The conjugate of a Coefficient will return itself if it can detect it's real. We can do so for constant and spline coefficient. Also, for function, we look at the output type hint if present.
- Also removed the unused `_cdc` method.
- Have Hamiltonian compute `H rho - rho H.dag()` to support non-Hermitian Hamiltonian.

**Related issues or PRs**
Partial fix of #2651, the super (anti)commutator function could still be nice to have.


qutip/core/coefficient.py
```diff
@@ -480,21 +480,26 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
     cdef_cte = ""
     init_cte = ""
     copy_cte = ""
+    _eq__cte = ""
```
hodgestar:
```suggestion
    eq_cte = ""
```
Nitpick: I would just call this `eq_cte`. A leading underscore in a local variable conventionally indicates an unused value.

qutip/core/coefficient.py
```diff
@@ -480,21 +480,26 @@ def make_cy_code(code, variables, constants, raw, compile_opt):
...
     for i, (name, val, ctype) in enumerate(constants):
         cdef_cte += "        {} {}\n".format(ctype, name[5:])
         copy_cte += "        out.{} = {}\n".format(name[5:], name)
         init_cte += "        {} = cte[{}]\n".format(name, i)
+        _eq__cte += "            c_other.{} == {},\n".format(name[5:], name)
```
hodgestar:
```suggestion
        _eq__cte += "            other.{} == {},\n".format(name[5:], name)
```
Could we just name this `other` here and avoid doing `StrCoeffcient c_other = other` in `__eq__`?

qutip/core/coefficient.py
```diff
@@ -576,12 +581,10 @@ def __eq__(self, other):
         if self is other:
             return True
         # If compiled with cython 0.X, self and other could be inverted.
-        if (
-            not isinstance(self, StrCoefficient) or
-            not isinstance(other, StrCoefficient)
-        ):
+        if type(self) is not type(other):
```
hodgestar, line 584: Thinking through this change:

Previously `self` and `other` could be different sub-classes of `StrCoefficient`. Now they must be the same sub-class.

With the change they must be the same sub-class.

That seems valid.

qutip/core/coefficient.py
Ericgig: I tried it first, not that simple. 
`other` is a python object of any type so we can't give it a ctype in the function declaration.
The attributes are all in the cython namespace, so we need to eventually cast it to access them.

Two variables is the simplest solution, but having two methods could also work.

qutip/core/coefficient.py
Ericgig: `comp_cte`?

qutip/core/coefficient.py
hodgestar: Ah, that makes sense. Perhaps add a comment like `# cast other to StrCoefficient`?

qutip/core/coefficient.py
hodgestar: `eq_cte` seems more obvious to me. `comp` seems to  imply `compare` which in Python would imply `<`, `>`, `>=`, etc.

BoxiLi:
This seems to raise errors when array-based coefficients are used with different array lengths?


There are some new errors in the qutip-qip tests.

```
FAILED tests/test_device.py::test_numerical_evolution[LinearSpinChain-CNOT] - ValueError: operands could not be broadcast together with shapes (8,) (5,)
FAILED tests/test_device.py::test_numerical_evolution[CircularSpinChain-CNOT] - ValueError: operands could not be broadcast together with shapes (8,) (5,)
FAILED tests/test_device.py::test_numerical_evolution[SCQubits-ISWAP] - ValueError: operands could not be broadcast together with shapes (381,) (1361,)
FAILED tests/test_device.py::test_numerical_evolution[SCQubits-CNOT] - ValueError: operands could not be broadcast together with shapes (141,) (121,)
FAILED tests/test_device.py::test_numerical_circuit[ASAP-DispersiveCavityQED] - ValueError: operands could not be broadcast together with shapes (17,) (7,)
FAILED tests/test_device.py::test_numerical_circuit[ASAP-LinearSpinChain] - ValueError: operands could not be broadcast together with shapes (12,) (20,)
FAILED tests/test_device.py::test_numerical_circuit[ASAP-CircularSpinChain] - ValueError: operands could not be broadcast together with shapes (12,) (7,)
FAILED tests/test_device.py::test_numerical_circuit[ASAP-SCQubits] - ValueError: operands could not be broadcast together with shapes (2101,) (1601,)
FAILED tests/test_device.py::test_numerical_circuit[ALAP-DispersiveCavityQED] - ValueError: operands could not be broadcast together with shapes (17,) (7,)
FAILED tests/test_device.py::test_numerical_circuit[ALAP-LinearSpinChain] - ValueError: operands could not be broadcast together with shapes (12,) (20,)
FAILED tests/test_device.py::test_numerical_circuit[ALAP-CircularSpinChain] - ValueError: operands could not be broadcast together with shapes (12,) (7,)
FAILED tests/test_device.py::test_numerical_circuit[ALAP-SCQubits] - ValueError: operands could not be broadcast together with shapes (1601,) (2101,)
FAILED tests/test_device.py::test_numerical_circuit[None-DispersiveCavityQED] - ValueError: operands could not be broadcast together with shapes (21,) (7,)
FAILED tests/test_device.py::test_numerical_circuit[None-LinearSpinChain] - ValueError: operands could not be broadcast together with shapes (13,) (20,)
FAILED tests/test_device.py::test_numerical_circuit[None-CircularSpinChain] - ValueError: operands could not be broadcast together with shapes (13,) (7,)
FAILED tests/test_device.py::test_numerical_circuit[None-SCQubits] - ValueError: operands could not be broadcast together with shapes (2161,) (1641,)
FAILED tests/test_noise.py::TestNoise::test_collapse_with_different_tlist - ValueError: operands could not be broadcast together with shapes (100,) (2,)
```

Ericgig:
Thank @BoxiLi, I was used to see that test failing so I did not check.
It's now fixed, but there is still a failure that is not a warning: `test_phaseflip_correction_simulation`
Can this be another bug in this PR or can it fail randomly?

Edit: that test is failing elsewhere too, so not related to this.

BoxiLi:
Yes, indeed, this is something that seems to fail randomly. Unrelated here. I will investigate.

Ericgig:
If you have it in one of your branch, I would prefer you to make a PR for it.

PR #2664, opened 2025-03-25
Update shape in qobjevo dag/trans

Ericgig:
**Description**
There is a bug in `QobjEvo.dag` and `QobjEvo.trans`, where the shape was not updated, only the dims and data.
The Qobj created when calling the function would have the right shape, so it was missed in the tests.


pmenczel (approved):
Thank you for the quick fix!

PR #2667, opened 2025-03-27
Use slots to speed up numpy_backend

Ericgig:
**Description**
Speed up the numpy backend class using `__slots__` as proposed by Simon in #2509.

### DIFF
--- a/qutip/core/numpy_backend.py
+++ b/qutip/core/numpy_backend.py
@@ -2,8 +2,46 @@ from ..settings import settings
 
 
 class NumpyBackend:
+    __slots__ = (
+        "_qt_np",
+        "abs",
+        "angle",
+        "arccos",
+        "array",
+        "ceil",
+        "conj",
+        "cos",
+        "cumsum",
+        "diff",
+        "e",
+        "histogram",
+        "imag",
+        "inf",
+        "inner",
+        "linalg",
+        "log",
+        "log2",
+        "max",
+        "maximum",
+        "ones",
+        "pi",
+        "prod",
+        "real",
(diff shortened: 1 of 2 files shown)


qutip/core/numpy_backend.py
```diff
@@ -2,8 +2,46 @@
...
+        "zeros_like",
+    )
+
     def _qutip_setting_backend(self, np):
         self._qt_np = np
+        for slot in self.__slots__:
+            if slot != "_qt_np":
+                setattr(self, slot, getattr(np, slot))
 
     def __getattr__(self, name):
```
hodgestar, line 46: Is the `__getattr__` still needed? Is it to allow access to attributes that are not in `__slots__`? If so, do we actually want to allow arbitrary attribute access on the backend? Or only the ones we specify here?

Ericgig: It will help external contributors, ideally they should not need to know how this work to contribute.
I also fear that some call in a line that is missed in tests will creep up eventually and with this it will still work.

qutip/tests/core/test_options.py
```diff
@@ -1,14 +1,21 @@
...
+
 # Mocking JAX to demonstrate backend switching
-mock_jax = Mock
-mock_jax.sum = Mock(return_value="jax_sum")
+class Mock_JAX:
+    def sum(*a, **kw):
+        return "jax_sum"
+
+    def __getattr__(self, _):
+        return "other"
```
hodgestar, line 15: Do we need the `__getattr__` in the tests?

Issue #2646, opened 2025-03-05
Instantiating a qt.QobjEvo with a qt.coefficient


kylewamer:
It is possible to instantiate a qt.QobjEvo object with a qt.coefficient. However, the "args" argument QobjEvo does not behave as expected. Is this a bug, or is it incorrect to pass a qt.coefficient object to qt.QobjEvo? I am using qutip v5.0.4

```
import qutip as qt
A = qt.sigmaz()
def f(t, args):
    return t + args["g"]

coeff = qt.coefficient(f, args={"g":1})

evo_1 = qt.QobjEvo([A, coeff])
evo_2 = qt.QobjEvo([A, coeff], args={"g:2}) # Updating the args dictionary does nothing.

assert evo_1(1) == evo_2(1) 
```

Ericgig:
Thank you for reporting.
`Coefficient` can be used to build `QobjEvo`, ignoring `args` is a bug.

The `arguments` method can be used to overwrite (all) the `args` for now:

```
import qutip as qt
import numpy as np
A = qt.sigmaz()

def f(t, g):
    return t + g

coeff1 = qt.coefficient(f, args={"g":1})
coeff2 = qt.coefficient(f, args={"g":2})

evo_1 = qt.QobjEvo([[A, coeff1], [A, coeff2]])  # 2 different args kept.
evo_1.arguments(g=1)  # Can't update only a specific coeff args
evo_2 = 2 * A * coeff1

assert evo_1(1) == evo_2(1)
```

kylewamer:
Thanks for the workaround. Should I rename this issue as a bug report, or can you update the tag?

Ericgig:
The cause is that the `coefficient` function does not update `args`: 
```
f = lambda t, v: v
coeff1 = coefficient(f, args={"v": 0})
coeff2 = coefficient(coeff1, args={"v": 1})
assert coeff2(0) == 0  # should have been updated to 1
```

Marking as good first issue, should be an easy fix + one test.

PR #2668, opened 2025-03-27
Fix QobjEvo args bug in coefficient function

Picsidust:
### Issue Reference: #2646

This PR ensures that `args` is correctly initialized when creating a `Coefficient`. Previously, arguments passed to `QobjEvo` with a `Coefficient` were not properly updated. This was caused by the function using a mutable default argument (`args={}`), which could lead to unintended mutations across instances.

### Fix:
- Changed `args={}` to `args=None` and initialized inside the function.
- Ensured `args.copy()` is used to avoid modifications to original input.
- Added tests to confirm the fix.

✅ Tests pass successfully!

### DIFF
--- a/qutip/test_qobjevo.py
+++ b/qutip/test_qobjevo.py
@@ -1,0 +1,10 @@
+import pytest
+from qutip import QobjEvo
+
+def test_qobjevo_return_updates_args():
+    """Test that _return correctly updates args"""
+    qobj = QobjEvo([("sigmax()", lambda t, args: args["a"])], args={"a": 1})  # Initial args
+    qobj_updated = qobj._return({"a": 2})  # Should update args
+
+    assert qobj_updated.args["a"] == 2, "Args were not updated correctly"
+
--- a/qutip/core/coefficient.py
+++ b/qutip/core/coefficient.py
@@ -61,8 +61,8 @@ def coefficient(
     base: CoefficientLike,
     *,
     tlist: ArrayLike = None,
-    args: dict = {},
-    args_ctypes: dict = {},
+    args: dict = None,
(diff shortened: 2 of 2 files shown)


pmenczel (changes requested):
Hi and thank you for your contribution. If you look [here](https://github.com/qutip/qutip/pull/2668/files), you will see that you are asking to change 180 files in the repository, so something must have gone wrong. Please remove all unrelated changes from the PR. (Perhaps you only need to restore the `.gitignore` file?)

Ericgig (commented):
Thank you for the PR.
Safer input is good, but it's not the cause of the bug in #2646.
The issue is in the function `_return`, it should update the `args` when provided.

Could you add a test that would failed before your fix. You can use the bug report example for it.

Picsidust:
Yes sure I can.

qutip/test_qobjevo.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+from qutip import QobjEvo
+
+def test_qobjevo_return_updates_args():
+    """Test that _return correctly updates args"""
+    qobj = QobjEvo([("sigmax()", lambda t, args: args["a"])], args={"a": 1})  # Initial args
+    qobj_updated = qobj._return({"a": 2})  # Should update args
+
+    assert qobj_updated.args["a"] == 2, "Args were not updated correctly"
```
Ericgig, line 9: This should be added in `qutip/tests/core/test_qobjevo.py`. Here it's not ran with the automated tests.

This test is not releated to the bug in the original issue. `QobjEvo._return` does not exist. You can't retrieve the arguments from a `QobjEvo` (`QobjEvo.args` does not exist).

PR #2671, opened 2025-03-30
Add documentation for 64-bit integer indices in QuTiP.

Picsidust:
### Description
This PR adds documentation for building QuTiP with 64-bit integer indices. Previously, this information was only available in the changelog, but now a dedicated guide has been added under `doc/advanced/64bit_indices.rst`.

#### Changes Made:
✅ Created a new documentation file: `doc/advanced/64bit_indices.rst`
✅ Updated `doc/advanced/index.rst` to include the new guide
✅ Ensured documentation builds successfully (`make html`)

This guide provides a step-by-step explanation of how to enable 64-bit indexing in QuTiP using:

```bash
python -m build --wheel --config-setting="--global-option=--with-idxint-64"


Ericgig (commented):
Looking good.
Could you remove the advanced/index.rst file  and add a towncrier entry.

Ericgig (changes requested):
Not sure why, but the towncrier file is seen as a binary file by github.
Why did you choose to call it `news/next/doc/2671.feature`?
Creating a new top level directory for a `good first issue` is almost always wrong.
The right location is written in the [contribution guide](https://github.com/qutip/qutip/blob/master/doc/development/contributing.rst#changelog-generation).

PR #2678, opened 2025-04-10
Enable again to plot states with RGBA tuple color on the Bloch sphere

PositroniumJS:
**Description**
Before #2445, it was unwanted but possible to plot a state as a point on the Bloch sphere using a color as a RGB or RGBA tuple. This PR makes it possible again and also possible now when plotting as a vector. It also makes possible to specify a color as a string for all the states.

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -4,8 +4,7 @@ import os
 from typing import Literal
 
 import numpy as np
-from numpy import (outer, cos, sin, ones)
-
+from numpy import cos, ones, outer, sin
 from packaging.version import parse as parse_version
 
 from . import Qobj, expect, sigmax, sigmay, sigmaz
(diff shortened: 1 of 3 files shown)


Ericgig (commented):
Thank you!

How hard would it be to add a test for all supported colors format?
I fear that if we don't add a test, nothing would stop us from breaking it once again in the future...

PositroniumJS:
> Thank you!
> 
> How hard would it be to add a test for all supported colors format? I fear that if we don't add a test, nothing would stop us from breaking it once again in the future...

I need to learn how writing tests works but I don't think that it is really complicated. (But here, nothing was broken, it is rather than an unexpected behavior was working, plotting the states as vectors would have raised an error, what was not the case for plotting as points).

qutip/bloch.py
```diff
@@ -404,14 +403,19 @@ def add_states(self, state: Qobj,
             colors = np.asarray(colors)
 
             if colors.ndim == 0:
+                colors = np.repeat(colors, state.shape[0])
+
+            elif (colors.ndim == 1
+                  and np.isdtype(colors.dtype, ("integral", "real floating"))):
```
Ericgig, line 409: `isdtype` is a numpy 2.0 feature. But we can't drop support for numpy 1.X yet

PositroniumJS: Ok, then should I only check if it is `np.float64` or `np.int64` even if other bit numbers (and integer types) would have worked?

Ericgig: Maybe check that it's not string?

Ericgig: `np.issubdtype(colors.dtype, np.number)`?

PositroniumJS: Good catch, I did not expect it to be available with numpy 1. I can use it with `np.integer` and `np.floating`

PR #2684, opened 2025-04-23
Fix micro python version in 3.12 fallback test

Ericgig:
**Description**
One of the tests is failing, timing out after finishing the tests:

```
(48156 durations < 1s hidden.  Use -vv to show these durations.)
========== 16036 passed, 103 skipped, 6 xfailed in 1160.36s (0:19:20) ==========
Error: The action 'Run tests' has timed out after 60 minutes.
```

It's happening after merging dysolve and python 3.12.10 being release.
This is to test if the python release had an impact.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,14 +64,14 @@ jobs:
           # Use conda-forge to provide Python 3.11 and latest numpy
           - case-name: p312, numpy fallback
             os: ubuntu-latest
-            python-version: "3.12.9"
+            python-version: "3.12"
             numpy-requirement: ">=1.26,<1.27"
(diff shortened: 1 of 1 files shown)


pmenczel:
My guess is that the problem is MPI related. The 3.12 fallback setup is the only one where we test MPI. With the MPI stuff, I remember it happening before multiple times that something hangs, and pytest captures any error messages there might be before that and doesn't print them because it would only do so at the end of the test.

I haven't been able to reproduce locally. Is there a way we can run the runner with "-s" flag for pytest?

Ericgig:
@pmenczel, you're unto something. Removing the mpi tests and 3.12.10 also fix the issue.

PR #2685, opened 2025-04-30
Update openmpi version used in tests

pmenczel:
We are including the MPI tests in only one of the test runs. I am changing that to be the one with the latest python and numpy (which it was always supposed to be, but it didn't get updated when python 3.13 came out). I am also changing openmpi, which was pinned to `<5` due to dependency issues, to the latest version.

This gets rid of some warning messages, and it seems to fix the problem with the test timing out. I still do not know exactly what caused that problem, but I am fine if things are working with the latest openmpi / mpi4py versions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,6 +46,9 @@ jobs:
             numpy-build: ">=2.0.0"
             numpy-requirement: ">=2.0.0"
             pypi: 1
+            # Install mpi4py to test mpi_pmap
+            # Should be enough to include this in one of the runs
+            includempi: 1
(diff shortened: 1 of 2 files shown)


Ericgig (approved):
As a solution for the test, this should be sufficient. 
Should we add a note in the documentation or we expect it should not happen to users?

pmenczel:
That failing test can't be related, right?

Ericgig:
No idea, that's a new one to me...

pmenczel:
Today it's working again. I guess it was a temporary problem somewhere not on our side

PR #2688, opened 2025-05-06
Fix parameter name inconsistency in qsave

omnster:
The function definition and body used to use the parameter `name`, while the docstring used to refer to `filename`.

This commit changes the instances of `name` to `filename` to be consistent with the docstring.


Ericgig (commented):
Thank you for the fix.
The towncrier file should not have the `.rst` extension.
Otherwise everything is good.

omnster:
Sorry, such a silly mistake.  It's fixed now.

Ericgig (approved):

Ericgig:
Thank you

PR #2694, opened 2025-05-15
Fix bug were args are not properly updated at coefficient creation.

### LINKED ISSUE #2646: Instantiating a qt.QobjEvo with a qt.coefficient

Ericgig:
**Description**
Fix bug were args are not properly updated at coefficient creation when initialised from another coefficient.
fix #2646

### DIFF
--- a/qutip/tests/core/test_coefficient.py
+++ b/qutip/tests/core/test_coefficient.py
@@ -161,6 +161,17 @@ def test_CoeffCallArguments(base, tol):
     _assert_eq_over_interval(coeff, expected, rtol=tol)
 
 
+def test_coefficient_update_args():
+
+    def f(t, a):
+        return a
+
+    coeff1 = coefficient(f, args={"a": 1})
+    coeff2 = coefficient(coeff1, args={"a": 2})
+
+    assert coeff2(0)
+
+
 @pytest.mark.parametrize(['style'], [
     pytest.param("func", id="func"),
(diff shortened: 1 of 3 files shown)


qutip/tests/core/test_coefficient.py
```diff
@@ -161,6 +161,17 @@ def test_CoeffCallArguments(base, tol):
...
 
+def test_coefficient_update_args():
+
+    def f(t, a):
+        return a
+
+    coeff1 = coefficient(f, args={"a": 1})
+    coeff2 = coefficient(coeff1, args={"a": 2})
+
+    assert coeff2(0)
```
pmenczel, line 172:
```suggestion
    assert coeff2(0) == 2
```

pmenczel (approved):

Issue #2691, opened 2025-05-12
partial trace and tensor products for energy-restricted space


magzpavz:
### Problem Description

The functionality of the excitation-number-restricted operators are fairly limited. 

### Proposed Solution

I have written code for myself to implement these operations. They are only for sparse matrices and are based on python loops. If there is interest in this feature/if I need it myself, I may convert my functions to cython. 

### Additional Context

One reason not to do this is because partial trace and tensor product are weird in ENR space and having this functionality may lead people astray. In particular, if you tensor product with identity, and then partial-trace the identity out, the resulting state is not proportional to the original state because the size of the state space of each mode depends on the other modes. 
However, if you're dealing with density matrices and with kets, this is less of an issue because you are not (necessarily) trying to represent an infinite-dimensional operator in this highly truncated space. This is my use case for these functions.

nwlambert:
Thanks @magzpavz yes I think we would be interested!  If its useful we have a preliminary partial trace function hidden in the end of this tutorial notebook for some years  
https://github.com/qutip/qutip-notebooks/blob/master/examples/excitation-number-restricted-states-jc-chain.ipynb
but haven't gotten around to checking, polishing and incorporating it.  

The tensor issue also sounds very useful.  If you decide to contribute, feel free to make a draft PR for both, and we can give some feedback.

PR #2696, opened 2025-05-16
partial trace and tensor product for energy-restricted space

magzpavz:
**Checklist**

- [x] Please read [Contributing to QuTiP Development](http://qutip.org/docs/latest/development/contributing.html)
- [x] Please add tests to cover your changes if applicable.
- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](http://qutip.org/docs/latest/development/contributing.html#changelog-generation) for more information).

**Description**
The purpose of these changes is to have partial trace and tensor support in ENR space. I have only implemented these for Qobj's with CSR backing. As part of this, I made functions to calculate the number of states in an ENR space and to convert between state labels and indexes. Previously, this could only be done by fully enumerating all the states. If the state space is large, this can be a substantial speed-up. 
I have set these functions (`enr_nstates`, `enr_state2idx`, and `enr_idx2state`) to be cached, which I'm not sure is ok as I do not see any other cached functions in qutip.

I tested a few different implementations of the tensor product by running them on randomly-generated operators. One uses my new cached index-calculating functions (this is the included version) and one used the existing `enr_state_dictionaries`. The included version is faster for operators with sparsity at or below 1%. However, it is terrible if the  indexing functions are not cached. 

I'm still not sure how to handle the fact that ENR space is weird. Right now I just explain this problem in the function description. 

**Related issues or PRs**
Issue #2691

### DIFF
--- a/qutip/core/energy_restricted.py
+++ b/qutip/core/energy_restricted.py
@@ -534,70 +534,65 @@ def enr_tensor(*args: Qobj, newexcitations: int = None,
 
     excitations = [(getattr(q._dims[0], 'n_excitations', None)
                     or q._dims[1].n_excitations) for q in args]
-    dimslists = [[tuple(q.dims[i]) for q in args] for i in range(2)]
-    # top level of list is row vs col
-    # second level is which of the input Qobj's
 
     # get the new max excitations
     newexcitations = newexcitations or min(excitations)
-    newdims = [tuple(itertools.chain(*dimsublist)) for dimsublist in dimslists]
+    newdims = [tuple(itertools.chain(*(q.dims[i] for q in args)))
+               for i in range(2)]
 
     trunccount = 0
-    currop = {}  # keep track as we build up the tensor product
+    out = {((), ()): 1.0}  # initialize a dict to tensor the first Qobj with
 
-    # initialize currop with the first operator
-    mat = args[0].data.as_scipy().tocoo()
-    for row_idx, col_idx, val in zip(mat.row, mat.col, mat.data):
-        row_state = tuple(args[0]._dims[0].idx2dims(row_idx))
-        col_state = tuple(args[0]._dims[1].idx2dims(col_idx))
-        if sum(row_state) > newexcitations or sum(col_state) > newexcitations:
-            if truncate:
-                trunccount += 1
-                continue
-            else:
-                missingstates = (str(state) for state in [row_state, col_state]
-                                 if sum(state) > newexcitations)
-                msg = (
-                    "state(s) " + ", ".join(missingstates) +
-                    " are not in the new restricted state space. " +
-                    "Set `truncate=True` to ignore these entries."
-                )
-                raise ValueError(msg) from None
-        currop[row_state, col_state] = val
-
-    # loop over the rest of the operators
-    for q in args[1:]:
-        oldop = currop
-        currop = {}
-        mat = q.data.as_scipy().tocoo()
-        for row_idx, col_idx, val in zip(mat.row, mat.col, mat.data):
-            row_state = tuple(q._dims[0].idx2dims(row_idx))
-            col_state = tuple(q._dims[1].idx2dims(col_idx))
-            for (row_state_old, col_state_old), oldval in oldop.items():
-                row_state_new = row_state_old + row_state
-                col_state_new = col_state_old + col_state
-                if (sum(row_state_new) > newexcitations
-                        or sum(col_state_new) > newexcitations):
-                    if truncate:
-                        trunccount += 1
-                        continue
-                    else:
-                        missingstates = (str(state) for state
-                                         in [row_state_new, col_state_new]
-                                         if sum(state) > newexcitations)
-                        msg = (
-                            "state " + ", ".join(missingstates) +
-                            " is not in the new restricted state space. " +
-                            "Set `truncate=True` to ignore these entries."
-                        )
-                        raise ValueError(msg) from None
-                currop[row_state_new, col_state_new] = oldval * val
+    # loop over the the operators
+    for q in args:
+        out, trunccount = _enr_tensor_qobj_with_dict(
+            out, q, newexcitations, truncate, trunccount)
 
     if trunccount > 0 and verbose:
         print(f"Truncated {trunccount} entries.")
 
-    return _enr_qobj_from_dict(currop, newdims, newexcitations,
+    return _enr_qobj_from_dict(out, newdims, newexcitations,
                                isherm=all(q.isherm for q in args))
+
+
+def _enr_tensor_qobj_with_dict(d: dict, q: Qobj, newexcitations: int,
+                               truncate, trunccount=0):
+    """
+    Helper function to do the tensor product between a dictionary (which
+    represents a Qobj) and a Qobj.
+    """
+    out = {}
+    mat = q.data.as_scipy().tocoo()
+
+    # loop over nonzero elements of the qobj
+    for row_idx, col_idx, val in zip(mat.row, mat.col, mat.data):
+        # convert the indices to the state tuples
+        row_state = tuple(q._dims[0].idx2dims(row_idx))
+        col_state = tuple(q._dims[1].idx2dims(col_idx))
+
+        # loop over the nonzero elements of the dictionary
+        for (row_state_old, col_state_old), oldval in d.items():
+            # concatenate the states
+            row_state_new = row_state_old + row_state
+            col_state_new = col_state_old + col_state
+
+            if (sum(row_state_new) <= newexcitations
+                    and sum(col_state_new) <= newexcitations):
+                out[row_state_new, col_state_new] = oldval * val
+            elif truncate:
+                trunccount += 1
+            else:
+                missingstates = (str(state) for state
+                                 in [row_state_new, col_state_new]
+                                 if sum(state) > newexcitations)
+                msg = (
+                    "state " + ", ".join(missingstates) +
+                    " is not in the new restricted state space. " +
+                    "Set `truncate=True` to truncate these entries."
+                )
+                raise ValueError(msg) from None
+
+    return (out, trunccount)
 
 
 def _enr_qobj_from_dict(op_dict, dims, excitations, **kwargs):
--- a/qutip/core/energy_restricted.py
+++ b/qutip/core/energy_restricted.py
@@ -596,28 +596,24 @@ def _enr_tensor_qobj_with_dict(d: dict, q: Qobj, newexcitations: int,
 
 
 def _enr_qobj_from_dict(op_dict, dims, excitations, **kwargs):
-    if np.prod(dims[0]) == 1:  # 1-dimensional (e.g. bra)
-        rowspace = Space(list(dims[0]))
-        nrows = 1
-        row_idxlist = [0]*len(op_dict)
-    else:
-        rowspace = EnrSpace(dims[0], excitations)
-        nrows = rowspace.size
-        row_idxlist = [rowspace.dims2idx(coord_state[0])
-                       for coord_state in op_dict.keys()]
-
-    if np.prod(dims[1]) == 1:  # 1-dimensional (e.g. ket)
-        colspace = Space(list(dims[1]))
-        ncols = 1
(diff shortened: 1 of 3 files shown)


pmenczel:
Hi and thank you for contributing, this is very welcome! Some preliminary comments:

* It surprises me that anything would be faster than a dictionary lookup. Especially with the cache, which in my mind is just another dictionary again... The creation of the dictionary may be slow, but it seems like the dictionaries are still there in your version, right?
* I don't think there are problems with having caching, but there should probably be a maximum cache size.
* I am wondering if functions like [ptrace](https://github.com/qutip/qutip/blob/master/qutip/core/qobj.py#L1104) and [tensor](https://github.com/qutip/qutip/blob/master/qutip/core/tensor.py#L42) should be modified to "magically" work with ENR states.

magzpavz:
- Depending on the size of the state space, generating the full state lookup dictionary can take a lot of time. If your density matrix or operator or ket is sparse (I tested 0.1%--10%), it may be wasteful to generate the whole dictionary. If you are working with identity or annihilation operators, you eventually need every element anyway, so it'd make sense to generate them at the beginning. However, if you're working with density matrices, you are less likely to span the whole space, so it's more efficient to only calculate `state2idx` and `idx2state` as needed. (It'd be nice to integrate the two caches so that when you generate the full dictionaries it also populates the function caches, but that does not seem possible with the built-in python caches). To be clear, my functions are slower if you're generating the entire dictionaries. 
- I agree on maximum cache size. However, the maximum cache size cannot be changed after function definition. Can we make a setting or something to let the user change the max cache size when they import qutip? That way we can set the default relatively low while still allowing people to work with larger spaces if needed.
- The most reliable way to see if a `Qobj` is ENR is to check the `_dims` property.

pmenczel:
> To be clear, my functions are slower if you're generating the entire dictionaries.

I see, okay, then it makes sense to me. So you would suggest to remove the dictionaries and replace them with your new functions everywhere?

> * I agree on maximum cache size. However, the maximum cache size cannot be changed after function definition. Can we make a setting or something to let the user change the max cache size when they import qutip? That way we can set the default relatively low while still allowing people to work with larger spaces if needed.

Since qutip doesn't have any persistent settings, and the module is loaded as soon as qutip is imported, this might be difficult. Perhaps someone else will chime in with their thoughts.

> * The most reliable way to see if a `Qobj` is ENR is to check the `_dims` property, so we could add checks for this. As I've written `enr_ptrace` and `enr_tensor`, they require the user to prove the `excitations` parameter, but this is also available in the `_dims` property, so I can remove that. I'd like to keep the other keyword arguments (`truncate`, `newexcitations`,`verbose`) for `enr_tensor`, so that would require adjusting `tensor` to accept keyword arguments.

I am not necessarily saying that we have to modify the existing `ptrace` and `tensor` functions, I just wanted to start a discussion about it. New users might expect these functions to work with ENR states just like they do normally... I wonder (a) can we put some default values for these other keyword arguments that will work in most cases, (b) since these functions have some surprising behaviors with ENR states, it might even be less confusing if users have to write `enr_tensor` and `enr_ptrace` explicitly. Perhaps @nwlambert has some thoughts on this?

nwlambert:
Thanks for the contribution, looks great.  From the discussion with paul, it sounds like your new functions are optimal for constructing states/density operators, so could be used in `enr_fock` and `enr_thermal_dm` while `enr_destroy` etcv should be left as is?  

For `ptrace`, `tensor`, vs `enr_ptrace`, `enr_tensor`, etc, I know that eric wanted a general push to make all of qutip function with enr states, so in the long term this is desirable.   But the existing functions also don't satisfy this, so perhaps its ok to keep them separate for them moment, and save integration for a later broader effort.   If at that point we need to include args in tensor and in other functions, etc, perhaps theres some smart general workaround (since many of these functions might be effected).    

I don't have any comment on the cache issues, perhaps @hodgestar  or @Ericgig have some comments on this and potential problems (all new to me).

the tensor function seems very useful.  the default for `newexcitations` is taken as the minimum of the composite systems and the default for truncation as false.  did you have a reason for this choice?  naively i am assuming this will then tend to throw errors, since it will end up truncating states from the larger system sometimes?  maybe i am just misunderstanding, didn't actually try yet!

nwlambert:
One other comment; we completely lack documentation on ENR states, apart from a couple of notebooks.  If you felt like contributing some here or in another PR it would be very welcome.  if not, no problem (its on the todo list!)

magzpavz:
Thanks for all the feedback. 

> So you would suggest to remove the dictionaries and replace them with your new functions everywhere?

>it sounds like your new functions are optimal for constructing states/density operators, so could be used in enr_fock and enr_thermal_dm while enr_destroy etcv should be left as is?

One thing I have realized since I first wrote these functions is that when you create an ENR `Qobj`, the dictionaries are always created when the `EnrSpace` is created, and the `Qobj`s are carrying them around. Therefore, I will rewrite my functions (and possibly the existing state and operator-creation functions) to use the dictionaries inside the `EnrSpace`s, which eliminates the need for my new idx<->state functions. I will leave them in the file.
Long-term (I really shouldn't spend time on this), I think it'd be best to get rid of the initialization of the dictionaries when unnecessary. See the first bullet point on [my previous comment](https://github.com/qutip/qutip/pull/2696#issuecomment-2891349036).

>the tensor function seems very useful. the default for newexcitations is taken as the minimum of the composite systems and the default for truncation as false. did you have a reason for this choice? naively i am assuming this will then tend to throw errors, since it will end up truncating states from the larger system sometimes? 

Yes, as currently written it will often throw errors. The reason for these defaults is that only when both `newexcitations` is the min of the composite systems and there is no truncation, is the tensor operation really "proper." If newexcitations is greater than the previous min, you may be expanding the state space to states previously truncated (consider tensoring together two space with dims [3,3] and max excitations 3. If the new max excitations is >3, the state (2,2,0,0) is in the new space but wasn't in the old). And truncation obviously removes information. I want to force the user to consider which of these options is less bad in their case. 

Re documentation: I could clean up and upload one of my research notebooks later. 
By the way, [this notebook](https://github.com/qutip/qutip-notebooks/blob/master/examples/excitation-number-restricted-states-jc-chain.ipynb) which was referred to me in my github issue does not work with qutip 5, which is the only one I have installed on my computer.

magzpavz:
I have made the change I mentioned in the previous comment to use the idx<->state dictionaries internal to the `Qobj`s. The ptrace and tensor operations are even faster now!

I have written tests and made a changelog file. I do not anticipate making any further changes to this PR.

nwlambert:
thanks @magzpavz, we will leave some final feedback/review after discussing at the admin meeting later this week

qutip/tests/test_enr_state_operator.py
```diff
@@ -197,3 +198,101 @@ def test_eigenstates_ENR():
...
+                         ])
+def test_enr_ptrace(dims, sel, n_excitations):
+    nstates_enr = _n_enr_states(dims, n_excitations)
+    # use qutip to make a random sparse Hermitian matrix w/ trace 1
+    rand_dm_mat = qutip.rand_dm(
+        nstates_enr, 0.05, distribution="herm", dtype="csr").data_as()
+
+    rho_in_enr = qutip.Qobj(rand_dm_mat,
+                            dims=[qutip.energy_restricted.EnrSpace(
+                                dims, n_excitations)]*2)
```
Ericgig, line 251: Why create a Qobj, extract the matrix and rebuild a new one?
You can update the dims of a Qobj:
`rand_dm_mat.dims = [qutip.energy_restricted.EnrSpace(dims, n_excitations)]*2`.

Ideally `qutip.rand_dm(EnrSpace(...), ...)` should work, we should fix...

qutip/tests/test_enr_state_operator.py
```diff
@@ -197,3 +198,101 @@ def test_eigenstates_ENR():
...
+def test_enr_tensor(dims_list, n_ex_list, isoper):
+    # isoper = 1 to test for operators, 0 to test for kets
+    nstates_enr_list = [_n_enr_states(dims, n_ex)
+                        for dims, n_ex in zip(dims_list, n_ex_list)]
+    dens_floor_list = 1/(np.array(nstates_enr_list)**(isoper + 1))
+    rand_mat_list = [scipy.sparse.random(
+        nstates, nstates*isoper + 1*(not isoper),
+        density=max(0.05, dens_floor), dtype="complex128")
+        for (nstates, dens_floor)
+        in zip(nstates_enr_list, dens_floor_list)]
```
Ericgig, line 285: style is particularly strange here.

qutip/tests/test_enr_state_operator.py
```diff
@@ -197,3 +198,101 @@ def test_eigenstates_ENR():
...
+
+    q_in_enr_list = [qutip.Qobj(
+        rand_mat,
+        dims=[qutip.energy_restricted.EnrSpace(dims, n_ex)]*(1 + isoper)
+        + [1]*(not isoper))
+        for rand_mat, dims, n_ex in zip(rand_mat_list, dims_list, n_ex_list)]
+    q_in_reg_list = [enr_to_reg(q_in_enr) for q_in_enr in q_in_enr_list]
+
+    q_out_enr = qutip.enr_tensor(*q_in_enr_list, newexcitations=sum(n_ex_list))
```
Ericgig, line 294: Are there tests where some spaces are truncated?

qutip/core/energy_restricted.py
```diff
@@ -46,6 +50,122 @@ def enr_state_dictionaries(dims, excitations):
...
+            for k in range(kmax+1))
+    else:  # in general, need to iterate over all subsets
+        return sum((-1)**k * math.comb(excitations + m - sum(subset), m)
+                   for k in range(kmax+1)
+                   for subset in itertools.combinations(dims, k)
+                   if sum(subset) < excitations+m)
+
+
+@lru_cache(maxsize=4096)
+def enr_state2idx(dims, excitations, state):
```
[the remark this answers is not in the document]
magzpavz: Since I refactored the ptrace and tensor functions to use the internal state dictionaries, we can remove this function (see comment under `enr_nstates`)

qutip/core/energy_restricted.py
```diff
@@ -288,3 +408,212 @@ def enr_identity(dims, excitations, *, dtype=None):
...
+    out : Qobj
+        A Qobj instance that represents the partially-traced `rho` in the
+        excitation-number restricted state. The maximum number of
+        excitations is the same as for the input.
+    """
+    if rho.shape[0] != rho.shape[1]:
+        raise ValueError(
+            "enr_ptrace is only defined for square density matrices")
+    if rho.dtype is not _data.CSR:
+        raise TypeError("only implemented for CSR dtype")
```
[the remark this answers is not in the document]
Ericgig: We use the DIA format a lot too, so just CSR is too limited.
I would suggest to add a warning if it's dense, but still work. We want stuff to just work whatever the format.

qutip/tests/test_enr_state_operator.py
```diff
@@ -197,3 +203,148 @@ def test_eigenstates_ENR():
...
+                             ([5, 2, 3], [2, 1]),
+                             ([5, 2, 3], [0, 2]),
+                             ([5, 2, 3], [0, 1]),
+                             ([2]*6, [3, 2]),
+                             ([2]*6, [0, 2]),
+                             ([2]*6, [0, 1]),
+                             ([4, 4, 4], []),
+                             ([4, 4, 4], [0, 1, 2]),
+                         ])
+def test_enr_ptrace(dims, sel, n_excitations, dtype):
```
Ericgig:
```suggestion
                         ])
@pytest.mark.filterwarnings("ignore:enr_ptrace")  
def test_enr_ptrace(dims, sel, n_excitations, dtype):
```

nwlambert:
thanks @magzpavz and @Ericgig, I gave a quick try with the old tutorial notebook, and both ptrace and tensor do what i expect. I still feel like the default tensor usage is a bit odd, but I see that maybe there is no great choice there, and the warnings seem to make sense.

PR #2700, opened 2025-05-26
Creating operator from dense ket.proj return CSR when appropriate

Ericgig:
**Description**
We made it so `|ket><bra|` return a sparse when appropriate, but `basis(10, 2).proj()` would still return a dense matrix...
This update `proj` to match the matmul behavior and add a test.

Ideally, this should be the dispatcher that can choose the appropriate output type so we don't have to do the same in qutip-jax etc.

### DIFF
--- a/qutip/core/data/matmul.pyx
+++ b/qutip/core/data/matmul.pyx
@@ -965,16 +965,15 @@ cpdef Dia matmul_outer_dia_dense_sparse(Data left, Data right, double complex sc
 
 cpdef Data matmul_outer_dense_Data(Dense left, Dense right, double complex scale=1):
     out_type = settings.core["default_dtype"]
-    if out_type is None:
-        out_density = (
-            dense.nnz(left) * 1.0 / left.shape[0]
-            * dense.nnz(right) * 1.0 / right.shape[1]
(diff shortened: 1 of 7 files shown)


qutip/core/data/matmul.pyx
```diff
@@ -965,16 +965,15 @@ cpdef Dia matmul_outer_dia_dense_sparse(Data left, Data right, double complex sc
...
-            out_type = CSR
-        else:
-            out_type = Dense
-    return matmul(left, right, dtype=out_type)
+    out_density = (
+        dense.nnz(left) * 1.0 / left.shape[0]
+        * dense.nnz(right) * 1.0 / right.shape[1]
+    )
+    if out_density < 0.3 and out_type is not Dense:
+        return matmul(left, right, dtype=CSR)
```
nwlambert, line 972: just curious here, would this change mean DIA is not returned even if detault_dtype is dia?

Ericgig: With `detault_dtype_scope="full"` it is dia, but not with the normal "creation" scope.

nwlambert (approved):

PR #2707, opened 2025-06-23
Fix(solver): Correct `ntraj` mismatch in parallel `mcsolve` with mixed states

basnijholt:
**Checklist**

- [ ] Please read [Contributing to QuTiP Development](http://qutip.org/docs/latest/development/contributing.html)

**Description**
Describe here the proposed change.

**Related issues or PRs**
Please mention the related issues or PRs here. If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id, e.g. fix #1184


basnijholt:
<details>

<summary>This code snippet shows the problem in more detail (click to see it):
</summary>

```python
import numpy as np
import bisect
from operator import itemgetter

def _minimum_roundoff_ensemble(state_list, ntraj_total):
    """
    This is the problematic algorithm extracted from qutip.
    It calculates a list of trajectory counts from a total number.
    """
    # First, it throws out zero-weight states.
    filtered_states = [(index, weight)
                       for index, (_, weight) in enumerate(state_list)
                       if weight > 0]
    if len(filtered_states) > ntraj_total:
        raise ValueError(f'{ntraj_total} trajectories is not enough for '
                         f'initial mixture of {len(filtered_states)} '
                         'states')

    # It makes a first guess for each state's trajectory count by taking
    # the ceiling of the weight times the total number of trajectories.
    one_traj_states = []
    under_consideration = []
    current_total = 0
    for index, weight in filtered_states:
        guess = int(np.ceil(weight * ntraj_total))
        current_total += guess
        if guess == 1:
            one_traj_states.append(index)
        else:
            ratio = guess / (weight * ntraj_total)
            bisect.insort(under_consideration,
                          (index, weight, guess, ratio),
                          key=itemgetter(3))

    # CRITICAL FLAW: This loop only runs if current_total > ntraj_total.
    # If the sum of the guesses is *less* than the requested total,
    # this loop is skipped, and the incorrect, smaller sum is returned.
    while current_total > ntraj_total:
        index, weight, guess, ratio = under_consideration.pop()
        guess -= 1
        current_total -= 1
        if guess == 1:
            one_traj_states.append(index)
        else:
            ratio = guess / (weight * ntraj_total)
            bisect.insort(under_consideration,
                              (index, weight, guess, ratio),
                              key=itemgetter(3))

    # Finally, it arranges the results in a list.
    ntraj = [0] * len(state_list)
    for index in one_traj_states:
        ntraj[index] = 1
    for index, _, count, _ in under_consideration:
        ntraj[index] = count
    return ntraj

# --- Demonstration of the Failure ---

# 1. Define a list of states and weights that sum to 1.0 - 1e-5.
states = [
    ("state_0", 0.5),
    ("state_1", 0.5 - 1e-5),
]

# 2. Request a total number of trajectories.
ntraj_requested = 100000

print(f"Weights sum to: {sum(w for _, w in states)}")
print(f"Requested total trajectories: {ntraj_requested}")
print("-" * 40)

# 3. Run the algorithm.
calculated_ntraj_list = _minimum_roundoff_ensemble(states, ntraj_requested)
actual_total = sum(calculated_ntraj_list)

# 4. Explain the intermediate calculation.
guess1 = int(np.ceil(states[0][1] * ntraj_requested))
guess2 = int(np.ceil(states[1][1] * ntraj_requested))
print(f"Intermediate guess for state 0: ceil({states[0][1]} * {ntraj_requested}) = {guess1}")
print(f"Intermediate guess for state 1: ceil({states[1][1]} * {ntraj_requested}) = {guess2}")
print(f"Sum of initial guesses: {guess1 + guess2}")
print("-" * 40)


print(f"Calculated trajectory list: {calculated_ntraj_list}")
print(f"Actual sum of trajectories: {actual_total}")
print("-" * 40)

# 5. Show that the result is incorrect.
try:
    assert ntraj_requested == actual_total
    print("✅ The algorithm worked correctly.")
except AssertionError:
    print(f"❌ FAILURE: The algorithm returned a list that sums to {actual_total}, "
          f"not the requested {ntraj_requested}.")
```

</details>

Which prints:

```python
Weights sum to: 0.9999899999999999
Requested total trajectories: 100000
----------------------------------------
Intermediate guess for state 0: ceil(0.5 * 100000) = 50000
Intermediate guess for state 1: ceil(0.49999 * 100000) = 49999
Sum of initial guesses: 99999
----------------------------------------
Calculated trajectory list: [50000, 49999]
Actual sum of trajectories: 99999
----------------------------------------
❌ FAILURE: The algorithm returned a list that sums to 99999, not the requested 100000.
```

pmenczel:
Hi @basnijholt, thank you for bringing this to our attention and thank you for the provided fix.

In principle, we are requiring the weights provided to mcsolve to sum to one. Now in practice, they might not sum exactly to one due to floating point arithmetic. However, if the deviation from one is of the order of the machine precision, I would have thought it very unlikely to run into this problem. So I was wondering, did you encounter this problem in practice?

I would like to keep the actual number of trajectories that is used to be the same that the user requested. I am wondering if it would be sufficient to normalize the weights that are provided such that they do sum to one (up to machine precision. What do you think about that?

basnijholt:
Thanks for the quick reply @pmenczel!

Yes, this error occurred in a practical, long-running open quantum system simulation. The simulation is iterative, and at each step, the Hilbert space is truncated to keep the problem computationally tractable. This process of repeatedly taking partial traces and reconstructing the state can cause small numerical errors to accumulate, leading to a state whose trace deviates from 1.0 by more than machine precision.

While normalizing the state at each step is a possible workaround, the proposed fix makes the solver itself more robust. It ensures the solver uses the *actual* number of trajectories calculated by `_InitialConditions` as the single source of truth, preventing a crash even if the initial state isn't perfectly normalized due to these accumulated floating-point errors.

This makes `qutip` more resilient and easier to use for complex, iterative simulations where perfect normalization at every step can be difficult to maintain.

pmenczel:
> The simulation is iterative, and at each step, the Hilbert space is truncated to keep the problem computationally tractable. This process of repeatedly taking partial traces and reconstructing the state can cause small numerical errors to accumulate, leading to a state whose trace deviates from 1.0 by more than machine precision.

Cool, sounds you are doing something quite interesting!

My point of view is this: you are basically suggesting to support mixed initial conditions with unnormalized weights. Your solution seems okay to me if the sum of the weights is close to one, but not larger than one. However, if we remove the requirement that they sum to one, we probably should support arbitrary (positive) weights.

With your implementation, if the sum of weights is much smaller than one, the number of actually used trajectories will be very different from the number requested by the user. I do still think that the number of trajectories that are run should always be the requested number.  
In other words, the single source of truth should be the user-provided `ntraj`. If `_InitialConditions` generates a different number of trajectories, then that is a problem with `_InitialConditions`.

The most straightforward way to solve the problem might be to normalize the weights in the constructor of `_InitialConditions`. The original norm (sum of weights) would need to be saved and enter the correction weight in `get_state_and_weight`.

In any case, there would be at least one more thing to think about: mcsolve results can be [merged](https://github.com/qutip/qutip/blob/master/qutip/solver/multitrajresult.py#L698) and we would need to check that that doesn't break.

pmenczel:
I will close for now due to inactivity, but I raised the issue in #2880

PR #2708, opened 2025-06-23
Quantum System Library: First Few Models and Tests

Vanshaj0429:
## Quantum Systems Library Setup – Jaynes–Cummings Model Coming Next

This PR sets up the foundational structure for the Hamiltonian Library as part of the `dev.qsystem` branch in QuTiP.

### Summary of Additions
- [x] Created new module directory: `qutip/qutip/quantum_systems/`
- [x] Added `__init__.py` to initialize the module

### Next Steps (Planned in this branch)
- [x] Introduce the **Jaynes–Cummings model** as the first implemented system
- [x] Define model structure with support for LaTeX rendering 
- [x] Add minimal usage examples and unit tests

### Motivation

The goal of the Hamiltonian Library is to offer reusable, standardised implementations of commonly used quantum models, starting with cavity QED and spin chain systems.

The **Jaynes–Cummings model** is a natural starting point due to its importance in cavity QED and quantum optics.

### Feedback & Suggestions

Would appreciate thoughts on:
- Structural layout of the module
- Naming conventions and directory organisation
- Preferred implementation style (class-based vs functional)


Vanshaj0429:
## Adding quantum systems module

Hey everyone! This PR adds a new quantum systems module that I've been working on. The goal is to make it easier for users to create common quantum systems without having to build everything from scratch each time.

### What's included:

**Base class and qubit system:**
- Added a `QuantumSystem` base class that standardizes how we handle quantum systems (parameters, operators, pretty printing, etc.)
- Implemented a `qubit()` function for two-level systems with decay and dephasing
- Pretty straightforward but covers the most common use case

**Jaynes-Cummings model:**
- This one was more involved - implements cavity-atom interactions
- Supports both rotating wave approximation and the full Rabi model
- Includes various dissipation mechanisms (cavity decay, atomic decay, dephasing, thermal effects)
- Should handle most cavity QED scenarios people actually use

**Tests:**
- Added test files that are thorough but still readable
- Tests cover both the API and the actual physics (eigenvalues, commutation relations, etc.)
- Getting close to 100% coverage which is nice

I tried to keep the interface simple while still being flexible. Both functions return `QuantumSystem` objects so they work consistently, and users can still access all the underlying operators and Hamiltonians if they need to do custom stuff.

The main use case is letting people do things like:
```python
q = qubit(omega=2.0, decay_rate=0.1)
jc = jaynes_cummings(omega_c=1.0, omega_a=1.0, g=0.1, n_cavity=10)
```

Vanshaj0429:
@Ericgig, @nwlambert and @ajgpitch can you have a look at these commits, I think there are a few minor things that are left, apart from that looks fine to me.

Right now the `.pretty_print()` is not rendering latex, I am trying to figure out how this can be done.
 
Screenshot 2025-07-28 at 5 48 22 AM

I am getting pretty good coverage on the tests locally (using pytest) as well.

Can you have a look/review these commits.

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,80 @@
...
+        """Get eigenvalues and eigenstates"""
+        return self.hamiltonian.eigenstates()
+
+    @property
+    def ground_state(self) -> qt.Qobj:
+        """Get ground state"""
+        _, states = self.eigenstates
+        return states[0]
+
+    def pretty_print(self):
```
Ericgig, line 63: Look into `_repr_latex_`, it's a special method for ipython.

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,80 @@
...
+        return self.c_ops
+
+    def get_latex(self) -> str:
+        """Get LaTeX representation"""
+        return self.latex
+
+    @property
+    def dimension(self) -> int:
+        """Get Hilbert space dimension"""
+        return self.hamiltonian.shape[0] if self.hamiltonian else 0
```
Ericgig: It should return `self.hamiltonian.dims` or `self.hamiltonian._dims`

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+import qutip as qt
+from typing import Dict, List
```
Ericgig: The module typing is getting deprecated. You can use `dict` and `list` directly for type hints.

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class QuantumSystem:
+    """
+    General class for quantum systems
+
+    All quantum systems are instances of this class, configured by factory functions.
+    """
+
+    def __init__(self, name: str, **kwargs):
```
Ericgig: Why is only name and parameters used here?

The class should be usable after the `__init__`.
But presently, you could have a QuantumSystem without hamiltonian, totally unusable.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,192 @@
...
+    system.operators = operators
+    system.hamiltonian = hamiltonian
+    system.c_ops = c_ops
+    system.latex = latex
+
+    return system
+
+
+# Example usage and demonstrations
+if __name__ == "__main__":
```
Ericgig: This will have to be move to a notebook or documentation page .

qutip/quantum_systems/tests/test_qubit.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def test_default_qubit(self):
+        """Test qubit with default parameters"""
+        q = qubit()
+
+        assert isinstance(q, QuantumSystem)
+        assert q.name == "Qubit"
+        assert q.parameters["omega"] == 1.0
+        assert q.parameters["decay_rate"] == 0.0
+        assert q.parameters["dephasing_rate"] == 0.0
```
Ericgig: Is there a reason these default value are the right ones?
I would suggest to  test for the existence:
`assert "omega" in q.parameters`

qutip/quantum_systems/tests/test_qubit.py
```diff
@@ -0,0 +1,234 @@
...
+        assert q.dimension == 2
```
Ericgig: Qutip dimensions show the hilbert space. This is the shape.

qutip/quantum_systems/tests/test_qubit.py
```diff
@@ -0,0 +1,234 @@
...
+    @pytest.mark.parametrize("omega", [0.5, 1.0, 2.0, -1.0, 0.0])
+    def test_omega_parameter(self, omega):
+        """Test qubit with different omega values"""
+        q = qubit(omega=omega)
+
+        assert q.parameters["omega"] == omega
+
+        # Check Hamiltonian scaling
+        expected_H = 0.5 * omega * qt.sigmaz()
+        assert (q.hamiltonian - expected_H).norm() < 1e-10
```
Ericgig: `==` work for Qobj.

qutip/quantum_systems/tests/test_qubit.py
```diff
@@ -0,0 +1,234 @@
...
+    def test_pauli_properties(self, op_name):
+        """Test Pauli operator properties"""
+        q = qubit()
+        op = q.operators[op_name]
+
+        # Should be Hermitian
+        assert (op - op.dag()).norm() < 1e-10
+
+        # Should square to identity
+        assert (op * op - qt.qeye(2)).norm() < 1e-10
```
Ericgig: This feels like your testing the operators, not the no class...

qutip/quantum_systems/tests/test_qubit.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def test_edge_cases(self):
+        """Test edge cases"""
+        # Very small parameters
+        q_small = qubit(omega=1e-10, decay_rate=1e-10)
+        assert q_small.dimension == 2
+
+        # Large parameters
+        q_large = qubit(omega=1e10, decay_rate=1e5)
+        assert q_large.dimension == 2
```
Ericgig: How is that testing edge cases?

qutip/quantum_systems/tests/test_quantum_system.py
```diff
@@ -0,0 +1,298 @@
...
+    @pytest.mark.parametrize("name,params", [
+        ("Qubit", {"omega": 1.0}),
+        ("Oscillator", {"omega": 1.0, "n_levels": 10}),
+        ("Complex", {"freq1": 1.0, "freq2": 1.5, "coupling": 0.1}),
+    ])
+    def test_various_initializations(self, name, params):
+        """Test various initialization scenarios"""
+        system = QuantumSystem(name, **params)
+        assert system.name == name
+        assert system.parameters == params
```
Ericgig: Aren't these tests redundant?

qutip/quantum_systems/tests/test_quantum_system.py
```diff
@@ -0,0 +1,298 @@
...
+            assert (H - H.dag()).norm() < 1e-10
+
+            # Eigenvalues should be real
+            eigenvals = system.eigenvalues
+            assert all(np.isreal(eigenvals))
+
+            # Ground state should have lowest energy
+            ground_state = system.ground_state
+            ground_energy = qt.expect(H, ground_state)
+            assert abs(ground_energy - min(eigenvals)) < 1e-10
```
Ericgig: This does not test `QuantumSystem`...
H being hermitian or having real eigenvalues does not depend on any code written in `QuantumSystem`.
If you set H as `destroy(N)`, this would fail but not show any bug in `QuantumSystem`...

Ericgig (commented):
Are all tests needed / useful? It seems there is some quantity over quality going on.

Test files should be located with other tests files: `qutip/tests/...`. Creating a new directory for them would be good.

The `.coverage` file should not be included.

ajgpitch:
Is the spin chain code not in this PR? I am looking to add some comments to the code you shared in the blog.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,192 @@
...
+    rabi = jaynes_cummings(
+        omega_c=1.0,
+        omega_a=1.0,
+        g=0.2,
+        rotating_wave=False
+    )
+    print(f"LaTeX: {rabi.latex}")
+
+    # Example 4: Accessing operators and Hamiltonian
+    print("\n4. Accessing System Components:")
```
nwlambert: when this is finished/moved to a test or notebook (or docs), would be nice to show how to feed these too mesolve. hopefully at that point send the whole quantumsystem will work too of course

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @property
+    def eigenvalues(self) -> np.ndarray:
+        """Get eigenvalues of the Hamiltonian"""
+        return self.hamiltonian.eigenenergies()
+
+    @property
+    def eigenstates(self) -> tuple:
+        """Get eigenvalues and eigenstates"""
+        return self.hamiltonian.eigenstates()
```
nwlambert, line 55: just to keep track, i think eric suggested cacheing or storing these in a property?

qutip/quantum_systems/quantum_system.py
Vanshaj0429: I was able to get this done using `IPython.display` and  importing  `display`, `Markdown`, `Latex` from it.

qutip/quantum_systems/quantum_system.py
Vanshaj0429: Yes made the correction

qutip/quantum_systems/quantum_system.py
Vanshaj0429: Made the correction

qutip/quantum_systems/quantum_system.py
Vanshaj0429: We can include everything not just the hamiltonian, so this would change to something like:

```
def __init__(self, name: str, hamiltonian: qt.Qobj = None, 
             operators: dict = None, c_ops: list = None, 
             latex: str = "", **kwargs):
    """
    Initialize quantum system

    Parameters:
    -----------
    name : str
        Name/type of the quantum system
    hamiltonian : qt.Qobj, optional
        System Hamiltonian
    operators : dict, optional
        Dictionary of system operators
    c_ops : list, optional
        List of collapse operators
    latex : str, optional
        LaTeX representation of the system
    **kwargs : dict
        Additional system-specific parameters
    """
    self.name = name
    self.parameters = kwargs

    # Set system components with defaults
    self.hamiltonian = hamiltonian
    self.operators = operators if operators is not None else {}
    self.c_ops = c_ops if c_ops is not None else []
    self.latex = latex
```

And the model function would change something like:
```
def qubit(omega: float = 1.0, decay_rate: float = 0.0, 
          dephasing_rate: float = 0.0) -> QuantumSystem:
    """Create two-level system (qubit)"""
    
    # Build operators 
    operators = {
        'sigma_minus': qt.destroy(2),
        'sigma_plus': qt.create(2),
        'sigma_z': qt.sigmaz(),
        'sigma_x': qt.sigmax(),
        'sigma_y': qt.sigmay()
    }
    
    # Build Hamiltonian 
    hamiltonian = 0.5 * omega * operators['sigma_z']
    
    # Build collapse operators 
    c_ops = []
    if decay_rate > 0.0:
        c_ops.append(np.sqrt(decay_rate) * operators['sigma_minus'])
    if dephasing_rate > 0.0:
        c_ops.append(np.sqrt(dephasing_rate) * operators['sigma_z'])
    
    # LaTeX representation
    latex = r"H = \frac{\omega}{2}\sigma_z"
    
    # Create system with all components
    return QuantumSystem(
        name="Qubit",
        hamiltonian=hamiltonian,
        operators=operators,
        c_ops=c_ops,
        latex=latex,
        omega=omega,
        decay_rate=decay_rate,
        dephasing_rate=dephasing_rate
    )
```
Instead of how we do it now:
```
    # Create system
    system = QuantumSystem(
        "Qubit",
        omega=omega, decay_rate=decay_rate, dephasing_rate=dephasing_rate)
    ...
    ...
    ...
    ...
    
    # Set everything in the system
    system.operators = operators
    system.hamiltonian = hamiltonian
    system.c_ops = c_ops
    system.latex = latex
    
    return system
```

Does this make sense @Ericgig, @nwlambert??

Vanshaj0429:
I merged the `master` to `dev.qsystem`, it sort of brought all the other commits into the history. I read online that it is expected. Can you confirm?

Also I have pushed the code changes, have a look at the last three commits (https://github.com/qutip/qutip/pull/2708/commits/b402f1a3ac40830b38075be1ec881fb114d4bb9a, https://github.com/qutip/qutip/pull/2708/commits/0a9a6e430b75372653727684993995ac99687c9b) and could you please review it? @ajgpitch, @Ericgig and @nwlambert

Ericgig:
@Vanshaj0429 
Yes it's normal that it brought all the commits. What we wanted is to bring the bug fixes commits.
But yes, the diff  shows too much stuff. 
To clean that, I merged master into dev.qsystem, now if you merge `dev.qsystem` (from qutip) into yours, it should clean the diff and make the review easy.

Vanshaj0429:
> @Vanshaj0429 Yes it's normal that it brought all the commits. What we wanted is to bring the bug fixes commits. But yes, the diff shows too much stuff. To clean that, I merged master into dev.qsystem, now if you merge `dev.qsystem` (from qutip) into yours, it should clean the diff and make the review easy.

So I can simply do `git fetch upstream` and then `git pull upstream dev.qsystem`? OR do 

```bash
git fetch upstream
git merge --no-ff upstream/dev.qsystem
git push origin dev.qsystem
```
@Ericgig

Ericgig:
Both should work.
But you don't need the `-no-ff`.

Vanshaj0429:
I have fixed the failing tests. Can you run it again to see if it passes now? @Ericgig

qutip/__init__.py
```diff
@@ -34,6 +34,7 @@
 from .solver import *
 from .solver import nonmarkov
 import qutip.piqs.piqs as piqs
+from .quantum_systems import *
```
Ericgig: Let's have them in their namespace:
```suggestion
import quantum_systems
```

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,131 @@
...
+        omega_c: float = 1.0,
+        omega_a: float = 1.0,
+        g: float = 0.1,
+        n_cavity: int = 10,
+        rotating_wave: bool = True,
+        cavity_decay: float = 0.0,
+        atomic_decay: float = 0.0,
+        atomic_dephasing: float = 0.0,
+        thermal_photons: float = 0.0) -> QuantumSystem:
+    """
```
Ericgig: We use that format more often:
```suggestion
def jaynes_cummings(
    omega_c: float = 1.0,
    omega_a: float = 1.0,
    g: float = 0.1,
    n_cavity: int = 10,
    rotating_wave: bool = True,
    cavity_decay: float = 0.0,
    atomic_decay: float = 0.0,
    atomic_dephasing: float = 0.0,
    thermal_photons: float = 0.0,
) -> QuantumSystem:
    """
```

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,102 @@
...
+        """Get Hamiltonian"""
+        return self.hamiltonian
+
+    def get_c_ops(self) -> list[qt.Qobj]:
+        """Get collapse operators"""
+        return self.c_ops
+
+    def get_latex(self) -> str:
+        """Get LaTeX representation"""
+        return self.latex
```
Ericgig: Let's remove these.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,131 @@
+import numpy as np
+import qutip as qt
```
Ericgig: We don't use the shorten `qt` internally:
```suggestion
from qutip import tensor, qeye, destroy, sigmax, sigmay, sigmaz
```

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,102 @@
+import numpy as np
+import qutip as qt
```
Ericgig:
```suggestion
from qutip import Qobj
```

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,102 @@
...
+    def get_c_ops(self) -> list[qt.Qobj]:
+        """Get collapse operators"""
+        return self.c_ops
+
+    def get_latex(self) -> str:
+        """Get LaTeX representation"""
+        return self.latex
+
+    @property
+    def dimension(self) -> int:
```
Ericgig: `dimension` is  not an `int`, but a `list of list of int` or `list of list of list of int`. Confusing I know...

qutip/quantum_systems/quantum_system.py
```diff
@@ -0,0 +1,102 @@
...
+class QuantumSystem:
+    """
+    General class for quantum systems
+
+    All quantum systems are instances of this class, configured by factory functions.
+    """
+
+    def __init__(self, name: str, hamiltonian: qt.Qobj = None,
+                 operators: dict = None, c_ops: list = None,
+                 latex: str = "", **kwargs):
```
Ericgig: Many features don't work without the hamiltonian. Let make it required.
On the other side, the name could be made optional.

qutip/quantum_systems/qubit.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+import qutip as qt
```
Ericgig: Same

qutip/quantum_systems/qubit.py
```diff
@@ -0,0 +1,57 @@
...
+from .quantum_system import QuantumSystem  # Import the QuantumSystem class
```
Ericgig:
```suggestion
from .quantum_system import QuantumSystem
```

qutip/quantum_systems/qubit.py
```diff
@@ -0,0 +1,57 @@
...
+    H = (omega_a/2) * sigma_z
+
+    Parameters:
+    -----------
+    omega : float
+        Transition frequency
+    decay_rate : float
+        Relaxation rate (1/T1)
+    dephasing_rate : float
+        Dephasing rate (1/T2)
```
Ericgig: Let's add the default in the description.

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
+import pytest
+import numpy as np
+import qutip as qt
```
Ericgig:
```suggestion
import qutip
```

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+from qutip.quantum_systems.quantum_system import QuantumSystem
```
Ericgig:
```suggestion
from qutip.quantum_systems import QuantumSystem
```

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+        (qt.sigmax(), [[2], [2]]),
+        (qt.qeye(3), [[3], [3]]),
+        (qt.qeye(5), [[5], [5]]),
+        (qt.destroy(4), [[4], [4]]),
+        (qt.tensor(qt.sigmaz(), qt.sigmaz()), [[2, 2], [2, 2]]),
+    ])
+    def test_dimension_property(self, hamiltonian, expected_dim):
+        """Test dimension property with various Hamiltonians"""
+        system = QuantumSystem("Test")
+        system.hamiltonian = hamiltonian
```
Ericgig:
```suggestion
        system = QuantumSystem("Test", hamiltonian)
```

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+        assert system.dimension == expected_dim
+
+    def test_eigenvalues_qubit(self):
+        """Test eigenvalues for qubit systems"""
+        system = QuantumSystem("Test")
+        system.hamiltonian = 0.5 * qt.sigmaz()
```
Ericgig:
```suggestion
        system = QuantumSystem("Test", 0.5 * qt.sigmaz())
```

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+    def test_eigenvalues_harmonic_oscillator(self):
+        """Test eigenvalues for harmonic oscillator"""
+        system = QuantumSystem("Test")
+        n_levels = 5
+        system.hamiltonian = qt.num(n_levels) + 0.5 * qt.qeye(n_levels)
+
+        eigenvals = system.eigenvalues
+        expected = np.arange(n_levels) + 0.5
+        np.testing.assert_array_almost_equal(
+            sorted(eigenvals), sorted(expected))
```
Ericgig: `expected` is already sorted. The Hamiltonian eigenvalues should also be sorted if it's hermitian.
```suggestion
        np.testing.assert_array_almost_equal(eigenvals, expected)
```

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+        system = QuantumSystem("Test")
+
+        # Add operators
+        system.operators["sigma_x"] = qt.sigmax()
+        system.operators["sigma_z"] = qt.sigmaz()
+
+        assert "sigma_x" in system.operators
+        assert "sigma_z" in system.operators
+        assert system.operators["sigma_x"] == qt.sigmax()
+        assert system.operators["sigma_z"] == qt.sigmaz()
```
Ericgig: This is testing if a dict act as a dict...
Is that really needed.

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Add collapse operators
+        c_op1 = qt.destroy(2)
+        c_op2 = qt.sigmaz()
+        system.c_ops.append(c_op1)
+        system.c_ops.append(c_op2)
+
+        assert len(system.c_ops) == 2
+        assert system.c_ops[0] == c_op1
+        assert system.c_ops[1] == c_op2
```
Ericgig: This is testing that a list is working as a list...

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def test_latex_management(self):
+        """Test LaTeX string management"""
+        system = QuantumSystem("Test")
+
+        latex_str = r"H = \omega \sigma_z"
+        system.latex = latex_str
+
+        assert system.get_latex() == latex_str
+        assert system.latex == latex_str
```
Ericgig:
```
a = b
assert a == b
```
We don't need to test that...

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+        """Test parameter modification"""
+        system = QuantumSystem("Test", omega=1.0)
+
+        # Modify existing parameter
+        system.parameters["omega"] = 2.0
+        assert system.parameters["omega"] == 2.0
+
+        # Add new parameter
+        system.parameters["gamma"] = 0.1
+        assert system.parameters["gamma"] == 0.1
```
Ericgig: This test that python dict work as a dict...
You could argue for: `assert isinstance(system.parameters, dict)`, but don't test python language...

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def test_edge_cases(self):
+        """Test edge cases"""
+        # Large dimension system
+        system_large = QuantumSystem("Large")
+        system_large.hamiltonian = qt.qeye(50)
+        assert system_large.dimension == [[50], [50]]
```
Ericgig: How is that an edge case?
There is no branching on size, so no reason it don't work.
Except if it's too large to fit in memory. But then the Qobj creation will fail, not `QuantumSystem`.

ps. We are doing simulations of open system with Hamiltonian of size [32768 x 32768].

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Zero Hamiltonian
+        system_zero = QuantumSystem("Zero")
+        system_zero.hamiltonian = qt.qzero(3)
+        eigenvals = system_zero.eigenvalues
+        assert all(abs(e) < 1e-10 for e in eigenvals)
```
Ericgig: Also, how is that an edge cases?

qutip/tests/quantum_systems/test_quantum_system.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Identity Hamiltonian
+        system_identity = QuantumSystem("Identity")
```
Ericgig: Same

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
+import pytest
+import numpy as np
+import qutip as qt
```
Ericgig:
```suggestion
import qutip
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
...
+    def test_atomic_operators(self):
+        """Test atomic operator properties"""
+        jc = jaynes_cummings(n_cavity=3)
+
+        sigma_plus = jc.operators["sigma_plus"]
+        sigma_minus = jc.operators["sigma_minus"]
+
+        assert sigma_plus == sigma_minus.dag()
+
+    @pytest.mark.parametrize("omega_c,omega_a,g", [
```
Ericgig:
```suggestion
    @pytest.mark.parametrize("omega_c, omega_a, g", [
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
...
+        sigma_minus = jc.operators["sigma_minus"]
+        sigma_z = jc.operators["sigma_z"]
+
+        H_expected = (omega_c * a_dag * a +
+                      0.5 * omega_a * sigma_z +
+                      g * (a_dag + a) * (sigma_plus + sigma_minus))
+
+        assert jc.hamiltonian == H_expected
+
+    @pytest.mark.parametrize("decay_type,rate,operator_name", [
```
Ericgig:
```suggestion
    @pytest.mark.parametrize("decay_type, rate, operator_name", [
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
...
+            n_cavity=3)
+
+        # Should have both relaxation and excitation operators
+        cavity_relax_rate = cavity_decay * (1 + thermal_photons)
+        expected_relax_op = np.sqrt(cavity_relax_rate) * jc.operators["a"]
+        assert expected_relax_op in jc.c_ops
+
+        cavity_excite_rate = cavity_decay * thermal_photons
+        expected_excite_op = np.sqrt(
+            cavity_excite_rate) * jc.operators["a_dag"]
```
Ericgig:
```suggestion
        expected_excite_op = cavity_excite_rate**0.5 * jc.operators["a_dag"]
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
...
+
+        assert jc.latex == expected_latex
+
+    def test_ground_state(self):
+        """Test ground state properties"""
+        jc = jaynes_cummings(omega_c=1.0, omega_a=1.0, g=0.1, n_cavity=3)
+        ground_state = jc.ground_state
+
+        # Ground state should be normalized
+        assert abs(ground_state.norm() - 1.0) < 1e-10
```
Ericgig: Pytest has something for these cases:
```suggestion
        assert ground_state.norm() == pytest.approx(1., abs=1e-10)
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
...
+
+        n_c = jc.operators["n_c"]
+
+        # Test expectation values on basis states
+        for n in range(min(3, n_cavity)):
+            for atom_state in [0, 1]:
+                state = qt.tensor(
+                    qt.basis(
+                        n_cavity, n), qt.basis(
+                        2, atom_state))
```
Ericgig:
```suggestion
                state = qutip.basis([n_cavity, 2], [n, atom_state])
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,291 @@
...
+        required_attributes = [
+            'name',
+            'operators',
+            'hamiltonian',
+            'c_ops',
+            'latex',
+            'parameters']
+        for attr in required_attributes:
+            assert hasattr(jc, attr)
+
```
Ericgig: Remove, redundant to the previous `assert isinstance(jc, QuantumSystem)`.

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+    def test_default_qubit(self):
+        """Test qubit with default parameters"""
+        # Random values for parameters
+        omega_val = 0.5
+        decay_val = 0.1
+        dephasing_val = 0.05
+        q = qubit(
+            omega=omega_val,
+            decay_rate=decay_val,
+            dephasing_rate=dephasing_val)
```
Ericgig:
```suggestion
        q = qubit(
            omega=omega_val,
            decay_rate=decay_val,
            dephasing_rate=dephasing_val
        )
```

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Check Hamiltonian scaling
+        expected_H = 0.5 * omega * qt.sigmaz()
+        assert q.hamiltonian == expected_H
+
+        # Check eigenvalues
+        eigenvals = q.eigenvalues
+        expected_eigenvals = np.array([-omega / 2, omega / 2])
+        np.testing.assert_array_almost_equal(
+            sorted(eigenvals), sorted(expected_eigenvals))
```
Ericgig:
```suggestion
        np.testing.assert_array_almost_equal(eigenvals, expected_eigenvals)
```

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+        assert q.parameters["dephasing_rate"] == dephasing_rate
+
+        if dephasing_rate == 0.0:
+            assert len(q.c_ops) == 0
+        else:
+            assert len(q.c_ops) == 1
+            expected_dephasing_op = np.sqrt(dephasing_rate) * qt.sigmaz()
+            assert expected_dephasing_op in q.c_ops
+
+    @pytest.mark.parametrize("omega,decay_rate,dephasing_rate", [
```
Ericgig:
```suggestion
    @pytest.mark.parametrize("omega, decay_rate, dephasing_rate", [
```

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+        (1.0, 0.0, 0.1),    # Dephasing only
+        (1.0, 0.1, 0.1),    # Both mechanisms
+        (2.0, 0.2, 0.3),    # Different rates
+    ])
+    def test_combined_parameters(self, omega, decay_rate, dephasing_rate):
+        """Test qubit with combined parameter variations"""
+        q = qubit(
+            omega=omega,
+            decay_rate=decay_rate,
+            dephasing_rate=dephasing_rate)
```
Ericgig:
```suggestion
        q = qubit(
            omega=omega,
            decay_rate=decay_rate,
            dephasing_rate=dephasing_rate
        )
```

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Check collapse operators count
+        expected_c_ops_count = 0
+        if decay_rate > 0:
+            expected_c_ops_count += 1
+        if dephasing_rate > 0:
+            expected_c_ops_count += 1
+        assert len(q.c_ops) == expected_c_ops_count
+
+    def test_operators_completeness(self):
```
Ericgig: This is redundant to the one just under, let's remove.

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+        ("sigma_minus", qt.destroy(2)),
+        ("sigma_plus", qt.create(2)),
+        ("sigma_z", qt.sigmaz()),
+        ("sigma_x", qt.sigmax()),
+        ("sigma_y", qt.sigmay()),
+    ])
+    def test_operators_correctness(self, operator_name, expected_operator):
+        """Test that operators are correct"""
+        q = qubit()
+        assert q.operators[operator_name] == expected_operator
```
Ericgig: This could be a single test.

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+    def test_pauli_anticommutation(self, op1, op2):
+        """Test Pauli operator anticommutation relations"""
+        q = qubit()
+
+        op_1 = q.operators[op1]
+        op_2 = q.operators[op2]
+
+        # Different Pauli operators should anticommute
+        anticommutator = op_1 * op_2 + op_2 * op_1
+        assert anticommutator.norm() < 1e-10
```
Ericgig: remove

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+        excited = qt.basis(2, 1)
+
+        # σ_plus = (σ_minus)dagger
+        assert sigma_plus == sigma_minus.dag()
+
+        # Test basic actions
+        assert sigma_plus * ground == excited
+        assert sigma_minus * excited == ground
+        assert sigma_plus * excited == 0 * excited  # or qt.basis(2,0) * 0
+        assert sigma_minus * ground == 0 * ground   # or qt.basis(2,1) * 0
```
Ericgig: Redundant, remove.

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+        assert H_psi == eigenval_psi
+
+    def test_hamiltonian_structure(self):
+        """Test Hamiltonian structure and properties"""
+        q = qubit()
+
+        # Check dimension and Hermiticity
+        assert q.hamiltonian.shape == (2, 2)
+        H = q.hamiltonian
+        assert H == H.dag()
```
Ericgig:
```suggestion
        assert q.hamiltonian == q.hamiltonian.dag()
```

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+
+        required_attributes = [
+            'name',
+            'operators',
+            'hamiltonian',
+            'c_ops',
+            'latex',
+            'parameters']
+        for attr in required_attributes:
+            assert hasattr(q, attr)
```
Ericgig:
```suggestion
```

qutip/tests/quantum_systems/test_qubit.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # 4. Parameter ratios that might cause issues
+        q_large_decay = qubit(omega=0.1, decay_rate=100.0)
+        # Very fast decay compared to energy scale
+        assert q_large_decay.c_ops[0].norm() == np.sqrt(100.0)
+
+        # 5. Test that all operators remain valid
+        q_extreme = qubit(omega=1e10)
+        for op_name, op in q_extreme.operators.items():
+            assert op.check_herm() or op_name in ['sigma_plus', 'sigma_minus']
```
Ericgig: Remove

qutip/__init__.py
Vanshaj0429: Makes sense! Keeping them in their own namespace is cleaner and avoids potential conflicts.

qutip/quantum_systems/jaynes_cummings.py
Vanshaj0429: This was a result of autopep8, I have changed it.

qutip/quantum_systems/jaynes_cummings.py
Vanshaj0429: Okay, changed

qutip/quantum_systems/quantum_system.py
Vanshaj0429: Done

qutip/quantum_systems/quantum_system.py
Vanshaj0429: Yes, removed it

qutip/quantum_systems/quantum_system.py
Vanshaj0429: Yes actually, so simply a list then.

qutip/quantum_systems/quantum_system.py
Vanshaj0429:
```
def __init__(self, hamiltonian: Qobj, name: str = "Quantum System",
                 operators: dict = None, c_ops: list = None,
                 latex: str = "", **kwargs):
```
Some thing like this would suffice? And some same ordering change in the return statement in model functions.

qutip/quantum_systems/qubit.py
Vanshaj0429: Yes

qutip/quantum_systems/qubit.py
Vanshaj0429: Yes, corrected

qutip/quantum_systems/qubit.py
Vanshaj0429: Yes, added the defaults.

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Yes, made the change

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Yes made the correction, though just to confirm this import will take place from the `qutip/quantum_systems/init.py`?

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: made the change

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: made the change, also because we made the `hamiltonian` required in the constructor for `QuantumSystem`, I think this this would change to `system = QuantumSystem(0.5 * qt.sigmaz(), "Test")`

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: made the change

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Actually yes, this is not testing related to our class. We can remove it

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Removed this also, same issue

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Yes, can remove it

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Okay

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Corrected

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: I think can be completely removed as we are already testing `isinstance` in `test_default_qubit`.

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Yes, removed

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Okay

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Corrected

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Done

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Done

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Done

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: Okay, yes I see it

qutip/tests/quantum_systems/test_qubit.py
Vanshaj0429: I don't get what are you saying, something like :

```
def test_operators_correctness(self):
        """Test that operators are correct"""
        q = qubit()
        
        assert q.operators["sigma_minus"] == qt.destroy(2)
        assert q.operators["sigma_plus"] == qt.create(2)
        assert q.operators["sigma_z"] == qt.sigmaz()
        assert q.operators["sigma_x"] == qt.sigmax()
        assert q.operators["sigma_y"] == qt.sigmay()
```

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Yes makes sense

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: Same reason as above

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: This whole thing is kind of testing QuTip's mathematical properties instead. But since this QuantumSystem is a data container and doesn't have complex logic, it doesn’t really need edge case testing. I think we can remove the whole thing.

qutip/tests/quantum_systems/test_quantum_system.py
Vanshaj0429: It's not really an edge case, same reason as below comment

qutip/tests/quantum_systems/test_jaynes_cummings.py
Vanshaj0429: Done

qutip/tests/quantum_systems/test_jaynes_cummings.py
Vanshaj0429: Done

qutip/tests/quantum_systems/test_jaynes_cummings.py
Vanshaj0429: Done

qutip/tests/quantum_systems/test_jaynes_cummings.py
Vanshaj0429: Is using `np.sqrt` not ideal during testing?

qutip/tests/quantum_systems/test_jaynes_cummings.py
Vanshaj0429: Okay

qutip/tests/quantum_systems/test_jaynes_cummings.py
Vanshaj0429: Corrected

qutip/tests/quantum_systems/test_jaynes_cummings.py
Ericgig: `np.sqrt` is fine, but it's cleaner if it fit on one line.
Splitting the line inside a `(...)` for one element is not nice to the eye.

qutip/tests/quantum_systems/test_quantum_system.py
Ericgig: Yes, it will take it from the `__init__`.

qutip/tests/quantum_systems/test_quantum_system.py
Ericgig: Yes, no since it's a data container, it does not need that much testing.

qutip/tests/quantum_systems/test_qubit.py
Ericgig: yes

qutip/tests/quantum_systems/test_qubit.py
Ericgig: Ok

qutip/__init__.py
Vanshaj0429: > import quantum_systems

But this dosen't seem to work, gives an error no `quantum_systems` module found. But works with `from . import quantum_systems`

Vanshaj0429:
@Ericgig can you have a look, it was indeed annoying and makes the code messy, but I think it works!

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,168 @@
...
+def _create_sqrt_coefficient(rate):
+    """Helper function to create sqrt coefficient from decay rate"""
+    if isinstance(rate, Coefficient):
+        # Extract coefficient information and create sqrt version
+        try:
+            # Try to create a callable that returns sqrt of the coefficient evaluation
+            def sqrt_func(t, args):
+                return np.sqrt(rate.coeff(t, args) if hasattr(rate, 'coeff') else rate(t, args))
+            return coefficient(sqrt_func, args={})
+        except:
```
Ericgig: No empty `except`.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,168 @@
...
+from .quantum_system import QuantumSystem
+
+def _create_sqrt_coefficient(rate):
+    """Helper function to create sqrt coefficient from decay rate"""
+    if isinstance(rate, Coefficient):
+        # Extract coefficient information and create sqrt version
+        try:
+            # Try to create a callable that returns sqrt of the coefficient evaluation
+            def sqrt_func(t, args):
+                return np.sqrt(rate.coeff(t, args) if hasattr(rate, 'coeff') else rate(t, args))
```
Ericgig: Where did you take that `coeff` from?

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,168 @@
...
+            return coefficient(sqrt_func, args={})
+        except:
+            # Fallback: assume user provided the correct coefficient
+            return rate
+    else:
+        return np.sqrt(rate)
+    
+def jaynes_cummings(
+    omega_c: Union[float, coefficient] = 1.0,
```
Ericgig: Still the function, not the class.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,168 @@
...
+                return np.sqrt(kappa * thermal_photons)
+            sqrt_cavity_excite = coefficient(cavity_excite_func, args={})
+            c_ops.append(sqrt_cavity_excite * operators['a_dag'])
+        else:
+            # Both are float
+            cavity_excite_rate = cavity_decay * thermal_photons
+            c_ops.append(np.sqrt(cavity_excite_rate) * operators['a_dag'])
+
+    # Atomic spontaneous emission: sqrt(gamma) * sigma_minus
+    if isinstance(atomic_decay, Coefficient) or (isinstance(atomic_decay, (int, float)) and atomic_decay > 0.0):
```
Ericgig: Line too long. `isinstance` for detecting number is not a good idea. It could be a `np.float64`, tensorflow scalar, etc...

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,168 @@
...
+    # Total Hamiltonian
+    hamiltonian = H_cavity + H_atom + H_interaction
+
+    # Build collapse operators for dissipation
+    c_ops = []
+
+    # Cavity decay with thermal effects 
+    if isinstance(cavity_decay, Coefficient) or (isinstance(cavity_decay, (int, float)) and cavity_decay > 0.0):
+        # Cavity relaxation: sqrt(kappa(1 + n_th)]) * a
+        if isinstance(cavity_decay, Coefficient):
```
Ericgig: This would look better with:
```
if isinstance(cavity_decay, Coefficient):
   ...
elif cavity_decay > 0.0:
    ...
```

qutip/tests/quantum_systems/test_jaynes_cummings.py
```diff
@@ -0,0 +1,302 @@
...
+            g=g_coeff, cavity_decay=cavity_decay_coeff, atomic_decay=atomic_decay_coeff, n_cavity=3)
+
+        # Test that coefficients are stored
+        assert jc.parameters["omega_c"] == omega_c_coeff
+        assert jc.parameters["g"] == g_coeff
+        assert jc.parameters["cavity_decay"] == cavity_decay_coeff
+        assert jc.parameters["atomic_decay"] == atomic_decay_coeff
+
+        # Test that system is created successfully
+        assert isinstance(jc, QuantumSystem)
```
Ericgig, line 309: Test that the operators are the right format:
`assert isinstance(jc.hamiltonian, QobjEvo)`?

qutip/quantum_systems/jaynes_cummings.py
Vanshaj0429: Would you prefer I catch `Exception` instead of bare `except:`, or do you want me to identify the specific exception types that could occur in this context? @Ericgig

qutip/quantum_systems/jaynes_cummings.py
Vanshaj0429: I made an error with the QuTiP Coefficient API. I was confused about how to properly call Coefficient objects and mistakenly assumed there was a `.coeff()` method. I've corrected this - Coefficient objects are directly callable, so it should just be `rate(t, args)`

qutip/quantum_systems/jaynes_cummings.py
Vanshaj0429: Yes just realised, made the correction

qutip/quantum_systems/jaynes_cummings.py
Ericgig: No, please catch the exception you expect to happen.

Vanshaj0429:
@Ericgig made the changes, can you look again once to see if anything is missing?

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,172 @@
...
+def _create_sqrt_coefficient(rate):
+    """Helper function to create sqrt coefficient from decay rate"""
+    if isinstance(rate, Coefficient):
+        # Extract coefficient information and create sqrt version
+        try:
+            # Try to create a callable that returns sqrt of the coefficient evaluation
+            def sqrt_func(t, args):
+                return np.sqrt(rate(t, args))
+            return coefficient(sqrt_func, args={})
+        except(TypeError, AttributeError, ValueError):
```
Ericgig: I don't understand this...
In what situation would these error happen?
Could you add a test where this branch is taken and it does the expected behaviour.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,172 @@
...
+
+def _create_sqrt_coefficient(rate):
+    """Helper function to create sqrt coefficient from decay rate"""
+    if isinstance(rate, Coefficient):
+        # Extract coefficient information and create sqrt version
+        try:
+            # Try to create a callable that returns sqrt of the coefficient evaluation
+            def sqrt_func(t, args):
+                return np.sqrt(rate(t, args))
+            return coefficient(sqrt_func, args={})
```
Ericgig:
```suggestion
                return np.sqrt(rate(t, args))
                
            return coefficient(sqrt_func, args={})
```

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,172 @@
...
+    # Build collapse operators for dissipation
+    c_ops = []
+
+    # Cavity decay with thermal effects 
+    if isinstance(cavity_decay, Coefficient):
+        # cavity_decay is coefficient, thermal_photons is float
+        def cavity_relax_func(t, args):
+            kappa = cavity_decay(t, args)
+            return np.sqrt(kappa * (1 + thermal_photons))
+        sqrt_cavity_relax = coefficient(cavity_relax_func, args={})
```
Ericgig:
```suggestion
            return np.sqrt(kappa * (1 + thermal_photons))

        sqrt_cavity_relax = coefficient(cavity_relax_func, args={})
```

qutip/quantum_systems/qubit.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _create_sqrt_coefficient(rate):
+    """Helper function to create sqrt coefficient from decay rate"""
+    if isinstance(rate, Coefficient):
+        # Extract coefficient information and create sqrt version
+        try:
+            # Try to create a callable that returns sqrt of the coefficient evaluation
+            def sqrt_func(t, args):
+                return np.sqrt(rate(t, args))
+            return coefficient(sqrt_func, args={})
```
Ericgig:
```suggestion
            def sqrt_func(t, args):
                return np.sqrt(rate(t, args))
                
            return coefficient(sqrt_func, args={})
```

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,172 @@
...
+
+    H = omega_c * a_dag * a + (omega_a/2) * sigma_z + g * (a_dag * sigma_minus + a * sigma_plus)  [with RWA]
+    H = omega_c * a_dag * a + (omega_a / 2) * sigma_z + g * (a_dag + a) * (sigma_plus + sigma_minus) [without RWA]
+
+    Parameters:
+    -----------
+    omega_c : float or coefficient, default=1.0
+        Cavity frequency, can be constant or time-dependent
+    omega_a : float or coefficient, default=1.0
+        Atomic transition frequency, can be constant or time-dependent
```
Ericgig:
```suggestion
    omega_c : float or Coefficient, default=1.0
        Cavity frequency, can be constant or time-dependent
    omega_a : float or Coefficient, default=1.0
        Atomic transition frequency, can be constant or time-dependent
```
etc.

qutip/quantum_systems/jaynes_cummings.py
Vanshaj0429: I'm not sure when these specific exceptions would realistically occur with well-formed Coefficient objects. Looking at it again I think we can remove the try/except entirely and simply write it as:

```python
def _create_sqrt_coefficient(rate):
    """Helper function to create sqrt coefficient from decay rate"""
    if isinstance(rate, Coefficient):
        def sqrt_func(t, args):
            return np.sqrt(rate(t, args))
        return coefficient(sqrt_func, args={})
    else:
        return np.sqrt(rate)
 ```
 What do you think? or should I think of some test where we get those errors? @Ericgig

Comment (Ericgig) [reply qutip/quantum_systems/jaynes_cummings.py]: Whether you remove it or write a test, I leave it up to you.

I would just like to understand why you wrote that fallback in the first place.
Did you get all these errors? Did you get any error, don't remember what it was so just put common ones?

Comment (Vanshaj0429) [reply qutip/quantum_systems/jaynes_cummings.py]: I added the exception handling because I wasn't entirely certain about all the cases that QuTiP's Coefficient API would take, there could be scenarios where the coefficient evaluation fails (maybe numerical issues, argument mismatches, or coefficients that don't behave as expected), and upon such issue I wanted it return the original rate.

I did encounter TypeError, AttributeError when the coefficient function is not properly written

Comment (Ericgig) [reply qutip/quantum_systems/jaynes_cummings.py]: If the coefficient evaluation fails, then using it directly would not change anything. That should be an error passed to the user, not hidden.

Comment (Vanshaj0429) [reply qutip/quantum_systems/jaynes_cummings.py]: Yeah, I realised the issue when I read it again couple of times, thus the suggestion remove it

**Comment by Vanshaj0429** [comment]
@Ericgig, made the suggested changes, can you do a final review? I can't think of anything else that needs to be added.

File: qutip/quantum_systems/jaynes_cummings.py [hunk 66]
```diff
@@ -0,0 +1,170 @@
...
+            c_ops.append(np.sqrt(cavity_excite_rate) * operators['a_dag'])
+
+    # Atomic spontaneous emission: sqrt(gamma) * sigma_minus
+    # Check if it's a Coefficient
+    if isinstance(atomic_decay, Coefficient):
+        sqrt_atomic_decay = _create_sqrt_coefficient(atomic_decay)
+        c_ops.append(sqrt_atomic_decay * operators['sigma_minus'])
+    elif atomic_decay > 0.0:  
+        sqrt_atomic_decay = _create_sqrt_coefficient(atomic_decay)
+        c_ops.append(sqrt_atomic_decay * operators['sigma_minus'])
```
Ericgig: In the cavity decay case, it was not the same code for the Coefficient and scalar cases. Here it is the same...

```suggestion
    if isinstance(atomic_decay, Coefficient) or atomic_decay > 0.0:
        sqrt_atomic_decay = _create_sqrt_coefficient(atomic_decay)
        c_ops.append(sqrt_atomic_decay * operators['sigma_minus'])
```

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Atomic pure dephasing:  sqrt(gamma_phi)* sigma_z
+    if isinstance(atomic_dephasing, Coefficient):
+        sqrt_atomic_dephasing = _create_sqrt_coefficient(atomic_dephasing)
+        c_ops.append(sqrt_atomic_dephasing * operators['sigma_z'])
+    elif atomic_dephasing > 0.0:  
+        sqrt_atomic_dephasing = _create_sqrt_coefficient(atomic_dephasing)
+        c_ops.append(sqrt_atomic_dephasing * operators['sigma_z'])
```
Ericgig:
```suggestion
    if isinstance(atomic_dephasing, Coefficient) or atomic_dephasing > 0.0:
        sqrt_atomic_dephasing = _create_sqrt_coefficient(atomic_dephasing)
        c_ops.append(sqrt_atomic_dephasing * operators['sigma_z'])
```

qutip/quantum_systems/qubit.py
```diff
@@ -0,0 +1,81 @@
...
+        sqrt_decay_rate = _create_sqrt_coefficient(decay_rate)
+        c_ops.append(sqrt_decay_rate * operators['sigma_minus'])
+    
+    # Handle dephasing_rate: coefficient object OR positive numeric value
+    if isinstance(dephasing_rate, Coefficient):
+        sqrt_dephasing_rate = _create_sqrt_coefficient(dephasing_rate)
+        c_ops.append(sqrt_dephasing_rate * operators['sigma_z'])
+    elif dephasing_rate > 0.0:  
+        sqrt_dephasing_rate = _create_sqrt_coefficient(dephasing_rate)
+        c_ops.append(sqrt_dephasing_rate * operators['sigma_z'])
```
Ericgig: Here too.

Vanshaj0429: @Ericgig , made the changes in all the places

Vanshaj0429:
@Ericgig do you think it is ready to merge??

Ericgig (approved):
The code is good for a first version.

Ericgig:
@Vanshaj0429 running the tests. 
I saw we were forgetting the Towncrier entry.
When you add one, you can ask @nwlambert or @ajgpitch to review and merge.

The fail in python3.13 is unrelated to your code so it's fine.

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # Cavity operators (tensor with 2-level atomic system)
+    operators['a'] = tensor(destroy(n_cavity), qeye(2))
+    operators['a_dag'] = operators['a'].dag()
+    operators['n_c'] = operators['a_dag'] * operators['a']  # Photon number
+
+    # Atomic operators (tensor with cavity system)
+    operators['sigma_minus'] = tensor(qeye(n_cavity), destroy(2))
+    operators['sigma_plus'] = operators['sigma_minus'].dag()
+    operators['sigma_z'] = tensor(qeye(n_cavity), sigmaz())
```
nwlambert: if you are now just using sigma_- and sigma_+ in the hamiltonian, these pauli operators are slightly inconsistent with the Hamiltonian. As far as I can see only place you use one is in the dephasing, in which case the overall minus sign you are introducing probably doesn't matter since it gets squared away in the lindblad.  but to avoid some user confusion or errors, you could replace them with custom versions, `operators['sigma_z'] = operators['sigma_plus'] * operators['sigma_minus'] - operators['sigma_minus'] *  operators['sigma_plus']' ?

Vanshaj0429: So you are basically saying to keep it consistent and define the `sigma_z` in terms of `sigma_+` and `sigma_-`, because we have defined hamiltonian in those terms so to define `σ_z = σ_+σ_- - σ_-σ_+ = |e⟩⟨e| - |g⟩⟨g|`?
@nwlambert

nwlambert: yep exactly!

Vanshaj0429:
Made the changes, can you run the workflow again to see if it passes the towncrier check! @nwlambert

Vanshaj0429:
@nwlambert, sorry I forgot to amend the test earlier, hence it failed. Now made the change, can you please run it again.

Vanshaj0429:
@nwlambert, do you think it is ready to merge now?

qutip/quantum_systems/jaynes_cummings.py
```diff
@@ -72,7 +72,7 @@ def jaynes_cummings(
     # Atomic operators (tensor with cavity system)
     operators['sigma_minus'] = tensor(qeye(n_cavity), destroy(2))
     operators['sigma_plus'] = operators['sigma_minus'].dag()
-    operators['sigma_z'] = tensor(qeye(n_cavity), sigmaz())
+    operators['sigma_z'] = operators['sigma_plus'] * operators['sigma_minus'] - operators['sigma_minus'] * operators['sigma_plus']
     operators['sigma_x'] = tensor(qeye(n_cavity), sigmax())
     operators['sigma_y'] = tensor(qeye(n_cavity), sigmay())
```
nwlambert: sorry, i wasn't explicit enough in my previous comment; i meant to check all of the pauli operators.  sigma_x is invariant under this redefinition, but sigma_y probably needs a complex conjugate?

Vanshaj0429: Yeah, but I was also thinking do we really need to define sigma_x and sigma_y in this model because they are not being used anywhere? And if yes, then I think the standard relation between Pauli operator and ladder operator is 

`sigma_+ = (sigma_x + i*sigma_y)/2`
`sigma_- = (sigma_x - i*sigma_y)/2`

then,

`operators['sigma_x'] = operators['sigma_plus'] + operators['sigma_minus']`
`operators['sigma_y'] = 1j * (operators['sigma_minus'] - operators['sigma_plus'])`

@nwlambert

nwlambert: Yeah, just could also remove those two, and make sure how the model is defined is just explained in the docs.  Either is fine, upto you!

Vanshaj0429:
@nwlambert removed those operators and amended the test, no failing test. Can you run the workflow and then possibly merge if everything looks fine to you.

PR #2711, opened 2025-07-01
Add support for scipy 1.16

Ericgig:
**Description**
Update for scipy 1.16.

2 changes broke our tests:
- Faster dia operations: it does no longer clean the `0` after the operation. So `(dia - dia).nnz` is no longer 0 even if all values are zero. `==` raise a warning so `not any(dia != dia)` it is...
- New `sqrtm`, it raise a warning when the matrix is singular. I just filtered the warnings in the tests, expecting the result to still be right most of the time. But we should see it fail if the result it wrong.

### DIFF
--- a/qutip/tests/core/data/test_dia.py
+++ b/qutip/tests/core/data/test_dia.py
@@ -61,7 +61,7 @@ class TestClassMethods:
         """Test that __init__ can accept a scipy dia matrix."""
         out = dia.Dia(scipy_dia)
         assert out.shape == scipy_dia.shape
-        assert (out.as_scipy() - scipy_dia).nnz == 0
+        assert not np.any((out.as_scipy() != scipy_dia).data)
 
     def test_init_from_tuple(self, scipy_dia):
         """
@@ -71,7 +71,7 @@ class TestClassMethods:
         arg = (scipy_dia.data, scipy_dia.offsets)
         out = dia.Dia(arg, shape=scipy_dia.shape)
         assert out.shape == scipy_dia.shape
-        assert (out.as_scipy() - scipy_dia).nnz == 0
+        assert not np.any((out.as_scipy() != scipy_dia).data)
 
     @pytest.mark.parametrize('d_type', (
         _dtype_complex + _dtype_float + _dtype_int + _dtype_uint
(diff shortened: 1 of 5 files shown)


hodgestar:
The new SciPy warning seems wrong? The SciPy `sqrtm` claims to use Schur decomposition to calculate the matrix square root and that only fails if there are *two* zero eigenvalues. Perhaps worthwhile adding a note somewhere? Maybe in the implementation of sqrtm that calls out to SciPy?

qutip/tests/core/data/test_dia.py
```diff
@@ -61,7 +61,7 @@ def test_init_from_scipy(self, scipy_dia):
         """Test that __init__ can accept a scipy dia matrix."""
         out = dia.Dia(scipy_dia)
         assert out.shape == scipy_dia.shape
-        assert (out.as_scipy() - scipy_dia).nnz == 0
+        assert not np.any((out.as_scipy() != scipy_dia).data)
```
hodgestar: Maybe make a function for this in the test file so that they're easy to find and fix later if needed?

hodgestar (approved):
Happy for this to be merged to make tests pass. I left two small suggestions for making things nicer.

Ericgig:
@hodgestar 
The two zero eigenvalues condition does not appear in scipy documentation.
Where did you find it? Not sure if we should add a note for it...

I checked the remaining error and there seems to have a bug in sqrtm so I opened a issue with scipy: (scipy/23278).

hodgestar:
@Ericgig I found it by reading the Wikipedia page on [Square root of a matrix](https://en.wikipedia.org/wiki/Square_root_of_a_matrix#By_Schur_decomposition). SciPy uses the Schur decomposition, which reduces the problem to finding the square root of an upper triangular matrix. The upper triangular matrix is guaranteed to have a square root if at most one of its eigenvalues is zero (as a result of principle square roots of complex numbers always lying in one half of the complex plain). This latter point is explained on the same Wikipedia page but in the section on upper triangular matrices.

I also read the SciPy implementation and the warning is generated by an explicit check for whether the matrix is singular and is separate to the actual algorithm used.

hodgestar:
Great that the [SciPy issue](https://github.com/scipy/scipy/issues/23278) got fixed so quickly!

Ericgig:
@hodgestar 
I added an exception for scipy 1.16.0 at runtime. Without the bug fix, our tests can't work without implementing our own sqrtm... Is this fine? Should we wait for 1.16.1?

Issue #2595, opened 2025-01-08
Dimension mismatch in solving energy restricted eigenstates


ottosalmenkivi:
### Bug Description

Calling `Qobj.eigenstates()` of an operator created with energy restricted functionality raises an error. I believe this is due to the dimension refactoring of v5. The dimension of the solved eigenvectors are set by this list representation from `Qobj.dims` instead of the underlying `Dimension`/`Space` instances. Specifically here, the list is used instead of the `EnrSpace` instance, and thus information about the excitation limit is lost, raising a dimension/shape mismatch error in the Qobj constructor.

The simplest solution, inspired by another ENR issue #2387, would be to use `_dims` instead of `dims` in lines  1509-1512 (master at the time of writing) of `qobj.py`:
```
        if self.type == 'super':
            new_dims = [self._dims[0], [1]]
        else:
            new_dims = [self._dims[0], [1]*len(self.dims[0])]
```
I quickly checked that the superoperators work as well, but I am not sure how general the approach is. 

### Code to Reproduce the Bug

```shell
# Hamiltonian of two uncoupled qubits (harmonic oscillators limited to 2 levels) with energies e1 and e2, and only one excitation allowed in the system.

import qutip as qt
dims = [2, 2]
excitations = 1
e1, e2 = 1, 2
a1, a2 = qt.enr_destroy(dims, excitations)
H = e1*a1.dag()*a1 + e2*a2.dag()*a2
H.eigenstates()
```

### Code Output

```shell
ValueError                                Traceback (most recent call last)
Cell In[3], line 6
      4 a1, a2 = qt.enr_destroy(dims, excitations)
      5 H = e1*a1.dag()*a1 + e2*a2.dag()*a2
----> 6 H.eigenstates()

File ~\anaconda3\envs\bare_qt5\Lib\site-packages\qutip\core\qobj.py:1514, in Qobj.eigenstates(self, sparse, sort, eigvals, tol, maxiter, phase_fix)
   1512     new_dims = [self.dims[0], [1]*len(self.dims[0])]
   1513 ekets = np.empty((evecs.shape[1],), dtype=object)
-> 1514 ekets[:] = [Qobj(vec, dims=new_dims, copy=False)
   1515             for vec in _data.split_columns(evecs, False)]
   1516 norms = np.array([ket.norm() for ket in ekets])
   1517 if phase_fix is None:

[...]

PR #2713, opened 2025-07-07
ENR dimension bugfix for when QobjEvo is multiplied by Coefficient

ottosalmenkivi:
**Description**
As far as I understand, when multiplying a `QobjEvo` with a `Coefficient`, the latter is first turned into `QobjEvo` element with an identity matrix as the operator. When creating this operator, it's size is determined by the list-like `QobjEvo.dims`, instead of the dimension classes stored in the `_dims` attribute. The key difference is that the list-like dimension do not accommodate excitation-number restricted dimensions. Like in to many other similar instances, the solution is simple: add an underscore.

**Related issues or PRs**
Same fix in #2595


nwlambert:
Thanks @ottosalmenkivi this looks good to me (and kudos for adding a test too).  I think there's lots of small places like this where things need small fixes to work with enr states..  so its much appreciated.

PR #2718, opened 2025-07-30
Create `iadd` dispatched functions.

Ericgig:
**Description**
Create `iadd` dispatched function.
Here _inplace_ means that the memory space for `left` can be used in any way by the function. 
After the operation, `left` cannot be used anymore, the return is to be used.

Update integrator to use `iadd`. Useful for other dense representation such as GPU dense state.

### DIFF
--- a/qutip/core/data/add.pyx
+++ b/qutip/core/data/add.pyx
@@ -20,7 +20,7 @@ from qutip.core.data.convert import to as _to
 cnp.import_array()
 
 __all__ = [
-    'add', 'add_csr', 'add_dense', 'iadd_dense', 'add_dia',
+    'add', 'add_csr', 'add_dense', 'add_dia',
     'iadd', 'iadd_dense', 'iadd_dense_data_dense', 'iadd_data',
     'sub', 'sub_csr', 'sub_dense', 'sub_dia',
 ]
@@ -226,7 +226,7 @@ cpdef Data iadd_data(Data left, Data right, double complex scale=1):
 
 
 cpdef Dense iadd_dense_data_dense(Dense left, Data right, double complex scale=1):
-    """ Fix for the dispatcher, Dense left take priority. """
+    """ Helper function to manually set the priority of the dispatcher. """
     _check_shape(left, right)
     if scale == 0:
         return left


qutip/core/data/add.pyx
```diff
@@ -15,11 +15,13 @@ from qutip.core.data.csr cimport (
...
 )
 from qutip.core.data cimport csr, dense, dia
+from qutip.core.data.convert import to as _to
 
 cnp.import_array()
 
 __all__ = [
     'add', 'add_csr', 'add_dense', 'iadd_dense', 'add_dia',
+    'iadd', 'iadd_dense', 'iadd_dense_data_dense', 'iadd_data',
     'sub', 'sub_csr', 'sub_dense', 'sub_dia',
```
hodgestar:
```suggestion
    'add', 'add_csr', 'add_dense', 'add_dia',
    'iadd', 'iadd_dense', 'iadd_dense_data_dense', 'iadd_data',
    'sub', 'sub_csr', 'sub_dense', 'sub_dia',
```

`iadd_dense` was included twice.

qutip/core/data/add.pyx
```diff
@@ -214,6 +218,21 @@ cpdef Dense iadd_dense(Dense left, Dense right, double complex scale=1):
...
 
+cpdef Data iadd_data(Data left, Data right, double complex scale=1):
+    _check_shape(left, right)
+    if scale == 0:
+        return left
+    return add(left, right, scale)
+
+
+cpdef Dense iadd_dense_data_dense(Dense left, Data right, double complex scale=1):
+    """ Fix for the dispatcher, Dense left take priority. """
```
hodgestar: This doesn't feel like a fix for the dispatcher -- it's the dispatcher working as its supposed to. It just happens to make sense for `iadd` to convert right to dense if left is dense.

Ericgig:
```suggestion
    """ Helper function to manually set the priority of the dispatcher. """
```

hodgestar (approved):

hodgestar:
It should probably have a towncrier entry since it's an important new feature.

PR #2719, opened 2025-07-30
Pluging utilities

Ericgig:
**Description**
- Add `Dispatcher.register` a decorator for adding specialization.
- Add random generator input in random operator generators. 
  Not used here, but can be used to make the mathematics test reproductive.
- Add error message when settings `default_dtype_scope` without `default_dtype`.
- Fix `_group_specialisation` repr using removed group name.

### DIFF
--- a/qutip/core/data/dispatch.pyx
+++ b/qutip/core/data/dispatch.pyx
@@ -342,27 +342,36 @@ cdef class Dispatcher:
 
         Parameters
         ----------
-        dtype: tuple(type)
-            dtype of the function to register.
-            If all dispatched types are the same, a single entry is sufficient.
-            Otherwise each types should be passed.
+        *dtype: type
+            The data types for each of the dispatched arguments. There should
+            be one for each `Data` input and one for the output if also a
+            `Data` object.
+
+            As a convenience, if the specialisation uses the same data type for
+            all inputs, you can provide the type just once.
 
         Example
         -------
 
-        @func_dispatched.register(new_dtype)
-        def func_new_dtype(A: new_dtype, b: new_dtype) -> new_dtype:
+        @func.register(CSR)
+        def func_new_dtype(A: CSR, b: CSR) -> CSR:
             ...
 
         or
 
-        @func_dispatched.register(new_dtype, Dense, new_dtype)
-        def func_new_dtype(A: new_dtype, b: Dense) -> new_dtype:
+        @func.register(CSR, Dense, CSR)
+        def func_new_dtype(A: CSR, b: Dense) -> CSR:
             ...
 
         """
         if len(dtypes) == 1 and self._n_dispatch > 1:
             dtypes = dtypes * self._n_dispatch
+
+        if len(dtypes) != self._n_dispatch:
+            raise ValueError(
+                f"Dispatcher expects {self._n_dispatch} type arguments,"
+                f" but {len(dtypes)} were given."
+            )
 
         def _register(func):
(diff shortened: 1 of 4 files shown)


qutip/core/data/dispatch.pyx
```diff
@@ -336,6 +336,49 @@ cdef class Dispatcher:
...
+    def register(self, *dtypes):
+        """
+        Decorator for add_specialisations.
+
+        Parameters
+        ----------
+        *dtype: type
+            The data types for each of the dispatched arguments. There should
+            be one for each `Data` input and one for the output if also a
+            `Data` object.
```
hodgestar:
```suggestion
            The data types for each of the dispatched arguments. There should
            be one for each `Data` input and one for the output if that is also a
            `Data` object.
```

qutip/core/data/dispatch.pyx
```diff
@@ -336,6 +336,49 @@ cdef class Dispatcher:
...
+        """
+        if len(dtypes) == 1 and self._n_dispatch > 1:
+            dtypes = dtypes * self._n_dispatch
+
+        if len(dtypes) != self._n_dispatch:
+            raise ValueError(
+                f"Dispatcher expects {self._n_dispatch} type arguments,"
+                f" but {len(dtypes)} were given."
+            )
+
```
hodgestar, line 383: I suggest asserting that all elements of `dtypes` are sub-classes of Data to avoid confusing error messages in accidents like:
```
@func.register
def func_new_dtype(...):
    ...
```

qutip/core/data/dispatch.pyx
```diff
@@ -336,6 +336,49 @@ cdef class Dispatcher:
...
+        def _register(func):
+            self.add_specialisations([dtypes + (func,)])
```
hodgestar:
```suggestion
            self.add_specialisations([dtypes + (func,)], _defer=True)
```
Should we set `_defer=True` here to avoid continuously recalculating the specializations during import?

qutip/core/options.py
```diff
@@ -93,6 +93,10 @@ def _set_default_dtype_scope(new_range):
             "'default_dtype_scope' must be one of "
             "'creation', 'missing', 'full'"
         )
+    if new_range != "creation" and settings.core["default_dtype"] is None:
```
hodgestar, line 96: Why is okay not to specify the `default_type` when the scope is `creation`? It would seem that setting the scope without a dtype never makes sense?

qutip/tests/core/data/conftest.py
```diff
@@ -16,26 +16,30 @@ def shuffle_indices_scipy_csr(matrix):
     general, we attempt to shuffle, and if this returns the same order as
     before, we just reverse it to ensure it's different.
     """
+    if gen is None:
```
hodgestar, line 22: Let's add something about `gen` to the docstring.

qutip/core/options.py
Ericgig: Changing `default_dtype_scope` with other value has the dispatcher rebuilding the table and trying to create operation resulting in the `default_dtype`. It will fail and leave the user with incomplete dispatcher tables... With `creation` it create them to the most efficient version (according to weight that we rarely add) and everything work as expected.

Now that group alias exist (they were created after `default_dtype_scope`). We could / should change `default_dtype` default to `core` and remove support for `None`.
But since it does not cause any bug as is, I will leave that for another PR.

hodgestar (approved):

PR #2720, opened 2025-07-30
Add tsit5 integrator method

Ericgig:
**Description**
Add a new integrator using qutip data as state, tsit5.
The existing vern7, vern9 are high order an require a lot of derivatives and states, which use a lot of RAM.
The tsit5 use 7 steps and about half the memory of vern7.

It could be optimized more: it lack the dense output and one step is repeated.

### DIFF
--- a/qutip/solver/integrator/explicit_rk.pxd
+++ b/qutip/solver/integrator/explicit_rk.pxd
@@ -27,6 +27,7 @@ cdef class Explicit_RungeKutta:
     cdef readonly int max_numsteps
     cdef readonly bint interpolate
     cdef readonly object method
+    cdef dict butcher_tableau
 
     # Runge Kutta tableau and info
     cdef int rk_step, rk_extra_step, order, denseout_order
--- a/qutip/solver/integrator/explicit_rk.pyx
+++ b/qutip/solver/integrator/explicit_rk.pyx
@@ -19,13 +19,6 @@ import numpy as np
 __all__ = ["Explicit_RungeKutta"]
 
 
-euler_coeff = {
-    'order': 1,
-    'a': np.array([[0.]], dtype=np.float64),
-    'b': np.array([1.], dtype=np.float64),
-    'c': np.array([0.], dtype=np.float64)
-}
-
 rk4_coeff = {
     'order': 4,
     'a': np.array([[0., 0., 0., 0.],
@@ -100,9 +93,19 @@ cdef class Explicit_RungeKutta:
         * It also accept a tuple of runge-kutta coefficients.
         (See `Explicit_RungeKutta._init_coeff`)
     """
-    def __init__(self, QobjEvo qevo, double rtol=1e-6, double atol=1e-8,
-                 int nsteps=1000, double first_step=0, double min_step=0,
-                 double max_step=0, bint interpolate=True, method="euler"):
+    def __init__(
+            self,
+            QobjEvo qevo,
+            dict butcher_tableau=None,
+            str method="",
+            double rtol=1e-6,
+            double atol=1e-8,
+            int nsteps=1000,
+            double first_step=0,
+            double min_step=0,
+            double max_step=0,
+            bint interpolate=True,
+        ):
         # Function to integrate.
         self.qevo = qevo
         # tolerances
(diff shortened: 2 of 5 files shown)


qutip/solver/integrator/explicit_rk.pxd
```diff
@@ -27,6 +27,7 @@ cdef class Explicit_RungeKutta:
     cdef readonly int max_numsteps
     cdef readonly bint interpolate
     cdef readonly object method
+    cdef dict butcher_tableau
```
hodgestar: Should `butcher_tableau` be readonly too?

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -437,7 +436,7 @@ cdef class Explicit_RungeKutta:
     @cython.cdivision(True)
     cdef double _get_timestep(self, double t):
         """ Get the dt for the step. """
-        cdef double dt_needed = t - self._t_front
+        cdef double dt_needed = t - self._t_prev
```
hodgestar, line 444: Was this a bug?

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -122,17 +125,12 @@ cdef class Explicit_RungeKutta:
...
-            self._init_coeff(**method)
-        elif "vern7" == method:
-            self._init_coeff(**vern7_coeff)
-        elif "vern9" == method:
-            self._init_coeff(**vern9_coeff)
-        elif "rk4" == method:
+        if butcher_tableau is None:
             self._init_coeff(**rk4_coeff)
         else:
-            self._init_coeff(**euler_coeff)
```
hodgestar, line 134: Could we keep the Euler method. I'm happy with Runge-Kutta 4th Order being the default, but it's nice to have the Euler option precisely because its so simple.

qutip/tests/solver/test_integrator.py
```diff
@@ -184,8 +184,27 @@ def test_pickling_vern_methods(integrator):
...
         result1 = inter.integrate(t)[1].to_array()[0, 0]
         result2 = recreated.integrate(t)[1].to_array()[0, 0]
         assert result1 == result2 == expected
+
+@pytest.mark.parametrize('integrator',
+    [IntegratorVern7, IntegratorVern9, IntegratorTsit5],
+    ids=["vern7", 'vern9', 'tsit5']
+)
+def test_rk_options(integrator):
+    """Test whether VernN methods can be pickled and hence used in multiprocessing"""
```
hodgestar:
```suggestion
    """Test whether VernN and Tsitoura's methods can be pickled and hence used in multiprocessing"""
```

qutip/solver/integrator/explicit_rk.pxd
Ericgig: I don't see any use for it and `dict` are mutable, so it can be modified when readonly.

hodgestar: In Cython `readonly` is about accessibility from Python. So `cdef public butcher_tableau` would add a setter and a getter while `cdef readonly butcher_tableau` would add only a getter. I suggested adding a getter for consistency with the rest of the attributes.

qutip/solver/integrator/explicit_rk.pyx
Ericgig: Yes...
It work fine when for step under tolerance, both values are the same. But when the step was refused, it would return a `dt` a lot smaller than needed for the next step, often dt_needed would be 0, resulting in a step of `min_step`, default 1e-15.

It did not affect output (within tol), but wasted time doing absurdly small steps.

hodgestar: Good find!

qutip/solver/integrator/explicit_rk.pyx
Ericgig: I made the tableau input required and moved the both simple tableaux to python space.

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -72,6 +54,15 @@ cdef class Explicit_RungeKutta:
     qevo : QobjEvo
         The system to integrate: ```dX = qevo.matmul_data(t, X)``
 
+    butcher_tableau : dict
+        The coefficient as butcher tableau:
```
Ericgig:
```suggestion
        The coefficients as butcher tableau:
```

qutip/solver/integrator/explicit_rk.pxd
```diff
@@ -26,12 +26,12 @@ cdef class Explicit_RungeKutta:
     cdef readonly double rtol, atol, first_step, min_step, max_step
     cdef readonly int max_numsteps
     cdef readonly bint interpolate
-    cdef readonly object method
 
     # Runge Kutta tableau and info
     cdef int rk_step, rk_extra_step, order, denseout_order
-    cdef bint adaptative_step, can_interpolate
+    cdef bint adaptative_step, can_interpolate, fsal
```
hodgestar:
```suggestion
    cdef bint adaptative_step, can_interpolate, fsal
```
It would be nice to explain `fsal` a little in a comment and give it a better name. I.e. that it means that the last derivative from the previous step is the same as the first derivative from the next step. Since we use longer names for the other values, perhaps we should also rename it `first_derivative_same_as_last` although that seems long.

PR #2722, opened 2025-07-31
Fix test

Ericgig:
**Description**
I made an error in #2721 which made it skip all test.
Seeing only towncrier check failing, it looked good...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
         numpy-requirement: [""]
         scipy-requirement: [">=1.9"]
         coverage-requirement: ["==6.5"]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         numpy-requirement: [""]
         scipy-requirement: [">=1.9"]
         coverage-requirement: ["==6.5"]
-        pytest-extra-options: "-W ignore:Matrix:scipy.linalg._misc.LinAlgWarning"
+        pytest-extra-options: ["-W ignore:Matrix:scipy.linalg._misc.LinAlgWarning"]
```
hodgestar, line 41: How come this one has square braces and the others don't?

Ericgig: This is the entry in the matrix section, the other are in include section.

PR #2725, opened 2025-08-11
Krylov and Lanczos for Density Matrices

Langhaarzombie:
**Description**
Enables the calculation of time dynamics for density matrices using the Lanczos algorithm for the Krylov subspace method.


qutip/solver/krylovsolve.py
```diff
@@ -47,8 +48,8 @@ def krylovsolve(
         Hamiltonians. List of [:obj:`.Qobj`, :obj:`.Coefficient`] or callable
         that can be made into :obj:`.QobjEvo` are also accepted.
 
-    psi0 : :class:`.Qobj`
-        Initial state vector (ket)
+    rho0 : :class:`.Qobj`
+        Initial density matric or state vector (ket).
```
Ericgig:
```suggestion
        Initial density matrix or state vector (ket).
```

qutip/solver/krylovsolve.py
```diff
@@ -119,5 +120,11 @@ def krylovsolve(
...
     options["method"] = "krylov"
     options["krylov_dim"] = krylov_dim
-    solver = SESolver(H, options=options)
-    return solver.run(psi0, tlist, e_ops=e_ops)
+
+    if rho0.isket:
+        return sesolve(H, rho0, tlist, e_ops=e_ops, args=args,
+                       options=options)
+
+    solver = MESolver(H, None, options=options)
```
Ericgig: No `c_ops`? Why use Mesolve?

qutip/solver/integrator/krylov.py
```diff
@@ -237,4 +250,4 @@ def options(self, new_options):
         Integrator.options.fset(self, new_options)
 
 
-SESolver.add_integrator(IntegratorKrylov, 'krylov')
+Solver.add_integrator(IntegratorKrylov, 'krylov')
```
Ericgig: Can krylov do `mcsolve`? `Solver` include it.

Langhaarzombie: There are a few problems that arise with Krylov and `mcsolve`, so I have removed it for now. Better for future work.

qutip/solver/krylovsolve.py
Langhaarzombie: I am sorry, it took me a long time to figure this out but...
For the first step, `krylovsolve` shall be extended to include density matrices in closed systems. However, `sesolve` does not preserve hermiticity and thus `mesolve` is needed.

An actual question in that regard: Since the extension to DMs requires the inclusion into MeSolve, the tests also throw`c_ops` at `krylovsolve`. Since I wanted to leave support for that till the next step, is there a way to do this properly? I.e. test `krylovsolve` within the scope of `test_mesolve` but without `c_ops`?
Please let me know if my thinking here is wrong or if I am missing a best practice!

Langhaarzombie: fyi this concearns exactely the one test that is currently failing

Ericgig: You could remove the 
```
SESolver.add_integrator(IntegratorKrylov, 'krylov')
MESolver.add_integrator(IntegratorKrylov, 'krylov')
```
lines, and in `krylovsolve`, instead of passing the string for the method, pass the integrator directly.
`options["method"] = IntegratorKrylov`.
That way, most `sesolve` and `mesovle` tests will skip the krylov method.

Langhaarzombie: Thank you for the quick response!
However, when setting the Integrator directly I am getting a key error because in the base `Solver` class we have:
`
integrator = self.avail_integrators()[method]
`, where `method` directly corresponds to `options["method"]` as you suggested above.

So setting the object directly does not to work I think. Also directly setting the integrator property of the solver class is a bit of a mess because of the specific options.

Ericgig: It works for getting the integrator, but not in the options...

Then maybe the best would be to remove it from `all_ode_method` in `test_mesolve`.

Langhaarzombie: Thanks!
I also had to exclude kyrlov from some tests in `test_integrator` for the same reason. You can see [8bf0cc6](https://github.com/qutip/qutip/pull/2725/commits/8bf0cc65d0ff988ea78687069fbb0d7ae6c50710) for the changes. I hope this works with the general code styling.

Langhaarzombie:
Since a lot has changed and I am formally opening this for review, I am putting the checklist and a more detailled description for this PR.

The ``krylovsolve`` method gets extended to support calculations for both closed and open (Lindblad-like) systems. For that, I added the fully reorthogonalized Lanczos and Arnoldi algorithm. The former improves the standard Lanczos by greatly reducing numerical errors. The latter is required to optain a Krylov space for open systems. The user can still choose which algorithm to use, since they differ in computation time.

Additionally the solver was fixed in terms of handling different system dimensions and the *happy breakdown* correctly.

**Technical TODOs / ideas**

- [ ] We can discuss an increase in tests to make the method more robust
- [x] Potential code style changes for more efficient interplay with data layer and built in qutip functions
- [x] Change error estimation method (to be discussed)

**Contribution Checklist**

- [x] pep8 codestyle checks
- [x] Update docs

qutip/tests/solver/test_mesolve.py
```diff
@@ -711,3 +712,45 @@ def test_non_normalized_dm(rho0):
...
     np.testing.assert_allclose(result.expect[0], rho0.tr(), atol=1e-7)
+
+
+@pytest.mark.parametrize("kdim", [50, 1000])
+@pytest.mark.parametrize("algorithm", ['lanczos_fro', 'arnoldi'])
+def test_krylovsolve_mixed_state(kdim, algorithm):
+    # Kyrlov method works best with dense Hamiltonians.
+    # Since rand_herm(.) usually provides sparse matrices, we are summing
+    # multiple samplings.
+    H = np.sum([qutip.rand_herm(20) for _ in range(30)]) / 30
```
Ericgig: This is not the way to do this...

qutip/solver/krylovsolve.py
```diff
@@ -4,34 +4,36 @@
...
 def krylovsolve(
     H: Qobj,
-    psi0: Qobj,
+    rho0: Qobj,
     tlist: ArrayLike,
-    krylov_dim: int,
-    _e_ops = None,
-    _args = None,
-    _options = None,
+    krylov_dim: int = 0,
```
Ericgig: Since this is an helper function for the krylov method, I think the algorithm options should be available as a keyword argument here.

qutip/solver/integrator/krylov.py
```diff
@@ -1,150 +1,262 @@
...
+            raise ValueError("The option 'krylov_dim', must be an integer "
+                             "greater or equal zero.")
+
+        if self.options['algorithm'] == 'arnoldi' or not self._hermitian:
+            # Arnoldi is the only algorithm supporting open systems so far
+            self._algorithm = self._arnoldi_algorithm
+        elif self.options['algorithm'] == 'lanczos_fro':
+            self._algorithm = self._lanczos_full_reorth_algorithm
+        elif self.options['algorithm'] == 'lanczos':
+            self._algorithm = self._lanczos_algorithm
```
Ericgig: So, if a user with mesolve (c_ops, non-hermitian) ask for the "lanczos" algorithm, it will quietly ignore the user input and use "arnoldi" instead...

At a minimum. let's raise a warning and update the options in Result to reflect the algorithm used. Or you could also raise an error (ValueError or NotImplementedError).

qutip/solver/integrator/krylov.py
```diff
@@ -1,150 +1,262 @@
...
-            [T_subdiag[:j], T_diag[:j+1], T_subdiag[:j]],
+            [subdiag[:j-1], diag[:j], subdiag[:j-1]],
             [-1, 0, 1]
         )
-        krylov_basis = _data.Dense(np.hstack([psi.to_array() for psi in v]))
+        krylov_basis = _data.Dense(np.hstack([p.to_array() for p in v]))
 
         return krylov_tridiag, krylov_basis
 
+    def _lanczos_full_reorth_algorithm(self, psi):
```
Ericgig: Can't this be merged with "_lanczos_algorithm" to limit code duplication?

qutip/solver/integrator/krylov.py
```diff
@@ -164,12 +276,27 @@ def krylov_error(t):
...
+        self._size = state0.shape[0]
+
+        if self._mat_state:
+            state0 = _data.column_stack(state0)
+            if not self.system.issuper:
+                self.system = -1j * liouvillian(self.system)
+
+        if self.options["krylov_dim"] == 0:
+            d = self.system.shape[0]
+            self._krylov_dim = d**2 - d + 1 if self._mat_state else d
```
Ericgig: Could you explain this choice?

Ericgig: Also `krylov_dim` should be in the Result so the users so the user can see what was used.

qutip/solver/integrator/krylov.py
```diff
@@ -164,12 +276,27 @@ def krylov_error(t):
...
+
+        krylov_tridiag, krylov_basis = self._algorithm(state0)
+        self._krylov_state = \
+            self._compute_krylov_set(krylov_tridiag, krylov_basis)
 
         if (
-            krylov_tridiag.shape[0] <= self.options['krylov_dim']
-            or krylov_tridiag.shape == self.system.shape
+            krylov_tridiag.shape <= self.system.shape and
+            krylov_tridiag.shape[0] <= self._krylov_dim
```
Ericgig: Are you sure this is right?

qutip/solver/integrator/krylov.py
```diff
@@ -164,12 +276,27 @@ def krylov_error(t):
 
     def set_state(self, t, state0):
         self._t_0 = t
-        krylov_tridiag, krylov_basis = self._lanczos_algorithm(state0)
-        self._krylov_state = self._compute_krylov_set(krylov_tridiag, krylov_basis)
+        self._mat_state = state0.shape[1] > 1
```
Ericgig: The density matrices are already column stacked when passed to the integrator.
The initial state is a matrix (identity) when computing the propagator...

qutip/solver/integrator/krylov.py
```diff
@@ -1,150 +1,262 @@
...
         aux = _data.multiply(phases, e0)
-        return _data.matmul(U, aux)
-
-    def _compute_max_step(self, krylov_tridiag, krylov_basis, krylov_state=None):
+        statet = _data.matmul(U, aux)
+        if self._mat_state:
+            return _data.column_unstack_dense(statet, self._size, inplace=True)
+        else:
+            return statet
+
```
Ericgig, line 231: Too many spaces...

qutip/solver/integrator/krylov.py
```diff
@@ -179,12 +306,11 @@ def set_state(self, t, state0):
...
         ):
-            self._max_step = self._compute_max_step(
-                krylov_tridiag, krylov_basis, self._krylov_state,
-            )
+            self._max_step = self._compute_max_step(krylov_tridiag, krylov_basis)
 
     def get_state(self, copy=True):
-        return self._t_0, self._compute_psi(0, *self._krylov_state)
+        state = self._compute_psi(0, *self._krylov_state)
+        return self._t_0, state.copy() if copy else state
```
Ericgig: The state is created with `statet = _data.matmul(U, aux)` so this is always the unique reference. No need for a copy...

qutip/solver/integrator/krylov.py
```diff
@@ -193,13 +319,16 @@ def integrate(self, t, copy=True):
...
+                    "Maximum number of integration steps "
+                    f"({self.options['nsteps']}) exceeded"
+                )
             new_psi = self._compute_psi(self._max_step, *self._krylov_state)
             self.set_state(self._t_0 + self._max_step, new_psi)
 
         delta_t = t - self._t_0
         out = self._compute_psi(delta_t, *self._krylov_state)
-        return t, out
+        return t, out.copy() if copy else out
```
Ericgig: Same

qutip/solver/integrator/krylov.py
```diff
@@ -1,150 +1,262 @@
...
             return np.log(_data.norm.l2(
-                self._compute_psi(t, *krylov_state) -
-                self._compute_psi(t, *reduced_state)
+                _data.column_stack(self._compute_psi(t, *krylov_state)) -
+                _data.column_stack(self._compute_psi(t, *reduced_state))
             ) / self.options["atol"])
 
         # Under 0 will cause an infinite loop in the while loop bellow.
-        dt = max(self.options["min_step"], 1e-14)
+        dt = self.options["min_step"]
```
Ericgig: Why did you remove the safety check?

Langhaarzombie: Got lost during debugging. I added it again. Thanks!

qutip/solver/integrator/krylov.py
Langhaarzombie: I am not sure I undertand what you mean. Do you mean the indentation? Or should I condense the block of code?

qutip/solver/integrator/krylov.py
Langhaarzombie: Thank you, after checking it again, it was not right. Fixed below.

qutip/solver/integrator/krylov.py
Langhaarzombie: > Could you explain this choice?

The maximum krylov dimension for a pure state in a closed system is equal to the system size. For mixed states in open or closed systems, the maximum dimension of the Krylov space is d**2 - d + 1 where d is the system size. In case you are interested, [this is the paper](https://arxiv.org/pdf/2009.01862) (eq. 2.7).

Of course, we could simply put d**2 as the default and the algorithm will stop automatically anyways. Do you think this is better?

qutip/solver/integrator/krylov.py
Langhaarzombie: I added a warning now and kept the defaulting to Arnoldi.

I did that so that the tests run smoothly (since they just call the integrator with no extra options). I can change the tests in case you think an error is better here?

Langhaarzombie:
The latest version of the code should fix this simple code example.

Be aware that when you are using the master version of krylov solve, there is a bug that the actual krylov dimension that is used, is +1 of what you request (so 6 in your example).
Therefore, when you put dimension 5 in the new version of the solver, you get an error since there are too many integration steps (but at least the dimension is what you wanted). When you change the dimension to 6, you will get exactly the same result as with the master verison.

qutip/solver/integrator/krylov.py
Ericgig: There are 12 spaces before `self`, `krylov_tridiag`, etc. there should be 8.
The snipped is not showing the right part... L224~.

qutip/solver/integrator/krylov.py
Ericgig: Isn't krylov dimensions supposed to be smaller than the operator size for krylov method to be useful? Otherwise you could just diagonalize the operator directly.
I would have expected the size to be about `sqrt(d)` or an small fraction of the system size.
If there is no way to guess a good number, wouldn't it be better to simply raise an error saying `krylov_dim` is not valid?

Langhaarzombie: Yes, that is generally true. I added this *guess* just because the master version has a default dimension too. There it is `d-1`.

But leaving it up to the user sounds better indeed. So I will raise an error when `krylov_dim <= 0`.

qutip/solver/integrator/krylov.py
Langhaarzombie: After some thinking, I found a better solution maybe. Introducing `"auto"` as the default option for the solver keeps the tests at peace and lets us automatically choose Arnoldi if the system is non-Hermitian. Without bothering the user. If the wrong algorithm is chosen, then I will simply raise an error.

Let me know if that sounds better.

qutip/solver/integrator/krylov.py
Langhaarzombie: > Also `krylov_dim` should be in the Result so the users so the user can see what was used.

Before getting deep into this: is there a straight forward way to do this? From first inspection, having a custom `KrylovResult` also means having a specific `KrylovSolver` class (instead of just using MESolver).

Ericgig: The options are included in the result. If you raise and error instead of modifying, you don't need to do anything.

qutip/solver/integrator/krylov.py
Ericgig: Yes it could work.

qutip/tests/solver/test_mesolve.py
Langhaarzombie: Now it's replaced with the proper density argument.

I had to increase the tolerance to 1e-5. to accomodate a small error (compared to mesolve) in the expectation value over time that is oscillating around 0. At longer times (1000 times longer than in the test), the error stays withing 1e-3, so I deemed this increase reasonable. This error is essentially the same for the master version of the code.

Do let me know if that is still withing desired bounds.

qutip/solver/integrator/krylov.py
```diff
@@ -1,174 +1,266 @@
...
     method = 'krylov'
 
     def _prepare(self):
         if not self.system.isconstant:
-            raise ValueError("krylov method only support constant system.")
+            raise ValueError("Krylov method only supports constant systems.")
+
+        if self.options["krylov_dim"] <= 0:
+            raise ValueError("The option 'krylov_dim', must be an integer "
+                             "greater zero.")
```
Ericgig:
```suggestion
                             "greater than zero.")
```

qutip/solver/integrator/krylov.py
```diff
@@ -1,174 +1,266 @@
...
         phases = _data.Dense(np.exp(-1j * dt * eigenvalues))
         aux = _data.multiply(phases, e0)
         return _data.matmul(U, aux)
 
-    def _compute_max_step(self, krylov_tridiag, krylov_basis, krylov_state=None):
+    def _compute_max_step(
+        self,
+        krylov_tridiag,
+        krylov_basis,
+        krylov_state=None
```
Ericgig: Can we remove `krylov_state`?

qutip/solver/integrator/krylov.py
```diff
@@ -1,174 +1,266 @@
...
+        dt = np.power(num / bsprod, 1 / krylov_tridiag.shape[0])
+        return np.real(dt)
+        
     def set_state(self, t, state0):
         self._t_0 = t
-        krylov_tridiag, krylov_basis = self._lanczos_algorithm(state0)
-        self._krylov_state = self._compute_krylov_set(krylov_tridiag, krylov_basis)
+
+        if state0.shape[1] > 1 and not self.system.issuper:
+            self.system = -1j * liouvillian(self.system)
```
Ericgig: This is not used as you expect.
The integrator always receive density matrices as `column_stacked`.
Case where "state0.shape[1] > 1" are when the input state is an operator (`qeye`) to compute the propagator.
Also the system will be super for mesolve and not for sesolve.
So this will update the system when used to compute closed system propagators...

qutip/solver/integrator/krylov.py
```diff
@@ -217,8 +311,14 @@ def options(self):
 
         krylov_dim: int, default: 0
             Dimension of Krylov approximation subspaces used for the time
-            evolution approximation. If the defaut 0 is given, the dimension is calculated
-            from the system size N, using `min(int((N + 100)**0.5), N-1)`.
+            evolution approximation. If the defaut 0 is given, the full Krylov
+            space is calculated.
```
Ericgig:
```suggestion
            evolution approximation.
```

qutip/solver/krylovsolve.py
```diff
@@ -47,15 +50,23 @@ def krylovsolve(
...
         list of times for :math:`t`.
 
     krylov_dim: int
-        Dimension of Krylov approximation subspaces used for the time
-        evolution approximation.
+        Dimension of Krylov subspaces used for the time evolution approximation.
+        Note that for mixed states, the Krylov dimension can be larger than the
+        dimension of the Hamltonian since we are computing the Liouvillian.
+
+    algorithm: str
```
Ericgig:
```suggestion
    algorithm: str, default: "auto"
```

qutip/solver/integrator/krylov.py
```diff
@@ -1,174 +1,266 @@
...
-        else:
-            return dt
-
+            krylov_state = \
+                self._compute_krylov_set(krylov_tridiag, krylov_basis)
+
+        bsprod = np.prod(np.diag(krylov_tridiag.as_ndarray(), k=-1))
+        num = self.options["atol"] * factorial(krylov_tridiag.shape[0])
+        dt = np.power(num / bsprod, 1 / krylov_tridiag.shape[0])
+        return np.real(dt)
```
Ericgig: `max_step` and `min_step` are ignored...
You could enforce them, remove them from options or raise and error when outside...

doc/guide/dynamics/dynamics-krylov.rst
```diff
@@ -47,27 +48,65 @@ After a short time, the error starts to grow exponentially. However, this can be
...
+
+Same as for the pure states mentioned above, the Krylov basis vectors are
+calculated using the Lanczos algorithm (or alternative - see below).
+
+Alternative Algorithms for Kyrlov Construction
+----------------------------------------------
+
+As the number of degrees of freedom grow, the standard Lanczos algorithm shows
+significant errors due to numerical inaccuracies.
+This is why the *default* algorithm in QuTiP to construct the Krylov basis is
```
Ericgig, line 84: This default here + Arnoldi enforced later doesn't sound clear...
How about leaving the default explanation outside of the  algorithm explanation?

qutip/solver/integrator/krylov.py
Langhaarzombie: Good point, I actually put them back in, I think they are important to determine what a good krylov dimension are.

qutip/solver/integrator/krylov.py
Langhaarzombie: I understand, thank you! I just removed the code.

Langhaarzombie:
I implemented all the changes, thanks a lot!

As for the krylov dimension in the documentation: I added a paragraph to give some intuition, but it is really hard to determine the "best possible" dimension. I hope it helps a bit.

Ericgig (approved):
Seems good.
Thank you.

Ericgig:
@Langhaarzombie 
Could you fix the conflict.

PR #2726, opened 2025-08-11
Use local rtol in runge-kutta.

Ericgig:
**Description**
To compute the error in runge-kutta solver, we use the norm of the difference vector (difference between 2 approximations):

`error = norm(diff_state) / (atol + rtol * norm(state))`

However, the usual way to do it (looking at the source code of scipy and diffrax) is to apply the relative factor per element and average:

```
error = sqrt( 1/N * sum(
    (
        |diff_state[i]| 
        / (atol + rtol * |state[i]| )
    ) ** 2
))
```
They also accept `atol`, `rtol` per elements...

I created a new dispatched function for the error as it can't be obtained from available functions.


qutip/core/data/ode.pyx
```diff
@@ -0,0 +1,214 @@
...
+            f"incompatible matrix shapes {left.shape} and {right.shape}"
+        )
+    return 0
+
+
+cpdef double wrmn_error_dense(Dense diff, Dense state, double atol, double rtol) except -1:
+    """
+    Compute the weighted root mean square norm for the ode error estimation
+
+    error = sqrt( 1/N * sum( ( |diff[i]| / (atol + rtol * |state[i]|))**2 ))
```
hodgestar:
```suggestion
    Compute the weighted root mean square error norm:

        error = sqrt( 1/N * sum( ( |diff[i]| / (atol + rtol * |state[i]|))**2 ))
        
    This error norm is commonly used to estimate error in ODEs.
```

qutip/core/data/ode.pyx
```diff
@@ -0,0 +1,214 @@
...
+    """
+    cdef double sum = 0.
+    cdef size_t i, j, N = diff.shape[0] * diff.shape[1]
```
hodgestar:
```suggestion
    cdef size_t i, N = diff.shape[0] * diff.shape[1]
```
It looks like `j` is unused?

qutip/core/data/ode.pyx
```diff
@@ -0,0 +1,214 @@
...
+    name='wrmn_error',
+    module=__name__,
+    inputs=('diff', 'state'),
+    out=False,
+)
+wrmn_error.__doc__ =\
+    """
+    Compute the weighted root mean square norm for the ode error estimation
+
+    error = sqrt( 1/N * sum( ( |diff[i]| / (atol + rtol * |state[i]|))**2 ))
```
hodgestar:
```suggestion
    Compute the weighted root mean square error norm:

        error = sqrt( 1/N * sum( ( |diff[i]| / (atol + rtol * |state[i]|))**2 ))
        
    This error norm is commonly used to estimate error in ODEs.
```

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -520,12 +516,10 @@ cdef class Explicit_RungeKutta:
...
-        print("t_prev, y_prev, |y_prev|", self._t_prev,
-              'None' if self._y_prev is None else self._y_prev.to_array(),
-              self._norm_prev)
-        print("t_front, y_front, |y_front|", self._t_front,
-              'None' if self._y_front is None else self._y_front.to_array(),
-              self._norm_front)
+        print("t_prev, y_prev", self._t_prev,
+              'None' if self._y_prev is None else self._y_prev.to_array())
+        print("t_front, y_front", self._t_front,
+              'None' if self._y_front is None else self._y_front.to_array())
```
hodgestar, line 522: It would be nice to still print the error norm here for debugging.

qutip/tests/core/data/test_mathematics.py
```diff
@@ -1085,3 +1085,52 @@ def op_numpy(self, matrix):
...
+        """
+        left, right = data_l(), data_r()
+        expected = self.op_numpy(left.to_array(), right.to_array(), atol, rtol)
+        test = op(left, right, atol, rtol)
+
+        assert isinstance(test, out_type)
+        if issubclass(out_type, Data):
+            assert test.shape == expected.shape
+            np.testing.assert_allclose(test.to_array(), expected,
+                                       atol=self.atol, rtol=self.rtol)
```
hodgestar: Can the output of `wrmn_error` ever be of type `Data`? If not, let's delete this and add an assert that it is a float, perhaps.

pmenczel:
@Ericgig This PR has broken the HEOM tutorial notebook 2: https://github.com/qutip/qutip-tutorials/blob/main/tutorials-v5/heom/heom-2-fmo-example.md (reverting to https://github.com/qutip/qutip/commit/c2551f81b933b08bb535284ca7f2e37302080cc2 the notebook runs)

When calling `run` on the HEOM solver, we get
> IntegratorException: Step size becomes too small. Try increasing tolerance.

I am not sure how to fix the notebook. Increasing the `rtol` and `atol` that were set to `1e-18` in the notebook all the way to `1` still results in the same error. Also, I am worried that users might have similar problems after this change.

Ericgig:
I will look at it.

Was the tolerance really set to `1e-18`? Not a typo? 
The coefficient for the ode have less precision than that.
If it really worked with that, it should have been enough to show us that the error computation were bugged...

Ericgig:
The operator (HEOM liouvillian?) has most values of order `1e12`, but the highest at around `1e27`.
With `dt` minimum at `1e-18` the high values did not cancel enough to fit within tolerance.

The best would be to set units to have values of order 1 and keep default options.
But just setting `min_step` to `1e-20` is enough. It still runs in less than 20sec with average steps of `1e-15`

pmenczel:
Hi, thanks for having a look. Sorry, the tolerance was `1e-12` not `1e-18`, typo.

I have changed the notebook to use a bit more sane units and default values for the tolerances and min step, it seems fine now.

PR #2730, opened 2025-08-18
QS Library: Adding The Spin Chain Model

Vanshaj0429:
### Description  
This PR adds a **Linear Spin Chain model** to the new `quantum_systems` module.  
The model is implemented as a function in `linear_spin_chain.py` that builds the Hamiltonian, operators, collapse and latex representation for a nearest-neighbor spin chain with configurable length, coupling, and boundary conditions.  

### Implementation details  
- New file: `qutip/quantum_systems/linear_spin_chain.py`  
- Function-based implementation of the linear spin chain Hamiltonian  
- Integrates with the `QuantumSystem` framework introduced in PR #2708  
- Ready for test coverage and further extensions (open vs periodic chains, anisotropic couplings, etc.)  

### Dependency on PR #2708  
⚠️ **Important:** This PR depends on [#2708](https://github.com/qutip/qutip/pull/2708).  
That PR introduces the `QuantumSystem` base class and updates `__init__.py`, which this spin chain function relies on.  

Please merge #2708 first — once it is merged, this PR will only contain the new `linear_spin_chain.py` additions.  

### Next steps  
- Add tests under `qutip/tests/quantum_systems/test_linear_spin_chain.py`


Vanshaj0429:
@ajgpitch, @Ericgig and @nwlambert, opened this PR. I have one minor correction I have in my mind (will correct it asap). Could you please review the code.

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,311 @@
...
+    transition_frequency: float,
+) -> list:
+    """Build collapse operators for open system dynamics"""
+    c_ops = []
+
+    # Local dissipation on each site
+    for k in range(N):
+
+        # Spontaneous emission (relaxation)
+        if gamma_relaxation > 0.0:
```
nwlambert: just a quick comment, i think i mentioned it seems a bit overly complex to have both gamma_relaxation and gamma_thermal. I think simply supporting T=0 in gamma_thermal, so that purely spontaneous emission can be included, is enough.    supporting both means the spin-chain can be coupled to both a T=0 (spontaneous emission) and T!=0 bath simultaneously, which is fine as a model in some sense, but kinda overkill for what we need here (and maybe a bit misleading to the user)

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,311 @@
...
+    QuantumSystem
+        Configured linear spin chain system instance
+    """
+
+    # Validate inputs
+    if N < 2:
+        raise ValueError("Chain length N must be at least 2")
+    if N > 15:
+        print(
+            f"Warning: Large system size N={N}. Hilbert space dimension = 2^{N} = {2**N}"
```
nwlambert: i think this warning isn't necessary

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,311 @@
...
+    Create linear spin chain system
+
+    Creates a 1D chain of spin-1/2 particles with nearest-neighbor interactions.
+    Supports various spin models and open system dynamics.
+
+    Model Hamiltonians:
+    - Heisenberg: H = J * sum_i [S_i^x * S_{i+1}^x + S_i^y * S_{i+1}^y + S_i^z * S_{i+1}^z]
+    - XXZ: H = J * sum_i [S_i^x * S_{i+1}^x + S_i^y * S_{i+1}^y] + Jz * sum_i [S_i^z * S_{i+1}^z]
+    - XY: H = J * sum_i [S_i^x * S_{i+1}^x + S_i^y * S_{i+1}^y]
+    - Ising: H = J * sum_i [S_i^z * S_{i+1}^z]
```
nwlambert: mention the external B-fields here, and also that that allows transverse ising model?

Vanshaj0429:
@nwlambert, I have put the tests, can you review the last two commits.

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,339 @@
+import numpy as np
+from qutip import tensor, qeye, sigmax, sigmay, sigmaz, coefficient, Qobj, sigmap, sigmam
+from qutip.core.cy.coefficient import Coefficient
+from typing import Union
+from .quantum_system import QuantumSystem
+
+def _create_sqrt_coefficient(rate):
```
nwlambert: since the same function is used in jaynes_cummings, could import from there?

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,339 @@
...
+        Coupling strength for XY interactions (Sx*Sx + Sy*Sy terms)
+    Jz : float or coefficient, default=None
+        Z-coupling strength (Sz*Sz terms). If None, equals J for Heisenberg, 0 for XY
+    boundary_conditions : str, default="open"
+        Boundary conditions: "open" or "periodic"
+    B_x, B_y, B_z : float or coefficient, default=0.0
+        External magnetic field components
+    gamma_dephasing : float or coefficient, default=0.0
+        Pure dephasing rate (T2* process)
+    gamma_depolarizing : float or coefficient, default=0.0
```
nwlambert, line 107: an explicit latex description of all the noise channels would be useful (dephasing, polarizing, and thermal)

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,339 @@
...
+                c_ops.append(sqrt_thermal_up * operators[f"S_{k}_plus"])
+            else:
+                # T=0 case: pure spontaneous emission (only downward transitions)
+                sqrt_gamma_thermal = _create_sqrt_coefficient(gamma_thermal)
+                c_ops.append(sqrt_gamma_thermal * operators[f"S_{k}_minus"])
+
+    return c_ops
+
+
+def _generate_latex(
```
nwlambert, line 343: could we also print the noise lindblads here too? (this case is a bit more complex than JC, with dephasing/depolarizing, so printing what is used would be cool)

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,339 @@
...
+            c_ops.append(sqrt_depol_rate * operators[f"S_{k}_z"])
+
+        # Thermal bath coupling
+        if isinstance(gamma_thermal, Coefficient) or gamma_thermal > 0.0:
+            if temperature > 0.0:
+                # Finite temperature case
+                beta = 1.0 / temperature
+                exp_factor = np.exp(-beta * transition_frequency)
+                p_down = 1.0 / (1.0 + exp_factor)
+                p_up = exp_factor / (1.0 + exp_factor)
```
nwlambert, line 315: just to note, these are the statistics (and hence emission/absorption rates) for a spin-bath? This should also be explicitly mentioned in the doc strings/documentation.

nwlambert (changes requested):

qutip/quantum_systems/linear_spin_chain.py
nwlambert: just to clarify a bit, for a spin-system we can imagine it couples to a bath of spins or a bath of bosons.  the choice leads to slightly different emission/absorption rates since bosons follow bose-einstein statistics, but spin bath will follow fermi-dirac statistics, which is what you do here.  ok just to support the latter here, for simplicity, but just needs to be documented.

nwlambert:
The tests look ok to me.  the final one testing the coefficients could be made a bit more useful?  But am not sure how!  (just fyi, The failing tests are not to do with this PR I think, seems to be a coveralls issue)

qutip/quantum_systems/linear_spin_chain.py
Vanshaj0429: Yes, I could do that

qutip/quantum_systems/linear_spin_chain.py
Vanshaj0429: Yes, makes sense.

qutip/quantum_systems/linear_spin_chain.py
Vanshaj0429: Screenshot 2025-09-22 at 7 09 52 PM

I can do something like this

qutip/quantum_systems/linear_spin_chain.py
Vanshaj0429: Yes, I have mentioned this explicitly in docstring.

qutip/quantum_systems/linear_spin_chain.py
nwlambert: seems reasonable to me

qutip/quantum_systems/linear_spin_chain.py
ajgpitch: Move to `quantum_system.py`

Vanshaj0429:
@Ericgig can you review this please, I have added more detail in the docstring to make it more readable and moved the `_create_sqrt_coefficient` function to `quantum_system.py`. I think these were two prominent issues.

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,409 @@
...
+        temperature=temperature,
+        transition_frequency=transition_frequency,
+    )
+
+
+def _create_site_operator(N: int, site: int, single_op: Qobj) -> Qobj:
+    """Create operator acting on specific site in N-spin chain"""
+    op_list = [qeye(2)] * N
+    op_list[site] = single_op
+    return tensor(op_list)
```
Ericgig, line 226: This already exist in qutip: `qutip.core.tensor.expand_operator`.

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,409 @@
...
+    boundary_conditions = boundary_conditions.lower()
+
+    # Set default Jz values
+    if Jz is None:
+        if model_type == "heisenberg":
+            Jz = J
+        elif model_type == "xy":
+            Jz = 0.0
+        else:  # xxz, ising
+            Jz = J
```
Ericgig, line 146: This could simply be `Jz = J`.
With "xy", `Jz` is not used in create hamiltonian.

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,409 @@
...
+    # Determine number of interaction terms
+    num_interactions = N if boundary_conditions == "periodic" else N - 1
+
+    # Interaction Hamiltonian
+    H_interaction = 0
+
+    for k in range(num_interactions):
+        k_next = (k + 1) % N
+
+        if model_type in ["heisenberg", "xxz", "xy"]:
```
Ericgig, line 251: How is "heisenberg" different from "xxz"?
"heisenberg" with custom "Jz" is "xxz", "xxz" without custom "Jz" is "heisenberg".

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,409 @@
...
+    return c_ops
+
+
+def _generate_latex(
+    model_type: str,
+    J: float,
+    Jz: float,
+    B_x: float,
+    B_y: float,
+    B_z: float,
```
Ericgig, line 349: Type hints are inconsistent.

qutip/quantum_systems/linear_spin_chain.py
```diff
@@ -0,0 +1,409 @@
...
+    operators["S_y_total"] = sum(operators[f"S_{k}_y"] for k in range(N))
+    operators["S_z_total"] = sum(operators[f"S_{k}_z"] for k in range(N))
+    operators["S_plus_total"] = sum(operators[f"S_{k}_plus"] for k in range(N))
+    operators["S_minus_total"] = sum(operators[f"S_{k}_minus"] for k in range(N))
+
+    # Magnetization (alias for S_z_total)
+    operators["magnetization"] = operators["S_z_total"]
+
+    # Nearest-neighbor correlation operators
+    operators["correlation_xx_nn"] = sum(
```
Ericgig, line 171: Why no `yy` operators?
Let's use them in `_build_hamiltonian`.

qutip/tests/quantum_systems/test_linear_spin_chain.py
```diff
@@ -0,0 +1,227 @@
...
+    @pytest.mark.parametrize("boundary_conditions", ["open", "periodic"])
+    def test_boundary_conditions(self, boundary_conditions):
+        """Test different boundary conditions"""
+        lsc = linear_spin_chain(N=3, boundary_conditions=boundary_conditions)
+        
+        assert lsc.parameters["boundary_conditions"] == boundary_conditions
+        
+        # Both should create valid systems
+        assert lsc.hamiltonian is not None
+        assert lsc.hamiltonian.shape[0] == 8  # 2^3 for 3 spins
```
Ericgig, line 57: It would be good to have a test that check that all periodic operator are so.
I think `<101| H |000>` would be zeros if open and not if periodic. (B = 0)

qutip/tests/quantum_systems/test_linear_spin_chain.py
```diff
@@ -0,0 +1,227 @@
...
+    """Test suite for linear_spin_chain function"""
+
+    def test_default_heisenberg_chain(self):
+        """Test default Heisenberg chain creation"""
+        lsc = linear_spin_chain()
+        
+        assert isinstance(lsc, QuantumSystem)
+        assert lsc.name == "Linear Spin Chain (HEISENBERG)"
+        assert lsc.parameters["model_type"] == "heisenberg"
+        assert lsc.parameters["N"] == 4
```
Ericgig, line 18: These are strange default...

qutip/tests/quantum_systems/test_linear_spin_chain.py
```diff
@@ -0,0 +1,227 @@
...
+        
+        # XY model: only X and Y interactions
+        H_expected = 0
+        for k in range(N - 1):
+            H_expected += J * (ops[f"S_{k}_x"] * ops[f"S_{k+1}_x"] +
+                             ops[f"S_{k}_y"] * ops[f"S_{k+1}_y"])
+        
+        assert lsc.hamiltonian == H_expected
+
+    def test_ising_model_hamiltonian(self):
```
Ericgig, line 94: Test for xxz missing.

Ericgig (commented):
I don't think `linear_spin_chain()` should not be valid without any arguments.
At least let's make `N`, `model` and `J` required...

veronikakurth:
@Vanshaj0429 thank you for your work! Please let us know whether you have an intention and time address the rest of the code review. Since this PR has been inactive for more than 6 months, we will have to close it otherwise.

Vanshaj0429:
> @Vanshaj0429 thank you for your work! Please let us know whether you have an intention and time address the rest of the code review. Since this PR has been inactive for more than 6 months, we will have to close it otherwise.

Hi @veronikakurth, apologies for not responding sooner and for the prolonged inactivity on this PR.

In early November I came down with a respiratory infection that escalated into a serious illness, and I had to stop all work and take leave for medical treatment. I only started working again in late February, and even then had other pressing commitments to catch up on first. The situation unfortunately cost me a number of opportunities, and this PR was one of the things I wasn't able to get back to as quickly as I'd have liked.

Thank you for taking the time to flag this rather than simply closing it. I want to see this work through to completion and am happy to address the remaining review comments and should be able to hopefuly complete this in a week's time.

PR #2740, opened 2025-08-27
merge setup.cfg into pyproject.toml

Ericgig:
**Description**

Modern python package have all the metadata in the `pyproject.toml` instead of having it shared between it and the setup.cfg.
I merged the files keeping only the pyproject.


pyproject.toml
```diff
@@ -3,15 +3,79 @@ requires = [
...
+  {name = "Shahnawaz Ahmed"},
+  {name = "Neill Lambert"},
+  {name = "Eric Giguère"},
+  {name = "Boxi Li"},
+  {name = "Jake Lishman"},
+  {name = "Simon Cross"},
+  {email = "qutip-admin@googlegroups.com"},
+]
+classifiers = [
+    "Development Status :: 2 - Pre-Alpha",
```
hodgestar, line 36: Not for this PR, but we should change this. QuTiP is definitely stable now. :)

pyproject.toml
```diff
@@ -3,15 +3,79 @@ requires = [
...
+    "Intended Audience :: Science/Research",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Topic :: Scientific/Engineering",
+    "Operating System :: MacOS",
+    "Operating System :: POSIX",
+    "Operating System :: Unix",
+    "Operating System :: Microsoft :: Windows",
```
hodgestar, line 44: We should keep the license classifier `License :: OSI Approved :: BSD License`.

pyproject.toml
```diff
@@ -3,15 +3,79 @@ requires = [
...
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "qutip"
+dynamic = ["version"]
+description = "QuTiP: The Quantum Toolbox in Python"
+readme = "README.md"
+requires-python = ">=3.10"
+license = {file = "LICENSE.txt"}
```
hodgestar: Is there a way to also include the license name here directly? It's good to have something machine / human readable in addition to just a link to the text of the license.

pyproject.toml
```diff
@@ -3,15 +3,79 @@ requires = [
...
+  {name = "Arne Grimsmo"},
+  {name = "Nathan Shammah"},
+  {name = "Shahnawaz Ahmed"},
+  {name = "Neill Lambert"},
+  {name = "Eric Giguère"},
+  {name = "Boxi Li"},
+  {name = "Jake Lishman"},
+  {name = "Simon Cross"},
+  {email = "qutip-admin@googlegroups.com"},
+]
```
hodgestar, line 34: Is there a way to directly specify an email for the project? Having the project email address only be in the list of the authors seems not ideal.

pyproject.toml
Ericgig: I got a warning that it's deprecated when building when it's included.

pyproject.toml
Ericgig: All release branches are `Development Status :: 5 - Production/Stable`
But master is not meant to be released so stays this way.

pyproject.toml
Ericgig: Yes it seems strange to me too, but I got an error when I included both the file and the name.

pyproject.toml
Ericgig: I have not found one.
There is no project email field.
For the authors, we can have either name, email or both. So this match what we add in setup.cfg the best.

The project urls sections seems to accept any tags, so maybe we could add it here as a mailto link.

pyproject.toml
hodgestar: Which program generated the warning and what did it say? It's still listed at https://pypi.org/classifiers/.

hodgestar: Ah, I see the recommendation is to specify an SPDX license directly in the license field. Let me resolve this thread.

pyproject.toml
hodgestar: Ah, the recommendation is to use just the SPDX license name here. See https://spdx.org/licenses/ and https://packaging.python.org/en/latest/specifications/license-expression/. Perhaps just:
```
license = "BSD-3-Clause"
```

pyproject.toml
hodgestar: Having read the pyproject.toml guide, I think this looks good. Thanks.

pyproject.toml
```diff
@@ -3,15 +3,79 @@ requires = [
...
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "qutip"
+dynamic = ["version"]
+description = "QuTiP: The Quantum Toolbox in Python"
+readme = "README.md"
+requires-python = ">=3.10"
+license = {name = "BSD-3-Clause"}
```
hodgestar:
```suggestion
license = "BSD-3-Clause"
```
This should just be a string.

hodgestar:
I found references to `setup.cfg` in:

- doc/development/release_distribution.rst
- doc/development/contributing.rst
- setup.py

Would you mind updating those too?

doc/development/release_distribution.rst
```diff
@@ -57,7 +57,9 @@ We follow `NEP29`_ for minimum supported versions ::
     - All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
     - All minor versions of numpy and scipy released in the 24 months prior to the project, and at minimum the last three minor versions.
 
-If dependency versions need to be updated, update them in the master branch. The following files may need to be updated: `.github/workflows/tests.yml`, `setup.cfg` and `roadmap.rst`. Finally, ensure that PyPI wheels and conda builds cover at least these versions.
```
hodgestar, line 60: It looks like here `setup.cfg` should just be replaced with `pyproject.toml`.

doc/development/release_distribution.rst
```diff
@@ -57,7 +57,9 @@ We follow `NEP29`_ for minimum supported versions ::
...
+If dependency versions need to be updated, update them in the master branch.
+The following files may need to be updated: `.github/workflows/tests.yml` and `roadmap.rst`.
```
Ericgig:
```suggestion
The following files may need to be updated: `.github/workflows/tests.yml`, `pyproject.toml` and `roadmap.rst`.
```

pyproject.toml
```diff
@@ -3,15 +3,79 @@ requires = [
...
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "qutip"
+dynamic = ["version"]
+description = "QuTiP: The Quantum Toolbox in Python"
+readme = "README.md"
+requires-python = ">=3.10"
+license = {text = "BSD-3-Clause"}
```
hodgestar:
```suggestion
license = "BSD-3-Clause"
```
Setting `text = ...` specifies that the value is the full license document. We want exactly the suggestion above. Does the suggestion not work for some reason?

Ericgig: This is only valid with recent setuptools version, (I still used an older version without realising.)

Normally we say we support dependencies for 2 years...
It says it needs at minimum 77.0.3 which was released in March this year. Maybe it doesn't matter since it's not a runtime dependencies...

hodgestar: Ah, that is annoying. Maybe let's go back to `license = {file = ...}` then as more accurate than `text=`?

Ericgig: Now that I understand better, I prefer the new method. We won't have to touch it for longer.

Build time dependencies are not used when downloading from pip and when installing from source, unless you use `--no-build-isolation`, it will update it. So we are probably the only one who will be affected.

And we already depend on numpy 2.0.0 at build time.

hodgestar (approved):

hodgestar:
Other than the failing documentation build, looks good.

PR #2742, opened 2025-09-02
Better mkl finding for recent numpy

Ericgig:
**Description**
As @pmenczel found, `numpy.__config__` is mostly empty (all private) for recent numpy version.
Instead `numpy.show_config` can return a dict instead of printing and it has the needed information . (new with numpy 1.26)

Also, numpy and scipy can uses different blas libraries.

Updated `_blas_info` to read from `scipy.show_config` when available. It's used to detect the implementation of the eigen solvers, so the scipy inplementation is better.

Update `_find_mkl` to look at the blas/lapack implementation used by scipy, numpy then anaconda default location.

### DIFF
--- a/qutip/settings.py
+++ b/qutip/settings.py
@@ -8,12 +8,13 @@ from ctypes import cdll, CDLL
 import platform
 from glob import glob
 import numpy as np
+import scipy
 
 __all__ = ['settings']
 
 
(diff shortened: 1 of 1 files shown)


pmenczel (approved):
Thanks for the quick fix!

Issue #2659, opened 2025-03-21
Discussion: Plans for local multiplication


Sola85:
### Problem Description

I'm opening this issue to restart the discussion on possible directions for better support for local multiplications.
This topic was already brought up in https://github.com/qutip/qutip/pull/1405, but not persued any further due to licence issues and because Qutip 5 was in the development. Since then, 4 years have passed and I think it makes sense to re-open this discussion.

Background:
It is quite common for "large" states (be it kets or density matrices) to be acted upon by "small" channels, i.e. local operators or super-operators.
The currently preferred approach within Qutip seams to be to expand local operations to global operations using `tensor()` or `expand_operator()` such that the states and the channels have compatible dimension and then to perform a simple matrix multiply. 
This approach is inefficient because the intermediate channels are unnecessarily large and hence the subsequent matrix-multiply is unnecessarily ressource intensive.

I recently discovered `subsystem_apply()` which intends to solve this issue, however, this module too has its shortcomings:
- It is undocumented (despite using qutip for many years I only recently learned about it)
- It is not well tested (`subsystem_apply(ket, operator)` fails but none of the existing tests reveal this)
- The core operations that `subsystem_apply()` relies on are implemented in Python, making it just barely faster than `expand_operator()` followed by matrix-multiply.
- Its API is not very flexible. The `mask` parameter is nice, because it allows application of a single channel to multiple dimensions, but does not work with channels that act on more than one dimension. 
 

### Proposed Solution

I think there are two separate topics here:

### API
One possible improvement would be an API-change to e.g.: 
```subsystem_apply(state: Qobj, operators: list[Qobj], targets: list[list[int]])``` 
This would allow e.g. `subsystem_apply(state, [my_two_qubit_channel, my_single_qubit_channel], [[0, 1], [4]])`, which could represent a joint action of `my_two_qubit_channel` on dimensions 0 and 1, as well as `my_single_qubit_channel` on dimension 4. 
This would:
- still allow acting with the same channel on multiple sites (just like the exising `mask` parameter), by passing an operator multiple times
- enable multi-qudit channels to work
- be an interface similar to that of `expand_operator`.

### Implementation
The three relevant operations are "`local_channel*operator`", "`local_operator*operator`" and "`local_operator*vector`". For these I see a few options:
[...]

PR #2743, opened 2025-09-05
Local matmul

Ericgig:
**Description**
Add local multiplication at the cython level.

Add the function `local_matmul(operator, state, modes, dual=False)` which apply the operator on the listed mode of the state. 

`local_matmul(gates.cnot(), rand_ket([2] * 10), [2, 4])` 
is equivalent to 
`expend_operator(gates.cnot(), [2] * 10, [2, 4]) @ rand_ket([2] * 10)`

This does `operator @ state` product directly. If state is a dm, it does not apply it on the dual side with the same call.
The `dual` keyword reverse the order of operations: `state @ operator`.
Super operator on dm are applied on both side at once.

`local_matmul(sprepost(sigmax(), sigmax()), rand_dm([2] * 10), 2)` => 
`expend_operator(sigmax(), [2] * 10, [2]) @ rand_dm([2] * 10) @ expend_operator(sigmax(), [2] * 10, [2])`

This function act on Qobj.

There is also `target_mode_matmul(operator, state, modes, hilbert, hilbert_out, ...)` which is a dispatched function to act on the date layer.
That function can also transpose / conjugate the operator before applying it.


Ericgig:
Fix #2659 ?

qutip/core/local_matmul.py
```diff
@@ -0,0 +1,165 @@
...
+) -> Qobj:
+    """
+    Applies an operator to specific modes of a quantum state or operator.
+
+    This function performs a tensor product operation where a smaller
+    `operator` acts only on the specified `modes` of a larger `state`.
+
+    Super operator are applied on the same mode on both side of a density
+    matrix at once. The state must be a density matrix
+    (operator-ket are not supported).
```
hodgestar, line 22:
```suggestion
    This function performs a tensor product operation where a smaller
    `operator` acts only on the specified `modes` of a larger `state` (or density matrix).

    Super operators are supported but must be applied to a density matrix. Apply a super
    operator to an operator-ket is not supported. The pre- and post- actions of the super
    operator are applied to the same modes.
```

qutip/core/local_matmul.py
```diff
@@ -0,0 +1,165 @@
...
+        The indices specifying the modes on which the operator should act.
+
+    dual: bool, default: False
+        If True, the operator is applied from the right (state @ operator).
+        Ignored for superoperators.
+
+    Returns
+    -------
+    Qobj
+        Updated state as a Qobj.
```
hodgestar, line 43:
```suggestion
        The resulting state (or density matrix) as a Qobj.
```

qutip/core/local_matmul.py
```diff
@@ -0,0 +1,165 @@
...
+
```
hodgestar, line 44:
```suggestion

    Raises
    ------
    ValueError:
       When a target mode index is out of bounds for the state's dimensions, or the
       operator's input dimensions do not match the state's dimensions on the specified modes.
    TypeError:
        When the operator's input dimensions do not match the state's dimensions on the
        specified modes.
    NotImplementedError:
        When `local_matmul` is performed on a dimension that is not a tensor product space (e.g. an `ENR` state or a direct sum).
```

qutip/core/local_matmul.py
```diff
@@ -0,0 +1,165 @@
...
+        modes = [modes]
+
+    #TODO: Exotic dims could be supported if shared
+    operator._dims._require_pure_dims("local_matmul")
+    state._dims._require_pure_dims("local_matmul")
+    if not all(mode < N for mode in modes):
+        raise ValueError(
+            "Target mode index is out of bounds for the state's dimensions."
+        )
+    if not len(operator.dims[1]) == len(modes):
```
hodgestar, line 84: Is `operator.dims[1]` the right dimension when `dual=True`? Perhaps is should be `hilbert_oper_in`?

qutip/core/local_matmul.py
```diff
@@ -0,0 +1,165 @@
...
+
+    state: Qobj
+        The quantum state to be acted upon.
+
+    modes: int | list[int]
+        The indices specifying the modes on which the operator should act.
+
+    dual: bool, default: False
+        If True, the operator is applied from the right (state @ operator).
+        Ignored for superoperators.
```
hodgestar, line 38: Should we raise an error if `dual=True` for superoperators rather than ignoring it?

qutip/core/data/local_matmul.py
```diff
@@ -0,0 +1,247 @@
...
+    hilbert: list[int],
+    new_hilbert: list[int],
+    dual: bool = False,
+    transpose: bool = False,
+    conj: bool = False,
+) -> Dense:
+    """
+    Apply the operator to the target modes.
+
+    Roughtly equivalent to:
```
hodgestar, line 31:
```suggestion
    Roughly equivalent to:
```

qutip/core/data/local_matmul.py
```diff
@@ -0,0 +1,247 @@
...
+
+        expend_operator(operator, hilbert, modes) @ state
```
hodgestar, line 33:
```suggestion
        expand_operator(operator, hilbert, modes) @ state
```

qutip/core/data/local_matmul.py
```diff
@@ -0,0 +1,247 @@
...
+        out = one_mode_matmul_dual_dense_data(state, operator, hilbert, modes[0])
+    elif not dual:
+        out = n_mode_matmul_data_dense(
+            operator, state, hilbert, modes, new_hilbert, transpose, conj
+        )
+    else:
+        # For product from the right, we take the expend the hilbert space to
+        # includ the first index as an extra space.
+        # dims = [[2, 3], [5, 7]] -> hilbert [6, 5, 7] (C order)
+        # then contract with the operator transposed.
```
hodgestar, line 80:
```suggestion
        # For a product from the right, we take the expanded hilbert space to
        # include the first index as an extra space, e.g.
        # dims = [[2, 3], [5, 7]] -> hilbert [6, 5, 7] (C order),
        # and then contract with the transposed operator.
```

qutip/core/data/_local_matmul.pyx
```diff
@@ -0,0 +1,419 @@
...
+cdef Dense _null_dense(int rows, int cols, bint fortran=True):
+    cdef Dense out = Dense.__new__(Dense)
+    out.shape = (rows, cols)
+    out.data = <double complex *> 0
+    out._deallocate = False
+    out.fortran = fortran
+    return out
+
+
+cdef void _cut_dense(Dense base, Dense out, int loc):
```
hodgestar, line 73: Should we rename this to `_view_sliced_dense`?

qutip/core/data/_local_matmul.pyx
```diff
@@ -0,0 +1,419 @@
...
+
+    def __init__(self, hilbert, mode, hilbert_out=None):
+        if hilbert_out is None:
+            hilbert_out = hilbert
+        square_oper = hilbert == hilbert_out
+
+        if not square_oper:
+            if len(hilbert) != len(hilbert_out):
+                raise ValueError(
+                    "The number of hilbert subspaces the must stay constant."
```
hodgestar, line 226:
```suggestion
                    "The number of hilbert subspaces must stay constant."
```

qutip/core/local_matmul.py
```diff
@@ -0,0 +1,165 @@
...
+    (operator-ket are not supported).
+
+    Parameters
+    ----------
+    operator: Qobj
+        The operator to apply. Its Hilbert space dimensions must correspond to
+        the dimensions of the state on the targeted modes.
+
+    state: Qobj
+        The quantum state to be acted upon.
```
hodgestar, line 31: Let's make a note that the `state` is always converted to `Dense`.

qutip/core/local_matmul.py
Ericgig: What would be a good term for anything that can be a state in a solver?
For me "state" include ket, density matrix, operators (qeye for propagator), batched kets, etc. more general than operators. The only restriction is for it to be dense and matching dimensions.
"state or density matrix" makes it seems as if state stand for ket, not pretty much any Qobj.

hodgestar: What about `Updated solver state as a Qobj`?

hodgestar (approved):

qutip/core/local_matmul.py
Ericgig:
```suggestion
        Updated solver state as a Qobj
```

hodgestar (approved):

PR #2749, opened 2025-09-22
Correct type hint for __mul__ of Qobj

hbeukers:
**Description**
Correct the type hint of `__mul__` of the `Qobj` so that type checking is also correct for multiplying two `Qobj`s.

**Changelog**
Improve type hint of the `__mul__` method of `Qobj` (#2749, by hbeukers)

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -413,7 +413,7 @@ class Qobj:
     def __rsub__(self, other: Qobj | complex) -> Qobj:
         return self.__neg__().__add__(other)
 
-    def __mul__(self, other: complex) -> Qobj:
+    def __mul__(self, other: Qobj | complex) -> Qobj:
         """
         If other is a Qobj, we dispatch to __matmul__. If not, we
         check that other is a valid complex scalar, i.e., we can do


Ericgig:
It was a conscious choice not the include these hints.
We want to encourage the use of `@` for Qobj product.

hbeukers:
Thank you for your reply, I now also found the rationale in the documentation. I understand the preference fore the `@` symbol for these matrix multiplications. However, I find it confusing that type hints are used for this. Using a type checker plugin will give some of the following output:

image

In this case it is pyright, but also mypy will give an error message.

Would it be an idea to move the encouragement of using `@` to the docstring of the `__mul__` method as this clearly shows up by hovering over the `*` symbol? 

This would both give a clearer message for the encouragement (I now missed it for example) and reduce false positives when using type checkers on working qutip code.

Ericgig:
Yes, improving the docstring would be useful to decrease confusion.
Do you want to have a go at it?

hbeukers:
Yes, I'll update the pull request. Rereading my previous comment I realize that it is not so clear that I suggest to still update the type hint to be `Qobj | complex`. This will make type checkers not error unnecessarily. The updated docstring will cover the preferred behavior. Do you agree with this combination?

Ericgig:
Thank you for reporting the issue.
Replaced with #2888, which fix the docstring, not the hints.

PR #2751, opened 2025-09-24
Dimensions update, make ENR space objects work in more places

pmenczel:
I tried to find all places where old list-style `dims` are used and either replace them with `_dims` or, where the replacement is not straightforward, add a consistent error message if such functions are called with "impure" dimensions.

The only "impure" dims existing currently are ENR spaces. However, this PR is a bit of a preparation for going back to my "direct sums" PR, which would add a second type of impure dims. In any case, ENR space objects should now work in a lot more places. I added test cases with ENR spaces for some of the modules that should work now.

Since the PR touches many files, I try to briefly summarize the additional changes:
* The `sprepost` function was doing a weird thing, dropping all "1"s from the dims, but only for Qobj and not for QobjEvo. This seems like an unexpected side effect of `sprepost` and I have removed it. Instead, Qobj has a new method `drop_scalar_dims` which drops "1"s from the dims. (It is similar to `contract` but not the same.)
* One test in test_transfertensor and one piece of code in the user guide were relying on that behavior, and I have added calls to the new `drop_scalar_dims` method there.
* I moved the functionality of the non-public functions `collapse_dims_oper`, `collapse_dims_super` etc. in dimensions.py to the Dimensions class
* Solvers now have `sys_dims` (list) and `_sys_dims` (Dimensions)
* I added `einsum` to the apidoc
* Minor bugfixes throughout


qutip/core/superop_reps.py
```diff
@@ -205,12 +213,18 @@ def _super_tofrom_choi(q_oper):
...
     data = q_oper.data.to_array()
-    dims = q_oper.dims
-    new_dims = [[dims[1][1], dims[0][1]], [dims[1][0], dims[0][0]]]
-    d0 = np.prod(flatten(new_dims[0]))
-    d1 = np.prod(flatten(new_dims[1]))
-    s0 = np.prod(dims[0][0])
-    s1 = np.prod(dims[1][1])
+    dims_from = q_oper._dims[1].oper
+    dims_to = q_oper._dims[0].oper
+    outrep = 'super' if q_oper.superrep == 'choi' else 'choi'
```
Ericgig, line 218: Since the superrep is set here, we can remove it from the Qobj initialization.

qutip/core/superoperator.py
```diff
@@ -432,17 +423,21 @@ def sprepost(A, B):
     super : Qobj or QobjEvo
         Superoperator formed from input quantum objects.
     """
+    if not A.isoper or not B.isoper:
+        raise TypeError('Input is not a quantum operator')
+
```
Ericgig: Why add this condition? 
While rectangular super operator and not well supported, they are not forbidden.

doc/guide/guide-super.rst
```diff
@@ -83,6 +83,7 @@ Meanwhile, the :func:`.super_tensor` function implements the swap on the right,
 
    q = tensor(identity(2), basis(2))
    s_prep = sprepost(q, q.dag())
+   s_prep.drop_scalar_dims(inplace=True)  # clean up dims
```
Ericgig: This could confuse users if not explained somewhere.

qutip/solver/solver_base.py
```diff
@@ -284,10 +284,17 @@ def sys_dims(self):
         """
         Dimensions of the space that the system use:
 
-        ``qutip.basis(sovler.dims)`` will create a state with proper dimensions
+        ``qutip.basis(solver.dims)`` will create a state with proper dimensions
```
Ericgig: Let's fix the other error too:
```suggestion
        ``qutip.basis(solver.sys_dims)`` will create a state with proper dimensions
```

And maybe add a note it's not sufficient for exotic dimensions?

Ericgig (commented):
Looked at half of it, more to come tomorrow.

Have you tried to change `Qobj.dims` to return a dimension object?
I think it would break a lot of stuff, but never tried it.

qutip/core/superoperator.py
pmenczel: Ah, I added it only to make it match `spre` and `spost`. I will remove it again, since it is not necessary here.

I had never thought about it too much: the issue is that the dimensions of `spre` and `spost` are not actually fully determined by the dimensions of the operator `A` alone. (If A is m x n, then spre(A) could act on any n x k operator for arbitrary k.) That is why we require all operators to be square with these functions. I am adding little notes to the docs of `spre` and `spost`.

Note that `sprepost` will still fail with non-square QobjEvo.

doc/guide/guide-super.rst
pmenczel: I have added more explanation in the comment. Do you think there should be more explanation somewhere else?

pmenczel:
Thank you for the review.

> Have you tried to change `Qobj.dims` to return a dimension object? I think it would break a lot of stuff, but never tried it.

I did not try it... in this PR, I often needed to do more than just change `dims` to `_dims`, so I think it would have broken a lot of stuff indeed. After this PR, it should be easier to replace `_dims` back with `dims` if we ever want to do that. But we'd have to think about if that can be confusing for the user in some cases.

qutip/measurement.py
```diff
@@ -24,11 +24,11 @@ def _verify_input(op, state):
     if not isinstance(state, Qobj):
         raise TypeError("state must be a Qobj")
     if state.isket:
-        if op.dims[-1] != state.dims[0]:
+        if op._dims.from_ != state._dims.to_:
```
Ericgig, line 27:
```suggestion
        if not op._dims.issquare:
```

qutip/tests/test_entropy.py
```diff
@@ -177,6 +178,17 @@ def test_random_rho_sigma(self):
             self._simple_relative_entropy_implementation(rho, sigma, np.log)
         )
 
+    @pytest.mark.repeat(20)
+    def test_with_enr(self):
+        dims = EnrSpace([3, 3], 2)
+        rho = qutip.rand_dm(dims)
+        sigma = qutip.rand_dm(dims)
+        rel = qutip.entropy_relative(rho, sigma)
```
Ericgig, line 199: Should we test other entropy_types with enr?
Anything that you did not mark with `_require_pure_dims` or require qubit dimensions should work right.

Ericgig (commented):
Overall seems good. Not having to remove / fix existing tests is good.
We should probably add test with enr states in other places, but it's a good start.

I think the failing test is due to the seed fixing that you mentioned in the other PR. Adding some test had that one fall on a bad rng state... Maybe waiting for the other PR will fix them.

In `doc/apidoc/quantumobject.rst`, enr is named "Energy Restricted Operators" instead of "excitation-number restricted". Could you fix that while your at it?

qutip/measurement.py
pmenczel: This is not the same, the original code compares op.from and **state**.to

pmenczel:
> Overall seems good. Not having to remove / fix existing tests is good. We should probably add test with enr states in other places, but it's a good start.

There is certainly many more places where ENR tests could be added. I loosely followed the rules to only look at functions that access `dims` or functions from `dimensions.py`, and only add ENR tests where there are existing tests with at least a product space.

> I think the failing test is due to the seed fixing that you mentioned in the other PR. Adding some test had that one fall on a bad rng state... Maybe waiting for the other PR will fix them.

Yes, I think so.

> In `doc/apidoc/quantumobject.rst`, enr is named "Energy Restricted Operators" instead of "excitation-number restricted". Could you fix that while your at it?

Done. The module is still called `energy_restricted.py`... do you think we should rename it?

Ericgig (approved):
Look good. Thank you.

I would certainly be good to fix the file name eventually, but I don't want to stop this PR for it.

PR #2753, opened 2025-09-25
Fix tests

pmenczel:
Fixes two current problems with the tests:

1. Running `loky_pmap` and later `mpi_pmap` in the same session causes python to hang on exit. I have tracked the problem down to `loky` creating two "resource tracker" subprocesses that are never shut down. If I manually stop these subprocesses before calling `mpi_pmap`, the deadlock does not happen. Generally, I recommend not to use both loky and MPI in the same program, but to fix the problem with the test run, I added code to `mpi_pmap` to stop these resource trackers if they are running.
2. A few tests were calling `np.random.seed`, making it so that all randomness in all tests afterwards would also be determined by that seed. For example, in #2751, it happened that a flaky test (`TestMultiTrajResult::test_merge_result`) would now always fail. I have removed all calls to `np.random.seed` and I have marked the mentioned test as flaky.

When fixing the test of the fitting utilities, I noticed that the calculation of the root mean square error in the fitting routines was wrong. I have also fixed that, and then needed to adjust some tolerances.

### DIFF
--- a/qutip/solver/parallel.py
+++ b/qutip/solver/parallel.py
@@ -483,6 +483,26 @@ def mpi_pmap(task, values, task_args=None, task_kwargs=None,
 
     from mpi4py.futures import MPIPoolExecutor
 
+    # If loky is used (see loky_pmap), it starts two "resource tracker" child
+    # processes that are normally never shut down. However, using mpi4py
+    # when these child processes are still alive may lead to a deadlock.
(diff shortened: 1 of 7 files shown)


pmenczel:
Now all the test fails are just because of a coveralls issue, maybe it's just a temporary problem?

PR #2754, opened 2025-09-27
Fix color ordering for single points in Bloch sphere. Fixes issue #2681

### LINKED ISSUE #2681: Color ordering for single points in a Bloch sphere
### Bug Description

If you use a list of colors (i.e., colormaps), the ```add_points()``` method weirdly orders colors.

Image

### Code to Reproduce the Bug

```shell
from qutip import Bloch

x = [1, 0.5, 0, -0.5, -1]
y = [0, 0, 0, 0, 0]
z = [0, 0, 0, 0, 0]

b = Bloch()
b.point_color=list(['red', 'orange', 'yellow', 'cyan', 'blue'])
b.add_points([
    x,
    y,
    z
])
b.show()
```

### Code Output

```shell

```

### Expected Behaviour

Image

### Your Environment

```shell
QuTiP Version:      5.1.1
Numpy Version:      2.2.4
Scipy Version:      1.15.2
Cython Version:     None
Matplotlib Version: 3.10.1
Python Version:     3.12.10
Number of CPUs:     32
BLAS Info:          Generic
INTEL MKL Ext:      None
```

### Additional Context

My temporary solution is to remove ```[indperm]``` from these lines, but I don't know if I broke other parts of the code, so I didn't open a PR to solve this issue.

https://github.com/qutip/qutip/blob/4033695bd6e109cad576fcd647a926d42fb56878/qutip/bloch.py#L869-L879

QiLinXue:
**Description**
When choosing the color for points to be plotted on a Bloch sphere, there are 2 options that are broken:

1. By modifying `Bloch.point_color` directly. Note: code description seems to suggest this is deprecated, but the [docs](https://qutip.readthedocs.io/en/stable/guide/guide-bloch.html#configuring-the-bloch-sphere) seem to suggest it is still supported.
2. By passing a list of colours when calling `Bloch.add_points`.

In both of these cases, the order of points will get sorted by `indperm`, but the colours will not, leading to a mismatch.

For example,
```py
from qutip import Bloch

x = [1, 0.5, 0, -0.5, -1]
y = [0, 0, 0, 0, 0]
z = [0, 0, 0, 0, 0]

b = Bloch()
b.add_points([x, y, z], colors=['red', 'orange', 'yellow', 'cyan', 'blue'])
b.show()
```
and
```py
from qutip import Bloch

x = [1, 0.5, 0, -0.5, -1]
y = [0, 0, 0, 0, 0]
z = [0, 0, 0, 0, 0]

b = Bloch()
b.point_color=list(['red', 'orange', 'yellow', 'cyan', 'blue'])
b.add_points([
    x,
    y,
    z
])
b.show()
```
previously both did not work, and both cases are now fixed.

**Related issues or PRs**
Fixes #2681

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -874,8 +874,12 @@ class Bloch:
 
             if self._inner_point_color[k] is not None:
                 color = self._inner_point_color[k]
+                if isinstance(color, list) and len(color) > 1:
+                    color = [color[i] for i in indperm]
             elif self.point_color is not None:
                 color = self.point_color
+                if isinstance(color, list) and len(color) > 1:
+                    color = [color[i] for i in indperm]
             elif style in ['s', 'l']:
                 color = [self.point_default_color[
                     k % len(self.point_default_color)


Ericgig (commented):
Thank you for the fix.

The towncrier file is a plain text file, no `.rst` in the name.

Would you be able to add a test for this? There are tests in `qutip/tests/test_bloch.py`, that you could use as a base to make one.

qutip/tests/test_bloch.py
```diff
@@ -722,6 +722,25 @@ def test_clear(self, fig_test=None, fig_ref=None):
...
+        # Add point / color pairs in different order
+        b1.add_points([x, y, z], colors=colors)
+        b2.add_points([x[::-1], y, z], colors=colors[::-1])
+        b1.make_sphere()
+        b2.make_sphere()
+
+
+
+
+
```
Ericgig:
```suggestion


```

PR #2761, opened 2025-10-15
Add test for python 3.14

Ericgig:
**Description**
Add tests running on python 3.14.

Fix an incompatibility in compiling string coefficient with list subscription.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ jobs:
         # Test other versions of Python in special cases to avoid exploding the
         # matrix size; make sure to test all supported versions in some form.
         python-version: ["3.14"]
-        case-name: [python 3.14 - Latest]
+        case-name: ["python 3.14 - Latest"]
         condaforge: [1]
         numpy-build: [""]
         numpy-requirement: [""]


.github/workflows/tests.yml
```diff
@@ -30,9 +30,8 @@ jobs:
         os: [ubuntu-latest]
         # Test other versions of Python in special cases to avoid exploding the
         # matrix size; make sure to test all supported versions in some form.
-        python-version: ["3.11"]
-        case-name: [defaults]
-        # Version 2 not yet available on conda's default channel
+        python-version: ["3.14"]
+        case-name: [python 3.14 - Latest]
```
hodgestar:
```suggestion
        case-name: ["python 3.14 - Latest"]
```
I know quotes are not strictly needed here but I would add them for sanity given the string has punctuation.

.github/workflows/tests.yml
```diff
@@ -47,59 +46,54 @@ jobs:
...
             scipy-requirement: ">=1.9,<1.10"
             semidefinite: 1
-            condaforge: 1
             oldcython: 1
             oldmpl: 1
-            nomkl: 1
+            pypi: 1
             coveralls: 1
-            pytest-extra-options: "-W ignore:dep_util:DeprecationWarning -W \"ignore:The 'renderer' parameter of do_3d_projection\""
+            pytest-extra-options: "-W ignore:dep_util:DeprecationWarning -W \"ignore:The 'renderer' parameter of do_3d_projection\" -W ignore:Matrix:scipy.linalg._misc.LinAlgWarning"
```
hodgestar, line 90: What causes this warning?

qutip/core/coefficient.py
```diff
@@ -766,7 +766,8 @@ def parse(code, args, compile_opt):
         # If there is a subscript: a[b] int are always accepted to be safe
         # with TypeError.
         # Also comparison is not supported for complex.
-        accept_int = "SUBSCR" in dis.Bytecode(code).dis()
+        dis_code = dis.Bytecode(code).dis()
+        accept_int = "SUBSCR" in dis_code or "([])" in dis_code
```
hodgestar, line 770: What required this change?

hodgestar (commented):
Looks good. I asked some questions so that I know what the changes are about.

.github/workflows/tests.yml
Ericgig: Scipy 1.16 raise this warning when computing the matrix square root of a singular matrix. 
Many density matrix used in tests for fidelity, etc. are singular.

qutip/core/coefficient.py
Ericgig: In python 3.14, the bytecode of `my_list[i]` changed from `"SUBSCR"` to `"BINOPS 29 ([])` (Not sure for the 29).

Basically, there is little advantage of string over python function without types.
We type everything to complex per default, but list index must but int. Instead of parsing it ourself we ask python to parse the code and if a subscription is used, types it as int.
Very niche, but it ensure that it just work for users.

.github/workflows/tests.yml
hodgestar: Ah, this error. :)

qutip/core/coefficient.py
hodgestar: `BINOPS 29` is very obscure.

hodgestar (approved):

PR #2763, opened 2025-10-16
Increase SODE test stability

Ericgig:
**Description**
Improve reliability of the SODE tests.
They have been running with a fixed seed for a long time, but #2753 ensured different seed each runs.

The failure rate was close to 1% for the worst method...
With the number of tests, this was failing too often.

With these changes, I saw 1 failure in 5000 runs.


Ericgig:
Ps. This is the histogram of computed error in 5000 runs. 
We expect a measured error of 1.5, at 1 there is clearly a bug.
hist

PR #2766, opened 2025-10-20
Change measure to handle any iterable of Qobj instead of only lists

cpossel:
**Description**
'measure' function currently only allows Qobj or list[Qobj] as input for the operators. In the first case it invokes projective measurement, it the second case POVM measurement. Passing tuples or other iterables failed due to a single isinstance check even though all functions can in principle handle iterables without further adjustments.
With the proposed changes, any iterable[Qobj] can be used.

### DIFF
--- a/qutip/tests/test_measurement.py
+++ b/qutip/tests/test_measurement.py
@@ -228,10 +228,12 @@ def test_measure_observable(op, state):
                     pytest.param(PY, basis(2, 1), id="PY_ket2"),
                     pytest.param(PY, ket2dm(basis(2, 1)), id="PY_dm"),
 
-                    pytest.param(tuple(PZ), basis(2, 0), id="PZ_ket1"),
-                    pytest.param(tuple(PZ), basis(2, 1), id="PZ_ket2"),
-                    pytest.param(tuple(PZ), ket2dm(basis(2, 0)), id="PZ_dm1"),
-                    pytest.param(tuple(PZ), ket2dm(basis(2, 1)), id="PZ_dm2")])
+                    pytest.param(tuple(PZ), basis(2, 0), id="PZ_ket1_tuple"),
+                    pytest.param(tuple(PZ), basis(2, 1), id="PZ_ket2_tuple"),
+                    pytest.param(
+                        tuple(PZ), ket2dm(basis(2, 0)), id="PZ_dm1_tuple"),
+                    pytest.param(
+                        tuple(PZ), ket2dm(basis(2, 1)), id="PZ_dm2_tuple")])
 def test_measure(ops, state):
     """measure_povm: test on basis states using different projectors """
 


qutip/tests/test_measurement.py
```diff
@@ -227,7 +227,9 @@ def test_measure_observable(op, state):
 
                     pytest.param(PY, basis(2, 0), id="PY_ket1"),
                     pytest.param(PY, basis(2, 1), id="PY_ket2"),
-                    pytest.param(PY, ket2dm(basis(2, 1)), id="PY_dm")])
+                    pytest.param(PY, ket2dm(basis(2, 1)), id="PY_dm"),
+
+                    pytest.param(tuple(PZ), basis(2, 0), id="PZ_ket1_tuple")])
```
Ericgig: Let's add also a test for the `dm` case.

qutip/tests/test_measurement.py
```diff
@@ -1,14 +1,15 @@
...
-import scipy
 import pytest
-from math import sqrt
-from qutip import (
-    Qobj, basis, ket2dm, sigmax, sigmay, sigmaz, identity, num, tensor,
-    rand_ket
-)
+import scipy
+
+from qutip import Qobj, basis, identity, ket2dm, num, rand_ket, sigmax, sigmay, sigmaz, tensor
```
Ericgig: We use 80 characters line width.

qutip/measurement.py
```diff
@@ -47,8 +47,8 @@ def _measurement_statistics_povm_ket(state, ops, tol=None):
     state : :class:`.Qobj` (ket)
             The ket specifying the state to measure.
 
-    ops : list of :class:`.Qobj`
-      List of measurement operators :math:`M_i` (specifying a POVM such that
+    ops : iterable of :class:`.Qobj`
+      Iterable of measurement operators :math:`M_i` (specifying a POVM s.t.
       :math:`E_i = M_i^\dagger M_i`).
```
Ericgig:
```suggestion
    ops : iterable of :class:`.Qobj`
      Iterable of measurement operators :math:`M_i` (specifying a POVM
      such that :math:`E_i = M_i^\dagger M_i`).
```

qutip/tests/test_measurement.py
cpossel: I added it for all cases to keep the symmetry ;) 
Resolved by commits 0de2d81 and b85dfd4

qutip/tests/test_measurement.py
cpossel: There is always a line that slips the attention... I resolved it via removing the imports that aren't used either way.
Resolved by commit bbbb212

qutip/measurement.py
cpossel: I wanted to stay consistent with all the other cases, where 's.t.' is used. If writing out 'such that' is preferred, I'm fine with it, too (better readability). But I'm changing it everywhere for consistency.
Resolved by commit eaa3395

cpossel: @Ericgig Thanks for the review. I'm leaving it to you to mark the conversations as resolved, okay?

Ericgig (approved):
Thank you

PR #2771, opened 2025-10-30
Data layer support for block matrices (direct sum #1)

pmenczel:
This is the first part of my direct sums project, originally #2662 (see there for more background)

* Adds three functions dealing with block matrices to the data layer: `concat_blocks` (create large matrix from blocks), `slice` (extract block from large matrix), `insert_block` (overwrite a block in a larger matrix)
* Currently only includes `CSR` and `Dense` implementations. I am still not so familiar with `Dia`, any opinions on whether we need it here? Eventually, implementations in qutip-jax and qutip-cuquantum might be needed as well.
* The CSR version of `concat_blocks` is a slight generalization of `csr._from_csr_blocks`, which was used in HEOM. I have changed the HEOM code to use `concat_blocks` instead, so that `csr._from_csr_blocks` could be removed. Further changes to HEOM (including custom rhs dtype and time-dependent coupling operators) will be coming in a later PR.
* I have done a bit of performance testing of the HEOM rhs construction. This PR speeds up `gather()` by a small amount (<5%) in my tests. If we included another function `concat_blocks_homogeneous` where all blocks are assumed to be the same size, gather could be sped up a bit more (10-15%). I think the added complexity is not worth it, given that gather itself is just a small part of the whole rhs construction. Let me know if you think it would be good to include.
* In steadystate.py there used to be the comment
```# We don't have a function to overwrite part of an array, so```
I have rewritten that part with `insert_block`. Let me know if you know of any other places where these new functions could be useful in the existing codebase.

### DIFF
--- a/qutip/core/data/block_operations.pyx
+++ b/qutip/core/data/block_operations.pyx
@@ -96,7 +96,7 @@ cpdef CSR block_build_csr(
 
     # avoid modifying blocks in-place, but only copy the array if necessary
     cdef Data[:] blocks_copy = blocks
-    cdef bool copied = False
+    cdef bint copied = False
 
     for idx in range(num_ops):
         row = block_rows[idx]

### CODE REVIEW

File: qutip/core/data/block_operations.pyx [hunk 1]
```diff
@@ -0,0 +1,423 @@
...
+
+    # the resulting nnz is the sum of the nnz of data and the nnz of block,
+    # minus the number of non-zero elements in data that are overwritten
+    cdef base.idxint nnz = csr.nnz(data) + csr.nnz(block)
+    for row in range(above, above + block_height):
+        for idx_data in range(data.row_index[row], data.row_index[row + 1]):
+            if data.col_index[idx_data] < before:
+                continue
+            if data.col_index[idx_data] >= before + block_width:
+                break
```
Ericgig: While we try to keep the col index sorted, it's not certain.
We run tests for CSR with col index not sorted for many operation, we should also support / test these case.

qutip/core/data/block_operations.pyx
```diff
@@ -0,0 +1,423 @@
...
+    out[0] = 0
+    for i in range(n):
+        out[i + 1] = out[i] + array[i]
+    return out
+
+
+cpdef Dense concat_blocks_dense(
+    base.idxint[:] block_rows, base.idxint[:] block_cols, Data[:] blocks,
+    base.idxint[:] block_widths, base.idxint[:] block_heights
+):
```
Ericgig: This is good as an internal function, but it's not users friendly...
`idxint` is internal, I expect users to simply uses default `int`, which many not work...
Are you thinking of keeping it only internal, or adding a python wrapper with easier interface (and would sort CSR block etc.)?

I would invert `block_widths` and `block_heights` to match the order in `shape`.

qutip/core/data/block_operations.pyx
```diff
@@ -0,0 +1,423 @@
...
+
+    for idx in range(num_ops):
+        row = block_rows[idx]
+        column = block_cols[idx]
+        # check ops are ordered by (row, column)
+        if idx > 0 and (
+            row < block_rows[idx - 1] or
+            (row == block_rows[idx - 1] and column < block_cols[idx - 1])
+        ):
+            raise ValueError("The arrays block_rows and block_cols must be "
```
Ericgig, line 109: There is an error if blocks are not sorted, but what about repeated block?
With dense, the second entry would overwrite the first, but with csr, I think both entries would be inserted and we may end up with multiple values at one position...

Ericgig (commented):
Good work.
I added a few comments about a few edge case to fix.

I can help with adding support for "dia" matrix if needed.

qutip/core/data/block_operations.pyx
pmenczel: Just for clarity: these functions are intended to be internal. In the next PR I will add  `qt.direct_sum` etc, where they are used. `qt.direct_sum` will operate on `Qobj`, take care of the sorting etc.

I do think we need `idxint` in these functions. The point of `idxint` is to allow very large sparse matrices indexed by 64bit numbers, right? If we want to allow that, then we probably also need to allow large numbers of ADOs in the HEOM solver, that is, more than what fits in an int.

If you think it would be better, I could add a wrapper based on `int` also here, though.

Ericgig: `idxint` per default is `int32`, smaller than python `int` (`int64` / large int). 
The version of blas you get with numpy / scipy is compiled with 32 bits int so it's what we use.
For larger matrices, some manual compilation is needed.

If ` qt.direct_sum` is coming in the next PR, then using `Data` and `idxint` arrays here is fine.

hodgestar:
@pmenczel Thank you for splitting this out nicely & pushing the direct sum work forward. I'll do a review shortly but first some responses to the summary:

>     * Adds three functions dealing with block matrices to the data layer: `concat_blocks` (create large matrix from blocks), `slice` (extract block from large matrix), `insert_block` (overwrite a block in a larger matrix)

I would give these more descriptive names that clearly group the operations together. Maybe:

* `build_from_blocks`
* `extract_block`
* `overwrite_block`

Or perhaps:

* `block_build`
* `block_extract`
* `block_overwrite`
 
>     * Currently only includes `CSR` and `Dense` implementations. I am still not so familiar with `Dia`, any opinions on whether we need it here? Eventually, implementations in qutip-jax and qutip-cuquantum might be needed as well.

I doubt `Dia` works nicely when there are blocks present far from the diagonal of the full matrix, but there are cases like the HEOM where the blocks are close to the diagonals. Maybe worth it eventually, but I would vote to just leave it for later.

>     * The CSR version of `concat_blocks` is a slight generalization of `csr._from_csr_blocks`, which was used in HEOM. I have changed the HEOM code to use `concat_blocks` instead, so that `csr._from_csr_blocks` could be removed. Further changes to HEOM (including custom rhs dtype and time-dependent coupling operators) will be coming in a later PR.

Perfect.

>     * I have done a bit of performance testing of the HEOM rhs construction. This PR speeds up `gather()` by a small amount (<5%) in my tests. If we included another function `concat_blocks_homogeneous` where all blocks are assumed to be the same size, gather could be sped up a bit more (10-15%). I think the added complexity is not worth it, given that gather itself is just a small part of the whole rhs construction. Let me know if you think it would be good to include.

Let's leave it out for now and reconsider once the dust has settled for the direct sum support, which should improve HEOM performance significantly anyway.

>     * In steadystate.py there used to be the comment
>       `# We don't have a function to overwrite part of an array, so`
>       I have rewritten that part with `insert_block`. Let me know if you know of any other places where these new functions could be useful in the existing codebase.

I can't think of any others offhand.

qutip/core/data/block_operations.pyx
```diff
@@ -0,0 +1,423 @@
...
+            or block.shape[1] != block_widths[column]
+        ):
+            raise ValueError(
+                f"Block operator does not have the correct shape at row={row},"
+                f" column={column}."
+            )
+
+        if type(block) is not CSR:
+            block = <Data>convert.to(CSR, block)
+            blocks[idx] = block
```
hodgestar: After reading the note in `concat_blocks`, I'm assuming this is the bit of code that needs to modify `blocks` in place. I'm wondering why this is a performance issue. Sure, it's `O(n)` in the number of blocks, but the rest of the work is `O(n * number_of_entries_per_block)` which is much bigger.

Is it possible that we something else caused the big performance hit and that we could avoid modifying the passed lists of blocks?

qutip/core/data/block_operations.pyx
```diff
@@ -0,0 +1,423 @@
...
+    block_height, block_width = block.shape
+    if (
+        above < 0 or before < 0
+        or above + block_height > data_height
+        or before + block_width > data_width
+    ):
+        raise IndexError("Cannot insert block into data: doesn't fit.")
+
+    cdef cnp.ndarray data_array = data.to_array()  # copies
+    cdef cnp.ndarray block_array = block.as_ndarray()  # doesn't copy
```
hodgestar: We could easily also support `block.to_array()` here (e.g. via checking the block type) and then `insert_block_dense` could support `Data, Data, Dense`.

qutip/core/data/block_operations.pyx
pmenczel: I am a bit surprised as well. I have added the code and notebook for my benchmark on this branch: https://github.com/pmenczel/qutip/tree/direct-sum-heomtest
In particular, the modified `block_build_csr` that does not modify `blocks` in place can be found here: https://github.com/pmenczel/qutip/blob/683dbd31c9422b4c910d4c2259b6606909f639e9/qutip/core/data/block_operations.pyx#L170

The following table shows the time in ms for `_GatherHEOMRHS.gather()`, averaged over 5 runs on my laptop.

| | 816 ADOs | 3,876 ADOs | 15,504 ADOs | 54,264 ADOs | 170,544 ADOs | 490,314 ADOs |
| -- | -- | -- | -- | -- | -- | -- |
| old `csr._from_csr_blocks` | 0.8 | 4 | 22 | 88 | 318 | 1001 |
| new `overwrite_block` | 0.8 | 5 | 20 | 90 | 352 | 1050 |
| new `overwrite_block` (`blocks` not modified) | 1.5 | 10 | 40 | 170 | 651 | 1912 |

As I said, `_GatherHEOMRHS.gather()` is a relatively small fraction of the whole HEOM rhs construction time, but still, I didn't want to make the function twice as slow...

pmenczel:
Thank you both for the reviews. I have implemented all changes, other than the modifying `blocks` in place mentioned by @hodgestar. I have also renamed the functions as suggested by @hodgestar.

If we want to add support for `Dia` later, I would be happy about help, but let's leave it for now.

pmenczel:
> Happy with the fixes. There are some pep8 issues that would be nice to fix.

Ah, I've been meaning to ask why we are not running automated pep8 checks anymore. But I forgot to do it manually.
I have fixed the too long lines in the test. I don't know if there is a way to pep8 check the cython files?

qutip/core/data/block_operations.pyx
hodgestar: I'm still pretty surprised by the numbers. I wouldn't mind understanding better why copying the array of pointers doubles the time, but let's not let it hold up this MR.

I'm guess is that something we don't yet understand is happening (or at least that I don't understand).

pmenczel:
I have made a little change to avoid modifying `blocks` in place after all. I hope it is okay that I will merge now.

PR #2774, opened 2025-11-01
feat: speed up propagator for piecewise constant Hamiltonians

rafaelha:
This PR introduces `propagator_piecewise` that directly exponentiates piecewise constant Hamiltonians or Liouvillians. This is also integrated into `propagator`, where an optional argument `piecewise_t` is introduced.

Providing this optional argument generally speeds up the propagator computation, as shown by this simple benchmark:

image


qutip/solver/propagator.py
```diff
@@ -18,12 +18,101 @@
...
+    ValueError
+        If any collapse operator is time-dependent (not constant).
+    """
+
+    constant_c_ops = c_ops is None or all(op.isconstant for op in c_ops)
+
+    if not constant_c_ops:
+        raise ValueError(
+            "Collapse operators must be constant for piecewise propagation."
+        )
```
Ericgig: If the collapse operators are also piecewise constant, wouldn't is work?

qutip/solver/propagator.py
```diff
@@ -18,12 +18,101 @@
...
+
+    piecewise_times = {tt for tt in piecewise_t if tlist[0] < tt <= tlist[-1]}
+    eval_times = set(tlist)
+    times = sorted(eval_times | piecewise_times)
+
+    out = []
+    prev = times[0]
+    for nxt in times[1:]:
+        H_step = H(prev, args)
```
Ericgig: This feels risky:
```
def H(t):
    if t <= 1.:
        return H0
    else:
        return H1
```
```
def H(t):
    if t < 1.:
        return H0
    else:
        return H1
```
Are effectively the same Hamiltonian, but will give different results with this function.
It would be safer to take it between `prev` and `nxt`.

Ericgig (commented):
Thank you.
Seems like an interesting addition.

qutip/solver/propagator.py
rafaelha: That's a very good suggestion. I've implemented it and added a unit test.

qutip/solver/propagator.py
rafaelha: You're right. I removed that check and added a unit test.

qutip/solver/propagator.py
```diff
@@ -18,12 +18,92 @@
...
+            gen = liouvillian(H_step, c_ops_q)
+        else:
+            gen = H_step if H_step.issuper else -1j * H_step
+
+        if prev == times[0]:
+            U = qeye_like(gen)
+            if prev in eval_times:
+                out.append(U)
+
+        U = (gen * (nxt - prev)).expm() * U
```
Ericgig: Often the output time differences are constant (built with np.linspace).
So the step contribution is always the same.
It could be even faster if it did remember the previous computation.
```
if nxt - prev != prev_dt:
    prev_dt = nxt - prev
    dU = (gen * (nxt - prev)).expm()
U = dU @ U
```
Since time are sorted, initial `prev_dt` can be set to `-1` at the start and each time `nxt` is in `piecewise_times`.

This would speed up the cases where `eval_times` step are constant and smaller than `piecewise_times` ones at little cost.

(if you want to be fancy, you can add a tolerance in the check to account for floating point precision: `settings.core["atol"]`.)

Ericgig (approved):
Thank you, looks good.

I think it could be optimised some more for uniformly distributed `tlist`.

qutip/solver/propagator.py
rafaelha: Ok! I implemented that as well and added unit tests. Please have a look!

Ericgig (approved):
Thank you!

rafaelha:
> Thank you!

Ready to merge from my end! (I don't think I have access to merge myself)

amirebrahimi:
Thank you for this improvement, @rafaelha.

PR #2775, opened 2025-11-04
Fix missing core.tensor members in docs

cpossel:
**Description**
The functions `tensor_swap` and `expand_operator` from `qutip.core.tensor` have docstrings and were present in the documentation of QuTiP 4. But they can't be found anymore in QuTiP 5. So, they should be added to the APIdoc, too.

There might be some other functions missing (@nwlambert mentioned some ptrace function).
So, before merging this pull request, it might be beneficial to look for some other missing functions.

### DIFF
--- a/doc/changes/2775.bugfix
+++ b/doc/changes/2775.bugfix
@@ -1,0 +1,1 @@
+Fix mssing binding of docstrings of tensor_swap and expand_operator into APIdoc
--- a/doc/changes/2775.bugfix
+++ b/doc/changes/2775.bugfix
@@ -1,1 +1,1 @@
-Fix mssing binding of docstrings of tensor_swap and expand_operator into APIdoc
+Fix missing binding of docstrings of tensor_swap and expand_operator into APIdoc
--- a/qutip/core/tensor.py
+++ b/qutip/core/tensor.py
@@ -455,19 +455,20 @@ def expand_operator(
     targets: int,
     dtype: LayerType = None
 ) -> QobjOrQobjEvo:
-    """
-    Expand an operator to one that acts on a system with desired dimensions.
-    e.g.
-    ```
-    expand_operator(oper, [2, 3, 4, 5], 2) ==
-        tensor(qeye(2), qeye(3), oper, qeye(5))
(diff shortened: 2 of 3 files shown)


qutip/core/tensor.py
```diff
@@ -455,19 +455,20 @@ def expand_operator(
...
+
+        expand_operator(oper, [2, 3, 4, 5], 2) ==
+            tensor(qeye(2), qeye(3), oper, qeye(5))
+        expand_operator(tensor(oper1, oper2), [2, 3, 4, 5], [2, 0]) ==
+            tensor(oper2, qeye(3), oper1, qeye(5))
 
     Parameters
     ----------
-    oper : :class:`.Qobj`
+    oper : Qobj
```
Ericgig: Let's keep the hyperlinks.

doc/changes/2775.bugfix
```diff
@@ -0,0 +1 @@
+Fix mssing binding of docstrings of tensor_swap and expand_operator into APIdoc
```
Ericgig:
```suggestion
Fix missing binding of docstrings of tensor_swap and expand_operator into APIdoc
```

Ericgig (commented):
Sorry for the delay, I was expecting Neill to review.
Thank you for the realizing they were missing.

Please restore the hyperlinks.

qutip/core/tensor.py
cpossel: @Ericgig Removing the hyperlinks was an attempt to make it more consistent throughout the file. Cause tensor_swap and tensor_contract use Qobj while expand_operator and the other functions use :class:`.Qobj`... So, should I just revert those that I changed. Or also make the others consistent?

Ericgig: The link is preferred (at least the first time in a docstring).
Ideally, if you could add them, it would be appreciated.

cpossel: Alright. I added them wherever I considered it proper.

cpossel:
@Ericgig Thanks for the review. I restored the hyperlinks and fixed the typo.

Ericgig (approved):
Thank you!

PR #2776, opened 2025-11-04
Improve error message for coefficient function parameter naming

basnijholt:
## Problem

When users define time-dependent coefficient functions as `def func(time, args)` instead of `def func(t, args)`, QuTiP's auto-detection chooses the wrong calling style, producing an unhelpful error:

```
TypeError: func() missing 1 required positional argument: 'args'
```

This happens because QuTiP auto-detects calling style based on the exact signature:
- `def func(t, args)` → dict-style → calls `func(t, {"key": val})`
- `def func(time, args)` → pythonic-style → calls `func(time, key=val)` (wrong!)

## Solution

Add validation at `FunctionCoefficient` initialization to catch this mistake early with a clear error message:

```
TypeError: Coefficient function 'func' has signature (time, args). 
When using 'args' as a parameter name, the first parameter must be named 't' 
for dict-style: def func(t, args).
```

The check only triggers for exactly 2-parameter functions where the second parameter is named `args` and the first is not `t`. This avoids most false positives while catching the common mistake pattern.

Note: There is one false positive case where someone uses pythonic-style but names a parameter `args`, e.g., `def func(time, args): return time * args` with `args={"args": 2.0}`. This is an unusual pattern that violates QuTiP conventions and is inherently confusing. Users can bypass the check with explicit `function_style="pythonic"` if needed.

## Changes

- Modified `qutip/core/cy/coefficient.pyx`: Added validation in `FunctionCoefficient.__init__`
- Added `test_coefficient_parameter_naming()` documenting the behavior and edge cases

### DIFF
--- a/qutip/core/cy/coefficient.pyx
+++ b/qutip/core/cy/coefficient.pyx
@@ -198,6 +198,22 @@ cdef class FunctionCoefficient(Coefficient):
                 )
             _f_pythonic, _f_parameters = coefficient_function_parameters(
                 func, style=style)
+
+            # Detect common mistake: pythonic style but has 'args' parameter
+            # Only flag if EXACTLY 2 params - the classic mistake pattern
+            if _f_pythonic and _f_parameters is not None and 'args' in _f_parameters:
+                sig = inspect.signature(func)
+                param_names = list(sig.parameters.keys())
+                if len(param_names) == 2 and param_names[0] != 't':
+                    raise TypeError(
+                        f"Coefficient function '{func.__name__}' has signature "
+                        f"({', '.join(param_names)}). When using 'args' as a "
+                        f"parameter name, the first parameter must be named 't' "
+                        f"for dict-style: def {func.__name__}(t, args). "
+                        f"Or use pythonic style with different parameter names: "
+                        f"def {func.__name__}(t, {', '.join(p for p in param_names[1:] if p != 'args')})."
+                    )
+
             if _f_parameters is not None:
                 args = {k: args[k] for k in _f_parameters & args.keys()}
             else:
--- a/qutip/tests/core/test_coefficient.py
+++ b/qutip/tests/core/test_coefficient.py
@@ -489,3 +489,66 @@ def test_coefficient_parallel(map_func):
     for coeff in coeffs:
         assert isinstance(coeff, Coefficient)
         _assert_eq_over_interval(coeff, expected)
+
+
+def test_coefficient_parameter_naming():
+    """
+    Test that coefficient functions handle parameter naming correctly.
+
+    QuTiP auto-detects function style based on signature:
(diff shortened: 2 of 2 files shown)


qutip/core/cy/coefficient.pyx
```diff
@@ -198,6 +198,22 @@ cdef class FunctionCoefficient(Coefficient):
...
+                param_names = list(sig.parameters.keys())
+                if len(param_names) == 2 and param_names[0] != 't':
+                    raise TypeError(
+                        f"Coefficient function '{func.__name__}' has signature "
+                        f"({', '.join(param_names)}). When using 'args' as a "
+                        f"parameter name, the first parameter must be named 't' "
+                        f"for dict-style: def {func.__name__}(t, args). "
+                        f"Or use pythonic style with different parameter names: "
+                        f"def {func.__name__}(t, {', '.join(p for p in param_names[1:] if p != 'args')})."
+                    )
```
Ericgig, line 215: This should belong to the `coefficient_function_parameters` function.

qutip/tests/core/test_coefficient.py
```diff
@@ -489,3 +489,66 @@ def test_coefficient_parallel(map_func):
...
+    def pythonic_style(t, multiplier):
+        return t * multiplier
+
+    coeff2 = coefficient(pythonic_style, args={"multiplier": 2.0})
+    assert coeff2(0.5) == 1.0
+
+    # Case 3: COMMON MISTAKE - Using 'args' parameter but first param not 't'
+    # This should raise a helpful error because auto-detection will choose
+    # pythonic style (since first param isn't 't'), but the function expects
+    # dict style (it has a parameter named 'args').
```
Ericgig, line 523: You can remove the overly verbose AI comments.

Ericgig (commented):
Thank you for reporting the issue.
However, this feels a half-solution: the documentation is not clear and should be also improved.

- The documentation in `coefficient` only has example with `t` as the first parameter.
It could be interpreted that any function where the first parameter is not `t` is wrong and should raise an error. 
- `QobjEvo` only refer to `coefficient` but has no pythonic example.
- The guide says that `args` as the second argument is sufficient to be understood as old mode. (+ the exemple right before [qobjevo](https://qutip.readthedocs.io/en/latest/guide/dynamics/dynamics-time.html#qobjevo ) mixes both patterns and will only cause confusion.)

So the guide is wrong and the docstring is unhelpful. These need to be fixed for the problem to be fully resolved.

I would argue that changing it so the guide is right and `f(anything, args)` is understood as old mode would be a more satisfying solution.
But this could break someone code?
Maybe raising warning instead of an error asking not to use confusing pattern would be enough?

Ericgig:
Closed for inactivity.

PR #2782, opened 2025-11-06
Fix expected wheel number

Ericgig:
**Description**
Hot fix for 5.2.2 release.

The new py3.14 wheels are seen as extra and stopping the upload to pypi.


hodgestar (commented):
:+1:

PR #2784, opened 2025-11-06
Update build python version

Ericgig:
**Description**
When updating the build action for 3.14, I updated the runner python version, but forgot it in one place leading to an error just before uploading the wheels to pypi...

I already force merged this on the qutip-5.2.X branch for the release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,8 +105,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # between 13 and 14, mac changed from intel chip to apple silicon
-        os: [ubuntu-latest, windows-latest, macos-13, macos-latest]
+        # mac has 2 chipsets that need different build.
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -176,7 +176,7 @@ jobs:
       - name: Verify this is not a dev version
         shell: bash
         run: |
-          python -m pip install wheels/*-cp310-cp310-manylinux*.whl
+          python -m pip install wheels/*-cp311-cp311-manylinux*.whl
```
hodgestar, line 179: Why does the Python 3.10 wheel not exist (assuming that was the issue)? It's in the `CIBW_BUILD: "cp3{10,11,12,13,14}-*"` in the build step.

Ericgig: The wheel exist, but the python version used by the action is 3.11.
cibuildwheel 3 (needed to build for python 3.14) only support 3.11 and up.

hodgestar: Ah, got it.

hodgestar (approved):

PR #2785, opened 2025-11-07
Direct sums

pmenczel:
This is the main part of my direct sums project (more background in original PR #2662, first part was #2771). After this, I am planning two more PRs: one to update the HEOM solver, and one to make sure that direct sum dimensions work as expected wherever it is possible (like I did recently for ENR dimensions).

This adds a new module `qutip.core.direct_sum` with 4 functions: `direct_sum`, `direct_sum_sparse`, `direct_component` and `set_direct_component`. The main function `direct_sum` makes block matrices or block vectors, like so:
```python
In [1]: import qutip as qt

In [2]: op = qt.direct_sum([
   ...:     [qt.sigmax(),          None],
   ...:     [qt.basis(2, 1).dag(), 42  ],
   ...: ])

In [3]: op
Out[3]: 
Quantum object: dims=[([2], [1]), ([2], [1])], shape=(3, 3), type='oper', dtype=CSR, isherm=False
Qobj data =
[[ 0.  1.  0.]
 [ 1.  0.  0.]
 [ 0.  1. 42.]]
```
The blocks can be extracted again using `direct_component`:
```python
In [4]: qt.direct_component(op, 1, 0)
Out[4]: 
Quantum object: dims=[[1], [2]], shape=(1, 2), type='bra', dtype=CSR
Qobj data =
[[0. 1.]]
```
Similarly, `set_direct_component` can be used to overwrite a block. `direct_sum_sparse` is an alternative for large, sparse block matrices (e.g., HEOM). Also `QobjEvo` works everywhere.

In addition:
* `qutip.core.dimensions` gained a new class `SumSpace` to describe the dimensions of direct sum objects. In list representation, direct sums appear as tuples. I have fiddled a bit to make a SumSpace, where all spaces are the same, only store the space once and have a nicer string representation.
It took some thinking to decide whether `SumSpace(x)` with just a single space `x` should be the same as just `x`. I settled on "yes, except if x is a sum space itself". Some reasoning for this can be found in a comment.
* I updated `QobjEvo.compress` to discard zero elements. (Zero elements are common when extracting a component of a direct sum QobjEvo.)
* There is a small fix in `states.py` (because I wanted to test the `idx2dims` / `dims2idx` of SumSpace)
* Tests and docs have been added

Edit: removed `direct_sum.py` from this PR, only keeping changes to `dimension.py` and `QObjEvo`.

### DIFF
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -397,6 +397,11 @@ class MetaDims(type):
             else:
                 return proc_args[0]
 
+        # check for extra kwargs
+        proc_kwargs.pop("rep", None)  # extra "rep" is allowed
+        if len(proc_kwargs) > 0:
+            raise ValueError(f"Unexpected keyword arguments: {proc_kwargs}")
+
         proc_args = tuple([
             tuple(arg) if isinstance(arg, list) else arg
             for arg in proc_args
@@ -411,6 +416,11 @@ class MetaDims(type):
 class Space(metaclass=MetaDims):
     @classmethod
     def _process_args(cls, *args, **kwargs):
+        if "repeat" in kwargs:
+            repeat = kwargs.pop("repeat")
+            inner_space = Space(*args, **kwargs)
+            return SumSpace, [inner_space], {"repeat": repeat}
+
         if len(args) == 1 and isinstance(args[0], (list, tuple, _homtuple)):
             # From a list (or tuple) of int
             rep = kwargs.pop("rep", None)
--- a/qutip/tests/core/test_dimensions.py
+++ b/qutip/tests/core/test_dimensions.py
@@ -10,7 +10,7 @@ from qutip.core.dimensions import (
     flatten, unflatten, enumerate_flat, deep_remove, deep_map,
     dims_idxs_to_tensor_idxs, dims_to_tensor_shape, dims_to_tensor_perm,
     einsum, to_tensor_rep, from_tensor_rep,
-    Dimensions, Field, Space, SumSpace, SuperSpace, _homtuple
+    Dimensions, Compound, Field, Space, SumSpace, SuperSpace, _homtuple
 )
 from qutip.core.energy_restricted import EnrSpace
 
@@ -249,6 +249,20 @@ class TestSumSpace:
 
         assert SumSpace(Space(([2], [3])), repeat=1) is Space((([2], [3]),))
 
+    @pytest.mark.parametrize("space_arg", [
+        pytest.param(1, id="field"),
+        pytest.param(3, id="ket"),
+        pytest.param([3], id="ket-from-list"),
+        pytest.param([2, 3], id="compound"),
+        pytest.param([[2], [3]], id="superket"),
+    ])
+    @pytest.mark.parametrize("repeat", [None, 1, 3])
+    def test_repeat_space(self, space_arg, repeat):
+        assert (
+            Space(space_arg, repeat=repeat)
+            is SumSpace(Space(space_arg), repeat=repeat)
+        )
+
     def test_simplification(self):
         assert Space(([1],)) is Field()
         assert Space(([3, 2],)) is Space([3, 2])
@@ -458,3 +472,14 @@ def test_scalar_dims(list_dims, expected):
         assert Dimensions(list_dims).as_list() == expected
     with qutip.CoreOptions(auto_tidyup_dims=False):
         assert Dimensions(list_dims).as_list() == list_dims
+
+
+@pytest.mark.parametrize(["cls", "args", "kwargs"], [
+    pytest.param(SuperSpace, [Dimensions([[2], [3]])], {"superrep": "choi"},
+                 id="typo"),
+    pytest.param(Compound, [Space(2), Space(3)], {"repeat": 5},
+                 id="unsupported")
+])
+def test_extra_kwargs(cls, args, kwargs):
+    with pytest.raises(ValueError, match="Unexpected keyword arguments"):
+        cls(*args, **kwargs)


qutip/core/dimensions.py
```diff
@@ -649,6 +688,176 @@ def scalar_like(self) -> Space:
...
+        """
+        # The metaclass makes the following substitutions:
+        # - SumSpace(x, x, ..., x) --> SumSpace(x, repeat=n)
+        # - SumSpace(x, repeat=1)  --> SumSpace(x)
+        # Therefore, every possible sum space only has one internal
+        # representation, either with repeat = None or with repeat > 1.
+
+        if len(spaces) == 0 or (repeat is not None and repeat <= 0):
+            raise ValueError("Need at least one space for direct sum.")
+        if repeat and len(spaces) != 1:
```
Ericgig: Should there be a check for repeat, not being and int?

qutip/core/dimensions.py
```diff
@@ -649,6 +688,176 @@ def scalar_like(self) -> Space:
...
+    #
+    # The reason is the following. (a) With nested direct sums, for
+    # direct_component to always work as expected, we need to keep track of
+    # the order in which direct sums were taken. We therefore can't simplify
+    # SumSpace(SumSpace(...)) to SumSpace(...). (b) We expect that
+    # direct_sum(ket, scalar) always works, even if ket is a direct sum itself.
+    # Therefore, SumSpace(Field()) must simplify to Field().
+    #
+    # Simplifying if x is not a SumSpace itself allows the maximum flexibility
+    # while still allowing direct_component to function.
```
Ericgig, line 743: I do not understand why you can't simplify SumSpace(SumSpace(...)) to SumSpace(...)...

I expect `direct_sum([direct_sum([A, B]), C[)` to be the same as `direct_sum([A, B, C])`.
the same way `tensor([tensor([A, B]), C[)` is the same as `tensor([A, B, C])`.

I also expect `direct_component` to be able to extract the `A, B` part of `direct_sum([A, B, C])`.
Similar to how `ptrace` can keep / trace any subspace.

Do you have any example where this could not work?

qutip/core/dimensions.py
```diff
@@ -365,17 +383,37 @@ def __call__(cls, *args: SpaceLike, rep: str = None) -> "Space":
...
-        if args not in cls._stored_dims:
+        if cls is SumSpace:
+            if len(args) > 1 and all(arg == args[0] for arg in args):
+                repeat = len(args)
+                args = (args[0],)
+            kwargs = {'repeat': repeat}
+            stored_args = (*args, *kwargs.items())
+        else:
+            kwargs = {}
+            stored_args = args
```
Ericgig: `kwargs` and `stored_args` feels over complicated...
If you change the SumSpace.__init__ signature to `(self, spaces: tuple[Space], repeat=None)` then `args = (args, repeat)`, no need for new variables.
You could also remove the `repeat` and create the _homtuple here.

qutip/core/states.py
```diff
@@ -156,7 +158,7 @@ def basis(
     else:
         if not isinstance(n, list):
             n = [n]
-        if len(n) != len(dimensions.as_list()):
+        if dimensions._pure_dims and len(n) != len(dimensions.as_list()):
```
Ericgig, line 159: If `basis` can be used to create summed space ket, the documentation need to be updated to explain how.
If not shouldn't we raise and error is not `dimensions._pure_dims`?

qutip/core/dimensions.py
```diff
@@ -649,6 +688,176 @@ def scalar_like(self) -> Space:
...
+        )
+
+    def replace_superrep(self, super_rep: str) -> "Space":
+        if super_rep != 'super':
+            raise ValueError("Direct sums only accept superoperators"
+                             " in super representation.")
+        return self
+
+    def scalar_like(self) -> "Space":
+        return self.spaces[0].scalar_like()
```
Ericgig: Shouldn't this keep the SumSpace structure?

Ericgig (commented):
Partial review:
- QobjEvo changes look good.
- Can basis be used for non pure state?
- For dimensions, it seems quite complex... Are stacked `SumSpace` needed?
  Should we avoid `Compound(SumSpace)`? How about `SuperSpace(SumSpace)`? I would expect `SumSpace` should be _final_, can't be packaged anymore.

qutip/core/dimensions.py
pmenczel: You mean as in `SumSpace([space.scalar_like() for space in self.spaces])`? That would not be correct I think, since the result would not be 1-dimensional.

I guess, in principle we should check in the constructor that `scalar_like` is the same for all of the component spaces. But it felt a bit unnecessary to me...

pmenczel:
Thank you for the review @Ericgig 

**Regarding the change to `basis`**

> Can basis be used for non pure state?
> 
> If `basis` can be used to create summed space ket, the documentation need to be updated to explain how.
If not shouldn't we raise and error is not `dimensions._pure_dims`?

With the change here, it now works with sum space. It is a quite simple change so I thought why not. You are right that I should update the documentation. I will do that, except if you prefer that it only works with pure dims?

**Regarding stacked sum spaces:**

> For dimensions, it seems quite complex... Are stacked `SumSpace` needed?
> 
> I do not understand why you can't simplify SumSpace(SumSpace(...)) to SumSpace(...)...
> 
> I expect `direct_sum([direct_sum([A, B]), C[)` to be the same as `direct_sum([A, B, C])`.
> the same way `tensor([tensor([A, B]), C[)` is the same as `tensor([A, B, C])`.

I was thinking about writing a function that takes a state `psi` as argument, puts it into a direct sum `direct_sum([psi, ...])`, does some processing, and in the end extracts components again. For example in the goat algorithm in QOC, it constructs an initial state `direct_sum([psi, qzero_like(psi)])`, evolves it in time, and then needs to extract the two parts of the final state.

For that reason, I really wanted to ensure that `direct_component(direct_sum([psi, ...]), 0) == psi` always. If `psi` itself is a sum state, if we don't have stacked sum spaces, that would not work. Meaning that for example the goat function would have to keep track of the sum structure manually to extract the correct parts in the end.

However, you have a point that it works differently with `tensor` / `ptrace`. If you think it is not worth having the stacked sum spaces, I can change it.

> Should we avoid `Compound(SumSpace)`? How about `SuperSpace(SumSpace)`? I would expect `SumSpace` should be _final_, can't be packaged anymore.

Yes, I would avoid `Compound(SumSpace)`. I would make it so that `direct_sum([A, B]) & C == direct_sum([A & C, B & C])`. But I don't know if it is worth preventing this on the level of the dimensions.

I don't see a reason to avoid `SuperSpace(SumSpace)`. Perhaps it should be avoided if the sum space is super... on the other hand, also `SuperSpace(SuperSpace)` is currently possible.

> I also expect `direct_component` to be able to extract the `A, B` part of `direct_sum([A, B, C])`.
> Similar to how `ptrace` can keep / trace any subspace.

This is probably a good idea whether we keep stacked sum spaces or not, I will give it a try.

qutip/core/dimensions.py
Ericgig: Then why not return `Field` directly?
With `auto_tidyup_dims` it will always return that.
Without it, the order between sum order matter in an unexpected way:
  `SumSpace((SuperSpace, Field)).scalar_like() != SumSpace((Field, SuperSpace)).scalar_like()`.
Returning `Field` would make it at least consistent.

Ericgig:
For `basis`, as long as it's well documented, there is no reason limit to pure states.

I would prefer non-stacked SumSpace, just a single extra layer over what we already have, keeping it simple.
But the use case is understandable...

pmenczel:
I have updated the PR, sorry for the delay. I have added documentation for `basis` and a way to select multiple components in `direct_component`. I have implemented it by accepting slice instead of int induced, but I'm open for alternative suggestions.

Regarding the nested sum spaces, after a chat with @hodgestar and some more consideration, I would prefer to keep it as it is, if there are no strong objections. Note that also `direct_sum_sparse` wouldn't work as it currently does without nested spaces.

qutip/core/cy/qobjevo.pyi
```diff
@@ -39,7 +39,7 @@ class QobjEvo:
     def arguments(self, new_args: dict[str, Any]) -> None: ...
     @overload
     def arguments(self, **new_args) -> None: ...
-    def compress(self) -> QobjEvo: ...
+    def compress(self, *, _skip_coeff=False, _skip_qobj=False) -> QobjEvo: ...
```
Ericgig: pyi files are used for types hint for cython functions.
```suggestion
    def compress(self, *, _skip_coeff: bool = False, _skip_qobj: bool = False) -> QobjEvo: ...
```

qutip/core/dimensions.py
```diff
@@ -365,6 +375,40 @@ def __call__(cls, *args: SpaceLike, rep: str = None) -> "Space":
         if cls is SuperSpace:
             args = (*args, rep or 'super')
 
+        if cls is SumSpace:
```
Ericgig: Not a fan of adding too much code here. Ideally we should not need to modify this function to add new space types.

How about:
- changing the signature to `__call__(cls, *args: SpaceLike, **kwargs) -> "Space"`.
- Make the fast path check `elif len(args) == 1 and isinstance(args[0], Space) and not kwargs:`
- Add the class method `_process_args(*args, ***kwargs) -> tuple | (cls, dict)`.
  That function does sanity check on the kwargs and convert them to tuple, but if it detect that another type is better, (Compound to Field) it return the new class and kwargs to loop back.

That way each subclasses has it's own logic instead of pushing everything here.

pmenczel: This sounds like a great idea. I have tried implementing it, please have a look. I moved most code from the metaclasses into `_process_args` methods, and in the end, the same metaclass works for `Space` and `Dimensions`.

pmenczel:
I thought initially that it would be best to have the changes to `dimensions.py` and the new `direct_sum.py` in the same PR, but you ahve so far not commented on the `direct_sum` module. Should I take it out of this PR and make another one after?

Ericgig:
> I thought initially that it would be best to have the changes to `dimensions.py` and the new `direct_sum.py` in the same PR, but you ahve so far not commented on the `direct_sum` module. Should I take it out of this PR and make another one after?

It's as you wish, it's certainly could be moved out.

pmenczel:
> It's as you wish, it's certainly could be moved out.

I have taken it out.

qutip/tests/core/test_dimensions.py
```diff
@@ -229,11 +238,119 @@ def test_super(self, base, expected):
...
+    def test_drop_scalar_dims(self, space, expected):
+        assert space.drop_scalar_dims() == expected
+
+    @pytest.mark.parametrize(
+            "space_container", ["list", "tuple", "homtuple", "varargs"])
+    @pytest.mark.parametrize(
+            "space_type", ["none", "one-sum", "one-other", "hom", "inhom"])
+    @pytest.mark.parametrize(
+            "repeat", [None, 1, 3, 0, -1, 3.14])
+    def test_construct(self, space_container, space_type, repeat):
```
Ericgig: Could you split this test into valid / error cases, etc.
It's quite hard to follow what is tested.

qutip/core/dimensions.py
```diff
@@ -649,8 +667,218 @@ def scalar_like(self) -> Space:
         return Space([space.scalar_like() for space in self.spaces])
 
 
+class _homtuple(Sequence):
```
Ericgig: Can you move these before the meta dimensions or in an utils file, not between Space classes.

qutip/core/dimensions.py
```diff
@@ -322,68 +323,85 @@ def einsum(subscripts, *operands):
...
+            return SuperSpace, args, kwargs
+        if len(args) > 1 and all(isinstance(arg, Space) for arg in args):
+            # list of space: tensor product space
+            return Compound, args, kwargs
 
-    def from_list(
+        return cls, args, kwargs
+
+    @classmethod
+    def _process_list(
```
Ericgig: Do you dislike my names?

qutip/core/dimensions.py
```diff
@@ -322,68 +323,85 @@ def einsum(subscripts, *operands):
...
-            if cls is Compound and all(arg.size == 1 for arg in args):
-                cls = Field
-            elif cls is SuperSpace and args[0].type == 'scalar':
-                cls = Field
-
-        args = tuple([
-            tuple(arg) if isinstance(arg, list) else arg
-            for arg in args
-        ])
+class DimensionsMeta(type):
```
Ericgig: Did you rename this just to bug me?

qutip/core/dimensions.py
```diff
@@ -322,68 +323,85 @@ def einsum(subscripts, *operands):
...
+        if (
+            len(proc_args) == 1
+            and isinstance(proc_args[0], (Space, Dimensions))
+        ):
+            # Already a Space / Dimensions
+            rep = proc_kwargs.pop('rep', None)
+            if rep:
+                return proc_args[0].replace_superrep(rep)
+            else:
+                return proc_args[0]
```
Ericgig, line 398: There should be a check for extra kwargs.
Presently with `Space((2,), repeat=2)`, `repeat` quietly ignored...
Or if `SuperSpace(SuperSpace(...), superrep="choi")`, will not raise error nor update the representation.


```suggestion
        if proc_kwargs:
            # _process_args should have eaten all valid arguments.
            raise ValueError("Unsupported arguments")

        if (
            len(proc_args) == 1
            and isinstance(proc_args[0], (Space, Dimensions))
        ):
            # Already a Space / Dimensions
            return proc_args[0]
```

qutip/core/dimensions.py
```diff
@@ -649,8 +667,218 @@ def scalar_like(self) -> Space:
...
+    # SumSpace(SumSpace(...)) to SumSpace(...). (b) We expect that
+    # direct_sum(ket, scalar) always works, even if ket is a direct sum itself.
+    # Therefore, SumSpace(Field()) must simplify to Field().
+    #
+    # Simplifying if x is not a SumSpace itself allows the maximum flexibility
+    # while still allowing direct_component to function.
+
+    @classmethod
+    def _process_args(cls, *args, repeat=None, **kwargs):
+        if repeat is not None and repeat == 1:
```
Ericgig:
```suggestion
        if repeat == 1:
```

qutip/core/dimensions.py
```diff
@@ -649,8 +667,218 @@ def scalar_like(self) -> Space:
...
+
+        self.spaces = spaces
+        self._space_dims = _map_tuple(lambda space: space.size, spaces)
+        if self._repeat:
+            size = spaces[0].size * self._repeat
+        else:
+            self._space_cumdims_array = np.cumsum((0,) + self._space_dims)
+            size = self._space_cumdims_array[-1]
+
+        super().__init__(size)
```
Ericgig: Why not simple set the size here? You overwrite most of what `Space.__init__` set.
And `__setitem__` is an error it should have been `__setattr__`. So it can be skipped for now.

Ericgig (commented):
I would have expected that 
`Space((2,), repeat=3) -> Sum((Space(2),) * 3)`
and  maybe `Space([2], repeat=3) -> Compound([2, 2, 2])`.

Is it needed to build `SumSpace` from list? I feel it would be safer to build them always explicitly. Sometime tuple and list are used interchangeably, there are some issue with users using tuple instead of list to build QobjEvo.

qutip/core/dimensions.py
pmenczel: No. Obviously not 😢

To clarify: there used to be two metaclasses and now there's only one. I tried to come up with a fitting name for the one new metaclass. I realize now that the new name was almost the same as one of the old names, which is weird I guess. I will change the name to `MetaDims`

qutip/core/dimensions.py
pmenczel: I thought it was considered best practice to always call the parent's `__init__`, because for example the parent's `__init__` might change in the future? But I have changed it as you suggested.

qutip/core/dimensions.py
pmenczel: Currently, extra "rep" kwarg is already quietly ignored, and I think we want to keep it that way. I now raise a ValueError if there are extra/unsupported kwargs, except in the case of "rep".

pmenczel:
> I would have expected that `Space((2,), repeat=3) -> Sum((Space(2),) * 3)` and maybe `Space([2], repeat=3) -> Compound([2, 2, 2])`.

The second one seems like it might get complicated, I have left "repeat" only for sums for now. The first one works now, `Space(anything, repeat=x) -> SumSpace(Space(anything), repeat=x)`.

> Is it needed to build `SumSpace` from list? I feel it would be safer to build them always explicitly. Sometime tuple and list are used interchangeably, there are some issue with users using tuple instead of list to build QobjEvo.

Currently, it accepts varargs, tuple (or homtuple), and list. I would like to keep varargs and tuple; I only added the option for list because I thought it would make sense to treat list and tuple the same here. I would be okay to not accept list and raise an error instead if a list is passed. Could you confirm that that's what you meant, then I will make the change.

(Other than that, the PR is ready for review.)

Ericgig:
> > Is it needed to build `SumSpace` from list? I feel it would be safer to build them always explicitly. Sometime tuple and list are used interchangeably, there are some issue with users using tuple instead of list to build QobjEvo.
> 
> Currently, it accepts varargs, tuple (or homtuple), and list. I would like to keep varargs and tuple; I only added the option for list because I thought it would make sense to treat list and tuple the same here. I would be okay to not accept list and raise an error instead if a list is passed. Could you confirm that that's what you meant, then I will make the change.

I was speaking about not using `Space((2, 2))`, but only `SumSpace((2, 2))` to create `SumSpace`. I feel it would be safer to have `Space` not accept a tuple in case some mix list and tuple. 

For `SumSpace` I don't mind either way as it's new and clear what will be produced.

qutip/core/dimensions.py
Ericgig: I don't know if we should keep it that way, but let's leave it as is and discuss it later if / when we get to it.

pmenczel:
Thank you for your reviews, I will merge and move to the next part.

Regarding `Space(tuple)`: ah, I didn't understand your previous comment. I do like that the output of `qobj.dims` can be used as the input for `basis`, `rand_ket` and similar functions, but I also see the concern about typos. Maybe something to briefly chat about at the workshop, but I would probably be okay to not accept it.

PR #2792, opened 2025-12-08
Changing to src layout

Ericgig:
**Description**
During 2024's meeting, there was some discussion about changing from flat to src layout, to make it easier for new developers. 
This will cause conflict with other PRs, so I made the PR in a new branch for it (restored `dev.major` out of lack of imagination for names.)

I also restructured the tests: moved them outside the src package, inserted the config inside pyproject.toml and removed `testing.py`.


pyproject.toml
```diff
@@ -74,7 +74,8 @@ full = [
 ]
 
 [tool.setuptools]
-packages = { find = { include = ["qutip*"] } }
+packages = { find = { include = ["src/qutip*"] } }
+package-dir = { "" = "src" }
```
pmenczel, line 78: This didn't work for me, the following might be better:
```toml
[tool.setuptools.packages.find]
where = ["src"]
include = ["qutip*"]
```

(taken from [here](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages))

pmenczel (changes requested):
There might be a problem with the changed `pyproject.toml`. I have tried installing qutip from source from this branch:
* As an **editable install** (`pip install -e` or `setup.py develop`) things seem to work generally. There are 6 failed tests for me, which might however be stuff that we already fixed on the master branch. There is 5 times "TypeError: 'float' object cannot be interpreted as an integer" in `test_mesolve.py::TestMESolveStepFuncCoeff`, and there is an "AssertionError:  Not equal to tolerance..." in `/test_sesolve.py::TestSeSolve::test_sesolve_method[lsoda-unitary]`. I can provide more details if needed.
* As a **regular install** (with pip or with `setup.py install`), the qutip installation is empty except for `.so` files. See the following log:

```python
In [1]: import qutip

In [2]: qutip.about()
Cell In[2], line 1
----> 1 qutip.about()

AttributeError: module 'qutip' has no attribute 'about'
```

```bash
(tmp-env) paulm@Paul-XPS:~/qutip$ ls /home/paulm/miniconda3/envs/tmp-env/lib/python3.14/site-packages/qutip
_distributions.cpython-314-x86_64-linux-gnu.so  core  piqs  solver
(tmp-env) paulm@Paul-XPS:~/qutip$ ls /home/paulm/miniconda3/envs/tmp-env/lib/python3.14/site-packages/qutip/core/
_brtensor.cpython-314-x86_64-linux-gnu.so  _brtools.cpython-314-x86_64-linux-gnu.so  cy  data
```

I have suggested a change to the `pyproject.toml` below. After this change, also the regular install works for me (with the same 6 failing tests).

Minor documentation notes:
* In guide-basics.rst, we could remove the "Do not run QuTiP from the installation directory." warning?
* In development/contributing.rst, section "Testing", there are some wrong paths I think.

PR #2800, opened 2026-01-12
Add support for scipy 1.17

Ericgig:
**Description**
Scipy 1.17 released during the weekend. Here is the support patch.
Changes that affect us:
- `scipy.sparse.linalg.expm` doesn't work on dense array anymore...
- In the DOP853 integrator, the number of step isn't converted to int automatically. (`nsteps=1e4` now raises an error).
- With the LSODA integrator, the state is now modified inplace. All other integration methods returned a copy so we did not made one in the solver.

### DIFF
--- a/.github/workflows/tests_env.yml
+++ b/.github/workflows/tests_env.yml
@@ -58,4 +58,4 @@ jobs:
       - name: Run test
         run: |
           cd ../${{ matrix.repo }}
-          pytest tests -vv
+          pytest tests -vv -W "ignore::DeprecationWarning"


.github/workflows/tests.yml
```diff
@@ -105,6 +105,7 @@ jobs:
             python-version: "3.11"
             numpy-build: ">=2.0.0"
             numpy-requirement: ">=1.25,<1.26"
+            scipy-requirement: "==1.16.2"
```
hodgestar: Curious about why the SciPy version is pinned here.

qutip/core/data/expm.py
```diff
@@ -104,6 +104,7 @@ def expm_dense(matrix: Dense) -> Dense:
 expm.add_specialisations([
     (CSR, CSR, expm_csr),
     (CSR, Dense, expm_csr_dense),
+    (Dia, Dense, expm_csr_dense),
```
hodgestar: I wouldn't mind if we duplicated the `expm_csr_dense` function and named the second copy `expm_dia_dense`. It's surprising that `expm_csr_dense` has to also support Dia.

.github/workflows/tests.yml
Ericgig: Numpy 1.25 is too old for scipy 1.17, but it still install and only raise a warning:
`UserWarning: A NumPy version >=1.26.4 and <2.7.0 is required for this version of SciPy (detected version 1.25.2)`.
1.16.2 is the latest pre 1.17 version and we should keep it in a test for a while.

Ericgig: Changing it to any 1.16.X version.

Ericgig:
The failing macos test seems related, it's new from the release and consistent, but I can't reproduce it locally to make a fix...

PR #2802, opened 2026-01-14
Adjoint matmul in QobjEvo

amilsted:
This PR introduces a matrix-form master-equation solver variant, which has dramatically better performance when operators are dense, helping to close the gap to other packages such as QuantumOptics.jl and dynamiqs. It also adds some new "cast-complex-to-double" C++ low-level matmul routines, inspired by existing qutip matvec code, that vectorize better than the existing cython code (at least they are significantly faster on my arm Macbook Pro and, for CSR matrices, help to bring performance closer to the CSC routines used by QuantumOptics.jl).

The matrix-form master-equation solver introduces a new "rhs" class `LindbladMatrixForm` that uses matmul operations to implement the action of the Liouvillian on a density matrix without constructing explicit superoperator matrices. For dense operators of dimension n, this action scales as O(n^3) versus O(n^4) for the current "explicit superoperator matrix" approach. I have also found cases where using CSR representation with the matrix-form solver is faster than both CSR explicit superoperators and fully-dense matrix-form solves.

The matrix-form solver is currently opt-in via the `mesolve` API and is also usable via the new `MESolverMatrixForm` Solver class.

I am happy to provide benchmark code for both claimed improvements.

I worked with gen AI tools to produce this PR. I reviewed AI-generated changes in detail myself and iterated on the changes to ensure a good fit with existing paradigms in the qutip codebase and its tests.

I realize this is a fairly significant addition and fully expect to make possibly major adjustments during the review process! For example, should it be "LindbladMatrixForm" or "LiouvillianMatrixForm"? : )

**Update:** I split the original PR into two. This PR now just introduces underlying matmul routine changes. A second PR will bring in the master-equation solver.

### DIFF
--- a/doc/changes/2802.feature.rst
+++ b/doc/changes/2802.feature.rst
@@ -1,0 +1,1 @@
+Add adjoint (dag) in-place matmul for QobjEvo and optimize matmul ops on Apple Silicon.


qutip/solver/mesolve.py
```diff
@@ -280,3 +296,203 @@ def StateFeedback(
...
+
+class MESolverMatrixForm(Solver):
+    """
+    Master equation solver using matrix form of Lindblad equation.
+
+    Instead of building an n²×n² Liouvillian superoperator and vectorizing
+    the density matrix, this solver keeps ρ as an n×n matrix and computes
+    the RHS using matrix-matrix products:
+
+        dρ/dt = -i[H,ρ] + Σᵢ(cᵢρcᵢ† - ½{cᵢ†cᵢ,ρ})
```
pmenczel: You can add LaTeX equations also in docstring comments using `.. math::`. We avoid non-ascii characters.

doc/guide/dynamics/dynamics-options.rst
```diff
@@ -49,3 +49,20 @@ the :func:`.mesolve` and :func:`.mcsolve` function::
...
 
     >>> MCSolver(H0, c_op_list, options=options)
+
+
+Matrix-form solver option
+=========================
+
+The :func:`.mesolve` function supports a special ``matrix_form`` option that
+selects between the superoperator-based solver (:class:`.MESolver`) and the
+matrix-form solver (:class:`.MESolverMatrixForm`):
```
pmenczel: Have you considered adding the functionality to `MESolver` instead? The code would be clunkier, but it might be less surprising to the user. And I could imagine that `MESolver(.., options={"matrix_form": True})` will be easier to keep for backwards compatibility in the future than `MESolverMatrixForm(..)`. (But I am not sure about this.)

qutip/core/cy/_element.pyx
```diff
@@ -618,19 +676,19 @@ cdef class _ProdElement(_BaseElement):
         cdef double complex out = self._left.coeff(t) * self._right.coeff(t)
         return conj(out) if self._conj else out
 
-    cdef Data matmul_data_t(_ProdElement self, t, Data state, Data out=None):
+    cdef Data matmul_data_t(_ProdElement self, t, Data state, Data out=None, double complex scale=1):
```
pmenczel, line 681: If this overrides `matmul_data_t`, shouldn't it probably also override `adjoint_rmatmul_data_t`?

qutip/core/cy/lindblad_matrix_form.pyx
```diff
@@ -0,0 +1,230 @@
...
+    Computes the Lindblad master equation RHS in matrix form.
+
+    Instead of building an n²×n² superoperator and vectorizing the density
+    matrix, this class keeps ρ as an n×n matrix and computes the RHS using
+    matrix-matrix products:
+
+        dρ/dt = -i[H,ρ] + Σᵢ(cᵢρcᵢ† - ½{cᵢ†cᵢ,ρ})
+
+    This class mimics the QobjEvo interface (specifically the matmul_data
+    method) so it can be used as a drop-in replacement in solvers.
```
pmenczel: Clever :-)

qutip/solver/mesolve.py
```diff
@@ -280,3 +296,203 @@ def StateFeedback(
...
+
+    This approach can be more memory-efficient for large systems and avoids
+    building the full superoperator. For dense operators, the time complexity
+    also scales better with system size (n³ versus n⁴).
+    However, it requires more matrix multiplications per RHS evaluation.
+
+    **Note:**
+    - State must be kept as dense n×n matrix (no vectorization)
```
pmenczel: It wasn't quite clear to me what this comment means exactly, just that the initial state is not allowed to be vectorized?

pmenczel (commented):
Hi @amilsted, thank you very much for your contribution. It will take time to look at this in detail, but it looks like very good and thorough work. Benchmark code would be nice to have (and could potentially be repackaged into a tutorial notebook at some point?)

Storing superoperators not as a big Liouvillian matrix but in a more "analytic" form has been discussed internally for a while, and is a big goal for one of the next major releases. That would be a deeper change, applying not only to Lindblad-type superoperators... however, it will take some time. I imagine that it would make sense to add your feature to QuTiP now, even if it might become obsolete at some later time. We need to have some discussion within the admin team.

The following are just a few things I noticed scrolling through the code for the first time. (Also, please have a look at the warnings generated in the test runs; we treat warnings as errors.)

amilsted:
Looks like tests are failing because of warnings. I added coverage for the in-place behavior of the matmul routines, including cases where input and output matrices have different C/Fortran ordering choices (some of the existing code was actually broken for this case). These paths are slower and currently warn. Maybe these should actually be error paths? In any case, at the moment it looks like I just need to teach the tests to expect warnings here.

qutip/solver/mesolve.py
amilsted: That's all this means. A case of "overly explicit AI comments", I think.

amilsted:
> Storing superoperators not as a big Liouvillian matrix but in a more "analytic" form has been discussed internally for a while, and is a big goal for one of the next major releases. That would be a deeper change, applying not only to Lindblad-type superoperators... however, it will take some time. I imagine that it would make sense to add your feature to QuTiP now, even if it might become obsolete at some later time. We need to have some discussion within the admin team.

Thank you for taking a look! It did occur to me that it might make sense to have a general matrix-form Liouvillian class. I hope that the matmul op work here is useful for future changes along these lines. It wouldn't be hard to generalize LindbladMatrixForm to be a general superoperator class. Could keep the same matmul efficiency if we e.g. handle `[H_nh, None]` and `[None, H_nh]` by treating `None` as a noop identity.

amilsted:
Maybe I should also add: the explicit superoperator form for the Liouvillian is still often the fastest for sparse martrices, a according to my benchmarks, so you may not want to drop it entirely!

Ericgig:
@amilsted 
With the size of the PR, would you accept to split it into chunk for easier review?
As a first step, only include the core changes with tests, (except the new lindblad_matrix_form.pyx.)
This should cut it in about half.

qutip/core/cy/_element.pyx
```diff
@@ -162,21 +165,76 @@ cdef class _BaseElement:
...
+
+        Parameters
+        ----------
+        t : double
+          The time, ``t``.
+
+        state : :obj:`~Data`
+          The state to multiply by the adjoint of the element term from the right.
+
+        out : :obj:`~Data` or ``NULL``
```
Ericgig: Let use the python term while we are at it.
```suggestion
        out : :obj:`~Data` or ``None``
```

qutip/core/cy/_element.pyx
```diff
@@ -162,21 +165,76 @@ cdef class _BaseElement:
...
+            data_t = self.data(t)
+            imatmul_dag_dense_data(<Dense>state, data_t, total_scale, <Dense>out)
+            return out
+        else:
+            # General case for non-Dense matrices
+            data_t = self.data(t)
+            data_adj = data_t.adjoint()
+            
+            if out is None:
+                return _data.matmul(state, data_adj, total_scale)
```
Ericgig: It would be better to use `_data.matmul_dag` here and let it choose how to do the computation. 
That way, optimized operation for backend (JAX) can be added without changing this.

qutip/core/cy/_element.pyx
Ericgig: I think it fine without `adjoint_rmatmul_data_t`.
This is a faster operation for vector: do `A @ (B @ v)` instead of `(A @ B) @ v`.
But `adjoint_rmatmul_data_t` is to be used with density matrix, so there is little to gain.

qutip/core/data/matmul.pyx
```diff
@@ -229,13 +247,13 @@ cpdef Dense matmul_csr_dense_dense(CSR left, Dense right,
         # the user and then transpose one of the arrays.  We prefer to have
         # `right` in Fortran-order for cache efficiency.
         if right.fortran:
-            tmp = out
-            out = out.reorder()
+            out_add = out
```
Ericgig, line 250: If I am right, the error you found in mix format was in this case.
To ensure to return the returned instance was the input out, we renamed it to `tmp`, did the computation in "F" mode and copied the buffer without reordering...

This way is quite cleaner. 
Thank you for catching and fixing.

qutip/core/data/matmul.pyx
```diff
@@ -865,6 +849,122 @@ cpdef Dia multiply_dia(Dia left, Dia right):
     return out
 
 
+cpdef Dense matmul_dag_dense_dia_dense(
+    Dense left, Dia right,
+    double complex scale=1, Dense out=None
+):
+    """
+    Compute out = scale * (left @ right†) where right is DIA.
```
Ericgig: I would prefer to only use characters easy to type on any keyboard:
`right.dag`, `right.adjoint`, `dag(right)`, etc.

Ericgig:
Thank you for adding mixed C, F order in the tests.

It would be good to have tests that check the warnings are properly raised:
`with pytest.warns(OrderEfficiencyWarning) as warning:`

I don't think they should be error. 
The design of the low level function is that it should just work, even if not optimal.

qutip/core/cy/_element.pyx
amilsted: Ah, good point.

qutip/core/data/matmul.pyx
amilsted: Yes, this was where I found the bug. There were other routines that already used the `out_add` approach so I just adopted it here too.

amilsted:
> Thank you for adding mixed C, F order in the tests.
> 
> It would be good to have tests that check the warnings are properly raised: `with pytest.warns(OrderEfficiencyWarning) as warning:`
> 
> I don't think they should be error. The design of the low level function is that it should just work, even if not optimal.

I added some code to just swallow these warnings in all cases. It's currently difficult to detect when they should or shouldn't occur, but I can work on doing that if you prefer.

I also did some refactoring as I wasn't quite happy with how the tests were structured. Now there is a separate test variant with its own id for out-of-place as well as each in-place ordering.

qutip/solver/mesolve.py
amilsted: Deferring mesolve to a second PR. Resolved in the other branch.

doc/guide/dynamics/dynamics-options.rst
amilsted: Deferring mesolve to a second PR.

qutip/solver/mesolve.py
amilsted: Deferring mesolve to a second PR.

qutip/core/cy/_element.pyx
amilsted: I'm matching the matmul_data_t docstring above. Shall I change that one too?

amilsted:
> @amilsted With the size of the PR, would you accept to split it into chunk for easier review? As a first step, only include the core changes with tests, (except the new lindblad_matrix_form.pyx.) This should cut it in about half.

I have now done this. This PR now only includes the matmul stuff. I will make a second for the mesolve stuff. Would you like me to do that now or wait until we are done with this?

For now, I have put the remaining changes here: https://github.com/amilsted/qutip/pull/1

qutip/core/data/matmul.pyx
amilsted: Done

qutip/core/cy/_element.pyx
amilsted: Done

qutip/solver/mesolve.py
amilsted: Resolved in the other branch.

doc/guide/dynamics/dynamics-options.rst
amilsted: We can discuss this later too, but it occurred to me that MESolver inherits from SESolver, which makes at least some sense when we are vectorizing rho, but perhaps less if we are not. Then again, SESolver does not do very much. MESolver overrides most of it, so maybe we could just not do that inheritance for clarity? I can't see any obvious reasons for the inheritance, although I guess it is potentially breaking for users to change this.

In any case, building the feature into MESolver doesn't seem too hard. There also seem to be some simplifications possible that would reduce LOC a bit, like putting a `vectorize_dm` flag into Solver so we can act on it in the _prepare_state and _restore_state methods.

amilsted:
Added some benchmark results in a branch here: https://github.com/amilsted/qutip/tree/matmul_bench

See the full results here: https://github.com/amilsted/qutip/blob/matmul_bench/BENCHMARKING.md

In matmul microbenchmarks:
* CSR matrix-matrix operations get significantly faster on Linux x86_64 and Apple Silicon (15-30%). This is important for the matrix-form solver in the upcoming second PR!
* CSR matrix-vector operations get significantly faster on Apple Silicon (15-20%) and are about the same on x86_64 (and shouldn't change as the SSE specialized routine is still used unchanged here)
* DIA matrix-matrix and matrix-vector gets significantly faster on Apple Silicon (15-30%). Matrix-vector is somewhat slower on x86_64 (~5%) and matrix-matrix is about the same. The slowdown seems to be due to use of intermediate variables to avoid writing to vectors all the time in the inner loop. This significantly helps Apple Silicon (more registers?), but slightly hurts x86_64.

In JC model dynamics (sesolve and vectorized mesolve, so only exercising **mat-vec** routines):
* ~5% performance improvement for CSR on Apple Silicon, up to ~15% with DIA.
* Essentially no change on x86_64 for most cases. Perhaps a small slowdown for very small Hilbert spaces, but could be noise.

qutip/core/cy/_element.pyx
Ericgig: As you wish. If you don't, I will do a quick fix once this is merged.

nwlambert:
Just wanted to add, thanks for this, looks cool!   I had a quick question, I remember looking at what dynamiqs were doing a while back, and they had a nice trick in the Lindblad construction to slightly reduce the number of matrix-matrix multiplications, at the the cost of some small numerical error (https://github.com/dynamiqs/dynamiqs/blob/fdc6c8913bba2b0a03eb0ad314ca01b5a519030b/dynamiqs/integrators/core/diffrax_integrator.py#L297 ).   Just curious if you tried this in your benchmarks, and if it was worthwhile.

amilsted:
> Just wanted to add, thanks for this, looks cool! I had a quick question, I remember looking at what dynamiqs were doing a while back, and they had a nice trick in the Lindblad construction to slightly reduce the number of matrix-matrix multiplications, at the the cost of some small numerical error (https://github.com/dynamiqs/dynamiqs/blob/fdc6c8913bba2b0a03eb0ad314ca01b5a519030b/dynamiqs/integrators/core/diffrax_integrator.py#L297 ). Just curious if you tried this in your benchmarks, and if it was worthwhile.

Oh man! Yes, I had come across this, but only vaguely remembered it existed. I think it's worth doing for the second PR :)

Slight changes to the number of matmul ops can indeed determine the answer to "which solver variant is best". There are many cases where it is a close call!

qutip/tests/core/data/test_mathematics.py
```diff
@@ -267,6 +269,23 @@ def case(type_, shape_case, generator_case):
...
+        try:
+            sig = inspect.signature(op)
+            supports_out = 'out' in sig.parameters
+        except (ValueError, TypeError):
+            # Can't introspect (e.g., built-in function), assume no out support
+            pass
+
+    # Determine out_prealloc values
+    out_prealloc_values = [None, "C", "F"] if supports_out else [None]
+
```
Ericgig: This is not the way...
This will cause problem if the `out` argument was ever used in another way or named something else.
It would also skip test were the output is not Dense ,cupy , jax, cuQuantum backends uses non-Dense states in our solvers and we should test them too.

Why not use `TernaryOpMixin` (or `ScaledTernaryOpMixin`) with `out` as simply the third input:
the operation is really `A @ B * scale + out`. With the output being the same instance as out a bonus.
The function tested will have to be wrapped since `scale` comes before `out` (or we could inverse them in the functions).
It would allow to add tests form other specialization easily (says we want to test with a CSR out) and keep `test_mathematically_correct` and `cases_type_shape_product` simpler.

qutip/tests/core/data/test_mathematics.py
```diff
@@ -507,18 +546,87 @@ def test_incorrect_shape_raises(self, op, data_l, data_r):
...
+            if scale is not None:
+                test = (op(left, right, scale, out=out) if out is not None
+                        else op(left, right, scale))
+            else:
+                test = (op(left, right, out=out) if out is not None
+                        else op(left, right))
+
+        if out is not None:
+            assert test is out
```
Ericgig: This can be skipped if tricky to add.
Since the initial implementation, we moved to having "inplace" is understood as the function can reuse the memory space of the `Data` object given if appropriate, but the output should always be used.

Ericgig (commented):
I am still running benchmarks, but I am happy with everything except the tests for cases with inputs. Auto detection from signature and adding an argument that you must check that it's not used anywhere except one specific case is not a clean solution.

qutip/tests/core/data/test_mathematics.py
amilsted: Point taken. That sounds like a better approach!

qutip/core/data/src/matmul_diag_vector.cpp
```diff
@@ -10,15 +10,31 @@ void _matmul_diag_vector(
...
-        out_dbl[i+1] += data_dbl[i] * vec_dbl[i+1];
-        out_dbl[i+1] += data_dbl[i+1] * vec_dbl[i];
+    double prod_re, prod_im;
+
+    if (scale == std::complex<double>(1.0, 0.0)) {
+        for (IntT i=0; i<length*2; i+=2){
+            prod_re = data_dbl[i] * vec_dbl[i] - data_dbl[i+1] * vec_dbl[i+1];
+            prod_im = data_dbl[i] * vec_dbl[i+1] + data_dbl[i+1] * vec_dbl[i];
+            out_dbl[i] += prod_re;
+            out_dbl[i+1] += prod_im;
```
Ericgig: In my benchmark, on linux, (gcc 11), this is 5% slower...
Asking an AI why, I got a plausible explanation with a comment that the compiling options `-ffast-math` allow the compiler to automatically use SIMD for complex operation... (haven't tried it yet).

Anyway, the 20% speedup on Mac is more impressive than the 5% slowdown on Linux, but since HPC cluster mostly use linux, I think most users will experience the slowdown...
Ideally, I would like to get the best of both world if possible.
I will try to play with the compilation option tomorrow.

qutip/core/data/matmul.pyx
```diff
@@ -551,19 +551,12 @@ cpdef Dense matmul_dense_dia_dense(Dense left, Dia right, double complex scale=1
...
+              out.data[(start_right + i) * strideC_out + row * strideR_out] += (
+                scale
+                * right.data[diag * right.shape[1] + i + start_right]
                 * left.data[(start_left + i) * strideC_in + row * strideR_in]
               )
 
-    if out is None and scale == 1.:
-        out = tmp
-    elif out is None:
-        imul_dense(tmp, scale)
```
Ericgig, line 562: The case `matmul_dia_dense_dense(A, B, scale!=1, out=None)` is ~ 50% slower now.
The scale is applied in a triple loop (col, row, num_diag) instead of only in a double loop here...
Since solver provide an out matrix and that case is faster, I think it's fine to let it slide.

qutip/core/data/src/matmul_diag_vector.cpp
amilsted: Yeah, as I mentioned this is probably because of the prod_re/im intermediate variables. This is what makes the Apple Silicon version faster. Removing these and working directly with the arrays seems to be slightly faster on x86_64 with gcc. It would be pretty easy to have both versions, depending on the arch, I suppose?

Ericgig: If you are comfortable with it, yes having both version would be nice.

qutip/tests/core/data/test_mathematics.py
amilsted: Actually, on reflection, I think I don't understand. Is your statement about the term "inplace" or the matmul functions specifically? In the matmul functions, are you saying it is in principle okay for them to return a Data object that is different to the `out` argument? I suppose this might apply for non-dense output?

qutip/core/data/src/matmul_diag_vector.cpp
amilsted: I have done this now, but my testing is not showing much benefit. Would be interested in your findings! It's strange, as the scale==1 path is now very similar to master.

qutip/tests/core/data/test_mathematics.py
amilsted: I have done this now with a `InPlaceMatmulMixin`, since the semantics of `TernaryOpMixin` were a bit different. It does seem a lot cleaner. What do you think?

This time I have verified that the rest of the tests also pass :)

qutip/core/data/src/matmul_diag_vector.cpp
amilsted: I even tried separating the scale == 1 case into a separate function, but it didn't make a measurable difference.

qutip/core/data/matmul.pyx
amilsted: I find that somewhat surprising! I had expected these operations to be memory bound, so that reading tmp back through would be worse than doing a little more compute within the diag op. But I guess we read the Dense array num_dia times anyway. I wonder if another approach that doesn't use an outer loop over the diags would be better overall...

qutip/core/data/src/matmul_diag_vector.cpp
amilsted: So, Claude did the leg work and identified the main culprit here. Due to the addition of the scale parameter in the QobjEvo matmul method, we were doing a `*` operation with a python object. I just committed a fix for this which restores DIA matmat and matvec performance on x86 for scale==1.

qutip/core/data/matmul.pyx
amilsted: Ok, so it turns out the "separate scaling loop" approach is better overall in tests. I have reverted the on-the-fly scaling for DIA. I am still a little surprised that allocating a tmp buffer in each matmul does not kill the speed, but that is indeed what the benchmarks say!

amilsted: I suppose this is the only allocation we are doing in a solve loop in efficient cases, so the allocator should stay warm between calls.

qutip/core/data/matmul.pyx
```diff
@@ -494,6 +509,11 @@ cpdef Dense matmul_dia_dense_dense(Dia left, Dense right, double complex scale=1
 
 cpdef Dense matmul_dense_dia_dense(Dense left, Dia right, double complex scale=1, Dense out=None):
     _check_shape(left, right, out)
+    # Use tmp buffer when scaling, applying the scaling in a separate pass.
+    # This is faster than applying scaling on the fly, likely because the simpler
+    # scale-free inner loop vectorizes better. The extra allocation required is
+    # worth it, accoinding to benchmarks.
+    cdef Dense tmp
```
amilsted, line 516:
```suggestion
```

qutip/core/data/matmul.pyx
amilsted: See updated results in https://github.com/amilsted/qutip/blob/matmul_bench/BENCHMARKING.md

qutip/core/cy/_element.pyx
```diff
@@ -162,18 +165,76 @@ cdef class _BaseElement:
...
             return _data.add[type(out), type(state), type(out)](
                 out,
                 _data.matmul[type(data_t), type(state), type(state)](
-                    data_t, state, self.coeff(t)
+                    data_t, state, total_scale
+                )
+            )
+
+
+
```
Ericgig:
```suggestion

```

qutip/tests/core/data/test_mathematics.py
```diff
@@ -801,25 +829,118 @@ def op_numpy(self, left, right):
...
+        assert result.shape == expected.shape
+        np.testing.assert_allclose(result.to_array(), expected,
+                                   atol=self.atol, rtol=self.rtol)
+
+
+class TestMatmulInPlace(InPlaceMatmulMixin):
+    """Test in-place matmul operations that support out parameter."""
+
+    def op_numpy(self, left, right, out):
+        return np.matmul(left, right)
```
Ericgig: Either remove this or make it the right operation...
Having it not matching the operation is confusing.

qutip/core/cy/_element.pyx
amilsted: I made the change

amilsted:
> Benchmarks look good on my side. If you could just remove the unused `op_matmul`, it's ready.
> 
> I saw you removed the towncrier entry. Knowing that there is a follow up PR, it's fine, but this alone is lot of work and deserve a mention.

Removed the op and added the towncrier entry.

Issue #2747, opened 2025-09-18
None-CSR sparse formats tends to run out of memory when used with steadystate()+large problem sizes


nwlambert:
### Bug Description

When calling 'steadystate()' for large Hilbert space sizes with operators built out of DIA, we often hit a memory error. This line:

https://github.com/qutip/qutip/blob/dbabe79ddcb5e50d97ec75dee77e93498c57a96c/qutip/solver/steadystate.py#L225

Seems to be the culprit.  Either needs an option to capture other sparse formats, or convert to CSR (which is done later in the steadystate function for DIA anyway?)

### Code to Reproduce the Bug

```shell

```

### Code Output

```shell

```

### Expected Behaviour

Sparse problems shouldn't run out of memory

### Your Environment

```shell
QuTiP Version:      5.3.0.dev0+8d2e163
Numpy Version:      1.26.4
Scipy Version:      1.14.1
Cython Version:     0.29.36
Matplotlib Version: 3.5.0
Python Version:     3.10.12
Number of CPUs:     12
BLAS Info:          INTEL MKL
INTEL MKL Ext:      /home/neill/qutip-dev310/lib/libmkl_rt.so.2
Platform Info:      Linux (x86_64)
Installation path:  /mnt/d/github/qutip/qutip
```

veronikakurth:
Hi Neill @nwlambert!
Could you please assign me to this issue? 
I would be interested in working on it and understanding the data and solver layers in QuTiP better. 
I would also appreciate if you could provide an example or give a hint to a size of a Hilbert space with which the mentioned out-of-memory error may occur. 
Thanks!

nwlambert:
about the size of the hilbert space, I dont quite remember what i was doing!  I will see if i can just make a simple example for it

nwlambert:
Around N=110 or 120 this example fails for me when using the Dia format:

```from qutip import *
N=120 #number of Fock states
H0 = sigmaz() & qeye(N) #atom energy
H1 = qeye(2) & (destroy(N).dag() * destroy(N))  #cavity energy
H2 = sigmax() & (destroy(N).dag() + destroy(N)) #interaction

H = (H0 + H1 + H2) #total Hamiltonian
print(H.dtype) #should be csr because of using sigmax() etc
c_ops = [qeye(2) & destroy(N)] #collapse operators, usually Dia
#print(c_ops[0].dtype) #should be dia    

rhoss_csr = steadystate(H, c_ops)

[...]

PR #2803, opened 2026-01-20
Fix for "None-CSR sparse formats tends to run out of memory when used with steadystate()+large problem sizes"

veronikakurth:
**Description**
Enables performing computations on sparse matrices of high dimensionalities in the "direct" method of the steady-state solver without running into out-of-memory allocation problems. 

**Related issues or PRs**
Fixes issue #2747.

### DIFF
--- a/qutip/solver/steadystate.py
+++ b/qutip/solver/steadystate.py
@@ -216,20 +216,20 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
 
 
 def _steadystate_direct(A: Qobj, weight: float, **kw):
-    # Find the weight, no good dispatched function available...
-    if weight:
-        pass
     # Convert Dia to CSR for cleaner diagonal matrix representation:
     # without zeros or uninitialised padded elements, which is especially
     # relevant for multi-diagonal cases
     if isinstance(A.data, _data.Dia):
         A = A.to("csr")
-    # Compute the weight for sparse or other matrices
-    if isinstance(A.data, _data.CSR):
-        weight = np.mean(np.abs(A.data.as_scipy().data))
-    else:
-        A_np = np.abs(A.data.to_array())
-        weight = np.mean(A_np[A_np > 0])
+
+    if not weight:
+      # Calculate weight if not provided by user
+      # (currently, no good dispatched function is available)
+      if isinstance(A.data, _data.CSR):
+          weight = np.mean(np.abs(A.data.as_scipy().data))
+      else:
+          A_np = np.abs(A.data.to_array())
+          weight = np.mean(A_np[A_np > 0])
 
     # Add weight to the Liouvillian
     # L[:, 0] = A[:, 0] + vectorized(eye * weight).T


qutip/solver/steadystate.py
```diff
@@ -215,29 +215,32 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
...
     if weight:
         pass
-    elif isinstance(A.data, _data.CSR):
-        weight = np.mean(np.abs(A.data.as_scipy().data))
+    # Calculate the weight for sparse or dense matrices
+    if hasattr(data, "as_scipy") and callable(data.as_scipy):
+        weight = np.mean(np.abs(data.as_scipy().data))
     else:
-        A_np = np.abs(A.full())
+        A_np = np.abs(data.as_ndarray())
```
Ericgig: What if data as neither `as_scipy` or `as_ndarray`? Such as `JaxArray` from qutip-jax.
When using method from Data object, unless you check the format, you can only expect that the methods from the parent `Data` class exist.

qutip/solver/steadystate.py
```diff
@@ -237,7 +238,7 @@ def _steadystate_direct(A, weight, **kw):
     weight_vec = _data.column_stack(_data.diag([weight] * n, 0, dtype=dtype))
     first_row = _data.block_extract(A.data, 0, 1, 0, N, dtype=dtype)
     L = _data.block_overwrite(
-        A.data, _data.add(first_row, weight_vec.transpose()), 0, 0, dtype=dtype
+       A.data, _data.add(first_row, weight_vec.transpose()), 0, 0, dtype=dtype
```
Ericgig: Why did you remove a space here?

veronikakurth:
> I forgot, we can't average on a Dia matrix `data` directly. The dia matrix `data` is a (num_diag, num_col) shaped array. For example, in a NxN tri diagonal matrix, the main diagonal use the full N entries in the `data` row, but the sub diagonal use N-1 values. The extra entries are not used and could contain random garbage...
> 
> ```
> qt.create(5).data_as().data
> array([[1.00000000e+000+0.00000000e+000j,
>         1.41421356e+000+0.00000000e+000j,
>         1.73205081e+000+0.00000000e+000j,
>         2.00000000e+000+0.00000000e+000j,
>         1.10520054e+165+6.50070601e-308j]])  # This value is random!
> ```

This is trickier than I expected from this issue but I see the problem now. I have one idea in mind, will keep you updated once I try it out. :) 
Thanks for the insight! 
This is also something definitely to catch in tests more often, such tridiagonal cases, for example.

veronikakurth:
Hi Eric!
I tried creating different diagonal matrices of different structures that would underlie a quantum object and checked how easy it would be to take a mean value on them correctly. My conclusion that the safest and cleanest way to proceed (without operator overloading, an importance of which became even more apparent to me while working on this issue!) is to operate on the SciPy's CSR representation (after converting Dia to QuTiP's CSR).

Anything else, e.g. representing Dia as SciPy's Dia, requires "cleaning" from zeros or padded close-to-zero values (like in your example; but I am not 100% sure whether those would always be close-to-zero values and we can reliably filter entries that don't belong into the original matrix). Moreover, to clean from unwanted values requires a dense representation, which would also always lead to memory issues in case of large systems. 

P.S. It also feels more natural to delegate this conversion to CSR into `def steadystate` but I am a bit unsure about where exactly it would fit and how it may interfere with the default dtype scope. 
Please let me know what you think. Thank you!

qutip/solver/steadystate.py
```diff
@@ -215,14 +215,20 @@ def steadystate(A, c_ops=[], *, method='direct', solver=None, **kwargs):
     return rho_ss
 
 
-def _steadystate_direct(A, weight, **kw):
+def _steadystate_direct(A: Qobj, weight: float, **kw):
     # Find the weight, no good dispatched function available...
     if weight:
         pass
```
Ericgig: If `weight` is given, it will now be overwritten right?

veronikakurth: Fair remark! I corrected the changes so that weight is computed only if it is not provided. 
I also left the comment about the unavailability of good dispatched functions: I wanted to ask what is meant by this? How exactly is the weight supposed to be passed by user? (Right now I think it's not used but just wanted to get a brief idea about it. I guess my type hinting (float) for weight might be incorrect.)

Ericgig:
The failing build documentation is not related.
Thank you  for the fix.

PR #2810, opened 2026-01-23
Update cibuildwheel

Ericgig:
**Description**
Update cibuildwheel to use recent multilinux image.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -48,4 +48,4 @@ sphinxcontrib-serializinghtml==1.1.10
 traitlets==5.14.3
 urllib3==2.6.3
 wcwidth==0.2.13
-wheel==0.44.0
+wheel==0.46.2


.github/workflows/build.yml
```diff
@@ -113,6 +113,9 @@ jobs:
       # Numpy and SciPy do not supply wheels for i686 or win32 for
       # Python 3.10+, so we skip those:
       CIBW_SKIP: "*-musllinux* *-manylinux_i686 *-win32"
+      # Force recent version of manylinux
+      # manylinux2014 no longer supported by scipy and numpy
+      CIBW_MANYLINUX_X86_64_IMAGE: manylinux_2_28
```
hodgestar, line 118: Idea: Would it be better to skip the unsupported manylinux versions as we do above for, e.g., `*-manylinux_i686`?

Ericgig: I will try. Let's see what it does.

Ericgig: It doesn't work.
We could skip all x86_64 with CIBW_SKIP (`*manylinux_x86_64`), but not multilinux2014 image...

hodgestar: Thanks for giving the idea a try!

PR #2811, opened 2026-01-28
Matrix-form master-equation solver

amilsted:
This PR introduces a matrix-form master-equation solver variant, which has dramatically better performance when operators are dense, helping to close the gap to other packages such as QuantumOptics.jl and dynamiqs.

The matrix-form master-equation solver introduces a new "rhs" class LindbladMatrixForm that uses matmul operations to implement the action of the Liouvillian on a density matrix without constructing explicit superoperator matrices. For dense operators of dimension n, this action scales as O(n^3) versus O(n^4) for the current "explicit superoperator matrix" approach. I have also found cases where using CSR representation with the matrix-form solver is faster than both CSR explicit superoperators and fully-dense matrix-form solves.

The matrix-form solver is currently opt-in via the mesolve API and is also usable via the MESolver class.

This is a follow-on PR from https://github.com/qutip/qutip/pull/2802. It incorporates some suggestions from the initial comments, including the idea to assume rho is Hermitian and use the [trick from dynamiqs](https://github.com/dynamiqs/dynamiqs/blob/fdc6c8913bba2b0a03eb0ad314ca01b5a519030b/dynamiqs/integrators/core/diffrax_integrator.py#L297) to reduce the number of matmul ops (question: should this be optional so that we can use non-hermitian "states"?).

### DIFF
--- a/qutip/core/cy/lindblad_matrix_form.pyx
+++ b/qutip/core/cy/lindblad_matrix_form.pyx
@@ -150,6 +150,7 @@ cdef class LindbladMatrixForm(QobjEvo):
                                             self._buffer_size,
                                             fortran=rho_dense.fortran)
         temp_dense = <Dense>self._temp_buffer
+        temp_dense.fortran = rho_dense.fortran
 
         # Compute A = -i H_nh @ rho + 0.5 sum(L @ rho @ Ld)
         # Then drho/dt = A + A.dag()


setup.py
```diff
@@ -216,6 +216,7 @@ def _extension_extra_sources():
             'qutip/core/data/src/matmul_csr_vector.cpp',
             'qutip/core/data/src/matmul_csr_dense.cpp',
             'qutip/core/data/src/matmul_diag_vector.cpp',
+            'qutip/core/data/src/matmul_csr_dense.cpp',
```
amilsted:
```suggestion
```

qutip/core/cy/lindblad_matrix_form.pyx
```diff
@@ -0,0 +1,244 @@
...
+
+        # Use pre-computed H_nh if provided (e.g., from unpickling)
+        if _H_nh is not None:
+            self.H_nh = _H_nh
+        # Construct non-Hermitian Hamiltonian: H_nh = H - (i/2) * sum(dag(c_i) * c_i)
+        # This allows us to write: drho/dt = -i(H_nh * rho - rho * dag(H_nh)) + sum(c_i * rho * dag(c_i))
+        elif len(c_ops) > 0:
+            H_nh = H.copy()
+            for c_op in self.c_ops:
+                H_nh = H_nh - (0.5j) * (c_op.dag() * c_op)
```
Ericgig, line 89: Since you are not using inplace operation, the copy should not be needed.

qutip/core/cy/lindblad_matrix_form.pyx
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            self.H_nh = H
+
+        self._dims = H._dims
+        self.shape = H.shape
+        self.type = 'oper'
+        self.issuper = True
+
+        # Check if all operators are time-independent
+        self.isconstant = (H.isconstant and
```
Ericgig: Why not check is `self.H_nh` isconstant?

qutip/core/cy/lindblad_matrix_form.pyx
```diff
@@ -0,0 +1,244 @@
...
+            c_op.matmul_data(t, rho_dense, temp_dense, 0.5)
+            c_op.adjoint_rmatmul_data(t, temp_dense, out_dense)
+
+        # Add Hermitian conjugate: out += out.dag()
+        # Copy out to temp, conjugate temp in-place, then add with transpose
+        # (transpose is achieved by flipping fortran flag for iadd_dense)
+        cdef size_t nbytes = n * n * sizeof(double complex)
+        memcpy(temp_dense.data, out_dense.data, nbytes)
+        iconj_dense(temp_dense)
+        temp_dense.fortran = not temp_dense.fortran  # transpose via flag
```
Ericgig: This low level code should not be here.

qutip/core/cy/lindblad_matrix_form.pyx
```diff
@@ -0,0 +1,244 @@
...
+        # -i H_nh @ rho
+        self.H_nh.matmul_data(t, rho_dense, out_dense, -1j)
+
+        # +0.5 sum(L @ rho @ Ld)
+        cdef QobjEvo c_op
+        for i in range(self.num_collapse):
+            c_op = <QobjEvo>self.c_ops[i]
+            imul_dense(temp_dense, 0)
+            c_op.matmul_data(t, rho_dense, temp_dense, 0.5)
+            c_op.adjoint_rmatmul_data(t, temp_dense, out_dense)
```
Ericgig: Even for inplace operation, the output should be used, not the side effect.

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -10,10 +10,28 @@ from qutip.core.data.mul cimport imul_data
...
+    LindbladMatrixForm, avoiding the performance penalty of object calls.
+    """
+    if type(qevo) is QobjEvo:
+        return (<QobjEvo>qevo).matmul_data(t, state, out)
+    elif type(qevo) is LindbladMatrixForm:
+        return (<LindbladMatrixForm>qevo).matmul_data(t, state, out)
+    else:
+        # Fallback for other objects that implement matmul_data interface
+        return qevo.matmul_data(t, state, out)
+
```
Ericgig: While supporting QobjEvo as the only rhs format is restrictive and we have been working around for a while in different places, this is not the way... 
The ODE solver should not have to know what kind of algorithm we are using for the "matmul" operation. If / when other mean of computing the matmul operation, for any solver / matrix format, are created, they will should be added here, making quickly a mess.

Why did you not make `LindbladMatrixForm` a `QobjEvo`? Any unused functions could simply raise a `NotImplementError`. This is the workaround used in qutip-cuquantum, etc.

Of course we need to fix this on our side, so I don't want to stop the PR for this, but I also don't want it...

qutip/core/data/adjoint.pyx
```diff
@@ -110,6 +110,15 @@ cpdef Dense conj_dense(Dense matrix):
     return out
 
 
+cpdef Dense iconj_dense(Dense matrix):
```
Ericgig: Why `iconj_dense` and not `iadjoint_dense`?

qutip/solver/mesolve.py
```diff
@@ -39,9 +39,10 @@ def mesolve(
...
 
     Evolve the state vector or density matrix (``rho0``) using a given
-    Hamiltonian or Liouvillian (``H``) and an optional set of collapse operators
-    (``c_ops``), by integrating the set of ordinary differential equations
-    that define the system. In the absence of collapse operators the system is
+    Hamiltonian or Liouvillian (``H``) and an optional set of collapse
+    operators (``c_ops``), by integrating the set of ordinary differential
+    equations that define the system. In the absence of collapse operators
+    the system is
     evolved according to the unitary evolution of the Hamiltonian.
```
Ericgig: This is unrelated to the feature, so better skip it, but if you fix it, fix it properly...
```suggestion
    the system is evolved according to the unitary evolution of the 
    Hamiltonian.
```

qutip/solver/mesolve.py
```diff
@@ -124,11 +125,16 @@ def mesolve(
...
         - | nsteps : int
-          | Maximum number of (internally defined) steps allowed in one ``tlist``
-            step.
+          | Maximum number of (internally defined) steps allowed in one
+            ``tlist`` step.
         - | max_step : float
-          | Maximum lenght of one internal step. When using pulses, it should be
+          | Maximum length of one internal step. When using pulses, it
+            should be
             less than half the width of the thinnest pulse.
```
Ericgig: Here too, the fix is strange...

qutip/solver/mesolve.py
```diff
@@ -231,12 +241,34 @@ def __init__(
...
+
+        if matrix_form:
+            if H.issuper:
+                raise TypeError(
+                    "matrix_form=True cannot be used with superoperator H"
+                )
+            self._vectorize_state = False
+            H = QobjEvo(H) if isinstance(H, Qobj) else H
+            c_ops_qevo = [QobjEvo(c) if isinstance(c, Qobj) else c
+                         for c in c_ops]
```
Ericgig: QobjEvo has a pass through if the input is already a one.
```suggestion
            H = QobjEvo(H)
            c_ops_qevo = [QobjEvo(c) for c in c_ops]
```

qutip/solver/solver_base.py
```diff
@@ -56,8 +59,11 @@ class Solver:
     def __init__(self, rhs, *, options=None):
         if isinstance(rhs, (QobjEvo, Qobj)):
             self.rhs = QobjEvo(rhs)
+        elif hasattr(rhs, 'matmul_data') and hasattr(rhs, '_register_feedback'):
+            # Duck-typed RHS (e.g., LindbladMatrixForm)
+            self.rhs = rhs
```
Ericgig: Let's not start adding duck-typing. It will only make it harder for us in the future.
If we need a new attribute of the rhs, we will need to modify the check here and any object that use this path.

qutip/solver/solver_base.py
```diff
@@ -121,17 +127,21 @@ def _prepare_state(self, state):
             # refer to the ODE tolerance and some integrator do not use it.
             and np.abs(norm - 1) <= settings.core["atol"]
             # Only ket and dm can be normalized
-            and (self.rhs._dims[1] == state._dims or state.shape[1] == 1)
+            and (self.rhs._dims[1] == state._dims or state.shape[1] == 1
+                 or (not self._vectorize_state and state.isoper))
```
Ericgig: Why is this here?

doc/guide/dynamics/dynamics-master.rst
```diff
@@ -302,3 +302,48 @@ calculated with the lines following code
...
+Or when using :class:`.MESolver` directly:
+
+.. code-block:: python
+
+    >>> solver = MESolver(H, c_ops, options={"matrix_form": True})
+    >>> result = solver.run(rho0, tlist, e_ops=[op])
+
+The matrix-form solver supports all the same features as the standard solver,
+including time-dependent Hamiltonians and collapse operators, expectation value
+feedback, and all available ODE integrators.
```
Ericgig:
```suggestion
feedback, and most available ODE integrators.
```
There is one method that require the super operator: "diag".

doc/guide/dynamics/dynamics-options.rst
```diff
@@ -49,3 +49,19 @@ the :func:`.mesolve` and :func:`.mcsolve` function::
...
+``matrix_form`` option that selects between the superoperator-based solver
+and the matrix-form solver:
+
+.. code-block:: python
+
+    >>> # Use the matrix-form Lindblad solver
+    >>> result = mesolve(H, rho0, tlist, c_ops, options={"matrix_form": True})
+
+See :ref:`master-matrix-form` for more details on when to use the matrix-form
+solver.
```
Ericgig: This file does not describe specific solver options...

Ericgig (commented):
Sorry for the delay, I wanted a break after the other review.

The feature itself is great, but this part feels somewhat forced.

qutip/core/cy/lindblad_matrix_form.pyx
amilsted: Ah, this is obsolete. In a previous design, there was some mutation, but not anymore.

qutip/core/cy/lindblad_matrix_form.pyx
amilsted: Again a holdover from a previous approach. Will fix.

qutip/solver/integrator/explicit_rk.pyx
amilsted: I suppose we could make it a QobjEvo. I will give that a try!

qutip/core/cy/lindblad_matrix_form.pyx
amilsted: Sorry, I don't understand what you mean. Here we are accumulating the result into the out_dense array. What side effect are you referring to?

qutip/core/cy/lindblad_matrix_form.pyx
amilsted: Sure, we can put this in a separate function.

qutip/core/data/adjoint.pyx
amilsted: Are you asking why I chose not to implement iadjoint_dense here? We could do `iadjoint_dense(input, output)` that accumulates into output, sure.

qutip/solver/mesolve.py
amilsted: Sorry. This is an AI choice I did not catch!

doc/guide/dynamics/dynamics-options.rst
amilsted: Fair. I will look for a better place for this.

qutip/core/cy/lindblad_matrix_form.pyx
Ericgig: Not take for granted that the `out` is modified in-place, but use the output:
```
            temp_dense = imul_dense(temp_dense, 0)
            temp_dense = c_op.matmul_data(t, rho_dense, temp_dense, 0.5)
            out_dense = c_op.adjoint_rmatmul_data(t, temp_dense, out_dense)
```
Presently, it's hard coded for Dense only, but we may want to use it with other format which have other rules for in-place.

qutip/core/data/adjoint.pyx
Ericgig: You implemented `iconj_dense`, but do a transposition of the same object after using it to end-up with the adjoint. Why split the logic?

Yes, `iadjoint_dense(input, output)` would work. It would remove the Dense internal from the LindbladMatrixForm.

amilsted: I did this, but with `iadjoint_dense(matrix, tmp)` that adds the adjoint of `matrix` to itself, as this seems a little more efficient.

qutip/solver/solver_base.py
amilsted: Now that we only have QobjEvo's, this is gone, but I removed the QobjEvo(rhs) in case rhs is already a QobjEvo, which means we no longer copy that input data. This seemed cleaner. Let me know if it's a problem not to copy here.

qutip/solver/integrator/explicit_rk.pyx
amilsted: I did this and reverted the changes to explicit_rk.

amilsted:
Put some benchmark results here: https://github.com/amilsted/qutip/blob/matmul_bench/BENCHMARKING.md#end-to-end-coupled-transmons-scqubits

qutip/core/data/adjoint.pyx
```diff
@@ -110,6 +111,23 @@ cpdef Dense conj_dense(Dense matrix):
     return out
 
 
+cpdef Dense iadjoint_dense(Dense matrix, Dense tmp):
```
Ericgig: With that name, I would expect matrix to become the adjoint, but not the sum with it's adjoint...

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -10,6 +10,7 @@ from qutip.core.data.mul cimport imul_data
 from qutip.core.data.tidyup import tidyup_csr
 from qutip.core.data.norm import frobenius_data
 from qutip.core.data.ode cimport cy_wrmn_error
+from qutip.core.cy.qobjevo cimport QobjEvo
```
Ericgig: Why?

qutip/solver/mesolve.py
```diff
@@ -231,12 +239,33 @@ def __init__(
...
+
+        if matrix_form:
+            if H.issuper:
+                raise TypeError(
+                    "matrix_form=True cannot be used with superoperator H"
+                )
+            self._vectorize_state = False
+            H = QobjEvo(H)
+            c_ops_qevo = [QobjEvo(c) for c in c_ops]
+            from qutip.core.cy.lindblad_matrix_form import LindbladMatrixForm
```
Ericgig: The import goes at the top with the other.

qutip/solver/mesolve.py
```diff
@@ -231,12 +239,33 @@ def __init__(
...
+            rhs = LindbladMatrixForm(H, c_ops_qevo)
+        else:
+            rhs = H if H.issuper else liouvillian(H)
+            rhs += sum(c_op if c_op.issuper else lindblad_dissipator(c_op)
+                       for c_op in c_ops)
 
         Solver.__init__(self, rhs, options=options)
 
+    def _prepare_state(self, state):
```
Ericgig, line 262: For a pure state, _prepare_state convert the ket to a density matrix. 
Doing this check here before that will cause unwanted errors.

doc/guide/dynamics/dynamics-master.rst
```diff
@@ -302,3 +302,48 @@ calculated with the lines following code
...
+Or when using :class:`.MESolver` directly:
+
+.. code-block:: python
+
+    >>> solver = MESolver(H, c_ops, options={"matrix_form": True})
+    >>> result = solver.run(rho0, tlist, e_ops=[op])
+
+The matrix-form solver supports all the same features as the standard solver,
+including time-dependent Hamiltonians and collapse operators, expectation value
+feedback, and most available ODE integrators.
```
Ericgig, line 349: Maybe add a note that it need hermitian dm as state. While it's the usual case, propagator / correlation use other format.

qutip/tests/solver/test_lindblad_matrix_form.py
```diff
@@ -0,0 +1,134 @@
...
+        c = np.sqrt(0.1) * qutip.destroy(N)
+        c_ops = [c]
+
+        rhs = LindbladMatrixForm(QobjEvo(H), [QobjEvo(c) for c in c_ops])
+
+        H_nh_expected = H - 0.5j * (c.dag() * c)
+        H_nh_actual = rhs.H_nh(0.0)
+
+        diff = (H_nh_actual - H_nh_expected).norm()
+        assert diff < 1e-12
```
Ericgig: Identity is defined for Qobj"
```suggestion
        with qutip.CoreOptions(atol=1e-12):
            assert H_nh_actual == H_nh_expected
```

qutip/tests/solver/test_lindblad_matrix_form.py
```diff
@@ -0,0 +1,134 @@
...
+        # Time-dependent Liouvillian superoperator (built once)
+        L = qutip.liouvillian(QobjEvo(H), [QobjEvo(c) for c in c_ops])
+
+        for t in [0.0, 0.5, 1.0, 2.0]:
+            drho_matrix = rhs_matrix.matmul_data(t, rho_dense)
+
+            drho_super_vec = L(t) * rho_vec
+            drho_super = qutip.vector_to_operator(drho_super_vec)
+
+            diff = np.max(np.abs(drho_matrix.to_array() - drho_super.full()))
```
Ericgig: `np.testing.assert_allclose`?
`assert drho_matrix == drho_super`?

qutip/tests/solver/test_mesolve.py
```diff
@@ -95,47 +102,48 @@ def testME_CteDecay(self, cte_c_ops):
...
+        options = {"progress_bar": None, "matrix_form": matrix_form}
         medata = mesolve(H, psi0, self.tlist, c_op_list, e_ops=[H],
                          args={"kappa": self.kappa}, options=options)
         expt = medata.expect[0]
         actual_answer = 9.0 * np.exp(-2 * self.kappa *
                                      (1.0 - np.exp(-self.tlist)))
         np.testing.assert_allclose(actual_answer, expt, atol=me_error)
 
-    def testME_TDH_TDDecay(self, H, c_ops):
+    def testME_TDH_TDDecay(self, H, c_ops, matrix_form):
```
Ericgig: The tests are already quite long.
I am not sure testing `matrix_form` everywhere is really needed.
Maybe limit it to enough cases to cover all `matrix_form` feature / lines of code, once of twice.

qutip/solver/mesolve.py
amilsted: The isherm check only happens if isoper is True, so only for density-matrix input. If we get a ket here, we just pass it through and it will get turned into a (necessarily hermitian) dm.

qutip/tests/solver/test_mesolve.py
amilsted: I have reduced the tests quite a bit.

qutip/core/data/adjoint.pyx
amilsted: Changed to `iadd_adjoint_dense`

qutip/core/data/adjoint.pyx
```diff
@@ -110,6 +111,23 @@ cpdef Dense conj_dense(Dense matrix):
...
+    cdef size_t nbytes = matrix.shape[0] * matrix.shape[1] * sizeof(double complex)
+    memcpy(tmp.data, matrix.data, nbytes)
+    cdef size_t ptr
+    with nogil:
+        for ptr in range(matrix.shape[0] * matrix.shape[1]):
+            tmp.data[ptr] = _conj(tmp.data[ptr])
+    tmp.fortran = not tmp.fortran
+    iadd_dense(matrix, tmp, 1)
+    tmp.fortran = not tmp.fortran
+    return matrix
```
Ericgig, line 129: Since it's available from python space, we need to check the shape, otherwise it could cause a segfault if tmp is too smaller than matrix.
```suggestion
cpdef Dense iadd_adjoint_dense(Dense matrix, Dense tmp):
    """Add the adjoint to matrix in-place: matrix += matrix.dag()

    Uses tmp as scratch space. Both must be square with the same shape.
    """
    if matrix.shape[0] != tmp.shape[0] or matrix.shape[1] != tmp.shape[1]:
        raise ValueError("tmp and matrix must have the same shape")
    cdef size_t nbytes = matrix.shape[0] * matrix.shape[1] * sizeof(double complex)
    memcpy(tmp.data, matrix.data, nbytes)
    cdef size_t ptr
    with nogil:
        for ptr in range(matrix.shape[0] * matrix.shape[1]):
            tmp.data[ptr] = _conj(tmp.data[ptr])
    tmp.fortran = not matrix.fortran
    matrix = iadd_dense(matrix, tmp, 1)
    return matrix
```

qutip/core/cy/lindblad_matrix_form.pyx
```diff
@@ -0,0 +1,240 @@
...
+        else:
+            out_dense = <Dense>out
+            out_dense = imul_dense(out_dense, 0)
+
+        # Allocate temp buffer on first use
+        if self._temp_buffer is None:
+            self._temp_buffer = dense.zeros(self._buffer_size,
+                                            self._buffer_size,
+                                            fortran=rho_dense.fortran)
+        temp_dense = <Dense>self._temp_buffer
```
Ericgig, line 152:
```suggestion
        temp_dense = <Dense>self._temp_buffer
        temp_dense.fortran = rho_dense.fortran
```

Issue #2808, opened 2026-01-23
Extend matrix-to-scalar functions: add `mean` and `mean_abs` (Cython)


veronikakurth:
### Problem Description

Application of `np.mean` and `np.mean(np.abs)` to instances of `Data` hierarchy is a rather frequent operation in QuTiP (e.g. in solvers). For this, one has to always manually convert a matrix to a representation suitable for numpy operations, which leads to repeatable logic in the code, especially in cases when a method is supposed to work with various data types not sharing the same interface. 

### Proposed Solution

We could implement our own `mean` and `mean_abs` functions and make use of the dispatcher pattern to overload them for different QuTiP data types. 
This could be done in a manner similar to existing matrix-to-scalar functions (`qutip/core/data/norm`) and their dispatching. 

### Additional Context

It is a result of the discussion to this issue: [#2747-1](https://github.com/qutip/qutip/issues/2747#issuecomment-3771561688), [#2747-2](https://github.com/qutip/qutip/issues/2747#issuecomment-3773215671)

### Development plan

Implement operators for 
- [x] Dense (easiest!)
- [x] CSR
- [x] Dia

Particular correctness aspects
- [x] Make sure the implementation is robust to numerical noise for every data type (test on Eric's example with `A - (B/3) - (B/3) - (B/3)`)

Performance
- [ ] ~~Scalability testing: check how fast mean is computed for higher dimensions and tune Cython code if necessary~~

Advanced (optional, to discuss)
- [ ] Should we support axis-wise mean value computation?

veronikakurth:
@Ericgig I would appreciate if you could assign me to work on this issue, unfortunately I cannot do this 🙃

Shrivadhu:
@Ericgig  could you please assign me also to resolve this issue please

veronikakurth:
@Shrivadhu This issue is being worked on already, sorry!

Issue #1349, opened 2020-08-25
Consider removal or rework of `auto_tidyup`


jakelishman:
Currently, many `Qobj` operations are tagged as "tidied up" operations, which means that if the global option `auto_tidyup` is set to `True` (its default), then small values (< 1e-15) are truncated from the output matrices.  This generally ends up with "neater" looking matrices when output; there are far fewer numbers like `1 + 4.516584e-17j` reported, when that's basically just `1`.

There are, however, a couple of problems because of this.

The first is speed; after every (very simple)_ operation, such as `matmul` or `add`, the tidier runs on the matrix.  This typically has `O(nnz)` complexity for sparse matrices, and `O(n * n)` complexity for a dense square matrix with `n` rows.  This adds very significant overhead on to simple operations.

Second is numerical precision.  By dropping small terms, we are effectively turning numerical imprecision into a "feature" as opposed to a bug.  Because QuTiP 5 uses the new data-layer to do most of its operations, rather than constructing intermediary `Qobj` instances, many mathematical operations will not give the same result if the data layer is used directly compared to if `Qobj` is used at every step.

For example, let's say we have three random `Qobj`:
```python
   a, b, c = [qutip.rand_herm(5) for _ in [None]*3]
```
Now, in general,
```python
   a + b + c != qutip.Qobj(a.data + b.data + c.data)
```
because `a + b + c` will invoke the `tidyup` code twice, whereas the right-hand side will not invoke it at all.  This means that despite the fact that `Qobj.__add__` uses the data-layer addition underneath, there will be numerical differences using the QuTiP default settings.

I suspect these sort of problems are behind the recent test instabilities on `dev.major` with regards to tolerances.

----

My vote is for a removal of `auto_tidyup`, or at the very least, changing the default value of the option to `False`.  We still want nice values for display, so perhaps we could consider having an option to produce a tidied up copy when calling `__repr__`, instead?

Interested to hear what others think about this too.

Ericgig:
The way I see it, `tidyup` is to remove 0 elements from the sparse matrix, keeping `nnz` low when making the liouvillian etc.
So it should do nothing on Dense data layer.

If we do 
```
a = qutip.rand_herm(5)
a - a 
```
The matrix is empty, but it has still nnz elements saved without `tidyup`. It need to be periodically cleaned to keep sparse matrix sparse. We could lower the number method calling it, but we should keep it at key points.

jakelishman:
[...]

PR #2817, opened 2026-02-09
Issue 2808 matrix mean

veronikakurth:
**Checklist**

- [x] Please read [Contributing to QuTiP Development](https://qutip.readthedocs.io/en/stable/development/contributing.html)
- [x] Please add tests to cover your changes if applicable.
- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](https://qutip.readthedocs.io/en/stable/development/contributing.html#changelog-generation) for more information).

**Description**
Introduces operator overloading for functions `mean_nonzero` (mean of non-zero values) and `mean_abs_nonzero` (mean of absolute non-zero values) for complex matrices (qutip.data.Dia, qutip.data.CSR, qutip.data.Dense).
Identification of close-to-zero values that are discarded from the computation takes place based on the global `qutip.settings.core['atol']` parameter.

**Related issues or PRs**
Solves #2808


veronikakurth:
Hi Eric! @Ericgig 
I would appreciate if you could have a look at the changes (also, at #2824 - it is based on the branch from this PR, the only unique change is in `steadystate.py`), where the direct method of the steady state solver is refactored to use the mean-abs function). There are a couple of design considerations below that I wanted to share with you and ask for your opinion. 

While working on this issue, one of the main topics that I introduced myself into and which turned out to be not a trivial one was handling _values that are small to zero_. Namely, answering the key question: "How do we distinguish “physically small” values from “numerical noise”?"

Given the discussion [here](https://github.com/qutip/qutip/issues/1349) about the autotidyup, the entire debate seems to have a bit of a philosophical flavor! But philosophy aside, concrete decisions had to be made in the context of this PR. 
**TLDR**: I treat values that are close to global `atol` of QuTiP as being close to zero and the ones that should be discarded from mean calculation (or rather _adjusted mean (abs)_ calculation - here we deal with a very QuTiP-specific definition of mean! Further I will refer to the introduced functions as to _adjusted mean_). This filtering-out functionality is also conceptually different from the audotidyup, it serves a different purpose. But the key question is shared.

Below are the questions I considered & answers + arguments, please feel free to let me know if you find them sound. As well as to point to anything I might have missed when addressing requirements of this implementation.

1. Should _adjusted mean_ ignore all values close to zero?

**Answer**: Yes. **Why**: Since the calculation of the weight in the direct method of the steady state solver requires calculating mean of non zero values of a matrix.

*Note regarding the interface:*
To avoid confusion on the computation result, the introduced functions are dispatched as `mean_nonzero` and `mean_abs_nonzero`. 

2. How to identify values that are close to zero? 

**Criterion**: as "being close" to global absolute tolerance from QuTiP's settings (to `qt.settings.core['atol']`).

- [ ] **Open question**: Do we want to let a user configure the tolerance or should it be always in line with global `atol`? In a prospect of this function being used in other contexts, for example (cannot think of any right now, so it might be a very hypothetical case. And if so, it would be still possible to use a context manager "with qt.CoreOptions(atol=...):" if it has to be tuned anywhere).

3. Should the result of the adjusted mean computation be rounded to zero if it's close to zero? 
**Answer**: No. **Why**: I believe this side effect would be undesirable for the steady state use case, so we just return the result as it is. If it's needed, it can be handled on a higher level. 

**Regarding tests:**
- In addition to a standard test case based on `_GenericOpMixin` class, I added `test_atol_boundary` tests to be able to check how the implementation filters out values close to the absolute-tolerance boundary ("noisy values"). Test input matrices are very simple, though, and maybe a subject to be extended to involve more diverse matrix structures. 
**Rationale**: I believe matrices generated randomly by `tests/core/data/conftest.py` don't cover "numerical noise cases" since they sample from the uniform distribution between 0 and 1 and given the maximum dimension, getting values close to zero is not very likely. And since we don't want to rely on chance when testing for correctness, I think having dedicated tests is a more reliable way to ensure that the _adjusted mean_ implementation works as expected. 

Or maybe I missed something in the QuTiP's test cases that already covers a similar goal?

**Regarding performance:**
I didn't test for it explicitly but I tried to minimize interactions with Python via avoiding calls to NumPy (hence no use of `np.isclose` and replacing it with a custom function). What I couldn't completely avoid is having potential areas in the code where GIL may be used: since I have to access `atol` from a Python's attribute, I had to introduce `with gil` blocks to `nogil`-functions. But since it's not done in a tight loop, it shouldn't be a bottleneck, I believe!

**Regarding the PR:** 
I guess adding a new Cython module to `core/data` sub-package can be considered to be a new feature. Do you think we need examples? It's not really a user facing function but maybe you have suggestions on how to improve dev docs for it. 

Thank you!

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,199 @@
...
+    double abs(double complex z) nogil
+
+cdef inline bint isclose(double complex z, double atol) nogil:
+    return abs(z.real) <= atol and abs(z.imag) <= atol
+
+cdef inline int int_max(int a, int b) noexcept nogil:
+    return a if a > b else b
+
+cdef inline int int_min(int a, int b) noexcept nogil:
+    return a if a < b else b
```
Ericgig: Just use built-in `min` and `max`...

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,199 @@
+# cython: language_level=3
+# cython: boundscheck=False, wraparound=False, initializedcheck=False
+from qutip import settings
+from qutip.core.data cimport base, CSR, Dia, Dense
+
+cdef extern from "<complex>" namespace "std":
+    double abs(double complex z) nogil
+
+cdef inline bint isclose(double complex z, double atol) nogil:
```
Ericgig: `issmall`?
With `isclose`, I would expect 2 values and both absolute and relative tolerance.

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,199 @@
...
+
+# This module is meant to be accessed by dot-access (e.g. mean.mean_csr).
+__all__ = []
+
+cpdef double complex mean_csr(CSR matrix) noexcept nogil:
+    cdef base.idxint nnz = 0
+    cdef double atol
+
+    with gil:
+        atol = settings.core['atol']
```
Ericgig: You could simply not use `nogil` for `mean_csr`.
Since it will only be called from python in steadystate, having the function as nogil will have little impact.

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,199 @@
...
+    
+    if nnz == 0:
+        return 0.0
+    return mean / <double complex>nnz
+
+cpdef double complex mean_dense(Dense matrix) noexcept nogil:
+    cdef double atol
+
+    with gil:
+        atol = settings.core['atol']
```
Ericgig: How about:
```suggestion
cpdef double complex mean_dense(Dense matrix, double atol=-1) noexcept:
    if atol < 0:
        atol = settings.core['atol']
```
Passing tolerance explicitly would be better good for more control.

qutip/tests/core/data/test_mean.py
```diff
@@ -0,0 +1,83 @@
+import pytest
+import numpy as np
+import qutip as qt
```
Ericgig:
```suggestion
import qutip
```
No `qt`.

qutip/tests/core/data/test_mean.py
```diff
@@ -0,0 +1,83 @@
...
+        return matrix[mask].sum() / nnz
+
+    specialisations = [
+        pytest.param(mean_csr, CSR, numbers.Complex),
+        pytest.param(mean_dia, Dia, numbers.Complex),
+        pytest.param(mean_dense, Dense, numbers.Complex),
+    ]
+    @pytest.mark.parametrize("op, dtype, _", specialisations)
+    @pytest.mark.parametrize("scale", [0.999, 1.0, 1.001])
+    def test_atol_boundary(self, op, dtype, _, scale):
```
Ericgig: `_,` still needed?

qutip/core/data/mean.pyx
veronikakurth: This is a good idea, thanks!

qutip/core/data/mean.pyx
veronikakurth: You're right, sorry. I shouldn't have followed `norm.pxd` here since we don't have name collisions with dispatched functions here. Wanted to remove it and forgot!

qutip/core/data/mean.pyx
veronikakurth: Changed to `is_small`, thanks for the catch!

qutip/core/data/mean.pyx
veronikakurth: I see your point, thanks! My argument for leaving `nogil` in signatures, however, would be for the case if this function is ever used for some other computation in the data layer in the future. Then performance might matter.

qutip/core/data/mean.pyx
veronikakurth: Added!

qutip/tests/core/data/test_mean.py
veronikakurth: Unfortunately, it's needed
1) in the first argument to `pytest.mark.parametrize` since I use `specialisations` array, each entry of which will be unpacked into three values;
2) in the signature of `test_atol_boundary` since once it's mentioned in `parametrize`, it must become a part of the test function's signature, too.

To make it look nicer, I thought we actually can give the third unpacked value a name (`return_type`) and make use of it by asserting the function returns the value of a correct type.
Please check it out now. :)

veronikakurth:
@Ericgig I addressed your comments, thanks for the code review!
I hope the CI tests will succeed this time, there was some problem with Coveralls - jobs failed with an HTTP-error (502 Server Error). Might be a network glitch / server overload, I guess the best strategy for now is just to wait a bit.

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,196 @@
...
+cdef inline bint is_small(double complex z, double atol) nogil:
+    return abs(z.real) <= atol and abs(z.imag) <= atol
+
+cdef double complex _mean_generic(
+    double complex* data,
+    size_t start,
+    size_t end,
+    double atol
+) noexcept nogil:
+    cdef base.idxint i, count = 0
```
Ericgig:
```suggestion
    cdef size_t i, count = 0
```
When doing `for i in range(start, end)` if both `start` and `end` are `size_t` (int64), but `i` is int32, this could cause errors.
Also the total number of element can be larger that `base.idxint`'s max. Each dimension maximum is `base.idxint`, so there can be up to `base.idxint`'s max square elements.

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,196 @@
...
+    cdef double complex total = 0
+
+    for i in range(start, end):
+        if not is_small(data[i], atol):
+            total += data[i]
+            count += 1
+    return total / <double complex>count if count > 0 else 0.0
```
Ericgig:
```suggestion
    return total / count if count > 0 else 0.0
```

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,196 @@
...
+cpdef double complex mean_csr(CSR matrix, double atol=-1) noexcept nogil:
+    # Take the global absolute tolerance in case not provided by user
+    if atol < 0:
+        with gil:
+            atol = settings.core['atol']
+
+    cdef base.idxint nnz = 0
+
+
+    nnz = matrix.row_index[matrix.shape[0]]
```
Ericgig:
```suggestion
    cdef base.idxint nnz = 0

    nnz = matrix.row_index[matrix.shape[0]]
```

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,196 @@
...
+    return _mean_generic(matrix.data, 0, nnz, atol)
+
+cpdef double complex mean_dia(Dia matrix, double atol=-1) noexcept nogil:
+    # Take the global absolute tolerance in case not provided by user
+    if atol < 0:
+        with gil:
+            atol = settings.core['atol']
+    cdef int offset, diag, start, end, col = 1
+    cdef double complex cur_el
+    cdef base.idxint nnz = 0
```
Ericgig:
```suggestion
    cdef size_t nnz = 0
```

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,196 @@
...
+        for col in range(start, end):
+            cur_el = matrix.data[diag * matrix.shape[1] + col]
+            if is_small(cur_el, atol):
+                continue
+            mean += cur_el
+            nnz += 1
+    
+    if nnz == 0:
+        return 0.0
+    return mean / <double complex>nnz
```
Ericgig:
```suggestion
    return mean / nnz
```

qutip/core/data/mean.pyx
Ericgig: I would not expect it to be used much if at all outside of steadystate...
For performance, `cdivision=True` would be more impactful.

qutip/core/data/mean.pyx
veronikakurth: I understand the argument with using `size_t` in general (thanks for making the point about it!) but wouldn't be using `base.idxint` for `nnz` be more consistent with respect to how we define this variable in sparse matrix types?

Ericgig: If there are 3 non zero diagonal, nnz can go as high as 3 times the max of `base.idxint`. Overflow in C will not raise errors, but return wrong results.
CSR *should* raise an error when created with more than 2147483647 values. (Not sure it's tested.) But Dense and Dia do not have this limit.

qutip/core/data/mean.pyx
veronikakurth: I see the point, thank you so much! Using `size_t` seems to be indeed a safer option, especially for non-sparse data types.
I was wondering regarding the general design, though: except for giving a possibility to explicitly control if indices are 32- or 64-bit, what are other reasons to introduce `base.idxint` and not use just `size_t` everywhere where we need positive values for indices? I also see that it's being used heavily during memory allocation in matrices' constructions but probably there is even more to that.

qutip/core/data/mean.pyx
veronikakurth: I removed `nogil` from the signatures as you suggested; I guess it was a cargo-cult programming from my side, sorry! The impact is indeed questionable, especially given a potential overhead from `with gil` statement. 

Didn't know about `cdivision` as a compiler directive! I added it. I didn't do benchmarking (probably would be an overkill for this PR), but it sounds conceptually like a very valid thing.

qutip/core/data/mean.pyx
Ericgig: blas, lapack function do not accept size_t most of the time.
They are the mathematic libraries we rely on as much as possible since they are more optimized than anything we can write. 
The best would be to auto-detect the blas support and adapt `base.idxint` to it, but not always easy. Especially if you think we compile on the github runner, but use the users local libraries.

veronikakurth: Great, thank you so much for the explanation! Turned out to be not a trivial question after all despite my doubts. :)

veronikakurth:
@Ericgig I think all the remarks / suggestions are addressed by now, feel free to let me know if there's anything that would prevent the PR from being merged. Thanks a lot for your code review! :pray:

pmenczel:
Thank you for the work @veronikakurth, just two brief comments: (1) I think the `mean.pyx` might be missing an `__all__` declaration (2) The docstrings of `mean_nonzero` and `mean_abs_nonzero` should perhaps explain the `atol` parameter and its default value.

Edit: ah and (3), you probably want to do a `from .mean import *` in the `__init__.py` like all the other modules?
Edit again: huh sorry, please forget about points 1 and 3, I must have ctrl-f'd in the wrong place yesterday

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,194 @@
...
+
+@cdivision(True)
+cpdef double mean_abs_dia(Dia matrix, double atol=-1) noexcept:
+    # Take the global absolute tolerance in case not provided by user
+    if atol < 0:
+        atol = settings.core['atol']
+    cdef int offset, diag, start, end, col = 1
+    cdef double complex cur_el
+    cdef double mean_abs = 0
+    cdef base.idxint nnz = 0
```
Ericgig:
```suggestion
    cdef size_t nnz = 0
```

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,194 @@
...
+
+        for col in range(start, end):
+            cur_el = matrix.data[diag * matrix.shape[1] + col]
+            if is_small(cur_el, atol):
+                continue
+            mean_abs += abs(cur_el)
+            nnz += 1
+    if nnz == 0:
+        return 0.0
+    return mean_abs / <double>nnz
```
Ericgig:
```suggestion
    return mean_abs / nnz
```

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,194 @@
...
+
+cpdef double complex mean_dense(Dense matrix, double atol=-1) noexcept:
+    # Take the global absolute tolerance in case not provided by user
+    if atol < 0:
+        atol = settings.core['atol']
+
+    return _mean_generic(
+        matrix.data,
+        0,
+        matrix.shape[0] * matrix.shape[1],
```
Ericgig: Conversion to int64 must be done before the product.
```suggestion
        <size_t> matrix.shape[0] * <size_t> matrix.shape[1],
```

qutip/core/data/mean.pyx
```diff
@@ -0,0 +1,194 @@
...
+
+cpdef double mean_abs_dense(Dense matrix, double atol=-1) noexcept:
+    # Take the global absolute tolerance in case not provided by user
+    if atol < 0:
+        atol = settings.core['atol']
+
+    return _mean_abs_generic(
+        matrix.data,
+        0,
+        matrix.shape[0] * matrix.shape[1],
```
Ericgig:
```suggestion
        <size_t> matrix.shape[0] * <size_t> matrix.shape[1],
```

qutip/core/data/__init__.py
```diff
@@ -18,6 +18,7 @@
 from .linalg import *
 from .matmul import *
 from .make import *
+from .mean import *
```
Ericgig: `__all__` is empty, (which I believe is the right choice), but nothing will imported with this.
`from . import mean` would be better.

veronikakurth: Updated!

veronikakurth:
@Ericgig @pmenczel Addressed the next round of comments, thanks!

PR #2821, opened 2026-02-17
Fix entrypoints iteration bug in about() and improve optional plugin handling

vishwas-droid:
This PR fixes a NameError in qutip.about() caused by iterating over
`entry_points` instead of `entrypoints`.

Additionally, plugin loading failures (e.g. missing optional dependencies
like jaxlib for qutip_jax) are handled gracefully with a warning instead
of causing a crash.

This improves robustness of the plugin discovery mechanism and ensures
optional QuTiP family packages do not break core functionality.

### DIFF
--- a/qutip/about.py
+++ b/qutip/about.py
@@ -11,6 +11,8 @@ import numpy
 import scipy
 import inspect
 import qutip
+import warnings
+
 from qutip.settings import _blas_info, settings
 
 
@@ -74,7 +76,7 @@ def about():
 
     if not entrypoints:
         print("No QuTiP family packages installed.")
-    import warnings
+
 
     for ep in entrypoints:
         try:
--- a/qutip/about.py
+++ b/qutip/about.py
@@ -74,9 +74,18 @@ def about():
 
     if not entrypoints:
         print("No QuTiP family packages installed.")
+    import warnings
 
     for ep in entrypoints:
-        family_mod = ep.load()
+        try:
+            family_mod = ep.load()
+        except Exception as exc:
+            warnings.warn(
+                f"Skipping QuTiP family package '{ep.name}' "
+                f"due to import error: {exc}"
+            )
+            continue
+
         try:
             pkg, version = family_mod.version()
         except Exception as exc:


veronikakurth:
@vishwas-droid Thanks for your contribution! But can you please provide steps to reproduce the issue you are trying to solve within this PR?

vishwas-droid:
@veronikakurth Thanks for pointing this out 

To reproduce the issue:

-Install QuTiP without the optional plugins.
-Run qutip.about() in a Python session.
-If any optional plugin is missing or not properly installed, the entry point iteration can raise an error and interrupt the process.

The PR makes the plugin discovery more defensive so that missing optional packages don’t affect core functionality.

Happy to add a small repro script or test if that would help!

pmenczel:
Hi @vishwas-droid

> This PR fixes a NameError in qutip.about() caused by iterating over entry_points instead of entrypoints.

I am a bit confused, I cannot find any change like this in this PR.

Adding a `try`/`except` guard for the `ep.load()` might be a good idea, but is there a reason to not just use the same exception handling that already exists in the line below?

veronikakurth:
> I am a bit confused, I cannot find any change like this in this PR.
I also believe the problem statement is wrong. 

> Adding a try/except guard for the ep.load() might be a good idea, but is there a reason to not just use the same exception handling that already exists in the line below?

I would agree to the proposed changes for handling loading of modules. Having a closer look at it, I think it also makes sense to have two `try/except` clauses since they correspond to two different exception handling cases:

1. Handling a loading issue 
2. Handling absence of `version` function in the module

One could pack it all into one `try-except` clause but having two makes our exception handling a bit more granular.

However, I have some changes to the code to propose, they follow next.

qutip/about.py
```diff
@@ -75,8 +77,17 @@ def about():
...
 
+
     for ep in entrypoints:
-        family_mod = ep.load()
+        try:
+            family_mod = ep.load()
+        except Exception as exc:
+            warnings.warn(
+                f"Skipping QuTiP family package '{ep.name}' "
+                f"due to import error: {exc}"
```
veronikakurth, line 87: Import error is not the only type of error that may emerge.
Possible exceptions within `importlib.metadata.EntryPoint.load` are:
1. ImportError                                                                                                                                                                              
2. AttributeError                                                                                                                                                                             
3. ModuleNotFoundError                                                                                                                                                                             
4. Any exception that occurs during module import/execution

Hence, I'd suggest to not be specific about type of error in the hard-coded text of the warning and make it more general instead of "import error".

veronikakurth (changes requested):

veronikakurth:
@vishwas-droid there has been no activity in this PR since the code review; do you plan to come back to it? Otherwise, I'd suggest we close it.

vishwas-droid:
Apologies for delay iil work on it tonight

veronikakurth:
No activity.

PR #2825, opened 2026-02-25
Use frame_width for wireframe linewidth in Bloch sphere

### LINKED ISSUE #2396: Setting the linewidth of the Bloch sphere wireframe
### Problem Description

The `frame_width` attribute of the `Bloch` class is used for drawing the equatorial plane, but not for drawing the wireframe (see the `plot_back` method of the `Bloch` class). This is not what I was expecting from reading the documentation.

### Proposed Solution

A better behaviour would probably be either tu use this attribute for both, or to have two separate attributes.

### Alternate Solutions

_No response_

### Additional Context

_No response_

Chinmay-Tangal:
**Description**

The `frame_width` attribute of the `Bloch` class was used for the equator 
lines but not for the wireframe in `plot_back` and `plot_front`. This made 
the behavior inconsistent — users expected `frame_width` to control all 
frame lines, but the wireframe was unaffected.

### Changes

- Added `linewidth=self.frame_width` to `plot_wireframe` call in `plot_back()`
- Added `linewidth=self.frame_width` to `plot_wireframe` call in `plot_front()`

**Related issues or PRs**

Fixes #2396

### DIFF
--- a/doc/changes/2825.bugfix
+++ b/doc/changes/2825.bugfix
@@ -1,0 +1,1 @@
+The wireframe in ``plot_back`` and ``plot_front`` now respects the ``frame_width`` attribute, making it consistent with the equator lines.


Ericgig:
Plotting issue are up to taste, but I think having the secondary lines of the frame to be smaller than the main ones is the better look. And seeing no other maintainer picked this for a month, I think most agree.

Still thank you trying to close the issue.

PR #2826, opened 2026-02-28
fix: metrics: Added spectral decomposition to hermitian matrices

### LINKED ISSUE #2822: `LinAlgWarning: Matrix is singular` when calculating `fidelity` of pure state density matrices
### Bug Description

When calculating `qutip.fidelity(A, B)` where both A and B are density matrices (type='oper') and at least one of them represents a pure state (or has zero eigenvalues), it triggers a `scipy.linalg.LinAlgWarning: Matrix is singular` warning.

### Code to Reproduce the Bug

```shell

```

### Code Output

```shell

```

### Expected Behaviour

The `qutip.fidelity(A, B)` function should compute and return the correct fidelity value (e.g., 1.0 for identical pure states) silently, without raising any `scipy.linalg.LinAlgWarning`. Because pure state density matrices are perfectly valid, positive semi-definite (PSD) physical objects, their square root is mathematically well-defined.

### Your Environment

```shell
QuTiP Version:      5.2.3
Numpy Version:      2.4.2
Scipy Version:      1.16.3
Cython Version:     None
Matplotlib Version: 3.10.8
Python Version:     3.11.13
Number of CPUs:     20
INTEL MKL Ext:      None
```

### Additional Context

_No response_

mudit06mah:
**Description**
Did minor tweaks in the `fidelity` function to avoid warning: `LinAlgWarning: Matrix is singular` for pure states

**Changes**
- Changed `ket2dm(A)` to `A.proj()` for pure states (as sqrt(P) = P for projectors)
- Added spectral decomposition for hermitians in `exmp.py`  

**Related issues or PRs**
Fixes #2822

### DIFF
--- a/qutip/core/data/expm.py
+++ b/qutip/core/data/expm.py
@@ -157,10 +157,7 @@ def sqrtm_dense(matrix, /, isherm=None) -> Dense:
 
 
 sqrtm = _Dispatcher(
-    _inspect.Signature([
-        _inspect.Parameter('matrix', _inspect.Parameter.POSITIONAL_ONLY),
-        _inspect.Parameter('isherm', _inspect.Parameter.KEYWORD_ONLY, default=None),
-    ]),
+    sqrtm_dense,
     name='sqrtm',
     module=__name__,
     inputs=('matrix',),


pmenczel:
Hi @mudit06mah and thanks for your contribution! Could you please remove the `temp.py` file from the PR?

I just had a few comments / questions:
* Regarding "Changed `ket2dm(A)` to `A.proj()` for pure states": it is okay, but `ket2dm` and `proj` are exactly the same (`ket2dm` calls `proj`).
* You are changing the code in the fidelity function to avoid calling `sqrtm`, where the warning occurs. However, this warning will still be there if a user calls `sqrtm` directly for some reason. Perhaps it would make more sense to eliminate the root problem and modify `sqrtm` directly?
* I worry about performance, computing eigenvalues and eigenvectors might be rather slow. Have you checked this?

mudit06mah:
@pmenczel Thank You for your review! My apologies for accidentally pushing the `temp.py` file.

Concerning your points:

- You're correct, both `ket2dm()` and `proj()` are identical, I'm still sticking to `proj()` since it is more direct.
- I have now shifted my solution from `fidelity()` directly to `Qobj.sqrtm()` 
- Both `scipy.linalg.sqrtm` and reconstructing the square matrix using the eigenvalues and eigenvectors take O(N^3) time complexity while the latter is a better approach for hermitian density matrices.

qutip/core/qobj.py
```diff
@@ -992,6 +992,22 @@ def sqrtm(
         """
         if self._dims[0] != self._dims[1]:
             raise TypeError('sqrt only valid on square matrices')
+        
+        is_herm = self._isherm if self._isherm is not None else False
+
+        if is_herm:
+            evals, ekets = self.eigenstates(sparse=sparse, tol=tol, maxiter=maxiter)
+            # Clip tiny negative values to 0 to handle numerical noise
+            sqrt_evals = np.sqrt(np.maximum(evals, 0))
```
Ericgig: Not all hermitian Qobj are density matrix.
Negative eigenvalues are not always numerical error.

qutip/core/qobj.py
```diff
@@ -992,6 +992,22 @@ def sqrtm(
         """
         if self._dims[0] != self._dims[1]:
             raise TypeError('sqrt only valid on square matrices')
+        
+        is_herm = self._isherm if self._isherm is not None else False
+
+        if is_herm:
+            evals, ekets = self.eigenstates(sparse=sparse, tol=tol, maxiter=maxiter)
```
Ericgig: You should look into the `output_type` argument of `eigenstates` or into using `_data.eigs` directly.

qutip/core/qobj.py
```diff
@@ -992,6 +992,22 @@ def sqrtm(
...
+            # Clip tiny negative values to 0 to handle numerical noise
+            sqrt_evals = np.sqrt(np.maximum(evals, 0))
+            # Reconstruct the square root matrix.
+            data = _data.zeros(self.data.shape[0], self.data.shape[1])
+            for i in range(len(evals)):
+                if sqrt_evals[i] > 0:
+                    term = ekets[i] * ekets[i].dag()
+                    data = _data.add(data, _data.mul(term.data, sqrt_evals[i]))
```
Ericgig: Look at the `_data.add` documentation.

mudit06mah:
@Ericgig Regarding your points:

- I agree that `Qobj.sqrtm()` should essentially be a wrapper on top of `_data.sqrtm()`, the backend call for `_data.sqrtm()` has not been removed for all cases but only for singular hermitian matrices
- My earlier approach was indeed slower in comparison, hence I have removed the python-level loop I had earlier for reconstruction and instead used `_data.matmul()`. #https://rubenvannieuwpoort.nl/posts/the-spectral-theorem-for-hermitian-matrices.
- I now cast eigenvalues to complex numbers, hence computation of `sigmaz().sqrtm()` can now work correctly.

pmenczel:
I did some rudimentary performance testing. For larger systems, this seems to speed up `sqrtm` of Hermitian operators, but for small systems, it's a bit slower.

image
image

mudit06mah:
@pmenczel @Ericgig I have added changes to `expm.py` instead of `Qobj.py`, can you PTAL again? 👾

qutip/core/data/expm.py
```diff
@@ -140,6 +145,16 @@ def logm_dense(matrix: Dense) -> Dense:
 def sqrtm_dense(matrix) -> Dense:
     if matrix.shape[0] != matrix.shape[1]:
         raise ValueError("can only compute logarithm square matrix")
+
+    # Spectral Decomposition for Hermitian matrices to avoid SciPy's singularity warning
+    if isherm_dense(matrix):
+        evals, evecs = eigs_dense(matrix, isherm=True)
+        sqrt_lambda = np.sqrt(evals.astype(complex))
+        diag_csr = diag(sqrt_lambda, 0, matrix.shape)
+        diag_dense = to(Dense, diag_csr)
```
Ericgig:
```suggestion
        diag_dense = diag(sqrt_lambda, 0, matrix.shape, dtype=Dense)
```
or
```suggestion
        diag_dense = dense.diags(sqrt_lambda, 0, matrix.shape)
```
Every function with multiple dispatch on the output has a dtype kwarg added. It's not documented...

qutip/core/data/expm.py
```diff
@@ -140,6 +145,16 @@ def logm_dense(matrix: Dense) -> Dense:
 def sqrtm_dense(matrix) -> Dense:
```
Ericgig:
```suggestion
def sqrtm_dense(matrix, isherm=None) -> Dense:
```

qutip/core/data/expm.py
```diff
@@ -140,6 +145,16 @@ def logm_dense(matrix: Dense) -> Dense:
...
     if matrix.shape[0] != matrix.shape[1]:
         raise ValueError("can only compute logarithm square matrix")
+
+    # Spectral Decomposition for Hermitian matrices to avoid SciPy's singularity warning
+    if isherm_dense(matrix):
```
Ericgig:
```suggestion
    if (isherm if isherm is not None else isherm_dense(matrix)):
```

qutip/core/data/expm.py
Ericgig, line 161: Let's just have the Dispatcher extract the signature for us.
This should be enough to add the `isherm` argument:
```suggestion
sqrtm = _Dispatcher(
sqrtm_dense,
name='sqrtm',
```

mudit06mah:
@Ericgig I have made the changes you suggested, can you please review this again?

Should I also open an issue to add documentation for the `dtype` keyword?

mudit06mah:
> I hoped for a little more polish, but it's in an acceptable state.

Hi @Ericgig! Could you be more specific as to where I can improve in my code/implementation, would be great for future references!
I am willing to make further changes to your satisfaction :)

Ericgig:
- The `to` is still imported, but no longer useful.
- In `Qobj.sqrtm`, it could pass the `isherm` flag to `_data.sqrtm`.
- [`matmul_dag_dense`](https://github.com/qutip/qutip/blob/927aff166bd19112f58fa9869b536ac86d107cbd/qutip/core/data/matmul.pyx#L1097) could be used instead of `matmul_dense(v_mid, adjoint_dense(evecs))`. Internally `A @ B.dag` is quite common, so we have an specialized function for it.

mudit06mah:
@Ericgig Sorry for the overlook on the imports and `Qobj.sqrtm` function calls, I have made all the changes you have suggested.

Ericgig (approved):
Thank you!

qutip/core/qobj.py
```diff
@@ -993,6 +993,7 @@ def sqrtm(
         if self._dims[0] != self._dims[1]:
             raise TypeError('sqrt only valid on square matrices')
         return Qobj(_data.sqrtm(self._data),
+                    isherm=self._isherm,
```
pmenczel: This is probably a typo, `isherm` should be passed to `_data.sqrtm` and not to the `Qobj` constructor. Note that the square root of a Hermitian operator might not be Hermitian itself, so passing `isherm=self._isherm` to the new Qobj is definitely incorrect.

pmenczel:
Additionally: In PR #2711 we made a change to our tests to ignore that `LinAlgWarning`. If you feel up to it, it would be great if you could try to see if that change can be reverted now without breaking the tests.

qutip/core/qobj.py
mudit06mah: Wow, Thank you for catching this.

mudit06mah:
@pmenczel I have now removed the warning supression for `LinAlgWarning` in workflows! can you PTAL again?

mudit06mah:
@pmenczel I have added `isherm` to the dispatcher so the earlier workflow error should be resolved now! can you PTAL again?

pmenczel:
Thank you for fixing the yaml syntax. The failing qutip-jax test run was not due to the dispatcher thing, but due to the [qutip-jax implementation of `sqrtm`](https://github.com/qutip/qutip-jax/blob/177aa4c5401e968f1e7589fd9f929b4e4197719b/src/qutip_jax/unary.py#L117) missing the `isherm` parameter. (That would have to be fixed in a separate PR in the qutip-jax repo.) Since @Ericgig asked earlier to just have the Dispatcher extract the signature for us, perhaps you could undo the change in 2f1cc81?

mudit06mah:
> The failing qutip-jax test run was not due to the dispatcher thing, but due to the [qutip-jax implementation of `sqrtm`](https://github.com/qutip/qutip-jax/blob/177aa4c5401e968f1e7589fd9f929b4e4197719b/src/qutip_jax/unary.py#L117) missing the `isherm` parameter. (That would have to be fixed in a separate PR in the qutip-jax repo.)

Understood,
I actually have the qutip-jax repo set up as well, should I open a PR with the changes there?

mudit06mah:
> (That would have to be fixed in a separate PR in the qutip-jax repo.)

@pmenczel I have created a PR in qutip-jax for the same: https://github.com/qutip/qutip-jax/pull/99
PTAL!

PR #2829, opened 2026-03-05
Qobj: Deprecate Qobj @ numpy.array

### LINKED ISSUE #2547: Qobj multiplication overloading yields inconsistent results when combined with numpy objects
### Bug Description

Realized that I was getting inconsistent results in my Hamiltonian simulation code, and was testing using basic matrix operations. Some pieces that qutip was not doing easily, I used numpy objects and operations to complete. 

Qutip documentation suggests using * operator for matrix/vector multiplication https://qutip.org/docs/4.7/guide/guide-states.html -- but this will yield undesired results when the types of all objects are not rigorously checked. 

### Code to Reproduce the Bug

```shell
import qutip
import numpy as np

A = qutip.Qobj([[0, 1], [1, 0]])
B = qutip.Qobj([[1, 2], [3, 4]])

# qutip objects
print(A * B) # gives expected results
print(A @ B) # gives expected results

C = np.array([[0, 1], [1, 0]])
D = np.array([[1, 2], [3, 4]])

# numpy objects
print(C * D) # gives wrong results
print(C.dot(D)) # gives expected results
print(C @ D) # gives expecting results
```

### Code Output

```shell
Quantum object: dims=[[2], [2]], shape=(2, 2), type='oper', dtype=Dense, isherm=False
Qobj data =
[[3. 4.]
 [1. 2.]]
Quantum object: dims=[[2], [2]], shape=(2, 2), type='oper', dtype=Dense, isherm=False
Qobj data =
[[3. 4.]
 [1. 2.]]
[[0 2]
 [3 0]]
[[3 4]
 [1 2]]
[[3 4]
 [1 2]]
```

### Expected Behaviour

Would prefer that qutip A*B yields the same results as numpy, and A @ B becomes the default matrix operator, or another function is suggested. Consistency with numpy would be better.

### Your Environment

```shell
Current admin team: Alexander Pitchford, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, Boxi Li, Jake Lishman, Simon Cross, Asier Galicia, Paul Menczel, and Patrick Hopf.

QuTiP Version:      5.0.4
Numpy Version:      2.0.1
Scipy Version:      1.14.1
Cython Version:     None
Matplotlib Version: 3.9.2
Python Version:     3.12.6
Number of CPUs:     12
BLAS Info:          Generic
INTEL MKL Ext:      False
Installation path:  /Users/mkalp/.conda-envs/qutip-env/lib/python3.12/site-packages/qutip
```

### Additional Context

_No response_

mudit06mah:
**Description**
This PR deprecates `Qobj @ numpy.array` operation.

**Related issues or PRs**
fixes #2547 

**Steps to test**
Code:
```python
import qutip
import numpy as np

A = qutip.Qobj([[0, 1], [1, 0]])
C = np.array([[0, 1], [1, 0]])

print(A @ C)
```

Output:
```shell
/home/mochi/Projects/qutip/temp.py:7: DeprecationWarning: Support for Qobj @ numpy.array has been deprecated and will be removed in qutip 5.4. Please use Qobj(A) @ B instead.
  print(A @ C)
Qobj data =
[[1. 0.]
 [0. 1.]]
```

### DIFF
--- a/doc/changes/2829.bugfix
+++ b/doc/changes/2829.bugfix
@@ -1,0 +1,1 @@
+Qobj: Updated __matmul__ to throw deprecation warning for Qobj @ numpy.array operation
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -455,6 +455,13 @@ class Qobj:
 
     def __matmul__(self, other: Qobj) -> Qobj:
         if not isinstance(other, Qobj):
+            warnings.warn(
+                "Support for Qobj @ numpy.array has been deprecated "
+                "and will be removed in qutip 5.4. "
+                "Please use Qobj(A) @ B instead.",
+                DeprecationWarning,
+                stacklevel=2
+            )
             try:
                 other = Qobj(other)
             except TypeError:
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -459,8 +459,7 @@ class Qobj:
                 "Support for Qobj @ numpy.array has been deprecated "
                 "and will be removed in qutip 5.5 or later. "
                 "Please use Qobj(A) @ B instead.",
-                DeprecationWarning,
-                stacklevel=2
+                FutureWarning
             )
             try:
                 other = Qobj(other)
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -457,7 +457,7 @@ class Qobj:
         if isinstance(other, np.ndarray):
             warnings.warn(
                 "Support for Qobj @ numpy.array has been deprecated "
-                "and will be removed in qutip 5.4. "
+                "and will be removed in qutip 5.5 or later. "
                 "Please use Qobj(A) @ B instead.",
                 DeprecationWarning,
                 stacklevel=2
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -454,7 +454,7 @@ class Qobj:
         return self.__mul__(other)
 
     def __matmul__(self, other: Qobj) -> Qobj:
-        if not isinstance(other, Qobj):
+        if isinstance(other, np.ndarray):
             warnings.warn(
                 "Support for Qobj @ numpy.array has been deprecated "
                 "and will be removed in qutip 5.4. "
@@ -464,8 +464,12 @@ class Qobj:
             )
             try:
                 other = Qobj(other)
-            except TypeError:
+            except Exception:
                 return NotImplemented
+
+        if not isinstance(other, Qobj):
+            return NotImplemented
+
         new_dims = self._dims @ other._dims
         if new_dims.type == 'scalar':
             return _data.inner(self._data, other._data)
--- a/qutip/tests/core/test_qobjevo.py
+++ b/qutip/tests/core/test_qobjevo.py
@@ -407,17 +407,18 @@ def test_mul_vec(all_qevo):
 def test_matmul(all_qevo):
     "QobjEvo matmul oper"
     mat = np.random.rand(N, N) + 1 + 1j * np.random.rand(N, N)
+    matQobj = Qobj(mat)
     matDense = Qobj(mat).to(_data.Dense)
     matF = Qobj(np.asfortranarray(mat)).to(_data.Dense)
     matCSR = Qobj(mat).to(_data.CSR)
     op = all_qevo
     for t in TESTTIMES:
         Qo1 = op(t)
-        assert_allclose((Qo1 @ mat).full(),
+        assert_allclose((Qo1 @ matQobj).full(),
                         op.matmul(t, matF).full(), atol=1e-14)
-        assert_allclose((Qo1 @ mat).full(),
+        assert_allclose((Qo1 @ matQobj).full(),
                         op.matmul(t, matDense).full(), atol=1e-14)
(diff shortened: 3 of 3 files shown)


qutip/core/qobj.py
```diff
@@ -477,6 +477,15 @@ def __matmul__(self, other: Qobj) -> Qobj:
             copy=False
         )
 
+    def __rmatmul__(self, other) -> Qobj:
+        if not isinstance(other, Qobj):
```
pmenczel: This condition should always be true: if `other` were a `Qobj`, then it would have tried `other.__matmul__(self)` first. I would suggest to just remove the condition.

pmenczel (commented):
Thank you for the contribution. Other than the comment below, this seems like a good addition. (@Ericgig is there any reason not to add this / why it hasn't been added yet?)

Note also, the changelog file should not have the ".rst" ending; that's why the documentation build is failing.

Ericgig:
Do we want to encourage / officially support operations between `Qobj` and `ndarray`?

If we do, what about `Qobj + ndarray`, `Qobj - ndarray`, `Qobj & ndarray`, etc.
Also tests should be added for all new officially supported operations...

If not, should we remove it from `__matmul__`?

pmenczel:
> Do we want to encourage / officially support operations between `Qobj` and `ndarray`?

It's an a bit difficult question. In principle, I would be for removing it altogether, but I don't know if that could be a breaking change for users. However, having it only for the one order of operands is definitely weird. @Ericgig what do you think is best?

(And sorry @mudit06mah that #2547 was maybe somewhat misleading for an open issue.)

mudit06mah:
> (And sorry @mudit06mah that #2547 was maybe somewhat misleading for an open issue.)

No worries, Since there is still need for changes (whether to add support or remove it completely) so I don't think it was invalid to keep it open.
Once a decision has been made I'll make changes accordingly 👾

mudit06mah:
@Ericgig @pmenczel Any Updates on this?

Ericgig:
Personally I am for not supporting any operations with numpy array with the present Qobj. 

But for this kind of design decision, it will be better to discuss with the other qutip developers, we will meet in a few weeks.

doc/changes/2829.bugfix
```diff
@@ -0,0 +1 @@
+Added a __rmatmul__() function to Qobj class to successfully handle numpy.ndarray @ Qobj operations
```
veronikakurth: Would it be considered a bug fix or a feature, though?

veronikakurth:
> Personally I am for not supporting any operations with numpy array with the present Qobj.
> 
> But for this kind of design decision, it will be better to discuss with the other qutip developers, we will meet in a few weeks.

I added a new label (`core-design-review`) to this PR and a related issue for us to track such topics that require a core maintainers discussion before the actual work starts or changes get merged. So that we don't lose them out of sight.

hodgestar:
@mudit06mah: Thank you for raising this issue. As mentioned in #2547 we decided to deprecate support for numpy arrays in `@` and ask users to do `Qobj(A) @ B` if `A` is a numpy array.

You're welcome to make a PR for the deprecation if you like, otherwise one of the admin team will get to it.

mudit06mah:
> You're welcome to make a PR for the deprecation if you like, otherwise one of the admin team will get to it.

Sure, I'd change this PR only to deprecate it instead :)

mudit06mah:
@hodgestar @Ericgig @pmenczel @veronikakurth I have changed this PR to deprecate the operation instead :D PTAL

qutip/core/qobj.py
```diff
@@ -456,11 +454,22 @@ def __rmul__(self, other: complex) -> Qobj:
         return self.__mul__(other)
 
     def __matmul__(self, other: Qobj) -> Qobj:
-        if not isinstance(other, Qobj):
+        if isinstance(other, np.ndarray):
+            warnings.warn(
+                "Support for Qobj @ numpy.array has been deprecated "
+                "and will be removed in qutip 5.4. "
```
veronikakurth: It might be a bit of a short timeline for deprecation; @Ericgig would you say it would be better to communicate that the support will be removed in "qutip 5.5 or later"?

mudit06mah: I have changed the warning message, Could you please review it again?

qutip/core/qobj.py
```diff
@@ -456,11 +454,22 @@ def __rmul__(self, other: complex) -> Qobj:
         return self.__mul__(other)
 
     def __matmul__(self, other: Qobj) -> Qobj:
-        if not isinstance(other, Qobj):
+        if isinstance(other, np.ndarray):
+            warnings.warn(
+                "Support for Qobj @ numpy.array has been deprecated "
+                "and will be removed in qutip 5.5 or later. "
+                "Please use Qobj(A) @ B instead.",
+                DeprecationWarning,
```
Ericgig: It's better to use `FutureWarning`, `DeprecationWarning` are hidden per default so the average user will miss them.

mudit06mah: I see, I changed the stack level to 2 to make the `DeprecationWarning` visible earlier. I have changed it to `FutureWarning` now.

Ericgig (approved):

pmenczel (approved):

Issue #2384, opened 2024-04-04
Support for direct products


pmenczel:
### Problem Description

Currently, there is no great support for working with direct product spaces in qutip. For that reason, both the HEOM solver and qutip-qoc contain code manipulating data layer matrices directly.

Most commonly, direct product spaces arise because of coupled differential equations. For example, we might have two states $|\psi\rangle$ and $|\psi'\rangle$ satisfying

$$ \partial_t \begin{pmatrix} |\psi\rangle \\\\ |\psi'\rangle \end{pmatrix} = -i \begin{pmatrix} H & 0 \\\\ H' & H \end{pmatrix} \begin{pmatrix} |\psi\rangle \\\\ |\psi'\rangle \end{pmatrix} $$

then the object $( |\psi\rangle, |\psi'\rangle )$ is a direct product.

Some applications:
* HEOM is a coupled differential equation for the state $\rho$ of an open system, and a number of auxiliary density operators $\rho_{\mathbf n}$.
* The goat algorithm in qutip-qoc works by integrating coupled Schroedinger equations for two ore more unitary operators.
* In full counting statistics, we can find the first n derivatives of a moment-generating function by solving coupled Lindblad equations for n+1 operators.
* A direct product state could potentially be used instead of a tensor product state in cases where we know that two systems are only coupled classically and there will never be any coherences between them.

### Proposed Solution

I think that we would need the following:
* A way to construct objects such as $( |\psi\rangle, |\psi'\rangle )$ and the matrix in the equation above from their constituents.
* A way to describe the dimensions of these objects.
* A way to extract the constituents from these objects at the end of the calculation.
* The differential equations mentioned in the applications above should be solvable with `sesolve` and/or `mesolve`.
* Note that in the first three applications, we actually have direct products of operator spaces (which might make things complicated?)

### Alternate Solutions

If all the involved Hilbert spaces have the same dimension, then $\mathcal H \times \cdots \times \mathcal H \simeq \mathbb C^n \otimes \mathcal H$. In other words, a "big matrix" can be obtained by tensoring a "small matrix" with another appropriate matrix, for example

$$ \begin{pmatrix} H & 0 \\\\ H' & H \end{pmatrix} = \begin{pmatrix} 1 & 0 \\\\ 0 & 1 \end{pmatrix} \otimes H + \begin{pmatrix} 0 & 0 \\\\ 1 & 0 \end{pmatrix} \otimes H' . $$

[...]

PR #2833, opened 2026-03-06
Direct sums #2

pmenczel:
Main part of my direct sums project. This was originally part of #2785, but I split it up to make it easier to review. After this PR, I am planning two more: one to update the HEOM solver and one to update the rest of the repo to accept sum qobjs and sum spaces where it makes sense.

This adds a new module `qutip.core.direct_sum` with 4 functions: `direct_sum`, `direct_sum_sparse`, `direct_component` and `set_direct_component`. The main function `direct_sum` makes block matrices or block vectors, like so:
```python
In [1]: import qutip as qt

In [2]: op = qt.direct_sum([
   ...:     [qt.sigmax(),          None],
   ...:     [qt.basis(2, 1).dag(), 42  ],
   ...: ])

In [3]: op
Out[3]: 
Quantum object: dims=[([2], [1]), ([2], [1])], shape=(3, 3), type='oper', dtype=CSR, isherm=False
Qobj data =
[[ 0.  1.  0.]
 [ 1.  0.  0.]
 [ 0.  1. 42.]]
```
The blocks can be extracted again using `direct_component`:
```python
In [4]: qt.direct_component(op, 1, 0)
Out[4]: 
Quantum object: dims=[[1], [2]], shape=(1, 2), type='bra', dtype=CSR
Qobj data =
[[0. 1.]]
```
Similarly, `set_direct_component` can be used to overwrite a block. `direct_sum_sparse` is an alternative for large, sparse block matrices (e.g., HEOM). Also `QobjEvo` works everywhere.

### DIFF
--- a/qutip/tests/core/test_direct_sum.py
+++ b/qutip/tests/core/test_direct_sum.py
@@ -19,6 +19,39 @@ def _ramp(t, args):
     return t
 
 
+def _check_meta(result, result_type, result_dims, dtype):
+    assert isinstance(result, (Qobj, QobjEvo))
+    assert result.type == result_type
+    assert result._dims == result_dims
+    assert result.shape == result_dims.shape
+    if result_type in ["operator-ket", "operator-bra", "super"]:
+        assert result.superrep == "super"
+    if isinstance(result, Qobj):
+        assert result.dtype.__name__ == dtype
+    else:
+        assert result(0).dtype.__name__ == dtype
+
+
+def _assert_equal(qobj1, qobj2):
+    if isinstance(qobj1, QobjEvo) and isinstance(qobj2, QobjEvo):
+        # == for QobjEvo tests identity, not equality
+        # this should be enough to check equality
+        assert qobj1(0) == qobj2(0)
+        assert qobj1(1) == qobj2(1)
+    elif isinstance(qobj2, Number):
+        # if we put in a plain scalar, we will get out a Qobj
+        assert qobj1 == Qobj(qobj2)
+    elif qobj2 is None:
+        assert qobj1 == Qobj(0) or qobj1 == qzero_like(qobj1)
+    elif qobj1.type == "operator-ket":
+        # accept getting out vectorized operators
+        assert qobj1 == qobj2 or vector_to_operator(qobj1) == qobj2
+    elif qobj1.type == "operator-bra":
+        assert qobj1 == qobj2 or vector_to_operator(qobj1.dag()) == qobj2
+    else:
+        assert qobj1 == qobj2
+
+
 @pytest.mark.parametrize(
     ["arguments", "result_type", "result_dims"], [
         pytest.param([basis(2, 0)],
@@ -53,18 +86,7 @@ def _ramp(t, args):
 @pytest.mark.parametrize("dtype", ["CSR", "Dense"])
 def test_linear(arguments, result_type, result_dims, dtype):
     result = direct_sum(arguments, dtype=dtype)
-
-    assert isinstance(result, (Qobj, QobjEvo))
-    assert result.type == result_type
-    assert result._dims == result_dims
-    assert result.shape == result_dims.shape
-    if result_type in ["operator-ket", "operator-bra"]:
-        assert result.superrep == "super"
-    if dtype is not None:
-        if isinstance(result, Qobj):
-            assert result.dtype.__name__ == dtype
-        else:
-            assert result(0).dtype.__name__ == dtype
+    _check_meta(result, result_type, result_dims, dtype)
 
     for i in range(len(arguments)):
         cmp = direct_component(result, i)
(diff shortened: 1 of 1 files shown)


qutip/core/direct_sum.py
```diff
@@ -0,0 +1,676 @@
...
+from typing import Any, Union, overload
+
+import numpy as np
+
+
+__all__ = ['direct_sum', 'direct_sum_sparse',
+           'direct_component', 'set_direct_component']
+
+
+QobjLike = Union[Number, Qobj, QobjEvo]
```
Ericgig: To me `...Like` means anything that can be converted to `...`: `Space(SpaceLike)`.

But here it does not match `Qobj(QobjLike)`...

qutip/core/direct_sum.py
```diff
@@ -0,0 +1,676 @@
...
+
+ARRAY_ONE = np.full((1,), 1, dtype=_data.base.idxint_dtype)
```
Ericgig, line 25: Couldn't it support normal `int` as `dtype` or list inputs? Using `idxint_dtype` outside cython feels wrong...

qutip/core/direct_sum.py
```diff
@@ -0,0 +1,676 @@
...
+                             f" ({row}, {col}).")
+
+    return _do_direct_sum(dict(sorted(qobjs.items())), sum_dimensions, dtype)
+
+
+def _do_direct_sum(
+        qobjs: dict[tuple[int, int], QobjLike],
+        sum_dimensions: Dimensions,
+        dtype: LayerType = None
+):
```
Ericgig, line 354:
```suggestion
def _do_direct_sum(
    qobjs: dict[tuple[int, int], QobjLike],
    sum_dimensions: Dimensions,
    dtype: LayerType = None
):
```

qutip/core/direct_sum.py
```diff
@@ -0,0 +1,676 @@
...
+    on :math:`\mathbb C^2 \oplus \mathbb C^3`.
+
+    Parameters
+    ----------
+    qobjs : dict
+        Mapping that assigns a component to the given (row, col) location in
+        the block matrix. Missing locations are filled with zero blocks. The
+        components may be Python numbers, :class:`.Qobj`, or :class:`.QobjEvo`.
+    sum_dimensions : list of tuples or lists
+        Dimensions of the resulting Qobj, like in the creation of a Qobj.
```
Ericgig, line 321: How about adding a note about list vs tuple for dims here?

qutip/core/direct_sum.py
pmenczel: I see, that makes sense to me but I can't come up with a better name right now. I didn't use `QobjLike` in many places, I have removed it for now and just written it out explicitly everywhere.

qutip/core/direct_sum.py
pmenczel: [These functions](https://github.com/qutip/qutip/blob/master/qutip/core/data/block_operations.pyx#L30) need arrays of `idxint` as parameters. I don't think it can be converted automatically.

pmenczel:
I have updated the PR.

qutip/tests/core/test_direct_sum.py
```diff
@@ -0,0 +1,399 @@
...
+def test_matrix(arguments, result_type, result_dims, dtype):
+    result = direct_sum(arguments, dtype=dtype)
+
+    assert isinstance(result, (Qobj, QobjEvo))
+    assert result.type == result_type
+    assert result._dims == result_dims
+    assert result.shape == result_dims.shape
+    if result_type == "super":
+        assert result.superrep == "super"
+    if dtype is not None:
```
Ericgig: None is not a case from dtype.

qutip/tests/core/test_direct_sum.py
```diff
@@ -0,0 +1,399 @@
...
+    assert result.type == result_type
+    assert result._dims == result_dims
+    assert result.shape == result_dims.shape
+    if result_type in ["super", "operator-ket", "operator-bra"]:
+        assert result.superrep == "super"
+    if dtype is not None:
+        if isinstance(result, Qobj):
+            assert result.dtype.__name__ == dtype
+        else:
+            assert result(0).dtype.__name__ == dtype
```
Ericgig: This is repeated in multiple tests, how about a `check_meta` function?

qutip/tests/core/test_direct_sum.py
```diff
@@ -0,0 +1,399 @@
...
+
+    matrix = direct_sum([[sigmax(), None], [None, sigmay()]])
+    for args in (
+        [], [0], [-1], [-1, 0], [0, -1], [2, 0], [0, 2], [2, 2], [0, 0, 0]
+    ):
+        with pytest.raises(IndexError):
+            direct_component(matrix, *args)
+        with pytest.raises(IndexError):
+            set_direct_component(matrix, None, *args)
+            print(f"\nargs: {args}\n")
```
Ericgig: Left over debug print

pmenczel:
Note: closes #2384

PR #2834, opened 2026-03-07
Fixed and improved plot_spin_distribution

### LINKED ISSUE #2431: Change color map in plot_spin_distribution
### Describe the Issue!

Hi all,

I want to change the color map and contour linestyles and linewidth in `plot_spin_distribution_3d(Q, THETA, PHI)` (see below figure), but any effort fails. I would appreciate any suggestion. 

Husimi_Q

mudit06mah:
**Description**
This PR fixes a bug in `plot_spin_distribution` while also improving on the function by adding better `cmap` support and `**kwargs`

**Changes**
- Earlier `mpl.colors.Normalize()` was called inside `if cmap is None` block which caused `facecolors=cmap(norm(P))` to throw `UnboundLocalError: cannot access local variable 'norm' where it is not associated with a value`. This PR changes the if block accordingly. 
- Function now lets user pass name of a registered colormap instead of only a colormap object, eg: `cmap='jet'`.
- Added `**kwargs` to allow user to pass args like `linewidth`, `linestyle`.

Code:
```
fig, ax = plot_spin_distribution(Q, THETA, PHI, "3d", cmap='jet')
```
Now renders:
image

**Related issues or PRs**
Fixes #2431

### DIFF
--- a/doc/changes/2834.bugfix
+++ b/doc/changes/2834.bugfix
@@ -1,1 +1,1 @@
-Fixes UnboundLocalError in plot_split_distribution and improved function by changing cmap implementation and adding kwargs. 
+Fixes UnboundLocalError in plot_spin_distribution and improved function by changing cmap implementation and adding kwargs. 


mudit06mah:
@Ericgig @pmenczel Can you please review this PR?

doc/changes/2834.bugfix
```diff
@@ -0,0 +1 @@
+Fixes UnboundLocalError in plot_split_distribution and improved function by changing cmap implementation and adding kwargs.
```
pmenczel:
```suggestion
Fixes UnboundLocalError in plot_spin_distribution and improved function by changing cmap implementation and adding kwargs. 
```

PR #2836, opened 2026-03-09
ENH: restore OpenMP configuration surface

### LINKED ISSUE #1904: Restore openmp support on `dev.major`

arnavk23:
**Description**
This PR restores the OpenMP configuration surface for QuTiP by re-enabling OpenMP controls that were effectively disabled.

**Related issues or PRs**
Fixes #1904

### DIFF
--- a/doc/changes/1904.feature
+++ b/doc/changes/1904.feature
@@ -1,0 +1,1 @@
+Restores OpenMP support controls by re-enabling the build flag (``--with-openmp``/``CI_QUTIP_WITH_OPENMP``), exposing ``settings.has_openmp`` detection, and adding ``CoreOptions(use_openmp=..., openmp_thresh=...)`` toggles for runtime configuration.
--- a/qutip/core/options.py
+++ b/qutip/core/options.py
@@ -134,6 +134,12 @@ class CoreOptions(QutipOptions):
         ``auto_tidyup`` parameter above) and the default value of ``atol`` used
         in :meth:`Qobj.tidyup`.
 
+    use_openmp : bool [True]
+        Enable OpenMP-enabled kernels when they are available.
+
+    openmp_thresh : int [10000]
+        Minimum matrix size threshold before OpenMP kernels should be used.
+
     function_coefficient_style : str {"auto"}
         The signature expected by function coefficients. The options are:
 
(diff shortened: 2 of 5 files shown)


Ericgig:
This feature was left deactivated on purpose.

Ericgig:
Thank you for the interest in contributing.
We will try to open new good first issue or other reasonable contribution options.
Easy one have all been taken recently.

Openmp is not about just activating it, it's about packaging, interaction with other parallelism method, reviewing and adding it everywhere useful, etc.
It's not a quick fix.

PR #2837, opened 2026-03-12
ENH: Add plot_expect and plot_photocurrent to Result classes

Chinmay-Tangal:
Implements visualisation methods on Result objects as requested in #2617.

### Changes

**`Result.plot_expect()`**: Plot expectation values from deterministic solvers (`mesolve`, `sesolve`, etc.). Infers legend labels from `e_ops` dict keys when available. Uses solver name as default title.

**`MultiTrajResult.plot_expect()`**: Extends `plot_expect` for multi-trajectory solvers (`mcsolve`, `nm_mcsolve`). `show_trajectories=N` overlays the first N individual trajectories. `show_average` controls whether the mean is plotted. Warns gracefully when trajectories are requested but were not saved.

**`McResult.plot_photocurrent()`**: Plots the average photocurrent computed from collapse events, one curve per collapse operator.

All methods follow the existing `fig`/`axes` keyword-only convention from `visualization.py`. Default behavior requires no additional arguments.

### Usage

```python
# Basic usage with dict labels
result = mesolve(H, psi0, tlist, e_ops={"n": a.dag() * a})
fig, axes = result.plot_expect()

# MC result with trajectory overlay
result = mcsolve(H, psi0, tlist, c_ops, e_ops=[a.dag() * a],
                 ntraj=50, options={"keep_runs_results": True})
fig, axes = result.plot_expect(show_trajectories=10)

# Photocurrent
fig, axes = result.plot_photocurrent()

# Customisation
fig, axes = result.plot_expect(title="Cavity decay",
                               xlabel="t (μs)", ylabel="⟨n⟩")

### CODE REVIEW

**Comment by pmenczel** [comment]
Hi and thank you for your contribution. There was an open pull request for this issue already: https://github.com/qutip/qutip/pull/2724

The other PR seems to be inactive, but there was already a lot of discussion there. Perhaps you could review if any of the discussion or code there could still be useful here.

If this PR ends up getting merged, I would suggest crediting both @Chinmay-Tangal and the author of the other PR in the changelog.

Edit: but let me add that I like the customization options here, and the trajectory plotting as well! For the trajectories, it would be cool if the style of the trajectories was distinct from the style of the average plot. For example, maybe one wants the trajectories to be all the same color with low alpha, and then the average in a different color on top of all that.

**Comment by Chinmay-Tangal** [comment]
Thanks for the review @pmenczel! I've gone through the discussion and code on #2724.

Updated the trajectory styling so individual trajectories are visually distinct from the average — they now default to gray, low alpha, thin lines, while the average uses the standard color cycle with thicker lines. Users can fully customise trajectory appearance via the `trajectory_kwargs` parameter:

```python
result.plot_expect(
    show_trajectories=10,
    trajectory_kwargs={"color": "blue", "alpha": 0.2, "linestyle": "--"},
)

**Comment by Chinmay-Tangal** [comment]
Also credited @famous111 in the changelog for the initial work on #2724.

**Review by pmenczel — COMMENTED** [review]
I have played around with it a bit, and I think it works well and it is very cool! Two main comments:
* The new functions currently don't show up in the apidoc for McResult and NmmcResult, could you please fix that? ([see here](https://github.com/qutip/qutip/blob/master/doc/apidoc/solver.rst?plain=1#L60))
* Thank you for adding tests! You have put them in a new test module (`test_result_plot.py`), but I think it will be easier to quickly find them in the future if they were just added to the existing `test_results.py`, perhaps grouped as a class?

Optional suggestions:
* There could be an option to plot each expectation value in a separate plot, this was requested by someone in the previous PR.
* This is very useful functionality and it would be a shame if users don't use it, so I am wondering how to advertise it best. One nice option would be to add a small section in the Users' Guide, perhaps either in the `guide-visualization.rst` or as a new point in the "Time Evolution and Quantum System Dynamics" section.

**Comment by Chinmay-Tangal** [comment]
Thanks @pmenczel! Addressed all items:

**Required:**
- Fixed apidoc: `plot_expect` and `plot_photocurrent` now appear in McResult and NmmcResult documentation
- Moved all visualisation tests into the existing `test_results.py`

**Optional (implemented):**
- Added `separate_axes` parameter to `plot_expect` for one subplot per expectation value
- Added Users' Guide section on visualising solver results

File: doc/guide/dynamics/dynamics-visualization.rst [hunk 1]
```diff
@@ -0,0 +1,104 @@
...
+methods.  These make it easy to quickly inspect expectation values
+without writing boilerplate matplotlib code.
+
+Basic Usage
+===========
+
+After running any solver, call :meth:`~qutip.solver.result.Result.plot_expect`
+on the result:
+
+.. code-block:: python
```
pmenczel:
```suggestion
.. plot::
    :context: reset
```

doc/guide/dynamics/dynamics-visualization.rst
```diff
@@ -0,0 +1,104 @@
...
+                           e_ops=[a.dag() * a])
+    result.plot_expect()
+
+Dictionary Labels
+=================
+
+When ``e_ops`` is passed as a dictionary, the keys are used as legend labels
+automatically:
+
+.. code-block:: python
```
pmenczel:
```suggestion
.. plot::
    :context: close-figs
```
(same for all following code blocks)

qutip/tests/solver/test_results.py
```diff
@@ -528,3 +528,235 @@ def test_merge_nmmcresult(self, p):
 
         np.testing.assert_almost_equal(
             merged.trace, p * ensemble1.trace + (1 - p) * ensemble2.trace)
+
+
+# ── Visualisation tests ───────────────────────────────────────────
+
+mpl = pytest.importorskip("matplotlib", reason="matplotlib required")
+mpl.use("Agg")
+import matplotlib.pyplot as plt  # noqa: E402
```
pmenczel: Unfortunately, this will skip the entire module if matplotlib is not found. Perhaps we could do something like this?

```suggestion
@pytest.fixture
def plt():
    return pytest.importorskip("matplotlib.pyplot", reason="matplotlib required")
```

Note that the `mpl.use("Agg")` should be unnecessary (we are not setting that in `test_visualization` either, for example, but we set it globally in the workflows where necessary)

qutip/solver/result.py
```diff
@@ -364,6 +364,155 @@ def __repr__(self):
...
+                    self.times, expectation, label=label, **plot_kwargs
+                )
+
+            axes.set_xlabel(xlabel)
+            axes.set_ylabel(ylabel)
+
+            if title is None:
+                title = self.solver
+            if title is not None:
+                axes.set_title(str(title))
```
pmenczel:
```suggestion
            axes.set_title(title or self.solver)
```
should be equivalent, matplotlib does the `str()` for us.

(same in the other functions)

qutip/solver/result.py
```diff
@@ -364,6 +364,155 @@ def __repr__(self):
...
+                    for i in range(n_e_ops)
+                ])
+            elif fig is None:
+                if not hasattr(axes, '__len__'):
+                    axes = np.array([axes])
+                fig = axes[0].get_figure()
+
+            for i, (expectation, label) in enumerate(
+                zip(self.expect, labels)
+            ):
```
pmenczel: How about
```suggestion
            for ax, expectation, label in zip(axes, self.expect, labels):
```

qutip/solver/multitrajresult.py
```diff
@@ -694,6 +694,252 @@ def __repr__(self):
...
+                            self.times, exp,
+                            label=traj_label,
+                            **traj_kw,
+                        )
+
+            # --- average on each subplot ---
+            if show_average and avg_expect:
+                for i, (exp, label) in enumerate(
+                    zip(avg_expect, labels)
+                ):
```
pmenczel:
```suggestion
                for ax, exp, label in zip(axes, avg_expect, labels):
```

pmenczel (changes requested):
I think we are getting very close. I attach a few small comments on the code below.

Thank you very much for also including a users guide section and the separate axes option! I do see one problem with the separate axes: if we manually make a horizontal layout, like for example
```python
fig, ax = plt.subplots(1, 3, figsize=(12, 4))
result.plot_expect(fig=fig, axes=ax, separate_axes=True)
plt.tight_layout()
plt.show()
```
then it looks quite weird to have the title only on the first plot, and the x axis label only on the last.

One option might be to put labels on all axes, and call [label_outer](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.label_outer.html#matplotlib.axes.Axes.label_outer) on all axes. This removes the inner axes labels on all shared axes. I am not 100% sure whether that will work well in all scenarios, please let me know if you have additional insight.

Another straightforward option would be to put labels on all axes, except in the case where we have called `subplots` ourselves with a vertical layout.

Chinmay-Tangal:
Addressed all review comments:

- Switched to `.. plot::` directives with `:context:` in the Users' Guide
- Changed `plt` to a pytest fixture so the module isn't skipped when matplotlib is missing
- Simplified title setting to `axes.set_title(title or self.solver)`
- Replaced `enumerate(zip(...))` with `zip(axes, ...)`
- Used `label_outer()` for `separate_axes` so horizontal layouts display correctly

qutip/tests/solver/test_results.py
```diff
@@ -528,3 +528,240 @@ def test_merge_nmmcresult(self, p):
...
             merged.trace, p * ensemble1.trace + (1 - p) * ensemble2.trace)
+
+
+# ── Visualisation tests ───────────────────────────────────────────
+
+@pytest.fixture()
+def plt():
+    mpl = pytest.importorskip("matplotlib",
+                              reason="matplotlib required")
+    mpl.use("Agg")
```
pmenczel: Thank you for updating the tests. Could you just remove this line, please?

qutip/solver/result.py
```diff
@@ -364,6 +364,149 @@ def __repr__(self):
...
+
+            for label, expectation in zip(labels, self.expect):
+                axes.plot(
+                    self.times, expectation, label=label, **plot_kwargs
+                )
+
+            axes.set_xlabel(xlabel)
+            axes.set_ylabel(ylabel)
+
+            axes.set_title(title or self.solver)
```
pmenczel: I'm sorry, I think I made a mistake here. Now, there is no way to disable the title... Can we just add
```python
if title is None:
    title = self.solver
```
at the top of the function, and then call `set_title(title)` everywhere? That way, the title can be disabled with `title=""`.


```suggestion
            axes.set_title(title)
```

qutip/solver/result.py
```diff
@@ -364,6 +364,149 @@ def __repr__(self):
...
+                )
+                ax.set_ylabel(label)
+                if show_legend:
+                    ax.legend()
+
+            for ax in axes:
+                ax.set_xlabel(xlabel)
+                ax.set_title(title or self.solver)
+            for ax in axes:
+                ax.label_outer()
```
pmenczel: After trying it out, I don't think this is working well. Actually, `label_outer` also removes axis labels from axes that are not shared...

Let's not overcomplicate things and just remove this `label_outer`. Every plot will have its own title and axis labels then. That is fine for a quick utility method like this.

pmenczel:
```suggestion
                if show_legend:
                    ax.legend()
                ax.set_xlabel(xlabel)
                ax.set_title(title)
```

pmenczel (changes requested):
Hi and thank you for making the changes! See comments below: my suggestion for the title setting was not correct, sorry. Also, the `label_outer` did not work as expected, unfortunately.

I have only commented on the `result.py` module. Could you please also make the analogous changes in `multitrajresult.py`?

Chinmay-Tangal:
Addressed all comments:

- Removed `mpl.use("Agg")` from the plt fixture.
- Fixed title handling: `title = self.solver` set once at the top, `set_title(title)` everywhere — `title=""` now disables the title.
- Removed `label_outer()` — each subplot gets its own xlabel and title.
- Applied same changes to both `result.py` and `multitrajresult.py`.

pmenczel (approved):
Thank you again for your updates. I have made a few more small changes to improve the look of the plots, hope that is okay.

From my side, it is ready to merge now, but I will wait a little bit in case there are comments from other maintainers.

Chinmay-Tangal:
Thanks @pmenczel for the review and the additional polish — the plots look great. Happy to have this ready!

Ericgig (approved):
Work great.
Thank you.

doc/changes/2617.feature
```diff
@@ -0,0 +1 @@
+Added ``plot_expect()`` to ``Result`` and ``MultiTrajResult``, and ``plot_photocurrent()`` to ``McResult``, for visualising solver output directly from result objects. Based on initial work by @famous111 in PR #2724.
```
veronikakurth, line 1: We didn't notice the number of the PR in the name of this changelog is wrong, oops!
Such things may mess up release changelogs.

PR #2838, opened 2026-03-13
Unified caching

### LINKED ISSUE #453: Unified caching of expensive attributes
Currently, ``Qobj`` has several attributes which are somewhat expensive to compute:

- ``type``
- ``isherm``
- ``istp``
- ``iscp``
- ``ishp`` (added in #452)

Of these, ``type`` and ``isherm`` are cached by ``Qobj._type`` and ``Qobj._isherm``, respectively. These should be unified, so that the other three are also cached properly. Moreover, we should add unit tests to make sure that this caching is properly invalidated under at least some reasonably common manipulations. For instance, with the caching of ``type``, it can sometimes fail in unexpected ways to build an object up then redefine its ``dims``.

QubitGoblin31:
Description

Qobj has five expensive-to-compute properties — ishp, iscp, istp, iscptp, and isherm — but only isherm (via _isherm) and isunitary (via _isunitary) were cached. The other three carried # FIXME: this needs to be cached in the same ways as isherm comments. Additionally, no setter (dims, superrep, data) invalidated any cache, leading to silently stale results:

J = 2 * rho_out
print(J.type)           # caches 'oper'
J.dims = [[[2], [2]], [[2], [2]]]
J.superrep = 'choi'
print(J.type)           # wrongly prints 'oper'
This PR:

Adds _ishp, _iscp, _istp, _iscptp cache attributes, initialized to None in __init__ and propagated in copy()
Rewrites ishp, iscp, istp, iscptp to use the same lazy-cache pattern as isherm
Invalidates the appropriate caches in dims.setter, superrep.setter, and data.setter, with comments in each explaining the reasoning
Adds TestQobjCPTPCaching covering: correct values with cache reset, invalidation by each setter, cache preservation across copy(), and a regression test for the stale-type scenario above
Related issues or PRs

Fixes #453

### DIFF
--- a/qutip/core/qobj.py
+++ b/qutip/core/qobj.py
@@ -1381,7 +1381,7 @@ class Qobj:
             Quantum object with small elements removed.
         """
         atol = atol or settings.core['auto_tidyup_atol']
-        self.data = _data.tidyup(self.data, atol)
+        self._data = _data.tidyup(self._data, atol)
         return self
 
     def drop_scalar_dims(self, inplace: bool = False) -> Qobj:
@@ -1932,6 +1932,9 @@ class Qobj:
     @property
     def iscptp(self) -> bool:
         if self._iscptp is not None:
+            return self._iscptp
+        if self._iscp is not None and self._istp is not None:
+            self._iscptp = self._iscp and self._istp
             return self._iscptp
         if not (self.issuper or self.isoper):
             self._iscptp = False
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -246,7 +246,7 @@ class TestQobjCPTPCaching:
         rho_out = qutip.partial_transpose(rho, [0, 1])
         J = 2 * rho_out
         # Access properties while still an oper to populate the cache.
-        _ = J.type
+        _ = J.isunitary
         _ = J.ishp
         _ = J.iscp
         _ = J.istp
@@ -303,17 +303,6 @@ class TestQobjCPTPCaching:
         assert copied._iscp == channel._iscp
         assert copied._istp == channel._istp
         assert copied._iscptp == channel._iscptp
-
-    def test_type_correct_after_dims_superrep_change(self):
-        # Regression test for the bug in the issue: accessing .type before
-        # setting dims/superrep must not cause stale results afterwards.
-        rho = qutip.ket2dm(qutip.bell_state())
-        rho_out = qutip.partial_transpose(rho, [0, 1])
-        J = 2 * rho_out
-        assert J.type == 'oper'   # populates cache
-        J.dims = [[[2], [2]], [[2], [2]]]
-        J.superrep = 'choi'
-        assert J.type == 'super'
 
 
 def test_QobjDimsShape():


qutip/core/qobj.py
```diff
@@ -343,12 +369,22 @@ def data(self, data: _data.Data):
             raise ValueError('Provided data do not match the dimensions: ' +
                              f"{self._dims.shape} vs {data.shape}")
         self._data = data
+        # New matrix data invalidates every property derived from matrix
+        # values.  Note: dtype conversion in __init__ assigns _data directly
+        # to avoid triggering this (storage format changes do not affect any
+        # of these properties).
```
Ericgig, line 376: The setter is also used in tidyup and it should not reset the flag.

qutip/tests/core/test_qobj.py
```diff
@@ -201,6 +201,121 @@ def test_QobjUnitaryOper():
...
+        assert_cptp_properties(ket, ishp=False, iscp=False, istp=False, iscptp=False)
+
+    def test_cache_invalidated_by_dims_setter(self):
+        # Build an oper, cache its properties, then promote it to a super via
+        # dims + superrep. The cached values must be cleared.
+        rho = qutip.ket2dm(qutip.bell_state())
+        rho_out = qutip.partial_transpose(rho, [0, 1])
+        J = 2 * rho_out
+        # Access properties while still an oper to populate the cache.
+        _ = J.type
```
Ericgig:
```suggestion
        _ = J.isunitary
```
`type` is not cached.

qutip/tests/core/test_qobj.py
```diff
@@ -201,6 +201,121 @@ def test_QobjUnitaryOper():
...
+        _ = channel.iscp
+        _ = channel.istp
+        _ = channel.iscptp
+        copied = channel.copy()
+        assert copied._ishp == channel._ishp
+        assert copied._iscp == channel._iscp
+        assert copied._istp == channel._istp
+        assert copied._iscptp == channel._iscptp
+
+    def test_type_correct_after_dims_superrep_change(self):
```
Ericgig: `type` is no longer cached...

qutip/core/qobj.py
```diff
@@ -1831,62 +1868,78 @@ def dnorm(self, B: Qobj = None) -> float:
...
+            # Theory of Quantum Information (Fall 2011), theorem 5.4.
+            tr_oper = qobj.ptrace([0])
+            self._istp = np.allclose(tr_oper.full(), np.eye(tr_oper.shape[0]),
+                                     atol=settings.core['atol'])
+        return self._istp
 
     @property
     def iscptp(self) -> bool:
+        if self._iscptp is not None:
+            return self._iscptp
```
Ericgig, line 1935: `iscptp` is composed of `iscp` and `istp`, if both are defined, but not `iscptp`, it does not need to be fully recomputed.

qutip/core/qobj.py
```diff
@@ -300,12 +304,18 @@ def __init__(
...
+                      dims=self._dims,
+                      isherm=self._isherm,
+                      isunitary=self._isunitary,
+                      dtype=self.dtype,
+                      copy=True)
+        # use get_attr for jax
+        result._ishp = getattr(self, '_ishp', None)
+        result._iscp = getattr(self, '_iscp', None)
+        result._istp = getattr(self, '_istp', None)
+        result._iscptp = getattr(self, '_iscptp', None)
```
Ericgig, line 317: While this may fix the immediate error, a fix should be made on qutip-jax side.
I don't think the properties are tested there and it would fail at any time `self._ishp` etc. are used.

pmenczel: I believe this is where the fix would need to be made on the qutip-jax side: https://github.com/qutip/qutip-jax/blob/master/src/qutip_jax/qutip_trees.py

To be clear: that would be a separate PR in the qutip-jax repo. Here, it is okay to just do `self._ishp` etc instead of the `getattr` calls. (That's why I unresolved the conversation.)

pmenczel: Hi @QubitGoblin31, are you still working on this? The PR is almost ready, only this one point is missing. It is not required for you to do the necessary change in qutip-jax yourself.

Ericgig:
Review not addressed for a month...

PR #2839, opened 2026-03-15
ENH: Add offset support for wigner

### LINKED ISSUE #1417: Offset support in Wigner and other visualizations
For coherent states of large average photon number it is convenient to use the `offset` parameter as in http://qutip.org/docs/latest/apidoc/functions.html?highlight=wigner#qutip.states.coherent

However, `wigner` and others do not support that parameter and would just give wrong results.

Functions like http://qutip.org/docs/latest/apidoc/functions.html?highlight=wigner#qutip.wigner.wigner , http://qutip.org/docs/latest/apidoc/functions.html?highlight=wigner#qutip.visualization.plot_wigner_fock_distribution , and others could support an `offset` parameter or the `offset` parameter can be stored in the qobject.

**Describe alternatives you've considered**
Calculating these functions manually. It works but it is pretty annoying.

mudit06mah:
**Description**
This PR adds the `offset` parameter to the wigner function, thus improving performance.

**Changes**
- Adds `offset` arg to `wigner` and helper private method: `_wigner_clenshaw`, `_wigner_iterative` and `_wigner_laguerre`
- Adds `offset` arg to `plot_wigner` and `plot_flock_distribution`

**Related issues or PRs**
Fixes #1417 

**Benchmarking**
Code snippet used:
```
import time
import numpy as np
from qutip import coherent, wigner

def benchmark(N, alpha, offset=0, method='laguerre'):
    state = coherent(N, alpha, offset=offset)
    xvec = np.linspace(5, 15, 100)
    
    start_time = time.perf_counter()
    wigner(state, xvec, xvec, method=method, offset=offset)
    end_time = time.perf_counter()
    
    return end_time - start_time

time_base = time_offset = 0

for i in range(50):
    time_base += benchmark(150, 10.0, offset=0)

for i in range(50):
    time_offset += benchmark(60, 10.0, offset=70)

print(f"Performance Report ({time.strftime('%Y-%m-%d')})")
print("-" * 30)
print(f"Baseline (N=150): {time_base:.4f} seconds")
print(f"Offset (N=60):   {time_offset:.4f} seconds")
print(f"Speedup:         {time_base / time_offset:.2f}x faster")
```

For Laguerre:
```
Baseline (N=150): 917.6025 seconds
Offset (N=60):   229.2310 seconds
Speedup:         4.00x faster
```

For clenshaw:
```
Baseline (N=150): 19.4652 seconds
Offset (N=60):   9.2073 seconds
Speedup:         2.11x faster
```

**Notes for Reviewers**
- This fix doesn't work well for iterative method, offset also causes change in the plot.
- This PR doesn't add support for the fft method
- This PR doesn't add tests (if necessary I'll add test cases as well)

### DIFF
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -721,7 +721,7 @@ def test_wigner_offset_consistency(method):
     W_base = qutip.wigner(state_base, xvec, yvec, method=method, offset=0)
     W_offset = qutip.wigner(state_offset, xvec, yvec, method=method, offset=30)
 
-    np.testing.assert_allclose(W_base, W_offset, atol=1e-3)
+    np.testing.assert_allclose(W_base, W_offset, atol=1e-10)
 
 
 def test_wigner_offset_sparse_consistency():
--- a/qutip/wigner.py
+++ b/qutip/wigner.py
@@ -379,7 +379,7 @@ def _wigner_laguerre(rho, xvec, yvec, g, parallel, offset=0):
         for m in range(M):
             m_phys = m + offset
             if abs(rho[m, m]) > 0.0:
-                W += real(rho[m, m] * (-1) ** m * genlaguerre(m_phys, 0)(B))
+                W += real(rho[m, m] * (-1) ** m_phys * genlaguerre(m_phys, 0)(B))
             for n in range(m + 1, M):
                 n_phys = n + offset
                 if abs(rho[m, n]) > 0.0:
@@ -397,11 +397,7 @@ def _par_wig_eval(args):
     Private function for calculating terms of Laguerre Wigner function
     using parfor.
     """
-    if len(args) == 5:
-        m, rho, A, B, offset = args
-    else:
-        m, rho, A, B = args
-        offset = 0
+    m, rho, A, B, offset = args
 
     W1 = zeros(np.shape(A))
     m_phys = m + offset


qutip/wigner.py
```diff
@@ -330,36 +345,35 @@ def _wigner_laguerre(rho, xvec, yvec, g, parallel):
...
-
-                    if m == n:
-                        W += real(rho[m, m] * (-1) ** m * genlaguerre(m, 0)(B))
-
-                    elif n > m:
-                        W += 2.0 * real(rho[m, n] * (-1) ** m *
-                                        (2 * A) ** (n - m) *
-                                        sqrt(factorial(m) / factorial(n)) *
-                                        genlaguerre(m, n - m)(B))
+        for m in range(len(rho.data.indptr) - 1):
```
pmenczel: What happened to the parallelization here?

pmenczel:
Thank you for yet another contribution! Just a preliminary review.

> This fix doesn't work well for iterative method, offset also causes change in the plot.

This is a bit suprising, shouldn't it be exactly equivalent with and without offset? (Note that I am not deeply familiar with this stuff.)

> This PR doesn't add support for the fft method

Okay, but let's add an error message or warning if the fft method is selected an a non-zero offset is specified.

> This PR doesn't add tests (if necessary I'll add test cases as well)

It would be good to add some test cases including offset to the `test_wigner` test suite.

mudit06mah:
@pmenczel Thank you for your Review!

I have now:
- Added an error message for the fft method.
- Added test cases for the offset parameter.
- I mistakenly removed parallelization when I made the changes, since those changes didn't get caught by the tests I realized that a test case for parallelization would also be beneficial, so I have added one test case for that.
- Improved some of the surrounding code formatting to match pycodestyle.

Regarding the iterative method:
From my understanding, the iterative method does not save computation as it is computed from the vaccum state all the way up to the final state (|0>, |1>, |2>, ... |n>) hence we cannot really benefit from using `offset` by skipping some of the initial states or computing backwards, we have to start from 0.
Also every step of the computation adds a slight floating point error(noise) which accumulates to give a visually different diagram.

Can you please review this again?

qutip/wigner.py
```diff
@@ -369,19 +397,27 @@ def _par_wig_eval(args):
     Private function for calculating terms of Laguerre Wigner function
     using parfor.
     """
-    m, rho, A, B = args
+    if len(args) == 5:
```
pmenczel: I don't think this function is ever called with 4 arguments

qutip/wigner.py
```diff
@@ -332,34 +352,42 @@ def _wigner_laguerre(rho, xvec, yvec, g, parallel):
...
+                                        (B))
+
     else:
         # for dense density matrices
-        B = 4 * abs(A) ** 2
         for m in range(M):
+            m_phys = m + offset
             if abs(rho[m, m]) > 0.0:
-                W += real(rho[m, m] * (-1) ** m * genlaguerre(m, 0)(B))
+                W += real(rho[m, m] * (-1) ** m * genlaguerre(m_phys, 0)(B))
```
pmenczel:
```suggestion
                W += real(rho[m, m] * (-1) ** m_phys * genlaguerre(m_phys, 0)(B))
```

qutip/tests/test_wigner.py
```diff
@@ -692,3 +692,46 @@ def test_spin_wigner_overlap(spin, pure, n=5):
...
+
+    state_offset = qutip.rand_ket(40)
+    padded_data = np.pad(state_offset.full(), ((30, 0), (0, 0)),
+                         mode='constant')
+    state_base = qutip.Qobj(padded_data)
+
+    W_base = qutip.wigner(state_base, xvec, yvec, method=method, offset=0)
+    W_offset = qutip.wigner(state_offset, xvec, yvec, method=method, offset=30)
+
+    np.testing.assert_allclose(W_base, W_offset, atol=1e-3)
```
pmenczel:
```suggestion
    np.testing.assert_allclose(W_base, W_offset, atol=1e-10)
```
Passes for me (after fixing the Laguerre typo)

pmenczel (changes requested):
Thank you, I just found one typo in the laguerre method and had a few other small comments on the code.

> Regarding the iterative method:
From my understanding, the iterative method does not save computation as it is computed from the vaccum state all the way up to the final state (|0>, |1>, |2>, ... |n>) hence we cannot really benefit from using offset by skipping some of the initial states or computing backwards, we have to start from 0.
Also every step of the computation adds a slight floating point error(noise) which accumulates to give a visually different diagram.

I agree that the iterative method is not faster with offset. I don't understand where the floating point error would come from, since both computations are identical. In fact, when I run your `test_wigner_offset_consistency("iterative")` test, the `W_base` and `W_offset` results are exactly identical. (Anyway, that is even better, then.)

mudit06mah:
@pmenczel I guess you're correct about the noise part, what I encountered was a visual difference between the base and offset plots:
image

Code snippet i used:
```
import numpy as np
from qutip import coherent, plot_wigner

xvec = np.linspace(-2, 10, 100)
yvec = np.linspace(-5, 5, 100)
alpha = 4.0

state_base = coherent(35, alpha)
state_offset = coherent(20, alpha, offset=10)

fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12, 5))

plot_wigner(state_base, xvec=xvec, yvec=yvec, method='iterative', offset=0, ax=ax1)
ax1.set_title("Baseline (N=35, offset=0)\nMethod: Iterative")

plot_wigner(state_offset, xvec=xvec, yvec=yvec, method='iterative', offset=10, ax=ax2)
ax2.set_title("Offset (N=20, offset=10)\nMethod: Iterative")

plt.tight_layout()
plt.show()
```

It is creating some sort of ripples after I introduce the offset parameter, while it is mathematically identical bit-for-bit.

PR #2841, opened 2026-03-16
Add Qobj.full_tensor convenience method

### LINKED ISSUE #2353: Add Qobj method for getting data as a numpy array shaped like a tensor rather than an array
### Problem Description

There is no built in method for converting a Qobj into a numpy array such that the `shape` of the numpy array reflects the `dims` of the Qobj. While fixing this is a oneliner with `np.reshape`, there are many opportunities for mistakes here, such as making sure the numpy reshape convention (row-major or column-major) matches the Qutip convention.

### Proposed Solution

Add a `full_tensor` and `get_data_tensor` method to `Qobj` which behave like `full` and `get_data` except the returned numpy array has a shape equal to `dims` flattened into a 1D list.

### Alternate Solutions

Instead of a new method, `full` and `get_data` could be given an optional parameter `keep_shape: bool = false`

### Additional Context

_No response_

sovopr:
## Summary
- add `Qobj.full_tensor()` as a thin wrapper around `qutip.core.dimensions.to_tensor_rep`
- add a regression test covering the tensor-shaped ndarray output for a composite state

## Testing
- `MPLCONFIGDIR=/tmp/matplotlib .venv/bin/python -m pytest qutip/tests/core/test_qobj.py -q`

Closes #2353

### DIFF
--- a/qutip/tests/core/test_qobj.py
+++ b/qutip/tests/core/test_qobj.py
@@ -1348,27 +1348,17 @@ def test_data_as():
             id="ket-no-auto-tidyup",
         ),
         pytest.param(
-            True,
-            lambda: qutip.tensor(qutip.qeye(2), qutip.qeye(3)),
-            (2, 3, 2, 3),
-            id="operator-auto-tidyup",
-        ),
-        pytest.param(
             False,
-            lambda: qutip.tensor(qutip.qeye(2), qutip.qeye(3)),
-            (2, 3, 2, 3),
+            lambda: qutip.tensor(qutip.qeye(2), qutip.qeye(3), qutip.qeye(1)),
+            (2, 3, 1, 2, 3, 1),
             id="operator-no-auto-tidyup",
         ),
         pytest.param(
-            True,
-            lambda: qutip.to_super(qutip.tensor(qutip.qeye(2), qutip.qeye(3))),
-            (2, 3, 2, 3, 2, 3, 2, 3),
-            id="super-auto-tidyup",
-        ),
-        pytest.param(
             False,
-            lambda: qutip.to_super(qutip.tensor(qutip.qeye(2), qutip.qeye(3))),
-            (2, 3, 2, 3, 2, 3, 2, 3),
+            lambda: qutip.to_super(
+                qutip.tensor(qutip.qeye(2), qutip.qeye(3), qutip.qeye(1))
+            ),
+            (2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1),
             id="super-no-auto-tidyup",
         ),
     ],


qutip/core/qobj.py
```diff
@@ -859,6 +859,20 @@ def full(
...
+        Return the dense ndarray representation of the quantum object reshaped
+        as a tensor according to its dimensions.
+
+        Returns
+        -------
+        data : numpy.ndarray
+            Dense ndarray representation of the quantum object reshaped as a
+            tensor according to its dimensions.
+        """
+        from qutip.core.dimensions import to_tensor_rep
```
Ericgig: `to_tensor_rep` is already imported in qobj.py.

qutip/core/qobj.py
```diff
@@ -859,6 +859,20 @@ def full(
...
+        """
+        Return the dense ndarray representation of the quantum object reshaped
+        as a tensor according to its dimensions.
+
+        Returns
+        -------
+        data : numpy.ndarray
+            Dense ndarray representation of the quantum object reshaped as a
+            tensor according to its dimensions.
+        """
```
Ericgig, line 878: The same short description is used twice...

Maybe adding some examples would be good.

qutip/tests/core/test_qobj.py
```diff
@@ -1332,6 +1332,15 @@ def test_data_as():
     assert "dia_matrix" in str(err.value)
 
 
+def test_full_tensor():
+    with qutip.CoreOptions(auto_tidyup_dims=False):
+        state = qutip.tensor(qutip.basis(2, 0), qutip.basis(2, 1))
+        tensor = state.full_tensor()
+
+    assert isinstance(tensor, np.ndarray)
+    assert tensor.shape == (2, 2, 1, 1)
```
Ericgig: This would deserve more than one test.
Maybe both with and without "auto_tidyup_dims".
Also different type: ket, operator, superoperator...

sovopr:
Thanks for the review! I've addressed the feedback in the latest commits:
Removed the duplicate import.
Cleaned up the docstring and added an example showing the expected output.
Expanded the test suite using pytest.mark.parametrize to cover kets, operators, and superoperators, both with and without auto_tidyup_dims.
Let me know if anything else needs tweaking!

qutip/tests/core/test_qobj.py
```diff
@@ -1332,6 +1332,56 @@ def test_data_as():
...
+            lambda: qutip.tensor(qutip.basis(2, 0), qutip.basis(2, 1)),
+            (2, 2, 1, 1),
+            id="ket-no-auto-tidyup",
+        ),
+        pytest.param(
+            True,
+            lambda: qutip.tensor(qutip.qeye(2), qutip.qeye(3)),
+            (2, 3, 2, 3),
+            id="operator-auto-tidyup",
+        ),
```
Ericgig: Auto tidyup dims does not apply for this operator.
You could add a dummy `qutip.qeye(1)` that will be cleaned or remove this case. (and the super + tidy later)

sovopr: Good point about the operator dimensions! I removed the auto_tidyup=True tests for the operators/superoperators since the 1s are natively preserved in those structures, but I kept the dummy qeye(1) cases for auto_tidyup=False to ensure full_tensor still handles them correctly. I have Pushed the update!

sovopr:
thank you for reviewing  @Ericgig! would it be fine I drop a mail to you regarding my GSoC proposal?

Ericgig:
Better contact the qutip-admin mailing list. I will probably not have time to respond.

sovopr:
Done! Thanks @Ericgig.

PR #2845, opened 2026-03-17
Fix Bloch attributes not updating after clear() (#1106)

### LINKED ISSUE #1106: Updating Bloch sphere attributes seems impossible

suryanshcode2601:
## Summary
Fixes issue #1106 where `Bloch` sphere attributes such as `view` and
`vector_color` are not respected when updated between animation frames
after calling `clear()`.

## Root cause
Two bugs combined to cause this:

1. `clear()` was resetting `self.vector_color` and `self.point_color`
   which are user-set attributes, not data. `clear()` should only wipe
   plotted data (points, vectors, annotations etc.), not attributes.

2. `render()` only applied `self.view` when creating axes for the first
   time (`if self.axes is None`). On subsequent renders the axes already
   existed, so the updated `self.view` was never passed to matplotlib.

## Changes
- Removed `self.vector_color` and `self.point_color` from `clear()`
- Added `self.axes.view_init(azim=self.view[0], elev=self.view[1])`
  after axes creation in `render()` so view is always re-applied

## Testing
Ran `pytest qutip/tests/test_bloch.py -v` locally:
- 24 passed, 0 failed

Style check:
- `pycodestyle qutip/bloch.py` — clean

## Related
- Fixes #1106

### DIFF
--- a/doc/changes/2845.bugfix
+++ b/doc/changes/2845.bugfix
@@ -1,0 +1,2 @@
+```
+Fixed ``Bloch`` attributes such as ``view`` and ``vector_color`` not being respected after ``clear()`` is called (issue #1106).


doc/changes/2843.bugfix
```diff
@@ -0,0 +1 @@
+Fixed ``Bloch`` attributes such as ``view`` and ``vector_color`` not being respected after ``clear()`` is called (issue #1106).
```
veronikakurth, line 1: This changelog item refers to another PR, please leave only the relevant one.

qutip/bloch.py
```diff
@@ -314,8 +314,6 @@ def clear(self):
         self.point_alpha = []
         self.vector_alpha = []
         self.annotations = []
-        self.vector_color = []
```
veronikakurth, line 317: Why are these properties removed?

PR #2847, opened 2026-03-20
bloch: Add kwargs to vectors and points

### LINKED ISSUE #1105: Three issues with customizing plots on Bloch spheres
Hello again. Since these are very simple and related issues I will adress them in a single post.

**1) Adding control over plotted line width**
When we create a Bloch sphere with `b = Bloch()` and add points, `pnts_list`, to it with `b.add_points(pnts_list, meth = 'l')` we can use this handy `meth` parameter to instead of plotting points (which corresponds to the value  `'s'`) connect them with line segments (which corresponds to the value  `'l'`).

We can control de size of the points with `b.point_size` but we can't do it for lines (when `meth = 'l'` is been used). It would be nice to have that option, to allow control over the width of the lines plotted just like with points. In that way we could prevent it to be confused with the frame of the bloch sphere in a black and white image.

Maybe a good way of solving this would be to use the value of `b.point_size` for that. Or maybe a better option would be to have a new `b.line_size` parameter to allow for different sizes between points and lines (which makes everything more customizable).
_______________________________________________________________

**2) Adding styles for markers and lines for plotting**
In [the code](http://qutip.org/docs/4.0.2/modules/qutip/bloch.html) for the add_points function it is written that

```
meth : str {'s', 'm', 'l'}
            Type of points to plot, use 'm' for multicolored, 'l' for points
            connected with a line.
```

First of all, it is not true that `meth = 'l'` gives points connected by lines (which in pyplot would be done with the `linestyle='-'` and `marker='o'` parameters) but it gives lines without the point markers (`linestyle='-'` and `marker='None'` in pyplot). But this is not important since it is just a comment on a function.

The thing that matters is that it would be great if we had a `b.linestyle` and `b.marker` methods for the Bloch sphere instead of a `meth` parameter which can only show either points or lines without them, just like/similar to the [line styles](https://matplotlib.org/3.1.1/gallery/lines_bars_and_markers/linestyles.html) and [marker styles](https://matplotlib.org/3.1.1/api/markers_api.html) of pyplot. Or any other idea that could allow the user to create more complicated graphs on the Bloch sphere.

_______________________________________________________________

**3) Adding edge color parameter for plotted points**
The `b.point_color` method changes the color of the points but it doen't change the edge color of those marker points. This is important since right now one can only choose between two restrictive options; either we plot the points with large sizes ([like in this image](https://i.imgur.com/pGhU55e.png)) so the chosen color can be seen, or we plot see the dots as black points if we chose a smaller size for them ([like in thsi image](https://i.imgur.com/5Ql5sKM.png)). If I want to plot small points in red I just can't since the black edge will dominate in the plot.

Just like in pyplot, it would be good to have an obtion to toggle the edge or choose the edge color. Or if that is too complicated, at least make the marker edge color the same as `b.point_color`.

mudit06mah:
**Description**
This PR adds `**kwargs` to arguments of: `add_state`, `add_vector`, `add_points`
Hence making the interface consistent among the methods `add_state`, `add_vector`, `add_points`, `add_arc` and `add_line`.

**Related issues or PRs**
Fixes #1105 and part of #1913 

**Results**
Used code snippet:
```
import numpy as np
from qutip import Bloch, basis

b = Bloch()

vec_x = [1, 0, 0]

b.add_vectors(vec_x, 
              linewidth=10,
              linestyle='-.', 
              facecolor='pink', 
              edgecolor='blue')

state_z = basis(2, 0)
b.add_states(state_z, 
             kind='point', 
             s=500,            
             marker='*',       
             facecolor='gold', 
             edgecolor='red',  
             linewidth=2)      

points = [
    [0, 1, 0],
    [0, -1, 0],
]
points_array = np.array(points).T

b.add_points(points_array, 
             meth='l', 
             linewidth=4,
             linestyle=':',
             color='green',
             marker='D',
             markersize=12,
             markeredgecolor='black')

b.show()
plt.show()
```

Output:
image

### DIFF
--- a/qutip/bloch.py
+++ b/qutip/bloch.py
@@ -466,7 +466,7 @@ class Bloch:
         alpha : float, default=1.
             Transparency value for the vectors. Values between 0 and 1.
 
-        kwargs : dict
+        **kwargs : dict
             Additional parameters for matplotlib.patches.FancyArrowPatch.
 
         """
@@ -523,7 +523,7 @@ class Bloch:
             or escape backslashes
             e.g. "$\\\\langle x \\\\rangle$".
 
-        kwargs :
+        **kwargs :
             Options as for mplot3d.axes3d.text, including:
             fontsize, color, horizontalalignment, verticalalignment.
 
@@ -865,7 +865,7 @@ class Bloch:
             alpha = self.vector_alpha[k]
             color = self.vector_color[k]
 
-            if color is None or (isinstance(color, list) and color == None):
+            if color is None:
                 idx = k % len(self.vector_default_color)
                 color = self.vector_default_color[idx]
 


qutip/bloch.py
```diff
@@ -843,17 +864,29 @@ def plot_vectors(self):
 
             alpha = self.vector_alpha[k]
             color = self.vector_color[k]
-            if color is None:
+
+            if color is None or color == [None]:
```
veronikakurth: Having `color.all() == [None]` in the second operand would fix the failing tests.

mudit06mah: It appears that since I have not installed ipython many of the test cases were skipped when I was running the tests locally.
I have added an instance check for list since this check is actually invalid for numpy arrays, this passes all the test cases successfully :)

veronikakurth: Same for me; it took me time to understand why I cannot reproduce CI failures despite having matplotlib installed. I think this silent skipping could be improved but that's another topic! 
The change looks good to me, thanks!

veronikakurth:
@pmenczel the changes look good to me, I also ran the proposed code snippet - now the Bloch sphere visualization is more customizable w.r.t. its graphical elements (points and lines, basically). 

**UPD**: I will check the PR once again and submit my final review/approval in order to follow a more clear review workflow.

qutip/bloch.py
```diff
@@ -449,6 +466,9 @@ def add_vectors(self, vectors, colors=None, alpha=1.0):
         alpha : float, default=1.
             Transparency value for the vectors. Values between 0 and 1.
 
+        kwargs : dict
```
veronikakurth: Would suggest to keep to one kwargs documentation style (e.g. "**kwargs", as in other functions)

veronikakurth: Other than that, I think it's good to go.

mudit06mah: I have made the necessary changes (changed docstring of another function as well)

veronikakurth: Thank you!

veronikakurth:
@Ericgig @pmenczel The changes look good to me - can anyone please merge it? 
Thanks!

qutip/bloch.py
```diff
@@ -843,17 +864,29 @@ def plot_vectors(self):
 
             alpha = self.vector_alpha[k]
             color = self.vector_color[k]
-            if color is None:
+
+            if color is None or (isinstance(color, list) and color == None):
```
Ericgig: What is the situation where color is a list and equal to None?

mudit06mah: This should actually be `[None]`, and it was for handling cases when `color[k]` was None since add_state wrapped around the color in a list, but I think we can just change it to `if color is None:` since I have changed the add_state code to pass on None.

Ericgig (approved):

PR #2848, opened 2026-03-24
Qobj __add__ / __sub__ type annotations

JacobHast:
**Description**
Add type annotations to `core.qobj._require_equal_type` to facilitate proper typing for `Qobj.__add__` and `Qobj.__sub__`.


Ericgig:
Why did you add these types hints?
I already have working types hints with `Qobj.__add__`...
They don't work for you?

JacobHast:
> Why did you add these types hints? I already have working types hints with `Qobj.__add__`... They don't work for you?

Which IDE do you use? It doesn't work for me in VSCode:
image

pmenczel:
For me in VSCode (with Pylance):

image

JacobHast:
Interesting! I did some testing and found that I don't get correct hints because I'm using the [ty extension](https://marketplace.visualstudio.com/items?itemName=astral-sh.ty) which overwrites Pylance. If I go back to Pylance I get proper hints like you guys, so apparently Pylance does a better job at filling in the blanks for cases like this. 

However, it still looks like there's some difficulty for some common type-checkers in recognizing proper types here. E.g., for me mypy fails to raise an error in this example:

```
import qutip

x = qutip.Qobj()

def accepts_string(s: str):
    return

accepts_string(x + x) # mypy doesn't complain

accepts_string(x) # mypy complains

y = x + "dummy_string" # mypy doesn't complain

```

So the added type annotations to `_require_equal_type` should still make type-checking more robust.

qutip/core/qobj.py
```diff
@@ -39,15 +40,16 @@
     ('oper', 'ket'),
 }
 
+T = TypeVar('T')
```
Ericgig, line 43: Since the output is always `Qobj`, why not using it directly instead of TypeVar?
Is it to future proof?

JacobHast: In principle the output of the function wrapped by `_require_equal_type` could be anything, so I didn't want make assumptions in case it would be used for other methods than `__add__` and `__sub__` in the future. But you're right that since in the current code `_require_equal_type` is only used by `__add__` and `__sub__` the output could also be `Qobj`. Either is fine by me, if you prefer one way or the other :)

PR #2853, opened 2026-04-03
docs: clarify fmmesolve spectra_cb documentation and relationship to BosonicEnvironment

### LINKED ISSUE #2828: Inconsistency in the definition of the spectral density
### Describe the Issue!

Hi! I am new to this project, and trying to calculate dissipative Floquet evolution. It seems that the definition of the spectral density in the `BosonicEnvironment` module may be inconsistent with the one used in the `fmmesolve` module. 
In the `BosonicEnvironment` module, as stated in [BosonicEnvironment](https://qutip.readthedocs.io/en/stable/guide/environments/bosonic.html#theory),
$$J(\omega)\coloneqq\pi\sum_k|g_k|^2\delta(\omega-\omega_k)$$
However, in the `fmmesolve` module, according to [fmmesolve](https://qutip.readthedocs.io/en/stable/guide/dynamics/dynamics-floquet.html),
$$\gamma_{\alpha\beta k}=2\pi J(\Delta_{\alpha\beta k})|X_{\alpha\beta k}|^2$$
which seems to imply that
$$J(\omega)\coloneqq\sum_k|g_k|^2\delta(\omega-\omega_k)$$
In fact, I think a similar issue already appears in the reference [Grifoni98](https://www.sciencedirect.com/science/article/pii/S0370157398000222#aep-abstract-id38).
In Eqs. (233), (234), and (249), $J$ appears to include the factor $\pi$, while in Eq. (252) the  $J$ used there seems not to include this factor.
I am not sure whether my derivation is incorrect or whether there is indeed an inconsistency here. I would greatly appreciate any clarification.

Patbby:
## Summary

This PR clarifies the documentation for `fmmesolve()` regarding the `spectra_cb` parameter, resolving the confusion reported in issue #2828 about inconsistent definitions of spectral density between `BosonicEnvironment` and `fmmesolve`.

## Problem

Issue #2828 reported that the spectral density `J(ω)` appears to have different definitions in:
- `BosonicEnvironment` documentation: `J(ω) = π Σ g_k² δ(ω - ω_k)`
- `fmmesolve` documentation: `γ = 2π J(Δ) |X|²`

This caused confusion about whether the same `J` could be used in both contexts.

## Solution

The documentation now clarifies that:
1. `spectra_cb` expects the **noise power spectrum** `S(ω)`, not the spectral density `J(ω)`
2. These are related by: `S(ω) = 2π J(ω) [n_th(ω) + 1]` (finite temperature) or `S(ω) = 2 J(ω)` (zero temperature)
3. When using `BosonicEnvironment`, use the `.power_spectrum()` method to obtain the correct spectrum

## Changes

- Expanded `spectra_cb` parameter description with formula `γ = S(Δ) |X|²`
- Added note distinguishing noise power spectrum from spectral density
- Added example showing correct usage with `BosonicEnvironment.power_spectrum`
- Added Notes section explaining Floquet-Markov master equation context
- Added relationship formula between `S(ω)` and `J(ω)`

## Related Issue

Resolves #2828

## Checklist

- [x] Improved docstrings with clear explanations
- [x] Added practical code examples
- [x] Added changelog entry
- [x] Follows qutip documentation style
- [x] Created feature branch (not master)

*Created by AI agent OpenClaw*

### DIFF
--- a/qutip/solver/floquet.py
+++ b/qutip/solver/floquet.py
@@ -759,31 +759,6 @@ def fmmesolve(
         Other options could be supported depending on the integration method,
         see `Integrator <./classes.html#classes-ode>`_.
 
-    Notes
-    -----
-    The Floquet-Markov master equation describes the dynamics of a quantum
-    system subject to a periodic driving Hamiltonian and weak coupling to
-    a thermal environment. The dissipation is characterized by the noise
-    power spectrum :math:`S(\omega)` provided via `spectra_cb`.
-
-    **Relationship to BosonicEnvironment:**
-
-    When using :class:`.BosonicEnvironment` to describe the environment,
-    note that `spectra_cb` expects the noise power spectrum :math:`S(\omega)`,
-    not the spectral density :math:`J(\omega)`. These are related by:
-
-    .. math::
-        S(\omega) = 2 J(\omega) [n_{th}(\omega) + 1] \quad \text{for } \omega > 0
-
-    where :math:`n_{th}(\omega)` is the thermal occupation number.
-    At zero temperature, this simplifies to :math:`S(\omega) = 2 J(\omega)`.
-
-    Use :meth:`.BosonicEnvironment.power_spectrum` to obtain the correct
-    spectrum for `fmmesolve`:
-
-    >>> env = DrudeLorentzEnvironment(T=0.1, lam=0.01, gamma=0.1)
-    >>> spectra_cb = [env.power_spectrum]
-
     Returns
     -------
     result: :class:`.Result`


qutip/solver/floquet.py
```diff
@@ -734,6 +759,31 @@ def fmmesolve(
         Other options could be supported depending on the integration method,
         see `Integrator <./classes.html#classes-ode>`_.
 
+    Notes
```
Ericgig: This note would fit better in the [guide](https://qutip.readthedocs.io/en/latest/guide/dynamics/dynamics-floquet.html#the-floquet-markov-master-equation-in-qutip). We have the physic description there. In the guide, the noise spectral-density description is lacking and moving this there would be great.

qutip/solver/floquet.py
```diff
@@ -734,6 +759,31 @@ def fmmesolve(
...
+    power spectrum :math:`S(\omega)` provided via `spectra_cb`.
+
+    **Relationship to BosonicEnvironment:**
+
+    When using :class:`.BosonicEnvironment` to describe the environment,
+    note that `spectra_cb` expects the noise power spectrum :math:`S(\omega)`,
+    not the spectral density :math:`J(\omega)`. These are related by:
+
+    .. math::
+        S(\omega) = 2 \pi J(\omega) [n_{th}(\omega) + 1] \quad \text{for } \omega > 0
```
pmenczel: The "\pi" here is wrong (compare to your zero-temperature formula two lines below).

Patbby:
@Ericgig @pmenczel Thank you both for the reviews!

**Fixed the mixed PR issue:**

I realized this branch accidentally included commits from PR #2852. I have cleaned it up by creating a fresh branch and cherry-picking only the two commits belonging to this PR, then force-pushed. The PR now only contains the intended changes to floquet.py and the 2828.doc changelog.

**Regarding the review comments:**

1. π formula error - Already fixed (line 776 now correctly shows S(ω) = 2 J(ω) without the π)

2. Moving Notes to guide - Happy to do that in a follow-up PR if preferred, but the current docstring note addresses the immediate confusion from issue #2828.

Please let me know if any other changes are needed!

qutip/solver/floquet.py
Ericgig: > 2. Moving Notes to guide - Happy to do that in a follow-up PR if preferred, but the current docstring note addresses the immediate confusion from issue [Inconsistency in the definition of the spectral density #2828](https://github.com/qutip/qutip/issues/2828).

Sorry, but it's not as much that I want you to add more information in the guide, but remove the redundant note from the docstring. The note here and the one in the spectra_cb description are very similar and only one is needed.

Also it does not match the flow of the docstring... Why is is between the attributes and the return? Why does it explain what Floquet-Markov master equation is? If you feel it's insufficient, please improve it and the start, but do not add another description in the middle of the docstring...

Ericgig:
Sorry, but I will close this mostly because we don't know if it's always right or not...
`fmmesolve` take a temperature to apply on top of the spectra.
The bath power_spectrum is build from the spectral density and the temperature.
If the temperature is `0` in both case, then your example is probably right, but in other cases, it's probably not. `fmmesolve`  probably expect the `spectral_density` and apply the temperature itself and passing the power_spectrum with the temperature will effectively apply it twice...

So if you can derive the solver equations and prove that `power_spectrum` and how it interact with the solver temperature, we would be interested in seeing it.

But as is, I don't get the feeling that you us (nor us) have enough understanding to properly explain what needed. (Please feel free prove us wrong, we need a floquet expert.)

PR #2855, opened 2026-04-04
Add linux, windows arm64 to wheels build target

### LINKED ISSUE #2832: Publish Python Wheels for Linux arm64
### Problem Description

Currently, Linux Arm64 pre-built wheels are not published for qutip. This adds extra steps to install the package on those platforms because it must be compiled from source. I have observed very positive performance results using qutip on Linux Arm64 cloud compute hosts, but because there is no Linux Arm64 wheel, the install process for qutip is more complicated than with other platforms. 

### Proposed Solution

Github now provides Windows Arm64 and Linux Arm64 runners for github actions. 
https://github.blog/changelog/2026-01-29-arm64-standard-runners-are-now-available-in-private-repositories/

It should be straightforward to add Linux arm64 and optionally windows-arm64 to the list of platforms in the github actions workflows that build packages for and test the qutip library. The "ubuntu-24.04-arm" runner would be equivalent to the ubuntu-latest amd64 runner that is currently used by the qutip workflows.

### Alternate Solutions

The current approach of building from source produces a usable package, however it is not ideal in use cases like mine. I'm building a multi-platform Docker image and building the arm docker image takes 45 minutes when running on amd64 because I need to perform the compilation under qemu. Switching the docker build to run on github arm64 runners reduced the compile time to 5 minutes, but this is still much longer than the amd64 docker build that is able to use a pre-compiled qutip wheel. 

I see that there is an aarch64 version package of qutip available on conda-forge, but my team has migrated to use uv for package management, which would not be compatible with conda-forge.

### Additional Context

Thanks for reading my request, let me know if anything is unclear.

Mayank447:
**Description**
Add ubuntu-24.04-arm, windows-11-arm as wheels build targets.

For reference:
https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

**Related issues or PRs**
Fixes #2832

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,6 +54,16 @@ jobs:
             # Should be enough to include this in one of the runs
             includempi: 1
 
+          - case-name: p314 ubuntu arm
+            os: ubuntu-24.04-arm
+            python-version: "3.14"
+            numpy-build: ">=2.2.0"
+            numpy-requirement: ">=2.2.0"
+            pypi: 1
+            # Install mpi4py to test mpi_pmap
+            # Should be enough to include this in one of the runs
+            includempi: 1
+
           - case-name: p312, numpy fallback
             os: ubuntu-latest
             python-version: "3.12"
@@ -87,6 +97,7 @@ jobs:
             pypi: 1
             coveralls: 1
             pytest-extra-options: "-W ignore:dep_util:DeprecationWarning -W \"ignore:The 'renderer' parameter of do_3d_projection\""
+
           # Mac
           # Mac has issues with MKL since september 2022.
           - case-name: macos
@@ -110,7 +121,7 @@ jobs:
             pytest-extra-options: "-W ignore:Matrix:scipy.linalg._misc.LinAlgWarning"
 
           - case-name: Windows
-            os: windows-latest
+            os: windows-11-arm
             python-version: "3.14"
             numpy-build: ">=2.0.0"
             numpy-requirement: ">=2.0.0"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -145,6 +145,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: conda-incubator/setup-miniconda@v3
+        if: matrix.os != 'windows-11-arm'
+        # Currently miniconda doesn't support window-11 arms
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
@@ -218,8 +220,13 @@ jobs:
           fi
 
       - name: Package information
+        shell: bash
         run: |
-          conda list
+          if [[ "${{ matrix.os }}" == "windows-11-arm" ]]; then
+            pip list
+          else
+            conda list
+          fi
           python -c "import qutip; qutip.about()"
           python -c "import qutip; print(qutip.settings)"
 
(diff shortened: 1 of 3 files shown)


Mayank447:
@Ericgig @pmenczel This PR is ready for review. The failing test cases due to stochastic solvers are unrelated to this PR. For the arm version everything passes.

As a side note ideally we should pin the random tests in the solvers to a seed.

Ericgig:
I tried to run the build wheel action and the windows arm wheels does not build...
It seems to be trying to compile numpy from sources:
```
...
[139/324] Compiling C object numpy/_core/_multiarray_umath.cp310-win_arm64.pyd.p/meson-generated_einsum_sumprod.c.obj
[140/324] Compiling C object numpy/_core/libloops_autovec.dispatch.h_baseline.a.p/meson-generated_loops_autovec.dispatch.c.obj
[141/324] Compiling C object numpy/_core/_multiarray_umath.cp310-win_arm64.pyd.p/meson-generated_lowlevel_strided_loops.c.obj
      ninja: build stopped: subcommand failed.
      INFO: autodetecting backend as ninja
      INFO: calculating backend command to run: C:\Tools\Ninja\ninja.EXE
      [end of output]
```

Mayank447:
Upon looking into it, the issue seems to be that there is no official python installer for 3.10 for windows arm. From 3.11 there is. Numpy wheels for windows arm are for python 3.11+. That's why it was trying to compile numpy from source and failing.

@Ericgig should we bump up the minimum python version or skip wheels build for windows arm for 3.10

Ericgig:
For now, you can remove 3.10 from the build wheels, but leave it in the test and requirement, etc.
3.10 is old enough that we can remove it's support according to our rules, but I will discuss it with other qutip maintainers before making a call.

Mayank447:
@Ericgig Done. The wheels for windows arm should build now.

Also could you rerun the test for build-documentation. Although I didn't touch the yaml file for it, the CI run appears to be broken.

Ericgig:
Every wheels build fine.
Thank you.

Mayank447:
@Ericgig I had one doubt. Is there a reason why we are not building wheels for musllinux (Alpine Linux)? Numpy, scipy and cython provide wheels for `musllinux` and `qutip-qoc` seems to be doing that too.

Edit: Also adding `abi3` as fallback for future python releases might be beneficial, currently we are compiling python version-specific wheels

BoxiLi:
qutip-qoc does that just because we were lazy and didn't add any filters. Let's not use it as a standard here. Eric might be able to answer the concrete question.

Ericgig:
Sorry, all I can say is it was like that before I joined the team (as far as I remember).

For a long time, conda was the only place to get qutip binaries for windows, there was a complain/ issue and now it work. I guess we are just waiting for a musllinux user to ask for it.

On conda forge, we also support ppc, but there are no musllinux tags.

.github/workflows/tests.yml
```diff
@@ -134,6 +145,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: conda-incubator/setup-miniconda@v3
+        if: matrix.os != 'windows-11-arm'
+        # Currently miniconda doesn't support window-11 arms
```
Ericgig: By skipping this step, windows-11-arm does not setup the desired python, it uses 3.13 instead of 3.14.

Mayank447: Good catch, I have added `setup-python` CI for `windows-11-arm`. I have verified the same, this resolved the issue.

Ericgig (approved):
Thank you.

The ReadtheDocs failure is fixed in the master branch.
The 3.10 failure is a new one... not related to changes here.

PR #2859, opened 2026-04-06
OhmicEnvironment without mpmath

Ericgig:
**Description**
`mpmath` is only needed for one function `zeta` that is not available for complex number in scipy. An efficient equation to implement it is available on [wikipedia](https://en.wikipedia.org/wiki/Euler%E2%80%93Maclaurin_formula#Examples).

This implementation is not as robust as mpmath's when s is smaller than 1, but works well in the range needed for OhmicEnvironment.

### DIFF
--- a/doc/changes/2859.misc
+++ b/doc/changes/2859.misc
@@ -1,0 +1,1 @@
+Add fallback for missing mpmath module for OhmicEnvironment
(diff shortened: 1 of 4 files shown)


pmenczel:
Just a note that zeta function with complex argument is also under development in scipy, see https://github.com/scipy/scipy/issues/14073

However, it will take a while until we can use that, so I think this would be a good addition that allows us to remove mpmath dependency now.

Ericgig:
> Just a note that zeta function with complex argument is also under development in scipy, see https://github.com/scipy/scipy/issues/14073Just a note that zeta function with complex argument is also under development in scipy, see https://github.com/scipy/scipy/issues/14073

Presently it's only for riemann zeta (q=0). Support for what we need seems far away.

hodgestar (approved):

PR #2862, opened 2026-04-08
Add doc notes to sigmam and sigmap

### LINKED ISSUE #2494: Definition of CNOT
### Bug Description

The act of the CNOT gate is not what it is expected. In particular it changes the target qubit when the control qubit is in the ground state instead of excited state given that in qutip basis(2,1) is the ground state. If this is on purpose it needs to be clearly stated in the tutorials or the documentation

### Code to Reproduce the Bug

```shell
from qutip import *
from qutip_qip.circuit import QubitCircuit
from qutip_qip.operations import (Gate, berkeley, cnot, cphase, csign, fredkin,
                                  gate_sequence_product, globalphase, iswap,
                                  molmer_sorensen, phasegate, qrot, rx, ry, rz,
                                  snot, sqrtiswap, sqrtnot, sqrtswap, swap,
                                  swapalpha, toffoli)
test=cnot()*tensor(basis(2,1),basis(2,1))
print(test)
```

### Code Output

```shell
Quantum object: dims=[[2, 2], [1, 1]], shape=(4, 1), type='ket', dtype=Dense
Qobj data =
[[0.]
 [0.]
 [1.]
 [0.]]
```

### Expected Behaviour

Quantum object: dims=[[2, 2], [1, 1]], shape=(4, 1), type='ket', dtype=Dense
Qobj data =
[[0.]
 [0.]
 [0.]
 [1.]]

### Your Environment

```shell
QuTiP Version:      5.0.2
Numpy Version:      1.26.4
Scipy Version:      1.11.4
Cython Version:     None
Matplotlib Version: 3.8.0
Python Version:     3.11.0
Number of CPUs:     8
BLAS Info:          Generic
INTEL MKL Ext:      False
Installation path:  /opt/anaconda3/lib/python3.11/site-packages/qutip
```

### Additional Context

_No response_

BoxiLi:

**Description**
Add doc notes for sigmax and sigmap operators.

**Related issues or PRs**
Fix #2494

### DIFF
--- a/doc/changes/2862.doc
+++ b/doc/changes/2862.doc
@@ -1,0 +1,1 @@
+Add doc notes to sigmam and sigmap
--- a/qutip/core/operators.py
+++ b/qutip/core/operators.py
@@ -357,6 +357,11 @@ def sigmap(*, dtype: LayerType = None) -> Qobj:
         Storage representation. Any data-layer known to ``qutip.data.to`` is
         accepted.
 
+    Notes
+    -----
+    Equivalent to :func:`.create` when acting on a qubit defined by
+    :func:`.basis`.
+
     Examples
     --------
     >>> sigmap() # doctest: +SKIP
@@ -379,6 +384,11 @@ def sigmam(*, dtype: LayerType = None) -> Qobj:
     dtype : type or str, optional
         Storage representation. Any data-layer known to ``qutip.data.to`` is
         accepted.
+
+    Notes
+    -----
+    Equivalent to :func:`.destroy` when acting on a qubit defined by
+    :func:`.basis`.
 
     Examples
     --------


pmenczel:
I would suggest adding a note like this to all sigmax, sigmay, sigmaz, sigmam, sigmap, and perhaps being a bit more detailed. (I think it is a very common point of confusion when learning qutip.) Perhaps along the following lines?

> In QuTiP, usually `basis(N, 0)` represents the ground state. However, for historical reasons, the Pauli operators in QuTiP are defined such that `basis(2, 1)` is the ground state and `basis(2, 0)` the excited state. That is, `create(2)` is equivalent to `sigmam()` and `destroy(2)` is equivalent to `sigmap()`.

BoxiLi:
But for sigmax, sigmay, and sigmaz, those conventions are clear. Also, we have the matrix there explicitly. The only confusion comes from sigma_mines (plus), which is not really lowering/rising the qubit state (but spin!). And this is also not wrong, the pauli matrices are defined like this everywhere.

That is why I find adding this to all those Pauli matrices unessenary but maybe we should add the sentence you suggested to https://qutip.readthedocs.io/en/qutip-5.2.x/guide/guide-basics.html#states-and-operators

Also say that `basis(2,0) is spin up in the convention and basis(2,1) spin down.`

Ericgig:
I agree with adding the explanation to the guide.

Also, we could try to push the function `qutip.ket` more.
`qutip.ket("uud...")` give the expected form for up and down qubits, but I don't think we use it a lot in the examples...

paultutorialdemo:
> But for sigmax, sigmay, and sigmaz, those conventions are clear

I disagree with this... I recently had a bug, essentially because I expected `sigmaz() @ basis(2, 0) == -basis(2, 0)`. Oops...

However, I would be okay with having an explanation not on all the Pauli matrices, but for example in the docstring of `basis`.

Also adding something to the guide, and pushing `ket` more, both sound good to me.

(Edit: sorry this is @pmenczel on the wrong account)

PR #2864, opened 2026-04-09
HEOM Solver Refactor with Direct Sum

pmenczel:
The direct sum module was just added to qutip (#2833). This PR updates the HEOM solver to make use of this functionality in order to build the generator. This has two main advantages: (1) coupling operators can now be time-dependent (`QobjEvo`), and (2) the HEOM generator can now have any data type in principle. (The block operation functions must be implemented in the data layer for the dtype. Usually it won't make sense to use Dense or Dia, but this is perhaps useful for plugins such as qutip-jax or qutip-cuquantum.)

In my tests with CSR, the performance seems roughly the same as before.
Edit: I ran into a case where the rhs construction got slower by >300%. I am looking into it.


qutip/solver/heom/bofin_solvers.py
```diff
@@ -649,48 +682,63 @@ def __init__(self, H, bath, max_depth, *, odd_parity=False, options=None):
...
             liouvillian(H) if H.type == "oper"  # hamiltonian
             else H  # already a liouvillian
         )
+        if isinstance(self.L_sys, QobjEvo):
+            self.L_sys.compress(_skip_coeff=True)
 
-        self._sys_shape = int(np.sqrt(self.L_sys.shape[0]))
-        self._sup_shape = self.L_sys.shape[0]
+        # Dimensions that L_sys acts on
+        self._sys_dims_ = self.L_sys._dims[0].oper
```
hodgestar, line 695:
```suggestion
        self._sys_dims = self.L_sys._dims[0].oper
```
If this already exists without the underscore we need to choose a better name. :)

Issue #2414, opened 2024-05-01
Check matplotlib requirement


Ericgig:
### Describe the Issue!

We still have matplotlib minimum version requirement at 1.2.1.
We added animations functions, updated the bloch sphere etc. since that version was set and it's probably no longer sufficient.

Prtm2110:
I believe we can set the minimum required version of Matplotlib to 2.0.0, as it includes all the necessary animation features qutip is currently using. I would like to contribute to resolving this issue.

Mayank447:
@Ericgig This issue can be closed as the minimum matplotlib version has been bumped to 3.6 now

PR #2875, opened 2026-04-14
Increase minimum python version to 3.11

Ericgig:
**Description**
Python 3.10 is getting old and we can't build arm wheel for it.
Using py3.11 also allows us to use ExceptionGroup and some new type hints feature.

Updating wheel to build post 3.10 only is in #2855


.github/workflows/tests.yml
```diff
@@ -64,20 +64,9 @@ jobs:
             oldmpl: 1
             coveralls: 1
 
-          - case-name: p311, numpy fallback, no mkl
```
hodgestar, line 67: Do we check the `no mkl` and `numpy fallback` cases in another case?

pyproject.toml
Mayank447, line 61:
```suggestion
graphics = ["matplotlib>=3.6"]
```

matplotlib 3.5 only supported Python 3.7-3.10, I tried locally and am unable to install it on Python 3.11 on mac. The same is true for windows https://github.com/matplotlib/matplotlib/issues/28671

pyproject.toml
```diff
@@ -64,8 +64,7 @@ semidefinite = ["cvxpy>=1.0", "cvxopt"]
 tests = ["pytest>=5.2", "pytest-rerunfailures"]
```
Mayank447, line 64:
```suggestion
tests = ["pytest>=7.2", "pytest-rerunfailures"]
```

Same reason as above.

pyproject.toml
Mayank447, line 62:
```suggestion
runtime_compilation = ["cython>=0.29.34", "filelock", "setuptools"]
```

First Cython to support Python 3.11 is `0.29.34`, cython `0.29.20` supported python 2.6 to 3.8 as per PyPI

Mayank447, line 63:
```suggestion
semidefinite = ["cvxpy>=1.3", "cvxopt"]
```

Mayank447, line 47:
```suggestion
    "numpy>=1.23.2",
```

Numpy 1.22 only has wheels for Python <= 3.10

Mayank447, line 48:
```suggestion
    "scipy>=1.9.2,!=1.16.0,!=1.17.0",
```

Same reason as above

Mayank447 (changes requested):
@Ericgig I was trying to fix the matplotlib version issue in #2414. The one line change for the same can be included in this PR. Also there were some related dependency version update I found

pyproject.toml
Ericgig: Thank you for looking into that.

.github/workflows/tests.yml
Ericgig: Yes, we still have no mkl and numpy fallback for each OS after removing this one.

Mayank447 (approved):
LGTM

#2414 can also be closed when this PR is merged. All the animation functions seem to work.

hodgestar (approved):
Thanks, @Ericgig.

veronikakurth:
This PR has implications for other packages of the QuTiP family like `qutip-qoc`. This is how tests fail on `main` there at the moment (the branch is the copy of main): https://github.com/veronikakurth/qutip-qoc/actions/runs/25781447343/job/75724835621
Tests with Python 3.10 have to be removed there then, I will do it in `qutip-qoc`.
Other repositories should be checked for this as well.

Mayank447:
> This PR has implications for other packages of the QuTiP family like `qutip-qoc`.

I think it might be better to update the minimum Python version requirements for QuTiP family packages as Python 3.10 EOL is during fall this year.

PR #2881, opened 2026-04-15
fix: normalize weights in _minimum_roundoff_ensemble for non-normalized mixed states

### LINKED ISSUE #2880: mcsolve fails with non-normalized initial mixed state
### Bug Description

### Code to Reproduce the Bug

```python
import qutip as qt
import numpy as np
```

### Code Output

```shell
```

### Expected Behaviour

### Your Environment

```shell
```

### Additional Context

xyaz1313:
## Description

Fixes #2880

When a mixed state is provided as the initial condition for , the  algorithm allocates trajectories proportional to each state's weight. However, it assumes the weights sum to exactly 1.0.

If the density matrix is not perfectly normalized (e.g.,  due to numerical error), the algorithm allocates too few trajectories:

- **Before fix**:  → only 8 trajectories allocated, later causes `IndexError: State id 8 must be smaller than number of trajectories 8`
- **After fix**: weights normalized to  → , then trimmed to 10 by the existing while loop

## Fix

Added weight normalization after filtering zero-weight states in :

```python
total_weight = sum(w for _, w in filtered_states)
if total_weight > 0 and abs(total_weight - 1.0) > 1e-10:
    filtered_states = [(idx, w / total_weight)
                       for idx, w in filtered_states]
```

This ensures the allocation algorithm always works with weights summing to 1.0, regardless of numerical precision in the input density matrix.

## Reproduction (from issue)

```python
import qutip as qt
import numpy as np
```

Previously crashed with `IndexError`. Now runs 10 trajectories correctly.

## Testing

- Verified the fix logic mathematically for the reported case
- The normalization is a no-op when weights already sum to 1.0 (common case)
- Existing while-loop trimming handles over-allocation from normalization

### DIFF
--- a/qutip/solver/multitraj.py
+++ b/qutip/solver/multitraj.py
@@ -448,6 +448,14 @@ class _InitialConditions:
         filtered_states = [(index, weight)
                            for index, (_, weight) in enumerate(state_list)
                            if weight > 0]
+
+        # Normalize weights so they sum to 1, preventing too few trajectories
+        # when the input density matrix is not perfectly normalized.
+        total_weight = sum(w for _, w in filtered_states)
+        if total_weight > 0 and abs(total_weight - 1.0) > 1e-10:
+            filtered_states = [(idx, w / total_weight)
+                               for idx, w in filtered_states]
+
         if len(filtered_states) > ntraj_total:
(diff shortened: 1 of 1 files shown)


pmenczel:
Why did you open 2 PRs?

PR #2884, opened 2026-04-16
Fix mcsolve crash with non-normalized mixed initial state

### LINKED ISSUE #2880: mcsolve fails with non-normalized initial mixed state

Ai-chan-0411:
## Summary

When `mcsolve` receives a density matrix whose eigenvalues do not sum to exactly 1.0 (e.g. `0.5 * fock_dm(2, 0) + 0.25 * fock_dm(2, 1)` has trace 0.75), the `_minimum_roundoff_ensemble` method in `_InitialConditions` allocates fewer trajectories than requested. This happens because the ceil-based initial guesses sum to less than `ntraj_total`, so the trimming loop never runs and the total trajectory count ends up too low. Downstream, this causes an `IndexError` when trajectory IDs exceed the allocated count.

The fix normalizes the weights within `_minimum_roundoff_ensemble` so the distribution algorithm always assigns exactly `ntraj_total` trajectories. A tolerance of `1e-12` avoids unnecessary normalization when weights are already close to 1. The original un-normalized weights are preserved in `state_list` and continue to be used by `get_state_and_weight` for the correction factor, so the physical meaning of the result is unaffected.

Includes a regression test that reproduces the exact scenario from the issue report.

Closes #2880

## Changes

- `qutip/solver/multitraj.py`: Normalize filtered weights in `_minimum_roundoff_ensemble` before trajectory allocation
- `qutip/tests/solver/test_mcsolve.py`: Add `test_mcsolve_non_normalized_mixed_state` regression test

### DIFF
--- a/qutip/solver/multitraj.py
+++ b/qutip/solver/multitraj.py
@@ -444,10 +444,18 @@ class _InitialConditions:
         Calculate a list ntraj from the given total number, under contraints
         explained above. Algorithm based on https://stackoverflow.com/a/792490
         """
-        # First we throw out zero-weight states
+        # First we throw out zero-weight states and normalize the
+        # remaining weights so they sum to one. This is necessary because
+        # the input density matrix may not be perfectly normalized due to
+        # numerical error. Without normalization, the algorithm can assign
+        # too few trajectories, leading to an IndexError later.
         filtered_states = [(index, weight)
                            for index, (_, weight) in enumerate(state_list)
                            if weight > 0]
+        total_weight = sum(weight for _, weight in filtered_states)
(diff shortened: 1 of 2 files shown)


veronikakurth:
The reason for closing this PR can be understood from [our stance on good-first-issue contributions](https://github.com/qutip/qutip/issues/2880#issuecomment-4257481718).

Issue #1537, opened 2020-07-31
Bug  in function spectrum_correlation_fft or wrong documentation


denisvion:
Is'nt there a bug in spectrum_correlation_fft ?
The documentation of the function does not specifiy whether one should use symetrized correlations functions at both positive and negative times or only positive times.
But the documentation gives an example with positive times only.
The code below show that in this case, the spectrum of an exponentially decaying correlation function is not Lorentzian as it should be.
-----------------------------------
@author: denis vion
"""
from numpy import *
from qutip import*
from matplotlib.pyplot import *

taus=linspace(0, 50, 5000)
g1s=exp(-taus)
figure()
ax = subplot(1,1,1)
ax.plot(taus, g1s)
ax.set_xlim([0,5])
ax.set_ylim([0, 1])
ws, spec = spectrum_correlation_fft(taus, g1s) # spectrum is not lorentzian!
ws2, spec2 = spectrum_correlation_fft(hstack([-taus[:0:-1], taus]),hstack([g1s,g1s[-2::-1]])) # dirty fix outside 
figure()
ax = subplot(1,1,1)
ax.semilogy(ws, spec,'r')
ax.semilogy(ws2 ,spec2,'b')
ax.set_title('spectrum_correlation_fft(correlations at positive times) is not Lorentzian !')
ax.set_xlim([-100,100])
ax.set_ylim([0.0001,4])
-------------------------------------
![BugSpectrum](https://user-images.githubusercontent.com/12784991/89054759-a3b01f00-d359-11ea-974c-cebc7f763cbb.png)

nwlambert:
Looking again at spectrum_correlation_fft(), given that we seem to be implicitly assuming c(t) is always symmetric in time (or more generally, c(t) = c*(-t)), perhaps we can update this to use the full two-sided fft, like we do in https://github.com/qutip/qutip/blob/5a8ea7ffa68525f664c1b0d4b338691cf7f8f8c4/qutip/core/environment.py#L2167

if no objections we could make this a good first issue

pmenczel:
To summarize the discussion: currently, `spectrum_correlation_fft` currently internally uses a one-sided fft. The two-sided one, which is already implemented in qutip in the environment module (linked by @nwlambert) converges better, so we would like to change `spectrum_correlation_fft`.

matti-sarjala:
A small way to make the change easier to review might be to separate the **numerical implementation** from the **user-facing input convention**.

Possible minimal acceptance criteria:

[...]

PR #2887, opened 2026-04-18
Clarify spectrum_correlation_fft documentation

mitre88:
Updates the docstring to explain that for physically accurate spectra, the correlation function should include both positive and negative time delays, and that the FFT assumes periodicity which can cause artifacts when only positive times are provided.

Addresses #1537.

### DIFF
--- a/qutip/solver/spectrum.py
+++ b/qutip/solver/spectrum.py
@@ -64,6 +64,15 @@ def spectrum_correlation_fft(tlist, y, inverse=False):
     Calculate the power spectrum corresponding to a two-time correlation
     function using FFT.
 
+    This function computes the discrete Fourier transform of the correlation
+    data. For a physically accurate spectrum, the correlation function should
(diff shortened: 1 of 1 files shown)


nwlambert:
Thanks @mitre88!  I think this only fixes half the problem though, and introduces some logical problems....

If users use the current function with time steps [-T,..., T], the function itself assumes these are a one sided tlist so still multiplies the final spectrum result by '2', which will give an incorrect overall magnitude to the result.

I feel like we have a few options, not sure which is best

- stick with only accepting as input positive times [0,...T], internally extend the tlist to [-T,...T], and do the full FFT (without taking the final 2*Real[] part). 
- Stick with the current logic, and add an option for [-T,....,.T], with a options arg or something.
- Check for the range of the tlist and choose which option to take.

I feel like the first option is the cleanest, and keeps backwards compatibility?  not sure if there is a reason not to do it that way...

PR #2896, opened 2026-05-12
Bug.pip in test

Ericgig:
**Description**
Use conda to install pip in case it's not available with setup-miniconda by default.

Also increased the oldest matplotlib version to match the one we officially support now that we don't support python 3.10.


hodgestar (approved):

veronikakurth:
Thank you very much for the fix!
I'm still trying to understand how this error came into place. Based on the solution that works, can we conclude that miniconda currently doesn't support the latest macOS?

PR #2901, opened 2026-05-14
Clarify the to_chi Pauli convention

mzl2233:
This clarifies the Pauli-basis convention used by to_chi, including the ordering, column-stacking detail, and the unnormalised scaling of the returned chi matrix. It also adds a small sigmay example that shows the expected Y,Y entry so the sign convention is visible in the user guide.\n\nVerification: ran git diff --check and a lightweight documentation-content check. A local editable install was attempted but could not complete in this environment because the project metadata is rejected by the installed setuptools version; importing from the checkout also cannot load the unbuilt Cython extensions.


veronikakurth:
One of the numerous spam activity PRs in this and other open-source projects. Closing.

PR #2902, opened 2026-05-14
Fix Pauli Y sign in to_chi basis

mzl2233:
This fixes the Pauli-Y sign convention used when building the super-Pauli basis for chi and Pauli-transfer conversions. The basis now keeps the usual Y orientation rather than its complex conjugate, and a focused regression test checks that the Y column matches the vectorized Pauli-Y operator. Verification was attempted with `python3 -m pytest qutip/tests/core/test_superop_reps.py -q -k positive_y_column`; it currently cannot run in this checkout because the local C extension module `qutip.core.data._local_matmul` is not built, and `python3 setup.py build_ext --inplace` is blocked by the installed setuptools validation of the repository's existing license metadata.

### DIFF
--- a/qutip/core/superop_reps.py
+++ b/qutip/core/superop_reps.py
@@ -33,7 +33,7 @@ from ..typing import DimensionLike
 _SINGLE_QUBIT_PAULI_BASIS = (
     identity(2).to(_data.CSR),
     sigmax().to(_data.CSR),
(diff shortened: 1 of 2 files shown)


veronikakurth:
Purely AI-generated contribution. No AI usage statement provided. Closing.

PR #2903, opened 2026-05-14
Cover the Pauli Y basis sign

adminlip:
This adds regression coverage for the Pauli Y basis used by the superoperator representation helpers. The test checks that the generated super-Pauli basis keeps the conventional positive Y column instead of accepting a conjugated sign, which documents the behavior discussed in the issue.


veronikakurth:
Purely AI-generated contribution. Duplicate of #2902. Closing.

PR #2890, opened 2026-04-22
Improve test tolerance to lower random failures rate

Ericgig:
**Description**
Improve test tolerances to reduce random failures in tests:
- `test_satisfy_eigenvalue_equation`: The sparse solver had trouble converging on degenerate `0` Eigenvalues.
  The error could happen with `scipy.sparse.linalg.eigh` directly, with it not finding the desired eigenvalues or error around 1e-6 instead of machine precision. 
  Avoiding `0` eigenvalues avoid the issue.
- `test_run_from_experiment_open`: decrease the step length make it stable enough for the tests. The error is of order `dt` and can accumulate quickly.
- `test_fsesolve_fallback`: Not fixed, unable to reproduce the error...
- `test_zeta`: I tested it more than in the original PR and 1e-14 tolerance was too tight. I loosened the tolerance and also keep input farther from hard to converge cases.

fix #2876

PR #2913, opened 2026-05-27
Improve test tolerance to lower random failures rate [part 2]

### LINKED ISSUE #2876: Fix tests for eigenstates, stochastic and floquet solvers
### Bug Description

During the last month, the automated tests ran ~500 times and we had 6 random failures:

`qutip/tests/core/test_eigenstates.py::test_satisfy_eigenvalue_equation[degenerate-dense-True]` - failed 3 times
`qutip/tests/solver/test_stochastic.py::test_run_from_experiment_open[True-euler]` - failed twice
`qutip/tests/solver/test_floquet.py::test_fsesolve_fallback` - failed once

We should fix them. _Possible solution_: "loosen the tolerances or change the input generation (size, density, etc.) until the failure rate is acceptable, without making it so tolerant that anything would pass"

### Code to Reproduce the Bug

No related code to reproduce it - due to random nature of tests, the above-mentioned errors can be observed _occasionally_ in CI/CD tests and is not related to code changes of a PR.

### Code Output

From _[ubuntu-latest, python3.11, p311, numpy fallback, no mkl]_ and _[ubuntu-latest, python3.12, p312, numpy fallback]_:

```shell
FAILED qutip/tests/solver/test_stochastic.py::test_run_from_experiment_open[False-euler] - RuntimeWarning: overflow encountered in square
```

From _[ubuntu-latest, python3.14, python 3.14 - Latest]_:

```shell
FAILED qutip/tests/solver/test_floquet.py::TestFloquet::testFloquetUnitary - assert (0.9999981730...34135176e-05j) == 1.0 ± 5.0e-05
  comparison failed
  Obtained: (0.9999981730805094+5.2031557134135176e-05j)
  Expected: 1.0 ± 5.0e-05
FAILED qutip/tests/solver/test_stochastic.py::test_run_from_experiment_open[False-euler] - RuntimeWarning: overflow encountered in square
```

### Expected Behaviour

Tests don't fail anymore

### Your Environment

```shell
Dependencies installed in CI/CD jobs w.r.t. to platform and Python version
```

### Additional Context

Based on the discussion in #1268

veronikakurth:
**Description**
This is the continuation of addressing #2876, started by @Ericgig  in #2890. 

Upon repeating `test_fsesolve_fallback` > 2000-3000 times, I managed to reproduce the error occasionally occurring in the test workflows.
(`pytest qutip/tests/solver/test_floquet.py::test_fsesolve_fallback --count=2000`)

The only source of randomness in the test is in the initial ket generation, so I had a look at an example of its value upon the test's failure:

```
Quantum object: dims=[[2], [1]], shape=(2, 1), type='ket', dtype=Dense
Qobj data =
[[ 0.43238738+0.90113114j]
 [-0.02923969-0.01220112j]]
```
As we see, the second value is very small in magnitude. 

When multiplying matrices to get inner products in [to_floquet_basis](https://github.com/qutip/qutip/blob/master/qutip/solver/floquet.py#L280), such tiny values may lead to cancelling out larger complex values.

Knowing we want to keep randomness in tests, I'd suggest constraining it a bit to make the test stabler. We could introduce a "minimum weight" for the coefficients of a vector, for example. 

I experimented with minimum weight values and found out the following:
`1e-4` makes it a bit too loose -> for ~5000 tests, there are 1-2 failed tests. Without constraining weights, there are 2-3 failed tests per 5000. There is no significant difference between having a guard on weights and not having one. 

Going much higher in the number of trials doesn't increase the average number of failed tests significantly, though. 
**Proposal**
Based on the observed failure rate when doing local repeated runs, I believe `1e-2` should do well as a weight constraint (with the latter, I didn't get any error per 10K test runs). With this, we still cover a significant amount of initial kets, but make errors even more unlikely than they are now. 

**Related issues or PRs**
Fix #2876

**AI Tools Usage Disclosure**
Claude Sonnet 4.7 was used to analyze the Floquet solver codebase and find an exact place where catastrophic cancellation may occur if the initial density matrix/ket contains too small values.

### DIFF
--- a/qutip/tests/solver/test_floquet.py
+++ b/qutip/tests/solver/test_floquet.py
@@ -334,16 +334,9 @@ class TestFloquet:
             assert (min(abs(Xs - Xpm_m1)) < 1e-4)
             idx += 1
 
-def safe_rand_ket(N, min_weight=1e-2):
-    while True:
-        psi = rand_ket(N)
-        weights = np.abs(psi.full().flatten())**2
-        if np.min(weights) > min_weight:
-            return psi
-
 def test_fsesolve_fallback():
     H = [sigmaz(), lambda t: np.sin(t * 2 * np.pi)]
-    psi0 = safe_rand_ket(2)
+    psi0 = rand_ket(2)
     ffstate = fmmesolve(H, psi0, [0, 1], T=1.).final_state
-    fstate = sesolve(H, psi0, [0, 1]).final_state
+    fstate = sesolve(H, psi0, [0, 1], options={"atol": 1e-9}).final_state
     assert (ffstate - fstate).norm() < 1e-5
--- a/qutip/tests/solver/test_floquet.py
+++ b/qutip/tests/solver/test_floquet.py
@@ -334,10 +334,16 @@ class TestFloquet:
             assert (min(abs(Xs - Xpm_m1)) < 1e-4)
             idx += 1
 
+def safe_rand_ket(N, min_weight=1e-4):
+    while True:
+        psi = rand_ket(N)
+        weights = np.abs(psi.full().flatten())**2
+        if np.min(weights) > min_weight:
+            return psi
 
 def test_fsesolve_fallback():
     H = [sigmaz(), lambda t: np.sin(t * 2 * np.pi)]
-    psi0 = rand_ket(2)
+    psi0 = safe_rand_ket(2)
     ffstate = fmmesolve(H, psi0, [0, 1], T=1.).final_state
     fstate = sesolve(H, psi0, [0, 1]).final_state
     assert (ffstate - fstate).norm() < 1e-5
--- a/qutip/tests/solver/test_floquet.py
+++ b/qutip/tests/solver/test_floquet.py
@@ -334,7 +334,7 @@ class TestFloquet:
             assert (min(abs(Xs - Xpm_m1)) < 1e-4)
             idx += 1
 
-def safe_rand_ket(N, min_weight=1e-4):
+def safe_rand_ket(N, min_weight=1e-2):
     while True:
         psi = rand_ket(N)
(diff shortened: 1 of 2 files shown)


qutip/tests/solver/test_floquet.py
```diff
@@ -334,10 +334,16 @@ def noise_spectrum(omega):
             assert (min(abs(Xs - Xpm_m1)) < 1e-4)
             idx += 1
 
+def safe_rand_ket(N, min_weight=1e-2):
+    while True:
```
Ericgig: Avoid `while True` without safeguard for infinite loops.
Presently it's fine since the default for `rand_ket`'s density is 100% filled, but if that default where to be changed, it could never end.

Ericgig (commented):
Nice find!

I would suggest to use options to increase accuracy of the solver instead of avoiding some states. With the example above, `fmesolve` is 2e-6 above analytical, `sesolve` 8.1e-6 under.  So tightening the tolerance in sesolve's options a little would be enough.

veronikakurth:
@Ericgig Thank you for your analysis of the problem - I agree that the approach you suggest is better than filtering out the inputs. It seems that adjusting `atol` for `sesolve` by one order of magnitude helped - I ran tests ~10000 times, and there was not a single failure on non-safeguarded inputs. :)

qutip/tests/solver/test_floquet.py
veronikakurth: Thanks for the tip!

veronikakurth:
@Ericgig P.S. I'm aware of the problem with `qutip-qoc` in the "family projects" pipeline and am trying to figure out the sudden failure to discover the matplotlib dependency there.

Ericgig (approved):
Thank you.

Issue #2819, opened 2026-02-10
Repeated calls to mesolve in a loop cause increasing memory usage


M0rtamor:
## Description

I want to find a steadystate by evolving an initial state for some time chunck, checking if its converged and if not, continue evolving in a loop.

When doing this, the memory usage increases after each iteration. When using gc.collect() at the end, this ram usage stays constant, however using del on every variable does not. To me it seems like it is caused by mesolve() not freeing some amount of memory after finishing.

## Environment

QuTiP version: 5.2.2
Python version: Python 3.13.3
OS: Windows 11
SciPy version: 1.16.3
NumPy version: 2.3.5

## Minimal Reproducible Example

```python
import qutip as qt
import numpy as np
import psutil
import gc

def check_convergence(val1: float, val2: float, atol: float = 1e-6, rtol: float = 1e-4):
    if np.abs(val2 - val1) < atol + rtol * np.abs(val2):
        return True

    return False

process = psutil.Process()
w0 = 1
wd = 1
kappa = 0.1
alpha = 8
nph = 120
a = qt.destroy(nph)

c_ops = [np.sqrt(kappa)*a]
e_ops = [a]

H = w0 * a.dag() * a + wd * (a.dag() + a)
L = qt.liouvillian(H, c_ops=c_ops)
rho0 = qt.coherent_dm(nph, alpha)

dt = 4
chunk_index = 0
chunk_size = 5
expect = []

memory = []

while True:
    tlist = np.arange(chunk_index*chunk_size*dt, (chunk_index+1)*chunk_size*dt + 1, dt)
    result = qt.mesolve(L, rho0, tlist, e_ops=e_ops, options={"store_final_state": True, "store_states": False})
    for val in result.expect[0][:-1]:
        expect.append(val)

    rho0 = result.final_state
    chunk_index += 1

    memory.append(process.memory_info().rss / (1024**2))
    if check_convergence(np.abs(result.expect[0][-2]), np.abs(result.expect[0][-1])):
        expect.append(result.expect[0][-1])
        break
    
    # gc.collect()

print("used memory after each iteration: ", memory)
```

Output:

```
used memory after each iteration:  [117.92578125, 121.45703125, 124.8203125, 128.12109375, 131.26953125, 134.609375, 137.93359375, 141.25, 144.609375, 147.9609375]
```

## Fix:

Output with gc.collect() uncommented:

```
[...]

PR #2916, opened 2026-06-11
Add `Integrator` entry point and developper documentation

Ericgig:
**Description**
Change integrator to accept other format that QobjEvo.
Remove the  `support_time_dependant`, `supports_blackbox` flags for a `RHS_format` value.
This `RHS_format` indicate the format of the RHS: arbitrary function, constant matrix or the solver itself.

Now most integrator can take any callable with the right signature as the derivative. `krylov`, `diag` take a matrix directly.

I also added an internal documentation section for the integrators.

- Made change to the memory management: avoided integrator to keep a one of it's method as an attribute. This create a reference loop and stop easy de-allocation. Should help with #2819, but does not fully solve the issue.

- Added types hint for public facing Integrator method.

- Removed the `method` attribute, it's explicitly passed in `add_integrator` so redundant.

This is the first of a set of PR adding internal documentation for the solvers.

**AI Tools Usage Disclosure**
Gemini: review + Rephrased the documentation `.rst`.

### DIFF
--- a/doc/internals/solver/index.rst
+++ b/doc/internals/solver/index.rst
@@ -1,0 +1,68 @@
+.. _solver_internals:
+
+Quantum Solvers and Integrators Architecture
+############################################
+
+The solver framework in QuTiP is responsible for simulating the time-evolution of quantum systems.
+Built directly on top of the core data layer,
+the architecture enforces a separation of concerns between physical equations and numerical solvers.
+
+This document serves as an overview of the internal architecture of QuTiP's solvers,
+guiding core developers and contributors on how the system is structured,
+how data flows through it, and how to extend it.
+
+
+Core Architecture and Separation of Concerns
+============================================
+
+The framework decouples physical models from their numerical integration backends, dividing
+tasks between two primary base classes:
+
+1. **The Physics Layer (:class:`Solver`)**: This layer represents the physical equation
+   governing the system's dynamics. Each distinct quantum equation—such as the Schrödinger
+   equation, Lindblad Master Equation, or Bloch-Redfield equations—maps to a unique
+   subclass inheriting from the abstract base :class:`Solver` class (e.g., :class:`SESolver`,
+   :class:`MESolver`, :class:`BRSolver`). The solver class is responsible for setting up
+   the system operators and defining the physical derivative.
+
+2. **The Numerical Layer (:class:`Integrator`)**: This layer handles the low-level
+   numerical implementation of Ordinary Differential Equations (ODEs) or Stochastic
+   Differential Equations (SDEs). The solver communicates the physical state and derivative
+   rules down to the :class:`Integrator`. The selection of which numerical scheme to run
+   is entirely dictated by the solver's configuration options through the ``"method"`` parameter.
+
+Generally, integrators are general-purpose ODE solvers. However, to maximize performance,
+certain integrators are specialized to specific physical representations,
+where the boundary between physical assumptions and numerical routines is intentionally blended.
+
+
+User-Facing Interfaces vs. Internal Infrastructure
+==================================================
+
+To maintain an intuitive user experience, QuTiP hides internal architectural complexity
+behind high-level functional wrappers:
+
+- Functions such as :func:`sesolve` or :func:`mesolve` serve as convenience entry points for the end user.
+  In a single call, these helper functions instantiate the appropriate underlying :class:`Solver` class,
+  run the evolution, and return a completed results container.
+- Multiple functional entry points can map back to the same physical equations.
+  For instance, :func:`sesolve`, :func:`fsesolve`, and :func:`krylovsolve`
+  all represent closed-system Schrödinger dynamics, using :class:`SESolver` but different integrators.
+- End users typically interact exclusively with top-level functions or the core :class:`Solver` class interface.
+  The numerical :class:`Integrator` and the configuration parsing via :class:`SolverOptions` operate silently under the hood.
+  Once execution concludes, the resulting :class:`Result` instance is returned
+  to the user and expected to be used as a read-only data container.
+
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Sections
+
+   terminology
+   .. motivation
+   .. solver
+   integrator
+   .. integrator_stochastic
+   .. result
+   .. solveroptions
+   .. feedback
--- a/doc/internals/solver/index.rst
+++ b/doc/internals/solver/index.rst
@@ -59,10 +59,13 @@ behind high-level functional wrappers:
    :caption: Sections
 
    terminology
-   .. motivation
-   .. solver
    integrator
-   .. integrator_stochastic
-   .. result
-   .. solveroptions
-   .. feedback
+
+..
+  Sections to add later:
+  motivation
+  solver
+  integrator_stochastic
+  result
+  solveroptions
+  feedback
--- a/doc/internals/solver/index.rst
+++ b/doc/internals/solver/index.rst
@@ -5,7 +5,7 @@ Quantum Solvers and Integrators Architecture
 
 The solver framework in QuTiP is responsible for simulating the time-evolution of quantum systems.
 Built directly on top of the core data layer,
-the architecture enforces a separation of concerns between physical equations and numerical solvers.
+the architecture enforces a separation of concerns between physical model and numerical solvers.
 
 This document serves as an overview of the internal architecture of QuTiP's solvers,
 guiding core developers and contributors on how the system is structured,
@@ -31,7 +31,7 @@ tasks between two primary base classes:
    rules down to the :class:`Integrator`. The selection of which numerical scheme to run
    is entirely dictated by the solver's configuration options through the ``"method"`` parameter.
 
-Generally, integrators are general-purpose ODE solvers. However, to maximize performance,
+Usually, integrators are general-purpose ODE solvers. However, to maximize performance,
 certain integrators are specialized to specific physical representations,
 where the boundary between physical assumptions and numerical routines is intentionally blended.
 
--- a/doc/internals/solver/integrator.rst
+++ b/doc/internals/solver/integrator.rst
@@ -107,14 +107,14 @@ Derivative Format
 
 While ordinary ODE interfaces strictly expect functions to represent the right-hand side (RHS),
 quantum solvers in QuTiP can profit from alternative formats.
-Every integrator exposes an :attr:`RHS_format` class attribute string that dictates
+Every integrator exposes an :attr:`rhs_format` class attribute string that dictates
 the type of RHS object required during initialization:
 
 * **"callable"**: The integrator expects a function matching the signature
-  ``rhs(t: float, state: Data) -> Data``. This is the default for most ODE methods.
-  For specific Cython-backed ODE implementations, if the passed function is a method of
-  :class:`QobjEvo`, it will bypass standard Python overhead and hook directly into the
-  Cython bindings of ``QobjEvo.matmul`` using a ``RHS`` object.
+  ``rhs(t: float, state: Data) -> Data`` returning the derivative of the `state` at time `t`.
+  This is the default for most ODE methods. For specific Cython-backed ODE implementations,
+  if the passed function is the method of :meth:`QobjEvo.matmul_data`, it will bypass standard
+  Python overhead and hook directly into the Cython bindings of ``QobjEvo.matmul`` using a ``RHS`` object.
 * **"matrix"**: The integrator accepts a complex matrix formatted as a
   :class:`qutip.core.data.Data` object, solving the linear matrix differential equation
   $\frac{dX}{dt} = \text{RHS} \times X$.
@@ -134,7 +134,7 @@ Every subclass inheriting from :class:`Integrator` must define the following cla
    by the numerical backend along with their default values.
    Upon instantiation, this is copied to an instance-level ``options`` dictionary.
 
-.. attribute:: RHS_format
+.. attribute:: rhs_format
 
    A string identifier defining the expected structural format of the Right-Hand Side (RHS)
    derivative function (e.g., ``"callable"``, ``"matrix"``, or ``"solver"``).
@@ -160,7 +160,7 @@ Let's build a simple linear step integrator that updates the current state by ad
     from qutip.solver.integrator import Integrator
 
     class LinearStepODE(Integrator):
-        RHS_format = "callable"
+        rhs_format = "callable"
         name = "Simple Linear Step"
         method = "linear"
         # An integrator may also define `integrator_options` here. If the attribute is not set,
@@ -220,9 +220,9 @@ derivative added directly into the existing ``out`` object. The function is
 permitted to reuse, modify, or overwrite the ``out`` container as needed.
 This pattern eliminates the considerable overhead of frequent memory allocations
 and deallocations inside tight numerical integration loops. The returned object
-does not need to the same as the ``out`` input if immutable or otherwise not reusable.
+does not need to the same as the ``out`` input if it's immutable or otherwise not reusable.
 
-These specialized integrators are marked as ``RHS_format="callable"``.
+These specialized integrators are marked as ``rhs_format="callable"``.
 Internally, they rely on the ``RHS`` helper class.
 This class acts as an interface layer that safely wraps an ordinary,
 out-of-place python derivative function and transforms it into one that behaves
--- a/doc/internals/solver/integrator.rst
+++ b/doc/internals/solver/integrator.rst
@@ -110,11 +110,11 @@ quantum equation in QuTiP can profit from alternative formats.
 Every integrator exposes an :attr:`RHS_format` class attribute string that dictates
 the type of RHS object required during initialization:
 
-* **"callable"**: The integrator expects a function matching the signature ``rhs(t, Data) -> Data``
-  or ``rhs(t, Data, out=Data)``. This is the default for most ODE methods.
+* **"callable"**: The integrator expects a function matching the signature
+  ``rhs(t: float, state: Data) -> Data``. This is the default for most ODE methods.
   For specific Cython-backed ODE implementations, if the passed function is a method of
   :class:`QobjEvo`, it will bypass standard Python overhead and hook directly into the
-  Cython bindings of ``QobjEvo.matmul``.
+  Cython bindings of ``QobjEvo.matmul`` using a ``RHS`` object.
 * **"matrix"**: The integrator accepts a complex matrix formatted as a
   :class:`qutip.core.data.Data` object, solving the linear matrix differential equation
   $\frac{dX}{dt} = \text{RHS} \times X$.
@@ -205,3 +205,46 @@ register it using the solver class's target method:
 Once added, the method is immediately accessible via the high-level functional
 wrappers by configuring the options map (e.g., ``options={"method": "linear"}``).
 Registering the class with the parent :class:`Solver` base class makes it globally available to all solvers except stochastic solver.
+
+
+The RHS class
+-------------
+To optimize performance, certain Cython-backed integrators reuse memory when
+computing derivatives by supporting an in-place functional signature:
+
+.. code-block:: python
+
+    rhs(t: float, state: Data, out: Data) -> Data
+
+In this paradigm, the operation should be interpreted as returning the computed
+derivative added directly into the existing ``out`` object. The function is
+permitted to reuse, modify, or overwrite the ``out`` container as needed.
+This pattern eliminates the considerable overhead of frequent memory allocations
+and deallocations inside tight numerical integration loops. The returned object
+does not need to the same as the ``out`` input if immutable or otherwise not reusable.
+
+These specialized integrators are marked as ``RHS_format="callable"``.
+Internally, they rely on the ``RHS`` helper class.
+This class acts as an interface layer that safely wraps an ordinary,
+out-of-place python derivative function and transforms it into one that behaves
+correctly when subjected to in-place calls.
+
+Within the QuTiP's solver, physical derivatives are usually
+computed via :meth:`QobjEvo.matmul_data`. The integrators automatically detects
+that method and bind directly to the low-level Cython implementation with
+in-place support.
+
+If you are providing a custom Python callable, you must explicitly wrap your
+function inside an ``RHS`` instance to leverage in-place performance optimizations:
+
+
+.. code-block:: python
+
+    from qutip.solver.integrator._rhs import RHS
+
+    def custom_diff(t, state, out):
+        # Perform in-place operations directly on the 'out' data object
+        out += ...
+        return out
+
+    derivative = RHS(custom_diff, inplace=True)
--- a/doc/internals/solver/integrator.rst
+++ b/doc/internals/solver/integrator.rst
@@ -1,0 +1,207 @@
+.. _integrator_internal:
+
+Integrator Architecture and Design
+##################################
+
+The numerical solution of differential equations is a cornerstone of QuTiP's
+time-evolution solvers. Because different quantum systems present varying behaviours,
+no single numerical algorithm is optimal for all scenarios.
+
+The :class:`Integrator` framework provides a unified, backend-agnostic interface
+that wraps external Ordinary Differential Equation (ODE) libraries. This design
+allows solvers to select routines from multiple distinct libraries through a single,
+interchangeable class.
+
+
+Motivation
+==========
+
+While Python packages like SciPy offer a robust collection of integration routines,
+they present several architectural limitations when applied directly to quantum dynamics:
+
+* **Inconsistent API Interfaces**: Across SciPy, interfaces vary drastically
+  (e.g., the older :class:`scipy.integrate.ode` vs. the modern :meth:`scipy.integrate.solve_ivp`),
+  and some legacy algorithms do not natively support complex numbers without manual
+  splitting into real and imaginary parts.
+* **Hardware Limitations**: Standard SciPy integration routines are strictly
+  bound to CPU execution, preventing seamless acceleration via GPUs or distributed architectures.
+* **Specialized Quantum Algorithms**: Certain high-performance routines tailored
+  for quantum mechanics —such as Krylov subspace methods or custom unitary propagators—
+  are too domain-specific to be included in general-purpose mathematical libraries.
+
+The :class:`Integrator` class resolves these challenges by abstracting numerical backends
+into an interchangeable component, ensuring that switching from a SciPy Fortran solver to
+a GPU-accelerated backend requires only changing a configuration string.
+
+
+Core Design
+===========
+
+The abstract base class :class:`Integrator` serves as the common template for all
+numerical integration backends.
+
+Deterministic Time-Evolution
+----------------------------
+For standard solvers, the integrator acts sequentially. Given an initial time $t_0$
+and an initial state vector or density matrix, it propagates the state forward to
+requested target times. This behaviour is governed by four core methods:
+
+* :meth:`set_state(t, state)`: Configures the initial conditions of the
+  integrator, seeding it with the starting time and the state array.
+* :meth:`get_state()`: Returns the current internal numerical state of the integration
+  loop as a tuple of ``(t, state)``.
+* :meth:`integrate(t)`: Steps the numerical solver explicitly to the target time ``t``
+  using the internal state configuration.
+* :meth:`run(tlist)`: A generator method that yields successive ``(t, state)`` tuples at
+  each timestamp specified in the user-provided array ``tlist``.
+
+.. note::
+   The output state returned by :meth:`get_state`, :meth:`integrate`, and :meth:`run`
+   is consistently packed as a tuple: ``(t, state)``.
+
+.. note::
+   All inputs and outputs for states are handled as :class:`qutip.core.data.Data` objects.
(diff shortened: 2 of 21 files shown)


doc/internals/solver/index.rst
```diff
@@ -0,0 +1,71 @@
...
+
+2. **The Numerical Layer (:class:`Integrator`)**: This layer handles the low-level
+   numerical implementation of Ordinary Differential Equations (ODEs) or Stochastic
+   Differential Equations (SDEs). The solver communicates the physical state and derivative
+   rules down to the :class:`Integrator`. The selection of which numerical scheme to run
+   is entirely dictated by the solver's configuration options through the ``"method"`` parameter.
+
+Generally, integrators are general-purpose ODE solvers. However, to maximize performance,
+certain integrators are specialized to specific physical representations,
+where the boundary between physical assumptions and numerical routines is intentionally blended.
```
hodgestar:
```suggestion
Usually, integrators are general-purpose ODE solvers. However, to maximize performance,
certain integrators are specialized to specific physical representations,
where the boundary between physical assumptions and numerical routines is intentionally blended.
```

doc/internals/solver/index.rst
```diff
@@ -0,0 +1,71 @@
+.. _solver_internals:
+
+Quantum Solvers and Integrators Architecture
+############################################
+
+The solver framework in QuTiP is responsible for simulating the time-evolution of quantum systems.
+Built directly on top of the core data layer,
+the architecture enforces a separation of concerns between physical equations and numerical solvers.
```
hodgestar:
```suggestion
The solver framework in QuTiP is responsible for simulating the time-evolution of quantum systems.
Built directly on top of the core data layer,
the architecture enforces a separation of concerns between physical model and numerical solvers.
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+##################################
+
+The numerical solution of differential equations is a cornerstone of QuTiP's
+time-evolution solvers. Because different quantum systems present varying behaviours,
+no single numerical algorithm is optimal for all scenarios.
+
+The :class:`Integrator` framework provides a unified, backend-agnostic interface
+that wraps external Ordinary Differential Equation (ODE) libraries. This design
+allows solvers to select routines from multiple distinct libraries through a single,
+interchangeable class.
```
hodgestar, line 13:
```suggestion
The :class:`Integrator` framework provides a unified, backend-agnostic interface
that wraps external or custom Ordinary Differential Equation (ODE) libraries. This design
allows solvers to select routines from multiple distinct libraries through a single,
interchangeable class.
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+Activating the newly assigned options requires an explicit call to :meth:`reset()`.
+This tears down the active backend instance and restarts it with the new options
+while preserving the active state and time.
+
+
+Derivative Format
+=================
+
+While ordinary ODE interfaces strictly expect functions to represent the right-hand side (RHS),
+quantum equation in QuTiP can profit from alternative formats.
```
hodgestar:
```suggestion
While ordinary ODE interfaces strictly expect functions to represent the right-hand side (RHS),
quantum solvers in QuTiP can profit from alternative formats.
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+Every integrator exposes an :attr:`RHS_format` class attribute string that dictates
+the type of RHS object required during initialization:
+
+* **"callable"**: The integrator expects a function matching the signature
+  ``rhs(t: float, state: Data) -> Data``. This is the default for most ODE methods.
+  For specific Cython-backed ODE implementations, if the passed function is a method of
+  :class:`QobjEvo`, it will bypass standard Python overhead and hook directly into the
+  Cython bindings of ``QobjEvo.matmul`` using a ``RHS`` object.
```
hodgestar: It would be helpful to add a sentence describing what the arguments and return value of `rhs` are. Perhaps:

The ``rhs`` function returns the derivative of the `state` at time `t`.

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Every subclass inheriting from :class:`Integrator` must define the following class-level attributes:
+
+.. attribute:: integrator_options
+
+   A class-level dictionary defining the comprehensive set of configuration keys supported
+   by the numerical backend along with their default values.
+   Upon instantiation, this is copied to an instance-level ``options`` dictionary.
+
+.. attribute:: RHS_format
```
hodgestar:
```suggestion
.. attribute:: rhs_format
```
I would personally rename this to all lowercase. Uppercase is not typically used in Python attribute names.

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+
+
+Creating New Integrators
+========================
+
+To implement a custom numerical backend, create a subclass derived from :class:`Integrator`.
+At a minimum, your class must override :meth:`integrate`, :meth:`set_state`, :meth:`get_state`,
+and either :meth:`__init__` or :meth:`_prepare`:
+
+.. code-block:: python
```
hodgestar, line 158:
```suggestion

Let's build a simple linear step integrator that updates the current state by adding `dt * derivative(t, state)`. This integrator won't provide very accurate results, but it will nicely illustrate what's required.

.. code-block:: python
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+
+    from qutip.solver.integrator import Integrator
+
+    class LinearStepODE(Integrator):
+        RHS_format = "callable"
+        name = "Simple Linear Step"
+        method = "linear"
+        # Uses default empty dict if no specialized integrator_options are defined
```
hodgestar:
```suggestion
        # An integrator may also define `integrator_options` here. If the attribute is not set,
        # a default empty dict is assumed, indicating that their are not special options for
        # this integrator.
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+
+        def _prepare(self):
+            # __init__ automatically prepares the local options dict,
+            # stores the rhs in self.derivative, and initializes self._is_set to False.
+            # Apply option here!
```
hodgestar:
```suggestion
            # Apply any specialized options here!
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+
+.. code-block:: python
+
+    from qutip.solver import SESolver
+
+    SESolver.add_integrator(LinearStepODE, "linear")
+
+Once added, the method is immediately accessible via the high-level functional
+wrappers by configuring the options map (e.g., ``options={"method": "linear"}``).
+Registering the class with the parent :class:`Solver` base class makes it globally available to all solvers except stochastic solver.
```
hodgestar, line 206:
```suggestion

Registering the class with the parent :class:`Solver` base class makes it globally available to all solvers except stochastic solver.
```
Just making this a separate paragraph.

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+.. code-block:: python
+
+    rhs(t: float, state: Data, out: Data) -> Data
+
+In this paradigm, the operation should be interpreted as returning the computed
+derivative added directly into the existing ``out`` object. The function is
+permitted to reuse, modify, or overwrite the ``out`` container as needed.
+This pattern eliminates the considerable overhead of frequent memory allocations
+and deallocations inside tight numerical integration loops. The returned object
+does not need to the same as the ``out`` input if immutable or otherwise not reusable.
```
hodgestar:
```suggestion
and deallocations inside tight numerical integration loops. The returned object
does not need to the same as the ``out`` input if is immutable or otherwise not reusable.
```

doc/internals/solver/integrator.rst
```diff
@@ -0,0 +1,244 @@
...
+These specialized integrators are marked as ``RHS_format="callable"``.
+Internally, they rely on the ``RHS`` helper class.
+This class acts as an interface layer that safely wraps an ordinary,
+out-of-place python derivative function and transforms it into one that behaves
+correctly when subjected to in-place calls.
+
+Within the QuTiP's solver, physical derivatives are usually
+computed via :meth:`QobjEvo.matmul_data`. The integrators automatically detects
+that method and bind directly to the low-level Cython implementation with
+in-place support.
```
hodgestar, line 234:
```suggestion
Within the QuTiP's solver, physical derivatives are usually
computed via :meth:`QobjEvo.matmul_data`. The integrators automatically detect
that method and bind directly to the low-level Cython implementation with
in-place support.
```

doc/internals/solver/terminology.rst
```diff
@@ -0,0 +1,61 @@
...
+See the rest of the guide for more in depth information.
+
+Quantum system
+    Set of operators and physical parameters used to build the dynamic equations.
+    Common components are the Hamiltonian, collapse or jump operators,
+    bath coupling operators and environment/spectral noise configurations.
+
+method
+    The identifier string used to select the numerical integration technique for a time-evolution run.
+    Can refer to ODE or SDE integration algorithm, depending on the solver.
```
hodgestar:
```suggestion
    May refer to an ODE or SDE integration algorithm, depending on the solver.
```

doc/internals/solver/terminology.rst
```diff
@@ -0,0 +1,61 @@
...
+    Ordinary Differential Equation.
+
+      $$ \frac{dX}{dt} = f(t, X) $$
+
+SDE
+    Stochastic Differential Equation:
+
+      $$ dX = f(t, X) dt + \sum_i g_i(t, X) dW_i $$
+
+    where $dW_i$ represent independent Wiener processes modeling Gaussian white noise.
```
hodgestar:
```suggestion
    where $dW_i$ represents an independent Wiener processes modeling Gaussian white noise.
```

doc/internals/solver/terminology.rst
```diff
@@ -0,0 +1,61 @@
...
+
+RHS
+    Right-Hand Side
```
hodgestar:
```suggestion
    Right-Hand Side.
```

doc/internals/solver/terminology.rst
```diff
@@ -0,0 +1,61 @@
...
+    The representation describing the derivative in a differential equation
+    (i.e., the $f(t, X)$ term in an ODE).
+
+Feedback
+    The term we use for quantum dynamics where the underlying quantum system parameters
+    depend explicitly on the instantaneous state of the system, introducing a form of non-linearity.
+    For example, when its Hamiltonian depends on the expectation value of an operator or the state itself: $H = H(t, \rho)$.
+    Different :class:`Solver` provide can provide different feedback format.
```
hodgestar:
```suggestion
    For example, when its Hamiltonian depends on the expectation value of an operator or the state itself: $H = H(t, \rho)$.
    Different :class:`Solver` sub-classes may provide different feedback formats.
```

doc/internals/solver/terminology.rst
```diff
@@ -0,0 +1,61 @@
...
+    * Non-deterministic evolution introduces pseudo-random variables,
+      yielding structurally unique trajectories on successive invocations.
+
+Stochastic
+    While broadly synonymous with non-deterministic behavior, in this section,
+    "stochastic" specifically denotes simulations driven by true Stochastic Differential Equations (SDEs)
+    containing Gaussian noise channels (e.g., :class:`SMESolver`).
+    Conversely, while Monte Carlo simulations (:class:`MCSolver`) are non-deterministic due to discrete,
+    random quantum jumps, their trajectories are piecewise-deterministic and are typically
+    excluded when discussing "stochastic evolution" within this architecture framework.
```
hodgestar:
```suggestion
    excluded when discussing "stochastic evolution".
```
Even outside of the context of QuTiP, I would not consider Monte Carlo methods necessarily indicating that the underlying evolution is stochastic.

hodgestar (commented):
I've reviewed the documentation, which is really great. Will move on to the code next.

qutip/solver/integrator/_rhs.pyx
```diff
@@ -0,0 +1,81 @@
...
+        a callable. This function can be both inplace or not.
+
+    inplace: bool
+        If ``derivative`` is a callable, whether that function take the output
+        inplace or not.
+
+    """
+    def __init__(
+        self,
+        derivative: Callable[[float, Data], Data] | QobjEvo,
```
hodgestar:
```suggestion
        derivative: Callable[[float, Data], Data] | Callable[[float, Data, Data], Data] | QobjEvo,
```
Should we explicitly add the inplace signature here too?

doc/internals/solver/integrator.rst
Ericgig: it's?

qutip/solver/integrator/explicit_rk.pyx
```diff
@@ -50,8 +51,9 @@ cdef class Explicit_RungeKutta:
 
     Parameters
     ----------
-    qevo : QobjEvo
-        The system to integrate: ```dX = qevo.matmul_data(t, X)``
+    deriv : callable, RHS
+        The system to integrate: ```dX = deriv(t, X)``, either as a callable
+        with signature (t, state) or a RHS objectself.
```
hodgestar:
```suggestion
        with signature (t, state) or a RHS object.
```

qutip/solver/integrator/integrator.py
```diff
@@ -25,8 +29,10 @@ class Integrator:
 
     Parameters
     ----------
-    system: qutip.QobjEvo
-        Quantum system in which states evolve.
+    derivative: callable | QobjEvo
+        Function to integrate.
+        Note that specific integrator may require other input as controlled by
+        the ``RHS_format`` attribute.
```
hodgestar:
```suggestion
        Note that specific integrators may require other input types. These types are specified by
        the ``RHS_format`` attribute.
```

qutip/solver/integrator/integrator.py
```diff
@@ -36,19 +42,17 @@ class Integrator:
...
+        Which format the ODE integrator rhs is used by the integration method.
+        - "callable": The integrator expect a function with signature
+          ``rhs(t, Data) -> Data``.
+          Some cython integrator use ``RHS`` object for cython binding.
+          "callable" is the default for most ODE methods.
+        - "matrix": The integrator take a constant matrix as a ``Data`` object
+          as the rhs and will solve the equation ``dX/dt = RHS @ X``.
+        - "solver": The integrator take the Solver instance that created it and
+          build the RHS itself. These are limited to integration method that
+          mixes the physics of the problem and the numerics.
```
hodgestar:
```suggestion
    RHS_format : {"callable", "matrix", "solver"}
        Which format the ODE integrator expects for its RHS:
        - "callable": The integrator expects a function with the signature
          ``rhs(t, Data) -> Data``.
          Some cython integrators use the RHS object for cython binding.
          "callable" is the default for most ODE methods.
        - "matrix": The integrator takes a constant matrix as a ``Data`` object
          as the RHS and will solve the equation ``dX/dt = RHS @ X``.
        - "solver": The integrator takes the Solver instance that created it and
          builds the RHS itself. Typically used only by integration methods that
          mix the system physics and the integrator numerics.
```

qutip/solver/integrator/integrator.py
```diff
@@ -219,16 +232,17 @@ def arguments(self, args):
         args : dict
             New arguments
         """
+        raise Exception("Remove this method")
```
hodgestar: I guess this is still a TODO?

qutip/solver/integrator/integrator.py
```diff
@@ -59,16 +63,21 @@ class Integrator:
     # Dict of used options and their default values
     integrator_options = {}
     _options = None
-    # Can evolve time dependent system
-    support_time_dependant = None
-    # Whether the integrator used the system QobjEvo as a blackbox
-    supports_blackbox = None
+    # How the rhs is passed to the integrator.
+    # "Solver", "matrix", or "callable".
```
hodgestar:
```suggestion
    # "solver", "matrix", or "callable".
```
It also wasn't clear to me in my first pass through the code where support for these three constants lies and how a new type would be added. Are they defined by the new `RHS` object so that adding more requires extending the `RHS` object? Or somewhere else?

qutip/solver/integrator/integrator.py
```diff
@@ -200,7 +211,9 @@ def run(self, tlist):
         for t in tlist[1:]:
             yield self.integrate(t, False)
 
-    def reset(self, hard=False):
+        #TODO: why cut tlist[0] here, not in solver?
```
hodgestar, line 214: I didn't fully understand this TODO.

qutip/solver/solver_base.py
```diff
@@ -286,7 +286,22 @@ def _get_integrator(self):
             integrator = method
         else:
             raise ValueError("Integrator method not supported.")
-        integrator_instance = integrator(self.rhs, self.options)
+        if integrator.RHS_format == "callable":
```
hodgestar: Ah, I see support for the different RHS formats is defined here.

hodgestar (commented):
Still need to review `_rhs.pxd` and `_rhs.pyx`, and tests, but the rest is has had a first pass.

qutip/solver/solver_base.py
```diff
@@ -286,7 +286,22 @@ def _get_integrator(self):
...
+            integrator_instance = integrator(
+                self.rhs.matmul_data, self.options
```
hodgestar, line 291: Perhaps not for this PR, but I was wondering how to create a solver if one has the RHS callable but not a `Qobj` or `QobjEvo`? E.g. Let's say I use the proposed QuTiP symbolic layer to generate a JAX or other jitted function containing the derivative, how could I pass that to a solver and integrator?

hodgestar: Perhaps a new `"symbolic"` format would be needed?

qutip/solver/integrator/integrator.py
Ericgig: Why do we start from tlist[1] and not tlist[0]?
Sure it does no computation for tlist[0], but I feel it would be more natural.

qutip/solver/integrator/integrator.py
Ericgig: Yes removing, when the RHS is a callable, the integrator can't access the args.

qutip/solver/solver_base.py
Ericgig: No symbolic format if it creates a callable, just pass it to the integrator.
But at the solver level, some changes will be needed to support symbolic.

I plan to make a pass on the Solver too, splitting the building of the function better. Hopefully that will make it easier to fit.

qutip/solver/integrator/_rhs.pyx
```diff
@@ -0,0 +1,81 @@
...
+
+cdef class RHS:
+    """
+    Container for the derivative function in qutip's cython integrator.
+
+    When it's a python callable, wrap it and add inplace support.
+
+    When it's a bound method of QobjEvo: bind the ``QobjEvo.matmul_data``
+    cython call directly as the derivative.
+    **This ignore the exact method used.**
```
hodgestar: I'm not sure what this sentence is intended to say. Is it meant to say that the `RHS` class abstracts away the difference between a callable and a QObjEvo? Or something else?

qutip/solver/integrator/_rhs.pyx
```diff
@@ -0,0 +1,81 @@
...
+
+    Parameters
+    ----------
+    derivative: Callable[[float, Data, ...], Data] | QobjEvo
+        Function to integrate.
+        Can be either a QobjEvo where QobjEvo @ state is the derivative or
+        a callable. This function can be both inplace or not.
```
hodgestar:
```suggestion
        a callable. This function may be either inplace or not.
```

qutip/solver/integrator/_rhs.pyx
```diff
@@ -0,0 +1,81 @@
...
+
+    inplace: bool
+        If ``derivative`` is a callable, whether that function take the output
+        inplace or not.
```
hodgestar:
```suggestion
        When ``derivative`` is a callable, whether the function supplied takes the output
        argument in-place or not.
```

qutip/solver/integrator/_rhs.pyx
```diff
@@ -0,0 +1,81 @@
...
+
+    """
+    def __init__(
+        self,
+        derivative: Callable[[float, Data], Data] | Callable[[float, Data, Data], Data] | QobjEvo,
+        inplace: bool=False
```
hodgestar:
```suggestion
        inplace: bool=False,
```

qutip/solver/integrator/_rhs.pyx
```diff
@@ -0,0 +1,81 @@
...
+        """
+        if self.qevo_derr:
+            return self.qevo.matmul_data(t, state, out=out)
+
+        if self.inplace:
+            if out is None:
+                out = zeros_like(state)
+            out = self.derivative(t, state, out)
+        elif out is not None:
+            out = iadd(self.derivative(t, state), out)
```
hodgestar, line 83: Do we need to add a `scale` parameter to `apply`? If seems strange to always do `out = out + derivative(...)` rather than `out = out + scale * derivative`.

qutip/solver/integrator/integrator.py
hodgestar: I would say that we don't include `tlist[0]` because presumably we set `t` to `tlist[0]` when calling `set_state` and that we now one to integrate to each next `t`.

I feel the behaviour is helpful, but I agree it's all a bit implicit.

qutip/tests/solver/test_integrator.py
```diff
@@ -13,48 +17,150 @@
...
+
+class TestIntegratorCallable:
+    @pytest.fixture(params=[
+        method
+        for method, integrator in Solver.avail_integrators().items()
+        if integrator.rhs_format == "callable"
+    ])
+    def method(self, request):
+        # Callable are the general method that should all be registered with
+        # the general Solver
```
hodgestar:
```suggestion
        # The callables returned are the general method registered with
        # the base Solver class.
```

qutip/tests/solver/test_integrator.py
```diff
@@ -13,48 +17,150 @@
...
         return request.param
 
-    @pytest.fixture(params=list(MESolver.avail_integrators().keys()))
-    def me_method(self, request):
+    def _check_integrator(self, integrator_instance, state0, analytical):
+        integrator_instance.set_state(0, state0)
+        tlist = np.linspace(0, 2, 21)
```
hodgestar:
```suggestion
        tlist = np.linspace(0, 2, 21)
        integrator_instance.set_state(tlist[0], state0)
```
Perhaps makes the use of `tlist[0]` more obvious.

qutip/tests/solver/test_integrator.py
```diff
@@ -13,48 +17,150 @@
...
+    x = state.to_array()
+    der = np.array([
+        [1, t],
+        [x[1, 0] * -0.1, t**2],
+        [-0.1j * t * x[2, 0], t**3],
+    ])
+    der = _data.Dense(der)
+    if out is not None:
+        der += out
+    return der
```
hodgestar:
```suggestion
    if out is not None:
        out += der
    return out
```
It feels more correct to add the derivative to `out` here, but maybe there is some reason not too?

qutip/tests/solver/test_integrator.py
```diff
@@ -13,48 +17,150 @@
...
+        t_front, state = integrator_instance.mcstep(5)
+        t_target = (3 + t_front) / 2
+        t_in, state = integrator_instance.mcstep(t_target)
+        assert t_in == t_target
+        assert_allclose(analytical(t_in, state0), state.to_array(), atol=5e-5)
+
+    # When the derivative is C ordered, the messa
+    # @pytest.mark.filterwarnings("ignore:cannot stack columns")
+
+    @pytest.mark.parametrize(['derivative', "analytical"], [
```
hodgestar:
```suggestion
    @pytest.mark.parametrize(["derivative", "analytical"], [
```
Just for consistency.

qutip/tests/solver/test_integrator.py
```diff
@@ -13,48 +17,150 @@
...
+        assert_allclose(analytical(t, state0), state.to_array(), atol=5e-5)
+
+        t_front, state = integrator_instance.mcstep(5)
+        t_target = (3 + t_front) / 2
+        t_in, state = integrator_instance.mcstep(t_target)
+        assert t_in == t_target
+        assert_allclose(analytical(t_in, state0), state.to_array(), atol=5e-5)
+
+    # When the derivative is C ordered, the messa
+    # @pytest.mark.filterwarnings("ignore:cannot stack columns")
```
hodgestar: Should we remove this comment now?

qutip/solver/integrator/_rhs.pyx
hodgestar: The tests seem to pass without this, but I'm not sure what scales the derivative by the time step. Probably I am being stupid and missing something obvious.

qutip/tests/solver/test_integrator.py
```diff
@@ -13,48 +17,150 @@
...
 
-    @pytest.fixture(params=list(SESolver.avail_integrators().keys()))
-    def se_method(self, request):
+def analytical_2(t, x0):
+    return ((t * mat).expm() @ qutip.Qobj(x0)).full()
+
+derivative_3 = RHS(derivative_1, inplace=True)
+
+
+class TestIntegratorCallable:
```
hodgestar: Should we move `derivate_1`, `derivative_2`, `derivative_3` and `analytical_1` and `analytic_2` inside the body of the `TestIntegratorCallable` class like we do for the other test class derivatives?

hodgestar (commented):
Review done!

qutip/tests/solver/test_integrator.py
Ericgig: Fixing manually, der is returned:
```
    if out is not None:
        out += der
        der = out
    return der
```

qutip/tests/solver/test_integrator.py
Ericgig: Inside the class, the functions become a methods of the class and not stand alone function. It's an extra layer that I felt better avoiding.

qutip/solver/integrator/integrator.py
Ericgig: We had to have the state go in the integrator once:
```
        results.add(
            tlist[0],
            self._restore_state(self._integrator.get_state()[1], copy=False)
        )
```
https://github.com/qutip/qutip/blob/cd925a43285d48d255a95579ea67cc9d93dd3508/qutip/solver/solver_base.py#L206

So the initial state when through the same transformation than the other before being added to result. The integrator can return the state in another format: (scipy always convert to dense, etc.) and with cuquantum, this step added the internal hilbert space of the state.

qutip/solver/integrator/_rhs.pyx
Ericgig: This is no longer true, correcting.

qutip/solver/integrator/_rhs.pyx
Ericgig: In all our uses, `out` is always a zero buffer when passed here.
The derivative is usually reused at multiple step, so it's not accumulated in the state when computed, but returned as a stand alone.
`out[:] =  derivative(...)` should be a better image. (But it uses matmul while is `out += ...`)

qutip/solver/integrator/integrator.py
hodgestar: It's a good point about integrating to `tlist[0]` being free if the solver state already has `t = tlist[0]`. Something for a different PR?

qutip/solver/integrator/_rhs.pyx
hodgestar: Hmm. If `out` is assumed to be always a zero buffer, I would note that in the docstring of `apply`.

qutip/tests/solver/test_integrator.py
hodgestar: Mostly I suggested moving them in because it's how the other test examples in the file are handled and it did seem nice to group things together with where they are used. Is having them be methods annoying for some reason?

Ericgig: Somewhat annoying, in `parametrize`, it's not a method and lose the locals `mat` variable...
They can be set if not used directly in the parametrization.

hodgestar (approved):
Approved!

Ericgig:
Made the PR to qutip-jax to match these changes.
Will make one to qutip-cuquantum too before merging.

Ericgig:
@BoxiLi @mudit06mah 
I would like to merge this soon (before 5.4), but it breaks qutip-jax.
I already make the fix: qutip/qutip-jax#102.
When both are merged, it should not affect qutip-qip or qutip-qoc, but you might see error if version are fixed on one side only.

veronikakurth:
@Ericgig, thank you a lot, Eric!
Is it fine to add it to the 5.3.1 release?

Ericgig:
No, it break other packages...

PR #2917, opened 2026-06-16
Add user survey link to the sidebar

veronikakurth:
**Description**
Adding a sidebar notice linking to QuTiP User Survey 2026.
It is to be removed in a couple of months. 
image

**AI Tools Usage Disclosure**
ChatGPT: consulting on common ways to add temporary links and their pros/cons; generating a template for the survey notice.


Ericgig (approved):
Thank you, it should help a lot.

I have some nick-picks, all subjective, feel free to address or ignore:
- The text color: light grey over darker grey, I think it would be better if there was more contrast. 
- Maybe adding a background color to the block to make it more visible.
- I would put the box under the menu.

veronikakurth:
Thank you, Eric. I think all those are good points. :) At the end of the day, all the nick picks contribute to a better general impression, so why not optimise it a bit? 
Here is how it looks now: 
image
Feel free to comment on it.
Is the background color in the box not too bright/distracting?

PR #2918, opened 2026-06-16
Internal doc for SolverOptions

Ericgig:
**Description**
Follow up to #2916, add the options section.

I made some change to the class while working on the doc:

- Use weakref for the update method of the solver. We had a circular reference in `solver_instance._options.feedback.__self__ is solver_instance`.
- Improved the printing look.
- Changed `feedback` to `on_update`, to keep the feedback term for ODE feedback.
- Do not directly descent from `dict`: `pop` or `clear` would break it. But I needed to add a lot of dict method instead, not sure I like it.

--------------------------------------------

While I am at it, do we want to add static typing?

If we used dataclass format instead of `dict` for options, we could have static typing.
It would require to split solver and integrator options:
```
mesolve(..., solver_options=mesolve.SolverOptions(store_final_state=True))
sol = SESolver(H, options=options)
sol.solver_options.store_final_state = True  # Would be static typed.
sol.integrator_options.atol = 1e-5  # Since this would depend on the method, static typing would not work
```

We could have `sesolve.IntegratorOptions(method)` return the right options class in a in a type checker friendly way (I don't want to pollute the namespace making these directly available), but it would not include types for plug-in integrator.
(AI suggest to have a pyi file and have plug-ins modify it on install...)

It could be made backward compatible with both `.options` dict and `.solver_options` data classes attached to solvers (or have all of them support both item and attribute interface.)

@hodgestar @veronikakurth 
Do you have any opinion on whether it would be worth doing? 

------------------------------------

**AI Tools Usage Disclosure**
Gemini for documentation writing and design planning.


veronikakurth:
@Ericgig Thank you for sharing your work, Eric!
I need more time to think about it and to understand the changes to the class as well as the exact problem it is solving, but I fully support the idea of having an internal documentation about solver options. 

I have a hunch we could make our life a bit easier if we could somehow separate the tasks of maintaining the solver's state and validation. For validation, I was wondering if we could adopt using Pydantic models. Pydantic is a very mature library for data validation, pretty much adopted in the industry, so it's safe to rely on it, and it might be helpful for us. For example, using it would already remove the need to manually check for extra keywords. 

They have a lot of features, but I think the most relevant for our case could be "Dynamic model creation"
https://pydantic.dev/docs/validation/latest/concepts/models#dynamic-model-creation 
I'd need to play around a bit to see if I can come up with a concrete usage proposal and see if it could give us any benefits for the validation part (and if those benefits would be worth introducing a new dependency). 

P.S. It's been a while since I used Pydantic, and I mostly used it for static models in the past, so please take my suggestion with a grain of salt for now.

Ericgig:
@veronikakurth

I am suggesting adding static typing, not data validation.
- Static typing will allows users to have their editor autocomplete the possible options name with type hints.
- Data validation would move the error sooner: when the options is set instead of used.

I believe static typing will be more useful to users, they won't need to remember or check options key when using a solver.  As the one who made them, I remember that "adams" method and "dop853" have 80% the same option, but I can't says which one differ without checking. 

It's quite rare to update and options and not use it quickly after, so the users should see the error quickly. Also if we check ourself, we can add custom error messages when the options is used. (max_step = 0 is no limit, etc.) But it would be good to have some and not let the error comes from too deep. (scipy.integrate.ode only output an integer error code)

I will look at data validation too. Probably home made, pydantic is too much for what we need and I would prefer not add more dependencies that needed.

qutip/solver/options.py
```diff
@@ -1,76 +1,105 @@
 __all__ = []
 
 import warnings
+import weakref
 
 
-class _SolverOptions(dict):
+class _SolverOptions():
```
veronikakurth: Stylistic remark: I guess parentheses become redundant with this change.

veronikakurth: An idea: Should we let `_SolverOptions` inherit from `MutableMapping`? Or does anything speak against it? Some functions already implement the dictionary-like behaviour, namely, `__getitem__`, `__setitem__`, so we could also explicitly declare that contract for the class.
An advantage would be that we would not need to implement `keys()`, `items()`, etc., methods ourselves.

veronikakurth:
@Ericgig Thank you very much for your detailed explanation. Now, I understand the central concept better. 

I agree that having static typing would be of high value for users. It is indeed inconvenient to consult the documentation of the dependency (in this case, scipy) to check for the exact arguments set (read: "oftentimes it's also pain"). I'd also be glad to have the solution to this for `qutip-qoc` since it's also relevant there in the context of optimizer options (or also solver options if we allow choosing it for power users).
I'd be happy to participate in further discussion and hope to help shape the solution. 

I also understand and respect your point regarding keeping dependencies minimal. 

Side-remark: I think, though, I'm a bit confused by the juxtaposition of static typing vs validation. I see that the former is the more important topic in this PR, but I believe they both can coexist; it's not "either or" - just as you pointed out, for example, we may want to implement catching errors of passing the wrong options on a high level. For this, though, I think it would make sense to fix scipy dependency to `< 2.0` in order not to break validation if there are changes in scipy's interfaces.

qutip/solver/options.py
Ericgig: I went back to inherit from dict. It's simpler to remove the unwanted methods.

qutip/solver/options.py
```diff
@@ -32,51 +32,49 @@ def __init__(
...
+        self[key] = self._default[key]
         if (updater := self._on_update()) is not None:
             updater(key)
 
-    def __contains__(self, key):
-        return key in self._dict
-
-    def keys(self):
-        return self._dict.keys()
+    def pop(self):
```
veronikakurth: To be consistent with `dict`'s interface, `.pop` has to have `key` argument
(https://docs.python.org/3/library/stdtypes.html#dict.pop)

hodgestar:
> AI suggest to have a pyi file and have plug-ins modify it on install...

That sounds like a terrible idea. :/

doc/internals/solver/solveroptions.rst
```diff
@@ -0,0 +1,85 @@
...
+The configuration of QuTiP's physical solvers and numerical integrators is
+handled via specialized dictionary objects. Rather than exposing raw Python
+dictionaries directly, the framework utilizes an internal subclass,
+:class:`_SolverOptions`.
+
+
+Motivation
+==========
+
+Simple dictionary or dataclass lacks the feature needed to
```
hodgestar:
```suggestion
A simple dictionary or dataclass lacks the features needed to:
```

doc/internals/solver/solveroptions.rst
```diff
@@ -0,0 +1,85 @@
...
+
+* **Preventing Silent Silent Bugs**: Raw Python dictionaries accept any key-value
+  pair without validation. A simple typo, such as typing ``"abstol"`` instead of
+  ``"atol"``, would be silently ignored by a standard dictionary, causing the
+  underlying integrator to fall back to default values without ever alerting the user.
```
hodgestar:
```suggestion
* **Prevent silent bugs**: Raw Python dictionaries accept any key-value
  pair without validation. A simple typo, such as typing ``"abstol"`` instead of
  ``"atol"``, would be silently ignored by a standard dictionary, causing the
  underlying integrator to fall back to default values without ever alerting the user.
```

doc/internals/solver/solveroptions.rst
```diff
@@ -0,0 +1,85 @@
...
+* **Integrator State Synchronization**: Modern differential equation solvers are
+  stateful; they maintain adaptive step histories, interpolation tables, and
+  multi-step coefficients. If a user modifies an integration parameter *after*
+  a simulation has started stepping forward, the internal state of the underlying
+  ODE engine must be forcefully reset to prevent numeric instability or crashes.
+  A standard dictionary cannot broadcast its updates to intercept these mid-run changes.
```
hodgestar:
```suggestion
* **Integrator state synchronization**: Modern differential equation solvers are
  stateful; they maintain adaptive step histories, interpolation tables, and
  multi-step coefficients. If a user modifies an integration parameter *after*
  a simulation has started stepping forward, the internal state of the underlying
  ODE engine must be forcefully reset to prevent numeric instability or crashes.
  A standard dictionary cannot broadcast its updates to intercept these mid-run changes.
```

doc/internals/solver/solveroptions.rst
```diff
@@ -0,0 +1,85 @@
...
+* **Namespace Isolation**: Different integration backends support entirely
+  different sets of configuration options. For example, a specialized Krylov subspace
+  propagator expects parameters like Krylov dimension limits, which are completely
+  meaningless to a standard SciPy ``zvode`` integrator.
```
hodgestar:
```suggestion
* **Namespace isolation**: Different integration backends support entirely
  different sets of configuration options. For example, a specialized Krylov subspace
  propagator expects parameters like Krylov dimension limits, which are completely
  meaningless to a standard SciPy ``zvode`` integrator.
```

doc/internals/solver/solveroptions.rst
```diff
@@ -0,0 +1,85 @@
...
+
+The :class:`_SolverOptions` class acts as an intelligent configuration firewall.
+By operating as a reactive, validating proxy, it shields developers from managing
+manual synchronization loops and protects users from configuration errors, ensuring
+that options can safely change dynamically throughout a system's evolutionary life cycle.
```
hodgestar:
```suggestion
The :class:`_SolverOptions` class addresses these short comings.
By operating as a reactive, validating proxy, it shields developers from managing
manual synchronization loops and protects users from configuration errors, ensuring
that options can safely change dynamically throughout a system's evolutionary life cycle.
```

doc/internals/solver/solveroptions.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Core Features
+-------------
+
+The :class:`_SolverOptions` class inherits directly from Python's built-in
+:repr:`dict`, behaving like a standard key-value map while injecting four
+crucial architectural features:
```
hodgestar:
```suggestion
The :class:`_SolverOptions` class inherits directly from Python's built-in
:repr:`dict`, behaving like a standard key-value map while adding four
important features:
```

qutip/solver/options.py
```diff
@@ -11,19 +12,22 @@ class _SolverOptions(dict):
     ----------
     default : dict
         Default dict, only keys in this will be accepted.
-    feedback : callable, ``f(keys : set) -> None``, optional
+    on_update : callable, ``f(keys : set) -> None``, optional
```
hodgestar, line 15: Nice renaming of this parameter. The new name is much clearer.

qutip/solver/options.py
```diff
@@ -38,39 +42,58 @@ def __setitem__(self, key, val):
...
+
+    def clear(self):
+        raise RuntimeError("Can't remove options")
+
+    def update(self, *args, **kwargs):
+        tmp = {}
+        tmp.update(*args, **kwargs)
+        for key, val in tmp.items():
+            # Ensure all keys are present
+            self[key] = val
```
hodgestar, line 66: This currently leaves the options in a partially updated state if one of the updates is invalid. Is it worthwhile making the update not partially succeed?

qutip/solver/options.py
```diff
@@ -38,39 +42,58 @@ def __setitem__(self, key, val):
...
 
     def copy(self):
         return self.__class__(
             self._default,
-            self._feedback,
+            None,
```
hodgestar, line 75:
```suggestion
            self._on_update(),
```
I'm not sure if we should avoid copying over the update function, but if we don't we should document that it isn't copied somewhere.

hodgestar (commented):
Reviewed. I made some comments, but generally looks good.

hodgestar:
> While I am at it, do we want to add static typing?

I'm not sure. Definitely not in this PR, I would say. Maybe if we can find a nice implementation?

qutip/solver/options.py
Ericgig: We should not copy this. Only the instance in `solver_instance._options` should have this.
Changing option on the copy will not change the original one and therefore the new value will not be visible to the solver.

Ericgig:
> I'm not sure. Definitely not in this PR, I would say. Maybe if we can find a nice implementation?

I tried a few ways, but did not find anything pleasing. It would be breaking changes or having 
```
def sesolve(
    ..., 
    options: dict=None, 
    solver_options: SESolveOptions = None, 
    integrator_options: ODEOptions = None,
)
```
With ODEOptions an union of multiple classes making static type checker unhelpful...


For now, I am leaving it as is...

hodgestar (approved):
Happy for this to be merged. Thanks, @Ericgig!

PR #2920, opened 2026-06-21
Document tensor_contract dims indexing

### LINKED ISSUE #385: Better documentation of `tensor_contract`
The usage of the `tensor_contract` function is explained in [QuTiP Example: Superoperators, Pauli Basis and Channel Contraction](http://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/example-superop-contract.ipynb).

It is explained in terms of the graphical notation in http://arxiv.org/abs/1111.6950. However, there is no figure explaining how the "tensor wires" are numbered. In the notebook it says "Numbering the tensor wires from 0 to 3, this corresponds to a tensor_contract argument of (1, 3)." But it is not clear what the convention is... does the numbering follow in an anticlockwise fashion, starting with 0 at the top left?

A figure in the notebook showing the numbering convention would be extremely helpful.

jeonghoon-ad:
Closes #385.

This documents how `tensor_contract` integer arguments map to `Qobj.dims` in
the reduced-channel guide. The added text spells out the ordinary-operator case
`[[a0, a1], [b0, b1]]`, then connects the existing `(1, 3)` example to the
nested superoperator dims returned by `to_super(identity([2, 2]))`.

The patch is docs-only.

Testing:

- `git diff --check`
- local smoke check with `qutip 5.2.3`:
  - `tensor_contract(Qobj(..., dims=[[2, 2], [2, 2]]), (1, 3)).dims == [[2], [2]]`
  - `tensor_contract(to_super(identity([2, 2])), (1, 3)).dims == [[[2], [2]], [[2, 2], [2, 2]]]`

Not run:

- Sphinx/docutils parse. The active local Python environment does not have
  `sphinx` or `docutils` installed.

### DIFF
--- a/doc/changes/2920.doc
+++ b/doc/changes/2920.doc
@@ -1,0 +1,1 @@
+Document how `tensor_contract` indices map to flattened `Qobj.dims` entries.
--- a/doc/guide/guide-super.rst
+++ b/doc/guide/guide-super.rst
@@ -66,8 +66,37 @@ We can think of the :math:`\scriptstyle \rm CNOT` here as a system-environment r
    :width: 2.5in
 
 
-The two tensor wires on the left indicate where we must take a tensor contraction to obtain the measurement map.
-Numbering the tensor wires from 0 to 3, this corresponds to a :func:`.tensor_contract` argument of ``(1, 3)``.
+The integer arguments to :func:`.tensor_contract` address the scalar entries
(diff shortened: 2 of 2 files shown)


Ericgig (commented):
Could you add a towncrier entry and re-check the contribution guideline?

Ericgig (approved):
Thank you

Issue #2924, opened 2026-07-01
Audit & refactor usages of SciPy's sparse matrix [deprecation]


veronikakurth:
Deprecation news from SciPy:
> SciPy sparse is shifting from a sparse matrix interface to a **sparse array** interface. In the next few releases we  expect to deprecate the sparse matrix interface. For documentation of the matrix interface, see the [spmatrix interface docs](https://docs.scipy.org/doc/scipy/reference/sparse.spmatrix_api.html#spmatrix-api). For guidance on converting existing code to sparse arrays, see [Migration from spmatrix to sparray](https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html#migration-to-sparray).

Despite QuTiP not relying on `scipy.sparse` for data storage, there are a couple of places to be audited to ensure the code will not break by the time of the final deprecation of sparse matrices - as of now, the 2nd of July 2026, the exact date for this deprecation seems not to be fixed. But we should be ready in advance.

**Namely**, 
1. `as_scipy()` / `to_scipy` conversions 
2. Qobj construction from user input (providing `scipy.sparse.csr_matrix`)
3. Usage of `scipy.sparse.linalg`

**Plan:**
- **Short-term**: Support both spmatrix and sparray for backward compatibility
- **Long-term**: Upon getting warnings on exact plans to deprecate spmatrix (we'll definitely notice it in CI), remove spmatrix usage -> make sure QuTiP users are warned about the change, too.

veronikakurth:
The final adaptations for compatibility with `csr_array` are made within #2992. Once it's complete, this task can be closed - till a later point when SciPy fully deprecates `csr_matrix`.

Issue #2922, opened 2026-06-29
Deprecation warnings break tests for Python 3.14


veronikakurth:
### Bug Description

Currently, we get the following error for all test workflows using Python 3.14 because of changes in the interface of `scipy.sparse. block_diag`.
The general strategy of Scipy is to switch from using sparse matrices to sparse arrays till v1.20.

For qutip, the issue affects only `qutip/piqs/piqs.py` and `qutip/tests/piqs/test_piqs.py`.
We have to refactor those files by following this migration guide: https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html 

```
 DeprecationWarning: `block_diag` is switching to the sparse array interface.
E           
E                   For the case where input arrays are numpy arrays, this function is
E                   switching to returning a sparse array instead of sparse matrix.
E                   Recover the sparse matrix return value by making one input a sparse matrix.
E                   For example, block_diag([coo_matrix(A), B]).
E                   Avoid this message for sparse array output using block_diag([coo_array(A), B]).
E                   For more information, see the spmatrix to sparray migration guide
E                   https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html
E           
E                   This function will be changed no earlier than v1.20.

../../../miniconda3/envs/test/lib/python3.14/site-packages/scipy/sparse/_construct.py:1485: DeprecationWarning
_________________________ TestDicke.test_purity_dicke __________________________

```

### Code to Reproduce the Bug

Run `qutip/tests/piqs/test_piqs.py` with Python 3.14 on any OS in CI.
**Problem**: Cannot reproduce it locally.

### Expected Behaviour

No errors upon using `block_diag` function from `scipy.sparse`

### Your Environment

```shell
QuTiP Version:      5.4.0.dev0+f343ee3
Numpy Version:      2.5.0
Scipy Version:      1.18.0
Cython Version:     3.2.6
Matplotlib Version: 3.11.0
Python Version:     3.14.6
Number of CPUs:     3
BLAS Info:          blas
INTEL MKL Ext:      None
Platform Info:      Darwin (arm64)
Installation path:  /Users/runner/work/qutip/qutip/qutip

Installed QuTiP family packages
-------------------------------

No QuTiP family packages installed.
```

PR #2925, opened 2026-07-02
Fix deprecation warning from scipy.sparse.block_diag in piqs.py

veronikakurth:
**Description**
Minimal fix to silence the warning when using the `block_diag` function from `scipy.sparse`:

```
 DeprecationWarning: `block_diag` is switching to the sparse array interface.
E           
E                   For the case where input arrays are numpy arrays, this function is
E                   switching to returning a sparse array instead of sparse matrix.
E                   Recover the sparse matrix return value by making one input a sparse matrix.
E                   For example, block_diag([coo_matrix(A), B]).
E                   Avoid this message for sparse array output using block_diag([coo_array(A), B]).
E                   For more information, see the spmatrix to sparray migration guide
E                   https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html
E           
E                   This function will be changed no earlier than v1.20.
```
The current warning is thrown because the _input_ to `block_diag` is dense (`numpy` array).
For the time being, the _output_ of `block_diag` would still be a sparse matrix (till SciPy v1.20). But after v1.20, the return type of `block_diag` would be forced to a sparse array. To guarantee that the output stays a sparse matrix (as `Qobj`'s constructor expects it, as of now), we adapt the input to `block_diag` to be sparse (one sparse argument is already enough).

For now, this is enough to fix the warning, but mid-term we have to audit all the places in qutip where sparse matrices are expected and migrate the code to use sparse arrays instead (#2924)

**Related issues or PRs**
#2922 

**AI Tools Usage Disclosure**
Claude Sonnet to better understand the scope of SciPy's deprecation plans and possible solutions for qutip-piqs / qutip in general.


veronikakurth:
@Ericgig, I'd appreciate your review, thanks!
The transition to sparray support, I'd suggest tackling in another PR since it's more overarching.

Ericgig (approved):
Thank you.

I would suggest to go to array immediately since we know we will need to fix it again soon.

PR #2926, opened 2026-07-02
Data: Add einsum and Dispatcher for einsum

mudit06mah:
**Description**
This PR adds the einsum function to the data layer as well as a dispatcher for einsum.
Also modifies einsum function in `dimensions.py` to call the einsum method in data layer.

**Related issues or PRs**
#2340 

**AI Tools Usage Disclosure**
- Gemini 3.1 Pro: Used for Debugging


veronikakurth:
@mudit06mah Thank you!
Please sync your branch with master; the fix for tests failing on Python 3.14 jobs is there. :)

qutip/core/data/einsum.py
```diff
@@ -0,0 +1,65 @@
...
+        The einsum subscript equation.
+    rest_operands : tuple of Data
+        The remaining QuTiP data objects to contract.
+    tensor_shapes : tuple of tuples
+        The N-D tensor shapes for each operand.
+    out_shape : tuple, optional
+        The final 2D shape of the resulting Data object.
+"""
+
+einsum.add_specialisations([
```
veronikakurth, line 74: I think we may need to add another specialisation that reflects a scalar as an output. Because Dense -> Dense is not the only possible input-output relationship.

qutip/core/data/einsum.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
+from .dense import Dense
+from .dispatch import Dispatcher as _Dispatcher
+
+__all__ = ['einsum', 'einsum_dense']
+
+
+def einsum_dense(
```
veronikakurth, line 8: This function's correctness is tested via tests for `dimensions.py`, which is great. But I'd argue that we still need a new test suite in accordance with how we test specialisations of data layer modules. See an example: https://github.com/qutip/qutip/blob/master/qutip/tests/core/data/test_norm.py#L10

qutip/core/data/einsum.py
```diff
@@ -0,0 +1,65 @@
...
+        op0, /,
+        subscripts,
+        rest_operands,
+        tensor_shapes,
+        out_shape=None
+):
+    """
+    Dense CPU specialization for einsum.
```
veronikakurth: Maybe the CPU/GPU distinction is not necessary to mention in the qutip core; just Dense should be fine.

qutip/core/data/einsum.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    operands = (op0,) + tuple(rest_operands)
```
veronikakurth: To fix the current error in CI tests (failing on CSR type due to a missing `as_ndarray` attribute), you may want to convert `rest_operands` explicitly to Dense before `einsum_csr` and `einsum_dia` are in place. (The Dispatcher will [convert](https://github.com/qutip/qutip/blob/75cd612c7551e4c431f51a895f7d7d4c2cd6ae94/qutip/core/data/dispatch.pyx#L89) `op0` to a dispatched type (Dense) if it gets something else, like CSR, as `op0` is the only parameter passed to `inputs` and hence it's the only operand that gets converted to a casted type; `rest_operands` are ignored from conversion.)

It is not the cleanest approach since it will have to be refactored later when einsum is defined for other data classes. But at least `op0` and `rest_operands` will be aligned in their data types within `einsum_dense`.

qutip/core/data/einsum.py
Ericgig: For the signature, how about:
```
def einsum_dense(op0, *rest_operands, *, subscripts, ...)
```
We could add support for `*args` to the dispatcher.

qutip/core/data/einsum.py
Ericgig: The dispatcher can't handle that.
You could just not dispatch on the output or return a 1x1 Dense and have the einsum in dimension extract the scalar.

qutip/core/dimensions.py
```diff
@@ -303,6 +304,8 @@ def einsum(subscripts, *operands):
         separated list of subscript labels.
     operands: list of array_like
         These are the arrays for the operation.
+    out_dims: tuple of int
+        The final 2D shape of the state.
```
veronikakurth: Currently, the name of the parameter and its description are not aligned. I'd reformulate the description to avoid confusion on the user's side. Here, we are dealing with dimensions, not with the shape. Also, let's make sure the type in the signature aligns as well: `dims_to_tensor_perm` expects `out_dims` as a list (or Dimensions). So does the einsum.

I'd also suggest avoiding the term "state": after all, the output of einsum can be not only a ket, but a quantum object of a different structure: operator, bra, superoperator (? this we don't know yet, I'm not sure it's covered in tests right now).

Thank you!

qutip/core/data/einsum.py
veronikakurth: When discussing adding the tests for specialisations today, we touched upon the current ambiguity in the function's output type and the issues it poses. Mudit will update the contract of `einsum_dense` to return only Dense and fix the downstream usages of it to extract the scalar. 

FIY @BoxiLi @mudit06mah

qutip/core/dimensions.py
```diff
@@ -312,15 +315,91 @@ def einsum(subscripts, *operands):
...
-    if result.shape == ():
-        return result
-    dims = [
-        [d for d in result.shape[:result.ndim // 2]],
-        [d for d in result.shape[result.ndim // 2:]]
-    ]
-    return from_tensor_rep(result, dims)
+    data_operands = tuple(op.data for op in operands)
+
+    if out_dims is None:
```
veronikakurth: As discussed earlier, please refactor this logic into a helper function for inferring dimensions. Thanks!

qutip/core/data/einsum.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Parameters
+    ----------
+    op0 : Data
+        The first operand, used for type dispatching.
+    subscripts : str
+        The einsum subscript equation.
+    rest_operands : tuple of Data
+        The remaining QuTiP data objects to contract.
+    tensor_shapes : tuple of tuples
```
veronikakurth, line 64: `tensor_perms` and `out_perm` are lacking in the parameter list.

qutip/core/data/einsum.py
BoxiLi: > We could add support for *args to the dispatcher.

We will probably not do this in this GSoC, maybe later, when the requirement is still there.

Ericgig: Touching the dispatcher can wait, but it would be good to set it so all Data object can put as positional argument so we don't have to modify every use when we do.

qutip/tests/core/data/test_einsum.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+import pytest
+
+from qutip.core import data
+from qutip.core.data import Dense, einsum_dense
+
+
+class TestEinsum:
```
veronikakurth, line 8: @Ericgig I've been thinking whether we have to be strict about following the `_GenericMixin` interface here. If so, then which one? Einsum doesn't fall into a category of a unary/binary/ternary operation due to variability of the arguments number. It's also a bit tricky to elegantly adapt `op_numpy` function here. 

Can we leave this test case without inheritance for now and come back to it when we extend the dispatcher to the variable number of arguments? I think then we will need to also add another abtraction to `test_mathematically.py`, the one that would match einsum and other similar functions better.

Ericgig: For now, you could use the unary one since only the first data object it used for the dispatching if it's easier that making your own from scratch.
It's probably fine to make one unique for just einsum. If other similar pop up we can rethink then, but einsum has some condition on inside structure. It's unique enough that generic pattern will probably not be enough to fully test it.

qutip/core/data/einsum.py
veronikakurth: @mudit06mah Just in case not every follow-up on this comment results in a notification - let's still address this suggestion by Eric before we merge. 

Afterwards, I'd be glad to go on and approve the PR to be merged soon, and we can verify with Eric one last time if there are no more important things left to be fixed.

qutip/tests/core/data/test_einsum.py
```diff
@@ -0,0 +1,107 @@
...
+                np.array([[0.0, 1.0j], [-1.0j, 0.0]], dtype=complex),
+                id="transpose",
+            ),
+            pytest.param(
+                "ijkl->kjil",
+                [(2, 2, 2, 2)], [(0, 1, 2, 3)], (1, 0, 2, 3), (4, 4),
+                [np.zeros((16,), dtype=complex).reshape(4, 4)],
+                np.zeros((16,), dtype=complex).reshape(4, 4),
+                id="tensor_permutation",
+            ),
```
BoxiLi, line 78: This test is a bit too trival, permutation of a zero matrix. We can get something more specific
```suggestion
            pytest.param(
                "ijkl->kjil",
                [(3, 2, 2, 2)], [(2, 0, 1, 3)], (1, 2, 0, 3), (4, 6),
                [np.arange(24, dtype=complex).reshape(6, 4)],
                np.array(
                    [
                        [0, 1, 8, 9, 16, 17],
                        [4, 5, 12, 13, 20, 21],
                        [2, 3, 10, 11, 18, 19],
                        [6, 7, 14, 15, 22, 23],
                    ],
                    dtype=complex,
                ),
                id="tensor_permutation",
            ),
```

BoxiLi (commented):
It might be trivial, but it might help to add another test (or let AI come up with one) that 
- doesn't have uniform dimension 2, like [2, 3] to test the shape handling.
- A superoperator example in test_dimensions.py to verify the support.

E.g. to test_einsum in test_dimensions.py
```python
    pytest.param("ijkl->kjil", [qutip.tensor(qutip.sigmam(), qutip.sigmaz())], qutip.tensor(qutip.sigmap(), qutip.sigmaz())),
    pytest.param(
        "ijk->jik",
        [qutip.Qobj(
            np.arange(6, dtype=complex).reshape(6, 1),
            dims=[[[2], [3]], [1]],
        )],
        qutip.Qobj(
            np.array([[0], [2], [4], [1], [3], [5]], dtype=complex),
            dims=[[[3], [2]], [1]],
        ),
        id="operator_ket_permutation",
    ),
    pytest.param(
        "ijkl->jikl",
        [qutip.Qobj(
            np.arange(36, dtype=complex).reshape(6, 6),
            dims=[[[2], [3]], [[2], [3]]],
        )],
        qutip.Qobj(
            np.array([
                [0, 1, 2, 3, 4, 5],
                [12, 13, 14, 15, 16, 17],
                [24, 25, 26, 27, 28, 29],
                [6, 7, 8, 9, 10, 11],
                [18, 19, 20, 21, 22, 23],
                [30, 31, 32, 33, 34, 35],
            ], dtype=complex),
            dims=[[[3], [2]], [[2], [3]]],
        ),
        id="superoperator_permutation",
    ),
 ```

File: qutip/core/dimensions.py [hunk 10]
```diff
@@ -303,24 +375,59 @@ def einsum(subscripts, *operands):
...
+        out_dims = _infer_out_dims(subscripts, operands)
+
+    tensor_shapes = tuple(
+        dims_to_tensor_shape(op.dims)
+        for op in operands
+    )
```
BoxiLi: Very minor issue, `dims_to_tensor_shape` actually calls `dims_to_tensor_perm` internally. So it is computed twice. But this is not heavy computation. Ok

qutip/core/dimensions.py
```diff
@@ -303,24 +375,59 @@ def einsum(subscripts, *operands):
...
+
+    tensor_perms = tuple(
+        dims_to_tensor_perm(op.dims)
```
BoxiLi: I think the reason we need this is purely for superoperators and operator-ket, is that true? If so, we need to add somet tests and maybe leave a short comment here.

qutip/core/data/einsum.py
BoxiLi: I think we can update with this, just switching the order of rest_operands and subscripts. 
```python
def einsum_dense(
        op0, /,
        *rest_operands,
        subscripts,
        tensor_shapes,
        tensor_perms,
        out_perm,
        out_shape=None
):
    """
    Dense specialization for einsum.
    """
    operands = (op0,) + rest_operands
    from .convert import to as _to
```
This makes sure that the dispatcher still uses op0 to define the type. Even if there are CSR in other operands.
Then we can call in dimensions.py
```python
result_data = einsum_dense(
    *data_operands,
    subscripts=subscripts,
    tensor_shapes=tensor_shapes,
    tensor_perms=tensor_perms,
    out_perm=out_perm,
    out_shape=out_shape
)
```

This will make the later migration to a proper support for mulitple positonal arguments.

Maybe this address @Ericgig 's suggestions.

BoxiLi:
Also, if not `out_dim` is given, does qutip.einsum still return the right answer? E.g. what if the input is a superoperator? Is `out_dim` then required?

qutip/tests/core/test_dimensions.py
```diff
@@ -447,7 +447,70 @@ def test_dims_comparison():
...
+            ], dtype=complex),
+            dims=[[[3], [2]], [[2], [3]]],
+        ),
+        id="superoperator_permutation",
+    ),
+    pytest.param(
+        "abcd,cdef->abef",
+        [
+            qutip.Qobj([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]], dims=[[2, 2], [2, 2]]),
+            qutip.ket2dm(qutip.tensor(qutip.basis(2, 0), qutip.basis(2, 1))),
```
veronikakurth: For better readability, would it be possible to organise these objects into named fixtures or would parameterisation not work then?

BoxiLi:
@mudit06mah Hey. There seems to be quite a big algorithm that goes into the dimension inference now. Looks like quite a complicated solution. What was the reason, and do we really need that?

veronikakurth:
> @mudit06mah Hey. There seems to be quite a big algorithm that goes into the dimension inference now. Looks like quite a complicated solution. What was the reason, and do we really need that?

And if there is a good reason for this, especially for the case of superoperators, it would be important to have more comments in the code to make it more readable for other maintainers.

BoxiLi:
It seems that the BFS graph solving in dim inference is trying to address a very rare use case: multiplication contraction together with partial transpose. This is allowed in einsum, but it makes dimension handling very convoluted. 

Maybe we should limit the einsum feature to multiplication + contraction, and refuse transpose. Permutation can be optional. Because in qutip we have well-established methods to deal with transposition and permutation. It does not make much sense to support it here under the pricing of maintaining a little BFS search algorithm. Thoughts?

BoxiLi:
https://github.com/BoxiLi/qutip/tree/einsum-simplify-dims

I tried one implementation, not completely tested

qutip/core/dimensions.py
```diff
@@ -291,7 +292,97 @@ def _frozen(*args, **kwargs):
...
+        ))
+
+    input_subs = inputs_part.split(",")
+
+    char_to_dim = {}
+    char_occurrences = {}
+    for op_idx, (op, sub) in enumerate(zip(operands, input_subs)):
+        row_len = len(op.dims[0])
+        for i, char in enumerate(sub):
+            is_row = i < row_len
```
BoxiLi: There seems to be an `IndexError` on composite (multi-subsystem) superoperators / operator-kets / operator-bras

`row_len = len(op.dims[0])` is only correct when `dims[0]` is a flat list of ints. For composite superoperators/operator-kets/operator-bras, `dims[0]` is nested and `len(op.dims[0])` undercounts the real flattened index count, so indexing later reads past the end of `op.dims[1]`.


```python
L1 = qutip.spre(qutip.tensor(qutip.sigmax(), qutip.sigmay()))
L2 = qutip.spost(qutip.tensor(qutip.sigmaz(), qutip.sigmax()))
einsum('ijklABCD,ABCDmnop->ijklmnop', L1, L2)
```
```
IndexError: list index out of range
```

Suggest reusing the existing `flatten`/`dims_to_tensor_shape`/`dims_to_tensor_perm` helpers already in this file instead of re-deriving the row/col split from raw `len()`.

I am not sure if this is the best design but maybe something like

```python
row_flat = flatten(op.dims[0])
col_flat = flatten(op.dims[1])
row_len = len(row_flat)
```

The test superoperator_permutation was removed because it is pure permutation, but we still need tests to cover  superoperators / operator-kets / operator-bras.

qutip/core/dimensions.py
```diff
@@ -303,24 +394,63 @@ def einsum(subscripts, *operands):
...
+        tensor_shapes=tensor_shapes,
+        tensor_perms=tensor_perms,
+        out_perm=out_perm,
+        out_shape=out_shape
+    )
+
+    # Extract scalar from the 1x1 Data object to
+    # fulfill the Dense output contract.
+    if out_dims is None:
+        return complex(result_data.to_array()[0, 0])
```
BoxiLi: I think this `to_array` will break jit compilation, `to_array` calls `full`.  We maybe can use `extract (result_data, copy=False)[0, 0]` from `qutip.core.data.extract`.

We need a test for scalar value in qutip-jax for `qutip.einsum` with jit.

qutip/core/dimensions.py
```diff
@@ -443,10 +445,13 @@ def einsum(subscripts, *operands, out_dims=None):
...
     )
 
-    # Extract scalar from the 1x1 Data object to
-    # fulfill the Dense output contract.
+    # Extract scalar from the 1x1 Data object.
     if out_dims is None:
-        return complex(result_data.to_array()[0, 0])
+        scalar_val = extract(result_data, copy=False)[0, 0]
+        try:
+            return complex(scalar_val)
```
BoxiLi: Try `np.complex`

BoxiLi:
And add the towncrier

qutip/core/dimensions.py
```diff
@@ -443,10 +445,13 @@ def einsum(subscripts, *operands, out_dims=None):
         out_shape=out_shape
     )
 
-    # Extract scalar from the 1x1 Data object to
-    # fulfill the Dense output contract.
+    # Extract scalar from the 1x1 Data object.
     if out_dims is None:
-        return complex(result_data.to_array()[0, 0])
+        scalar_val = extract(result_data, copy=False)[0, 0]
```
BoxiLi: extract(result_data)[0, 0]

qutip/core/dimensions.py
mudit06mah: Even while using `np.complex128()` we won't be able to remove the exception handling.

More context:
https://docs.jax.dev/en/latest/errors.html#jax.errors.ConcretizationTypeError

BoxiLi:
I think there might still be some safeguards missing.

```python
A = Qobj([[1, 2],
          [3, 4]])

B = Qobj([[5],
          [6]])

result = einsum("ij,jk->ki", A, B)
print(result.full())
```

This is actually something we said in doc that we do not support (transpose), but we didn't really check this output only transpose and raise an error

This currently outputs a result without transpose. So no error raised, but a wrong result.

```python
[[17],
 [39]]
```

Could you
- Try to find a solution to this, either support and change the doc, or don't support and raise an error.
- Give us a summary here which cases are not supported, and where they are checked and rejected, in your words?

qutip/core/dimensions.py
BoxiLi: What if we just remove this part, the result return a numpy scalar. The returned result will just be numpy scalar or jax.numpy scalar. Just not python scalar

mudit06mah:
@BoxiLi Here's a summary of cases that are not supported:

- Single operand, No contraction  (`ij -> ij`): Rejected at line 353-359
- Single operand with output reorder (`ij -> ji`): Rejected at line 353-359
- Row <-> Row or Col <-> contraction (`cdef,ghef->cdgh`): Rejected at line 362-369
- Output col before row ordering (`ij, jk->ki`): Rejected at line line 384-394 (The case you mentioned)

BoxiLi:
@Ericgig Would you like to give a quick look and see if there are any further problems that need to be fixed before merging?

qutip/core/data/einsum.py
```diff
@@ -0,0 +1,72 @@
...
+    einsum_dense,
+    name='einsum',
+    inputs=('op0',),
+    out=True
+)
+einsum.__doc__ = """
+    Data layer implementation of Einstein summation.
+
+    Parameters
+    ----------
```
Ericgig, line 57: Parameters don't match function inputs.

Ericgig (commented):
Looking good.

Docstring out of sync.

It's getting large enough that it could be moved out of dimensions.py into it's own file.

veronikakurth:
> It's getting large enough that it could be moved out of dimensions.py into its own file.

I agree - for example, into `qutip/core/einsum.py`?

BoxiLi:
We need `from qutip.core.einsum import einsum` in the `__init__` of core

BoxiLi (approved):

BoxiLi:
@veronikakurth Please let me know if you have further comments to address

veronikakurth:
@mudit06mah, merged it. Thank you for your work!

PR #2927, opened 2026-07-05
Fixed mcsolve trajectory allocation for unnormalized mixed states

### LINKED ISSUE #2880: mcsolve fails with non-normalized initial mixed state

RahulShastri003:
## Description

Fixes #2880.

`mcsolve` accepts the initial state as a `Qobj`. When the initial state is a mixed density matrix, `MCSolver.run` decomposes it into eigenstates and eigenvalues. This creates a list of pure-state components in the form `(psi, p)`.

The mixed-state trajectory code assumes that these weights sum to one. If the input density matrix is not normalized, the total weight can be smaller or larger than one. This can make `_minimum_roundoff_ensemble` allocate a different number of trajectories from the requested `ntraj`. The solver still creates seeds for the requested number of trajectories, which can later cause an `IndexError`.

This PR keeps the original eigenvalue weights for calculating the result states, but uses normalized weights for trajectory allocation. This makes sure the requested number of trajectories is allocated while keeping the original trace in the output states.

For direct ensemble input through `MCSolver.run`, where the user passes a list of `(state, weight)` pairs, the existing assumption that the weights sum to one is unchanged.

## Tests
Added a parameterized test for unnormalized mixed density matrices with traces below and above one.
The test checks that:
- `mcsolve` completes without an error.
- The requested number of trajectories is allocated.
- The total number of trajectories across the initial states matches `ntraj`.
- The states in `result.states` keep the trace of the original input state.

Tested with:
```bash
python -m pytest qutip/tests/solver/test_mcsolve.py
```

## AI Tools Usage Disclosure
OpenAI ChatGPT/Codex: Used to better understand the issue, improve the PR description, and provide suggestions/autocomplete in VS Code. I reviewed and implemented all the final changes.

### DIFF
--- a/qutip/tests/solver/test_mcsolve.py
+++ b/qutip/tests/solver/test_mcsolve.py
@@ -658,9 +658,23 @@ def test_mixed_equals_merged(improved_sampling, p):
     )
 
 
-def test_mcsolve_unnormalized_mixed_state():
+@pytest.mark.parametrize(
+    "weight_0, weight_1, expected_trace",
+    [
+        (0.5, 0.25, 0.75),
+        (0.75, 0.5, 1.25),
+    ],
+)
+def test_mcsolve_unnormalized_mixed_state(
+    weight_0, weight_1, expected_trace
+):
     ntraj = 10
-    initial_state = 0.5 * qutip.fock_dm(2, 0) + 0.25 * qutip.fock_dm(2, 1)
+
+    # Test density matrices with traces below and above one.
+    initial_state = (
+        weight_0 * qutip.fock_dm(2, 0)
+        + weight_1 * qutip.fock_dm(2, 1)
+    )
 
     result = mcsolve(
         qutip.sigmaz(),
@@ -671,8 +685,10 @@ def test_mcsolve_unnormalized_mixed_state():
         options={"progress_bar": False},
         seeds=1234,
     )
-
-    assert initial_state.tr() == 0.75
     assert result.num_trajectories == ntraj
     assert sum(result.ntraj_per_initial_state) == ntraj
+    assert all(
+        state.tr() == pytest.approx(expected_trace)
+        for state in result.states
+    )
     
--- a/qutip/tests/solver/test_mcsolve.py
+++ b/qutip/tests/solver/test_mcsolve.py
@@ -656,3 +656,23 @@ def test_mixed_equals_merged(improved_sampling, p):
         sum(merged_result.runs_weights + merged_result.deterministic_weights)
         == pytest.approx(1.)
     )
+
+
+def test_mcsolve_unnormalized_mixed_state():
+    ntraj = 10
+    initial_state = 0.5 * qutip.fock_dm(2, 0) + 0.25 * qutip.fock_dm(2, 1)
+
+    result = mcsolve(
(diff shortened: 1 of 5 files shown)


pmenczel:
It looks good to me. Could you please add the towncrier file as described in the PR template?

veronikakurth:
@RahulShastri003, thank you! Please fill out the AI tools usage disclosure.

RahulShastri003:
@pmenczel Thanks! I have now added the Towncrier entry in 'doc/changes/2927.bugfix'.

RahulShastri003:
@veronikakurth Thanks! I have updated the AI tools usage disclosure to the PR description.

pmenczel:
Thank you. I am sorry, but one more thing: you have added a test were the trace of the initial density matrix is less than one. Perhaps also include a test where it is larger than one?

qutip/tests/solver/test_mcsolve.py
```diff
@@ -656,3 +656,23 @@ def test_mixed_equals_merged(improved_sampling, p):
...
+        qutip.sigmaz(),
+        initial_state,
+        np.linspace(0, 1, 100),
+        [qutip.sigmam()],
+        ntraj=ntraj,
+        options={"progress_bar": False},
+        seeds=1234,
+    )
+
+    assert initial_state.tr() == 0.75
```
veronikakurth: I think it's not necessary to check the initial state's trace - it would be the responsibility of a test for the trace function to do so. Also, `initial_state` is not data that `mcsolve` can modify. This information, however, would be useful to mention in a comment before the initial state definition.

pmenczel: It would however make sense to check `state.tr() == 0.75` for all states in `result.states`.

RahulShastri003: Thanks @veronikakurth, that makes sense. I have modified the solution a little. Instead of modifying `initial_state` directly, I now use normalized state weights only for `ntraj` allocation and keep the original weights unchanged for the result states.

@pmenczel, thank you. I have also added a check for all states in `result.states`, and the parameterized test now covers initial states with traces both below and above one.

pmenczel (approved):
Thank you, it looks good to me. I have made a few minor changes for polishing:
* Cleaned up the logic of `MCSolver.run` a bit
* We prefer not to fix rng seeds in the tests, I have removed the seed
* I have copied the test over to `nm_mcsolve` which now also works with unnormalized mixed initial conditions

@veronikakurth please have another quick look :)

veronikakurth (approved):
@pmenczel Thank you very much! I agree with the changes. I also think the condition for normalization now should work "cleaner" due to taking into account the absolute tolerance parameter. 

P.S. Minor stylistic remark (based on the way we used this method, I think both approaches are functionally equivalent): `_minimum_roundoff_ensemble` method could also be as well decorated with `staticmethod` since it requires neither class nor instance data. 

P.P.S. To attribute both contributors, we should merge this PR **without squashing**.

PR #2930, opened 2026-07-12
fix: correct AttributeError in to_choi/to_chi error message

steps-re:
**Description**

`to_choi` and `to_chi` both raise `TypeError` when given a `Qobj` whose
`type` is neither `'super'` nor `'oper'`. Building that error message
formats `q_oper.choi`, which doesn't exist on `Qobj`, so the code raises
`AttributeError` instead and the real problem (an unsupported input type)
never gets reported.

`to_chi`'s message also said "to Choi" even though it's raised inside
`to_chi`.

Fix: format `q_oper.superrep` (which does exist) instead of `q_oper.choi`,
and fix `to_chi`'s message to say "Chi".

Added a regression test (`test_to_choi_to_chi_unsupported_type_raises_TypeError`)
covering both functions.

**Related issues or PRs**

None found. Not a duplicate of an open issue or PR.

**AI Tools Usage Disclosure**

Claude (Anthropic), Claude Code: used to locate the bug, verify it live
against a compiled build of qutip (built from source, `pip install -e .`
with cython/numpy/scipy), confirm `Qobj` has no `.choi` attribute, write
the one-line fix in both functions, add a regression test, and run the
existing `test_superop_reps.py` suite (100/100 passing, including the new
test) plus `pycodestyle` on the changed lines.


Ericgig (approved):
Thank you!



I though we updated all these to f-string...

PR #2931, opened 2026-07-12
fix: correct a batch of small correctness bugs across core and solvers

steps-re:
**Description**

Six small, independent correctness fixes, each verified against a source
build of `master` and each in its own commit with a regression test where
one adds value.

1. **`Qobj.logm` error message** (`qutip/core/qobj.py`). The non-square
   `TypeError` said "expm is only valid for square operators", copy-pasted
   from `expm`. Now says "logm".

2. **`HEOMResult.final_ado_state`** (`qutip/solver/heom/bofin_solvers.py`).
   The property guarded on `self._final_ado_state` but returned
   `self._final_state`, i.e. the plain density matrix rather than the
   `HierarchyADOsState`. This branch is reached when `store_final_state` is
   set with `store_states` off, so the property returned the wrong object
   and type. Now returns `self._final_ado_state`.

3. **`MCSolver.StateFeedback`** (`qutip/solver/mcsolve.py`). The feedback
   constructors were called with `open=open`, passing the builtin `open`
   as the flag (always truthy, so the state was treated as a density
   matrix) and silently dropping `prop`. MCSolver evolves kets, so this now
   mirrors `SESolver`: `open=False, prop=prop`. The stale docstring that
   documented `open` is updated to `prop`.

4. **`qdiags` unitary check** (`qutip/core/operators.py`). The main-diagonal
   branch tested `np.abs(diagonals) - 1 <= atol`, i.e. `|d| <= 1 + atol`,
   so a diagonal of modulus below 1 (e.g. 0.5) was flagged unitary. Now
   tests the distance from 1: `np.abs(np.abs(diagonals) - 1) <= atol`.

5. **`qzero` / `qzero_like` hermiticity** (`qutip/core/operators.py`). Both
   set `isherm=True` unconditionally, but both accept rectangular dims and a
   non-square matrix cannot be Hermitian. `isherm` is now set only when the
   operator is square.

6. **scipy lsoda excess-work message** (`qutip/solver/integrator/scipy_integrator.py`).
   Two adjacent string literals concatenated to "...on this call.Try to
   increase...", dropping the space between sentences. Space restored. (No
   test: asserting exact wording of an error string is brittle and the fix
   is a one-character change.)

Full test suites for every touched module pass (core qobj/operators/superop_reps,
HEOM, integrator, mcsolve), and the new tests fail on unmodified `master` and
pass with these changes. Changed lines are `pycodestyle`-clean.

**Related issues or PRs**

None found. Separate from and independent of my message-fix PR #2930.

**AI Tools Usage Disclosure**

Claude (Anthropic), Claude Code: used to triage an audit of candidate bugs,
execute-verify each one against a source build of qutip (`pip install -e .`
with cython/numpy/scipy), reproduce each failure and confirm the fix, write
the fixes and regression tests, and run the affected test suites. Several
audit candidates were dropped after verification showed they were false
positives (e.g. `dag`/`trans` dims are correctly swapped via `Dimensions`,
`str_env` is a module global, `unit()` forwards `kwargs` correctly,
`Bloch.add_vectors` does append, `set_ylim(0.5,)` is valid). I reviewed and
understand every change and take responsibility for it.


qutip/solver/mcsolve.py
```diff
@@ -927,17 +927,17 @@ def StateFeedback(
         default : Qobj or qutip.core.data.Data, default : None
             Initial value to be used at setup of the system.
 
-        open : bool, default False
-            Set to ``True`` when using the monte carlo solver for open systems.
+        prop : bool, default : False
```
Ericgig: That one is wrong. Mcsolve can evolve open states ([mix of collapse with jump and lindblad dissipators](https://nbviewer.org/urls/qutip.org/qutip-tutorials/tutorials-v5/time-evolution/016_smesolve-inefficient-detection.ipynb),) but cannot be used to compute propagator.

Ericgig (commented):
Overall good.
Nice to have these small issues fixed.
But the one in mcsolve was not a bug.

Not sure about the HEOM one, @pmenczel should be able to confirm.

pmenczel:
Regarding the HEOM one, it looks good to me, but I noticed another problem there that could ideally be fixed as well.

Usually, the result objects in qutip always have all of the attributes, even if the corresponding data is not stored. For example. a `Result` has the attribute `states` even if `store_states` is set to False; in that case, `states` will be `[]`.

Therefore, the condition in [line 394](https://github.com/steps-re/qutip/blob/4bbcff027a6b4b34844f4b220c3fbe6540285670/qutip/solver/heom/bofin_solvers.py#L394) should be removed (so that lines 395-396 are always called). Then, no `AttributeError` can be raised inside `final_ado_state(self)`.

qutip/tests/solver/test_mcsolve.py
```diff
@@ -565,6 +565,21 @@ def test_feedback(func, kind):
     assert np.all(result.expect[0] > 4. - tol)
 
 
+def test_state_feedback():
```
Ericgig: Need to revert that too...

pmenczel:
Thank you. I approve, except for the `mcsolve` thing.

steps-re:
Thanks @Ericgig, @pmenczel. Done:

- **mcsolve**: reverted the `StateFeedback` change entirely (`mcsolve.py` is now byte-identical to master), and just removed `test_state_feedback` that was added alongside it. You're right that mcsolve evolves open states but can't compute a propagator, so it shouldn't have been touched.
- **HEOM**: `final_ado_state` now returns `self._final_ado_state`, and per @pmenczel's note the `store_ados` guard in `_post_init` is dropped so `_final_ado_state`/`ado_states` are always initialized (no `AttributeError`).

Remaining changes are the other correctness fixes (`logm` message, `qdiags` unitary check, `qzero`/`qzero_like` Hermitian, lsoda message). Should be ready for another look.

Ericgig (commented):
You did not ran the tests locally before pushing...
There is still a broken test:
_______________ TestHEOMResult.test_create_ado_states_attribute ________________

self = <qutip.tests.solver.heom.test_bofin_solvers.TestHEOMResult object at 0x7ff309df7250>

    def test_create_ado_states_attribute(self):
        options = fill_options()
        result = HEOMResult(e_ops=[], options=options)
>       assert not hasattr(result, "final_ado_state")
E       AssertionError: assert not True
E        +  where True = hasattr(<HEOMResult\n  Solver: None\n  Number of e_ops: 0\n  State not saved.\n>, 'final_ado_state')

qutip/tests/solver/heom/test_bofin_solvers.py:1797: AssertionError

steps-re:
Yeah, my bad. I'd over-corrected `_post_init` to always allocate those attrs, which is what broke `test_create_ado_states_attribute`. Backed that out, so `_post_init` is identical to master again and only the actual fix stays: `final_ado_state` returns `self._final_ado_state` instead of `self._final_state`. Ran the whole `TestHEOMResult` class this time, all green.

steps-re:
ran it against the pushed commit and `TestHEOMResult` is 7/7 green, full `test_bofin_solvers.py` passes 127/127. `_post_init` is back to identical-with-master now, so the only real change left is the `final_ado_state` return fix. should be good to re-run CI.

Ericgig (approved):

pmenczel (approved):
Sorry, I missed that the test is also wrong. The behavior is not fully consistent now, but it is a very minor issue and outside the original PR scope, so I will go ahead and merge.

PR #2933, opened 2026-07-15
fix(heom): always initialize final_ado_state/ado_states on HEOMResult

steps-re:
Follow-up to #2931. As @pmenczel noted there, qutip result objects expose all attributes even when the corresponding data is not stored (e.g. `Result.states` is `[]` when `store_states=False`), but `HEOMResult` only created `_final_ado_state` and `ado_states` when `store_ados=True`. Accessing `result.final_ado_state` with `store_ados=False` therefore raised `AttributeError` instead of returning `None`.

This removes the guard in `_post_init` so both attributes always exist, updates `test_create_ado_states_attribute` to the always-present convention (it previously asserted the attributes were absent), and adds a regression test covering the populated-result case.

Ran `qutip/tests/solver/heom/test_bofin_solvers.py` locally against a source build: 128 passed. Both updated tests fail on current `master` without the fix.

**AI Tools Usage Disclosure**

Claude (Anthropic), Claude Code: used to implement the change suggested in the #2931 review, update the affected tests, and run the HEOM test suite against a source build. Reviewed before submission.

### DIFF
--- a/qutip/solver/heom/bofin_solvers.py
+++ b/qutip/solver/heom/bofin_solvers.py
@@ -391,9 +391,12 @@ class HEOMResult(Result):
         super()._post_init()
 
         self.store_ados = self.options["store_ados"]
-        if self.store_ados:
-            self._final_ado_state = None
-            self.ado_states = []
+        # Always initialize these attributes so that accessing them does not
+        # raise an AttributeError when store_ados is False (result objects
+        # expose all attributes even when the corresponding data is not
+        # stored, e.g. Result.states when store_states is False).
+        self._final_ado_state = None
+        self.ado_states = []
 
     def _e_op_func(self, e_op):
         """ Convert an e_op into a function ``f(t, ado_state)``. """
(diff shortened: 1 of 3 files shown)


qutip/solver/heom/bofin_solvers.py
```diff
@@ -391,9 +391,12 @@ def _post_init(self):
         super()._post_init()
 
         self.store_ados = self.options["store_ados"]
-        if self.store_ados:
-            self._final_ado_state = None
-            self.ado_states = []
+        # Always initialize these attributes so that accessing them does not
+        # raise an AttributeError when store_ados is False (result objects
+        # expose all attributes even when the corresponding data is not
+        # stored, e.g. Result.states when store_states is False).
```
pmenczel: This belongs in the changelog (where it already is) not in comments. Not every attribute needs a comment explaining why we don't not initialize it. Future readers of the code and comments will not know that these attributes used to be not initialized.

pmenczel (commented):
Thank you for making the change. The PR is mostly fine, but the code comments are a bit weird (typical of LLM). I have added a more detailed note below for one of the comments; similar things apply for the other comments (in the tests).

steps-re:
fair point, thanks. those comments were explaining the change instead of the code. removed all three, including the two in the tests. the why stays in the changelog entry. full heom suite still green locally, 145/145.

pmenczel (approved):
Looks good to me, thank you.

PR #2934, opened 2026-07-16
fix(solver): use two-sided FFT in spectrum_correlation_fft

### LINKED ISSUE #1537: Bug  in function spectrum_correlation_fft or wrong documentation

Otto-Deviant1904:
Fixes #1537

## Summary
`spectrum_correlation_fft` used a one-sided FFT plus `2*Re()`. For a
one-sided exponential correlation that does not converge to a Lorentzian
as well as a full two-sided transform (as already used in
`qutip.core.environment._fft`).

## Changes
- If `tlist` is one-sided (`t >= 0`), build a Hermitian two-sided signal
  with `C(-t) = conj(C(t))`
- Apply a phase-corrected FFT and scale with `dt` (not `2*dt`)
- Document the one-sided input convention
- Add a Lorentzian regression test (peak-to-tail ratio)

## AI Tools Usage Disclosure
Assisted with investigation, implementation, and tests. I reviewed the
numerical change, checked fail→pass on the Lorentzian shape ratio, and
will handle review feedback myself.

Assisted-by: Cursor

Made with [Cursor](https://cursor.com)


qutip/solver/spectrum.py
```diff
@@ -67,33 +67,55 @@ def spectrum_correlation_fft(tlist, y, inverse=False):
...
+    if N < 2:
+        raise ValueError('tlist must contain at least two points for FFT.')
     dt = tlist[1] - tlist[0]
     if not np.allclose(np.diff(tlist), dt * np.ones(N - 1, dtype=float)):
         raise ValueError('tlist must be equally spaced for FFT.')
-    F = (N * scipy.fftpack.ifft(y)) if inverse else scipy.fftpack.fft(y)
-    # calculate the frequencies for the components in F
+
+    # Expand one-sided input (t >= 0) to a Hermitian two-sided correlation.
+    if tlist[0] >= -np.finfo(float).eps:
```
veronikakurth: Could you please explain this condition?

Otto-Deviant1904: That check decides whether `tlist` is one-sided (starts at/near \(t = 0\)) or already two-sided (starts at a clearly negative time).

`tlist[0] >= -np.finfo(float).eps` is a float-safe way of saying "first sample is \(\ge 0\)" without rejecting tiny negatives from rounding (`-1e-16` and friends). When it holds, we mirror with \(C(-t)=C^*(t)\) so the FFT sees a full Hermitian signal. If `tlist` already begins well below zero, we leave it alone and FFT as given.

Otto-Deviant1904:
@veronikakurth friendly ping when you have a chance.

Happy to clarify the tlist[0] / eps check further or adjust the test if that would help.

Otto-Deviant1904:
Rebased onto current master.

qutip/solver/spectrum.py
```diff
@@ -67,33 +67,55 @@ def spectrum_correlation_fft(tlist, y, inverse=False):
...
-        two-sided power spectrum 'S(w)'.
-
+        Angular frequencies ``w`` and the corresponding two-sided power
+        spectrum ``S(w)``.
+
+    Notes
+    -----
+    The transform uses a full two-sided FFT (same convention as the helper
+    in :mod:`qutip.core.environment`), which converges better than a
+    one-sided FFT with a factor-of-two correction.
```
Ericgig: Does that comment need to be user facing?

qutip/solver/spectrum.py
Ericgig: If someone use second units for a process that take femtoseconds, then `max(tlist)` will be of order 1e-15. Then using a constant, undocumented `np.finfo(float).eps` could cause issues...

(Yes it's better to choose units to use numbers of order ~1, but not everyone does it.)

qutip/solver/spectrum.py
```diff
@@ -67,33 +67,55 @@ def spectrum_correlation_fft(tlist, y, inverse=False):
...
-    idx = np.append(idx, np.where(f < 0.0))
-    idx = np.append(idx, np.where(f >= 0.0))
-    return 2 * np.pi * f[idx], 2 * dt * np.real(F[idx])
+    w = 2 * np.pi * f
+    # Phase factor for continuous FT when samples start at tlist[0] (typically
+    # -t_max after the one-sided expansion). Matches environment._fft.
+    F = F * np.exp(1j * w * (-tlist[0]))
+    idx = np.argsort(f)
+    # Two-sided sampling already includes both positive and negative times,
+    # so scale with dt (not 2*dt).
```
Ericgig: This comment makes little sense without looking at the previous version.
Documenting why you did some changes is not wrong, but it should not be in the code itself.

qutip/tests/solver/test_correlation.py
```diff
@@ -88,6 +88,31 @@ def test_spectrum_solver_equivalence_to_es(spectrum):
...
+    i10 = np.argmin(np.abs(ws - 10.0))
+    # Shape check: Lorentzian peak-to-tail ratio (independent of overall scale).
+    # One-sided FFT + 2*Re() historically undershoots this (~68 vs 101).
+    np.testing.assert_allclose(
+        spec[i0] / spec[i10],
+        expected[i0] / expected[i10],
+        rtol=0.05,
+    )
+    scale = expected[i0] / spec[i0]
+    band = np.abs(ws) < 20.0
```
Ericgig: Why not the compare the full array.

qutip/tests/solver/test_correlation.py
```diff
@@ -88,6 +88,31 @@ def test_spectrum_solver_equivalence_to_es(spectrum):
...
+    expected = 2 * gamma / (gamma**2 + ws**2)
+    i0 = np.argmin(np.abs(ws))
+    i10 = np.argmin(np.abs(ws - 10.0))
+    # Shape check: Lorentzian peak-to-tail ratio (independent of overall scale).
+    # One-sided FFT + 2*Re() historically undershoots this (~68 vs 101).
+    np.testing.assert_allclose(
+        spec[i0] / spec[i10],
+        expected[i0] / expected[i10],
+        rtol=0.05,
+    )
```
Ericgig: Why do this test at all?
It should be covered by the full array comparison.

qutip/tests/solver/test_correlation.py
```diff
@@ -88,6 +88,31 @@ def test_spectrum_solver_equivalence_to_es(spectrum):
...
+    scale = expected[i0] / spec[i0]
```
Ericgig: Why is this scale needed?
Is it a complex phase of norm 1?

Otto-Deviant1904:
Thanks @Ericgig — addressed in e67160b:

- Dropped the user-facing Notes block and the changelog-style inline comments
- One-sided detection now uses `tlist[0] > -0.5 * abs(dt)` so tiny physical time units (e.g. femtoseconds) are not compared against machine eps
- Test now does a full-array comparison to the analytical Lorentzian (no ratio check, no scale factor)

Happy to adjust further if anything still looks off.

qutip/solver/spectrum.py
Otto-Deviant1904: Removed — agreed it did not need to be in the public docstring.

qutip/solver/spectrum.py
Otto-Deviant1904: Switched to `tlist[0] > -0.5 * abs(dt)` so the threshold scales with the sampling step instead of machine eps.

qutip/solver/spectrum.py
Otto-Deviant1904: Removed those changelog-style comments.

qutip/tests/solver/test_correlation.py
Otto-Deviant1904: Done — the test now asserts against the full analytical Lorentzian array.

qutip/tests/solver/test_correlation.py
Otto-Deviant1904: Dropped the ratio check; covered by the full-array comparison.

qutip/tests/solver/test_correlation.py
Otto-Deviant1904: Dropped — after the two-sided FFT scaling, the absolute spectrum matches without a free scale factor.

Ericgig:
Sorry, it's good to fix the points I commented, but I asked "why?" explicitly to start a conversation.
Good first issues are there for learning and making connection to new contributors.
We don't feel we are getting to know you as a person and that you are learning much...

PR #2936, opened 2026-07-17
Optimize rouchon method for stochastic solvers

Ericgig:
**Description**
The rouchon method was quite inefficient, adding operators instead of state.
The new method gave a 2x to 4x improvement in my toy models.


veronikakurth:
@Ericgig I wonder if the changes here could be considered a feature :) 
If it's a significant speedup, I think it deserves a bit more attention in the changelogs.

Ericgig:
It's nothing new, no user code will change for this. I still can't recommend the `rouchon` method for the average user...

Not sure it deserve too much attention.

I renamed the changelog file to be a feature.

qutip/solver/sode/rouchon.py
```diff
@@ -118,34 +122,58 @@ def integrate(self, t, copy=True):
         if self._issuper:
             self.state = unstack_columns(self.state)
```
hodgestar, line 123: Could the unstack be done in `set_state` instead?

qutip/solver/sode/rouchon.py
```diff
@@ -118,34 +122,58 @@ def integrate(self, t, copy=True):
...
             self.t += dt
         if self._issuper:
             self.state = stack_columns(self.state)
 
         return self.t, self.state, np.sum(dW, axis=0)
 
+    def get_state(self, copy=True):
+        # self.state can be modified inplace
+        state = state.copy() if copy else state
+        return self.t, state, self.wiener
```
hodgestar:
```suggestion
        state = self.state.copy() if copy else self.state
        return self.t, state, self.wiener
```
`state` appears not to be defined here.

hodgestar (changes requested):

qutip/solver/sode/rouchon.py
Ericgig: Good catch.
The scary part is that tests are passing...

qutip/solver/sode/rouchon.py
Ericgig: The `unstack` can be, but not the stack.
`integrate` and `get_state` don't return the noise in the same format for stochastic solvers...
Probably need to fix that, but in another PR.

qutip/solver/sode/rouchon.py
hodgestar: Should we add a test? Or extend an existing one?

Ericgig: Added a test.

hodgestar: Woot.

hodgestar (approved):
Feels like a nice improvement, even if Rouchon is not yet easy to get good results from.

PR #2937, opened 2026-07-17
qfunc for large states

### LINKED ISSUE #2932: qfunc returns inaccurate results for large cohenret states
### Bug Description

The function ```qfunc()```  cannot accurately calculate the Husimi-Q function of coherent states with large average photon numbers. The plotted distribution is not circularly symmetric in this case.

### Code to Reproduce the Bug

```shell
import numpy as np
from qutip import num, displace, basis, qfunc

dim = 1000
n_op = num(dim)  
xvec = np.linspace(-30, 30, 400)

fig, ax = plt.subplots(layout='constrained')

coherent = displace(dim, alpha=13) * basis(dim, 0)
Q_coherent = qfunc(coherent, xvec, xvec)
ax.contourf(xvec, xvec, Q_coherent, levels=100, cmap='rainbow', vmin=0, vmax=np.max(np.abs(Q_coherent)))
ax.set_aspect('equal')
plt.show()
```

### Code Output

```shell

```

### Expected Behaviour

The Husimi-Q function of coherent states should always be a circularly symmetric Gaussian distribution.

### Your Environment

```shell
QuTiP Version:      5.2.3
Numpy Version:      2.4.2
Scipy Version:      1.17.1
Cython Version:     None
Matplotlib Version: 3.10.8
Python Version:     3.12.12
Number of CPUs:     12
INTEL MKL Ext:      None
```

### Additional Context

_No response_

Ericgig:
**Description**
Fix underflow in qfunc for large dims.

**Related issues or PRs**
fix #2932

### DIFF
--- a/qutip/wigner.py
+++ b/qutip/wigner.py
@@ -227,7 +227,7 @@ def wigner(psi, xvec, yvec=None, method='clenshaw', g=sqrt(2),
 
     W : array (2D)
         Values representing the Wigner function calculated over the specified
-        range [xvec,yvec]. The array is indexed such that W[j, k] corresponds to 
+        range [xvec,yvec]. The array is indexed such that W[j, k] corresponds to
         y-coordinate yvec[j] and x-coordinate xvec[k].
 
     yvex : array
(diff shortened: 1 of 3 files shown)


qutip/wigner.py
```diff
@@ -813,18 +813,32 @@ def __call__(self, state: Qobj):
...
     """
-    ns = np.arange(vector.shape[0])
+    size = min(cutoff, vector.shape[0])
+    ns = np.arange(size)
     out = np.polyval(
-        (0.5*g * vector / np.sqrt(scipy.special.factorial(ns)))[::-1],
+        (0.5 * g * vector[:size] / np.sqrt(scipy.special.factorial(ns)))[::-1],
         alpha_grid.grid,
     )
+    if vector.shape[0] > cutoff:
```
hodgestar, line 856: Could we do this `if` check straight away and not call `scipy.special.factorial` at all if we know it will return infinity?

Ericgig: The factorial approximation is good for large number.
For the first part of the vector, better use the real factorial function.

hodgestar: I meant we always call `out = np.polyval(...)` above, but then if the vector is bigger than the cutoff, we throw the result away and calculate a new one. We could just not do the first calculation when the shape is too big.

Or maybe I am not reading something properly?

Ericgig: It's not thrown away but added to it.
The polyval computation is always used.

hodgestar: Oh yes, I remember what I was thinking. So there is an `out += ...` are the end, but if `scipy.special.factorial(ns)` is `inf`, is `out` not 0 at that point? Maybe `scipy.special.factorial(ns)` is not quite `inf` though?

Just trying to understand what is happening here more fully.

Ericgig: factorial is inf from 171, the cutoff is as 170, so the part using polyval is the one which factorial is not inf and return a valid result.

hodgestar (approved):
Looks good to me now!

PR #2938, opened 2026-07-21
Scipy sparray migration

veronikakurth:
**Description**

**Motivation**

SciPy is deprecating `spmatrix` types in favor of the sparse array API:
https://docs.scipy.org/doc/scipy/dev/roadmap-detailed.html#sparse

**Our migration strategy:**

To make qutip's code future-proof in light of SciPy's migration-to-sparse-array plan, we switch the internal (i.e., not user-facing) API to work with sparse arrays instead of sparse matrices. With these changes, QuTiP uses `sparray`  types by default (`csr_array`, `dia_array`, `coo_array`, etc.)

**Backward compatibility**

We keep user-facing behavior backward compatible for as long as SciPy supports `*_matrix` types. `Qobj.data_as` and `extract` return a SciPy representation of `csr_matrix`/`dia_matrix` for sparse types by default, if not specified otherwise. This is to ensure that the downstream user code does not break prematurely

**List of changes**

The core changes mostly concern the data layer, which currently relies on `scipy.sparse.csr_matrix` and `scipy.sparse.dia_matrix` for backing inside CSR/Dia matrices. There, we make `csr_array` and `dia_array` the default backings. 

1. **Data-layer constructors (`csr.pyx`, `dia.pyx`)**  accept both matrix and array inputs. They build `csr_array`/`dia_array` internally. 
2. **`extract.py` / `Qobj.data_as`** - support `*_array` types, but keep returning `*_matrix` types by default - for backward compatibility.
3. In multiple modules, instead of assuming a concrete type, format-agnostic checks are done: `convert.pyx`, `linalg.py`,`metrics.py`, `qobj.py`, `energy_restricted.py`, `partial_transpose.py` `wigner.py`, `semidefinite.py`, etc.
4. Random-object generation (`random_objects.py`) is switched to return `*_array` types by default.

**Deprecation strategy**

QuTiP will support legacy`*_matrix` types for user-facing methods, namely, `extract` and `Qobj.data_as()`, as long as SciPy supports them. We will not throw deprecation warnings - this we expect to happen on SciPy's side. 

Nevertheless, we strongly recommend our users start adapting to the usage of the `sparray` interface in their code as soon as possible. 

On the level of qutip, the first step would be to explicitly request `*_array` types when working with `extract`/`Qobj.data_as`: for this, `csr_array` or `dia_array` has to be passed as an argument. 

Additional adaptations might be required if the resulting object is used in slicing or mathematical operations. Please refer to SciPy's migration guide for making your code future-proof: 

https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html#recommended-steps-for-migration

**Out-of-scope**
This PR doesn't complete the migration to the `sparray` interface fully. 
But in order to keep this PR more lightweight, the following topics will be tackled separately:
- MKL solver (`_mkl/spsolve.py`, `solver/heom/bofin_solvers.py`). Currently, it is forced to work with `csr_matrix` with 32-bit indices. A proper migration requires a more systematic approach to the low-level code. 
- piqs
- After SciPy's minimum version upgrade:
  - Use `diags_arrays` and `random_array` API

**Related issues or PRs**
#2924

**AI Tools Usage Disclosure**

- Used Claude Sonnet for detecting and analyzing parts of the codebase affected by the migration and for a better understanding of SciPy's API; for applying small but repetitive changes across multiple files.
- Used Claude Opus for troubleshooting MKL-solver-related errors when I got stuck. 

- Fully generated a test case for a forward compatibility check based on monkey patching
- [x] `qutip/tests/core/data/test_scipy_sparse.py` - Review it additionally


qutip/core/data/extract.py
```diff
@@ -39,52 +37,66 @@ def extract_dense(matrix, format=None, copy=True):
...
     """
-    if format not in [None, "scipy_csr", "csr_matrix"]:
+    if format not in [None, "scipy_csr", "csr_matrix", "csr_array"]:
         raise ValueError(
-            "CSR can only be extracted to 'csr_matrix'"
+            "CSR can only be extracted to 'csr_array' or 'csr_matrix'"
         )
-    csr_mat = matrix.as_scipy()
+    scipy_csr = matrix.as_scipy()
+    # TODO: remove upon SciPy deprecation of csr_matrix
```
veronikakurth, line 59: I leave these TODOs intentionally to quickly detect places in the code that require changes upon SciPy's deprecation.

qutip/random_objects.py
```diff
@@ -181,32 +181,19 @@ def _merge_shuffle_blocks(blocks, generator):
...
-            col.append(bcol.ravel())
-            data.append(block.ravel())
-        data = np.hstack(data)
-        row = np.hstack(row)
-        col = np.hstack(col)
-        matrix = sp.coo_matrix((data, (row, col)), shape=(N, N))
+        # Use block_diag to assemble the (row, col, data) triplets
+        bd = sp.block_diag([sp.coo_array(b.T) for b in blocks], format="coo")
+        # Relabel existing indices (~permute)
+        matrix = sp.coo_array((bd.data, (idx[bd.row], idx[bd.col])), shape=(N, N))
```
veronikakurth, line 188: Refactored this function to use `coo_array` since it makes it much simpler.

qutip/_mkl/spsolve.py
```diff
@@ -223,8 +223,8 @@ def mkl_splu(A, perm=None, verbose=False, **kwargs):
 
     Parameters
     ----------
-    A : csr_matrix
-        Sparse input matrix.
+    A : csr_matrix or csr_array
```
veronikakurth: I may leave out this module from this migration stage since switching to sparse array seems to cause tricky issues in the steady state solver in HEOM tests for Python 3.14 and the latest Ubuntu.

qutip/core/data/extract.py
```diff
@@ -1,11 +1,9 @@
 from . import Dense, CSR, Dia
 from .dispatch import Dispatcher as _Dispatcher
+from ._scipy_sparse import (
+    csr_as_matrix, dia_as_matrix, _warn_if_legacy_scipy_input,
+)
 import inspect as _inspect
-try:
-    from scipy.sparse import csr_array
```
veronikakurth, line 5: `csr_array` is supported since SciPy 1.8, and we require a higher version of SciPy; so this check was redundant and also not used.

qutip/core/data/reshape.pyx
```diff
@@ -158,7 +158,7 @@ cpdef list split_columns_dense(Dense matrix, copy=True):
 
 
 cpdef list split_columns_csr(CSR matrix, copy=True):
-    return [CSR(matrix.as_scipy()[:, k], copy=copy)
+    return [CSR(matrix.as_scipy()[:, [k]], copy=copy)
```
veronikakurth, line 161: Without this change, in some tests where `as_scipy` returned 1D values, indexing was faulty. 

Reference: https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html#sparse-migration-shapes-reductions

qutip/core/qobj.py
```diff
@@ -903,16 +903,18 @@ def data_as(self, format: str = None, copy: bool = True) -> Any:
...
 
         copy : bool {False, True}
             Whether to return a copy
 
         Returns
         -------
-        data : numpy.ndarray, scipy.sparse.matrix_csr, etc.
+        data : numpy.ndarray, scipy.sparse.csr(dia)_array, or
+               scipy.sparse.csr(dia)_matrix
             Matrix in the type of the underlying libraries.
```
Ericgig, line 955: Could be cupy array, jax array, cudensity Operator and more with other backend...

qutip/tests/core/data/test_csr.py
```diff
@@ -260,7 +260,7 @@ def test_as_scipy_is_correct_result(self, scipy_csr):
         input.
         """
         data_csr = data.CSR(scipy_csr)
-        assert isinstance(data_csr.as_scipy(), scipy.sparse.csr_matrix)
+        assert scipy.sparse.issparse(data_csr.as_scipy())
```
Ericgig, line 263: Add assert for the format?

qutip/random_objects.py
```diff
@@ -181,32 +181,19 @@ def _merge_shuffle_blocks(blocks, generator):
...
+        end = 0
         for block in blocks:
-            start = end
-            end = end + block.shape[0]
+            start, end = end, end + block.shape[0]
             brow, bcol = np.meshgrid(idx[start:end], idx[start:end])
             matrix[brow, bcol] = block
-    return _data.create(matrix, copy=False)
-
+    return _data.create(matrix, copy=None)
```
Ericgig, line 196:
```suggestion
    return _data.create(matrix, copy=None)

```

qutip/core/data/dia.pyx
```diff
@@ -1,4 +1,5 @@
 #cython: language_level=3
+
```
Ericgig: Why?

qutip/core/data/extract.py
```diff
@@ -105,7 +117,7 @@ def extract_dia(matrix, format=None, copy=True):
 extract.__doc__ =\
     """
     Return the common representation of the data layer object: scipy's
-    ``csr_matrix`` for ``CSR``, numpy array for ``Dense``, Jax's ``Array`` for
+    ``csr_array`` or ``csr_matrix`` (soon-to-be-deprecated; see https://docs.scipy.org/doc/scipy/dev/roadmap-detailed.html#sparse)  for ``CSR``, numpy array for ``Dense``, Jax's ``Array`` for
```
Ericgig: Line too wide. Maybe better in a note block.

qutip/core/data/extract.py
```diff
@@ -39,52 +37,66 @@ def extract_dense(matrix, format=None, copy=True):
...
+    if format not in [None, "scipy_csr", "csr_matrix", "csr_array"]:
         raise ValueError(
-            "CSR can only be extracted to 'csr_matrix'"
+            "CSR can only be extracted to 'csr_array' or 'csr_matrix'"
         )
-    csr_mat = matrix.as_scipy()
+    scipy_csr = matrix.as_scipy()
+    # TODO: remove upon SciPy deprecation of csr_matrix
+    if format == "csr_matrix":
+        _warn_if_legacy_scipy_input("csr")
```
Ericgig: Do we start warning immediately?
We still support from scipy 1.9 and it will only be deprecated when the minimum version becomes 1.18. Can't we let scipy raise the warning until then?

qutip/core/data/_scipy_sparse.py
```diff
@@ -0,0 +1,84 @@
...
+Helper module for the use of ``scipy.sparse`` backing containers in the data layer.
+
+In the light of SciPy's migration plans, ``spmatrix`` types will be replaced by ``sparray`` types.
+We adopt to this migration strategy by ensuring that qutip's data layer always builds ``sparray`` containers, and a legacy matrix is only produced when a user explicitly requests one at such public methods as ``extract`` / ``Qobj.data_as``.
+
+In our migration strategy, we want to ensure (hence this module) that:
+
+* we detect SciPy's CSR/Dia objects regardless of the exact implementation (csr_matrix/csr_array or dia_matrix/dia_array).
+* we correctly convert array to matrix views (upon user request). This is to ensure that before SciPy officially deprecates sparse matrix, users of qutip can still rely on its interface in their downstream code.
+"""
```
Ericgig, line 24: Pep8.

qutip/_mkl/spsolve.py
Ericgig: Do you know what is causing the issues?

qutip/core/data/dia.pyx
veronikakurth: It happened upon some reorganisation of imports; is unnecessary, will delete the change.

qutip/random_objects.py
veronikakurth: I will update the formatting, thanks!

Just for context on why I'm changing the `copy`'s value from `False` to `None`. I came across one error that was tricky to investigate. After I switched to sparse arrays in the `_merge_shuffle_blocks` function, I started to get a ValueError originating at a NumPy array creation in CSR (for column indices; `ValueError: Unable to avoid copy while creating an array as requested`). The error was coming from `test_energy_restricted.py::test_enr_ptrace`, where a random density matrix is created.

Turned out that some randomly generated matrices for that test (or rather sparse arrays) had 64-bit integer indices (which is a bit unclear to me why because if I understand the default behavior of SciPy correctly, the index type will be inferred based on the data; the main criterion should be the amount of nnz values: > `2**31`, which is never the case for qutip's tests). 

And since `copy` was `False` (being prohibitive semantics, in comparison to NumPy 1.x, where tests would not fail because copying would be performed if memory-sharing constraints are not met), and the index type of the passed SciPy object didn't match the expectations on `idxint_dtype` ([here](https://github.com/qutip/qutip/blob/master/qutip/core/data/csr.pyx#L116)).

So I aligned the flag to how this data creation would behave for NumPy 1.x, meaning that a copy will be created if it's not possible to create a SciPy representation otherwise. I don't know, though, if there are any reasons for us to follow a stricter strategy and disallow copying in places like this.  

Here's an example of a random DM for which the exception was raised in CSR's constructor:

- `matrix` -> a COOrdinate sparse array of dtype 'complex128' with 32 stored elements and shape (22, 22)
- `csr_m = matrix.tocsr()`
- `csr_m.indptr.dtype` -> `dtype('int64')`
- `CSR(csr_m, copy=False)` -> `ValueError`
  - With `copy=True` -> no error.

P.S. I think this was a good stress test on checking how qutip would handle SciPy's sparse objects with 64-bit indices. And my answer for now is: it will handle it without errors only if qutip was built with 64-bit index support. Otherwise, if `copy=True`, CSR/Dia will create a downcasted backing of a scipy matrix/array, or if `copy=False`, it will throw an error, which I think would be better to surround with exception handling, but this is separate from this PR. Given [this discussion](https://github.com/scipy/scipy/issues/16774), I think it would be important to ensure we are also ready to always work with 64-bit indices in qutip.

qutip/tests/core/data/test_csr.py
veronikakurth: A good idea, thanks! Added the same for test_dia.py.

qutip/core/data/extract.py
veronikakurth: I see your point; it might indeed be too early to start the warnings, especially if users are far from using SciPy 1.18, where they would anyway start receiving warnings from SciPy itself. 

We can wait with deprecation warnings. However, I wonder what would be a good way to manage the timeline for such changes, in order not to miss a fitting moment to add a deprecation warning from our side. I guess the tests for `extract` using `csr_matrix`/`dia_matrix` and failing upon SciPy's new deprecation warnings will be a reminder... Don't know if the best one.

qutip/tests/core/test_qobj.py
```diff
@@ -1304,8 +1304,13 @@ def test_groundstate():
 def test_data_as():
     qobj = qutip.qeye(2, dtype="CSR")
 
+    # Explicit format names select the scipy container type.
     assert scipy.sparse.isspmatrix_csr(qobj.data_as("csr_matrix"))
```
veronikakurth, line 1415: At some point, when SciPy deprecates `csr_matrix`, we will not be able to run this test for certain versions (preceded by a warning). 
I'm concerned that then we would need to add version-based branching for the test (or separate tests for different SciPy versions?) and `data_as`/`extract` to keep them backward compatible with older versions of SciPy that will still be supported for a while - at least before the minimum supported version would become 1.18.

qutip/_mkl/spsolve.py
veronikakurth: Unfortunately, not yet. I also couldn't reproduce it locally. The error in the CI is quite obscure and points to compiled MKL code down the stack. I'll try to trace operations on the sparse matrix at a higher level and see if I can detect a problem.

veronikakurth:
@Ericgig, a brief remark regarding the milestone: I would argue we should target this PR for minor 5.4.0 and not a patch version. With this PR, backward-compatible support for sparse arrays would be a _new_ functionality.

qutip/core/qobj.py
```diff
@@ -903,16 +903,18 @@ def data_as(self, format: str = None, copy: bool = True) -> Any:
         Parameters
         ----------
         format : str, default: None
-            Type of the output, "ndarray" for ``Dense``, "csr_matrix" for
-            ``CSR``. A ValueError will be raised if the format is not
-            supported.
+            Type of the output: "ndarray" for ``Dense``; "csr_array"/
+            "csr_matrix" for ``CSR``; "dia_array"/"dia_matrix" for ``Dia``.
+            When ``None``, sparse types return their native ``sparray`` backing.
```
veronikakurth: This is actually wrong if we want to have backward compatibility for the user code at this stage. We should return a sparse matrix by default as long as the minimum supported version of SciPy allows this.

qutip/tests/core/test_qobj.py
```diff
@@ -1304,8 +1304,18 @@ def test_groundstate():
...
+    # CSR data type despite sparse array based internal backing
+    # of CSR (as long as the minimum SciPy version supports it)
+    assert scipy.sparse.isspmatrix_csr(qobj.data_as())
     assert scipy.sparse.isspmatrix_csr(qobj.data_as("csr_matrix"))
     assert scipy.sparse.isspmatrix_csr(qobj.data_as(copy=False))
+    # Test array representation
+    csr_arr = qobj.data_as("csr_array")
+    assert isinstance(csr_arr, scipy.sparse.csr_array)
+    assert scipy.sparse.issparse(csr_arr) and csr_arr.format == "csr"
+    assert not scipy.sparse.isspmatrix_csr(csr_arr)
```
Ericgig: Aren't these 3 checks are redundant?
Is there any situation where the `isinstance` check passes but the other don't.

qutip/random_objects.py
```diff
@@ -502,11 +489,11 @@ def rand_ket(
     elif distribution == "haar":
         ket = rand_unitary(N, density, "haar", seed=generator) @ basis(N, 0)
     else:
-        X = scipy.sparse.rand(N, 1, density, format='csr',
+        X = scipy.sparse.random_array((N, 1), density=density, format='csr',
```
Ericgig: This exist from scipy 1.13. 
We are officially OK for requiring it.
But that would require updating the tests matrix and maybe bumping numpy min version.
I would prefer that change to be included in 5.4 over a micro release.

How do you see this.
- Make this a non user facing change PR and make another PR for 5.4 that change requirements?
- Have this be part of 5.4 milestone instead of 5.3.1.
- ???

veronikakurth: Thanks for the catch!
Sorry, I didn't notice that it is required by a higher SciPy. But speaking of which: we should indeed raise the minimum version. The recommended minimum version by SPEC 0 is currently 1.15 -- I think we could go for it. 
I agree that upgrading the minimum version should be a part of the minor release, though, not a patch one.

And speaking of this PR in general: I'd be fine to make it a part of the micro release, but I'm still more inclined to have it as a part of 5.4. Not so much because I want to have more time 🙃 , but rather because this entire internal migration to a new data type _feels_ to deserve more attention. With this, we also signal to the community that we are on the same page with the current roadmap of SciPy/NumPy, as other scientific Python projects do.

But speaking of the more objective arguments w.r.t. semantic versioning: we actually introduce, even if not so impressive, a new functionality: a user can get a `csr_array` representation upon calling `data_as` with the right argument. It might be well noticed by users who already adopted sparse arrays a while ago in their code. 

Let's decide by the release time when we exactly want to merge it, based on the general picture/summary. :) I will, in any case, pull out `random_array` + SciPy version upgrade from here and MKL solver adaptations.

qutip/tests/core/test_qobj.py
veronikakurth: This was indeed overprotective. Changed, thanks!

qutip/_mkl/spsolve.py
```diff
@@ -237,7 +237,7 @@ def mkl_splu(A, perm=None, verbose=False, **kwargs):
         solve method for solving with a given RHS vector.
 
     """
-    if not sp.isspmatrix_csr(A):
+    if not (sp.issparse(A) and A.format == "csr"):
```
veronikakurth, line 240: I added this check to let it go through the type error and reach the segmentation error, which has been very annoying.

veronikakurth: I keep getting a segfault error in the test `qutip/tests/solver/heom/test_bofin_solvers.py::TestHEOMSolver::test_pure_dephasing_model_bosonic_bath`. 
Debugging led me to `hsolver.steady_state()` from `bofin_solvers.py`. There, `mkl_spsolve` fails with a segfault. I think the culprit is matrix `A` which is a sparse array, although `bofin_solvers.py` should still operate on sparse matrices; somehow, in the test, sparse arrays come through. 

When I was checking matrix `A` in `mkl_spsolve`, I checked the following index types:
`A.indices.dtype` -> `dtype('int64')`
`A.indptr.dtype` -> `dtype('int64')`

On master, it's `int32` and, of course, the type is a sparse matrix. 

I was playing around with `np.int32 -> np.int64` in `mkl_spsolve` in the memory allocation parts, but it didn't help. 
My actual goal is to remove changes to anything MKL-related from this PR, which is probably, at this point, about fixing the tests, but I cannot find a way to do so.

Ericgig: Could you convert array to matrix at the start of `mkl_spsolve`?

It could already be wrong and the error is just hidden.
We had `eigh` failing with segfault from the second call at one time. The first call would always work perfectly. Never found the causes, `settings.eigh_unsafe` exist because of that.


Eventually we would need to look at https://github.com/haasad/PyPardiso or https://github.com/simpeg/pydiso.
If licenses are matching, we could just take the part we need, not sure how maintained they are...

veronikakurth: I will try it out, thanks for the idea. 🙌 

And thanks for the context, that’s good to know!

veronikakurth: Both libraries seem to be maintained with a similar frequency (not so often), but the latter one has newer activity. 
They also both assume csr_matrix for the sparse format, but sparsity checks look general enough to be compatible with sparse arrays. 
I will try them out.
But first things first: for now, getting sparse arrays out of the way in the MKL solver…

Ericgig: I think the important part is they don't assume type, but check them : https://github.com/simpeg/pydiso/blob/b6a78544db57e77dce295449acf28e4681ea0350/pydiso/_mkl_solver.in.pyx#L141, while we use [`ctypes`](https://docs.python.org/3/library/ctypes.html) and don't check the mkl version or compilation option and just assume everything is default and untouched for years...

veronikakurth: 🎉 It helped to explicitly convert the matrix to `csr_matrix` type before passing it to `mkl_spsolve`: 

https://github.com/qutip/qutip/pull/2938/changes#diff-ec29eb2b20157b4905fde0de947da42c0319e3d21dbfdbb44251e302553de639R1027

No segfault anymore. 

P.S. It seems that the way qutip interfaces with Pardiso is not ready to tackle `csr_array` with its 64-bit indices... 
I will tackle the topic of MKL separately, as discussed.

Ericgig (approved):

Ericgig:
Please check the rtd build before merging.

veronikakurth:
> Please check the rtd build before merging.

Thank you for your review, Eric.
I will have a look into RTD: retriggering the build didn't help, so I guess there might be a deeper issue hiding. Thanks!

veronikakurth:
@Ericgig docs should build now!
Found this warning when building docs locally:
`docstring of qutip.core.qobj.Qobj.data_as:24: WARNING: Block quote ends without a blank line; unexpected unindent. [docutils]`
Have to be more attentive to docstrings in the future.

PR #2886, opened 2026-04-17
Unified caching

QubitGoblin31:
**Description**
Cached isunitary, ishp, iscp, istp, iscptp, took off isattr for qutip_jax

**Related issues or PRs**
Pr: #2838 
Issue: #453

PR #2940, opened 2026-07-22
Fix solver output isherm for non-Hermiticity-preserving dynamics

### LINKED ISSUE #2410: Hermitian attribute (isherm) can be wrong on mesolve output states
### Bug Description

In some cases, the states returned by `mesolve` report to be Hermitian (`isherm == True`), but they are not. The two examples below show this happening with a non-Hermitian Hamiltonian, and with a custom Liouvillian that does not preserve Hermiticity.

### Code to Reproduce the Bug

```shell
import qutip as qt
H = qt.sigmaz() + 1j * qt.sigmax()
initial = qt.basis(2, 1)
res1 = qt.mesolve(H, qt.ket2dm(initial), [0, 1])
print(res1.final_state)

L = qt.liouvillian(qt.sigmaz()) + 1j * qt.lindblad_dissipator(qt.sigmap())
res2 = qt.mesolve(L, initial, [0, 1])
print(res2.final_state)
```

### Code Output

```shell
Qobj data =
[[-0.33333333+0.j          0.33333333-0.33333333j]
 [-0.33333333-0.33333333j  0.66666667+0.j        ]]

Qobj data =
[[0.23467737+0.42957377j 0.        +0.j        ]
 [0.        +0.j         0.27582588-0.42957377j]]
```

### Expected Behaviour

For both states, `isherm` should be `False`.

### Your Environment

```shell
QuTiP Version:      5.0.0.dev0+5fabdd3
Numpy Version:      1.26.3
Scipy Version:      1.12.0
Cython Version:     3.0.8
Matplotlib Version: 3.8.2
Python Version:     3.12.1
Number of CPUs:     20
BLAS Info:          Generic
INTEL MKL Ext:      False
```

### Additional Context

The source of the issue is that `_prepare_state` ([link](https://github.com/qutip/qutip/blob/master/qutip/solver/solver_base.py#L86)) and `_restore_state` ([link](https://github.com/qutip/qutip/blob/master/qutip/solver/solver_base.py#L99)) of the `solver_base` module assume that the dynamics preserves the Hermiticity of the state. The `isherm` attribute of the initial state is saved and copied into the output state.

We could try to detect whether the Hamiltonian is Hermitian / whether the Liouvillian is Hermiticity-preserving, but @Ericgig mentioned in https://github.com/qutip/qutip/pull/2408 that this solution could be difficult with time-dependent problems.

sankalpsthakur:
**Description**
Stop copying the initial state's `isherm` onto solver output. Cache it only when the RHS is known to preserve Hermiticity (standard ME construction; SME/SSE with operator `H`).

**Related issues or PRs**
Fixes #2410

Changelog: `doc/changes/2940.bugfix`. The earlier `doc/contributors.rst` docs flake is in #2946.

**AI Tools Usage Disclosure**
  - Grok / Codex: helped draft the solver `isherm` path and this description. I reviewed the change and the tests.


qutip/solver/solver_base.py
```diff
@@ -103,12 +103,12 @@ def _prepare_state(self, state):
             raise TypeError(f"incompatible dimensions {self.rhs.dims}"
                             f" and {state.dims}")
 
+        # Do not copy isherm from the initial state. Evolution under a
+        # non-Hermitian Hamiltonian or non-Hermiticity-preserving superoperator
+        # can break Hermiticity of density operators; Qobj.isherm recomputes
+        # and caches from the actual data when needed (see issue #2410).
```
Ericgig:
```suggestion
        # TODO: Add back setting isherm flag checking both the initial state and 
        # liouvillian's ishp flag.

**Review by Ericgig — COMMENTED** [review]
Checking the hermiticity is quite slow: it need to look at every values of the state not in order, (and it's probably not as optimized than other operations.) It's needed when computing the expectation values or printing the states, etc. In large state, it can be 100x slower to compute the hermiticity than the to compute the expectation value with a sparse state.

You can try with:
```
for N in [10, 30, 100, 300, 1000, 3000, 10000]:
    m = np.random.randn(N, N)
    a = qt.data.Dense(m + m.T)
    op = qt.num(N)
    print(N)
    %timeit qt.core.data.isherm_dense(a)
    a = qt.Qobj(a)
    %timeit qt.expect(op, a)
```


While it's better not to have the flag wrong, it's quite the drag for the usual case where the density matrices are all hermitian.
Better solutions would be to check if the Liouvillian preserve hermiticity, or optimize it's computation, or add an option to the solver to include the flag...

**Comment by sankalpsthakur** [comment]
Fair point on the isherm cost — added a TODO in solver_base.py. Correctness fix stays; happy to follow up on a cheaper path (Liouvillian check or solver option) separately.

File: qutip/solver/solver_base.py (hunk 1 above)
Comment (sankalpsthakur) [reply qutip/solver/solver_base.py]: Implemented in 0851666f8. Rather than calling `Qobj.ishp` on the assembled superoperator, `MESolver` now records whether the RHS was constructed entirely from Hamiltonian and collapse operators; that construction is Hermiticity-preserving, including for time-dependent inputs. The cached `isherm=True` metadata is retained only when that structural guarantee and a Hermitian initial state are both present. User-supplied superoperators remain uncached, so non-HP dynamics still recompute from the evolved data. I avoided a direct `ishp` probe because it performs a Choi conversion and was materially more expensive than the state check in a local benchmark. Regressions cover constant standard dynamics, time-dependent standard dynamics, and a time-dependent non-HP superoperator. Validation: `test_mesolve.py` + `test_sesolve.py`, 247 passed.

**Comment by sankalpsthakur** [comment]
RTD was failing on a GitHub API rate-limit 403 in the contributors plot (`fail_on_warning`) — fixed in f53478b0e.

File: doc/contributors.rst [hunk 2]
```diff
@@ -37,11 +37,16 @@ Developers
...
+    # GitHub anonymous API rate limits can 403 on shared RTD IPs; catch so
+    # fail_on_warning docs builds do not flake.
+    try:
+        url_object = urllib.request.urlopen(LINK_CONTRIBUTORS)
+        list_contributors = json.loads(url_object.read())
+        qutip_contributors = [element["login"] for element in list_contributors]
+        qutip_contributors = [s.lower() for s in qutip_contributors]
+        text = " ".join(qutip_contributors)
+    except Exception:
+        text = "qutip developers and contributors"
```
Ericgig: Unrelated to this PR...

Ericgig (commented):
The `contributors.rst` should be in it's own PR.

I like the solution, fast for normal cases and right for other.
Could you also add `_rhs_preserves_hermiticity` to SMESolver?
Just need to check `H`, the flag can be set for SSESolver too, the ket state will set isherm to false.

sankalpsthakur:
@Ericgig Split the contributors.rst change into #2946. Added `_rhs_preserves_hermiticity` for SMESolver/SSESolver (operator `H`; SSE kets still leave `isherm` unset).

Ericgig (approved):
That works for me.
Thank you.

pmenczel (commented):
It looks good to me, but I wanted to comment: if [this PR](https://github.com/qutip/qutip/pull/2886) was merged, then we could add something like
```python
or (H.issuper and H._ishp)
```
to the definition of `_rhs_preserves_hermiticity`. That PR has been added to the 5.3.1 milestone, so perhaps it will be merged soon?

Ericgig:
> It looks good to me, but I wanted to comment: if [this PR](https://github.com/qutip/qutip/pull/2886) was merged, then we could add something like
> 
> ```python
> or (H.issuper and H._ishp)
> ```
> 
> to the definition of `_rhs_preserves_hermiticity`. That PR has been added to the 5.3.1 milestone, so perhaps it will be merged soon?

That would be even better, but you would need to move the check before converting to `QobjEvo` as only the `Qobj` will have the flag.
I will wait a little more to merge #2886 to see if the contributor respond or other maintainers have comments. If not, I will merge early next week.

If you don't have the interest or time for it, don't feel forced.

sankalpsthakur:
CI red was coveralls.io HTTP 500 after a full test pass (27334 passed). Rebased onto master — please approve the new workflow runs (or merge ignoring the coverage-upload flake). Thanks!

Ericgig:
We know about coveralls.io failure, no need to force push for that.

I merged #2886, do you want to try Paul's suggestion or leave it as is?

sankalpsthakur:
Thanks @Ericgig — implemented Paul's suggestion now that #2886 is in.

`MESolver` treats a constant superoperator `H` as Hermiticity-preserving when `isinstance(H, Qobj) and H.ishp` (checked before any `QobjEvo` conversion). Time-dependent superoperator `QobjEvo` inputs still skip the fast path. Added a regression test for an `ishp` Liouvillian.

Ericgig (approved):
Looks good, thank you.

sankalpsthakur:
qutip-jax CI red: their Qobj subclass has no `_ishp` attribute, so `H.ishp` raised. Guarded with `getattr(H, 'ishp', False)` — subclasses without the attr take the safe path.

Ericgig:
qutip-jax's Qobj is the same as qutip, it has `ishp`, but no `_ishp`, so `getattr(H, 'ishp', False)` guard nothing.
`_ishp` is new and just needed a patch: https://github.com/qutip/qutip-jax/pull/104, you just have to wait.

Please relax with trying to having everything green. We know about flaky Coverall and doc build, we know that the PR we just merged has side effects. We know there are some other tests using random number that have a 0.1% fail rate. Can restart tests ourselves if it bother us.

Merging is slow in part because of timezone of the maintainer team. The active members are 13h apart, so it's often one exchange per day.

qutip/solver/mesolve.py
```diff
@@ -231,6 +231,19 @@ def __init__(
...
         self._num_collapse = len(c_ops)
+        # The Liouvillian assembled from Hamiltonian and collapse operators
+        # preserves Hermiticity by construction.  User-supplied superoperators
+        # are safe only when the Qobj is known to be Hermiticity-preserving
+        # (``Qobj.ishp``).  Check before any ``QobjEvo`` conversion — only
+        # ``Qobj`` carries that flag.  A time-dependent superoperator
+        # ``QobjEvo`` cannot be certified from a single-time ``ishp`` probe.
+        h_preserves = (
+            not H.issuper
+            or (isinstance(H, Qobj) and getattr(H, "ishp", False))
```
pmenczel:
```suggestion
            or (isinstance(H, Qobj) and H._ishp)
```
To my understanding: we do not want to access `ishp` here, which computes the Choi matrix and is slow. Only if `H` is already confirmed Hermitian-preserving (`_ishp`), we want to use that information. (And the `getattr` doesn't do anything as Eric has pointed out.)

pmenczel (changes requested):

qutip/solver/mesolve.py
sankalpsthakur: Updated to `H._ishp` so we only use the cached Hermitian-preserving flag and avoid the slow `ishp` Choi path.

sankalpsthakur:
Addressed @pmenczel’s review: `mesolve` now uses `H._ishp` (cached) instead of `getattr(H, \"ishp\", False)`.

qutip/solver/mesolve.py
pmenczel: Unfortunately things are a bit more tricky. Sorry for forgetting about this point in my previous comment (Eric had mentioned it earlier). With your current code, everything should be working if a user creates `MESolver` themselves. However, users typically use the `mesolve` function instead of creating the solver object manually, and there the user-provided `H` is always converted to a `QobjEvo`: https://github.com/sankalpsthakur/qutip/blob/656ac2af8f11eb0e271c06aa8ea74444dc9a7440/qutip/solver/mesolve.py#L148

If you would like to fix this properly, the two best options that I can see are (a) adding something like `or (H.isconstant and H(0)._ishp)`, (b) changing `mesolve` so that the `QobjEvo` conversion is only performed if necessary, (c) somehow moving the check before the conversion.

sankalpsthakur:
went with option (a) in efb1b24c — guarded on QobjEvo since Qobj has no isconstant. added a regression test that fails without it. solver and core suites green locally. ptal

pmenczel:
Thank you for making the change. However, what happened in your [last two commits](https://github.com/qutip/qutip/pull/2940/changes/efb1b24c28e4b7c32f86e56d8b71e5ad6d8adb51..b2b9a2ca8cb64071683582ba59f77fab131ae54c)? Why would you undo half of your changes? If you revert to [efb1b24](https://github.com/qutip/qutip/pull/2940/commits/efb1b24c28e4b7c32f86e56d8b71e5ad6d8adb51) and if you remove the `test_mesolve_preserves_ishp_flag_through_qobjevo_conversion` and `test_mesolve_does_not_certify_time_dependent_superoperator` tests added there, I will accept the PR.

(These two tests are (a) too specific to the current implementation, (b) wrong, since they claim to test behavior of mesolve but do not call mesolve at all, (c) unnecessary, since these points are already tested in the other tests that you have added. Hence I suggest to just remove them.)

sankalpsthakur:
> Thank you for making the change. However, what happened in your [last two commits](https://github.com/qutip/qutip/pull/2940/changes/efb1b24c28e4b7c32f86e56d8b71e5ad6d8adb51..b2b9a2ca8cb64071683582ba59f77fab131ae54c)? Why would you undo half of your changes? If you revert to [efb1b24](https://github.com/qutip/qutip/pull/2940/commits/efb1b24c28e4b7c32f86e56d8b71e5ad6d8adb51) and if you remove the `test_mesolve_preserves_ishp_flag_through_qobjevo_conversion` and `test_mesolve_does_not_certify_time_dependent_superoperator` tests added there, I will accept the PR.
> 
> 
> 
> (These two tests are (a) too specific to the current implementation, (b) wrong, since they claim to test behavior of mesolve but do not call mesolve at all, (c) unnecessary, since these points are already tested in the other tests that you have added. Hence I suggest to just remove them.)

Got it. Done

veronikakurth:
My two cents here with respect to documenting the `isherm` property: Do I understand correctly that if it's `False`, it doesn't necessarily imply that it's a non-Hermitian operator, but could also mean that _we cannot know_ without more precise checks? If `False` has this ambiguity, do you think it would make sense to document this semantics for the user a bit more precisely? @pmenczel 
Sorry if I misunderstood this.

pmenczel (approved):
Thank you, this looks good to me now.

@veronikakurth No, if `isherm` is False, that is supposed to mean that it is a non-Hermitian operator. If Hermitian-ness is unknown, then `_isherm` is None, and upon calling `Qobj.isherm` it will be checked whether it is Hermitian or not.  
Before this PR, we always set `isherm` to either True or False based on a guess, which could be wrong sometimes. With this PR, we set it to True if we are certain about it. Otherwise, we don't set it (that is, we keep `_isherm` as None).

veronikakurth:
> Thank you, this looks good to me now.
> 
> @veronikakurth No, if `isherm` is False, that is supposed to mean that it is a non-Hermitian operator. If Hermitian-ness is unknown, then `_isherm` is None, and upon calling `Qobj.isherm` it will be checked whether it is Hermitian or not. Before this PR, we always set `isherm` to either True or False based on a guess, which could be wrong sometimes. With this PR, we set it to True if we are certain about it. Otherwise, we don't set it (that is, we keep `_isherm` as None).

I see, thank you very much for the explanation!

veronikakurth:
@sankalpsthakur, thank you. Please fill out the PR template as expected for contributions.

sankalpsthakur:
> @sankalpsthakur, thank you. Please fill out the PR template as expected for contributions. 

Where do I find the template

Ericgig:
Look at [pull_request_template](https://github.com/qutip/qutip/blob/master/.github/pull_request_template.md) and the [contribution guidelines](https://github.com/qutip/qutip/blame/master/CONTRIBUTING.md).

doc/contributors.rst
sankalpsthakur: Moved that contributors.rst change to #2946.

qutip/solver/mesolve.py
sankalpsthakur: That's in the later commits. mesolve now checks H before the QobjEvo conversion.

Ericgig:
Thank you for adding the AI disclamer.
Why did you force push (without cleaning the commit messages) ?
The "Co-authored-by: Cursor" is still there...
Did you force-push for another reason?

sankalpsthakur:
force-pushed to rebase, missed stripping the cursor co-author trailer on 4 commits. rewrote messages only (no file diff) in 3a9bfa2, trailer's gone now.

veronikakurth:
@sankalpsthakur, thank you. 

For the future, please make sure you follow **all** the contribution guidelines before PR submission. This includes filling out the AI tools usage statement and having no ambiguity on whom belongs the authorship of the code. As stated in our [contribution guide](https://github.com/qutip/qutip/blob/master/CONTRIBUTING.md#3-copyright--legal), we do not allow "Co-authored by [LLM]" trailing messages due to copyright reasons.

veronikakurth:
@pmenczel, could you please have a (hopefully last) look at the changes? Because of the rebasing & force-pushing, we cannot be 100% sure that the changes are as before. I looked through them, and they make sense and seem to be as before (i.e., addressing the review); I also compared them to the local version of the branch, but it was a bit outdated, so I cannot rely on this comparison. 
Then I think we can give it a go.

pmenczel (approved):
Seems okay!

qutip/solver/solver_base.py
```diff
@@ -105,11 +105,13 @@ def _prepare_state(self, state):
...
         self._state_metadata = {
             'dims': state._dims,
-            # This is herm flag take for granted that the liouvillian keep
-            # hermiticity.  But we do not check user passed super operator for
-            # anything other than dimensions.
-            'isherm': not (self.rhs._dims == state._dims) and state._isherm,
         }
+        if (
+            self.rhs.issuper
+            and state.isherm
```
Ericgig, line 112: Computing propagator from mesolve directly will wrongly flag them as hermitian:
```
qutip.mesolve(
    H=qutip.num(2), 
    rho0=qutip.qeye([[2],[2]]),
    tlist=[0, 1], 
    c_ops=[qutip.destroy(2)]
).final_state`
```
```
Quantum object: dims=[[[2], [2]], [[2], [2]]], shape=(4, 4), type='super', dtype=Dense, isherm=True
Qobj data =
[[1.        +0.j         0.        +0.j         0.        +0.j   0.63212038+0.j        ]
 [0.        +0.j         0.32771053-0.51037797j 0.        +0.j    0.        +0.j        ]
 [0.        +0.j         0.        +0.j         0.32771053+0.51037797j  0.        +0.j        ]
 [0.        +0.j         0.        +0.j         0.        +0.j   0.36787962+0.j        ]]
```
The check if the state is a propagator should be kept:
```suggestion
            self.rhs.issuper
            and not (self.rhs._dims == state._dims)
            and state.isherm
```

Ericgig: Other than that, everything looks fine.

PR #2941, opened 2026-07-22
Clarify dims error for tuple outer container (QuTiP 5.3)

### LINKED ISSUE #2915: Breaking change with Dimensions in 5.3.0
### Bug Description

I think this was a breaking change: d8f32b57ab16740726c90720d752bd17c1cbe392 
Previously, specifying `dims` as lists to `Qobj` used to work. Now it doesn't, but also seems that we can't pass any form of `dims` to `Qobj` as the underlying `Dimensions` class has changed.

### Code to Reproduce the Bug

```shell
import numpy as np
from qutip import Qobj
rho = np.array([
    [0.9, 0.0],
    [0, 0.1]
])
Qobj(rho, dims=([2], [2]))
# Qobj(rho, dims=((2,), (2,)))  # Uncomment to get past the first error
```

### Code Output

```shell
403     raise ValueError(f"Unexpected keyword arguments: {proc_kwargs}")
    405 proc_args = tuple([
    406     tuple(arg) if isinstance(arg, list) else arg
    407     for arg in proc_args
    408 ])
--> 409 if proc_args not in cls._stored_dims:
    410     instance = cls.__new__(cls)
    411     instance.__init__(*proc_args)

TypeError: unhashable type: 'list'

--- If the other line is uncommented

    409 if proc_args not in cls._stored_dims:
    410     instance = cls.__new__(cls)
--> 411     instance.__init__(*proc_args)
    412     cls._stored_dims[proc_args] = instance
    413 return cls._stored_dims[proc_args]

TypeError: Dimensions.__init__() missing 1 required positional argument: 'to_'
```

### Expected Behaviour

Output (v5.2.3 produces this):
Image

 

### Your Environment

```shell
QuTiP Version:      5.3.0
Numpy Version:      2.4.6
Scipy Version:      1.17.1
Cython Version:     None
Matplotlib Version: 3.10.9
Python Version:     3.12.6
Number of CPUs:     12
INTEL MKL Ext:      None
```

### Additional Context

_No response_

sankalpsthakur:
## What / Why

After QuTiP 5.3, tuples inside `dims` denote direct-sum spaces. The outer dims container for `Qobj` must therefore be a **list**, e.g. `[[2], [2]]` or `[(2, 3), (2, 3)]`.

Users who still write the pre-5.3 form `dims=([2], [2])` (a length-2 **tuple**) hit an opaque:

```text
TypeError: unhashable type: 'list'
```

or a missing-argument error for related tuple forms. This PR:

1. Detects a length-2 tuple passed as the outer dims argument in `Dimensions._process_args` and raises a clear `TypeError` with the correct list form and a concrete suggestion.
2. Documents the list-outer / tuple-inner convention in the `Qobj` constructor docstring.

Fixes #2915

## Why this matters

The 5.3 `Dimensions` change is correct but breaks common pre-5.3 call patterns. A targeted error message saves debugging time without changing valid list-outer usage.

## Related

- Breaking-change report: #2915

## Validation

```bash
python -m pytest qutip/tests/core/test_dimensions.py::test_tuple_outer_dims_error_message -q
python -m pytest qutip/tests/core/test_dimensions.py -q
```

All passed locally (294 tests).

Manual checks:

```text
dims=([2], [2])  -> TypeError: ... Use dims=[[2], [2]] instead ...
dims=[[2], [2]]  -> OK
dims=[(2, 3), (2, 3)]  -> OK (direct-sum spaces)
```

## AI Tools Usage Disclosure

OpenAI Codex: Used to inspect the reported failure path, draft the targeted error handling, tests, and documentation, and review the resulting diff. I reproduced the issue, reviewed the changes, and ran the validation above.


Ericgig (approved):
Thank you.
I personally feel it's too much, but expected some issues with using tuple for SumSpace and if it help someone, I don't see many reason to refuse it.

veronikakurth:
I tested how the exception handling works when running the code-to-reproduce from the original issue; it looks good and actionable to me too. @pmenczel, maybe you could have a brief look too, as the main author of direct sums functionality. :)
@sankalpsthakur, thank you for your contribution. Please fill out the AI statement.

pmenczel:
Sorry for not seeing this earlier. To be honest, I don't think disallowing the `dims=([2], [2])` convention was intentional, I just wasn't aware of it. It probably would have been okay to just accept `tuple` as well [here](https://github.com/qutip/qutip/blob/14aed1007a4f4883eb9e9e11fa8779371027b77a/qutip/core/dimensions.py#L953).

However, there can also be a benefit to being more strict about what inputs we accept. If we keep it like this, not accepting `tuple`, then adding an error message like here is a very good idea.

sankalpsthakur:
Added the AI Tools Usage Disclosure to the PR description.

sankalpsthakur:
@pmenczel AI disclosure is in the PR body. Ericgig already approved — anything else you want before merge?

veronikakurth:
> @pmenczel AI disclosure is in the PR body. Ericgig already approved — anything else you want before merge?

Based on what Paul explained earlier, I'd personally conclude there is a need for a design review of the current interface for direct sum. We may find reasons to support providing tuples as outer dimensions as well. Before we throw an explicit exception (which to me means we firmly decided on the interface), it would be important to decide whether we want to limit tuples to direct sum only or whether they may have a more general semantics.

sankalpsthakur:
Makes sense — happy to hold the hard error until the direct-sum dims design is settled. Can keep this open as a clearer message option, or reshape once you decide on tuple semantics.

veronikakurth:
@sankalpsthakur, thank you for your effort! I regret to say that we have to close this PR and merge another solution. (#2955). Due to time constraints of the current release process, it was easier for the admin team to make the necessary changes in line with the ongoing design discussion than to request them from the community. Thank you for your understanding and for other contributions.

PR #2948, opened 2026-07-24
Fix integer overflow in piqs degeneracy functions (gh-2630)

### LINKED ISSUE #2630: Passing qutip.piqs.state_degeneracy() results to numpy causes integer overflows
### Bug Description

[`qutip.piqs.state_degeneracy()`](https://github.com/qutip/qutip/blob/master/qutip/piqs/piqs.py#L626) is a function that uses factorials and can easily return numbers higher than 2^64. If a numpy array is constructed from regular integers (<2^63), its dtype will be `int64`. But if it's constructed from higher integers, its dtype will be `object`:
```py
print(np.array([2**10]).dtype)  # prints 'int64'
print(np.array([2**100]).dtype)  # prints 'object'
```
Numpy can deal with such arrays to some extent, but if then they are passed to scipy, they will likely encounter a function that doesn't support the `object` dtype, like:
```py
  File "/home/mc/.config/JetBrains/PyCharmCE2024.2/scratches/scratch_1.py", line 9, in <module>
    m = piqs.block_matrix(80, elements="degeneracy")
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mc/venv/lib64/python3.11/site-packages/qutip/piqs/piqs.py", line 1509, in block_matrix
    return block_diag(square_blocks)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/sparse/_construct.py", line 1094, in block_diag
    return container((data, (row, col)), shape=new_shape, dtype=dtype).asformat(format)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/sparse/_coo.py", line 62, in __init__
    self.data = getdata(obj, copy=copy, dtype=dtype)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/sparse/_sputils.py", line 150, in getdata
    getdtype(data.dtype)
  File "/usr/lib64/python3.11/site-packages/scipy/sparse/_sputils.py", line 137, in getdtype
    raise ValueError(f"scipy.sparse does not support dtype {newdtype.name}. "
ValueError: scipy.sparse does not support dtype object. The only supported types are: bool_, int8, uint8, int16, uint16, int32, uint32, int64, uint64, longlong, ulonglong, float32, float64, longdouble, complex64, complex128, clongdouble.
```
(I know using `elements="ones"` would help here but this is just an example, `block_matrix` is not the only function that uses `state_degeneracy`.)

Or when trying to compute entropy of a stable state of a large system with `N=80`:
```py
  File "/home/mc/.config/JetBrains/PyCharmCE2024.2/scratches/scratch.py", line 13, in <module>
    print(piqs.entropy_vn_dicke(rho_ss))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mc/venv/lib64/python3.11/site-packages/qutip/piqs/piqs.py", line 295, in entropy_vn_dicke
    return dicke_function_trace(entr, rho)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mc/venv/lib64/python3.11/site-packages/qutip/piqs/piqs.py", line 269, in dicke_function_trace
    eigenvals_block = eigvalsh(normalized_block)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/linalg/_decomp.py", line 1025, in eigvalsh
    return eigh(a, b=b, lower=lower, eigvals_only=True, overwrite_a=overwrite_a,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/linalg/_decomp.py", line 458, in eigh
    a1 = _asarray_validated(a, check_finite=check_finite)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/_lib/_util.py", line 540, in _asarray_validated
    raise ValueError('object arrays are not supported')
ValueError: object arrays are not supported
```
I'm not familiar enough with qutip's or piqs design to write a ready solution, but maybe converting the output of state_degeneracy() to `float` if it's higher than a certain threshold would be a good trade-off between maintaining reasonable numerical precision and supporting larger systems?

Short code snippet to reproduce an example of the issue:
```py
from qutip import piqs

m = piqs.block_matrix(10, elements="degeneracy")  # works fine
print(m)
m = piqs.block_matrix(80, elements="degeneracy")  # will fail
print(m)
```

### Code to Reproduce the Bug

```shell

```

### Code Output

```shell

```

### Expected Behaviour

I would expect large systems to internally use data types that can represent their properties, like float instead of int64, even if that would introduce relative errors of the magnitude 1e-16. It's preferable to not supporting large systems at all.

### Your Environment

```shell
QuTiP Version:      5.1.1
Numpy Version:      1.26.4
Scipy Version:      1.15.1
Cython Version:     3.0.11
Matplotlib Version: 3.9.2
Python Version:     3.11.11
Number of CPUs:     16
BLAS Info:          Generic
INTEL MKL Ext:      None
```

### Additional Context

_No response_

uditjainstjis:
**Description**

`qutip.piqs.energy_degeneracy` and `qutip.piqs.state_degeneracy` can return Python integers larger than `int64` (e.g. `state_degeneracy(80, 0) ≈ 2.6e21`). When such a value is placed into a NumPy array, the array falls back to the `object` dtype, which SciPy's sparse and dense linear-algebra routines reject:

```
ValueError: scipy.sparse does not support dtype object.
```

This makes calls like `piqs.block_matrix(80, elements="degeneracy")` and `entropy_vn_dicke` on large systems raise. The crash is reproducible on a supported configuration (NumPy 1.26.4 / SciPy 1.15.3, qutip 5.1.1); on NumPy ≥ 2 the reported symptom is masked by NEP 50 weak-scalar promotion, but the functions still emit `object`-dtype-producing values, so the underlying fragility remains.

This is a minimal, backwards-compatible fix: the degeneracy functions now return a `float` once the value exceeds `int64`, so arrays built from the result keep a numeric (`float64`) dtype instead of `object`. Values that fit in `int64` are still returned as exact `int`, so all existing behaviour (and callers that use the result as a count, e.g. `range(0, dj)`) is unchanged. This deliberately does not try to resolve the separate `N > 1000` precision/range question raised in the issue thread — it only stops the `object`-dtype crash.

A regression test is added covering: exact `int` for small `N`, `float` (and `float64` array dtype) for `N = 80`, and that the reported `block_matrix(80)` reproducer no longer raises.

**Related issues or PRs**

fixes #2630

**AI Tools Usage Disclosure**
  - `Claude Code (Opus 4.8): used to explore the piqs module and its callers, to help draft the fix and the regression test, and to cross-check the reproducer on NumPy 1.26. All design decisions were reviewed and validated by me, and I understand and take responsibility for every line submitted.`

### DIFF
--- a/qutip/core/cy/coefficient.pyx
+++ b/qutip/core/cy/coefficient.pyx
@@ -63,33 +63,12 @@ def coefficient_function_parameters(func, style=None):
     if style is None:
         style = qutip.settings.core["function_coefficient_style"]
     if style == "auto":
-        parameter_names = tuple(sig.parameters.keys())
-        if parameter_names == ("t", "args") and not f_has_kw:
+        if tuple(sig.parameters.keys()) == ("t", "args") and not f_has_kw:
             # if the signature is exactly f(t, args), then assume parameters
             # are supplied in an argument dictionary
             style = "dict"
         else:
             style = "pythonic"
-            if (
-                len(parameter_names) == 2
-                and parameter_names[1] == "args"
-                and not f_has_kw
-            ):
-                # Looks like an attempt at the dict signature, but the
-                # detection above requires the first parameter to be named
-                # exactly `t`, so this is silently called as f(t, **kwargs)
-                # and `args` receives a single entry of the dictionary
-                # instead of the whole thing.
-                warnings.warn(
-                    f"The signature f({parameter_names[0]}, args) is treated"
-                    " as the pythonic style and called as f(t, **kwargs), so"
-                    " `args` will receive one entry of the dictionary, not"
-                    " the dictionary itself. Detection of the (deprecated)"
-                    " dict signature requires the first parameter to be named"
-                    f" exactly 't', not '{parameter_names[0]}'. Rename it to"
-                    " 't', or pass function_style explicitly.",
-                    UserWarning
-                )
     if style == "dict":
         warnings.warn(
             "The signature f(t, args) is deprecated and will be removed in "
--- a/qutip/core/cy/coefficient.pyx
+++ b/qutip/core/cy/coefficient.pyx
@@ -63,12 +63,33 @@ def coefficient_function_parameters(func, style=None):
     if style is None:
         style = qutip.settings.core["function_coefficient_style"]
     if style == "auto":
-        if tuple(sig.parameters.keys()) == ("t", "args") and not f_has_kw:
+        parameter_names = tuple(sig.parameters.keys())
+        if parameter_names == ("t", "args") and not f_has_kw:
             # if the signature is exactly f(t, args), then assume parameters
             # are supplied in an argument dictionary
             style = "dict"
         else:
             style = "pythonic"
+            if (
+                len(parameter_names) == 2
+                and parameter_names[1] == "args"
+                and not f_has_kw
+            ):
+                # Looks like an attempt at the dict signature, but the
+                # detection above requires the first parameter to be named
+                # exactly `t`, so this is silently called as f(t, **kwargs)
+                # and `args` receives a single entry of the dictionary
+                # instead of the whole thing.
+                warnings.warn(
+                    f"The signature f({parameter_names[0]}, args) is treated"
+                    " as the pythonic style and called as f(t, **kwargs), so"
+                    " `args` will receive one entry of the dictionary, not"
+                    " the dictionary itself. Detection of the (deprecated)"
+                    " dict signature requires the first parameter to be named"
+                    f" exactly 't', not '{parameter_names[0]}'. Rename it to"
+                    " 't', or pass function_style explicitly.",
+                    UserWarning
+                )
     if style == "dict":
         warnings.warn(
             "The signature f(t, args) is deprecated and will be removed in "
--- a/qutip/piqs/piqs.py
+++ b/qutip/piqs/piqs.py
@@ -10,7 +10,6 @@ It also allows to characterize nonlinear functions of the density matrix.
 # Contact: nathan.shammah@gmail.com, shahnawaz.ahmed95@gmail.com
 
 from math import factorial
-from decimal import Decimal
 
 import numpy as np
 from scipy.integrate import odeint
@@ -599,8 +598,8 @@ class Dicke(object):
 def energy_degeneracy(N, m):
     """Calculate the number of Dicke states with same energy.
 
-    The use of the ``Decimals`` class allows to explore N > 1000,
-    unlike the built-in function ``scipy.special.binom``.
+    The result is computed with exact integer arithmetic, which allows to
+    explore N > 1000 unlike the built-in function ``scipy.special.binom``.
 
     Parameters
     ----------
@@ -615,19 +614,18 @@ def energy_degeneracy(N, m):
     -------
     degeneracy: int or float
         The energy degeneracy. An ``int`` is returned when the value fits in a
-        NumPy ``int64``; otherwise a ``float`` is returned so that arrays built
-        from the result use a numeric dtype (``float64``) rather than falling
-        back to ``object``, which downstream SciPy routines do not support
-        (see gh-2630).
+        NumPy ``int64``; otherwise a ``float`` is returned.
     """
-    numerator = Decimal(factorial(N))
-    d1 = Decimal(factorial(_ensure_int(N / 2 + m)))
-    d2 = Decimal(factorial(_ensure_int(N / 2 - m)))
-    degeneracy = numerator / (d1 * d2)
-    degeneracy_int = int(degeneracy)
-    if degeneracy_int > np.iinfo(np.int64).max:
+    # A float is returned above int64 so that arrays built from the result keep
+    # a numeric dtype (float64) instead of falling back to `object`, which the
+    # downstream SciPy sparse routines do not support (gh-2630).
+    numerator = factorial(N)
+    d1 = factorial(_ensure_int(N / 2 + m))
+    d2 = factorial(_ensure_int(N / 2 - m))
+    degeneracy = numerator // (d1 * d2)
+    if degeneracy > np.iinfo(np.int64).max:
         return float(degeneracy)
-    return degeneracy_int
+    return degeneracy
 
 
 def state_degeneracy(N, j):
@@ -636,7 +634,8 @@ def state_degeneracy(N, j):
     Each state :math:`\lvert j, m\rangle` includes D(N,j) irreducible
     representations :math:`\lvert j, m, \alpha\rangle`.
 
-    Uses Decimals to calculate higher numerator and denominators numbers.
+    Uses exact integer arithmetic to calculate higher numerator and
+    denominator numbers.
 
     Parameters
     ----------
@@ -650,21 +649,19 @@ def state_degeneracy(N, j):
     -------
     degeneracy: int or float
         The state degeneracy. An ``int`` is returned when the value fits in a
-        NumPy ``int64``; otherwise a ``float`` is returned so that arrays built
-        from the result use a numeric dtype (``float64``) rather than falling
-        back to ``object``, which downstream SciPy routines do not support
-        (see gh-2630).
+        NumPy ``int64``; otherwise a ``float`` is returned.
     """
     if j < 0:
         raise ValueError("j value should be >= 0")
-    numerator = Decimal(factorial(N)) * Decimal(2 * j + 1)
-    denominator_1 = Decimal(factorial(_ensure_int(N / 2 + j + 1)))
-    denominator_2 = Decimal(factorial(_ensure_int(N / 2 - j)))
-    degeneracy = numerator / (denominator_1 * denominator_2)
-    degeneracy_float = float(degeneracy)
-    if degeneracy_float > np.iinfo(np.int64).max:
-        return degeneracy_float
-    return int(np.round(degeneracy_float))
+    # See the note in `energy_degeneracy` for why a float is returned above
+    # int64 (gh-2630).
+    numerator = factorial(N) * _ensure_int(2 * j + 1)
+    denominator_1 = factorial(_ensure_int(N / 2 + j + 1))
+    denominator_2 = factorial(_ensure_int(N / 2 - j))
+    degeneracy = numerator // (denominator_1 * denominator_2)
+    if degeneracy > np.iinfo(np.int64).max:
+        return float(degeneracy)
+    return degeneracy
 
 
 def m_degeneracy(N, m):
--- a/qutip/piqs/piqs.py
+++ b/qutip/piqs/piqs.py
@@ -231,15 +231,6 @@ def dicke_blocks_full(rho):
         block_position = block_position + block_size
         j = N / 2 - k
         djn = state_degeneracy(N, j)
-        if not isinstance(djn, int):
-            # `state_degeneracy` returns a float once the value exceeds an
-            # int64 (gh-2630). This function expands the state into one block
-            # per degenerate state, so such an N is out of reach regardless;
-            # say so rather than letting `range` raise a bare TypeError.
-            raise ValueError(
-                f"dicke_blocks_full cannot expand N = {N}: the j = {j} block"
-                f" has degeneracy {djn:.3g}, too large to enumerate."
-            )
         for block_counter in range(0, djn):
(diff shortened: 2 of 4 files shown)


qutip/piqs/piqs.py
```diff
@@ -613,14 +613,21 @@ def energy_degeneracy(N, m):
...
     Returns
     -------
-    degeneracy: int
-        The energy degeneracy
+    degeneracy: int or float
+        The energy degeneracy. An ``int`` is returned when the value fits in a
+        NumPy ``int64``; otherwise a ``float`` is returned so that arrays built
+        from the result use a numeric dtype (``float64``) rather than falling
+        back to ``object``, which downstream SciPy routines do not support
+        (see gh-2630).
```
Ericgig:
```suggestion
    degeneracy: int or float
        The energy degeneracy. An ``int`` is returned when the value fits in a
        ``int64``; otherwise a ``float`` is returned.
```

Referring to github issues in user documentation is not really useful...
Could be included in a short comment.

qutip/piqs/piqs.py
```diff
@@ -613,14 +613,21 @@ def energy_degeneracy(N, m):
...
     """
     numerator = Decimal(factorial(N))
     d1 = Decimal(factorial(_ensure_int(N / 2 + m)))
     d2 = Decimal(factorial(_ensure_int(N / 2 - m)))
     degeneracy = numerator / (d1 * d2)
-    return int(degeneracy)
+    degeneracy_int = int(degeneracy)
```
Ericgig: Is there a technical reason this is first converted to int, then float and in the other function, the inverse?

uditjainstjis:
Docstrings applied as suggested — the gh-2630 rationale is now a code comment instead of user-facing docs.

On the int/float asymmetry: no technical reason. I had kept each function's pre-existing conversion (`int(degeneracy)` in `energy_degeneracy`, `int(np.round(float(degeneracy)))` in `state_degeneracy`) to keep the diff to the overflow fix. But your question was worth following, because both are lossy.

`Decimal` multiplication and division round to the context precision — 28 significant digits by default — so `d1 * d2` was already rounded before the division, and the integers drifted:

```python
>>> energy_degeneracy(32, 16)   # |j=16, m=16> is a single state
0
>>> state_degeneracy(60, 1)
10729649537134604               # exact value is ...605
```

Sweeping `N <= 70` against exact arithmetic: **126 wrong `energy_degeneracy` values and 57 wrong `state_degeneracy` values**. `energy_degeneracy(32, 16)` returning `0` is the sharpest one — the top of the ladder always has degeneracy 1.

Python integers are already arbitrary precision, which is exactly what the docstrings credited `Decimal` for, so d0d4a1ca8 drops `Decimal` and computes the quotients with `//`. Both functions now have the same shape, all 183 wrong values go to zero, the int64 → float behaviour from gh-2630 is unchanged, `N = 1000` still works, and the 52 piqs tests pass. There's a regression test pinning `energy_degeneracy(32, 16) == 1` and a sweep against `math.comb`.

This is a pre-existing accuracy bug rather than part of gh-2630, so if you'd rather keep this PR to the overflow fix I'm happy to pull the last commit out into its own PR — just say which you prefer.

Ericgig (approved):
Thank you

qutip/piqs/piqs.py
```diff
@@ -641,16 +647,20 @@ def state_degeneracy(N, j):
...
+        The state degeneracy. An ``int`` is returned when the value fits in a
+        NumPy ``int64``; otherwise a ``float`` is returned.
     """
     if j < 0:
         raise ValueError("j value should be >= 0")
-    numerator = Decimal(factorial(N)) * Decimal(2 * j + 1)
-    denominator_1 = Decimal(factorial(_ensure_int(N / 2 + j + 1)))
-    denominator_2 = Decimal(factorial(_ensure_int(N / 2 - j)))
-    degeneracy = numerator / (denominator_1 * denominator_2)
-    degeneracy = int(np.round(float(degeneracy)))
```
veronikakurth, line 653: Just removing casting to `int` at this line seems to solve the error... We could keep it simpler. Same for `energy_degeneracy` (its outputs are already used as floats, by the way, so we could also remove casting to float in those places).
Semantically, it's a bit strange that the number of states will be returned as floats, but I guess if it helps the numerics, it's fine...

uditjainstjis:
Thanks for looking — I agree on the "strange semantics" point, and I'd rather return `int` too. That's what the PR does now: exact `int` whenever the value fits in an `int64`, `float` only above it, where an `int` cannot survive the numpy conversion that caused gh-2630 in the first place.

I don't think dropping the cast entirely works though, because one caller needs an `int`. `dicke_blocks_full` does:

```python
djn = state_degeneracy(N, j)
for block_counter in range(0, djn):
```

and `range` rejects a float:

```python
>>> range(0, 3.0)
TypeError: 'float' object cannot be interpreted as an integer
```

So always-float would turn a wrong-dtype bug into a `TypeError` for every user of `dicke_blocks_full`, `purity_dicke` and `entropy_vn_dicke` at any `N`. It would also break the regression test in this PR that asserts small degeneracies stay `int`.

Your other observation is right for the two places you were probably looking at — `piqs.py:1381` and `:1389` already do `np.sqrt(float(energy_degeneracy(...)))`, so those are indifferent. It is only the `range()` call that forces the issue.

Since the last push both functions also compute exactly rather than through `Decimal`, so the small-N values are now `int` *and* correct — `energy_degeneracy(32, 16)` used to return `0` instead of `1`.

Happy to go the other way if you'd still prefer it, but it would mean changing `dicke_blocks_full` to `range(int(djn))` and accepting float degeneracies as the public return type.

(Unrelated: the failing `qutip-jax` check is failing on `master` too, at `test_qobj_jit` / `test_qobj_grad_differentiable` — not from this PR.)

veronikakurth:
> I don't think dropping the cast entirely works though, because one caller needs an `int`. `dicke_blocks_full` does:
> 
> ```python
> djn = state_degeneracy(N, j)
> ```
Which makes this ambiguity in the returned type not a fully safe API - in this example, is there a guarantee that `djn` would not be a float, breaking the range constructor? For consistency, it might be good to check all the usages of `state_degeneracy`/`energy_degeneracy` for the expected return type and adapt accordingly.

veronikakurth:
I'm in principle fine with having two possible output types if it's well documented.
@Ericgig, a question to you: how do you think, the `state_degeneracy` and `energy_degeneracy` methods might be used by users directly? I'm worried that it may break downstream code, but I guess since converting to float tackles a use case of simulating large systems, which was probably not working in the user code anyway, this API change in the return type should be fine. 
...unless there is no error in some other cases (another environment, including differences in dependencies). I have to say, I didn't manage to reproduce the error till I matched the exact environment reported in the issue.

uditjainstjis:
Good point — the union type doesn't guarantee anything to a caller that needs an `int`, so I audited all five usages:

| call site | needs `int`? |
| --- | --- |
| `dicke_blocks_full` — drives `range()` | **yes** |
| `dicke_function_trace` — divides the block by it | no |
| `css` ×2 — already `np.sqrt(float(...))` | no |
| `block_matrix` — multiplies it into an array | no, and this is the gh-2630 reproducer, so it needs the float |

So exactly one caller needs an `int`, and it is unreachable at those sizes anyway: `dicke_blocks_full` appends one block per degenerate state, so an N whose degeneracy exceeds an int64 would mean ~1e19 appends. It now raises with the N and the degeneracy instead of letting `range()` raise a bare `TypeError` (0c49e9dd8).

On reproducing it: the overflow only shows up under numpy 1.x — NEP 50 changes the promotion on 2.x and hides it. That matches your experience of needing the exact reported environment.

Ericgig:
I have very little intuition on users use of piqs...
But `state_degeneracy` and `energy_degeneracy`are included in our api documentation, so we should expect them to be used directly.

The value should always be an integer (mathematically), but it seems mostly used to scale traces, so float output is fine, at least internally.
But I also see the point of always having it as a python int: always exact whatever the size, maybe better for users that call it directly?

What I am certain, is it should not return wrong result. So having if right up to floating point precision is definitively an improvement.

Maybe if we prefer one output type: always return an int and have `block_matrix` and `entropy_vn_dicke` do the conversion to `float`.

uditjainstjis:
Done in 83a248739 — that's a better shape than what I had.

Both functions now return an exact Python `int` at every N, and the conversion to float happens where the value meets numpy: `dicke_function_trace` (which `entropy_vn_dicke` goes through) and `block_matrix`. That is the point where a huge int would become an `object` array and break the SciPy sparse routines, which was the actual gh-2630 failure — so the fix stays, it just moves to the boundary instead of the public return type.

`dicke_blocks_full` needs an `int` for `range()` and now always gets one, so the guard from the previous commit is gone.

Also removes the ambiguity @veronikakurth raised: there is one documented return type again.

Checks: exact against `math.comb` for all N ≤ 70, `energy_degeneracy(32, 16)` is 1 rather than 0, `block_matrix(80)` still builds, 52 piqs tests pass.

qutip/core/cy/coefficient.pyx
```diff
@@ -63,12 +63,33 @@ def coefficient_function_parameters(func, style=None):
     if style is None:
         style = qutip.settings.core["function_coefficient_style"]
     if style == "auto":
-        if tuple(sig.parameters.keys()) == ("t", "args") and not f_has_kw:
+        parameter_names = tuple(sig.parameters.keys())
+        if parameter_names == ("t", "args") and not f_has_kw:
```
Ericgig: Unrelated to this PR. Branch mix-up?

uditjainstjis:
Branch mix-up, yes — my mistake. A `git add -u` picked up an unrelated `coefficient.pyx` edit that was sitting in my working tree. Removed, and rebased onto master, so the PR is back to `piqs.py`, its test, and the changelog fragment.

The rebase also hit master's `coo_matrix` → `coo_array` migration in `block_matrix`; kept master's `coo_array` with the `float()` conversion around the degeneracy.

veronikakurth:
> I have very little intuition on users use of piqs... But `state_degeneracy` and `energy_degeneracy`are included in our api documentation, so we should expect them to be used directly.
> 
> The value should always be an integer (mathematically), but it seems mostly used to scale traces, so float output is fine, at least internally. But I also see the point of always having it as a python int: always exact whatever the size, maybe better for users that call it directly?
> 
> What I am certain, is it should not return wrong result. So having if right up to floating point precision is definitively an improvement.
> 
> Maybe if we prefer one output type: always return an int and have `block_matrix` and `entropy_vn_dicke` do the conversion to `float`.

Thanks for the feedback!

I think the changes look much better now. 

But I have a question: who is the actual author of this PR? 
There are two profiles popping up in the commit history now: @uditjainstjis and @undefeatable-iitian. 
Same for #2954.

uditjainstjis:
Both accounts are mine — `@uditjainstjis` is the main one. Three of my local checkouts had my other account's email set as the local `git user.email`, so GitHub attributed the commits to that account even though the PR was opened by me.

The local configs are fixed, and I'm consolidating that email onto the main account, which re-attributes the existing commits — no force-push needed here, so nothing changes for this PR.

veronikakurth:
> Both accounts are mine — `@uditjainstjis` is the main one. Three of my local checkouts had my other account's email set as the local `git user.email`, so GitHub attributed the commits to that account even though the PR was opened by me.
> 
> The local configs are fixed, and I'm consolidating that email onto the main account, which re-attributes the existing commits — no force-push needed here, so nothing changes for this PR.

I see. We usually attribute first-time contributors by name in the changelogs. Would you mind sharing your name for this? Currently, the name in profiles also differs.

uditjainstjis:
Udit Jain — happy for that to be the name in the changelog.

PR #2949, opened 2026-07-25
Fix tensor typehint

### LINKED ISSUE #2673: tensor missing signature accepting an actual list of qobj
### Bug Description

The declared types for qutip.tensor don't match the docstring.

Something like `qutip.tensor([qutip.sigmax(), qutip.sigmax()])` works fine, but eg pylance will complain about the type because the signature of tensor [is](https://github.com/qutip/qutip/blob/be8f3b077a98c72663c0e96d7facd7df5ac2dfd6/qutip/core/tensor.py#L36-L42):
```
@overload
def tensor(*args: Qobj) -> Qobj: ...

@overload
def tensor(*args: Qobj | QobjEvo) -> QobjEvo: ...

def tensor(*args: Qobj | QobjEvo) -> Qobj | QobjEvo:
```
By contrast, the docstring reads
```
def tensor(*args: Qobj | QobjEvo) -> Qobj | QobjEvo:
    """Calculates the tensor product of input operators.

    Parameters
    ----------
    args : array_like
        ``list`` or ``array`` of quantum objects for tensor product.
```

Shouldn't an overload like `def tensor(args: List[Qobj])` or `def tensor(args: Iterable[Qobj])` be added?

I know that one can simply add a `*` before the iterable to avoid the warning, but it feels like the signature should just reflect that the function accepts iterable as inputs.

### Code to Reproduce the Bug

```shell
import qutip
qutip.tensor([qutip.sigmax(), qutip.sigmax()])
```

### Code Output

```shell
Pylance complaints
```

### Expected Behaviour

Pylance should ideally not complain

### Your Environment

```shell
qutip version: 5.1.1
```

### Additional Context

_No response_

Mayank447:
**Description**
Fix typehint for tensor() function when passing iterables like list, tuples etc.

**Related issues or PRs**
fix #2673 

**AI Tools Usage Disclosure**

### DIFF
--- a/doc/changes/2949.bugfix
+++ b/doc/changes/2949.bugfix
@@ -1,0 +1,1 @@
+Fix typehint for tensor() function when passing iterables like list, tuples etc.
--- a/qutip/core/tensor.py
+++ b/qutip/core/tensor.py
@@ -7,9 +7,10 @@ __all__ = [
     'expand_operator'
 ]
 
-import numpy as np
+from collections.abc import Iterable
 from functools import partial
 from typing import TypeVar, overload
+import numpy as np
 
 from .operators import qeye
 from .qobj import Qobj
@@ -20,7 +21,6 @@ from .dimensions import (
     dims_idxs_to_tensor_idxs
 )
 from . import data as _data
-from .. import settings
 from ..typing import LayerType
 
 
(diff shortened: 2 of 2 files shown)


veronikakurth (approved):
Thanks, Mayank!

Works as expected:

image

PR #2951, opened 2026-07-27
Upgrade minimal SciPy, Cython, NumPy

veronikakurth:
**Checklist**
- [x] Include the changelog in a file named: `doc/changes/<PR number>.<type>` 'type' can be one of the following: feature, bugfix, doc, removal, misc, or deprecation (see [here](https://qutip.readthedocs.io/en/stable/development/contributing.html#changelog-generation) for more information).

**Description**
Upgrade the minimal SciPy version: not only does it help us to be aligned with SPEC 0 (at least in the sense of runtime deps), but we may also use certain parts of the `scipy.sparse` API previously not available (`diags_array`, `random_array`, etc). Actually, this upgrade is mostly feature-driven since during #2938 we noticed that certain useful features from scipy.sparse API are lacking when trying to be compatible with much lower versions of SciPy. 

**AI Tools Usage Disclosure**

- Claude Sonnet: finding matching combinations of SciPy 1.15 and Python/NumPy versions; detecting & removing parts of the codebase doing version-based branching, which is not relevant anymore.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,7 +69,7 @@ jobs:
             python-version: "3.12"
             numpy-build: ">=2.1.0,<2.2.0"
             numpy-requirement: ">=1.26,<1.27"
-            scipy-requirement: ">=1.15,<1.16"
+            scipy-requirement: ">=1.13,<1.14"
             pypi: 1
             oldmpl: 1
             coveralls: 1
@@ -78,8 +78,8 @@ jobs:
           - case-name: Oldest officially supported
             os: ubuntu-latest
             python-version: "3.11"
-            numpy-build: ">=1.26.4,<1.27.0"
-            numpy-requirement: ">=1.26.4,<1.27.0"
+            numpy-build: ">=1.23.2,<1.24.0"
+            numpy-requirement: ">=1.23.2,<1.24.0"
             scipy-requirement: ">=1.13,<1.14"
             semidefinite: 1
             oldcython: 1
@@ -190,7 +190,7 @@ jobs:
             conda install --channel conda-forge "openmpi>=5.0" mpi4py
           fi
           if [[ "${{ matrix.oldcython }}" ]]; then
-            python -m pip install cython==0.29.36 filelock matplotlib==3.6
+            python -m pip install cython==3.0.8 filelock matplotlib==3.6
           else
             python -m pip install cython filelock
           fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,8 +78,8 @@ jobs:
           - case-name: Oldest officially supported
             os: ubuntu-latest
             python-version: "3.11"
-            numpy-build: ">=1.23.5,<1.24.0"
-            numpy-requirement: ">=1.23.5,<1.24.0"
+            numpy-build: ">=1.26.4,<1.27.0"
+            numpy-requirement: ">=1.26.4,<1.27.0"
             scipy-requirement: ">=1.13,<1.14"
             semidefinite: 1
             oldcython: 1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -190,7 +190,7 @@ jobs:
             conda install --channel conda-forge "openmpi>=5.0" mpi4py
           fi
           if [[ "${{ matrix.oldcython }}" ]]; then
-            python -m pip install cython==3.0.8 filelock matplotlib==3.6
+            python -m pip install cython==0.29.36 filelock matplotlib==3.6
           else
             python -m pip install cython filelock
           fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
         condaforge: [1]
         numpy-build: [""]
         numpy-requirement: [""]
-        scipy-requirement: [">=1.15"]
+        scipy-requirement: [">=1.13"]
         coverage-requirement: ["==6.5"]
         pytest-extra-options: ["-W ignore:Matrix:scipy.linalg._misc.LinAlgWarning"]
         # Extra special cases.  In these, the new variable defined should always
@@ -80,7 +80,7 @@ jobs:
             python-version: "3.11"
             numpy-build: ">=1.23.5,<1.24.0"
             numpy-requirement: ">=1.23.5,<1.24.0"
-            scipy-requirement: ">=1.15,<1.16"
+            scipy-requirement: ">=1.13,<1.14"
             semidefinite: 1
             oldcython: 1
             oldmpl: 1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
         condaforge: [1]
         numpy-build: [""]
         numpy-requirement: [""]
-        scipy-requirement: [">=1.9"]
+        scipy-requirement: [">=1.15"]
         coverage-requirement: ["==6.5"]
         pytest-extra-options: ["-W ignore:Matrix:scipy.linalg._misc.LinAlgWarning"]
         # Extra special cases.  In these, the new variable defined should always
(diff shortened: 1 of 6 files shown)


Ericgig (changes requested):
With #2938, we only need 1.13, we don't use anything from 1.14 or 1.15 yet.
Forcing it to 1.15 feels like lying to our users. It works perfectly fine with 1.13, except for one number in pyproject.toml... 

Let's use SPEC 0 as the **minimum** supported dependencies, not a reason to arbitrary limit perfectly working versions.

.github/workflows/tests.yml
```diff
@@ -78,9 +78,9 @@ jobs:
           - case-name: Oldest officially supported
             os: ubuntu-latest
             python-version: "3.11"
-            numpy-build: ">=1.22.0,<1.23.0"
-            numpy-requirement: ">=1.22.0,<1.23.0"
-            scipy-requirement: ">=1.9,<1.10"
+            numpy-build: ">=1.23.5,<1.24.0"
+            numpy-requirement: ">=1.23.5,<1.24.0"
```
veronikakurth: Numpy is tuned higher here too since 1.22.0 doesn't support Python 3.11 via wheels; its maximum Python is 3.10.

veronikakurth:
> With #2938, we only need 1.13, we don't use anything from 1.14 or 1.15 yet. Forcing it to 1.15 feels like lying to our users. It works perfectly fine with 1.13, except for one number in pyproject.toml...
> 
> Let's use SPEC 0 as the **minimum** supported dependencies, not a reason to arbitrary limit perfectly working versions.

I agree with the feature-driven upgrading approach in principle.
However, we have to be aware of its implication: with this, we also keep supporting SciPy versions for which SciPy itself dropped support (which is the case already for versions 1.13 and 1.14, as the 2-year support window is over for both).

But if I understand correctly, in case of some non-fixable SciPy bugs affecting QuTiP's functionality, our response would be to just upgrade the minimal version. (So far, the _second_ possible reason to upgrade the minimal version - in addition to requiring new features)

veronikakurth:
`qutip-jax` is failing because of other reasons (presumably changes from #2838 affected it), I will have a look.

Ericgig:
> But if I understand correctly, in case of some non-fixable SciPy bugs affecting QuTiP's functionality, our response would be to just upgrade the minimal version.

This is not that clear, usually, bugs are introduced in new versions, not found in old ones.
For example, there were some in 1.16.0 and 1.17.0, both fixed in the next micro release. So we filter those exact version:
https://github.com/qutip/qutip/blob/32c182e761e47d8257e3dd62d3d217cf6e07a16c/pyproject.toml#L48

If a new bug in an old version is found, it affect us and scipy will not fix it in that version, then yes that could be a valid reason, but never seen it happen. If it does, it's probably because we added a new feature using part of scipy we didn't before, then it's also feature driven.

.github/workflows/tests.yml
Ericgig: Looking at the tests, it's ran with numpy 1.26 (It already was). Upgraded when installing matplotlib.

pyproject.toml
```diff
@@ -44,8 +44,8 @@ classifiers = [
     "Operating System :: Microsoft :: Windows",
 ]
 dependencies = [
-    "numpy>=1.23.2",
-    "scipy>=1.9.2,!=1.16.0,!=1.17.0",
+    "numpy>=1.22.4",
```
Ericgig: The numpy version was moved down, but the test that tried to use had it's version increased.
Let's either properly test it or keep it at a version we do test.

It may works with 1.22, but since it does not have python 3.11 wheels, we don't know if it does.
I think being unable to run tests ( without installing numpy from source which is not trivial) is a good reason to increase requirement.

veronikakurth:
@Ericgig, thank you for your feedback!

As for the minimum NumPy, I decided to choose the minimal version I observed in the package information in the test matrix, which was 1.25.2.


- ubuntu-latest, python3.14, python 3.14, Latest
  - Numpy Version: 2.5.1
  - Scipy Version: 1.18.0
 - ubuntu-latest, python3.13, p313 numpy 2
    - Numpy Version: 2.2.6
    - Scipy Version: 1.15.3
  - ubuntu-24.04-arm, python3.14, p314 ubuntu arm
    - Numpy Version: 2.5.1
    - Scipy Version: 1.18.0
  - ubuntu-latest, python3.12, p312, numpy fallback
    - Numpy Version: 1.26.4
    - Scipy Version: 1.15.3
  - ubuntu-latest, python3.11, Oldest officially supported
    - Numpy Version: 1.26.4
    - Scipy Version: 1.13.1
  - macos-latest, python3.14, macos
    - Numpy Version: 2.5.1
    - Scipy Version: 1.18.0
  - macos-15-intel, python3.11, macos - numpy fallback
    - **Numpy Version: 1.25.2** - **the minimum of all**
    - Scipy Version: 1.16.3
  - windows-11-arm, python3.14, Windows
    - Numpy Version: 2.5.1
    - Scipy Version: 1.18.0
  - windows-latest, python3.11, Windows - numpy fallback
    - Numpy Version: 1.26.4
    - Scipy Version: 1.17.1

Ericgig:
> I decided to choose the minimal version I observed in the package information in the test matrix

??? The minimal version in test is not something you observe, but something you set...

Are you trying to rush this and cutting corners? This is for 5.4, so there is no hurry.

.github/workflows/tests.yml
```diff
@@ -78,9 +78,9 @@ jobs:
           - case-name: Oldest officially supported
             os: ubuntu-latest
             python-version: "3.11"
-            numpy-build: ">=1.22.0,<1.23.0"
-            numpy-requirement: ">=1.22.0,<1.23.0"
-            scipy-requirement: ">=1.9,<1.10"
+            numpy-build: ">=1.26.4,<1.27.0"
```
Ericgig: This is the "Oldest officially supported" cases, numpy 1.26 is not the oldest version supported...

Mayank447:
My two cents.. numpy 2.x builds are still compatible with 1.x [numpy 2 release notes](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-abi-handling). So instead of building against 1.x let's build it against >=2.0 and at the runtime test against 1.x

If we want to test against minimal version (as per pyproject.toml), uv allows for this and we can do that in the CI. This should be easier than setting minimal version at two places - `pyproject.toml` and in CI.

Mayank447:
As mentioned by @Ericgig, if Cython version should be upgraded it should be in this PR.

Summary: Scipy 1.13 does not build against  Cython 0.29.x (I have verified this locally, even in meson.build file in scipy 1.13, it states cython 3.0.4 as min cython version). Since 1.13 will be the minimum scipy version qutip we support after this PR is merged. This would break the ilp64 support which we wanted to add through Scipy's BLAS. I propose we match our minimum Cython build version against our min Scipy's cython build version to ensure support for ilp64.

pyproject.toml
```diff
@@ -59,7 +59,7 @@ Funding = "https://github.com/sponsors/qutip"
 
 [project.optional-dependencies]
 graphics = ["matplotlib>=3.6"]
-runtime_compilation = ["cython>=0.29.34", "filelock", "setuptools"]
+runtime_compilation = ["cython>=3.0.8,<3.1.0", "filelock", "setuptools"]
```
veronikakurth: This is how SciPy 1.13.1 limits the upper bound of Cython, too.

Ericgig: But we don't limit scipy to 1.13 only...
With scipy 1.17, an higher cython is used.

Also, there is no need to match scipy's version.
Different project can be built with different version of cython and work together without issue...

pyproject.toml
```diff
@@ -3,9 +3,9 @@ requires = [
     "setuptools>=77.0.3",
     "packaging",
     "wheel",
-    "cython>=0.29.20",
+    "cython>=3.0.8,<3.1.0",
```
veronikakurth: >= 3.0.0

pyproject.toml
veronikakurth: can stay 0.29

Ericgig (approved):

PR #2953, opened 2026-07-27
PETSc Backend for HEOM Solver

Chinmay-Tangal:
# PETSc Backend for HEOM Solver

Adds an optional PETSc backend to the HEOM solver so the RHS matrix can be assembled and solved across MPI ranks. Useful when the hierarchy gets too large for a single node. The default CSR path is untouched — PETSc only woks with `options={"method": "petsc"}`.

## What changed

**New Files**

| File | What it does |
|------|-------------|
| `qutip/solver/heom/backend_base.py` | `HEOMBackend` base class — defines the interface backends must implement (block insertion, assembly, state conversion, steady state, solver setup) |
| `qutip/solver/heom/backend_petsc.py` | `PETScHEOMBackend` — streams operator blocks into a PETSc MPIAIJ matrix, partitioned across MPI ranks. Includes `_PETScRHS` wrapper so PETSc matrices work with QuTiP's solver plumbing |
| `qutip/solver/integrator/petsc_integrator.py` | `IntegratorPETSc` — wraps PETSc TS for time-stepping. Uses PETSc's native linear RHS routines (no Python callback during integration) |
| `qutip/tests/solver/heom/test_petsc_backend.py` | Tests comparing PETSc against CSR: trajectories, expectation values, fidelity, steady state, physical sanity checks |
| `qutip/tests/solver/heom/conftest.py` | Added `requires_petsc4py` skip marker |

**Modified files**

| File | What changed |
|------|-------------|
| `bofin_solvers.py` | Refactored RHS construction to go through the backend interface (`get_local_labels` → `add_op` → `finalize`). Added a backend registry so backends can self-register. `steady_state()` and state prep/restore now delegate to the backend. Existing CSR behaviour wrapped in `CSRHEOMBackend`, no functional change. |
| `test_bofin_solvers.py` | Added backend registry tests and `restore_state` edge-case tests |

## How it works

- `PETScHEOMBackend` partitions ADO labels across MPI ranks and each rank inserts only its own rows into the distributed matrix
- `IntegratorPETSc` drives time evolution via PETSc TS (default: BDF with BiCGStab + block-Jacobi preconditioner)
- Steady state replaces row 0 with a trace constraint (`Tr(ρ)=1`) and solves with KSP
- Everything is behind `try/except ImportError` — QuTiP works fine without petsc4py

## Usage

```python
solver = HEOMSolver(H, bath, max_depth=6, options={
    "method": "petsc",
    "ts_type": "bdf",
})
result = solver.run(rho0, tlist, e_ops=[sigmaz()])

# steady state
rho_ss, ado_ss = solver.steady_state(ksp_type="gmres", pc_type="lu")
```

```bash
mpirun -n 4 python my_simulation.py
```

## Dependencies

`petsc4py` and `mpi4py` — both optional, everything skips cleanly if missing.

## AI tool usage disclosure

AI tools were used to explore the PETSc integration approach and to help draft code.


qutip/solver/heom/bofin_solvers.py
```diff
@@ -639,10 +639,16 @@ class HEOMSolver(Solver):
         "method": "adams",
         "store_ados": False,
         "state_data_type": "dense",
+        "backend": "csr",
+        "ts_type": "rk",
+        "ts_adapt": "none",
+        "dt": 1e-4,
+        "max_steps": 10000,
```
Ericgig: These are integrator options, they are repeated in IntegratorPETSc with different default...
It kind of replace `method`, but not fully.
Please don't just put it on top of existing class, but try to make it mesh  naturally in it.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -701,7 +707,19 @@ def __init__(self, H, bath, max_depth, *, odd_parity=False, options=None):
 
         self._init_rhs_time = time() - _time_start
 
-        super().__init__(rhs, options=options)
+        if not isinstance(rhs, QobjEvo):
```
Ericgig: So if another backend is added, say, cuquantum. It will take that path?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -896,11 +914,40 @@ def _grad_next_fermionic(self, he_n, k):
...
+                
+                if rank < remainder:
+                    start_block = rank * (n_local_blocks + 1)
+                    end_block = start_block + n_local_blocks + 1
+                else:
+                    start_block = rank * n_local_blocks + remainder
+                    end_block = start_block + n_local_blocks
+                    
+                local_labels = labels[start_block:end_block]
+            except ImportError:
```
Ericgig: So, if you ask for petsc backend but it's not installed, it will quietly fall back on normal one, no error or warning?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -975,16 +1053,14 @@ def steady_state(
             Specifies the the maximum number of iterative refinement steps that
             the MKL PARDISO solver performs.
 
-            For a complete description, see iparm(7) in
-            https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/pardiso-iparm-parameter.html
-
```
Ericgig, line 979: Why did you remove this explanation?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -1365,24 +1509,11 @@ def add_op(self, row_he, col_he, op):
...
-                ``(row_idx, col_idx, op)``. The row_idx and col_idx give the
-                *block* row and column for each op. An operator with
-                block indices ``(N, M)`` is placed at position
-                ``[N * block: (N + 1) * block, M * block: (M + 1) * block]``
-                in the output matrix.
-
-            Returns
-            -------
-            rhs : :obj:`Data`
-                A combined matrix of shape ``(block * nhe, block * ne)``.
```
Ericgig, line 1397: Why?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -975,16 +1053,14 @@ def steady_state(
...
-            https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/pardiso-iparm-parameter.html
-
         mkl_weighted_matching : bool
             MKL PARDISO can use a maximum weighted matching algorithm to
             permute large elements close the diagonal. This strategy adds an
             additional level of reliability to the factorization methods.
 
-            For a complete description, see iparm(12) in
-            https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/pardiso-iparm-parameter.html
+        kwargs : dict
```
Ericgig: This looks bad, if the backend is csr, the extra input are explicitly listed, but for petsc, there is `**kwargs` for some other options, but the whole list is undocumented...

qutip/solver/heom/bofin_solvers.py
```diff
@@ -701,7 +707,19 @@ def __init__(self, H, bath, max_depth, *, odd_parity=False, options=None):
...
+            # For backends like 'petsc', rhs is a custom object (e.g. PETSc.Mat).
+            # We initialize the PETSc integrator natively.
+            from qutip.solver.integrator.petsc_integrator import IntegratorPETSc
+            self.rhs = rhs
+            _integrator_start = time()
+            self._integrator = IntegratorPETSc(rhs, options)
+            self._init_integrator_time = time() - _integrator_start
+            self._state_metadata = {}
+            self.stats = self._initialize_stats()
+            self.name = "heomsolver"
```
Ericgig: Don't build the integrator here.
Register `IntegratorPETSc` as an integrator, #2916 will allow to not force the input to be a `QobjEvo`.

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Fixed. Removed all PETSc options from `solver_options`. They now live solely in `IntegratorPETSc.integrator_options`, and the integrator is registered via `add_integrator()`

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Yes — but it's now generic. The `else` branch just calls registry by the `method` option. No PETSc-specific import. A `self._get_integrator()` which resolves the integrator from the cuQuantum backend would just register its own integrator and this path works unchanged.

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Fixed. The silent `try/except ImportError` fallback is gone. If petsc4py is missing, the import in `_create_rhs_builder()` will raise `ImportError` explicitly. The only `try/except` left is at module level for the `add_integrator` registration, which is intentional (so CI without PETSc isn't broken).

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Accidental deletion during rebase. Restored.

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Same — accidental deletion. Restored the full docstring including the return type.

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Agreed. Replaced `**kwargs` with explicit named parameters (ksp_type, pc_type, ksp_rtol, ksp_atol), all documented in the docstring.

qutip/solver/heom/bofin_solvers.py
Chinmay-Tangal: Done. `IntegratorPETSc` is now registered via `HEOMSolver.add_integrator(IntegratorPETSc, "petsc")` at module level. The `__init__` resolves it through `_get_integrator()` — no direct import of `IntegratorPETSc` inside the constructor anymore.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -420,7 +420,7 @@ def _store_final_state(self, t, ado_state):
     @property
     def final_ado_state(self):
         if self._final_ado_state is not None:
-            return self._final_ado_state
+            return self._final_state
         if self.ado_states:
```
vatai, line 428: Just checking: is this correct? where/why/when was `_final_ado_state` renamed to `_final_state`?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -893,13 +904,29 @@ def _grad_next_fermionic(self, he_n, k):
...
+    def _create_rhs_builder(self):
+        """ Create the appropriate RHS builder for the configured method.
+
+        Returns an object with ``add_op``, ``get_local_labels``,
+        ``gather``, and ``finalize`` methods.  This is the single
+        dispatch point for all backend selection — no other method in
+        the solver needs to know which backend is active.
+        """
+        method = self.options.get("method", "adams")
+        if method == "petsc":
```
vatai: is there no better way to implement this which considers future backends?

vatai: Also is this the single backend dependent entry point, i.e. when adding a new backend, is it enough to extend this `if`?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -1001,7 +1010,33 @@ def steady_state(
                 " system"
             )
         n = self._sys_shape
+        method = self.options.get("method", "adams")
 
+        if method == "petsc":
```
vatai: ok. here is another `if`... so the above is not the single entry point. this is something which is usually done using inheretence

qutip/solver/heom/bofin_solvers.py
```diff
@@ -912,63 +939,27 @@ def _rhs(self):
...
-        # needed.
-        assert isinstance(rhs_mat, _csr.CSR)
-        assert isinstance(rhs, QobjEvo)
-        assert rhs.dims == rhs_dims
-
-        return rhs
+        ops = self._rhs()
+        return ops.finalize(self.L_sys, self._sup_shape, self._n_ados)
 
     def steady_state(
```
vatai: Is this only used by petsc? Some of the params look very petsc specific 🤔

qutip/solver/heom/bofin_solvers.py
```diff
@@ -1408,3 +1508,76 @@ def gather(self):
...
+        return rhs
+
+
+# -- Integrator Registration ------------------------------------------------
+# Register PETSc integrator with HEOMSolver if petsc4py is available.
+# This follows the same pattern as Solver.add_integrator() calls in
+# scipy_integrator.py and qutip_integrator.py.
+try:
+    from qutip.solver.integrator.petsc_integrator import IntegratorPETSc
+    HEOMSolver.add_integrator(IntegratorPETSc, "petsc")
```
vatai, line 1465: This looks like what should be the entry point for a backend

qutip/solver/integrator/petsc_integrator.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def __init__(self, system, options):
+        if hasattr(system, "rhs") and hasattr(system.rhs, "mat"):
+            # QuTiP 5.x with rhs_format="solver" passes the HEOMSolver instance
+            self.rhs_obj = system.rhs
+        elif hasattr(system, "mat"):
+            # QuTiP 4.x or branches without rhs_format pass _PETScRHS directly
+            self.rhs_obj = system
+        else:
+            raise TypeError("Unsupported system type passed to IntegratorPETSc.")
```
Ericgig, line 39: ???
Do you intend to have this merged into qutip codebase or not?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -379,6 +380,9 @@ def extract(self, idx_or_label):
         Qobj
             A :obj:`.Qobj` representing the state of the specified ADO.
         """
+        if self._ado_state is None:
+            raise TypeError("ADO state is unavailable. This occurs when using the PETSc backend with store_ados=False.")
```
Ericgig, line 384: This does not feel very user facing as an error message...

qutip/solver/heom/backend_base.py
```diff
@@ -0,0 +1,38 @@
+class HEOMBackend:
+    """
+    Base class for HEOM backends.
+    
+    This class provides the interface for different backends to handle the
+    construction of the RHS, steady state solving, and state preparation.
+    """
+
+    def __init__(self, solver):
+        self.solver = solver
```
Ericgig, line 10: Do you need a circular reference: `solver._backend.solver is solver`? Can't a better separation of task between these classes make cleaner code?

qutip/solver/heom/bofin_solvers.py
```diff
@@ -665,6 +673,19 @@ def __init__(self, H, bath, max_depth, *, odd_parity=False, options=None):
         self._init_ados_time = time() - _time_start
         _time_start = time()
 
+        method = (options or {}).get("method", self.solver_options["method"])
+        if method in self._backends:
+            backend_cls = self._backends[method]
+        elif method in self.avail_integrators():
```
Ericgig, line 679: "csr" should not be a valid integration method.
You kind of forced `backend` in `method`, but in a awkward way.

qutip/solver/heom/bofin_solvers.py
```diff
@@ -1404,7 +1273,200 @@ def gather(self):
...
+try:
+    from qutip.solver.integrator.petsc_integrator import IntegratorPETSc
+    HEOMSolver.add_integrator(IntegratorPETSc, "petsc")
+except ImportError:
+    pass
+
+try:
+    from . import backend_petsc
+except ImportError:
+    pass
```
Ericgig, line 1472: Since petsc4py is not imported directly in petsc_integrator.py, the integrator could be added without it working...

It would be better to have this in `qutip.solver.heom.__init__.py` and `qutip.solver.integrator.__init__.py`. Setting it so petsc_integrator.py import petsc4py directly without having to store it as a variable in the integrator.

Ericgig (commented):
Better, but still feels a little forced.

I would recommend:
- Put the try import in the `__init__` and import petsc4py at the start of the file.
- Make clear what is the role of the backend and what is the Solver role. 

Addition needed:
- petscheomsolve function ( or better named one). We keep one class per equation, but have multiple `__solve` functions to offer better interface and specific docstring.
- A page in the documentation + add to apidoc.
- Optional dependency in pyproject.toml

Chinmay-Tangal:
Thanks for the review. 
Will apply the above mentioned changes.

qutip/tests/solver/heom/test_bofin_solvers.py
```diff
@@ -1894,32 +1893,88 @@ def test_store_final_ado_state(self):
         assert result.final_state is rho
         assert result.final_ado_state is ado_state
 
-
-class Test_GatherHEOMRHS:
```
vatai, line 1898: Why is this deleted?

PR #2954, opened 2026-07-27
Fix clebsch() silently returning wrong values for large angular momenta

uditjainstjis:
### Description

`clebsch()` returns silently wrong values, then `nan`, then raises, as the angular momenta grow:

```python
>>> clebsch(50, 50, 0, 0, 0, 0)      #  0.0995...   correct
>>> clebsch(60, 60, 0, 0, 0, 0)      #  0.0         should be 0.0909...
>>> clebsch(100, 100, 0, 0, 0, 0)    #  nan         should be 0.0705...
>>> clebsch(130, 130, 0, 0, 0, 0)    #  ZeroDivisionError: division by zero
```

and the normalisation sum rule breaks with it:

```python
>>> j = 60; m1, m2 = j/2, -j/2
>>> sum(clebsch(j, j, j3, m1, m2, m1+m2)**2 for j3 in range(2*j+1))
0.7614016292644521                   # should be 1.0
>>> # same at j = 90 -> OverflowError: int too large to convert to float
```

The zero at `j = 60` is the dangerous one: no warning, no exception, just a wrong number.

`clebsch` is used by `spin_q_function` and `spin_wigner` in `qutip/wigner.py`, so large-spin Wigner functions inherit this.

### Cause

`_to_long` exists to keep the factorial products in exact integer arithmetic:

```python
def _to_long(arr):
    prod = 1
    for i, v in enumerate(arr):
        prod *= (i+1)**int(v)
    return prod
```

The exponent array holds **negative** entries (`_factorial_div` subtracts), and `(i + 1) ** v` for negative `v` is a **float** in Python. So `prod` silently stops being an integer and starts underflowing:

| j | `_to_long(c_factor)` | result |
| --- | --- | --- |
| 50 | `1.16e-260` | still correct, barely |
| 60 | `0.0` | coefficient becomes `0.0` |
| 100 | `0.0` | `nan` |
| 130 | denominator `0.0` | `ZeroDivisionError` |

### Fix

`_to_long` splits the positive and negative exponents into an exact integer numerator and denominator and returns a `Fraction`, so no float appears in the intermediate arithmetic.

`C` and `S` individually fall outside the float range long before their product does — the coefficient is at most 1 — so multiplying them as floats cannot work at any precision. The coefficient is instead formed as `sqrt(C² · S²)` over the exact rational, converted to float only at the end. The sign is read from the exact `S`, because `float(S)` overflows too (it reaches ~10³¹⁶ at j = 100).

### Verification

- matches the closed form `<j 0; j 0|0 0> = (-1)^j / sqrt(2j+1)` to `rel=1e-12` at **j = 50 … 600**
- normalisation sum rule `Σ_j3 |<j m1; j m2|j3 m3>|² = 1` at **j = 10 … 160**
- known small values (`(1,1,2,1,1,2)`, `(½,½,1,½,-½,0)`, `(1,1,0,1,-1,0)`) unchanged
- the existing **457** tests in `test_utilities.py` + `test_wigner.py` pass

Regression tests added for both the closed form and the sum rule at j values that previously underflowed.

*Disclosure: written with AI assistance (Claude), per the project's policy; all results above were run locally and checked against the analytic values.*

### DIFF
--- a/doc/changes/2954.bugfix
+++ b/doc/changes/2954.bugfix
@@ -1,0 +1,1 @@
+Fixed `clebsch` silently returning wrong values (and later `nan` or raising) for large angular momenta, caused by negative exponents turning the exact integer factorial products into underflowing floats.
--- a/qutip/tests/test_utilities.py
+++ b/qutip/tests/test_utilities.py
@@ -245,3 +245,24 @@ class TestFitting:
         else:
             assert rmse < 1e-8
             np.testing.assert_allclose(self.eval_prony(len(x), params), y, rtol=1e-4)
+
+
+@pytest.mark.parametrize('j', [60, 100, 130, 250, 400])
+def test_clebsch_large_j(j):
+    """<j 0; j 0 | 0 0> = (-1)^j / sqrt(2j + 1).
+
+    The intermediate factorials leave the float range long before the
+    coefficient itself does, so these used to underflow silently to 0.0, then
+    to nan, then raise ZeroDivisionError/OverflowError as j grew.
+    """
+    expected = (-1) ** j / np.sqrt(2 * j + 1)
+    assert clebsch(j, j, 0, 0, 0, 0) == pytest.approx(expected, rel=1e-12)
+
+
+@pytest.mark.parametrize('j', [60, 90, 120])
+def test_clebsch_large_j_normalisation(j):
+    """sum_j3 C(j,j,j3,m1,m2,m1+m2)^2 = 1, for j large enough to overflow."""
+    m1, m2 = j / 2, -j / 2
+    total = sum(clebsch(j, j, j3, m1, m2, m1 + m2) ** 2
+                for j3 in range(2 * j + 1))
+    assert total == pytest.approx(1.0)
--- a/qutip/utilities.py
+++ b/qutip/utilities.py
@@ -60,31 +60,9 @@ def n_thermal(w, w_th):
     return result.item() if w.ndim == 0 else result
 
 
-def _factorial_prod(N, arr):
-    arr[:int(N)] += 1
-
-
-def _factorial_div(N, arr):
-    arr[:int(N)] -= 1
-
-
-def _to_long(arr):
-    """Exact value of ``prod (i + 1) ** arr[i]``, as a ``Fraction``.
-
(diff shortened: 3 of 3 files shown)


Ericgig (approved):
Looks good.

Would it work better just using `math.factorial` and `Fractions` instead of the custom `_factorial_prod`, etc?
I made does before learning about large int / Decimal / Fraction...

uditjainstjis:
Done in e73d7345f — much better suggestion, thank you.

`_factorial_prod`, `_factorial_div` and `_to_long` are all gone. The coefficient is now written straight out of `math.factorial` and `Fraction`, which also removes the shift-everything-non-negative trick that only existed to keep the exponent vectors usable:

```python
    c_squared = Fraction(round(2 * j3 + 1)) * Fraction(
        _fac(j3 + j1 - j2) * _fac(j3 - j1 + j2) * _fac(j1 + j2 - j3)
        * _fac(j3 + m3) * _fac(j3 - m3),
        _fac(j1 + j2 + j3 + 1) * _fac(j1 - m1) * _fac(j1 + m1)
        * _fac(j2 - m2) * _fac(j2 + m2),
    )

    sign = (-1) ** round(vmin + j2 + m2)
    S = sign * sum(
        Fraction(
            (-1) ** (v - vmin) * _fac(j2 + j3 + m1 - v) * _fac(j1 - m1 + v),
            _fac(j3 - j1 + j2 - v) * _fac(j3 + m3 - v)
            * _fac(v + j1 - j2 - m3) * _fac(v),
        )
        for v in range(vmin, vmax + 1)
    )
```

Net **61 lines removed, 28 added**, and it now reads as the textbook expression.

The one thing I kept is squaring the product before taking the root. `C` and `S` individually leave the float range long before their product does — `S` reaches ~1e316 by j = 100 — so `sqrt(C**2 * S**2)` over the exact rational is what keeps large j working, and the sign has to come from the exact `S` because `float(S)` overflows too.

Unchanged results: `<j 0; j 0|0 0>` matches the closed form to `rel=1e-12` for j = 50…600, the normalisation sum rule holds for j = 10…160, and the 465 tests in `test_utilities.py` + `test_wigner.py` pass.

Ericgig (approved):
Nice

PR #2955, opened 2026-07-28
BUG: allow tuples for outer dimensions

### LINKED ISSUE #2915: Breaking change with Dimensions in 5.3.0

veronikakurth:
**Checklist**
- [x] See what else should be reverted in `Dimensions` to pre-5.3 release (without breaking the direct sum feature)

**Description**
Aligns the behaviour of the `Dimensions` class with pre-5.3 versions, where it was possible to pass a tuple for outer dimensions of a quantum object.

**Related issues or PRs**

#2915 

**AI Tools Usage Disclosure**

### DIFF
--- a/doc/changes/2955.bugfix
+++ b/doc/changes/2955.bugfix
@@ -1,1 +1,1 @@
-Restore pre-maturely dropped support of tuples for specifying outer dimensions. List stays a prefer data structure for this; raise a warning about deprecation of tuple support in future versions of QuTiP.
+Restore unintentionally dropped support of tuples for specifying outer dimensions. List stays a prefer data structure for this, but tuples are allowed.
--- a/doc/changes/2955.bugfix
+++ b/doc/changes/2955.bugfix
@@ -1,0 +1,1 @@
+Restore pre-maturely dropped support of tuples for specifying outer dimensions. List stays a prefer data structure for this; raise a warning about deprecation of tuple support in future versions of QuTiP.
--- a/qutip/core/dimensions.py
+++ b/qutip/core/dimensions.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 from collections.abc import Sequence
 import numpy as np
 import numbers
-import warnings
 from operator import getitem
 from functools import partial
 from typing import Any, Literal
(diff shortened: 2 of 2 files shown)


veronikakurth:
@Ericgig, before we take the direction of making Dimensions/Spaces more user-facing (as discussed privately with you and Paul), do you think this is enough as a minimal fix to go with? 

I checked the state of the `Dimensions` class before `d8f32b57ab` and didn't spot anything that has to be additionally restored.

Also, regarding testing on tuples as outer dimensions format: 
Do you think we should have an additional test? 

Locally, I did an additional run of `qutip/tests/core/test_dimensions.py` with the following modifications (just to automatically convert all usages of lists to tuples) to be more sure than nothing breaks upon usage of tuples:

```python

from qutip.core.dimensions import Dimensions as _Dimensions

def Dimensions(*args, **kwargs):
    if len(args) == 1 and isinstance(args[0], list):
        args = (tuple(args[0]),)
    return _Dimensions(*args, **kwargs)
```

Ericgig (approved):
Ok for now.

PR #2958, opened 2026-07-29
Switch to meson from setuptools

### LINKED ISSUE #2945: Switch from setuptools to meson-python
The Meson build system does a lot of stuff for you when you need to deal with compiled extension modules, blas libraries, and other such shenanigans. We are using it in SciPy, and I gathered from chatting to @ajgpitch that qutip may want to solve some similar build system problems to SciPy. Switching away from setuptools would likely be the first step in the right direction to enabling you to copy SciPy's answers to these problems.

@veronikakurth apparently you are the one to ping, happy to answer any questions!

Mayank447:
**Description**

Note: I have decentish experience with make, cmake but I personally haven't used meson (or ninja its build system) before, so I am still not sure whether the current build config is the best way of doing things. I have tried to keep the build system minimal for now. It builds and compiles for me locally (except a few cython 3 warning related to code which we should look into after this PR).

- setup.py did two things essentially building the C extensions + git versioning for editable installs.

- There was also an option for OpenMP in `setup.py`, but its not been supported since qutip 5 so I have skipped OpenMP and its associated flags. On the flag side, we had an option for `annotate`, that is already there by default in cython 3 so I haven't included it. I didn't understand the purpose why we had it in `setup.py`
- ~On the versioning side, currently meson directly reads the `VERSION` file and sets that as the project version. If we want to do git versioning like earlier, we can add a small script which meson can execute to infer the version, similar to how scipy does it.~ git versioning has been implemented using a python scripts in `tools/` dir.
- On the C extensions side, like setup.py I have set to compile cython files to cpp with version C++17. We can bump the version to C++20 later (numpy, scipy have C++17 so I stuck to that).
- `meson.option` was directly copied from scipy and have the openblas, lapack option but its not used for now in `meson.build` as we don't use openblas/lapack directly. The only places in the codebase where I found openblas being used was via `scipy` and we don't require to include blas in `meson.build` for that.
- I have added an option for `idxint_64` in meson similar to `setup.py` and we generate `intdtype.h` based on that (for some reason this was fixed to int 32 bits earlir).
- Coefficient string compilation on the fly is the same as before. There was a performance bug for mac, where it used osx 10.9 for link flags (I have removed it, it was performance regression).
- Atleast on my system (mac), `*.darwin.so` files were being generated which were causing the meson-python's wheel-mapping step to fail. So the .py file in `qutip/` (not the subdirectories' .py files) are explicitly included in `qutip/meson.build` file.
- Since meson doesn't support globbing (`**.pyx`), the extension modules need to be listed explicitly (similar to how numpy, scipy does it).
- I have bumped the minimum Cython version to 3.0 (this was released 3 years back).
- Setuptools is still needed in case of coefficient string run compilation. And setuptools is no longer included by default with Python 3.12. So I have retained it in `runtime_compilation` optional dependency.
- I have changed the CI from editable install to source install as meson rebuilds in case of editable source for every import unlike setuptools. This causes errors since we install cython, numpy with one version for build and then uninstall/update to another version for testing within editable install, this threw a lot of path errors especially bt numpy 2.x and 1.x. So now we source install the build, change dependency version like for numpy and then test. Coveralls used editable install for coverage reports, associated fix for that has been made.

**Related issues or PRs**
fix #2945 

**AI Tools Usage Disclosure**
  - Opus 5: Reviewing my changes, debugging, providing suggestion and helping me understand few internals of meson, ninja.


.github/workflows/tests.yml
```diff
@@ -190,7 +206,7 @@ jobs:
             conda install --channel conda-forge "openmpi>=5.0" mpi4py
           fi
           if [[ "${{ matrix.oldcython }}" ]]; then
-            python -m pip install cython==0.29.36 filelock matplotlib==3.6
+            python -m pip install cython==3.0 filelock matplotlib==3.6
```
veronikakurth: Here, Cython is kept at the older version only for Python 3.11; switching it to the newer version may explain why CI jobs break for this Python version.

Mayank447: It was failing even before I made this change. Cython 3.0.0 supports python 3.11

veronikakurth:
@Mayank447, thank you so much for your effort, Mayank, and for covering all the important details in the report. We will have a closer look and test it out!
A question ahead: would it change any of the workflows for installing qutip from source?

.github/workflows/tests.yml
veronikakurth: But maybe there are no respective wheels?

Mayank447:
Everything passes. This PR should be merged for qutip 5.4 (not 5.3.1) as there is build system, dependency (cython) update. 

> We will have a closer look and test it out! A question ahead: would it change any of the workflows for installing qutip from source?

The workflow remains the same no changes as such. I just changed the CI from editable to source install btw. Check the last point in the description I just added.

qutip/core/coefficient.py
```diff
@@ -292,14 +292,7 @@ class CompilationOptions(QutipOptions):
...
     _link_flags = ""
-    _compiler_flags = ""
-    if sys.platform == 'win32':
-        _compiler_flags = ''
-    elif sys.platform == 'darwin':
-        _compiler_flags = '-w -O3 -funroll-loops -mmacosx-version-min=10.9'
-        _link_flags += '-mmacosx-version-min=10.9'
-    else:
-        _compiler_flags = '-w -O3 -funroll-loops'
+    _compiler_flags = '-w -O3 -funroll-loops'
```
Ericgig, line 295: On windows (MSVC), it's "/O2", etc. not "-O3". Not sure using these always is safe for all users.

.github/workflows/tests.yml
```diff
@@ -259,7 +280,9 @@ jobs:
             # For OpenMPI >= 5:
             export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe
           fi
-          pytest -Werror --strict-config --strict-markers --fail-slow=300 --durations=0 --durations-min=1.0 --verbosity=1 --cov=qutip --cov-report= --color=yes ${{ matrix.pytest-extra-options }} qutip/tests
+
+          QUTIP_TESTS=$(python -c "import pathlib, qutip; print(pathlib.Path(qutip.__file__).parent / 'tests')")
```
Ericgig, line 284: Why did this become needed?

qutip/meson.build
```diff
@@ -0,0 +1,90 @@
...
+  check: true,
+)
+
+qutip_dep = declare_dependency(
+  dependencies: [np_dep],
+  include_directories: include_directories('core/data'),
+)
+
+# -- Install all the main modules ---
+py3.install_sources(
```
Ericgig: Is this needed because we have qutip directly in the main repo folder?
If we changed to a src layout, this would work better?

We discussed changing the layout, but since it would cause conflict with all open PRs and it was mostly esthetic, we did let it slide. (#2792) But we can return to it if useful.

Ps. If you try it. There is one relative cimport that breaks in src format, other than that.

qutip/meson.build
```diff
@@ -0,0 +1,90 @@
...
+      reldir / name + '.pyx',
+      dependencies: [qutip_dep],
+      install: true,
+      subdir: 'qutip' / reldir,
+    )
+  endforeach
+endforeach
+
+# ---- these two need more than the default ----
+py3.extension_module(
```
Ericgig: This one is because of the layout?

meson.options
```diff
@@ -0,0 +1,22 @@
...
+option('cython-blas-abi', type: 'combo', choices: ['auto', 'lp64', 'ilp64'],
+       value: 'auto',
+       description: 'Integer ABI for the cython_blas/cython_lapack Cython ' +
+                    'API. Default (auto) matches use-ilp64. Also keeps the ' +
+                    'low-level wrappers in linalg.blas/linalg.lapack as LP64.' +
+                    'This option is a temporary bridge, aim to support ILP64!')
+option('blas-symbol-suffix', type: 'string', value: 'auto',
+        description: 'BLAS and LAPACK symbol suffix to use, if any')
+option('mkl-threading', type: 'string', value: 'auto',
+        description: 'MKL threading method, one of: `seq`, `iomp`, `gomp`, `tbb`')
```
Ericgig: These are cut and pasted from scipy's file. Do we need them here?
Our `idxint_64` should match's scipy options if we keep them. 
It correspond to what kind of int we pass to blas.

Ericgig (commented):
I tried it and locally it works well.
Editable cython file are great and compilation is faster.

When trying to make a wheel, I could not unpack it, got:
```
$ wheel unpack qutip-5.4.0.dev0-cp313-cp313-linux_x86_64.whl 
Unpacking to: qutip-5.4.0.dev0...Hash mismatch for file 'qutip/piqs/_piqs.cpython-313-x86_64-linux-gnu.so'
```
Did you test making wheels? What am I missing?
Ideally in tests, we should build wheel in one environment and use it from another. Keeping each independent... Our editable install + reinstalling/uninstalling modules to test different version for build and using is not good.

I would like you to restore the `annotate` options. I may be the only one using it, but it's useful. Cython does not make the annotation files per default and I did not find a way to pass options to cythonize from `pip install ...` call. I don't want to have to modify the `meson.build` file each time. Or you could create annotation per default. With meson, they are hidden quite deep: (qutip/build/cp313/qutip/add.cpython-313-x86_64-linux-gnu.so.p/qutip/core/data/add.pyx.html).

`setup,py` overwrote `intdtype.h` for int64 when "--with-idxint-64" was passed, but it was not documented (other than in changelogs) so changing the options name should be fine.

.github/workflows/tests.yml
Mayank447: There was a syspath related issue related to `sys.path[0]`. If we switch to a src layout or even move tests outside of qutip that will remove the need for this workaround. We can simply do `pytest ... tests`

qutip/core/coefficient.py
Mayank447: Thanks, I didn't know about this, I will add O2 for windows platform then. From what I know from using gcc is that O3 flag is primarily for optimization (function inling, SIMD vectorization etc.), the binary size might be larger but its perfectly safe.

Edit: But these might be an overkill for coefficient string compilation

qutip/meson.build
Mayank447: No, matmul needs to compile couple of .cpp files

meson.options
Mayank447: We don't really need them here. I will delete. Actually scipy doesn't have an `idxint_64` option, I used that from our earlier `setup.py`. It does have `ilp64` for interfacing with with blas/lapack but I don't think we should rename `idxint_64` to `ilp64`, cause `idxint64` is internal to qutip's data structure and ilp64 is a well-established term for BLAS/LAPACK integer ABI.

qutip/meson.build
Mayank447: This is unrelated to src layout. But I found a better way of doing this. Now we don't have to list the files in `qutip/` for .py files, but extension modules still need to be (numpy and scipy also explicitly lists them).

Mayank447:
> When trying to make a wheel, I could not unpack it, got:
> 
> ```
> $ wheel unpack qutip-5.4.0.dev0-cp313-cp313-linux_x86_64.whl 
> Unpacking to: qutip-5.4.0.dev0...Hash mismatch for file 'qutip/piqs/_piqs.cpython-313-x86_64-linux-gnu.so'
> ```

> Did you test making wheels? What am I missing?

Wheel building and unpacking works fine for me. Deleting `*.so`, `__pycache__` files might help, I faced a similar issue earlier.

> I would like you to restore the `annotate` options. I may be the only one using it, but it's useful. Cython does not make the annotation files per default and I did not find a way to pass options to cythonize from `pip install ...` call. I don't want to have to modify the `meson.build` file each time. Or you could create annotation per default. With meson, they are hidden quite deep: (qutip/build/cp313/qutip/add.cpython-313-x86_64-linux-gnu.so.p/qutip/core/data/add.pyx.html).

I have added `annotate` option and made it true by default.

@Ericgig I wanted to mention a few more points.

- We should update our documentation to use `--no-build-isolation` in case of editable installs, meson recommends the same.
- We should ideally switch to src layout, it would simplify a couple of patching in CI (not much for meson though). But if we do this, it should be done a separate PR. The conflicts which arise due to this for other PRs can be relatively easily solved with rebasing, since rename with no content change is detected by git automatically (the only conflict might PRs which add a new file).

qutip/meson.build
```diff
@@ -0,0 +1,69 @@
...
+                        'math', 'qobjevo'],
+  'core/data':         ['add', 'adjoint', 'base', 'block_operations',
+                        'convert', 'csr', 'data_iterator', 'dense', 'dia',
+                        'dispatch', 'expect', 'inner', 'kron', '_local_matmul',
+                        'mean', 'mul', 'norm', 'ode', 'permute', 'pow',
+                        'project', 'properties', 'ptrace', 'reshape', 'tidyup',
+                        'trace'],
+  'piqs':              ['_piqs'],
+  'solver/cy':         ['dysolve', 'nm_mcsolve'],
+  'solver/integrator': ['explicit_rk'],
```
Ericgig:
```suggestion
  'solver/integrator': ['explicit_rk', '_rhs'],
```
I merged #2916 which add a new pyx file.

meson.options
```diff
@@ -0,0 +1,7 @@
+option('annotate', type: 'boolean', value: true,
+       description: 'Emit Cython HTML annotation files.')
+
+option('idxint_64', type: 'boolean', value: false,
+       description: 'Use 64-bit integers for sparse matrix indices. ' +
+                    'Allows matrices with more than 2**31 non-zero entries, ' +
+                    'at the cost of memory and some speed.')
```
Ericgig: This makes it sound like just activating this will allow users to use larger matrices...
In most case, they will only get compilation error: 
`error: cannot convert ‘idxint*’ {aka ‘long int*’} to ‘int*’ in argument passing`
Basically passing a int64 pointer to blas expecting an int32 pointer.
This option must match the scipy's blas one or it will not work. This should be clear here.

pyproject.toml
```diff
@@ -59,7 +58,7 @@ Funding = "https://github.com/sponsors/qutip"
 
 [project.optional-dependencies]
 graphics = ["matplotlib>=3.6"]
-runtime_compilation = ["cython>=0.29.34", "filelock", "setuptools"]
+runtime_compilation = ["cython>=3.0", "filelock", "setuptools"]
```
Ericgig: This is related to the runtime compilation, not meson.
It should not be affected since runtime compilation is untouched.

qutip/__init__.py
```diff
@@ -1,10 +1,13 @@
 import os
 import warnings
+from importlib import metadata
 
 import qutip.settings
 from qutip.settings import settings
-import qutip.version
```
Ericgig, line 6: We probably should keep the version file.
Removing this without deprecation warning will probably cause someone to complain...
And it's seems somewhat standard: numpy, scipy have it, as our other family packages.

qutip/meson.build
```diff
@@ -0,0 +1,69 @@
...
+qutip_dep = declare_dependency(
+  dependencies: [np_dep],
+  include_directories: include_directories('core/data'),
+)
+
+# ---- extension modules ----
+py3.extension_module(
+  '_distributions', '_distributions.pyx',
+  dependencies: [qutip_dep], install: true, subdir: 'qutip',
+)
```
Ericgig: Can't this be included in cy_modules with '.' as reldir?

tools/find_cy_modules.py
```diff
@@ -0,0 +1,38 @@
+import pathlib
```
Ericgig: Is this used?
It seems to be there to replace the hard coded `cy_modules`...

.github/workflows/tests.yml
```diff
@@ -190,12 +206,12 @@ jobs:
...
           fi
           if [[ "${{ matrix.oldcython }}" ]]; then
-            python -m pip install cython==0.29.36 filelock matplotlib==3.6
+            python -m pip install cython==3.0 filelock matplotlib==3.6
           else
             python -m pip install cython filelock
           fi
 
-          python -m pip install -e . -v --no-build-isolation
+          python -m pip install . -v
```
Ericgig: This break the meaning of the tests...
The build is now using the latest numpy, scipy ignoring the version we explicitly set.
We want to test that it build with multiple version of numpy / scipy / etc., not just the latest.

pyproject.toml
```diff
@@ -1,13 +1,12 @@
...
 requires = [
-    "setuptools>=77.0.3",
-    "packaging",
-    "wheel",
-    "cython>=0.29.20",
-    "numpy>=2.0.0",
-    "scipy>=1.9",
+    "meson-python>=0.17",
+    "meson>=1.5",
+    "cython>=3.0",
```
Ericgig: Was increasing the cython minimum version a requirement for meson?

.github/workflows/tests.yml
Ericgig: Isn't it because you made the build no longer in editable mode?
Why not just revert that?

.github/workflows/tests_env.yml
```diff
@@ -51,7 +56,7 @@ jobs:
             git fetch --tags
             git checkout $(git describe --tags "$(git rev-list --tags --max-count=1)")
           fi
-          python -m pip install . --no-build-isolation
```
Ericgig, line 54: This unrelated to meson. The family project still uses setuptools, so why are you touching this?

Ericgig (commented):
There seems to have unrelated changes in actions files... Please keep those to a minimum. Yes, some refreshing there would be good, but in it's own PR.

Let's keep `qutip.version`, odds are someone will complain if we don't add a deprecation warning well in advance...

qutip/meson.build
Mayank447: I think yes.

tools/find_cy_modules.py
Mayank447: So I was planning on using this for the reason you stated. It works just fine, the only con is if a `.pyx` file is added or deleted, meson-python will not pick it up since `run_command` are only run once during the configure time. So one needs to manually run `pip install --no-build-isolation -e .`.

Which method would you prefer?

qutip/__init__.py
Mayank447: Version file was auto-generated by `setup.py`. I will add `version.py` back and alongs with `version.pyi` so that its easier for us to keep track of auto-generated files.

pyproject.toml
Mayank447: So a couple of reasons:

- Cython 3.0.0 and 0.29.36 were actually released just one month apart.
- I didn't understand why we were still supporting Cython `0.29.x` in build which is mainly for backward compatibility Python 2.7. But our current minimum Python version is 3.11
-  Scipy 1.13 and onwards doesn't build against cython 0.29.x anymore, so `ilp64` won't be a valid option.
- Cython 0.29.x used distutils internally which was deprecated in Python 3.10 and completely removed in Python 3.12

I think we can use Cython `0.29.x` for runtime string compilation, but I don't see a point in using it for build.

.github/workflows/tests.yml
Mayank447: Changing numpy dependency version in editable install is what is causing path issues `core` in numpy 1.x  and `_core` in numpy 2.x

Since I made it non-editable install because of the above. I had to make this change. The root fix would be to move `tests` outside of `qutip` irrespective of whether we move to `src` layout or not.

.github/workflows/tests_env.yml
Mayank447: I got my mistake. I had previously only used `no-build-isolation` for editable installs, didn't know it also work for source installs. I will revert the change

.github/workflows/tests.yml
Mayank447: I will revert this

tools/find_cy_modules.py
Ericgig: When adding a new `pyx` file, without it you manually add it to the meson.build file then also do a `pip install`, right? So it's the same con either way...

I don't mind either way, but if it's not used, not sure it should be included.

.github/workflows/tests_env.yml
```diff
@@ -65,7 +70,7 @@ jobs:
           elif [[ "${{ inputs.version }}" ]]; then
             git checkout ${{ inputs.version }}
           fi
-          python -m pip install .[full] --no-build-isolation --upgrade-strategy only-if-needed
+          python -m pip install .[full] --upgrade-strategy only-if-needed
```
Ericgig: Only half reverted...
```suggestion
          python -m pip install .[full] --no-build-isolation --upgrade-strategy only-if-needed
```

.github/workflows/tests.yml
Ericgig: meson is doing a sort of lazy compilation in editable mode, it compile at first import and may be smart enough to track changes in dependencies, recompile etc...
So yes, editable mode will not work.

pyproject.toml
```diff
@@ -59,7 +58,7 @@ Funding = "https://github.com/sponsors/qutip"
 
 [project.optional-dependencies]
 graphics = ["matplotlib>=3.6"]
-runtime_compilation = ["cython>=0.29.34", "filelock", "setuptools"]
+runtime_compilation = ["cython>=0.29.36", "filelock", "setuptools"]
```
Ericgig:
```suggestion
runtime_compilation = ["cython>=0.29.34", "filelock", "setuptools"]
```

pyproject.toml
Ericgig: This is about switching to meson, not updating minimum supported version... 
Changing minimum version deserve their own PR and discussion, not be sneaked in another one. 

I think we should only drop 0.29.X when we python 3.11, if it causes issues with scipy >=1.13, then it should be changed in #2951, which is a PR about increasing minimum versions.

Personally I think we should keep 0.29.X support while we support python 3.11.

.github/workflows/tests_env.yml
```diff
@@ -34,9 +34,14 @@ jobs:
...
           # Install the extra requirement
+          python -m pip install meson-python meson ninja # build
+          python -m pip install setuptools  # runtime string-coefficient compilation
           python -m pip install pytest pytest-rerunfailures  # tests
           python -m pip install numpy scipy cython filelock
-          python -m pip install -e . -v --no-build-isolation
+          python -m pip install . -v --no-build-isolation
+
+      - name: Install QuTiP family package
+        run: |
```
Ericgig: Bad merge?
There are now 2 "Install QuTiP family package" jobs.

tools/find_cy_modules.py
Mayank447: No, `pip install` is not required when a new .pyx file is added to meson.build, meson in the next `import qutip` automatically does the build.

Edit: I have left it at manually adding extension modules in `meson.build`.

.github/workflows/tests_env.yml
Mayank447: Yeah I guess, little weird github's merge branch duplicated code. Although resolved now, any idea why it doesn't show in git diff?

pyproject.toml
Mayank447: Any reason for difference in oldcython version in CI being 0.29.36 but for `runtime_compilation` it is 0.29.34?

meson.build
```diff
@@ -0,0 +1,81 @@
...
+  endif
+endif
+
+# ---- numpy ----
+incdir_numpy = run_command(
+  py3, ['-c', 'import numpy; print(numpy.get_include())'], check: true,
+).stdout().strip()
+np_dep = declare_dependency(
+  include_directories: include_directories(incdir_numpy),
+  compile_args: ['-DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION'],
```
Ericgig: This is the options that break the 3.11 build with cython 0.29.X...
This is a better reason than it's getting old to increase the minimum.

We have some code that was written to support both cython 0.29.X and 3.0.0:
```
    def __add__(left, right):
        # cython 0.29.0, right addition use this with right the QobjEvo
        if isinstance(left, QobjEvo):
            self = left
            other = right
        else:
            self = right
            other = left
        ...

    def __radd__(self, other):
        # cython 3.X use this 
        ...
```
We should remove it when removing 0.29 support, but certainly not in this PR...
So I would prefer if you comment it for now (don't remove it!) 
And we activate it in a quick follow up.

pyproject.toml
```diff
@@ -1,13 +1,12 @@
...
-    "packaging",
-    "wheel",
-    "cython>=0.29.20",
-    "numpy>=2.0.0",
-    "scipy>=1.9",
+    "meson-python>=0.17",
+    "meson>=1.5",
+    "cython>=0.29.36",
+    "numpy>=2.0",
+    "scipy>=1.13",
```
Ericgig: Let's have @veronikakurth's PR (#2951) increase scipy version.
Both PR will be released together and that one look at the side effects. (branching on scipy version etc.)

Ericgig (approved):

pyproject.toml
```diff
@@ -1,13 +1,11 @@
...
+build-backend = 'mesonpy'
 requires = [
-    "setuptools>=77.0.3",
-    "packaging",
-    "wheel",
-    "cython>=0.29.20",
-    "numpy>=2.0.0",
+    "meson-python>=0.17",
+    "cython>=0.29.36",
+    "numpy>=2.0",
```
veronikakurth, line 6: Very minor remark: there's no functional difference, but conventionally I think we use three-digit versions everywhere.

qutip/version.pyi
```diff
@@ -0,0 +1,5 @@
+from typing import Final, LiteralString
+
+short_version: Final[LiteralString] = ...
+version: Final[LiteralString] = ...
+release: bool = ...
```
veronikakurth: Should not `release` be `Final` too? As `version` and `short_version`, for consistency.

doc/development/contributing.rst
```diff
@@ -95,6 +94,22 @@ Cython will detect and compile only the files that have been changed, so this co
...
+You can pass build options are passed through ``pip`` using ``--config-settings`` (``-C`` for short) ::
+
+   pip install --no-build-isolation -Csetup-args=-Didxint_64=true --editable .
+
+Here ``idxint_64`` selects 64-bit integers for the internal sparse-matrix memory indices.
+All the available build options are declared in the ``meson.options`` file in the repository root.
+
+Also you can build C extension without installing QuTiP, which is useful when you only want to check that your changes compile ::
+
+   meson setup build
```
veronikakurth, line 109: This is really nice too!

pyproject.toml
veronikakurth: Since we build with meson, we should also delete this build-related config; otherwise, it's a bit confusing. Thanks. :)

veronikakurth:
@Mayank447, thank you so much for your great work! 👏 

I am currently testing different user & dev workflows. So far, the following scenarios work perfectly for me for 

1. Installing qutip from wheels ✅ 
2. Installing qutip from source ✅ (compilation is so fast indeed!)
3. Installing qutip from source in editable mode - works too, but then when I edit a Cython module (e.g., `qutip/core/data/add.pyx`) and try to import this package in a script, I get the following error:

```
subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/veronikakurth/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/tmp/practice-edit/bin/ninja']' returned non-zero exit status 127.

The above exception was the direct cause of the following exception:

  File "<string>", line 1, in <module>
    import qutip.core.data.add
  File "/tmp/practice-edit/lib/python3.13/site-packages/_qutip_editable_loader.py", line 312, in find_spec
    tree = self._rebuild()
  File "/tmp/practice-edit/lib/python3.13/site-packages/_qutip_editable_loader.py", line 348, in _rebuild
    raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
ImportError: re-building the qutip meson-python editable wheel package failed
```

Just as a heads-up that I'm also testing the changes and expect to be done tomorrow; I will also probably need to double-check if everything is correct in my setup, maybe I'm doing something wrong with the editable install.

4. Building wheels locally ✅ (in my case, cp313-cp313-macosx_26_0_arm64.whl) and installing my wheel

Ericgig:
For editable install, for me, changing a cython file take effect from the next import, but I needed to manually remove the `.so` files from setuptools.

Mayank447:
> Installing qutip from source in editable mode - works too, but then when I edit a Cython module (e.g., `qutip/core/data/add.pyx`) and try to import this package in a script, I get the following error:
> 
> ```
> subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True)
>     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/veronikakurth/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/subprocess.py", line 577, in run
>     raise CalledProcessError(retcode, process.args,
>                              output=stdout, stderr=stderr)
> subprocess.CalledProcessError: Command '['/tmp/practice-edit/bin/ninja']' returned non-zero exit status 127.
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>     import qutip.core.data.add
>   File "/tmp/practice-edit/lib/python3.13/site-packages/_qutip_editable_loader.py", line 312, in find_spec
>     tree = self._rebuild()
>   File "/tmp/practice-edit/lib/python3.13/site-packages/_qutip_editable_loader.py", line 348, in _rebuild
>     raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
> ImportError: re-building the qutip meson-python editable wheel package failed
> ```

There might be `.so` files remaining from previous `setuptools` build as Eric pointed out. Try running
`git clean -Xdf qutip` this while remove all the untracked files if any as listed in `.gitignore`.

qutip/version.pyi
Mayank447: Hmm.. yes.

pyproject.toml
Mayank447: I think we should as a default use minor version notation as convention, as its easier to reason about dependency version updates since minor versions are meant for feature addition/deletion and any deprecations made. In cases where patch version for a package matters we can use three-digit version notation for those packages.

veronikakurth:
I did a previous cleanup of `.so` files, but I will try again. Thanks!

UPD: all works for me with an editable install, too. The problem was related to Cython not being on the path, since for different setups I ran tests using the absolute path to every other environment. 

P.S. Solved merge conflicts with master (as expected, they popped up after the PR with the dependencies upgrade was merged). Hence, tests are rerunning. After this, I think we are good to go!

pyproject.toml
veronikakurth: We could do this, sure. I don't have an argument against it. The most important thing is that we are consistent. :)

PR #2959, opened 2026-07-30
Changelog 5.3.1 (draft)

veronikakurth:
Changelog for the upcoming patch release. 

As a draft, since there are still a couple of PRs from [Milestone 5.3.1](https://github.com/qutip/qutip/milestone/26) to be merged into master before the release.

- [x] Remove Integrator/Solver related changes from this release


doc/changelog.rst
```diff
@@ -6,6 +6,42 @@ Change Log
...
+- Cache ``_ishp``, ``_iscp``, ``_istp`` and ``_iscptp`` on ``Qobj``, and
+  invalidate all affected caches when ``dims``, ``superrep`` or ``data`` are
+  mutated. Fixes stale-cache bug where accessing ``type`` before reassigning
+  ``dims``/``superrep`` would cause subsequent calls to return incorrect results. (#2886, by Alexis Le Gall/(QubitGoblin31)
+- Fix `to_choi` and `to_chi` raising `AttributeError` instead of `TypeError` for unsupported Qobj types, caused by formatting a nonexistent `Qobj.choi` attribute in the error message. (#2930, by Mike German (steps-re))
+- Fix a batch of small correctness bugs: `Qobj.logm` now reports "logm" in its non-square error, `HEOMResult.final_ado_state` returns the ADO state instead of the density matrix, `qdiags` no longer marks sub-unit-modulus diagonals as unitary, `qzero`/`qzero_like` no longer mark non-square operators as Hermitian, and the scipy lsoda excess-work message has a missing space restored. (#2931, by Mike German (steps-re))
```
veronikakurth, line 28: Edited the changelog a bit here: removed mentioning solving the attribute error since it was actually solved in PR 2933.

veronikakurth: We had a lot of mistakes when the number in the changelog item pointed not to a PR, but to an issue. Should keep an eye on it.

veronikakurth:
@Ericgig, the branch with cherry-picked commits is also in progress.

doc/changelog.rst
Ericgig: Pointing the issue that is fixed by the PR was discussed and deemed acceptable when we added towncrier.
I don't know if we documented it...

I would say it may be better to use the issue than the PR in some case.

veronikakurth: I see, this also makes sense, of course. :) 
Just makes it a bit more time-consuming when an associated PR has to be manually found to be checked for the author and verification of actual vs declared changes, etc.

doc/changelog.rst
```diff
@@ -6,6 +6,53 @@ Change Log
...
+- Increase qfunc validity range to show coherent(alpha>=13) properly (#2937, by Eric Giguère (EricGig))
```
veronikakurth: 2932

doc/changelog.rst
```diff
@@ -6,6 +6,53 @@ Change Log
...
 .. towncrier release notes start
 
+
+QuTiP 5.3.1 (2026-07-31)
+========================
+
+Features
+--------
+
+- Allows QuTiP's integrator to support callable derivative instead of only QobjEvo. (#2916, by Eric Giguère (Ericgig))
```
Ericgig: This will break some stuff (qutip-jax, qutip-qoc), maybe user stuff if they implement their own integrator (very doubtful that anyone has).
But it should not be released as is. Making it backward compatible would not have been complicated, it just slipped out of mind.

doc/changelog.rst
Ericgig:
```suggestion
- Increase qfunc validity range to show coherent(alpha>=13) properly (#2932, reported by Dong-Sheng Liu (DS-Liu)
```
When fixing bug by reported by the community, I like to thank them like this.
My name appear often enough, better take the place to encourage participation from the community.

Ericgig: Or maybe listing both?

Ericgig (commented):
Look good, but 2916 should not be included as is.

doc/changelog.rst
veronikakurth: This is a great idea. I listed you both. :)
For the future, maybe we can move attributions to _Release notes_ only - I know it's not the most read thing, but there, we may also tag users, and they will get a notification about it.

doc/changelog.rst
veronikakurth: Removed it from the logs and `prepare_qutip_release_5.3.1` branch. 👌 Thank you for the explanation!

doc/changelog.rst
```diff
@@ -6,6 +6,51 @@ Change Log
...
 .. towncrier release notes start
 
+
+QuTiP 5.3.1 (2026-08-03)
+========================
+
+Features
+--------
+
+- Optimize "rouchon" method for stochastic solvers (#2936, by Eric Giguère (Ericgig))
```
Ericgig, line 16: I feel `Qobj.data_as("csr_array")` support also belong here. The details about our internal status should stay in misc, but that is a new options that some users could be interested in.

Ericgig (approved):

doc/changelog.rst
veronikakurth: That's a good point, thank you! Added it.

PR #2964, opened 2026-07-31
Fix spin_q_function returning nan for large spin

### LINKED ISSUE #2962: spin_q_function / spin_wigner overflow to inf for large spin (j ≳ 550)
### Bug Description

`spin_q_function` evaluates `binom(2*j, j+m)` in floating point and multiplies it by `cos(θ/2)**(2*(j+m)) * sin(θ/2)**(2*(j-m))`. The product is O(1), but the two factors are not individually representable: the binomial overflows to `inf` while the powers underflow, so the term becomes `inf` (or `nan`) for large spin.

`binom(2*j, j)` leaves the double range at roughly `2*j ≳ 1100`, i.e. **j ≳ 550** (Hilbert dimension ≳ 1100).

`spin_wigner` goes through the same coefficients.

### Code to Reproduce the Bug

No qutip build needed — this is the arithmetic from `wigner.py` lines 1009 and 1014:

```python
from scipy.special import binom
from math import cos, sin, pi, sqrt

j, theta = 600, pi / 2
term = binom(2*j, j) * cos(theta/2)**(2*j) * sin(theta/2)**(2*j)

print(binom(2*j, j))                  # inf
print(cos(theta/2)**(2*j))            # 2.41e-181
print(term)                           # inf
print(1/sqrt(pi*j))                   # 0.023032943298089033  <- true value
```

### Code Output

```shell
inf
2.4099198651030816e-181
inf
0.023032943298089033
```

### Expected Behaviour

A finite value. The coefficient is bounded, so the intermediates should not decide the range — evaluating the term in log space (`gammaln`/`betaln`, or `exp` of the summed logs) keeps it representable for any `j`.

This is the same shape as the `clebsch` bug fixed in #2954, where `C` and `S` each left the float range although their product was at most 1.

### Your Environment

qutip master, scipy 1.x. Reported from the arithmetic rather than an end-to-end run — I can't currently build qutip locally, so I haven't confirmed the resulting array contents in `spin_q_function` itself. Happy to send a patch if the log-space approach looks right to you.

uditjainstjis:
Fixes #2962.

### The problem

`spin_q_function` multiplies `binom(2*j, j+m)` by `cos(θ/2)**(2*(j+m)) * sin(θ/2)**(2*(j-m))`. The product is bounded, but neither factor is individually representable: the binomial overflows to `inf` around **j = 550** while the powers underflow to zero.

```python
>>> rho = Qobj(np.eye(1101) / 1101)      # j = 550, maximally mixed
>>> spin_q_function(rho, [np.pi/2], [0.0])[0]
array([[nan]])                           # exact value is 1/1101
```

j = 100 and j = 400 are fine; j = 550 and above are `nan`.

### The fix

Each term is accumulated in log space and exponentiated once — `gammaln` for the binomial, `xlogy` for the powers. `xlogy` is what makes the endpoints work: it keeps a zero exponent at zero rather than evaluating `0 * -inf`, so `theta = 0` and `theta = pi` behave.

### Verification

For the maximally mixed state the Q function is exactly `1/(2j+1)`, which gives an exact oracle:

| j | before | after |
| --- | --- | --- |
| 100 | 0.004975… ✓ | ✓ |
| 400 | 0.001248… ✓ | ✓ |
| 550 | `nan` | ✓ |
| 600 | `nan` | ✓ |
| 800 | `nan` | ✓ |

Correct to **5e-16** for j = 2 … 800, at `theta` = 0, π/4, π/2, 3π/4 and π.

No regression: against the previous implementation on random density matrices for j = 0.5 … 20, the largest difference is **3.6e-15**. The **161** tests in `test_wigner.py` pass, and a regression test is added for j = 550/600/800.

`spin_wigner` uses `_rho_kq` rather than these coefficients, so it is unaffected — I checked before widening the change.

*Disclosure: written with AI assistance, per the project's policy. Every number above was produced locally.*


Ericgig (approved):
Look good.
Thank you!

Fails are unrelated, the fix / cause is already merged in master.

PR #2968, opened 2026-08-04
qfunc using log space

Ericgig:
**Description**
Better version of https://github.com/qutip/qutip/issues/2932.
`qfunc` uses larges numbers that cancel each other, but independently diverge:

    ~ exp(-x**2) x**n / sqrt(n!)

The first to break was the factorial function, so in #2932, I used an approximation for it at large number.
That approximation is not perfect, from when we start using it (n=170) it's precise to under 1e-6.
But it's bad at small values, so I needed to use different code for small and larges values.

I rewrote it to do the computation in log scale: less branching, no approximation: `gammaln` is valid to machine precision, good with even larger state, no `cutoff` parameter only useful for tests.
It's speed is of a similar order of magnitude.

**AI Tools Usage Disclosure**
  - Gemini: Discussion to find error and comments.

### DIFF
--- a/qutip/tests/test_wigner.py
+++ b/qutip/tests/test_wigner.py
@@ -232,59 +232,31 @@ class TestHusimiQ:
         np.testing.assert_allclose(naive, qutip.QFunc(xs, ys, g)(state))
 
     def test_large_size(self):
-        xs = np.linspace(0, 30, 31)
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Nice implementation.

PR #2976, opened 2026-08-14
Backend support for heom solver

Chinmay-Tangal:
**Description**
This PR introduces a dynamic plugin architecture to `bofin_solvers.py` so we can easily swap out backends for the HEOM solver.

The main change is overloading the `__new__` factory method in the `HEOMSolver` base class. by doing this, we can now pass a `backend` argument and the code will lazy load the corresponding backend module dynamically using `importlib`.

**Advantages of this setup**
It makes it super easy to add future backends (like GPU or distributed solvers) down the line. you just drop in the new backend file and it registers itself. No need to touch the core solver logic.

**The PETSc Backend**
To utilize this new plugin architecture, this PR also introduces `backend_petsc.py` as the first dynamically loaded solver backend. By simply passing `backend="petsc"` to the standard `HEOMSolver ` call. 

Which introduces:
***Distributed Hierarchy Assembly:***
Instead of constructing the massive HEOM superoperator matrix in the memory of a single machine, `PETScHEOMSolver` distributes the matrix rows across multiple MPI ranks

***Custom PETSc Integrator:***
It implements `IntegratorPETSc` (a custom subclass of QuTiP's Integrator API). This allows the solver to seamlessly offload time-evolution to PETSc's Time Stepping (TS) module (efficiently handling stiff ODEs via implicit BDF solvers) and steady-state calculations to PETSc's Krylov Subspace (KSP) iterative linear solvers.

(Note: This contribution is under the Google Summer of Code 2026)

**AI usage disclosure**
 Ai tools were used to build understanding and intuition.


doc/guide/heom/petscbackend.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Because the coupled ordinary differential equations generated by HEOM are typically 
+stiff, the default PETSc integrator is the Backward Differentiation Formula (BDF). 
+You can customize the internal PETSc routines using an ``options`` dictionary 
+passed to the solver:
+
+.. code-block:: python
+
+    options = {
+        "ts_type": "bdf",       # Time Stepping method (e.g., 'bdf', 'cn', 'beuler')
```
Ericgig: Why isn't it `method`?
The `method` option decide which equation are used to solve the ODE problem.
ie. `options={"method":"bdf"}` to use the BDF method in mesolve.

`"method" control the ODE equation in every solver, except HEOM with the petsc backend, were it's renamed "ts_type".` It doesn't sound good.

doc/guide/heom/petscbackend.rst
```diff
@@ -0,0 +1,177 @@
...
+        pc_type="none",         # Preconditioner type
+        ksp_rtol=1e-8, 
+        ksp_atol=1e-8
+    )
+
+Running with MPI
+----------------
+
+To fully utilize the PETSc backend, your scripts must be executed using an MPI 
+launcher (e.g., ``mpiexec``, ``mpirun``, or ``srun``) rather than a standard
```
Ericgig:
```suggestion
launcher (e.g., ``mpiexec``, ``mpirun``, or ``srun``) with the standard 
```
`python` is still called...

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+
+class _PETScRHS:
+    """Wrapper around a PETSc.Mat that provides the interface expected
+    by IntegratorPETSc and HEOMSolver.
+    """
+
+    def __init__(self, mat, sys_size):
+        self.mat = mat
+        self.sys_size = sys_size
+        self.isconstant = True
```
Ericgig, line 27: This need to be well documented.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+                        )
+
+        mat.assemblyBegin()
+        mat.assemblyEnd()
+
+        return _PETScRHS(mat, self._sup_shape)
+
+    def steady_state(
+        self,
+        ksp_type="gmres", pc_type="none", ksp_rtol=1e-8, ksp_atol=1e-8,
```
Ericgig: No documentation on these input...

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+    def __call__(self, t):
+        return self
+
+
+class PETScHEOMSolver(HEOMSolver):
+    """
+    PETSc-based solver for the Hierarchical Equations of Motion (HEOM).
+    """
+    
+    name = "petsc"
```
Ericgig:
```suggestion
    name = "petsc heomsolve"
```
The physical equation being solved should be clear.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+    solver_options = HEOMSolver.solver_options.copy()
+    solver_options["method"] = "petsc"
```
Ericgig: This is not checked anywhere...

`HEOMSolver(..., options={"method":"bdf"}, backend="petsc")` will try to use scipy's bdf instead of petsc's.

You could create a `_get_integrator` method that always use petsc's integrator and redirect "method" to "ts_type". Or at least fail before building the RHS.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+        "rtol": 1e-6,
+        "ksp_type": "bcgs",
+        "pc_type": "bjacobi",
+        "ksp_atol": 1e-8,
+        "ksp_rtol": 1e-6,
+        "store_ados": False,
+    }
+
+    support_time_dependant = False
+    supports_blackbox = False
```
Ericgig: No longer needed.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+            self.rhs_obj = system.rhs
+        elif hasattr(system, "mat"):
+            self.rhs_obj = system
+        elif isinstance(system, QobjEvo) or (hasattr(system, "rhs") and isinstance(system.rhs, QobjEvo)):
+            self.rhs_obj = None
+        else:
+            raise TypeError("Unsupported system type passed to IntegratorPETSc.")
+
+        self._is_set = False
+        self._back = (np.inf, None)
```
Ericgig: That's for `mcstep`. I don't think HEOM monte carlo simulation will ever be a thing.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+        self.ts.setRHSFunction(PETSc.TS.computeRHSFunctionLinear)
+        self.ts.setRHSJacobian(PETSc.TS.computeRHSJacobianConstant, self.mat, self.mat)
+
+        self.ts.setType(self.options.get("ts_type", "bdf"))
+        self.ts.setTimeStep(self.options.get("dt", 1e-4))
+        self.ts.setMaxSteps(self.options.get("max_steps", 100000))
+        self.ts.setTolerances(
+            atol=self.options.get("atol", 1e-8),
+            rtol=self.options.get("rtol", 1e-6)
+        )
```
Ericgig: It would be better to not use `.get`. All options are filled in `__init__`, if any are missing, it's probably a typo in this PR.
Let's avoid using invisible default.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+        gathered_np = self.vec_seq.getArray()
+        if copy:
+            gathered_np = gathered_np.copy()
+
+        if self._gather_full:
+            shape = (self.mat.getSize()[1], 1)
+        else:
+            shape = (self.rhs_obj.sys_size, 1)
+
+        state_data = _data.Dense(gathered_np.reshape(shape))
```
Ericgig, line 575: This will always do a copy, so the copy higher is not needed.

Also, it would be better not to gather here...
It's probably possible to compute expectation values in a distributed way.
Probably too big a change for qutip/qutip, but we do it in qutip/qutip-cuquantum.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+        "ts_type": "bdf",    
+        "ts_adapt": "basic", 
+        "dt": 1e-4,          
+        "max_steps": 100000,
+        "atol": 1e-8,
+        "rtol": 1e-6,
+        "ksp_type": "bcgs",
+        "pc_type": "bjacobi",
+        "ksp_atol": 1e-8,
+        "ksp_rtol": 1e-6,
```
Ericgig, line 437: These options are not documented anywhere.

qutip/solver/heom/backend_petsc.py
```diff
@@ -0,0 +1,407 @@
...
+    def options(self):
+        """Supported options by PETSc TS."""
+        return self._options
+
+    @options.setter
+    def options(self, new_options):
+        Integrator.options.fset(self, new_options)
+
+    def _prepare(self):
+        self.PETSc = PETSc
```
Ericgig: Why not just use the global instance?

Ericgig (commented):
This approach is... shaky?

It makes it so that qutip's class are getting closer to abstract interface which can be taken over by different specialized backends... Not a bad idea in itself, but there are issues with the documentation. 

`PETScHEOMSolver` is not in the API doc so it's documentation is only in the guide. And it's not complete there. You list some ODE equation: "bdf", "cn", "beuler", but is that the full list of them? Where is the full list? What are other possible values of "ts_adapt"? HEAMSolver can solve time-dependant system, but not the petsc version. Since it uses the same docstring and API entry, it could cause confusion.

Ericgig (commented):
This approach is... shaky?

It makes it so that qutip's class are getting closer to abstract interface which can be taken over by different specialized backends... Not a bad idea in itself, but there are issues with the documentation. 

`PETScHEOMSolver` is not in the API doc so it's documentation is only in the guide. And it's not complete there. You list some ODE equation: "bdf", "cn", "beuler", but is that the full list of them? Where is the full list? What are other possible values of "ts_adapt"? HEAMSolver can solve time-dependant system, but not the petsc version. Since it uses the same docstring and API entry, it could cause confusion.

doc/guide/heom/petscbackend.rst
Chinmay-Tangal: Fixed. `PETScHEOMSolver` now overrides `_get_integrator()` which always selects `IntegratorPETSc` regardless of `options["method"]` and internally maps `"method" ` -> `"ts_type" ` via `options.setdefault("ts_type", user_method)`

doc/guide/heom/petscbackend.rst
Chinmay-Tangal: fixed

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: fixed. now it full docstring covering all parameters.

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: added the required docstring

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: removed both attributes

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: Removed. `IntegratorPETSc` already overrides `mcstep()` with a hard raise `NotImplementedError,` so _back was never reachable anyway.

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: The redundant conditional copy before `_data.Dense(...)` has been removed — `get_state` now always calls `.copy()` once on the raw NumPy array before wrapping it. Regarding distributed expectation values without gathering: agreed this is the right long-term direction, but scoping it out of this PR to keep the change reviewable. Happy to open a follow-up issue tracking the distributed gather improvement, referencing the qutip-cuquantum approach.

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: fixed (that was a typo)

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: fixed. Removed `self.PETSc = PETSc` from `_prepare().` All references in `_prepare(), get_state(), integrate()`, and the scatter setup now use the module-level `PETSc` directly.

qutip/solver/heom/backend_petsc.py
Chinmay-Tangal: fixed exactly as suggested

PR #2978, opened 2026-08-17
Remove cython 0.29 operator interface

Ericgig:
**Description**
Remove code added to support cython 0.29.X and 3.X in parallel.
Mostly cython 0.29.X do not use inverted operator `__radd__` but call `__add__` with reversed inputs: `__add__(other, self)`.
But with cython 3.X, reversed operator are supported.

We had cdef class which had both `__radd__` and `__add__` supporting both order, other that forced the old api with an option.

**Related issues or PRs**
Discussed in #2958

**AI Tools Usage Disclosure**


veronikakurth (approved):
I looked through all the changes and understood why it's necessary if we want to have Cython 3.x support. Thank you very much for taking care of it; I think the code becomes more Pythonic and much more succinct. The changes (which are mostly splitting the original operator's code into its left and right versions) look consistent to me with respect to how the code already was, maths- and types-wise. But maybe someone else wants to have a second look. This person is very welcome. :) 

P.S. I think your changes to handling Cython in `.github/workflows/tests.yml` should take precedence, so I undid related changes in the PR on upgrading dependencies.

.github/workflows/tests.yml
```diff
@@ -189,16 +189,15 @@ jobs:
...
-          fi
+
+          python -m pip install cython filelock
 
           python -m pip install -e . -v --no-build-isolation
 
           if [[ "${{ matrix.nocython }}" ]]; then
             python -m pip uninstall cython -y
+          elif [ "${{ matrix.oldcython }}" ]]; then
+            python -m pip install cython==0.29.36 filelock
```
Mayank447:
```suggestion
          elif [[ "${{ matrix.oldcython }}" ]]; then
            python -m pip install cython==0.29.36 filelock
```

This was not being tested earlier

Mayank447 (changes requested):

.github/workflows/tests.yml
```diff
@@ -80,7 +80,7 @@ jobs:
             python-version: "3.11"
             numpy-build: ">=1.22.0,<1.23.0"
             numpy-requirement: ">=1.22.0,<1.23.0"
-            scipy-requirement: ">=1.9,<1.10"
+            scipy-requirement: ">=1.11,<1.12"
```
veronikakurth: @Ericgig Can we resolve the merge conflict here by overriding SciPy's dependency from the master branch? 
I merged the PR with SciPy's upgrade, and there we decided on 1.13 as SciPy's floor, which also proved to work with Python 3.11. I think if we left 1.11 here, for Python 3.11 we would not be able to use some features of SciPy we want to use and hence would need version-based branching in code.
P.S. Whatever we do with Python 3.11, I guess it's not so important given that we would also try to drop support for it in 5.4, but for consistency, I guess 1.13 should be fine now. :)

Mayank447 (approved):

PR #2979, opened 2026-08-18
Test split build and run environment

Ericgig:
**Description**
In the test workflow, we installed QuTiP in editable mode with build-time dependencies, then modified the environment in-place to the run dependencies. I rewrote it so a QuTiP wheel is built in a virtual environment separate from the run time one. Modifying the environment in-place could cause conflict, module could be updated without us realizing and we had to look deep in the log to see the build versions.

I most tests cases, the wheel is build with the latest of everything (numpy, scipy, cython) since it's what we use when building wheels for releases. But I kept at least one test building with the older version on each OS.

I removed most boolean options: `oldcython, includempi, semidefinite, oldmpl, condaforge, coverall` for more general ones:
- `qutip-optional` 
  Optional dependencies for qutip:
  `qutip-optional = "[semidefinite]"` => `pip install qutip[semidefinite]` 
  replace `semidefinite` option.
- `pytest-extra-options`:
  Extra pip install:
  `pypi-extra-pkgs: "matplotlib==3.8.* "`. replace oldmpl
- `conda-extra-pkgs`
  Same for conda: includempi managed with this.

Conda forge and coverall are always on.

The mpi environment variable is also always set, even when mpi is not installed, in which case it does nothing.
(This is the drawback of not have a dedicated inlcudempi. Could be managed by some bash code.)

I added / modified tests cases, but it's temporary as it will be affected by #2978, #2951. 
Since the install is no longer in editable mode, I had to use Mayank447's method from #2958 to find the installed tests files.

Oldest matplotlib tested is 3.8, but we officially support 3.6, probably true, but hard to install without causing conflict with other packages...

**AI Tools Usage Disclosure**
  - Gemini: made a review.


.github/workflows/tests.yml
```diff
@@ -285,7 +293,7 @@ jobs:
           #     force coloured output in the terminal
 
       - name: Upload to Coveralls
-        if: ${{ matrix.coveralls }}
+        # if: ${{ matrix.coveralls }}
         env:
           GITHUB_TOKEN: ${{ secrets.github_token }}
           COVERALLS_FLAG_NAME: ${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.case-name }}
```
Mayank447, line 317: Coverage is not getting reported any longer on coveralls due to the working directory change.

https://coveralls.io/builds/81230460

.github/workflows/tests.yml
```diff
@@ -31,105 +31,133 @@ jobs:
...
             pypi: 1
+
+            qutip-optional: "[full]"
             # Install mpi4py to test mpi_pmap
             # Should be enough to include this in one of the runs
-            includempi: 1
+            # Both mpich and openmpi version 4 have caused problems in the past
+            # so we use openmpi version 5. Openmpi 5 and python 3.13-compatible
+            # version of mpi4py are currently only available on conda-forge:
+            conda-extra-pkgs: "openmpi>=5.0 mpi4py"
```
Mayank447:
```suggestion
            conda-extra-pkgs: '"openmpi>=5.0" mpi4py'
```

Bash would treat `>` as a stdout redirect.

.github/workflows/tests.yml
```diff
@@ -31,105 +31,133 @@ jobs:
...
+            qutip-optional: "[full]"
             pypi: 1
-            # Install mpi4py to test mpi_pmap
-            # Should be enough to include this in one of the runs
-            includempi: 1
 
-          - case-name: p312, numpy fallback
+          - case-name: p314 ubuntu arm mkl, version-2
+            os: ubuntu-24.04-arm
+            python-version: "3.13"
```
Mayank447: There is a mismatch in casename and python version. Also mkl doesn't exist on arm

.github/workflows/tests.yml
```diff
@@ -152,74 +180,64 @@ jobs:
...
-          else
-            python -m pip install cython filelock
+            conda install nomkl $CORE_DEP
           fi
 
-          python -m pip install -e . -v --no-build-isolation
+          mv qutip/qutip*.whl .
+          WHEEL=$( echo qutip*.whl )
+          echo $WHEEL
+          python -m pip install $WHEEL${{ matrix.qutip-optional }} -v
```
Mayank447:
```suggestion
          python -m pip install "$WHEEL${{ matrix.qutip-optional }}" -v
```

In this case, this shouldn't matter. But as a good practice, we should quote variable substitution in bash. There are few other places in the PR, where we should make this change.

Mayank447 (commented):
LGTM, few minor things

.github/workflows/tests.yml
```diff
@@ -31,105 +31,133 @@ jobs:
...
+            numpy-run: "==2.3.*"
+            scipy-run: "==1.16.*"
+            qutip-optional: "[full]"
             pypi: 1
-            # Install mpi4py to test mpi_pmap
-            # Should be enough to include this in one of the runs
-            includempi: 1
 
-          - case-name: p312, numpy fallback
+          - case-name: p314 ubuntu arm mkl, version-2
```
Ericgig:
```suggestion
          - case-name: p313 ubuntu arm, version-2
```

.github/workflows/tests.yml
Ericgig: That's annoying...
I guess this step simply can't find the generated files...
Moving them could fix the issue? Will try tomorrow.

Migrating to a src layout would also help.

Ericgig (commented):
Thank you for the review.

.github/workflows/tests.yml
Mayank447: Merging master into this branch should resolve this. I added coverage paths in pyproject.toml in #2958

But this is more of a temporary fix, we should look into migrating to a src layout (or atleast moving the `tests/` directory outside `qutip/`

Ericgig: Qutip family packages import utilities from these tests. So I will need to move some function outside the tests... 

I was looking but got distracted... Give me a week or two.

Ericgig: Looking at the coverage, it now works, but the different action files are no longer merged:
- python3.13/site-packages/qutip/core/qobj.py
- python3.14/site-packages/qutip/core/qobj.py
- envs/test/Lib/site-packages/qutip/core/qobj.py

All different entries on coveralls. Trying to find a solution. Pretty much trying AI suggestions until something works...

Ericgig: It seems it was already like that.
So coverall fixes will wait.

.github/workflows/tests.yml
```diff
@@ -31,105 +31,134 @@ jobs:
...
             # Should be enough to include this in one of the runs
-            includempi: 1
+            # Both mpich and openmpi version 4 have caused problems in the past
+            # so we use openmpi version 5. Openmpi 5 and python 3.13-compatible
+            # version of mpi4py are currently only available on conda-forge:
+            conda-extra-pkgs: '"openmpi>=5.0" mpi4py'
 
-          - case-name: p314 ubuntu arm
-            os: ubuntu-24.04-arm
+          - case-name: no conda, version-1
```
veronikakurth: Thank you very much for taking care of this restructuring.
I have a question about `version-x` in the names of some jobs: do we need them - are the job names not unique enough? Or is there some semantics behind it that I miss?

qutip/core/data/linalg.py
```diff
@@ -14,7 +14,7 @@ def inv_dense(data, /):
     if data.shape[0] != data.shape[1]:
         raise ValueError('Cannot compute the matrix inverse'
                          ' of a nonsquare matrix')
-    return Dense(scipy.linalg.inv(data.as_ndarray()), copy=False)
+    return Dense(scipy.linalg.inv(data.as_ndarray()), copy=None)
```
veronikakurth, line 17: Are these changes meant to be temporary, as mentioned in the description, or are they there to stay? 

I'm curious what the reason was behind changing it to a more permissive `None`, whether there were errors because of the prohibitive semantics of `copy=False` for NumPy 2.x. I had to think about the migration to the sparse array API: I remember having issues with memory sharing there when using`copy=False` (reasons being passing a sparse array with int64 indices where CSR/Dia would expect int32. But that's another topic)

So I wanted to ask you about whether we have some implicit principles of handling the `copy` parameter across the codebase. Namely, are there some cases when we want to be especially strict upstream and set `copy` to `False` (in NumPy's 2.x semantics) when creating `Data` objects?

I guess the prerequisite of being strict is being sure that it would be possible to share the memory.

.github/workflows/tests.yml
Ericgig: It's version minus 1, so numpy and scipy are older version as the -X increase.
And yes, with that naming scheme we should update the test with each numpy / scipy release, but it's already something we should do.

The test name is { os }{ python_version }{ case-name } so having "p314 ubuntu arm" as the case name is redundant, and I don't want it empty...

qutip/core/data/linalg.py
Ericgig: "Temporary" was for the tests case, testing with scipy 1.11 etc. Already removed when making the changes.

There was new warnings with `copy=False` needing a copy, not sure why we missed them before. 
My guess is that depending on scipy version, the index types could be kept or upgraded to int64.
If any supported scipy version can change the array type in `scipy.sparse.linalg.inv` , we have no choice to have `copy=False`.

I can't think of any place where we want to be strict. Not making copy is for optimization, having different Data object sharing memory can cause issue (`imul` on one changing the other unexpectedly) if not strictly managed. Such as inside a single function with only one reference kept at the end or cython code that we know users will not have access to private attribute...
But outside of that, `copy=None` is good. Also most of the data layer stuff was written before numpy 2.0, so the `copy=False` were set with the copy only if needed in mind.

.github/workflows/tests.yml
Ericgig: I am open to better naming scheme if you have an idea.

qutip/core/data/linalg.py
veronikakurth: Thank you for the explanation!

I think we will need to double-check how we use the flag where SciPy sparse objects are expected as input once SciPy drops the sparse matrix interface, and we have to remove it as well; any `copy=False` will throw an error if qutip is built with 32-bit integers for indices and an incoming object has 64-bit indices...
But that's another topic, sorry. We can discuss it in another context.

.github/workflows/tests.yml
veronikakurth: I see what you mean, thanks for the explanation. I'd be in favour of having information about scipy/numpy build-and-runtime versions in the name since this is a piece of information that is most likely to be checked in the logs... (Cython's version is also important, but we have to draw a line somewhere) 

However, I don't have an idea of how to make it more lightweight... Since something like this looks quite informative, but may be too crowded to be displayed nicely in the UI:

`ubuntu-latest | py3.12 | mkl |  build[np==2.0.*, sp==1.13.*] | run[np==2.1.*, sp==1.14.*]`

veronikakurth:
This test keeps failing on the latest Windows; strangely, it is so persistent even after rerun pipelines... 
`FAILED C:\Miniconda\envs\test\Lib\site-packages\qutip\tests\test_utilities.py::TestFitting::test_espira_II[False] - assert 7.826549132741106e-06 < 1e-08`

Ericgig:
`TestFitting` is quite unstable... It runs with a fix seed and if I remove that fixed seed, at least one of the fitting function fail about 25% of the test runs....

With the change of version (numpy changed the RNG at some point in recent years), that one seed no longer works...

veronikakurth (approved):
The changes look good to me, thank you! Let's see them in action (no pun intended 🙃 ); not only does the test configuration have a cleaner separation, but also for daily dev workflows (namely, using Actions), I personally see an immediate improvement: the names of the jobs are definitely more informative (as an example, it's easy to see where MKL dependencies are installed. This I found challenging before). I would still like to see more info about the scipy/numpy versions used for build/runtime, also in the job name, but I acknowledge that it's difficult to make it concise.

Mayank447 (approved):

PR #2986, opened 2026-08-20
Fix `bures_dist` returning NaN for a state compared with itself

### LINKED ISSUE #2985: `bures_dist` return Nan
### Bug Description

`bures_dist` can return Nan when computing the distance between the the same object.

### Code to Reproduce the Bug

```shell
import qutip as qt
dm = qt.rand_dm(3, seed=465)
qt.bures_dist(dm, dm)
```

### Code Output

```shell
np.float64(nan)
```

### Expected Behaviour

Should return `0`.

### Your Environment

```shell
Development version.
```

### Additional Context

_No response_

adisivaprasad:
Fixes #2985.

## Problem

`bures_dist(A, B) = sqrt(2 * (1 - F))` where F is the fidelity. For numerically identical states, the eigenvalue-sum in `fidelity` can come out as `1 + 1e-16` due to floating-point error, making the argument of `sqrt` negative and producing `NaN`:

```python
import qutip as qt
dm = qt.rand_dm(3, seed=465)
qt.bures_dist(dm, dm)  # nan, expected 0
```

The same numerical-overrun defect also affects `bures_angle`, where `arccos(1 + 1e-16)` is `NaN`.

## Fix

Clamp the computed value in the same style `hellinger_dist` already uses (`np.maximum(0, ...)` there — see the existing comment about avoiding NaN from numerical error):

- `bures_dist`: `np.sqrt(2 * np.maximum(0, 1 - fidelity(A, B)))`
- `bures_angle`: `np.arccos(np.minimum(1, fidelity(A, B)))`

## Tests

- `Test_bures_dist`: orthogonal states give `sqrt(2)`; identical states give exactly 0, including the issue's reproducer `rand_dm(3, seed=465)`.
- `Test_bures_angle`: identical states give 0 (same numerical path).

I verified the defect and the fix numerically: replicating qutip's fidelity computation on random density matrices shows `fidelity(rho, rho) > 1` at the ~1e-16 level for ~45% of random draws (i.e. `sqrt` of a negative number — the reported `NaN`), and 0 occurrences after clamping.

Also adds a towncrier changelog fragment (`2985.bugfix`).

### DIFF
--- a/doc/changes/2985.bugfix
+++ b/doc/changes/2985.bugfix
@@ -1,0 +1,2 @@
+Fix ``bures_dist`` and ``bures_angle`` returning NaN when comparing a state to
+itself because floating-point error could push the fidelity slightly above one.
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -345,7 +345,10 @@ def bures_dist(A, B):
         B = B.proj()
     if A._dims != B._dims:
         raise TypeError('A and B do not have same dimensions.')
-    dist = np.sqrt(2 * (1 - fidelity(A, B)))
+    # np.maximum() is to avoid nan appearing sometimes due to numerical error
+    # making the fidelity slightly larger than one, which would make the
+    # argument of the square root negative.
+    dist = np.sqrt(2 * np.maximum(0, 1 - fidelity(A, B)))
     return dist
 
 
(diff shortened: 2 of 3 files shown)


Ericgig:
Did not follow the contribution guidelines and PR template...

PR #2988, opened 2026-08-23
Fix docstring parameter names that do not match the signature

karpovantonme:
**Description**

17 docstrings name a parameter the signature does not have.

Most are a rename that never reached the docs: `concurrence` documents `state` while the signature is `rho`, `wigner` documents `state` for `psi`, `set_state` documents `state0` for `state`, `start` the other way round. Two are typos, `superrop` for `superrep` in `rand_super` and `rand_super_bcsz`, and `keys` for `key` in `add_integrator`. Two are leftovers: `molmer_sorensen` documents a `target` argument that is not in the signature at all, and `Qobj.proj` documents `Q` but takes only `self`, so both blocks are dropped rather than renamed.

Found with a docstring checker I maintain, then opened each of the 17 against its signature by hand. Nothing here changes behaviour.

There is an 18th hit in `MCSolver.StateFeedback` that I left out, because it looks like a code bug rather than a docs one and needs a maintainer decision. Filing it separately.

**Related issues or PRs**

None.

**AI Tools Usage Disclosure**
  - `Claude (Anthropic): drafting the patch and this description. The mismatches came from a static checker, and I read all 17 sites against their signatures before including them.`

### DIFF
--- a/doc/changes/2988.doc
+++ b/doc/changes/2988.doc
@@ -1,0 +1,1 @@
+Fix 17 docstrings that name a parameter the signature does not have.
--- a/qutip/animation.py
+++ b/qutip/animation.py
@@ -601,7 +601,7 @@ def anim_schmidt(kets, theme='light', splitting=None,
 
     Parameters
     ----------
-    ket : :class:`.Result` or list of :class:`.Qobj`
+    kets : :class:`.Result` or list of :class:`.Qobj`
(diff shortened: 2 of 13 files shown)


veronikakurth (approved):
Thank you!

Ericgig (approved):

PR #2989, opened 2026-08-23
Fix MCSolver.StateFeedback being built as an open-system feedback

karpovantonme:
**Description**

`MCSolver.StateFeedback` cannot be used at all at the moment:

```python
>>> MCSolver.StateFeedback().open
<built-in function open>
```

`StateFeedback` passes `open=open` to `_QobjFeedback`, and `open` there resolves to the builtin, which is always truthy. The feedback is then built as if the solver evolved a density matrix, while `MCSolver` evolves a ket, so a run dies on the dimensions:

```
ValueError: Provided dimensions do not match the data: (2, 2) vs (2, 1)
```

The same code with `SESolver.StateFeedback()` runs fine.

Second thing in the same three lines: `prop` is in the signature and never reaches the feedback, so `StateFeedback(prop=True).prop` comes back `False`.

The neighbours spell out the intent: `sesolve` passes `open=False, prop=prop`, `mesolve` passes `open=True, prop=prop`. `MCSolver` is documented as a "Monte Carlo Solver of a state vector", so this uses `open=False, prop=prop`. If `prop` was deliberately dropped here, say so and I will cut that half.

The docstring documented `open` as a parameter, which it never was in this signature; that block is replaced with `prop`.

Two tests added, both fail on master.

**Related issues or PRs**

Found alongside #2988, which is docstrings only. This one changes behaviour, so it is separate.

**AI Tools Usage Disclosure**
  - `Claude (Anthropic): drafting the patch, the tests and this description. I reproduced the failure and the fix against qutip 5.3.1 before opening this.`

### DIFF
--- a/qutip/solver/mcsolve.py
+++ b/qutip/solver/mcsolve.py
@@ -955,17 +955,17 @@ class MCSolver(MultiTrajSolver):
         default : Qobj or qutip.core.data.Data, default : None
             Initial value to be used at setup of the system.
 
-        open : bool, default False
-            Set to ``True`` when using the monte carlo solver for open systems.
-
         raw_data : bool, default : False
             If True, the raw matrix will be passed instead of a Qobj.
             For density matrices, the matrices can be column stacked or square
             depending on the integration method.
+
+        prop : bool, default : False
+            Set to ``True`` when the states are propagators.
         """
         if raw_data:
-            return _DataFeedback(default, open=open)
-        return _QobjFeedback(default, open=open)
+            return _DataFeedback(default, open=False, prop=prop)
+        return _QobjFeedback(default, open=False, prop=prop)
 
 
 class _unpack_arguments:
--- a/qutip/solver/mcsolve.py
+++ b/qutip/solver/mcsolve.py
@@ -938,7 +938,7 @@ class MCSolver(MultiTrajSolver):
         cls,
         default: Qobj | _data.Data = None,
         raw_data: bool = False,
-        prop: bool = False
+        open: bool = False,
     ):
         """
         State of the evolution to be used in a time-dependent operator.
@@ -955,17 +955,17 @@ class MCSolver(MultiTrajSolver):
         default : Qobj or qutip.core.data.Data, default : None
             Initial value to be used at setup of the system.
 
+        open : bool, default : False
+            Set to ``True`` when using the monte carlo solver for open systems.
+
         raw_data : bool, default : False
             If True, the raw matrix will be passed instead of a Qobj.
             For density matrices, the matrices can be column stacked or square
             depending on the integration method.
-
-        prop : bool, default : False
-            Set to ``True`` when the states are propagators.
         """
         if raw_data:
-            return _DataFeedback(default, open=False, prop=prop)
-        return _QobjFeedback(default, open=False, prop=prop)
+            return _DataFeedback(default, open=open)
+        return _QobjFeedback(default, open=open)
 
 
 class _unpack_arguments:
--- a/qutip/tests/solver/test_mcsolve.py
+++ b/qutip/tests/solver/test_mcsolve.py
@@ -566,9 +566,6 @@ def test_feedback(func, kind):
 
 
 def test_state_feedback():
-    # MCSolver evolves a ket, so StateFeedback must not be built as an open
-    # system feedback. Before the fix it was created with the builtin `open`,
-    # which is always truthy, and the run failed on the state dimensions.
     psi0 = qutip.basis(2, 0)
     H = qutip.QobjEvo(
         [qutip.sigmaz(), [qutip.sigmax(), lambda t, state: 0.1]],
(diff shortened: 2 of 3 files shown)


qutip/solver/mcsolve.py
```diff
@@ -955,17 +955,17 @@ def StateFeedback(
...
 
-        open : bool, default False
-            Set to ``True`` when using the monte carlo solver for open systems.
-
         raw_data : bool, default : False
             If True, the raw matrix will be passed instead of a Qobj.
             For density matrices, the matrices can be column stacked or square
             depending on the integration method.
+
+        prop : bool, default : False
```
veronikakurth: The changes to the docstring and passing parameters downstream make sense; there was an inconsistency indeed. 

@Ericgig, we don't have tests (yet) to check for expected behaviour when using `StateFeedback` for `MCSolver`, so it's a bit unclear what the `prop` argument would be used for... Would it make sense in the context of a Monte Carlo solver at all? 

Thanks in advance for your feedback.

qutip/tests/solver/test_mcsolve.py
```diff
@@ -565,6 +565,27 @@ def test_feedback(func, kind):
     assert np.all(result.expect[0] > 4. - tol)
 
 
+def test_state_feedback():
+    # MCSolver evolves a ket, so StateFeedback must not be built as an open
+    # system feedback. Before the fix it was created with the builtin `open`,
```
veronikakurth: It's usually a bit redundant to explain how the code used to work before the fix _in comments_, unless it's _really_ important to note down as a dev note for current/future maintainers; let's leave these kinds of comments for changelogs and the PR description.

qutip/tests/solver/test_mcsolve.py
```diff
@@ -565,6 +565,27 @@ def test_feedback(func, kind):
...
+    )
+    solver = qutip.MCSolver(
+        H, c_ops=[qutip.sigmam()], options={"map": "serial"}
+    )
+    result = solver.run(psi0, np.linspace(0, 1, 3), ntraj=2)
+    assert len(result.states) == 3
+
+
+def test_state_feedback_prop():
+    # `prop` was accepted by the signature and then never passed on.
```
veronikakurth: Same remark as before: it's redundant to elaborate on past buggy behaviours in comments to the code.

veronikakurth (changes requested):

qutip/solver/mcsolve.py
```diff
@@ -955,17 +955,17 @@ def StateFeedback(
...
+            Set to ``True`` when the states are propagators.
```
Ericgig: Thank you for finding the bug.
But mcsolve can never create propagator state, it can however evolve open state.
(https://nbviewer.org/urls/qutip.org/qutip-tutorials/tutorials-v5/time-evolution/017_smesolve-jc-photocurrent.ipynb)

So the fix should be to fix the signature no the documentation.

karpovantonme:
Fixed the signature instead, thanks -- `open` is back in it and now actually reaches the feedback, `prop` is gone since mcsolve never propagates. Also dropped the test comments about the old behaviour as @veronikakurth asked. `pytest -k feedback` passes locally

qutip/tests/solver/test_mcsolve.py
```diff
@@ -565,6 +565,25 @@ def test_feedback(func, kind):
     assert np.all(result.expect[0] > 4. - tol)
 
 
+def test_state_feedback():
+    psi0 = qutip.basis(2, 0)
+    H = qutip.QobjEvo(
+        [qutip.sigmaz(), [qutip.sigmax(), lambda t, state: 0.1]],
+        args={"state": qutip.MCSolver.StateFeedback()},
```
Ericgig, line 572: In this test, `qutip.MCSolver.StateFeedback()` could stay a that object and never be resolved to a Qobj and it would not catch it. You can put check in the coefficient function:
```
def _coeff(t, state):
    assert isinstance(state, Qobj)
    ...
```
to confirm it works as expected.

karpovantonme: Done, the coefficient now keeps what it receives and the test asserts every one of them is a ket.

Checking that it actually catches the bug turned up something worth recording: it only does without a default. On the pre-fix code `StateFeedback()` raises `ValueError: Provided dimensions do not match the data: (2, 2) vs (2, 1)` on the first trajectory, while `StateFeedback(default=psi0)` runs clean, because `check_consistency` lands in the rectangular-state branch and takes the dims from the default. So the test is parametrized over both, and the default case at least exercises that path.

Old code: 2 failures. Current branch: 132 passed in `test_mcsolve.py`.

Ericgig (approved):
Thank you!

veronikakurth (approved):

veronikakurth:
@karpovantonme, thank you!
As per the project's rules, we require human authorship of contributions. This must be declared via the AI usage disclosure (which is done in this PR). Additionally, in case of assistance from AI tools, contributors' commits must also follow [the trailing message guidelines](https://github.com/qutip/qutip/blob/master/CONTRIBUTING.md#3-copyright--legal) for copyright reasons. Please make sure it holds true for your PR.

Ericgig:
@karpovantonme 
The last commit message it the only thing blocking the PR... could you reword it according to the guideline?

Issue #2952, opened 2026-07-27
Migrate MKL solver to third-party PARDISO wrapper and enable support of SciPy's sparse arrays


veronikakurth:
Motivated by a discussion with @Ericgig during work on #2938.

Expected pitfalls:
- Running into segfault errors when passing a `csr_array` to the solver: the problem seems to be its 64-bit indices that are not handled in `qutip/_mkl/spsolve.py`

Most likely, we need to refactor our custom MKL Pardiso binding; ideally, we would replace it with a usage of a well-maintained library (PyPardiso, etc) and not manage ctypes pointers ourselves.

veronikakurth:
Working on it at the moment. 
I think `PyPardiso` will not be the right choice for us: it turns out it _doesn't support complex numbers_ at the moment (thanks to the team putting the disclaimer into the README - this saves us time!). 
`pydiso` should be a better fit; I'm trying it out.

PR #2992, opened 2026-08-27
Refactoring MKL-backed sparse solver: integration with pydiso

veronikakurth:
**Checklist**

**Description**

Migrates `qutip/_mkl_spsolve.py` to the usage of `pydiso`'s MKLPardisoSolver - a wrapper handling all the low-level interfacing with Intel's PARDISO. 

Introduces an additional third-party dependency (`pydiso`).

What `pydiso` overtakes
- CSR normalization
- triangular extraction
- index-width selection
- factorization
- handling PARDISO errors
- memory ownership

**Migration/refactoring checklist:**

**Essentials:**

**Dependencies & detection part**

- [x] Change the mechanism of detecting MKL support 
	- Changed semantics: `has_mkl` setting now means `pydiso.mkl_solver` is importable
	- Deprecated: `mkl_lib_location` property and its setter
	    - [ ] **Open question**: Shall we deprecate or remove it all? Is it plausible that users could use this property to refer to a custom MKL location? 
- [x] Update documentation:
	- doc/guide/guide-steady.rst
	- doc/guide/guide-settings.rst
	- [ ] **Open question**: Should we add a new dev documentation piece? I was thinking of providing more information about how we handle `iparm` values (which are qutip-defined, which are pydiso's defaults, which are configurable by the user), and the docstring felt like the wrong place for it. But I think it is important to capture it somewhere.
- [x] CI tests: install MKL solver dependency
	- **After pydiso’s release**, we will be able to use conda-forge’s compiled pydiso package and not do installation from source (currently, we are relying on pydiso’s meson build, and it’s a bit painful in CI). So the future changes will be:
		- pyproject.toml: mkl = ["pydiso>=0.2.0"]
		- tests.yaml: 
			conda install --channel conda-forge \
			    blas=*=mkl \
			    "numpy${{ matrix.numpy-build }}" \
			    "scipy${{ matrix.scipy-requirement }}" \
			    "pydiso>=0.2.0"
	- Currently, there are workarounds in pyproject.toml and tests.yaml. We install pydiso from source because the changes rely on the unreleased code. It is hacky since we need to provide MKL headers and pkg-config metadata, not only the runtime library; it’s something that conda’s wheels would take care of otherwise. 

**Programmatic part**
- [x] Delegate communication to PARDISO via third-party solver wrapper (pydiso) 
- [x] Preserve user interface to 
	- mkl_splu (however, renamed into `MKLFactorization` - more Pythonic)
	- mkl_spsolve. Made keyword arguments explicit.
- [x] Differentiate between pydiso defaults, QuTiP defaults, and user overrides (see the open question in **Update documentation**)
    - [x] Reduced the number of arguments we override: the major subset of those is already handled by `pydiso` the way we expect
- [x] Check the actual usage of user-facing iparm values: `perm`, `max_iter_refine`, `scaling_vectors`, `weighted_matching`
	- Checked this based on how `mkl_spsolve` is used in the qutip’s codebase -  it’s very unlikely that someone is using, e.g. the steady state solver since they are not really documented. Most of the arguments are used with default values.
	- [x] Simplified arguments handling on qutip's side by removing `scaling_vectors` and `weighted_matching` - those are handled already correctly on pydiso's side based on the matrix type.
- [x] Define limitations of pydiso w.r.t. qutip’s use case
	- **Issue**: Before https://github.com/simpeg/pydiso/pull/25, it was not possible to pass `iparm` overrides. This was a problem since we override certain default values of `iparm` in a different way than `pydiso`. With the changes in `pydiso`'s master, we can now pass all parameters the way we want, but it has a drawback that we have to rely on the master version before their release (I pinned it to the commit)
	- **Issue**: ~~However, it is still not possible to use permutations.~~
		- [x] **Open question**: do we want to support user-defined permutations? Currently, `perm` is always `None` upon calling `mkl_spsolve` in qutip (e.g., in HEOM). 
		- [x] **UPD**: with the latest [PR](https://github.com/simpeg/pydiso/pull/27) in `pydiso`, it will be possible to pass permutations. So why not keep supporting them? However, add tests!

**Extra:**

- [x] **Open question**: Are there additional `iparm` overrides we want to allow to be passed by a user? Can we imagine a power user of qutip for such a case? Concretely, somebody who would choose the MKL solver for a steady state and pass arguments corresponding to [docs](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2026-0/pardiso-iparm-parameter.html#PARDISO-IPARM37) (in a more user-friendly form)
   - **Decision**: For now, keep things simple; we can provide more control later. Users can provide a permutation, define the number of refinement iterations, and describe their matrix differently via `hermitian` and `posdef`. 
- [ ] **Open question**: `pydiso` supports more matrix types than our wrapper can infer via the `_mkl_matrix_type` wrapper. Do we want to adapt our matrix type inference to support more matrix types? Are there use cases in qutip for **1, 3, 6** (Namely, `real_structurally_symmetric`, `complex_structurally_symmetric`, `complex_symmetric`)?
- [x] Switched from `csr_matrix` to `csr_array`
	- [x] Do adaptations to `csr_array` in HEOM and respective tests
- [ ] **Open question**: Should we remove the `_mkl/spmv.py` module? It is not imported anywhere. 
- Not tested yet, but it should be possible to use `pydiso` also on Windows.
   - [ ] **Open question**: Shall we add a new CI test case for Windows with MKL?

**What is still lacking:**

Add more tests for:
- [ ] HEOM with MKL
   - [x] Correctness check against steady state solver with SciPy
 - [ ] Using `csr_array` in `qutip/tests/test_mkl.py`: add extra tests. `csr_array` is implicitly tested via HEOM, but we should make steady state tests also more complete
- [ ] Nonsymmetric matrices (i.e., when `hermitian=False`)
	- E.g. in `qutip/tests/test_mkl.py` `hermitian` is set to 1
	- Test the following matrix types:
	   - 11: real, unsymmetric
	   - 13: real, unsymmetric, imaginary part is exactly zero
	   - 13, 2, 4
	   - 
- [ ] User permutations
    - [x] Trivial: identity
    - [ ] Non-trivial: 

- [x] Ensure proper logging: do we want to keep pydiso’s or our logging? Currently is commented out. 
   - Decision: keep pydiso's logging as it's very well handled and lets us know what we already wanted to know before, plus a bit more. 

**Related issues or PRs**

#2952

**AI Tools Usage Disclosure**
  - `Claude Opus 5`: 
     - for the initial discussion about PARDISO and understanding better how qutip interfaces with PARDISO and what should and could be delegated to `pydiso`. 
     - for finding a new way to do MKL detection in `settings.py`
  - `GPT-5.6`: 
     - for rewriting docstrings based on the initial version of those I composed with the info from Intel's docs and from pydiso's source code - it was important to check for iparm defaults; 
     - for adaptation of the documentation where MKL-based solver is mentioned (fully let the model do so); 
     - for finding a way to install `pydiso` from source in CI (see pain points in the description above).
     - for tests for different matrices types 

Speaking of the implementation in general, I used `Claude` as an assistant tool that would review my changes and suggest a better way, which I would iteratively accept (usually writing myself) after understanding every suggestion.  
I definitely learned a lot from this piece of work.


.github/workflows/tests.yml
```diff
@@ -188,7 +188,22 @@ jobs:
             pip install "numpy${{ matrix.numpy-build }}"
             pip install "scipy${{ matrix.scipy-requirement }}"
           elif [[ -z "${{ matrix.nomkl }}" ]]; then
-            conda install blas=*=mkl "numpy${{ matrix.numpy-build }}" "scipy${{ matrix.scipy-requirement }}"
+            conda install --channel conda-forge \
```
veronikakurth, line 238: Most of the changes are temporary: these are needed to install pydiso from source, which turned out to be a bit of a pain in the CI environment.
Once `pydiso` does a release that contains changes we rely on (i.e., for overriding iparms), changes in tests.yml will be reduced to installing pydiso>=0.2.0 from wheels.

mkl-devel, pkg-config, setuptools-scm and patching of `PKG_CONFIG_PATH` / `.pc` will go away...

.github/workflows/tests.yml
```diff
@@ -208,7 +223,14 @@ jobs:
 
           python -m pip install cython filelock
 
-          python -m pip install . -v --no-build-isolation
+            if [[ -z "${{ matrix.pypi }}" && -z "${{ matrix.nomkl }}" ]]; then
+              python -m pip install "setuptools-scm[toml]>=6.2"
```
veronikakurth: As previously mentioned, setuptools-scm installation will also go away once we switch from source installation to wheels; the other changes will stay

doc/guide/guide-steady.rst
```diff
@@ -80,16 +80,59 @@ system. To do so, there are multiple solvers available: ``
...
+QuTiP can use the Intel oneMKL PARDISO sparse direct solver through the
+optional `pydiso <https://github.com/simPEG/pydiso>`_ package. PARDISO can
+provide substantially better performance than SciPy's SuperLU solver for
+large sparse systems.
+
+Install QuTiP with MKL support using:
+
+.. code-block:: bash
+
+   pip install "qutip[mkl]"
```
Ericgig: Does this works?
From my understanding, with pip, it will try to install pydiso from source and will probably raise an error if mkl is missing (or if they don't have a compiler.) 

It might be good to point to pydiso installation instruction with a note to look there if unable to install. 
Or/and maybe show conda's installation. It can have non-python depedencies, so it can install mkl if missing.

veronikakurth: This, of course, doesn't work; I'm sorry, I haven't gone through those AI-generated docs yet. Thanks for pointing this out.

As workarounds in the CI config showed, installing pydiso from source doesn't work smoothly indeed: one has to do additional steps for configuring detection of MKL headers and pkg-config metadata... 
I think we should recommend conda installation, as pydiso does. 
And link to their README for the installation from source - for those who might need it.

PR #2993, opened 2026-08-27
Fix `bures_dist` returning `nan` for identical density matrices

### LINKED ISSUE #2985: `bures_dist` return Nan

Surjasa:
**Description**
`bures_dist` returned `nan` when called with the same density matrix because floating-point roundoff could make the fidelity slightly greater than 1.
Consequently, the argument to the square root became negative.
This change clamps the square-root argument to zero when it is slightly negative, preventing `nan` and returning the mathematically correct distance of zero. A regression test was added for identical density matrices.

**Related issues or PRs**
Fixes #2985 

**Tests**
- `python -m pytest qutip/tests/core/test_metrics.py -q`
- `git diff --check`

**AI Tools Usage Disclosure**
  - `I used Github Copilot and gemini-flash to understand the issue, and existing implementation. Also, to understand and troubleshoot the development environment, and review the code and test changes`

### DIFF
--- a/qutip/tests/core/test_metrics.py
+++ b/qutip/tests/core/test_metrics.py
@@ -195,11 +195,19 @@ class Test_hellinger_dist:
         assert hellinger_dist(rho_sim, sigma) == pytest.approx(dist, abs=tol)
 
 
-
 class Test_bures_dist:
-    def test_state_with_itself(self):
-        state = rand_dm(3, seed=465)
-        assert bures_dist(state, state) == pytest.approx(0, abs=1e-6)
+    def test_state_with_itself(self, state):
+        dist = bures_dist(state, state)
+        assert np.isfinite(dist)
+        assert dist == pytest.approx(0, abs=1e-12)
+
+    def test_state_with_itself_pure_and_mixed_edges(self):
+        psi = basis(2, 0)
+        rho = qeye(2) / 2
+        assert np.isfinite(bures_dist(psi, psi))
+        assert bures_dist(psi, psi) == pytest.approx(0, abs=1e-12)
+        assert np.isfinite(bures_dist(rho, rho))
+        assert bures_dist(rho, rho) == pytest.approx(0, abs=1e-12)
 
 
 class Test_average_gate_fidelity:
--- a/qutip/tests/core/test_metrics.py
+++ b/qutip/tests/core/test_metrics.py
@@ -194,10 +194,14 @@ class Test_hellinger_dist:
         assert hellinger_dist(rho, sigma) + tol > dist
         assert hellinger_dist(rho_sim, sigma) == pytest.approx(dist, abs=tol)
 
+
+
 class Test_bures_dist:
     def test_state_with_itself(self):
         state = rand_dm(3, seed=465)
         assert bures_dist(state, state) == pytest.approx(0, abs=1e-6)
+
+
 class Test_average_gate_fidelity:
     def test_identity(self, dimension):
         id = qeye(dimension)
--- a/qutip/tests/core/test_metrics.py
+++ b/qutip/tests/core/test_metrics.py
@@ -194,7 +194,10 @@ class Test_hellinger_dist:
         assert hellinger_dist(rho, sigma) + tol > dist
         assert hellinger_dist(rho_sim, sigma) == pytest.approx(dist, abs=tol)
 
-
+class Test_bures_dist:
+    def test_state_with_itself(self):
+        state = rand_dm(3, seed=465)
+        assert bures_dist(state, state) == pytest.approx(0, abs=1e-6)
 class Test_average_gate_fidelity:
     def test_identity(self, dimension):
         id = qeye(dimension)
--- a/doc/changes/2985.bugfix
+++ b/doc/changes/2985.bugfix
@@ -1,0 +1,1 @@
+Fixed ``bures_dist`` returning ``nan`` for identical density matrices.
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -345,7 +345,7 @@ def bures_dist(A, B):
         B = B.proj()
     if A._dims != B._dims:
         raise TypeError('A and B do not have same dimensions.')
-    dist = np.sqrt(2 * (1 - fidelity(A, B)))
+    dist = np.sqrt(2 * np.maximum(0, 1 - fidelity(A, B)))
     return dist
 
 
--- a/qutip/core/metrics.py
+++ b/qutip/core/metrics.py
@@ -345,7 +345,9 @@ def bures_dist(A, B):
         B = B.proj()
     if A._dims != B._dims:
         raise TypeError('A and B do not have same dimensions.')
-    dist = np.sqrt(2 * np.maximum(0, 1 - fidelity(A, B)))
+    if A == B:
+        return 0
+    dist = np.sqrt(2 * np.maximum(0, 1 - np.real(fidelity(A, B))))
     return dist
 
 


Ericgig:
Hi @Surjasa,
Thank you for your contribution.
Could you tell us more about you? Why are you interested in this project?

We use "Good first issue" to get to know new contributors, but all I get is that you use AI, which is not very unique these days...

darksoul-240:
Hi @Ericgig, so I was going through the PR which @Surjasa had made, and tracing down the issue going through your docs, where fidelity() seems to be returning a value slightly greater than one due to floating point round-off. Since the value of Fidelity is mathematically bounded between 0 and 1, wouldn't it be preferable if we could clamp the value at the result of fidelity() instead of only protecting the square root argument in bures_dist. This would also prevent the same numerical violation from propagating to other calculations. Would this be better to handle issue, or could there be a reason to have it in bures_dist itself

Surjasa:
Hi, @Ericgig, I am a CS undergrad student currently in 3rd year, I have done some quantum computing courses(Linkedin: https://www.linkedin.com/in/anika-islam-surjasa-/), and want to do research in this line that's why I chose this project to contribute. I want to learn how things work in this line. Also, I don't have the best idea of this project's code yet, that's why i had to use Ai to understand the issue and code. But I reassure you, I changed the code after totally understanding it, nothing was done automatically by any Ai.

Surjasa:
Hi @darksoul-240, I clamped 'bures_dist' in the same way as I saw `hellinger_dist` sqrt is clamped. I think what you said about clamping the value at the result of fidelity() instead of only protecting the square root argument in `bures_dist`, seems like the better way to go, as fidelity should be between 0 and 1, also many other metrices uses it, and if it is allowed to leak a value just above 1, then other calculations can become invalid or inconsistent. So, I think that we can keep the clamp on `bures_dist` because it is protecting the specific sqrt argument from going negative due to that tiny roundoff, and lastly add clamp at fidelity to restore its valid domain.

Ericgig (commented):
@Surjasa, I wish you success in your studies.

I don't mind the clamping being in either `bures_dist` or `fidelity`. Having it once in `fidelity` would centralize the solution, but fidelity returning number slightly higher than one should not be much of an issue when used directly and there is research using non-physical states (non-hermitian, trace != 1). 

Anyway, the change in `bures_dist` is good, but the tests needs more work:
- Extra empty line before the test.
- Don't fix the seed.
- Tolerance of 1e-6 feels too high.
- Why use a class for one test...

Surjasa:
@Ericgig, Thank you! 
when i was creating this PR it said the code should follow, PEP 8, Python’s official style guide, so just to be sure there is no spacing issue left i ran `python -m pycodestyle qutip/core/metrics.py qutip/tests/core/test_metrics.py` test, and it showed E302 errors, it was caused by missing blank lines around the test class i added, that's why i added the blank lines and checked with `git diff --check`. Yes, for bures_dist test maybe using the tolerance value 1e-12 would have been more appropiate than using 1e-6, a stricter check with tolerance 1e-12 instead of 1e-6 is better, i made a mistake because i tried to follow the other tests. As for adding a new class, i think putting it at module scope as a single regression test would be better if we don't want extra class for just one test. I saw the other matric tests and tried following how they are written, that' s why i added the test before `hellinger_dist test` class. i will keep in mind not to fix the seed. If you want me to fix and add these changes, just let me know. Thank you.

veronikakurth:
@Surjasa, thanks for your interest in qutip!

 > If you want me to fix and add these changes, just let me know. 

Just a quick tip from my side: If you've already received concrete feedback (~ code review), like the one from Eric, please go ahead and try to address it in your branch without waiting for an explicit approval to do so -- unless there's a need for clarification; for this, of course, a discussion is necessary. Acting upon feedback would bring the next feedback loop closer. This would be quite a usual workflow for open-source contributions. 

Good luck!

Surjasa:
thank you @veronikakurth for the tip!

Hi @Ericgig, I have implemented your given feedback. I have removed the fixed seed regression test , and added property test and edge case tests. The fixed-seed regression test, it made the random state reproducible, the state chosen by that seed did not trigger the numerical instability, and the test passed, but in case of property test when i ran the test for random states it failed and showed assertion error. This assertion error was seen here in test line `assert dist == pytest.approx(0, abs=1e-12)`, it was because of fidelity drifting slightly below 1 for a valid self comparison. In case of A==B, the bures distance is supposed to be zero, but because of fidelity not being exactly 1, the bures distance did not return 0. That's why i used a condition for this case. Also, numerically intermediate matrix operations produced a tiny imaginary part due to floating-point noise so `np.real` is used. Edge case test is added because the Bures distance is computed from fidelity, and fidelity is numerically delicate when the state is at the boundary of the quantum-state space. The two opposite extremes pure state and maximally mixed state will trigger floating-point issues when square roots or eigenvalues are taken, as it is working with values extremely close to 0 or 1. So the edge-case test is added to ensure the implementation remains numerically stable at the extreme corners of state space.

qutip/tests/core/test_metrics.py
```diff
@@ -195,11 +195,19 @@ def test_monotonicity(self, dimension):
...
 
-
 class Test_bures_dist:
-    def test_state_with_itself(self):
-        state = rand_dm(3, seed=465)
-        assert bures_dist(state, state) == pytest.approx(0, abs=1e-6)
+    def test_state_with_itself(self, state):
+        dist = bures_dist(state, state)
+        assert np.isfinite(dist)
+        assert dist == pytest.approx(0, abs=1e-12)
```
Ericgig, line 202: Is there any situation where `np.isfinite` can be False but not `dist == pytest.approx(0, abs=1e-12)`?

qutip/tests/core/test_metrics.py
```diff
@@ -195,11 +195,19 @@ def test_monotonicity(self, dimension):
...
+
+    def test_state_with_itself_pure_and_mixed_edges(self):
+        psi = basis(2, 0)
+        rho = qeye(2) / 2
+        assert np.isfinite(bures_dist(psi, psi))
+        assert bures_dist(psi, psi) == pytest.approx(0, abs=1e-12)
+        assert np.isfinite(bures_dist(rho, rho))
+        assert bures_dist(rho, rho) == pytest.approx(0, abs=1e-12)
```
Ericgig, line 210: These are common states, not really deserving the edge cases designation.

Better edges cases would be slightly imperfect dm:
```
    trace_not_one = rand_dm(N) * (1+1e-8)
    non_hermitian = rand_dm(N) + Qobj(np.random.randn(N, N) + 1j * np.random.randn(N, N) ) * 1e-8
```
States which are the result of `mesolve` can accumulate such error and could easily cause error with metric functions. So it can be interesting to check. The default tolerance of solver is at `1e-8`, so this is a decent scale for the imperfection. 

With the `A == B` path, some tests need to avoid it to use the clamping...
```
    a = rand_dm(N)
    b = a  * (1+1e-8)
```

Ericgig (commented):
Good improvement. 
Not fixing the seed let you find more issues (small imaginary part) and test the limits of the tolerance.

But adding the `A == B` path means you now need more tests, for case that it's used and not, and all test using `bures_dist(state, state)` makes it that the clamping is never used.

ps.
I would have been happy without the `A==B` and an higher tolerance if you could explain it with some statistical analysis: Ran 10000 times, max error was X,, rounded up to Y for the test tolerance. One test function, no class.

Surjasa:
> 
> But adding the `A == B` path means you now need more tests, for case that it's used and not, and all test using `bures_dist(state, state)` makes it that the clamping is never used.
> 
> ps. I would have been happy without the `A==B` and an higher tolerance if you could explain it with some statistical analysis: Ran 10000 times, max error was X,, rounded up to Y for the test tolerance. One test function, no class.
>
@Ericgig, I ran statistical analysis for 10,000 randomly generated states. maximum self distance: 3.9891162751362981e-6
Rounded-up tolerance: approximately `4e-6`, as for 200,000 random states i got maximum: 1.0070980412017298e-5
Rounded-up tolerance: approximately `1.1e-5`. I have also noticed and checked `1e-6` tol works for the random states bures self dist test. It means the tolerance here can differ with the randomly generated states and sample size. That's why i  replaced that test with deterministic tests for identical and orthogonal states. The deterministic test always uses the same states, so the observed numerical error remains stable across runs. Its expected values are known: `bures_dist(ket0, ket0) = 0`, `bures_dist(ket0, ket1) = sqrt(2)`. Also, in case of the random test, the generated inputs vary and observed maximum error varies with it, so using deterministic tests seem fit here.

qutip/tests/core/test_metrics.py
Surjasa: no specific case was found where `np.isfinite(dist)` failed or can be False but the approximate-zero assertion passed. For NaN and infinities, both assertions fail.

qutip/tests/core/test_metrics.py
Surjasa: yes, i removed A== B path as it was avoiding using the clamp. As a better edge case, the trace-imperfect test is added to verify that bures_dist remains finite and non-negative when given a density matrix with a small trace normalization error, similar to numerical imperfections that can occur in solver output

Ericgig:
The original bug was that `bures_dist(A, A)` returned NaN for some `A`. Now it's only tested with a trivial cases once...

Don't forget pep8 and good test names.

Surjasa:
@Ericgig, I corrected the tests according to your feedback, this time i didn't keep the tolerance limit that was changing with the random states num. i used `assert dist >=0` instead.

Ericgig (approved):
New tests are ok.
Thank you.

Issue #1898, opened 2022-05-25
Update user guide content of documentation to QuTiP 5


nathanshammah:
### Problem Description

Update user guide content of documentation to QuTiP 5, which contains many changes that may not be captured by the latest versions of the documentation, e.g., the content in https://qutip.org/docs/latest/guide/guide.html. 

### Proposed Solution

- Read the documentation files on the QuTiP 5 branch, at https://github.com/qutip/qutip/tree/dev.major
- Build locally the documentation on the branch and make sure it builds
- Read the documentation files and check that they match with qutip-5 branch code, so that the documentation matches the actual existing code, similarly to what happens for QuTiP 4.7, https://qutip.org/docs/latest/guide/guide.html and locally test 
- Make changes where needed.

Anika-Roy:
Hello! is this open for contributions?

Ericgig:
We made good progress updating the documentation for v5, we are mostly missing pages for new features and porting from v4 to v5, which would be hard to write without being up to date with recent progress.
However if you want to take a look with fresh eyes, fix error, improve wording, fix broken links, etc. it would be greatly appreciated.

rawsun007:
@Ericgig — taking you up on "fix error, improve wording, fix broken links" with the mechanical half, since that part can be checked rather than judged. Results against `master` (1517dc7):

**Fixed in #2997**

- Four Sphinx cross-references naming module paths that no longer exist: `qutip.operators.sigmaz` (module removed in v5), `qutip.propagator.propagator` and `qutip.stochastic.smesolve` (both moved under `qutip.solver`), and `qutip.core.data.csr.zeroes` (the function is `zeros`). The `smesolve` one was a straggler — three other references already use the solver path.
- Four dead external links: `qutip.org/docs/latest/` (404, the `user guide` target in the internals index), `qutip.org/qutip-tutorials/index-v5.html` (404), SimulaQron's `github.io` docs page (404), and `qinfer.org` (domain no longer resolves). Replacements all probed at 200.

**Filed separately as #2998**

The internals data-layer docs carry 99 cross-references over 36 targets in 9 files, and `doc/apidoc/` documents the data layer nowhere, so those have nothing to resolve against. That needs a decision about whether the data layer belongs in the API reference, which is yours rather than mine.

**Checked and deliberately left alone**

[...]

PR #2997, opened 2026-09-01
doc: repoint stale API references and dead links

rawsun007:
Four Sphinx cross-references in the docs name module paths that no longer exist, so they render as plain literals rather than links. Related to #1898 — @Ericgig invited "fix error, improve wording, fix broken links" there, and these are the mechanical half of "documentation matches the actual existing code".

| Reference | Where | Reality |
| --- | --- | --- |
| `~qutip.operators.sigmaz` | `doc/guide/guide-measurement.rst` | `qutip/operators.py` was removed in v5; `sigmaz` is in `qutip/core/operators.py` |
| `qutip.propagator.propagator` | `doc/guide/guide-visualization.rst` | now `qutip/solver/propagator.py` |
| `qutip.stochastic.smesolve` | `doc/guide/dynamics/dynamics-stochastic.rst` | now `qutip/solver/stochastic.py` |
| `~qutip.core.data.csr.zeroes` | `doc/internals/data-layer/types.rst` | the function is `zeros` (`csr.pyx`: `cpdef CSR zeros`) |

The `smesolve` one is a straggler: three other references in the docs already use `qutip.solver.stochastic.smesolve`, so this file was missed when the rest were updated. The `zeroes` one is a plain spelling slip.

## How these were found

I resolved every `:func:`/`:class:`/`:meth:`/`:obj:` target in `doc/` against the names actually defined in the source tree (parsing the `.py` and `.pyx` files rather than importing, so the check runs against the working tree without a build). These four are the only remaining references whose leading module segment does not exist.

Deliberately untouched: `qutip.Qobj`, `qutip.QobjEvo` and `qutip.Dispatcher` are classes rather than modules, and the 44 `qutip.data.*` references resolve through the `from .core import *` re-export — my checker flags all of those, and all of them are fine.

I have not built the docs locally, so I have not confirmed how Sphinx currently renders each one; the claim here is only that the targets do not exist in the source and the replacements do.

I will add the `doc/changes/<PR>.doc` towncrier fragment once this has a number, following #2988 and #2969.

Disclosure: written with Claude Code, and the commit carries `Assisted-by: Claude Code (Opus 5)`.

## Second commit: four dead links

I also checked every external URL in `doc/` (288 unique). Four are genuinely unreachable rather than bot-blocked:

| Link | Where | Now |
| --- | --- | --- |
| `qutip.org/docs/latest/` (404) | `doc/internals/index.rst`, the `user guide` target | the current guide on readthedocs |
| `qutip.org/qutip-tutorials/index-v5.html` (404) | `doc/development/ideas/quantum-error-mitigation.rst` | the tutorials index, which resolves |
| SimulaQron's `github.io` docs page (404) | `doc/frontmatter.rst` | its repository |
| `qinfer.org` (domain no longer resolves) | `doc/frontmatter.rst` | the QInfer repository |

Every replacement was probed and returns 200.

Deliberately **not** touched: `sciencedirect.com`, `pubs.acs.org`, `doi.org`, `mpich.org` and a StackOverflow answer all return 403 to scripted requests but load fine in a browser. That is bot protection, not rot, and "fixing" them would be wrong.

Happy to split this into a separate PR if you would rather keep the API-reference fixes and the link fixes apart.

### DIFF
--- a/doc/changes/2997.doc
+++ b/doc/changes/2997.doc
@@ -1,0 +1,1 @@
+Fix four cross-references that named modules removed or renamed in v5: ``qutip.operators.sigmaz``, ``qutip.propagator.propagator``, ``qutip.stochastic.smesolve``, and the misspelled ``qutip.core.data.csr.zeroes``.
--- a/doc/changes/2997.doc
+++ b/doc/changes/2997.doc
@@ -1,1 +1,1 @@
-Fix four cross-references that named modules removed or renamed in v5: ``qutip.operators.sigmaz``, ``qutip.propagator.propagator``, ``qutip.stochastic.smesolve``, and the misspelled ``qutip.core.data.csr.zeroes``.
+Fix four cross-references that named modules removed or renamed in v5 (``qutip.operators.sigmaz``, ``qutip.propagator.propagator``, ``qutip.stochastic.smesolve``, and the misspelled ``qutip.core.data.csr.zeroes``), and four dead documentation links.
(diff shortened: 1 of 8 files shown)


Ericgig (approved):
Thank you!

PR #3000, opened 2026-09-04
Move reusable testing component to qutip.testing

Ericgig:
**Description**
When moving to a src layout, the tests will be moved out of the qutip packages.
Operator mixin classes are used in family project, so I moved them to qutip.testing.

- Moved `UnitaryMixin`, `TestAdd`, etc. to qutip.testing.
- Added a fixture for a numpy random generator. All arrays created by `_GeneralMixin` uses that generator and the seed is printed in pytest error report. The idea is to keep the randomness while making the test reproducible, so if a rare error happen, we can more easily reproduce it. Ideally I would like to use that generator for every test using random number. But too much changes for one PR.
- Reworked the tests generation script: 
  - cases creation function can be set by class.
  - Added an extra level of parametrization for non-data input: `add`'s scale, etc.
  - Move the result check in it's own method: allow classes to overwrite it.

**AI Tools Usage Disclosure**
  - Gemini: various searches and reviews.


hodgestar (approved):

qutip/testing/mixin.py
```diff
@@ -0,0 +1,1184 @@
...
+        ParameterSet at the end so that it will get correctly parametrised.
+
+    types: Tuple[Type]
+        A tuple of data.Data subclass types (not instances).  This defines the
+        inputs to this particular specialisation under test.  There should be
+        as many types as there are Data arguments to `op`.
+
+    shapes: Iterable[Tuple[(int, int) | ParameterSet[(int, int)]]]
+        An iterable containing several sets of shapes to parameterise over.
+        Each element of the iterable should be s tuple of shapes, or
```
veronikakurth: "a tuple" (little typo)

qutip/testing/mixin.py
```diff
@@ -0,0 +1,1184 @@
...
+            if out_type is not None:
+                id_ += "->" + out_type.__name__
+                args += [out_type]
+            cases.append(pytest.param(op, *args, id=id_))
+    return cases
+
+
+def build_extra_cases(extra_param):
+    """
+    Create the combination of all extra_param entrys.
```
veronikakurth: Typo: entrys -> entries

qutip/testing/mixin.py
```diff
@@ -0,0 +1,1184 @@
...
+
+    extra_param = {
+        "scale": [0, 1],
+        "alpha": [1, pytest.param(1j, id="complex")],
+    }
+    with create the 4 cases for each values of scale and alpha.
```
veronikakurth: with -> will

qutip/testing/mixin.py
```diff
@@ -0,0 +1,1184 @@
...
+    If the entrise are pytest.param, id and mask will be kept.
```
veronikakurth: entries

veronikakurth (approved):
Thank you for the neat refactoring; the entire testing ansatz of qutip looks much cleaner. I especially like the approach to test case generation, using the factory approach. 

I personally found it hard to find the mixins every time when testing functions on the data layer. Nice that they are in a visible place now...

Regarding the global seed: this is a cool idea! I hope it will help us investigate failing test cases occasionally. A bit unsure yet how to use it in practice, though.

PR #3002, opened 2026-09-15
Minimal adaptation for testing Python 3.15 compatibility

veronikakurth:
**Checklist**

**Description**

As I see it now, the only programmatic change we have to do is to avoid using `fork` as a process start method. Since Python 3.14, `forkserver` is the default method for Unix/POSIX platforms. We probably ignored the deprecation warnings about it before... 

Currently, if we stay with `fork`, we get the following tests failing

```
test_stochastic.py::test_ssesolve_method
test_stochastic.py::test_smesolve_methods
test_parallel.py::test_map
test_parallel.py::test_map_accumulator
test_parallel.py::test_map_pass_error
test_parallel.py::test_map_store_error
test_parallel.py::test_map_early_end
```

I.e., all the tests that use a parallel map.

The error is not a real error but a deprecation warning:

```
DeprecationWarning: This process (pid=31575) is multi-threaded; use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

We could also ignore the deprecation warnings, but I don't think it would be the most sustainable approach.

However, whether using `forkserver` for Linux instead of `fork` would lead to significant performance degradation -- this is something to benchmark... I guess the fact that `forkserver` works in a single-threaded way makes it safer, but slower.

Currently, the changes in `qutip/solver/parallel.py` would lead to the following OS vs process start method selection:

Linux -> forkserver
macOS -> spawn
Windows -> spawn

@Ericgig, please let me know what you think.

Some changes to workflow configs (i.e., tackling the case of not having Miniconda with Python 3.15 - which is in release candidate mode) are temporary and will be adapted after the official Python release. 

**Related issues or PRs**

**AI Tools Usage Disclosure**
  - `GPT-5.6`: for discussion of the usage of parallel map from the `multiprocessing` module and default process start methods for different platforms; for finding the way to use a release candidate version of Python in CI.


.github/workflows/tests.yml
```diff
@@ -165,23 +173,24 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - uses: conda-incubator/setup-miniconda@v4
-        if: matrix.os != 'windows-11-arm'
+        if: matrix.os != 'windows-11-arm' && !matrix.no-conda
```
Ericgig: This could be simply `!matrix.no-conda` if the flag is set in "windows-11-arm".

qutip/solver/parallel.py
```diff
@@ -15,12 +15,8 @@
 from qutip.ui.progressbar import progress_bars
 from qutip.settings import available_cpu_count
 
-if sys.platform == 'darwin':
-    mp_context = multiprocessing.get_context('fork')
-elif sys.platform == 'linux':
-    # forkserver would handle threads better, but is much slower at starting
-    # the executor and spawning tasks
-    mp_context = multiprocessing.get_context('fork')
+if sys.platform == 'linux':
```
Ericgig: The `spawn` method was causing bugs at some point, mostly when used in jupyter. See #1202.
We had to use `fork` on mac when `spawn` became the default because of that.

It was a bug with older version of python, but we have not check if it was fixed in between, so this will need to be tested locally in jupyter, (for all versions / OS if the change is not for 3.15 only.)

veronikakurth: Thank you for your feedback, Eric. It was useful to learn about the historical context behind the decisions in qutip, as well as how default [process start methods](https://docs.python.org/3/library/multiprocessing.html?utm_source=chatgpt.com#contexts-and-start-methods) evolved in the CPython standard over time...

From this, I may conclude that there are a couple of topics to investigate when deciding whether to follow the standard:

1) correctness of tests
2) performance
3) error-free parallel map runs in Jupyter notebooks

As for correctness, it looks like using a default start method (via `multiprocessing.get_context()`) for all the platforms and Python versions doesn't break anything ([CI](https://github.com/qutip/qutip/actions/runs/35062559961)). (For now, a couple of jobs are excluded because of unmerged fixes to tests, so it's not 100% confidence yet). It's also a better practice according to the official docs of Python. If the start method is overridden, it should be documented since there might be incompatibilities in case qutip is used within an application with another multiprocessing context. If we keep overriding contexts, we should document it (maybe we did and I missed it?).

As for performance, I noticed on macOS locally that tests with parallel map run slower when `spawn` is used instead of `fork`. On the other hand, in the CI, the execution time of all tests was not higher than on master. However, this is too little information to judge how performance would be affected once `fork` is no longer used. We'd need proper benchmarking there.

As for Jupyter notebooks: the issue with using `spawn` persists across platforms and different Python versions. Same error everywhere about not being able to pickle functions. I think this would be the main reason to keep using `fork` for Darwin as a compromise between safety and correctness. As well, we would need to keep `fork` for Python >= 3.14 on Linux (where `forkserver` is default) if correctness of running parallel simulations in notebooks is more important for us than thread safety. For now, I'm not particularly convinced that we should prioritise convenience over following safety standards - would not it be better to recommend running parallel simulations in regular Python scripts? When performance matters, the Jupyter notebook approach is not the best one to follow.

(Here is a [link](
https://github.com/veronikakurth/qutip-tutorials/actions/runs/35065974466) to a temporary workflow that tests Jupyter notebooks using parallel map. Qutip's version is linked to this PR's branch)

Also, I think we will not be able to fix the problem for notebooks running on Windows, though: there, only `spawn` is available... So the solution to ensure 3) will be partial.

.github/workflows/tests.yml
veronikakurth: Neat suggestion, thanks!

qutip/solver/parallel.py
Ericgig: It needs to work in a notebook: we use it in tutorials, jupyter running on a remote server (google collab, academic cluster) is common. Project using qutip also have notebook tutorials...

For heavy computation a script is preferred, but for 30 sec runs a notebook is often used. In parallel this is 2~3 sec, no issues, but 30 sec is annoying. And we will get complain that it suddenly no longer work. 

We should probably add tests running in notebook (run qutip-tutorial as part of the PR checks?)

We could have the context to use in `qutip.settings`. Detect if it's running in a notebook or not and change the default accordingly + let users overwrite it.
And / or change `"map":"parallel"` to use loky if installed and fall back on multiprocessing if not.
Adding `"map":"multiprocessing"` for explicit usage (in tests). We don't lose anything and it mostly just work for users. (loky works in notebook on windows)

Now I don't know if loky is slower that multiprocessing or not, but if the worker setup is observable, the work probably did not need to run in parallel.

`fork` is risky with cuda libraries, open file / database connections, other parallel method already running when forking. It's properly contained in our use cases. It may break with qutip-cuquantum, but it's also very inefficient there and I made a batching map to replace it.

veronikakurth: > We should probably add tests running in a notebook (run qutip-tutorial as part of the PR checks?)

I think it's a good idea. We should know if notebooks fail before merging to `master`...
For now, I added an extra job into the family packages workflow; it's not the final version (and should go in a separate PR), but it will give us the possibility to check if changes are compatible with tutorials more quickly.

P.S. I also think it would be nice to test tutorials on different Python versions. E.g., such changes to contexts will go unnoticed until the Python version for tutorials is lifted to 3.15.

veronikakurth: Thanks a lot for your feedback and for sharing more context on that. 🙌  I agree it should work in the notebooks too; those are all very valid cases...

I'm more inclined to explore the option of controlling the context via `qutip.settings`; it sounds like the most transparent one from the user's perspective, but it would be good to compare different approaches. Thank you for the ideas; I'll keep you updated.

PR #3003, opened 2026-09-15
Remove branching over mpl<=3.3, increase minimum matplotlib version

Ericgig:
**Description**
We still had branching to support matplotlib<3.3...
Remove these and increase the minimum version. (Technically we could keep from 3.7, 3.6 is hard to install without fixing the version of every dependencies, but 3.7 works with the oldest )
Also made sure installing matplotlib did not change numpy version as it caused errors in some tests.

**Related issues or PRs**
Should fix recent failures in test due to matplotlib increasing the runtime version of numpy in other PRs.

**AI Tools Usage Disclosure**


.github/workflows/tests.yml
```diff
@@ -96,7 +96,7 @@ jobs:
             qutip-optional: "[semidefinite]"
             # Installing matplotlib 3.8 with it's latest dependencies raise this warnings
             pytest-extra-options: "-W ignore::pyparsing.warnings.PyparsingDeprecationWarning"
-            pypi-extra-pkgs: "matplotlib==3.8.*"
+            pypi-extra-pkgs: "matplotlib==3.10.*"
```
veronikakurth, line 97: If we pin matplotlib to 3.10 here, the filtering of the deprecation warning is not necessary anymore (checked it by commenting out `pytest-extra-options`)

qutip/bloch.py
veronikakurth: This condition could be removed as well :) As well as the import of `parse_version`. This would be the last use of it

qutip/visualization.py
```diff
@@ -71,15 +64,15 @@ def _is_fig_and_ax(fig, ax, projection='2d'):
             if projection == '2d':
                 ax = fig.add_subplot(1, 1, 1)
             else:
-                ax = _axes3D(fig)
+                ax = fig.add_axes(Axes3D(fig, auto_add_to_figure=False))
```
veronikakurth: I spent a bit of time figuring out the status of the `auto_add_to_figure` argument. Apparently, its default value is `False` [starting from version 3.5](https://github.com/matplotlib/matplotlib/blob/c982eb789036a7f537f919ebf9d7943bfc88b2d3/doc/api/prev_api_changes/api_changes_3.4.0/behaviour.rst#axes3d-automatically-adding-itself-to-figure-is-deprecated). I think we don't need to add explicitly. I played around with the Bloch sphere in a notebook, with and without `auto_add_to_figure`: the result is the same. 

This flag is also used in `qutip/visualization.py::_is_fig_and_ax`: it can be safely removed there as well. Tests from `qutip/tests/test_visualization.py` pass locally for me in this case.

veronikakurth (approved):

Ericgig:
@veronikakurth Thank you for the review.

We still have branch for version >= 3.8 and >= 3.9. 
We used to test from 3.7 and technically 3.9 old enough to be dropped.

So we can set the minimum anywhere between 3.7 and 3.10. 
How about 3.9 to remove all branching (hoping that 3.11 won't bring new ones)?

veronikakurth:
> @veronikakurth Thank you for the review.
> 
> We still have branch for version >= 3.8 and >= 3.9. We used to test from 3.7 and technically 3.9 old enough to be dropped.
> 
> So we can set the minimum anywhere between 3.7 and 3.10. How about 3.9 to remove all branching (hoping that 3.11 won't bring new ones)?

Let’s give 3.9 a go! And see if nothing breaks. If we could remove all the branching, it would make code even more cleaner, I think it’s a good motivation for the upgrade. :) Apart from that, 3.8 has been there since 2023. It’s usually not our first argument to upgrade, but having the minimum a bit newer is also a good thing.

veronikakurth (approved):
Everything also works with Matplotlib 3.9 as the minimum! Thank you. :)

PR #3007, opened 2026-09-18
Restore the raw-string prefix on the LaTeX \times macro

### LINKED ISSUE #3004: Error in Qobj latex printing
### Bug Description

Qobj don't display correctly

Image

### Code to Reproduce the Bug

```shell

```

### Code Output

```shell

```

### Expected Behaviour

Show the time operator.

### Your Environment

```shell
master branch
```

### Additional Context

_No response_

rawsun007:
**Description**

`_latex_real` builds scientific notation with an f-string that lost its raw prefix:

```python
return base + f"\times10^{{ {int(exp):d} }}"
```

`\t` is a tab, so the macro reaches LaTeX as a tab followed by `imes` and renders as a stray `imes`, exactly as in the screenshot on #3004. Restoring the `r` fixes it:

```python
return base + rf"\times10^{{ {int(exp):d} }}"
```

Built from `6f8df7f`, before and after:

```
>>> _latex_real(-1.115e-6)
'-1.115\times10^{ -6 }'     # a tab, then "imes"
'-1.115\\times10^{ -6 }'    # after
```

Only values outside `0.001 <= abs(x) < 1000` reach that branch, which is why a ket of small amplitudes is affected and ordinary operators are not. 5.3.1 is unaffected — it still had `r"\times10^{{ {:d} }}".format(int(exp))`; the prefix was lost converting to an f-string in 1e9cdb8e.

Also adds the first tests over this path: `_latex_real` across the scientific, plain, integral and zero branches, and a `_repr_latex_` case asserting `\times` is present and no literal tab is. There was no test touching `_repr_latex_` or `_latex_real` anywhere under `qutip/tests/`, which is why a one-character change slipped through. Reverting only `qobj.py` fails 4 of the 7.

`qutip/tests/core/test_qobj.py` is 216 passed. `pycodestyle` clean on the changed lines.

**Related issues or PRs**

fixes #3004

**AI Tools Usage Disclosure**
  - `Claude Opus 5 (Claude Code): diagnosed the issue from the screenshot, bisected the regression to 1e9cdb8e, wrote the one-character fix and the tests, and ran them against a local build of master.`


veronikakurth (approved):
This works, thank you!

Image

rawsun007:
The red `qutip-jax` job is not from this change. It fails at collection with 29 errors, all `KeyError: <class 'qutip_jax.jaxarray.JaxArray'>`, which is a data-layer registration mismatch between master and the plugin.

The same run on `Ericgig-patch-1` fails with the identical error, as do `feature.ULME` and `test_split_2`.

This PR is a raw-string prefix on a LaTeX macro plus tests for it; `qutip/tests/core/test_qobj.py` is 216 passing.

